PR #107, opened 2017-03-12
[wip] Live stream chat

justinswart:
### What

This adds chat functionality to live streams. It's a WIP but I'm going to be on a ✈️ for the next two days so it'd be nice to get some 👀 on this!

### Apologies

For the size of this. It was meant to be three separate PR's but some design changes along the way and the resultant refactoring just made working on 3 separate branches too complicated so I resolved conflicts and merged back into one.

### Things to review
- In order to auth chat we hit a KSR api endpoint for a hash which we then pass to the Live api when fetching the `LiveStreamEvent`. This then adds additional values to the `LiveStreamEvent` which we use for chat auth to Firebase.
  - This adds `liveAuthToken` to the `AppEnvironment` for persistence between sessions so that it's not necessary to fetch each time.
- `LiveStreamContainerViewController` was refactored to be a container of the `LiveStreamViewController` positioned at the top, then a `LiveStreamContainerPageViewController` which contains a tab strip switching between `LiveStreamEventDetailsViewController` and `LiveStreamChatViewController`.
  - Previously this entire view was scrollable but this made interacting with the chat table view difficult so the container view controller is now static with the paged children view controllers being scrollable.
- `LiveStreamViewController` still houses all of the Firebase listeners and thus handles all of the observation of chat messages and writing of messages back to Firebase. I've stuck this behind a protocol `LiveStreamChatHandler` in anticipation that we may want to move it to the `LiveStreamService` in order for us to test it with the `AppEnvironment`. I'll chat through this with @mbrandonw (this could also be a separate PR).
- Due to the way in which Firebase initially loads (hundreds of new message events may occur) we buffer these for 500ms and reload the chat table view if an amount of messages above a threshold is received in that time. Alternatively we animate the new messages in. This buffering logic is in `LiveStreamViewModel` and the reloading in `LiveStreamChatViewModel`.
- I'll add some in-line comments for the rest.

### Still needed
- Koala tracking (separate PR when I get a list of events for tracking).
- VC tests. I temporarily commented out VC tests for `LiveStreamContainerViewController` to get a ✅ until we can fix them. Having trouble getting the chat input view to layout correctly and also need to create a means to inject the `LiveStreamChatHandler` for testing.
- Provide `LiveStreamChatHandler` or similar via AppEnvironment.

### What does it look like?

screen shot 2017-03-12 at 6 56 13 pm
screen shot 2017-03-12 at 6 44 29 pm


LiveStream/ViewModels/LiveStreamViewModel.swift
```diff
@@ -396,3 +528,60 @@ private func startDateMoreThanFifteenMinutesAgo(event: LiveStreamEvent) -> Bool
 private func didEndNormally(event: LiveStreamEvent) -> Bool {
   return !event.liveNow && event.definitelyHasReplay
 }
+
+private enum TimeIntervalBufferEvent<T> {
```
justinswart: This is a pretty cool state machine approach that someone in the ReactiveSwift slack channel suggested to use for buffering on an interval. I just made a few small changes for our purposes. It may be overkill and perhaps we'd prefer that this is an operator in our ReactiveExtensions but it works!

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+
+      return vc
+  }
+
+  internal override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.tableView.dataSource = self.dataSource
+
+    self.tableView.transform = CGAffineTransform(scaleX: 1, y: -1)
```
justinswart: I know this is a cheeky hack but it really simplifies an otherwise annoying problem with table views! By flipping the table view on its head and inversely flipping all of its cells we get the following for free:

- Automatically pins the table view to the bottom for new messages being inserted at row `0`.
- Animates the table view up with the keyboard appearance.
- Keeps the `contentOffset` on rotation.
- It's really simple.

The alternative (and my prior attempt) was a bunch of inputs and outputs to the view model handling all of these events and scrolling to the bottom on new messages, etc. but it would sometimes get the offset wrong after rotation or not animate smoothly with the keyboard appearance, etc.

I welcome other/better approaches!

Library/DataSource/ValueCellDataSource.swift
```diff
@@ -71,21 +71,45 @@ open class ValueCellDataSource: NSObject, UICollectionViewDataSource, UITableVie
...
+   Inserts a single value at the beginning of the section specified.
+
+   - parameter value:     A value to prepend.
+   - parameter cellClass: The type of cell associated with the value.
+   - parameter section:   The section to append the value to.
+
+   - returns: The index path of the prepended row.
+   */
+  @discardableResult
+  public final func prependRow <
```
justinswart: Added `prependRow` for inserting the chat messages at the beginning each time. Also for animating the rows in we require the `IndexPath`. Despite this always being `row: 0, section: 0` figured it's best to return it here and on `appendRow` for consistency.

LiveStream/ViewModels/LiveStreamViewModel.swift
```diff
@@ -301,18 +329,106 @@ internal final class LiveStreamViewModel: LiveStreamViewModelType, LiveStreamVie
...
+    }.take(during: Lifetime(self.token))
+
+    let snapshots = self.viewDidLoadProperty.signal
+      .flatMap { [snapshot = self.receivedChatMessageSnapshotProperty.producer] in
+        snapshot
+          .start(on: environment.backgroundQueueScheduler)
+          .skipNil()
+    }
+
+    self.chatMessages = Signal.merge(
```
justinswart: So this signal buffers the chat messages and decodes the arrays of Firebase snapshots into arrays of `LiveStreamChatMessage`.

Feels like it would be useful to try and move this to the `LiveStreamService` so that it can be mocked in the `AppEnvironment`, however seeing as the service is a singleton I'm just wondering how we'd do that whilst keeping this signal clearly `LiveStreamEvent`-specific.

Kickstarter-iOS/Tests/Views/LiveStreamContainerViewControllerTests.swift
```diff
@@ -18,193 +18,193 @@ internal final class LiveStreamContainerViewControllerTests: TestCase {
     UIView.setAnimationsEnabled(true)
     super.tearDown()
   }
-
-  func testReplay() {
```
justinswart: Just temporary until we can fix these!

Library/ViewModels/LiveStreamChatViewModel.swift
```diff
@@ -0,0 +1,223 @@
...
+    self.userSessionStartedProperty.value = ()
+  }
+
+  public let chatInputViewHidden: Signal<Bool, NoError>
+  public let configureChatHandlerWithUserInfo: Signal<LiveStreamChatUserInfo, NoError>
+  public let dismissKeyboard: Signal<(), NoError>
+  public let didAuthorizeChat: Signal<Bool, NoError>
+  public let openLoginToutViewController: Signal<LoginIntent, NoError>
+  public let presentMoreMenuViewController: Signal<(LiveStreamEvent, Bool), NoError>
+  public let prependChatMessagesToDataSource: Signal<([LiveStreamChatMessage], Bool), NoError>
```
mbrandonw: could either create a custom struct for this tuple to describe what the `Bool` is for, or could call the output `prependChatMessagesToDataSourceAndReload`

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
```diff
@@ -0,0 +1,140 @@
...
+    super.bindViewModel()
+
+    self.moreButton.addTarget(self, action: #selector(LiveStreamChatInputView.more), for: .touchUpInside)
+    self.sendButton.addTarget(self, action: #selector(LiveStreamChatInputView.send), for: .touchUpInside)
+
+    self.moreButton.rac.hidden = self.viewModel.outputs.moreButtonHidden
+    self.sendButton.rac.hidden = self.viewModel.outputs.sendButtonHidden
+
+    self.viewModel.outputs.notifyDelegateMoreButtonTapped.observeValues { [weak self] in
+      self.flatMap { $0.delegate?.liveStreamChatInputViewDidTapMoreButton(chatInputView: $0) }
```
mbrandonw: more uses of `flatMap` for side effects in this file. can we correct them all to `doIfSome` or `if let`?

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
```diff
@@ -0,0 +1,140 @@
...
+    self.viewModel.outputs.notifyDelegateRequestLogin
+      .observeForControllerAction()
+      .observeValues { [weak self] in
+        self.flatMap { $0.delegate?.liveStreamChatInputViewRequestedLogin(chatInputView: $0) }
+    }
+
+    self.viewModel.outputs.placeholderText
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.textField.attributedPlaceholder = $0
```
mbrandonw: we could make a reactive extension for this

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
```diff
@@ -0,0 +1,140 @@
...
+    self.viewModel.outputs.notifyDelegateMoreButtonTapped.observeValues { [weak self] in
+      self.flatMap { $0.delegate?.liveStreamChatInputViewDidTapMoreButton(chatInputView: $0) }
+    }
+
+    self.viewModel.outputs.notifyDelegateMessageSent
+      .observeForUI()
+      .on(value: { [weak self] _ in
+        self?.textField.text = nil
+        self?.textField.resignFirstResponder()
+      })
```
mbrandonw: let's not inject side effects like this. should this instead be a separate output?

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    _ = self.moreButton
+      |> UIButton.lens.tintColor .~ .white
+  }
+
+  internal override func bindViewModel() {
+    super.bindViewModel()
+
+    self.moreButton.addTarget(self, action: #selector(LiveStreamChatInputView.more), for: .touchUpInside)
+    self.sendButton.addTarget(self, action: #selector(LiveStreamChatInputView.send), for: .touchUpInside)
```
mbrandonw: we prefer to keep only vm binding stuff in `bindViewModel`. any other initialization stuff should be in an initializer, `viewDidLoad`, `awakeFromNib`, and other one-time initialization endpoints

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+
+  private func openLoginTout() {
+    let vc = LoginToutViewController.configuredWith(loginIntent: .liveStreamSubscribe)
+    let nav = UINavigationController(rootViewController: vc)
+    nav.modalPresentationStyle = .formSheet
+
+    self.present(nav, animated: true, completion: nil)
+  }
+
+  @objc private func subscribe() {
```
mbrandonw: this should  be `subscribeButtonTapped`, and same for `goToProjectTapped`

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+      |> UIButton.lens.titleColor(forState: .normal) .~ .white
+      |> liveStreamGoToProjectStyle
+  }
+
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.viewModel.outputs.openLoginToutViewController
+      .observeValues { [weak self] _ in
+        self?.openLoginTout()
```
mbrandonw: this should be observed for controller action right?

Library/Keyboard.swift
```diff
@@ -35,7 +36,8 @@ public final class Keyboard {
     self.changeObserver.send(value: (
       frame.cgRectValue,
       duration.doubleValue,
-      UIViewAnimationOptions(rawValue: UInt(curve.rawValue))
+      UIViewAnimationOptions(rawValue: UInt(curve.rawValue)),
+      notificationName: notification.name
```
mbrandonw: cool, would love to remove if not needed

justinswart:
Great, thanks for going through all of that, let me make changes!

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuCancelCell.swift
```diff
@@ -0,0 +1,34 @@
...
+import Prelude
+
+internal final class LiveStreamContainerMoreMenuCancelCell: UITableViewCell, ValueCell {
+
+  @IBOutlet private weak var titleLabel: UILabel!
+
+  internal func configureWith(value moreMenuItem: LiveStreamContainerMoreMenuItem) {
+    self.titleLabel.text = localizedString(key: "Cancel", defaultValue: "Cancel")
+
+    self.selectedBackgroundView = self.selectionView
```
justinswart: It's the darker grey background of the more menu cell:

screen shot 2017-03-15 at 12 23 14 pm

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuCancelCell.swift
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  internal override func bindStyles() {
+    super.bindStyles()
+
+    _ = self
+      |> UITableViewCell.lens.backgroundColor .~ .hex(0x1B1B1C)
```
justinswart: It's the darker grey background of the more menu cell:

screen shot 2017-03-15 at 12 23 14 pm

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuCancelCell.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    self.separatorInset = UIEdgeInsets(leftRight: self.frame.size.width)
```
justinswart: Yip, it hides the separator only on the cancel cell.

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuIconTextCell.swift
```diff
@@ -0,0 +1,70 @@
...
+  @IBOutlet private weak var contentStackView: UIStackView!
+  @IBOutlet private weak var iconImageView: UIImageView!
+  @IBOutlet private weak var rootStackView: UIStackView!
+  @IBOutlet private weak var subtitleLabel: UILabel!
+  @IBOutlet private weak var titleLabel: UILabel!
+
+  internal func configureWith(value moreMenuItem: LiveStreamContainerMoreMenuItem) {
+    switch moreMenuItem {
+    case .hideChat(let hidden):
+      self.iconImageView.image = UIImage(named: "speech-icon")
```
justinswart: 👍

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuIconTextCell.swift
```diff
@@ -0,0 +1,70 @@
...
+      self.subtitleLabel.text = hidden
+        ? localizedString(key: "Show_chat", defaultValue: "Show chat")
+        : localizedString(key: "Hide_chat", defaultValue: "Hide chat")
+    case .share:
+      self.iconImageView.image = UIImage(named: "share-icon")
+      self.titleLabel.isHidden = true
+      self.subtitleLabel.text = localizedString(key: "Share_live_stream",
+                                                defaultValue: "Share live stream")
+    default:
+      break
```
justinswart: Sure, but any other type should not be using this cell so I'll do something like this?

```
case .cancel:
      assertionFailure("Invalid type for this cell class.")
```

Kickstarter-iOS/DataSources/LiveStreamContainerPagesDataSource.swift
```diff
@@ -0,0 +1,47 @@
+import KsApi
+import Library
+
+internal final class LiveStreamContainerPagesDataSource: NSObject {
+  fileprivate var viewControllers: [UIViewController]?
```
justinswart: Yip, fixed!

Kickstarter-iOS/Views/Controllers/LiveStreamContainerMoreMenuViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+
+  internal override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.tableView.dataSource = self.dataSource
+    self.tableView.isScrollEnabled = false
+    self.automaticallyAdjustsScrollViewInsets = false
+
+    let emptyFooterView = UIView(frame: CGRect())
+    self.tableView.tableFooterView = emptyFooterView
```
justinswart: Empty footer just ensures that we won't see any trailing separators at the bottom of the table view, but seeing as it's not a scrollable table view I'll remove it.

Kickstarter-iOS/Views/Controllers/LiveStreamContainerMoreMenuViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+    if let menuItem = self.dataSource[indexPath] as? LiveStreamContainerMoreMenuItem {
+      switch menuItem {
+      case .cancel:
+        self.delegate?.moreMenuViewControllerWillDismiss(controller: self)
+      case .share(let liveStreamEvent):
+        self.delegate?.moreMenuViewControllerDidShare(controller: self, liveStreamEvent: liveStreamEvent)
+      case .hideChat(let hidden):
+        self.delegate?.moreMenuViewControllerDidSetChatHidden(controller: self, hidden: !hidden)
+      default:
+        return
```
justinswart: Yeah turns out this one was already exhaustive, removed it.

Kickstarter-iOS/Views/Controllers/LiveStreamContainerMoreMenuViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+        CGSize(width: self.tableView.frame.size.width, height: 9999)).height
+    })
+
+    self.tableView.contentInset = UIEdgeInsets(
+      top: self.tableView.frame.size.height - contentHeight
+    )
+  }
+
+  internal override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+    if let menuItem = self.dataSource[indexPath] as? LiveStreamContainerMoreMenuItem {
```
justinswart: 👍

Kickstarter-iOS/Views/Controllers/LiveStreamContainerMoreMenuViewController.swift
```diff
@@ -0,0 +1,102 @@
...
+      case .share(let liveStreamEvent):
+        self.delegate?.moreMenuViewControllerDidShare(controller: self, liveStreamEvent: liveStreamEvent)
+      case .hideChat(let hidden):
+        self.delegate?.moreMenuViewControllerDidSetChatHidden(controller: self, hidden: !hidden)
+      default:
+        return
+      }
+    }
+
+    tableView.indexPathForSelectedRow.flatMap { tableView.deselectRow(at: $0, animated: true) }
```
justinswart: 👍

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
justinswart: woops, indeed

Kickstarter-iOS/Views/Controllers/LiveStreamContainerPageViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+        case .info:
+          self?.liveStreamEventDetailsViewController.flatMap {
+            self?.pageViewController?.setViewControllers([$0], direction: direction, animated: true,
+                                                        completion: nil)
+          }
+        }
+    }
+  }
+
+  @objc private func info() {
```
justinswart: Guess I've always just named actions that way from [these](http://nshipster.com/ibaction-iboutlet-iboutletcollection/) suggestions and that we're being more explicit in our view model and tests about exactly what 'event' took place. But all good! updated.

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+    self.subscribeButton.rac.accessibilityHint
+      = self.viewModel.outputs.subscribeButtonAccessibilityHint
+    self.subscribeButton.rac.accessibilityLabel
+      = self.viewModel.outputs.subscribeButtonAccessibilityLabel
+
+    self.viewModel.outputs.subscribeLabelHidden
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.subscribeLabel.alpha = $0 ? 0 : 1
+    }
```
justinswart: I'm guessing I needed it to still occupy the same space but not be shown, I'll test and make sure.

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+      .animateSubscribeButtonActivityIndicator
+
+    self.subscribeButton.rac.hidden = self.viewModel.outputs
+      .animateSubscribeButtonActivityIndicator
+
+    self.goToProjectButtonContainerView.rac.hidden = self.viewModel.outputs.goToProjectButtonContainerHidden
+
+    Signal.merge(
+      self.viewModel.outputs.showErrorAlert,
+      self.viewModel.outputs.showErrorAlert
```
justinswart: Yip that's a mistake!

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
justinswart: 👍

Library/AppEnvironment.swift
```diff
@@ -114,6 +121,7 @@ public struct AppEnvironment {
     koala: Koala = AppEnvironment.current.koala,
     language: Language = AppEnvironment.current.language,
     launchedCountries: LaunchedCountries = AppEnvironment.current.launchedCountries,
+    liveAuthToken: String? = AppEnvironment.current.liveAuthToken,
```
justinswart: Ok, this will change along with https://github.com/kickstarter/kickstarter/pull/6045

Library/ViewModels/LiveStreamContainerMoreMenuViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+  public let loadDataSource: Signal<[LiveStreamContainerMoreMenuItem], NoError>
+
+  public var inputs: LiveStreamContainerMoreMenuViewModelInputs { return self }
+  public var outputs: LiveStreamContainerMoreMenuViewModelOutputs { return self }
+}
+
+public enum LiveStreamContainerMoreMenuItem {
+  case hideChat(hidden: Bool)
+  case share(liveStreamEvent: LiveStreamEvent)
+  case cancel
```
justinswart: 👍

Kickstarter-iOS/Views/LiveStreamNavTitleView.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    guard let superview = self.superview else { return }
+
+    let newOrigin = CGPoint(x: (superview.frame.size.width / 2) - (stackViewSize.width / 2),
+                            y: self.frame.origin.y)
+
+    self.frame = CGRect(
+      origin: newOrigin,
+      size: CGSize(width: stackViewSize.width, height: Styles.grid(5))
+    )
```
justinswart: Can't remember, had issues with this, I'll move it back to parent.

Library/ViewModels/LiveStreamContainerMoreMenuViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+  public static func == (lhs: LiveStreamContainerMoreMenuItem, rhs: LiveStreamContainerMoreMenuItem) -> Bool {
+    switch (lhs, rhs) {
+    case (.hideChat(let lhsHidden), .hideChat(let rhsHidden)):
+      return lhsHidden == rhsHidden
+    case (.share(let lhsLiveStreamEvent), .share(let rhsLiveStreamEvent)):
+      return lhsLiveStreamEvent == rhsLiveStreamEvent
+    case (.cancel, .cancel):
+      return true
+    default:
+      return false
```
justinswart: Is there a nice way of doing that without having just stacks of combinations being tested?

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuIconTextCell.swift
mbrandonw: yeah, if `.cancel` happening means something really bad happened then an assertion failure would be good to make it complain loudly

Kickstarter-iOS/Views/Controllers/LiveStreamEventDetailsViewController.swift
mbrandonw: cool, if that's the case then might be worth making the outlet describe the alpha instead of the hidden state.

Kickstarter-iOS/Views/LiveStreamChatInputView.swift
justinswart: it should be! 👍

justinswart: i'd like to think i was testing something and forgot to make it an output 😊

Library/Keyboard.swift
justinswart: Yeah for some reason the height doesn't change, which is why I had to inspect the notification name:

```
Name(_rawValue: UIKeyboardWillShowNotification) frame: (0.0, 409.0, 375.0, 258.0)
Name(_rawValue: UIKeyboardWillHideNotification) frame: (0.0, 667.0, 375.0, 258.0)
```

Also - we should see who this behaves with [custom keyboards](http://stackoverflow.com/a/26004605) 😕

Library/DataSource/ValueCellDataSource.swift
justinswart: If we're happy to return the discardable `IndexPath` here, should we add it to the other functions for consistency?

Kickstarter-iOS/Views/Cells/LiveStreamChatMessageCell.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    _ = self.rootStackView
+      |> UIStackView.lens.alignment .~ .top
+      |> UIStackView.lens.spacing .~ Styles.grid(1)
+      |> UIStackView.lens.layoutMargins .~ .init(topBottom: CGFloat(Styles.grid(2)) / 2,
+                                                 leftRight: Styles.grid(2))
+      |> UIStackView.lens.layoutMarginsRelativeArrangement .~ true
+
+    _ = self.creatorTextLabel
+      |> UILabel.lens.text .~ localizedString(key: "Creator", defaultValue: "Creator")
```
justinswart: Opened yesterday, still waiting to confirm if we need all the strings before adding them: https://github.com/kickstarter/kickstarter/pull/6207

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    self.viewModel.outputs.prependChatMessagesToDataSourceAndReload
+      .observeForUI()
+      .observeValues { [weak self] chatMessages, reload in
+        let indexPaths = self?.dataSource.add(chatMessages, toSection: Section.messages.rawValue)
+        indexPaths.doIfSome { self?.insert($0, andReload: reload) }
+    }
+
+    self.viewModel.outputs.openLoginToutViewController
```
justinswart: I did a search and there's not much consistency on this but I'll rename

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuIconTextCell.swift
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  internal func configureWith(value moreMenuItem: LiveStreamContainerMoreMenuItem) {
+    switch moreMenuItem {
+    case .hideChat(let hidden):
+      self.iconImageView.image = Library.image(named: "speech-icon")
+      self.titleLabel.isHidden = true
+      self.subtitleLabel.text = hidden
+        ? localizedString(key: "Show_chat", defaultValue: "Show chat")
+        : localizedString(key: "Hide_chat", defaultValue: "Hide chat")
```
justinswart: https://github.com/kickstarter/kickstarter/pull/6207

Kickstarter-iOS/Views/Controllers/LiveStreamContainerMoreMenuViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+internal protocol LiveStreamContainerMoreMenuViewControllerDelegate: class {
+  func moreMenuViewControllerWillDismiss(controller: LiveStreamContainerMoreMenuViewController)
+  func moreMenuViewControllerDidShare(controller: LiveStreamContainerMoreMenuViewController,
+                                      liveStreamEvent: LiveStreamEvent)
+  func moreMenuViewControllerDidSetChatHidden(controller: LiveStreamContainerMoreMenuViewController,
+                                              hidden: Bool)
+}
+
+//swiftlint:disable:next type_name
+internal final class LiveStreamContainerMoreMenuViewController: UITableViewController {
```
justinswart: Yip, sent an email about this, I think we may just be able to get rid of it.

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuIconTextCell.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    self.selectedBackgroundView = UIView()
+      |> UIView.lens.backgroundColor .~ .ksr_navy_700
+  }
+
+  internal func configureWith(value moreMenuItem: LiveStreamContainerMoreMenuItem) {
+    switch moreMenuItem {
+    case .hideChat(let hidden):
+      self.iconImageView.image = Library.image(named: "speech-icon")
+      self.titleLabel.isHidden = true
```
justinswart: Probably something I needed before everything changed. Will fix if we are keeping these cells.

Kickstarter-iOS/Views/Cells/LiveStreamContainerMoreMenuCancelCell.swift
```diff
@@ -0,0 +1,33 @@
...
+
+  internal override func awakeFromNib() {
+    super.awakeFromNib()
+
+    self.selectedBackgroundView = UIView()
+      |> UIView.lens.backgroundColor .~ .ksr_navy_700
+  }
+
+  internal func configureWith(value moreMenuItem: LiveStreamContainerMoreMenuItem) {
+    self.titleLabel.text = localizedString(key: "Cancel", defaultValue: "Cancel")
```
justinswart: Yip, we won't need this if the menu goes.

Kickstarter-iOS/DataSources/LiveStreamContainerMoreMenuDataSource.swift
```diff
@@ -0,0 +1,37 @@
+import Library
+import LiveStream
+import Prelude
+import UIKit
+
+fileprivate enum Section: Int {
```
justinswart: This one is inside the data source?

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+    self.dismiss(animated: true) {
+      self.viewModel.inputs.didSetChatHidden(hidden: hidden)
+    }
+  }
+
+  func moreMenuViewControllerDidShare(controller: LiveStreamContainerMoreMenuViewController,
+                                      liveStreamEvent: LiveStreamEvent) {
+    self.delegate?.liveStreamChatViewController(self, willDismissMoreMenuViewController: controller)
+    self.dismiss(animated: true) {
+      self.shareViewModel.inputs.shareButtonTapped()
```
justinswart: Will wait to see if we're keeping this 👍

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+    self.viewModel.outputs.openLoginToutViewController
+      .observeForControllerAction()
+      .observeValues { [weak self] in
+        self?.openLoginTout(loginIntent: $0)
+    }
+
+    Keyboard.change
+      .observeForUI()
+      .observeValues { [weak self] change in
+        if change.notificationName == .UIKeyboardWillShow {
```
justinswart: Yeah, it works in the `CommentDialogViewController` to only observe `change.frame.minY` because it's only needed once when the view appears. Upon sending the modal is dismissed so there isn't any animation of the input field back to the bottom of the screen in that view.

All I get observing `change.frame.minY` is:
```
minY: 409.0
minY: 667.0
```

Ideally we'd want to use an input accessory view for this but it's weird when changing tabs for the input field to appear and disappear.

For now this is the best solution I could come up with.

Kickstarter-iOS/DataSources/LiveStreamContainerMoreMenuDataSource.swift
mbrandonw: i mean scoped inside the class:

```
class DataSource {
  enum Section {
```

reason being is that we typically scope `Section` to be `internal` so that it can be tested, and so then we need to namespace `Section` so that there aren't conflicts.

Kickstarter-iOS/Views/Cells/LiveStreamChatMessageCell.swift
```diff
@@ -0,0 +1,80 @@
+import Library
+import LiveStream
+import Prelude
+
+internal final class LiveStreamChatMessageCell: UITableViewCell, ValueCell {
+  private let viewModel: LiveStreamChatMessageCellViewModelType = LiveStreamChatMessageCellViewModel()
+
+  @IBOutlet private weak var avatarImageView: UIImageView!
```
justinswart: This is updated in the Storyboard 👍

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,268 @@
...
+    super.bindViewModel()
+
+    NotificationCenter.default
+      .addObserver(forName: .ksr_sessionStarted, object: nil, queue: nil) { [weak self] _ in
+        self?.viewModel.inputs.userSessionStarted()
+    }
+
+    NotificationCenter.default
+      .addObserver(forName: .UIDeviceOrientationDidChange, object: nil, queue: nil) { [weak self] _ in
+        self?.viewModel.inputs.deviceOrientationDidChange(
```
justinswart: Have moved these, not sure why this isn't collapsing.

Library/ViewModels/LiveStreamContainerPageViewModelTests.swift
```diff
@@ -0,0 +1,174 @@
...
+    self.pagedToPage.assertValueCount(0)
+    self.pagedToPageDirection.assertValueCount(0)
+
+    self.vm.inputs.configureWith(project: .template, liveStreamEvent: .template,
+                                 refTag: .projectPage, presentedFromProject: true)
+    self.vm.inputs.viewDidLoad()
+
+    self.vm.inputs.didLoadViewControllersIntoPagesDataSource()
+
+    XCTAssertTrue(self.pagedToPage.values[0].isInfoPage)
```
mbrandonw: we shouldn't reach into the test observers `values` like this. in fact it should be private.

instead we should construct the values we want to assert against.

Library/ViewModels/LiveStreamContainerPageViewModelTests.swift
```diff
@@ -0,0 +1,174 @@
...
+
+  func testLoadViewControllersIntoPagesDataSource() {
+    self.loadViewControllersIntoPagesDataSource.assertValueCount(0)
+
+    self.vm.inputs.configureWith(project: .template, liveStreamEvent: .template,
+                                 refTag: .projectPage, presentedFromProject: true)
+    self.vm.inputs.viewDidLoad()
+
+    self.loadViewControllersIntoPagesDataSource.assertValueCount(1)
+    XCTAssertTrue(self.loadViewControllersIntoPagesDataSource.values[0][0].isInfoPage)
```
mbrandonw: same here

Library/Koala/Koala.swift
```diff
@@ -1657,6 +1657,24 @@ public final class Koala {
...
+      .withAllValuesFrom(properties(liveStreamEvent: liveStreamEvent))
+      .withAllValuesFrom(
+        [
+          "context": stateContext(forLiveStreamEvent: liveStreamEvent).trackingString,
+          "error": error
+        ]
+    )
+
+    self.track(
+      event: "Live API Error",
```
justinswart: We can remove that but there isn't anything else to do with these errors, nothing user-actionable so I didn't want to show an alert but I figured we may want to know if a lot of these are happening over time. Not sure if it is trackable from api side.

PR card added for other events.

Kickstarter-iOS/Views/Controllers/LiveStreamChatViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  // MARK: Actions
+
+  @objc private func sendButtonTapped() {
+    self.viewModel.inputs.sendButtonTapped()
+  }
+
+  @objc private func textFieldChanged(_ textField: UITextField) {
+    textField.text.doIfSome { self.viewModel.inputs.textDidChange(toText: $0) }
```
justinswart: 👍

Kickstarter-iOS/Views/LiveStreamNavTitleView.swift
```diff
@@ -0,0 +1,96 @@
...
+  @IBOutlet private weak var numberOfPeopleWatchingStackView: UIStackView!
+
+  private weak var delegate: LiveStreamNavTitleViewDelegate?
+  private let viewModel: LiveStreamNavTitleViewModelType = LiveStreamNavTitleViewModel()
+
+  internal class func fromNib() -> LiveStreamNavTitleView {
+    return UINib(nibName: Nib.LiveStreamNavTitleView.rawValue, bundle: .framework)
+      .instantiate(withOwner: nil, options: nil)
+      //swiftlint:disable:next force_cast
+      .first as! LiveStreamNavTitleView
```
justinswart: Think we should assert here if this fails, it's kind've similar to our `configureCell` not finding a registered cell type?

Kickstarter-iOS/DataSources/LiveStreamContainerPagesDataSource.swift
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+  }
+
+  internal func index(forController controller: UIViewController) -> Int? {
+    return self.viewControllers.index(of: controller)
+  }
+
+  internal func controller(forPage page: LiveStreamContainerPage) -> UIViewController? {
+    return self.pages.index(of: page).flatMap { self.viewControllers[$0] }
```
justinswart: Aah yes I think was optional before, will update.

Kickstarter-iOS/Views/LiveStreamNavTitleView.swift
mbrandonw: either `as!` or `fatalError` is fine as long as a test will catch if something changed. is there a screenshot test covering this, even if indirectly through testing some other interface?

justinswart: No screenshot test because this is on the nav bar but I'll see about testing it as just a subview

Library/ViewModels/LiveStreamContainerViewModel.swift
```diff
@@ -124,202 +124,342 @@ LiveStreamContainerViewModelInputs, LiveStreamContainerViewModelOutputs {
...
 
-    self.navBarLiveDotImageViewHidden = hideWhenReplay
-    self.creatorAvatarLiveDotImageViewHidden = hideWhenReplay
-    self.numberWatchingBadgeViewHidden = hideWhenReplay
-    self.availableForLabelHidden = Signal.combineLatest(nonStarter, hideWhenLive).map { $0 || $1 }
+    let nonStarterState = updatedEventFetch.values()
+      .map { event in isNonStarter(event: event) }
+      .filter(isTrue)
+
+    let signInAnonymouslyEvent = Signal.merge(
```
justinswart: So as we spoke about (moved here for discussion and removed the fixme). The code from line 336 - 376 does the following:

- Signs into Firebase anonymously if `currentUser` is nil. This is to still update the number of people watching value anonymously.
- Signs in with our custom token if it's available.
- Observes the values from the producer to keep it alive as we don't do anything else with it.

Our discussion was how to test this and whether or not we want some kind of output for that.

justinswart:
In testing sign-in I noticed that [this](https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/ViewModels/RootViewModel.swift#L108) output i causing the `RootTabBarViewController` to reload all of its view controllers when the user session starts. This is a problem for chat because the chat and live stream disappears after sign-in.

Any ideas on how we can handle this?

Library/ViewModels/LiveStreamContainerViewModel.swift
```diff
@@ -124,202 +123,291 @@ LiveStreamContainerViewModelInputs, LiveStreamContainerViewModelOutputs {
...
-    self.availableForLabelHidden = Signal.combineLatest(nonStarter, hideWhenLive).map { $0 || $1 }
+    let nonStarterState = updatedEventFetch.values()
+      .map { event in isNonStarter(event: event) }
+      .filter(isTrue)
 
-    self.goToProject = configData
-      .takeWhen(self.goToProjectButtonPressedProperty.signal)
-      .map { project, liveStreamEvent, _, _ in
-        (project, liveStreamEvent.liveNow ? .liveStream : .liveStreamReplay)
+    let firebaseUserId = updatedEventFetch.values()
```
justinswart: So as we spoke about (moved here for discussion and removed the fixme). The code from line 302 - 324 does the following:

- Signs into Firebase to update the number of people watching value.
- If there is no token on the event object it signs in anonymously with an anonymous user ID
- If there is a token it authenticates as that user and returns their respective user ID.
- Observes the values from the producer to keep it alive as we don't do anything else with it.

Our discussion was how to test this and whether or not we want some kind of output for that. We aren't really too concerned if this errors out.

mbrandonw:
> In testing sign-in I noticed that this output is causing the RootTabBarViewController to reload all of its view controllers when the user session starts. This is a problem for chat because the chat and live stream disappears after sign-in.
> 
> Any ideas on how we can handle this?

i'm not sure why that would happen with chat. we do something similar for the pledge flow. the reward screen is a modal presented over the project page and you can login during the flow.

can you think of something different about those two use cases?

justinswart:
> i'm not sure why that would happen with chat. we do something similar for the pledge flow. the reward screen is a modal presented over the project page and you can login during the flow.
> 
> can you think of something different about those two use cases?

Oh, interesting, let me investigate a bit more

PR #26, opened 2016-12-23
Swift 3 compatibility for Library.framework

mbrandonw:
This one is a doozy!

This upgrades our `Library.framework` to Swift 3, which contains (most) of our view models and platform agnostic shared code. Right now `Library` builds, tests pass, and swiftlint passes except for some `FIXME`s I have placed at places that need attention.

Let's start by reviewing this, and I will work on the `Kickstarter.framework` swift 3 conversion separately and PR it against this. We probably need a way of divvying up this work, but for now I think it's chill just to scan it and call out anything weird. I'll be making a ton of notes in the PR comments to call out things I've learned some some problems I'm having...

10215497-116640902_13-s4-v1

### See also

https://github.com/kickstarter/Kickstarter-Prelude/pull/62
https://github.com/kickstarter/ios-ksapi/pull/3
https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/54

PR #110, opened 2017-03-15
Fix an over-tracking issue.

mbrandonw:
This is a p bad one :(

The over tracking issues we are having comes from a small logical typo from our Swift 3 #26 conversion. It's unfortunately on some untested code, but @stephencelis and I will fix that soon.

The issue is that we never remove successfully tracked events from the event queue, and so we keep on re-tracking those events as the user continues using the app more.


aln2129:
Thanks for digging into this @mbrandonw

PR #111, opened 2017-03-16
Backer dashboard part deux

theginbin:
### What's left:

- Expanding / Collapsing the header (removed all of that for now) - likely for a separate PR

### Future improvements:

- The logic for the Backer Dash header stuff should be moved into its own View and VM

screen shot 2017-03-15 at 9 52 09 pm
screen shot 2017-03-15 at 9 51 57 pm


Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
```diff
@@ -0,0 +1,57 @@
+import KsApi
+import Library
+import UIKit
+
+internal final class BackerDashboardPagesDataSource: NSObject, UIPageViewControllerDataSource {
+  private let viewControllers: [UIViewController]
+
+  internal init(delegate: UIViewController, sort: DiscoveryParams.Sort) {
```
mbrandonw: could we use `BackerDashboardProjectsViewControllerDelegate` as the type of `delegate` instead of `UIViewController`?

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    if let delegate = delegate as? BackerDashboardProjectsViewControllerDelegate {
+      backedController.configureWith(delegate: delegate, projectsType: .backed, sort: sort)
+      savedController.configureWith(delegate: delegate, projectsType: .saved, sort: sort)
+    }
+
+    var controllers: [UIViewController] = []
+    controllers[BackerDashboardTab.backed.rawValue] = backedController
+    controllers[BackerDashboardTab.saved.rawValue] = savedController
+    self.viewControllers = controllers
```
theginbin: just a note here... we don't have a concept of sections for our UIPageViewControllers, but I wanted to ensure there was a way to keep the order of pages (aka tabs) consistent. If you hate the word "tab" let me know.  So `BackerDashboardTab` represents a page section and this concept is used in place of `Int` everywhere a page is referenced in the code. This way, if we ever change the order of the enum, it will be changed everywhere that `BackerDashboardTab` is used. Probably a more elegant way to handle if I had the time to think more on it.

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
```diff
@@ -0,0 +1,57 @@
...
+    let backedController = BackerDashboardProjectsViewController()
+    let savedController = BackerDashboardProjectsViewController()
+
+    if let delegate = delegate as? BackerDashboardProjectsViewControllerDelegate {
+      backedController.configureWith(delegate: delegate, projectsType: .backed, sort: sort)
+      savedController.configureWith(delegate: delegate, projectsType: .saved, sort: sort)
+    }
+
+    var controllers: [UIViewController] = []
+    controllers[BackerDashboardTab.backed.rawValue] = backedController
```
mbrandonw: this is a bit dangerous since if we ever changed the order of the enum cases we would get a crash when we tried doing `controllers[1] = ...` before `controllers[0] = ...`.

rather than the order being determined by the order of the enum cases (which is fragile and doesnt play nicely with our alphabetizing fanaticism!), we could create a static array that determines the order for us:

```swift
case BackerDashboardTab {
  case backed
  case saved

  public static allTabs: [BackerDashboardTab] = [.backed, .saved]
}
```

and then anywhere we need to do something with the tabs that depends on the order we could do something like:

```swift
self.viewControllers = BackerDashboardTab.allTabs.map { tab in 
  switch tab {
  case .backed: return backedController
  case .saved: return savedController
  }
}
```

added bonus of this approach is that adding a new tab will cause compiler errors everywhere so we can correct it for free!

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
```diff
@@ -0,0 +1,57 @@
...
+    controllers[BackerDashboardTab.saved.rawValue] = savedController
+    self.viewControllers = controllers
+  }
+
+  internal func controllerFor(tab: BackerDashboardTab) -> UIViewController? {
+    return self.viewControllers[tab.rawValue]
```
mbrandonw: and then to do this we would need to look up the index of the tab from `allTabs`

Kickstarter-iOS/Tests/Views/BackerDashboardProjectsViewControllerTests.swift
```diff
@@ -0,0 +1,84 @@
+import Prelude
+import Result
+import XCTest
+@testable import Kickstarter_Framework
+@testable import KsApi
+@testable import Library
+
+// swiftlint:disable:next type_name
+internal final class BackerDashboardProjectsViewControllerTests: TestCase {
+  fileprivate var project1: Project!
```
mbrandonw: possible to do this without the IUO vars? could we just put the definitions of these projects right in line?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -289,3 +289,10 @@ private func first<VC: UIViewController>(_ viewController: VC.Type) -> ([UIViewC
       .flatMap { viewControllers[$0] as? VC }
   }
 }
+
+private func profileController() -> UIViewController? {
+  let showDash = AppEnvironment.current.config?.features["ios_backer_dashboard"] == .some(true)
+  return showDash
+    ? BackerDashboardViewController.instantiate() as UIViewController?
+    : ProfileViewController.instantiate() as UIViewController?
```
mbrandonw: i think this function can return a `UIViewController` and you can drop the `as UIViewController?`

Kickstarter-iOS/Views/Cells/BackerDashboardEmptyStateCell.swift
```diff
@@ -0,0 +1,38 @@
...
+  @IBOutlet private weak var iconImageView: UIImageView!
+  @IBOutlet private weak var messageLabel: UILabel!
+
+  internal func configureWith(value: ProfileProjectsType) {
+    switch value {
+    case .backed:
+      self.messageLabel.text = Strings.profile_projects_empty_state_message()
+      self.iconImageView.tintColor = .clear
+    case .saved:
+      self.messageLabel.text = localizedString(key: "todo",
```
mbrandonw: could you use the real key that you will add to `en.yml` that way it will just work once strings are updated even if we aren't using the `Strings.abc` helper function

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
```diff
@@ -0,0 +1,87 @@
...
+  @IBOutlet fileprivate weak var metadataLabel: UILabel!
+  @IBOutlet fileprivate weak var percentFundedLabel: UILabel!
+  @IBOutlet fileprivate weak var projectNameLabel: UILabel!
+  @IBOutlet fileprivate weak var projectImageView: UIImageView!
+  @IBOutlet fileprivate weak var progressStaticView: UIView!
+  @IBOutlet fileprivate weak var progressBarView: UIView!
+  @IBOutlet fileprivate weak var savedIconImageView: UIImageView!
+
+  internal func configureWith(value: Project) {
+    self.viewModel.inputs.project(value)
```
mbrandonw: we typically call this input something like `configureWith(project:)`

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewController.swift
```diff
@@ -0,0 +1,118 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+internal protocol BackerDashboardProjectsViewControllerDelegate: class {
+  /// Called when a project cell is tapped.
+  func profileProjectsGoToProject(_ project: Project, projects: [Project], reftag: RefTag)
```
mbrandonw: curious why you need to tell the delegate to go to the project. could you not present the project straight from this view controller?

Library/ViewModels/BackerDashboardProjectCellViewModel.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    self.metadataText = project.map(metadataString(for:))
+
+    self.metadataIconIsHidden = project.map { $0.state != .live }
+
+    self.percentFundedText = project.map(percentFundedString(for:))
+
+    self.progressBarColor = project.map(progressBarColor(for:))
+
+    self.savedIconIsHidden = project.map { $0.personalization.isStarred != .some(true) ?? true }
```
mbrandonw: no need to coalesce `?? true` since `$0.personalization.isStarred != .some(true)` returns an honest boolean.

woulda thought swift would warn about that tbh

Library/ViewModels/BackerDashboardViewModel.swift
```diff
@@ -0,0 +1,242 @@
...
+    self.embeddedViewTopConstraintConstant = self.sortBarIsHidden
+      .map { $0 ? 0.0 : Styles.grid(2) }
+
+    self.updateProjectPlaylist = self.profileProjectsUpdatePlaylistProperty.signal.skipNil()
+
+    self.viewWillAppearProperty.signal.filter(isFalse)
+      .observeValues { _ in AppEnvironment.current.koala.trackProfileView() }
+
+    self.backedProjectsButtonTappedProperty.signal
+    .observeValues { _ in AppEnvironment.current.koala.trackViewedProfileBackedTab() }
```
mbrandonw: i would think these events would be tracked in the viewwillappear of the projects controller and not here. you could imagine that there are maybe other ways to navigate between tabs besides tapping on a tab button (e.g. deep linking), and this tracking doesnt seem to track when you first navigate to the profile.

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
theginbin: I chose generic because the other 2 tabs will not be of this type (Following and Live Streams)

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
theginbin: word i like this!

Kickstarter-iOS/Tests/Views/BackerDashboardProjectsViewControllerTests.swift
theginbin: yeah i think i thought i would do more than 1 test but no need.

Kickstarter-iOS/Views/Cells/BackerDashboardEmptyStateCell.swift
theginbin: i don't have any copy yet, so not sure what the key will be.

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
theginbin: hmm whatcha mean, all our cells inherit from `ValueCell` and have to define `internal func configureWith(value: Value)`

theginbin: oh wait, you mean the input. nevermind!

mbrandonw: just that the view model's input would be `self.viewModel.inputs.configureWith(project: value)`

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewController.swift
theginbin: i don't. will fix!

Library/ViewModels/BackerDashboardProjectCellViewModel.swift
theginbin: right right

Library/ViewModels/BackerDashboardViewModel.swift
theginbin: i think it's just for tracking button taps on the tabs actually. there is a separate event for actually going to the Profile page initially.

theginbin: i can ask @ktman to clarify.

ktman: yeah we have a `Viewed Profile` event that triggers on initial profile load, this is just to distinguish between tabs. An alternative approach would be to make the event `Viewed Profile Tab` with a `type` property for Backed, Saved, etc etc

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewController.swift
```diff
@@ -38,7 +38,15 @@ internal final class BackerDashboardProjectsViewController: UITableViewControlle
...
-    self.refreshControl?.rac.refreshing = self.viewModel.outputs.isRefreshing
+    self.viewModel.outputs.isRefreshing
+      .observeForUI()
+      .observeValues { [weak self] isRefreshing in
+        if isRefreshing {
+          self?.refreshControl?.beginRefreshing()
+        } else {
+          self?.refreshControl?.endRefreshing()
+        }
+    }
```
mbrandonw: strange, what exactly didnt work? wonder if we need to fix the extension...

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
```diff
@@ -7,12 +7,10 @@ internal final class BackerDashboardPagesDataSource: NSObject, UIPageViewControl
 
   internal init(delegate: UIViewController, sort: DiscoveryParams.Sort) {
     let backedController = BackerDashboardProjectsViewController()
-    let savedController = BackerDashboardProjectsViewController()
+    backedController.configureWith(projectsType: .backed, sort: sort)
```
mbrandonw: we typically use a static `configuredWith` on controllers for instantiating them and configuring them in one swoop, e.g. https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/Views/Controllers/BackingViewController.swift#L35-L39

theginbin: we typically use that style when we are pulling something from a storyboard. This controller has no storyboard representation.

mbrandonw: true that we mostly use storyboards but don't think that's the reason we adopted that style. was more about getting a fully configured vc at once instead of instantiating it and then doing work on it

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewController.swift
theginbin: the output didn't trigger the rac extension when it was written that way. I'm not sure why.

Kickstarter-iOS/DataSources/BackerDashboardPagesDataSource.swift
theginbin: yes i understand why we choose to do this. pardon the oversight.

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -289,3 +289,10 @@ private func first<VC: UIViewController>(_ viewController: VC.Type) -> ([UIViewC
       .flatMap { viewControllers[$0] as? VC }
   }
 }
+
+private func profileController() -> UIViewController {
+  let showDash = AppEnvironment.current.config?.features["ios_backer_dashboard"] == .some(true)
```
mbrandonw: could we add a test for this? we have a few examples of testing features, but basically you just push a new config onto the environment:

```swift

    let config = .template
      |> Config.lens.features .~ ["ios_backer_dashboard": true]

    withEnvironment(config: config) {
```

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    _ = self.metadataIconImageView
+      |> UIImageView.lens.tintColor .~ .white
+
+    _ = self.progressStaticView
+      |> UIView.lens.backgroundColor .~ .black
+      |> UIView.lens.alpha .~ 0.15
+
+    _ = self.savedIconImageView
+      |> UIImageView.lens.tintColor .~ .init(white: 1.0, alpha: 0.99)
```
mbrandonw: is it possible to use `.white`? things with alpha cause ios to do additional graphics compositing work which can slow down rendering.

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewControllerTests.swift
```diff
@@ -0,0 +1,109 @@
+import Prelude
+import Result
+import XCTest
+@testable import Kickstarter_Framework
+@testable import KsApi
+@testable import Library
+
+// swiftlint:disable:next type_name
+internal final class BackerDashboardProjectsViewControllerTests: TestCase {
```
mbrandonw: oh wow, i was wondering why it felt so nice reviewing this pr and just realized that the test+implementation co-location means you alternate between the two when going down a pr! no more having tests all lumped together and implementation all together...

Library/ViewModels/BackerDashboardProjectCellViewModel.swift
```diff
@@ -0,0 +1,152 @@
...
+  public init() {
+    let project = projectProperty.signal.skipNil()
+
+    self.projectTitleText = project.map(titleString(for:))
+
+    self.photoURL = project.map { URL(string: $0.photo.full) }
+
+    self.progress = project.map { $0.stats.fundingProgress }
+
+    self.cellAccessibilityLabel = project.map { "\($0.name) \($0.state.rawValue)" }
```
mbrandonw: no need to do this now, but we should make a new localized string for this. maybe we can just card for now

Library/ViewModels/BackerDashboardProjectsViewModel.swift
```diff
@@ -0,0 +1,178 @@
...
+    self.isRefreshing = isLoading
+
+    self.emptyStateIsVisible = Signal.combineLatest(projectsType, self.projects)
+      .map { type, projects in
+        (projects.isEmpty, type)
+    }
+
+    self.goToProject = Signal.combineLatest(projectsType, self.projects)
+      .takePairWhen(self.projectTappedProperty.signal.skipNil())
+      .map { typeAndProjects, project in
```
mbrandonw: could `map(unpack)` to get a triple for this instead of a nested tuple

Kickstarter-iOS/ViewModels/RootViewModel.swift
theginbin: yes definitely. good suggestion.

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
theginbin: would you believe that when I use 100% white, the graphic turns a weird grey color?? I couldn't figure it out until I played with the alpha. It is resting on top of a UIView overlay with opacity, but other than that I cannot fathom why this gave issue. Thoughts?

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewControllerTests.swift
theginbin: oh that's brilliant

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
mbrandonw: oh jeez :/ is it a template image thing?

we could look at it irl, but that is really bizarre.

Library/ViewModels/BackerDashboardProjectCellViewModel.swift
theginbin: yes woops on non-localization here - there should be a string already as we translate project state for the metadata view. would be a good idea to run this on device and understand what label makes sense here actually (e.g. do we care to tell them the percent funded too?).

Library/ViewModels/BackerDashboardProjectsViewModel.swift
theginbin: forgot about good ol map unpack

Kickstarter-iOS/Views/Controllers/BackerDashboardProjectsViewControllerTests.swift
theginbin: add it to your article ;)

mbrandonw: You know me so well. Already did it!

> . It also aids in reviewing pull-requests since the implementation and test files will alternate instead of having them lumped together.

theginbin:
hey @mbrandonw I just removed the a11y label from this cell because i think it reads the cell correctly without it (aka [Project state]. [Project title]. [percent funded]. "Button." "Opens project.")
If that is cool, can you approve and I'll add the collapsing header stuff to a new PR? Thanks.

PR #112, opened 2017-03-17
Messages read fix

cdolm92:
## What

The unread indicator should be shown when the user has unread messages.  If the user has no unread messages the indicator should be hidden. 

However, when the user returns back to the inbox from reading their messages the indicator continues to be shown. To fix this we created a new input, setSelected(_:animated:) and stored a bool that indicates if there are unread messages to the current Environment's cache.


Kickstarter-iOS/Tests/Views/MessagesThreadsViewControllerTests.swift
```diff
@@ -0,0 +1,78 @@
+import Library
+import Prelude
+import Result
+import XCTest
+@testable import Kickstarter_Framework
+@testable import KsApi
+
+internal final class MessagesThreadViewControllerTests: TestCase {
```
mbrandonw: oh dang, nice! thanks for adding some screenshot tests!

Kickstarter-iOS/Tests/Views/MessagesThreadsViewControllerTests.swift
```diff
@@ -0,0 +1,78 @@
...
+      currentUser: backer,
+      mainBundle: Bundle.framework
+    )
+
+    Language.allLanguages.forEach { language in
+        withEnvironment(language: language) {
+          let controller = MessageThreadsViewController.configuredWith(project: project)
+
+          let (parent, _) = traitControllers(device: .phone4_7inch, orientation: .portrait, child: controller)
+          parent.view.frame.size.height = 600
```
mbrandonw: let's not set the height of this controller. we only like to do that if we have a lot of content we want to get into a single screenshot.

Library/ViewModels/MessageThreadsViewModel.swift
```diff
@@ -22,6 +22,9 @@ public protocol MessageThreadsViewModelInputs {
   /// Call when the view loads.
   func viewDidLoad()
 
+  /// Call when the view appears.
+  func viewWillAppear(animated: Bool)
```
mbrandonw: looks like we didnt end up using this endpoint so let's remove it

Kickstarter-iOS/Views/Controllers/MessageThreadsViewController.swift
```diff
@@ -46,14 +52,14 @@ internal final class MessageThreadsViewController: UITableViewController {
     self.tableView.tableFooterView?.rac.hidden = self.viewModel.outputs.loadingFooterIsHidden
 
     self.viewModel.outputs.messageThreads
-      .observeForControllerAction()
+      .observeForUI()
```
mbrandonw: nice thanks for catching these

Kickstarter-iOS/Tests/Views/MessagesThreadsViewControllerTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    AppEnvironment.pushEnvironment(
+      apiService: MockService(
+        fetchMessageThreadsResponse: messageThreads
+      ),
+      currentUser: backer,
+      mainBundle: Bundle.framework
+    )
+
+    Language.allLanguages.forEach { language in
```
mbrandonw: i don't think we need to generate screenshots for each language since this UI doesnt have any localized strings. it's all user generated content. this will help reduce the number of screenshots we have.

and maybe we could also do a screenshot for iPad?

Library/Tests/ViewModels/MessageThreadCellViewModelTests.swift
```diff
@@ -50,7 +50,20 @@ final class MessageThreadCellViewModelTests: TestCase {
...
+  func testUnreadIndicatorHidden() {
+    let thread = MessageThread.template
+
+    self.vm.inputs.configureWith(messageThread: thread)
+
+    self.vm.inputs.setSelected(false)
+    self.unreadIndicatorHidden.assertValues([false])
+
+    self.vm.inputs.setSelected(true)
+    self.unreadIndicatorHidden.assertValues([false, true])
```
mbrandonw: we could add a further test for the caching logic by simulating what happens when the cell scrolls off the page and back on. we do that by doing a `configureWith` with a different message thread, and then re-doing `configureWith` with this message thread. that requires having two message threads with different ids. wanna give that a shot?

Library/ViewModels/MessageThreadCellViewModel.swift
```diff
@@ -2,9 +2,11 @@ import KsApi
 import ReactiveSwift
 import ReactiveExtensions
 import Result
+import Prelude
```
mbrandonw: could we alphabetize this?

PR #103, opened 2017-02-27
Instrument updates, fix update load bug

luoser:
## what
* Instrumented a couple of update events 
* Fixed a navigation bug (Safari browser opened when it shouldn't) when you tap through Update indexes

Bug:

Fix:

PR #113, opened 2017-03-17
Remove unneeded update tracking.

mbrandonw:
The pr #103 added some tracking to updates that isn't needed since it's done on the server side. So just removing that and correcting a few small things.

### DIFF
--- a/Kickstarter-iOS/ViewModels/UpdateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/UpdateViewModel.swift
@@ -5,10 +5,10 @@ import Prelude
 import ReactiveSwift
 import Result
 
-public struct UpdateData {
+private struct UpdateData {
+  public let context: Koala.UpdateContext
   public let project: Project
   public let update: Update
-  public let context: Koala.UpdateContext
 }
 
 internal protocol UpdateViewModelInputs {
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/ViewModels/UpdateViewModel.swift
```diff
@@ -5,10 +5,10 @@ import Prelude
 import ReactiveSwift
 import Result
 
-public struct UpdateData {
+private struct UpdateData {
```
luoser: oh dang actually you don't even need to use this struct anymore since this `context` was for the unneeded events sry

mbrandonw: lol

Kickstarter-iOS/ViewModels/UpdateViewModel.swift
```diff
@@ -5,15 +5,14 @@ import Prelude
 import ReactiveSwift
 import Result
 
-public struct UpdateData {
-  public let project: Project
-  public let update: Update
-  public let context: Koala.UpdateContext
+fileprivate struct UpdateData {
```
luoser: you could also kill this `UpdateData` struct entirely and just use a pair for project and update if your heart desires :)

mbrandonw: yah for sure but just wanna do the minimal work right now. or you can clean up your own mess ;)

luoser: ༼ຈل͜ຈ༽

PR #104, opened 2017-03-06
Update and sync some global styles

theginbin:
### What

While working on updating the backer dashboard, I felt it necessary to separately update the entire app to match the new design in terms of navigation bar and background colors. As we keep going forward with new designs, it will be easier to start from a place where colors are in sync.

The main changes here are a white background for navigation bars and a white background color for screens that are not table view or collection view controllers. The table/collection view controllers have been updated to have a white foreground color and a darker grey background color. Navigation titles are now black. 

I'd like to get the navigation bar border image to be the same across the app (grey_500), but for the sake of time, am keeping the default shadow for now.

Note: Still have to run past @maggieKSR but wanted to get the code and screenshots out for review.

PR #114, opened 2017-03-17
Fixed live styles

mbrandonw:
This is just for posterity, but #104 messed with some nav bar styles on our live streaming controllers and this fixes that. Will merge when tests pass.

### DIFF
--- a/Kickstarter-iOS/Views/Storyboards/Activity.storyboard
+++ b/Kickstarter-iOS/Views/Storyboards/Activity.storyboard
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11762" systemVersion="15G1217" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
+<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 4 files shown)


Library/Styles/BaseStyles.swift
```diff
@@ -119,5 +125,12 @@ private let baseNavigationBarStyle =
     NSForegroundColorAttributeName: UIColor.black,
     NSFontAttributeName: UIFont.ksr_callout()
     ]
-    <> UINavigationBar.lens.translucent .~ false
```
theginbin: quick q, the default `translucent` value is `true`. how come this was removed?

mbrandonw: ah sorry, i thought the default was `false` and it seemed to work, but now i see it's wrong. fixing...

Kickstarter-iOS/Views/Storyboards/Activity.storyboard
```diff
@@ -385,7 +385,7 @@
                                                         <constraint firstAttribute="width" constant="80" id="vCB-Xb-z6f"/>
                                                     </constraints>
                                                 </imageView>
-                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="HERTZFELDT ON BLU-RAY" textAlignment="natural" lineBreakMode="tailTruncation" numberOfLines="2" baselineAdjustment="alignBaselines" minimumScaleFactor="0.69999999999999996" translatesAutoresizingMaskIntoConstraints="NO" id="peK-ts-qtm" customClass="StyledLabel" customModule="Library">
```
mbrandonw: sorry saw this `StyledLabel` getting logged to the console so wanted to remove. unrelated to this pr but still nice to get rid of

theginbin: oh nice, thanks!

Library/Styles/BaseStyles.swift
theginbin: yeah apple loves that translucent look

PR #100, opened 2017-02-23
new colors and new screens

theginbin:
🍡  Some color updates 🍡

PR #115, opened 2017-03-20
Fixed color for title in discovery card.

mbrandonw:
I forgot to include this in the 3.3.2 that we just released, so adding it now. Basically during the #100 pr we lost a bit of contrast in the title/blurb in discovery cards. Fixing that now...

screenshot 2017-03-20 14 06 34


theginbin:
thanks for doing this @mbrandonw!

PR #116, opened 2017-03-20
More email deep linking

mbrandonw:
### What

The captures the last types of links that can be embedded in our emails so that we can deep-link to projects in 48 hour reminder emails, project success emails, etc...


Library/Navigation.swift
```diff
@@ -558,16 +570,3 @@ private func oneToBool(_ string: String?) -> Decoded<Bool?> {
...
- is analagous to `zip` on sequences if you think of optionals as sequences with at most one element.
-
- - parameter a: An optional.
- - parameter b: An optional.
-
- - returns: An optional tuple.
- */
-private func zip <A, B> (_ x: A?, _ y: @autoclosure () -> B?) -> (A, B)? {
-  return x.flatMap { x_ in y().map { y_ in (x_, y_) } }
-}
```
mbrandonw: we no longer even use `zip` on optionals so getting rid of it. 

was fun while it lasted! maybe we can add to prelude some day

Library/Navigation.swift
```diff
@@ -221,6 +224,7 @@ extension Navigation {
 
 private let routes: [String:(RouteParams) -> Decoded<Navigation>] = [
   "/": emailClick,
+  "/mpss/:a/:b/:c/:d/:e/:f/:g": emailLink,
```
mbrandonw: would be nice  if we could match against urls with wildcards, like `/mpss/*`, but i don't have time to flesh that out and it seems all email urls are of this form with 7 path components after `mpss`

PR #117, opened 2017-03-22
Test co-location

mbrandonw:
Finishing off the series of PR's to co-locate all of our test files (see also https://github.com/kickstarter/ios-ksapi/pull/18, https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/60, https://github.com/kickstarter/Kickstarter-Prelude/pull/69).

This one was more difficult than the previous 3. The big reason is that we do a bunch of stuff in tests that we consider unacceptable in our implementation code, like force casts, which will trigger swiftlint errors. I think an argument could be made that this isn't right, and we should have the same standard of quality for tests as we do for all other code. However, just to get us started I decided to explicitly disable certain swiftlint rules in the files that have violations. I think going forward we should strive to not break any of our swiftlint rules, no matter the file.


mbrandonw:
a lil preview of what this looks like in practice:

screenshot 2017-03-22 17 44 27

theginbin:
very nice!

PR #119, opened 2017-03-25
Conform to new swiftlint rules

justinswart:
The latest swiftlint introduced some new rules, this PR just addresses them.

Most notable new rule is `discarded_notification_center_observer`. Turns out all of our block/closure `Notification` observers were leaking?

Most of the others are fairly innocuous but nice to have them tidied up.

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -154,6 +154,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .observeForUI()
       .observeValues { [weak self] in self?.findRedirectUrl($0) }
 
+    //swiftlint:disable discarded_notification_center_observer
     NotificationCenter.default
       .addObserver(forName: Notification.Name.ksr_sessionStarted, object: nil, queue: nil) { [weak self] _ in
         self?.viewModel.inputs.userSessionStarted()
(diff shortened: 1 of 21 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -154,6 +154,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .observeForUI()
       .observeValues { [weak self] in self?.findRedirectUrl($0) }
 
+    //swiftlint:disable discarded_notification_center_observer
```
justinswart: I've disabled this here seeing as `AppDelegate` lives forever?

mbrandonw:
> Most notable new rule is discarded_notification_center_observer. Turns out all of our block/closure Notification observers were leaking?

This isn't needed in iOS 9+. From https://developer.apple.com/library/content/releasenotes/MacOSX/WhatsNewInOSX/Articles/MacOSX10_11.html:

> NSNotificationCenter and NSDistributedNotificationCenter no longer send notifications to registered observers that may be deallocated. If the observer is able to be stored as a zeroing-weak reference the underlying storage stores the observer as a zeroing weak reference. Alternatively, if the object cannot be stored weakly (because it has a custom retain/release mechanism that would prevent the runtime from being able to store the object weakly) the object is stored as a non-weak zeroing reference. This means that observers are not required to un-register in their deallocation method.

justinswart:
Ah, I should read the docs occasionally :)

PR #120, opened 2017-03-26
Fix URL in README

tiagomartinho:
- Since tests are now collocated with the ViewModels the correct folder
  for the tests is the same as the ViewModels


mbrandonw:
good catch! thanks so much!

tiagomartinho:
Welcome 😄

PR #122, opened 2017-03-28
Clear first comment fix

cdolm92:
## WHAT

Empty states for comments is now hidden immediately after first comment is added. This is done by having emptyStateVisible also emit a boolean, that determines if there is a comment to display.

first-comment-clear-fix


Kickstarter-iOS/DataSources/CommentsDataSource.swift
```diff
@@ -14,8 +14,8 @@ internal final class CommentsDataSource: ValueCellDataSource {
              inSection: Section.comments.rawValue)
   }
 
-  internal func load(project: Project, update: Update?) {
-    self.set(values: [(project, update)],
+  internal func load(project: Project, update: Update?, visible: Bool) {
```
mbrandonw: seems like we could combine the functionality of this method with the other `load` method. essentially, if we hit `load(comments:project:loggedInUser)` with an empty array of comments then we use the empty state, and otherwise use the comments cell.

we would have to make that method also take an update, but i think overall it would be simpler.

Kickstarter-iOS/Views/Controllers/CommentDialogViewController.swift
```diff
@@ -21,7 +21,6 @@ internal final class CommentDialogViewController: UIViewController {
   @IBOutlet fileprivate weak var subtitleLabel: UILabel!
   @IBOutlet fileprivate weak var bodyTextView: UITextView!
   @IBOutlet fileprivate weak var postButton: UIBarButtonItem!
-  @IBOutlet fileprivate weak var loadingView: UIView!
```
mbrandonw: why was this removed?

PR #125, opened 2017-03-31
Filter and settings copy

cdolm92:
## WHAT 

New copy in discovery filters, settings, and notifications.


Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -199,7 +199,7 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
   fileprivate func showFavoriteCategoriesAlert(categoryName: String) {
     let alertController = UIAlertController(
       title: categoryName,
-      message: Strings.discovery_favorite_categories_alert_message(),
+      message: Strings.Access_all_your_favorite_categories_in_the_explore_dropdown(),
```
theginbin: This copy doesn't make sense as an alert after favoriting a category because the only way they can access this favorite button is after they've switched to a favorite category from the dropdown. Was this copy supposed to be somewhere else, like future onboarding? 
Note, the original copy is "Yes! You've saved this as one of your favorite categories."

Library/ViewModels/DiscoveryNavigationHeaderViewModel.swift
```diff
@@ -210,7 +210,11 @@ public final class DiscoveryNavigationHeaderViewModel: DiscoveryNavigationHeader
       .map { $0.selected ? Strings.discovery_favorite_categories_buttons_unfavorite_a11y_label() :
         Strings.discovery_favorite_categories_buttons_favorite_a11y_label() }
 
-    self.showFavoriteOnboardingAlert = strings.map { $0.subcategory ?? $0.filter }
+    self.showFavoriteOnboardingAlert = strings
+      .map { $0.subcategory != nil ?
+        Strings.category_name_subcategory_name_saved(
+          category_name: $0.filter, subcategory_name: $0.subcategory ??  "") :  $0.filter
+      }
```
theginbin: This condition would be easier to read if the clauses are on separate lines, ala:
```
.map { $0.subcategory != nil
 ? Strings.category_name_subcategory_name_saved(category_name: $0.filter, subcategory_name: $0.subcategory ??  "")
 : $0.filter
}

Comment (theginbin) [inline Library/ViewModels/DiscoveryNavigationHeaderViewModel.swift]: I think we should talk about the title for this alert, sorry I didn't see this on the main repo or I would've brought it up then. The string here is "%{category_name}/%{subcategory_name} saved." This means that if there is no subcategory being saved, the copy will read "All Art Projects/ saved." so there would need to be logic to include the "/" character. Even without the slash, "All Art Projects saved." needs to update to capitalize the "saved" text. I think we should regroup with sara on this.

File: Kickstarter-iOS/Views/Controllers/SettingsViewController.swift [hunk 3]
```diff
@@ -313,7 +313,7 @@ internal final class SettingsViewController: UIViewController {
 
     _ = self.wereAllEarsTitleLabel
       |> settingsTitleLabelStyle
-      |> UILabel.lens.text %~ { _ in Strings.profile_settings_rating_title() }
+      |> UILabel.lens.text %~ { _ in Strings.We_re_all_ears() }
```
theginbin: Can we chat with Sara about an alternate expression here or maybe lose it altogether? I think the period instead of an exclamation makes this title feel awkwardly serious.

screen shot 2017-04-03 at 11 34 53 am

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
theginbin: oh my bad, i now see the word "favorite" hah

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
ktman: @awakinglim you cool to keep the ! for We're all ears?

awakinglim: For localization clarity, let's change that section header to _Reviews_.

PR #126, opened 2017-04-02
Build and test using latest iOS version (10.3)

tiagomartinho:


mbrandonw:
Hey, thanks for the pr! This will require us to update screenshots since there are small pixel differences in 10.2 vs 10.3, so we're gonna take a bit more time before flipping the switch to 10.3.

Gonna close for now and we'll revisit soon. Thanks again!

PR #127, opened 2017-04-05
fix validation errors in .swiftlint.yml

jpsim:
SwiftLint is migrating to a more robust YAML parser (libYAML, see https://github.com/realm/SwiftLint/pull/1411). The previous parser wasn't spec-compliant and actually accepted invalid YAML sequences as input.

This commit makes .swiftlint.yml pass validation at http://www.yamllint.com.


mbrandonw:
Oh hey, thanks for this!

PR #130, opened 2017-04-10
Prevent login filtering Discovery

justinswart:
During #107 I picked up that logging in while a modal view controller was presented on `LiveStreamDiscoveryViewController` caused it to disappear. This was due to a `combineLatest` in `RootViewModel`'s `filterDiscovery` output.

Changing this to a `takePairWhen` solves the problem.

This will affect logging in to subscribe in the current live build when presented from live discovery.

### DIFF
--- a/Kickstarter-iOS/ViewModels/RootViewModelTests.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModelTests.swift
@@ -280,6 +280,28 @@ final class RootViewModelTests: TestCase {
 
     self.tabBarItemsData.assertValues([tabData, tabDataLoggedIn, tabDataLoggedIn, tabData, tabDataMember])
   }
+
+  func testSetViewControllers_DoesNotFilterDiscovery() {
+    self.filterDiscovery.assertValueCount(0)
+
+    let viewControllerNames = TestObserver<[String], NoError>()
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/ViewModels/RootViewModelTests.swift
```diff
@@ -280,6 +280,28 @@ final class RootViewModelTests: TestCase {
 
     self.tabBarItemsData.assertValues([tabData, tabDataLoggedIn, tabDataLoggedIn, tabData, tabDataMember])
   }
+
+  func testSetViewControllers_DoesNotFilterDiscovery() {
```
mbrandonw: this test wouldn't have failed with the previous code unfortunately.

i would suggest reverting the vm to go back to `combineLatest` and then get a failing test that the `takeWhen` fixes.

essentially what we are testing is:

* view loads and `filterDiscovery` doesnt emit
* we call `switchToDiscovery` explicitly (cause some other part of the app wanted to), and that causes `filterDiscovery` to emit
* later a user logs in (or out) causing the controllers to be reloaded, _but_ `filterDiscovery` does not emit again

justinswart: Ah of course, apologies rushed through it.

PR #132, opened 2017-04-13
Update KsApi

mbrandonw:
We've had a couple of nice additions to the prelude (https://github.com/kickstarter/Kickstarter-Prelude/pull/75 & https://github.com/kickstarter/Kickstarter-Prelude/pull/74), so just bring that into the main app.


Library/ViewModels/CheckoutViewModel.swift
```diff
@@ -274,7 +274,7 @@ public final class CheckoutViewModel: CheckoutViewModelType {
     }
 
     configData
-      .takeWhen(self.stripeTokenAndErrorProperty.signal.filter(isNotNil • first))
+      .takeWhen(self.stripeTokenAndErrorProperty.signal.filter(first >>> isNotNil))
```
mbrandonw: reads a bit nicer eh?

theginbin:
the double dot notation makes sense. i dig it!

PR #134, opened 2017-04-20
Fastlane deliver

justinswart:
Just added what I think we need to deliver to the app store via Fastlane again.

Would like to test this with our next submission.

Reference: https://docs.fastlane.tools/actions/#deliver

### DIFF
--- a/.fastlane/Fastfile
+++ b/.fastlane/Fastfile
@@ -61,6 +61,14 @@ platform :ios do
       skip_screenshots: true,
       skip_metadata: true
     )
+
+    slack(
+      slack_url: ENV["SLACK_WEBHOOK"],
+      message: "Successfully uploaded"
+        + " v" + get_version_number(xcodeproj: "Kickstarter.xcodeproj")
+        + " (" + get_build_number(xcodeproj: "Kickstarter.xcodeproj") + ")"
+        + " of the Kickstarter iOS app to iTunes Connect 🚀"
+    )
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
     - security find-identity -p codesigning
     - instruments -s devices
     - xcodebuild -showsdks
-    - gem install fastlane -NVv 2.26.1
+    - gem install fastlane -NVv 2.28.2
```
justinswart: I've upped the Fastlane version as it includes a fix we'll need:

> Handle dSYM uploading for HockeyApp (#8863) via Prata Catalin

Ref: https://github.com/fastlane/fastlane/releases/tag/2.28.0

.fastlane/Fastfile
```diff
@@ -48,13 +48,26 @@ platform :ios do
...
+      app_identifier: ENV["ITUNES_APP_IDENTIFIER"],
       app: ENV["ITUNES_APP_ID"],
       ipa: "Kickstarter.ipa",
-      team_id: ENV["ITUNES_TEAM_ID"]
+      team_id: ENV["ITUNES_TEAM_ID"],
+      skip_screenshots: true,
+      skip_metadata: true
+    )
+
+    slack(
```
justinswart: Figured it'd be nice to ping the channel when a new build is submitted to the store.

PR #135, opened 2017-04-20
Discovery card style update

cdolm92:
## WHAT
 Updated discovery project cards styles.  The funding, backers, and deadline title label text sizes were increased. The funding, backers, and deadline subtitle/title label text colors were changed. The funding subtitle/title label now is `.ksr_text_green_700` and the rest being `.ksr_text_navy_700`.  

The behavior of the `fundingProgressBarView` and `fundingProgressBarContainerView` were also changed. When a project is canceled or suspended the `fundingProgressBarContainerView` is now set to hidden.  When a project is unsuccessful only the `fundingProgressBarView` is set to hidden.

There are also new strings as some of the copy in the card changed.


Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -14,8 +14,8 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
   @IBOutlet fileprivate weak var deadlineTitleLabel: UILabel!
   @IBOutlet fileprivate weak var fundingProgressBarView: UIView!
   @IBOutlet fileprivate weak var fundingProgressContainerView: UIView!
-  @IBOutlet fileprivate weak var fundingSubtitleLabel: UILabel!
   @IBOutlet fileprivate weak var fundingTitleLabel: UILabel!
+  @IBOutlet fileprivate weak var fundingSubtitleLabel: UILabel!
```
mbrandonw: looks like we should move this label up to make it alphabetized

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewControllerTests.swift
```diff
@@ -102,4 +133,8 @@ internal final class DiscoveryPageViewControllerTests: TestCase {
 
   fileprivate let magicParams = .defaults
     |> DiscoveryParams.lens.sort .~ .magic
+
+  fileprivate let starredParams = .defaults
+    |> DiscoveryParams.lens.starred .~ true
```
mbrandonw: don't think we need this anymore

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -64,6 +64,12 @@ public protocol DiscoveryPostcardViewModelOutputs {
   /// Emits the text for the deadline title label.
   var deadlineTitleLabelText: Signal<String, NoError> { get }
 
+  /// Emits a boolean to determine whether or not to display funding progress container view.
+  var fundingProgressContainerViewHidden: Signal<Bool, NoError> { get }
+
+  /// Emits a boolean to determine whether or not to display the funding progress bar view.
+  var fundingProgressBarViewHidden: Signal<Bool, NoError> { get }
```
mbrandonw: these should be swapped to be alphabetized

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -197,6 +203,12 @@ public final class DiscoveryPostcardViewModel: DiscoveryPostcardViewModelType,
       .map { $0.personalization.friends == nil || $0.personalization.friends?.count ?? 0 == 0 }
       .skipRepeats()
 
+    self.fundingProgressContainerViewHidden = project
+      .map { $0.state == .canceled || $0.state == .suspended }
```
mbrandonw: beaut!

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -197,9 +203,17 @@ public final class DiscoveryPostcardViewModel: DiscoveryPostcardViewModelType,
...
+
+    self.fundingProgressBarViewHidden = project
+      .map { $0.state == .failed }
+
     // a11y
     self.cellAccessibilityLabel = project.map(Project.lens.name.view)
-    self.cellAccessibilityValue = project.map(Project.lens.blurb.view)
+
+    self.cellAccessibilityValue = Signal.combineLatest(project, self.projectStateTitleLabelText)
+      .map { project, projectState in "\(project.blurb)\(projectState)" }
```
mbrandonw: this interpolation might be weird if the blurb doesn't have a space/punctuation at the end, e.g. "This is some project blurbSuccessfully funded". instead we could interpolate it as `"\(project.blurb). \(projectState)"`.

also, `combineLatest` will cause some over emissions cause it will emit once for each of the signals, hence twice overall. a better operator would be `zip` which requires each signal to emit once before emitting again.

PR #138, opened 2017-04-26
Track sent chat message

justinswart:
### What

Koala tracking event for when a chat message is sent.

Also removed a redundant tracking method that should've been removed from #107.

### DIFF
--- a/Library/Koala/KoalaTests.swift
+++ b/Library/Koala/KoalaTests.swift
@@ -583,14 +583,12 @@ final class KoalaTests: TestCase {
 
     koala.trackLiveStreamChatSentMessage(
       project: .template,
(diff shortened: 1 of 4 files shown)


Library/Koala/KoalaTests.swift
```diff
@@ -571,4 +571,26 @@ final class KoalaTests: TestCase {
     XCTAssertEqual(["Cool Live Stream"], client.properties(forKey: "live_stream_name", as: String.self))
     XCTAssertEqual([startDate], client.properties(forKey: "live_stream_start_date", as: Double.self))
   }
+
+  func testTrackSentChatMessage() {
```
mbrandonw: nice! thanks for writing a dedicated koala tracking test! we don't do this enough

PR #140, opened 2017-04-26
Instrumentation for search featured projects

cdolm92:
### WHAT

Added ref tags: `search_popular_featured`, `search_popular`, `search_featured`, and `search`. We calculate the ref tag from the search query, list of displayed projects and the selected project. 

When the search screen first loads popular projects are displayed and  the the query `isEmpty`. If the user selects the project displayed on the `MostPopularSearchProjectCell` the ref tag should be `search_popular_featured`,  if they select a project displayed in the `SearchProjectCell` the ref tag will be `search_popular`.   

When the user makes a search the query is not empty. After search, If the user taps on the project displayed on the `MostPopularSearchProjectCell` the ref tag should be `search_featured`, if they select a project displayed in the `SearchProjectCell` the ref tag will be `search`.


Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -53,6 +54,106 @@ internal final class SearchViewModelTests: TestCase {
...
+        self.vm.inputs.searchFieldDidBeginEditing()
+        self.vm.inputs.searchTextChanged("robots")
+
+        self.scheduler.advance()
+        self.hasProjects.assertValues([true, false, true],
+                                      "Projects emitted immediately upon view appearing.")
+
+        self.vm.inputs.tapped(project: projects2[2])
+
+        self.goToRefTag.assertValues([refTag])
```
mbrandonw: I say we put `RefTag.search` directly in this assertion so that we don't have to look up at the top of the test to see what ref tag we are checking.

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -53,6 +54,106 @@ internal final class SearchViewModelTests: TestCase {
...
+      self.hasProjects.assertValues([true], "Projects emitted immediately upon view appearing.")
+
+      withEnvironment(apiService: MockService(fetchDiscoveryResponse: response2)) {
+
+        self.vm.inputs.searchFieldDidBeginEditing()
+        self.vm.inputs.searchTextChanged("robots")
+
+        self.scheduler.advance()
+        self.hasProjects.assertValues([true, false, true],
+                                      "Projects emitted immediately upon view appearing.")
```
mbrandonw: I think we could prob remove the `hasProjects` assertions since we have asserted on it in other tests and this test is mostly focused on the ref tag.

we could make the whole user script a bit succinct with:

```swift
self.vm.inputs.searchFieldDidBeginEditing()
self.vm.inputs.searchTextChanged("robots")
self.scheduler.advance()
self.vm.inputs.tapped(project: projects2[2])

self.goToRefTag.assertValues([refTag])
```

can do something similar in the other tests too

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -55,101 +55,69 @@ internal final class SearchViewModelTests: TestCase {
...
 
   func testSearch_RefTag() {
-    let refTag = RefTag.search
-
     let projects = (0...10).map { idx in .template |> Project.lens.id .~ (idx + 42) }
     let response = .template |> DiscoveryEnvelope.lens.projects .~ projects
-    let projects2 = (20...30).map { idx in .template |> Project.lens.id .~ (idx + 42) }
-    let response2 = .template |> DiscoveryEnvelope.lens.projects .~ projects2
+    let searchProjects = (20...30).map { idx in .template |> Project.lens.id .~ (idx + 42) }
+    let searchResponse = .template |> DiscoveryEnvelope.lens.projects .~ searchProjects
```
mbrandonw: ohhh even better!

PR #142, opened 2017-04-28
Live chat fix view count and messages fix

justinswart:
### What

During our testing we picked up that the view count on the live stream stopped updating and that chat messages would also randomly stop feeding through.

### tl;dr
- Number of people watching timeout check is a one-time thing now.
- `LiveStreamChatMessage`s that fail to decode will simply be discarded.
- We don't care about the format of the `userId` anymore on these messages.

### For historical purposes:

We chatted about this on Slack but the reason number of people was failing after a little while is because I had a 10 second timeout which would basically check if there was no response from that firebase path within 10 seconds of loading the stream it should switch to HLS. I didn’t realize however that it would expect a value _every_ 10 seconds or it would consider it to have timed out. So I’ve now just moved things around so that the timeout is a one-time check which stops once a value is received.

Also the reason chat messages stopped coming through was because one failed to decode so it error’d out the `SignalProducer`. In any case the fix for now is to just discard messages that fail to decode and to ignore the format of the `userId`.

The reason that the decode failed though is someone chatted with an unexpected `userId` format, this is the failing message:

```
{
  "message" : "i can hear you just fine in this good place",
  "name" : "s-d--p",
  "profilePic" : "https://s3.amazonaws.com/huzza-web/users/avatars/000/192/920/medium/open-uri20161110-28-5z15ob?1478820199",
  "timestamp" : 1493236931735,
  "userId" : "tGcsBej5jAcq7lAGvqliguL6deC3"
}
```

We expect user IDs to be an integer prefixed with `id_` like `id_12345` so this was failing.

Trello: https://trello.com/c/JO0a24aY/1994-live-view-count-messages-bug

### DIFF
--- a/LiveStream/Service/LiveStreamService.swift
+++ b/LiveStream/Service/LiveStreamService.swift
@@ -282,12 +282,13 @@ public struct LiveStreamService: LiveStreamServiceProtocol {
         )
 
         query.observe(.childAdded, with: { snapshot in
-          let chatMessage = (snapshot as FIRDataSnapshot?)
+          let tryChatMessage = (snapshot as FIRDataSnapshot?)
             .flatMap { $0 }
             .map(LiveStreamChatMessage.decode)
             .flatMap { $0.value }
             .map(Event<LiveStreamChatMessage, LiveApiError>.value)
-            .coalesceWith(.failed(.chatMessageDecodingFailed))
+
+          guard let chatMessage = tryChatMessage else { return }
 
           observer.action(chatMessage)
         })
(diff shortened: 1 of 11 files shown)


LiveStream/Service/LiveStreamService.swift
```diff
@@ -282,12 +282,13 @@ public struct LiveStreamService: LiveStreamServiceProtocol {
...
         query.observe(.childAdded, with: { snapshot in
-          let chatMessage = (snapshot as FIRDataSnapshot?)
+          let tryChatMessage = (snapshot as FIRDataSnapshot?)
             .flatMap { $0 }
             .map(LiveStreamChatMessage.decode)
             .flatMap { $0.value }
             .map(Event<LiveStreamChatMessage, LiveApiError>.value)
-            .coalesceWith(.failed(.chatMessageDecodingFailed))
+
+          guard let chatMessage = tryChatMessage else { return }
```
justinswart: It can error if firebase 'fails to initialize'

PR #143, opened 2017-04-30
Limit chat message input length

justinswart:
### What

Chat messages are limited to 140 characters but this isn't indicated in the UI. Messages longer than 140 characters are discarded by Firebase and an error is returned. So to improve this experience I've:

- Limited the input field to stop accepting text at 140 characters.
- Added a small label to indicate remaining characters.

It looks like this:


Library/ViewModels/LiveStreamChatViewModel.swift
```diff
@@ -204,6 +216,12 @@ LiveStreamChatViewModelOutputs {
       return AppEnvironment.current.currentUser != nil
     }
 
+    self.textFieldShouldChangeCharactersInRangeReturnProperty <~
+      textFieldShouldChangeCharactersInRangeProperty.signal.skipNil().map { currentString, range, string in
+        return (currentString as NSString)
+          .replacingCharacters(in: range, with: string).characters.count <= maxMessageLength
```
stephencelis: I always thought this delegate method was a bit awkward, and in this case a paste that brings things over the limit won't be applied/truncated. A couple solutions:

1. Allow users to go over the limit (a la Twitter), but turn the counter red and disable submission.
2. Have a VM output tracking what the current chat input is and use `.editingChanged` as an input that can handle the truncation.

The former seems a bit less troublesome to implement.

Library/ViewModels/LiveStreamChatViewModel.swift
```diff
@@ -225,6 +243,18 @@ LiveStreamChatViewModelOutputs {
           return Strings.Something_went_wrong_please_try_again()
         }
     }
+
+    self.chatInputViewMessageLengthCountLabelText = Signal.merge(
+      self.textProperty.signal,
+      self.viewDidLoadProperty.signal.mapConst("")
+      )
+      .map { $0.coalesceWith("") }
+      .map { "\($0.characters.count)/\(maxMessageLength)" }
```
stephencelis: Curious if this should be localized? Have we checked what Twitter does?

justinswart:
@stephencelis hey thanks for taking a look! You're right, should definitely consider a11y here 👍

Library/ViewModels/LiveStreamChatViewModel.swift
justinswart: True, pasted text over the limit would just do nothing 🤔 

I think `1` would be easiest and I could change the `x/x` to just be a number that decrements and then goes into negatives when it exceeded.

Library/ViewModels/LiveStreamChatViewModel.swift
justinswart: If I do as above and just have the number start at `140` decrement into negatives would localization still be necessary? I'm not sure what Twitter does.

justinswart:
@stephencelis something a little more like this?

Library/ViewModels/LiveStreamChatViewModel.swift
```diff
@@ -42,6 +48,12 @@ public protocol LiveStreamChatViewModelOutputs {
   /// Emits when the keyboard should dismiss on rotate.
   var dismissKeyboard: Signal<(), NoError> { get }
 
+  /// Emits when the message length count label should be hidden.
+  var chatInputViewMessageLengthCountLabelStackViewHidden: Signal<Bool, NoError> { get }
```
theginbin: congrats this might be the longest output name ever, haha!

PR #145, opened 2017-05-03
Handle live video being disabled during stream

justinswart:
During testing of live chat a bug was logged that the video [freezed up](https://trello.com/c/XwnVrJ60/2003-p3-live-video-freezes-while-chat-and-audio-continue) but that audio and chat continued. I did some testing on a not-so-great 3G connection and was able to reproduce this.

It seems that OpenTok's `OTSubscriberDelegate` (which we hadn't previously implemented) sends `subscriberVideoDisabled` and `subscriberVideoEnabled` with a `reason` argument which in this case is `qualityChanged`.

I've changed our `LiveVideoPlaybackState.playing` state to have an associated `Bool` value of `videoEnabled` so that we can handle this.

If OpenTok disables the video we hide the view and display text indicating that video will resume when the connection improves. Although I didn't see it resume during my testing, OpenTok [does say](https://tokbox.com/developer/guides/subscribe-stream/ios/#subscriber_video_disabled) that the video may resume.


Kickstarter-iOS/Views/Controllers/LiveVideoViewController.swift
```diff
@@ -231,3 +242,19 @@ extension LiveVideoViewController: OTSessionDelegate {
     self.viewModel.inputs.sessionDidFailWithError(error: error)
   }
 }
+
+extension LiveVideoViewController: OTSubscriberDelegate {
+  public func subscriberDidConnect(toStream subscriber: OTSubscriberKit) {}
+
+  public func subscriber(_ subscriber: OTSubscriberKit, didFailWithError error: OTError) {}
```
justinswart: This needs to be implemented for protocol conformance but we've not cared about this error'ing yet. I want to consider its effect and implement it in a separate PR.

Kickstarter-iOS/Views/Controllers/LiveStreamContainerViewController.swift
```diff
@@ -93,6 +93,7 @@ public final class LiveStreamContainerViewController: UIViewController {
     _  = self.loaderLabel
       |> UILabel.lens.font .~ .ksr_headline(size: 14)
       |> UILabel.lens.textColor .~ .white
+      |> UILabel.lens.textAlignment .~ .center
```
justinswart: This caused almost all the snapshots for this view controller to have to be re-recorded 😐

justinswart:
testplaybackstates_lang_en_state_playing_videodisabled 2x

Library/ViewModels/LiveVideoViewModelTests.swift
```diff
@@ -10,6 +10,13 @@ private struct TestOTStreamType: OTStreamType {
   fileprivate let streamId: String
 }
 private class TestOTErrorType: NSError, OTErrorType {}
+private struct TestOTSubscriberVideoEventReasonType: OTSubscriberVideoEventReasonType {
```
mbrandonw: don't think we need to suffix this with `Type` since it's a concrete implementation

Library/ViewModels/LiveStreamContainerViewModel.swift
```diff
@@ -416,14 +433,20 @@ LiveStreamContainerViewModelInputs, LiveStreamContainerViewModelOutputs {
       replayState.mapConst(Strings.The_replay_will_start_soon()),
       nonStarterState.mapConst(Strings.No_replay_is_available_for_this_live_stream()),
       self.viewDidLoadProperty.signal.mapConst(Strings.Loading()),
-      self.showErrorAlert
-    )
+      self.showErrorAlert,
+      videoEnabled.filter { !$0 }.mapConst(localizedString(
```
mbrandonw: could `filter(isFalse)` if you wanted to point-free.

Library/ViewModels/LiveVideoViewModelTests.swift
justinswart: ah right, lemme tidy that up

Library/ViewModels/LiveStreamContainerViewModel.swift
justinswart: of course! thanks 👍

PR #147, opened 2017-05-03
Auto-capitalize sentences in chat

justinswart:
Just a quick change to auto-capitalize sentences entered during chat.

Also ran `make strings` to get some new localizations for live stream copy.

Updated relevant screenshot tests.


justinswart:
Thanks @theginbin!

PR #149, opened 2017-05-08
Login to message creator

justinswart:
We have a small bug in that users can tap to message creators while they are logged out but then upon sending their message receive an error saying that they need to be logged in ([Trello](https://trello.com/c/nISUH4vr/1963-logged-out-ask-a-question-button-allows-user-to-start-drafting-message)).

This PR just adds the fix to present the `LoginToutViewController` when attempting to message the creator while logged out. It seems that there was a `.messageCreator` `LoginIntent` for this already so I just used that.

### DIFF
--- a/Library/ViewModels/ProjectCreatorViewModel.swift
+++ b/Library/ViewModels/ProjectCreatorViewModel.swift
@@ -63,8 +63,8 @@ ProjectCreatorViewModelOutputs {
 
     self.goToMessageDialog = project
       .takeWhen(messageCreatorRequest)
-      .map { (MessageSubject.project($0), .projectPage) }
       .filter { _ in AppEnvironment.current.currentUser != nil }
+      .map { (MessageSubject.project($0), .projectPage) }
 
     self.goToSafariBrowser = navigationAction
       .filter { $0.navigationType == .linkActivated }
--- a/Library/ViewModels/ProjectCreatorViewModel.swift
+++ b/Library/ViewModels/ProjectCreatorViewModel.swift
@@ -17,6 +17,9 @@ public protocol ProjectCreatorViewModelInputs {
 }
 
 public protocol ProjectCreatorViewModelOutputs {
+  /// Emits when the LoginToutViewController should be presented.
+  var goToLoginTout: Signal<LoginIntent, NoError> { get }
+
   /// Emits when we should navigate to the message dialog.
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ProjectCreatorViewModelTests.swift
```diff
@@ -65,6 +117,47 @@ final class ProjectCreatorViewModelTests: TestCase {
...
+        navigationType: .linkActivated,
+        request: messagesRequest,
+        sourceFrame: WKFrameInfoData(mainFrame: true, request: messagesRequest),
+        targetFrame: WKFrameInfoData(mainFrame: true, request: messagesRequest)
+      )
+    )
+    XCTAssertEqual(.cancel, policy)
+
+    self.goToMessageDialogContext.assertValues([])
+    self.goToMessageDialogSubject.assertValues([])
```
theginbin: For simplicity, you could just test this output in your `testGoToLoginTout_LoggedOut` function and just add this one line to show that the message dialog isn't opened. But no bigs if this separate function feels more clear to you.

Library/ViewModels/ProjectCreatorViewModel.swift
```diff
@@ -54,9 +57,14 @@ ProjectCreatorViewModelOutputs {
...
       .map { $0.navigationType == .other ? .allow : .cancel }
 
+    self.goToLoginTout = messageCreatorRequest.ignoreValues()
+      .filter { AppEnvironment.current.currentUser == nil }
+      .map { .messageCreator }
+
     self.goToMessageDialog = project
       .takeWhen(messageCreatorRequest)
       .map { (MessageSubject.project($0), .projectPage) }
+      .filter { _ in AppEnvironment.current.currentUser != nil }
```
justinswart: good call!

Library/ViewModels/ProjectCreatorViewModelTests.swift
justinswart: Makes sense, thanks!

mbrandonw:
nice! one additional suggestion. it's not too hard to also add the functionality of automatically opening the message dialog after a successful log in. that way the user doesn't have to tap the "message creator" button again. wanna give that a shot?

justinswart:
ah good idea, will do another PR for that 👍

PR #150, opened 2017-05-08
Nil delegate when removing subscriber view

justinswart:
Quick fix for a bug that crept in with #145 - because we now do set the `delegate` on our `OTSubscriber` views we need to nil that delegate just before removing the subscribers.


justinswart:
Merging with confidence!

PR #151, opened 2017-05-09
Update comments fix

justinswart:
We have a bug that causes Project comments to be shown on Updates when navigating to the `CommentsViewController` from a push notification ([Trello](https://trello.com/c/0ZMsHlpQ/1976-p2-update-comment-push-takes-you-creator-to-project-comments)).

I'm not sure if I'm understanding this 💯 but I think this may fix it.

As far as I can tell update comments are shown when configuring `CommentsViewModel` with an `Update` and a `nil` `Project` and vice versa.

Changes I've made:

- When navigating to `CommentsViewController` from the push notification only for Updates, only configure with an `Update`.
- Add an additional response to `MockService` so that we can differentiate between comments being fetched for a `Project` vs an `Update`. I think that tests weren't picking this up before because both used the same response, by changing this I could fail the test and see that `fetchComments(update:)` was no being called (Will open KsApi PR as well if we're going this route).

Suggestions welcomed if this is not entirely the correct solution!


mbrandonw:
oh nice, haven't looked through yet but can you also PR yer ksapi changes https://github.com/kickstarter/ios-ksapi/tree/update_comments_response. thanks for converting that to a proper result type!

justinswart:
Gonna merge this and PR the type change separately as it does change a few things which would be better to review in isolation, I think.

PR #154, opened 2017-05-10
Backer dash header

theginbin:
### WHAT

Some improvements to the Backer Dashboard, including collapsible header, swiping between tabs, and bug fixes.

profile-collapse-swipe


Library/ViewModels/BackerDashboardViewModel.swift
```diff
@@ -106,18 +122,25 @@ public final class BackerDashboardViewModel: BackerDashboardViewModelType, Backe
...
+      swipedToTab,
       self.backedProjectsButtonTappedProperty.signal.mapConst(.backed),
       self.savedProjectsButtonTappedProperty.signal.mapConst(.saved)
     )
 
+    self.currentSelectedTabProperty <~ self.navigateToTab
+
     self.setSelectedButton = Signal.merge(
-      self.backedButtonTitleText.skip(first: 1).mapConst(.backed),
+      self.backedButtonTitleText.skip(first: 1).mapConst(.backed).take(first: 1),
```
theginbin: For clarity, let me explain this line: 
We prefix the user fetch with the current user, which doesn't contain the number of backed projects (hi graphql!), so we have to skip the first emission and wait for the api to return with the full user data. I am adding the `.take(first: 1)` because when the user leaves this screen, we want to keep their place. In the current version, this is a bug - it resets the tab and indicator view to always select "backed".

justinswart:
This looks cool! Did you consider a `UITableView` with a header view and a section header view for the tabs? You'd get the pinning at the top for free and wouldn't need any pan gesture or animation code.

theginbin:
@justinswart not sure i understand how you could switch between table views that way? If you use the same table view and just refresh the data, then you wouldn't be able to save your place after scrolling in the different sections. Know what I mean?

justinswart:
@theginbin aah, good point! I guess it wouldn't work in this scenario.

Library/ViewModels/BackerDashboardProjectsViewModelTests.swift
```diff
@@ -55,8 +55,8 @@ internal final class BackerDashboardProjectsViewModelTests: TestCase {
       self.emptyStateProjectsType.assertValues([.backed])
       self.isRefreshing.assertValues([true, false])
 
-      self.vm.inputs.viewWillAppear(false)
-      self.isRefreshing.assertValues([true, false, true])
+      self.vm.inputs.viewWillAppear(true)
+      self.isRefreshing.assertValues([true, false, true], "Projects refresh after swipe animation.")
```
theginbin: It does keep your place after paginating and swiping between tabs (i've also tested swiping with paginating through the project navigator and all seems well), but I will doublecheck paginating several times with swipes.

theginbin: @mbrandonw so you're right, it jumps back anywhere from 1 to 6 cells back after reloading OR paginating. I didn't see it because I had `DiscoveryParams.lens.perPage` set to `20` for our test account with fewer backed projects. The higher I set this value, the better it seems. Thoughts on how to keep the positioning other than loading more projects up front?

PR #156, opened 2017-05-11
Prevent discovery filter crash on network error

justinswart:
I found a bug while trying to fix other bugs that I was attempting to reproduce in airplane mode. This uncovered the fact that tapping the Discovery Filter without any network connection could crash the app as it attempts to remove the row for the loader indicator that was never added if the network request immediately failed.

Changes in this PR:
- Only return an `IndexPath` to delete if a row was really removed.
- Only animate the row deletion if there's a row to delete.
- Reload the table view after adding the loader indicator row to be sure.

I also noticed that in the corresponding test file the VM was being used directly and not via it's protocol type so just corrected that 👍

### DIFF
--- a/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
+++ b/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
@@ -67,7 +67,7 @@ internal final class DiscoveryFiltersDataSource: ValueCellDataSource {
   }
 
   internal func deleteCategoriesLoaderRow() -> [IndexPath]? {
-    if self[section: Section.categoriesLoader.rawValue].count > 0 {
+    if !self[section: Section.categoriesLoader.rawValue].isEmpty {
       self.clearValues(section: Section.categoriesLoader.rawValue)
 
       return [IndexPath(row: 0, section: Section.categoriesLoader.rawValue)]
--- a/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
+++ b/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
@@ -66,8 +66,9 @@ internal final class DiscoveryFiltersDataSource: ValueCellDataSource {
              inSection: Section.categoriesLoader.rawValue)
   }
 
-  internal func deleteCategoriesLoaderRow() -> [IndexPath]? {
-    if !self[section: Section.categoriesLoader.rawValue].isEmpty {
+  internal func deleteCategoriesLoaderRow(_ tableView: UITableView) -> [IndexPath]? {
+    if self.numberOfSections(in: tableView) > Section.categoriesLoader.rawValue
+      && !self[section: Section.categoriesLoader.rawValue].isEmpty {
       self.clearValues(section: Section.categoriesLoader.rawValue)
 
       return [IndexPath(row: 0, section: Section.categoriesLoader.rawValue)]
--- a/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
+++ b/Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
@@ -66,10 +66,14 @@ internal final class DiscoveryFiltersDataSource: ValueCellDataSource {
              inSection: Section.categoriesLoader.rawValue)
   }
 
-  internal func deleteCategoriesLoaderRow() -> [IndexPath] {
-    self.clearValues(section: Section.categoriesLoader.rawValue)
+  internal func deleteCategoriesLoaderRow() -> [IndexPath]? {
+    if self[section: Section.categoriesLoader.rawValue].count > 0 {
+      self.clearValues(section: Section.categoriesLoader.rawValue)
 
-    return [IndexPath(row: 0, section: Section.categoriesLoader.rawValue)]
+      return [IndexPath(row: 0, section: Section.categoriesLoader.rawValue)]
+    }
+
+    return nil
   }
 
   internal func selectableRow(indexPath: IndexPath) -> SelectableRow? {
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewController.swift
```diff
@@ -200,10 +201,11 @@ internal final class DiscoveryFiltersViewController: UIViewController, UITableVi
...
   private func deleteCategoriesLoaderRow() {
-    self.filtersTableView.beginUpdates()
+    guard let deleteCategoriesLoaderRow = self.dataSource.deleteCategoriesLoaderRow() else { return }
 
-    self.filtersTableView.deleteRows(at: self.dataSource.deleteCategoriesLoaderRow(), with: .fade)
+    self.filtersTableView.beginUpdates()
+    defer { self.filtersTableView.endUpdates() }
 
-    self.filtersTableView.endUpdates()
+    self.filtersTableView.deleteRows(at: deleteCategoriesLoaderRow, with: .fade)
```
mbrandonw: does `deleteRows` crash if you use an empty array?

I bet it does. cause table views are the worst :/

Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
```diff
@@ -66,10 +66,14 @@ internal final class DiscoveryFiltersDataSource: ValueCellDataSource {
              inSection: Section.categoriesLoader.rawValue)
   }
 
-  internal func deleteCategoriesLoaderRow() -> [IndexPath] {
-    self.clearValues(section: Section.categoriesLoader.rawValue)
+  internal func deleteCategoriesLoaderRow() -> [IndexPath]? {
+    if !self[section: Section.categoriesLoader.rawValue].isEmpty {
```
mbrandonw: is this susceptible to another crash of accessing an index that doesn't exist?

theginbin:
sorry i didn't see this for some reason. looking now!

Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewController.swift
```diff
@@ -60,6 +60,7 @@ internal final class DiscoveryFiltersViewController: UIViewController, UITableVi
       .observeValues { [weak self] isVisible in
         if isVisible {
           self?.dataSource.loadCategoriesLoaderRow()
+          self?.filtersTableView.reloadData()
```
theginbin: I believe there was a conscious decision to not call `reloadData` here, but I don't quite remember why, but I see that this is essentially the fix. Have you tested on a device to be sure things look ok if you first go to the list, let it load, then select a category, and come back to the filters list? I did some delaying of the reloading in this function `reloadCategories(selectedRowId: Int?)`. It's entirely possible this is ok, but I would want to be sure it's visually tested. 

Another thing you might want to try for extra safety and slightly better UX... I see in the VM that this output is called when the categories fetch terminates. Maybe a better place for the `loaderIndicatorIsVisible` output to emit `false` is when `categoriesEvent.values()` is emitted instead of on the `terminated` state of the `SignalProducer`. Since we don't handle the error state with a message or something, it seems like better feedback to just keep the loading indicator spinning if the categories never return. What do you think?

Library/ViewModels/DiscoveryFiltersViewModelTests.swift
```diff
@@ -44,7 +44,7 @@ private let filmExpandableRow = expandableRowTemplate
 private let categories = [ Category.art, .illustration, .filmAndVideo, .documentary ]
 
 internal final class DiscoveryFiltersViewModelTests: TestCase {
-  private let vm = DiscoveryFiltersViewModel()
+  private var vm: DiscoveryFiltersViewModelType = DiscoveryFiltersViewModel()
```
theginbin: why not keep `let`? p.s. thank youuuu for adding the protocol!

justinswart: Ah, I changed it to `var` because I moved the instantiation to `setUp()` but forgot to change it back to `let`, thanks!

Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewController.swift
justinswart: it does! added to the `guard` to check that it's not empty.

Kickstarter-iOS/DataSources/DiscoveryFiltersDataSource.swift
justinswart: It could, yes.

I don't have direct access to `self.values` in the subclass of `ValueCellDataSource` though, could add a function to check if it's empty or of a section exists?

Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewController.swift
justinswart: Agree, I've changed to to hide on `values()`. Tested it on my device and behaviour seems the same with/without the reloading!

theginbin: cool!

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -91,12 +90,16 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
...
-              terminated: { loaderIsVisible.value = false })
+          .on(starting: { loaderIsVisible.value = true } )
           .map { $0.categories }
           .materialize()
       }
 
+    self.loadingIndicatorIsVisible = Signal.merge(
+      loaderIsVisible.signal,
+      categoriesEvent.values().mapConst(false)
+    )
```
theginbin: noice!

PR #158, opened 2017-05-12
Make an allow list of deep linkable routes

mbrandonw:
Right now our routing systems has a big ole list of URLs on Kickstarter that we want to recognize in a variety of situations, e.g. when inspecting requests in a web view, and when deep linking into the app.

Sometimes we want to use a different set of recognizable routes depending on the situation we are routing. In particular, when deep linking there is a smaller set of routes we want to recognize, and then anything we don't recognize can bump the user back to mobile safari.

That's what this PR does...

There are smarter ways to do this, and would love to try it out soon, but this will do the job and is well tested!

### DIFF
--- a/Library/Navigation.swift
+++ b/Library/Navigation.swift
@@ -211,7 +211,14 @@ public func == (lhs: Navigation.User, rhs: Navigation.User) -> Bool {
 
 extension Navigation {
   public static func match(_ url: URL) -> Navigation? {
-    return routes.reduce(nil) { accum, templateAndRoute in
+    return allRoutes.reduce(nil) { accum, templateAndRoute in
+      let (template, route) = templateAndRoute
+      return accum ?? parsedParams(url: url, fromTemplate: template).flatMap(route)?.value
+    }
+  }
+
+  public static func deepLinkMatch(_ url: URL) -> Navigation? {
+    return deepLinkRoutes.reduce(nil) { accum, templateAndRoute in
       let (template, route) = templateAndRoute
       return accum ?? parsedParams(url: url, fromTemplate: template).flatMap(route)?.value
     }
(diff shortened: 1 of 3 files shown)


Library/Navigation.swift
```diff
@@ -260,6 +267,33 @@ private let routes: [String:(RouteParams) -> Decoded<Navigation>] = [
   "/users/:user_param/surveys/:survey_response_id": userSurvey
 ]
 
+private let deepLinkRoutes: [String:(RouteParams) -> Decoded<Navigation>] = allRoutes.allValues(
```
mbrandonw: this is the allow-list of routes we'll use for deep linking. I'm using this lil private helper `Dictionary.allValues(from:)` to pluck out a certain set of keys from the `allRoutes` dictionary.

open to other names!

mbrandonw: oh actually, using Hoogle I found this function:

```haskell
restrictKeys :: Ord k => Map k a -> Set k -> Map k a
```

which is a better name and a better definition! maybe we should even add this to the prelude...

Library/Navigation.swift
```diff
@@ -260,6 +267,33 @@ private let routes: [String:(RouteParams) -> Decoded<Navigation>] = [
...
+    "/projects/:creator_param/:project_param/creator_bio",
+    "/projects/:creator_param/:project_param/dashboard",
+    "/projects/:creator_param/:project_param/description",
+    "/projects/:creator_param/:project_param/friends",
+    "/projects/:creator_param/:project_param/posts",
+    "/projects/:creator_param/:project_param/posts/:update_param",
+    "/projects/:creator_param/:project_param/posts/:update_param/comments",
+    "/projects/:creator_param/:project_param/surveys/:survey_param",
+    "/users/:user_param/surveys/:survey_response_id"
+  ]
```
theginbin: nice! only suggestion would be to put these keys into an enum since a string could potentially change in one of these places and not the other.

Issue #164, opened 2017-05-18
naming nitpick: "ViewModel"


AndrewSB:
I see that you use the name `ViewModel` for objects that have no inherent view (e.g. the `AppDelegateViewModel`) and I'm confused as to why. Can you explain why you chose to name objects with that role as `ViewModel`s?

To me, your `ViewModel`s seem like MVC controllers since they do many MVC control-y things; they don't simply map models into a viewable format. That comes with it's own host of issues of bloated controllers: anything that isn't view or data related becomes controller territory - persistence, networking, etc.
The iOS community has become very trigger happy calling objects that reduce code in a view controller a "view model"  ever since MVVM became popular

From my understanding, it seems like you folks are using your view controllers simply as views that ferry messages back and forth from a view model, and your view models are making all of the hard decisions. 
That to me no longer sounds like an object that maps data -> view (a view model), and more like a controller (terribly broad name, I agree)

The reason I'm asking is not to beat a dead horse or go into a naming holy war (please stop me if that's the direction this is heading in), but because I'm genuinely curious as to why you decided to classify your objects in the way you did 😛

mbrandonw:
lol boy oh boy have we talked about this quite a bit! we honestly don't really care about the name at all. we see our objects as just sugar around transforming user input signals to user observable output signals. it makes things easy to test (we have over 1,000 tests with over 7,000 assertions). we've even gone as far to model these things as just a pure function from input signals to output signals, like our [paginate function](https://github.com/kickstarter/ios-oss/blob/master/Library/Paginate.swift), which handles pagination for the entire app.

so yeah, we really don't wanna think about the naming of this thing at all. just wanna think of it as the "functional core" as mentioned in gary berhardt's talk https://www.destroyallsoftware.com/talks/boundaries

AndrewSB:
😂😂 totally on the same wavelength, I watched a bunch of the 2017 functional swift talks on Tuesday (thank you for hosting), and was planning on re-watching boundaries later today since it's been about a year since I last thought about berhardt's 

Paginate is really well done, it reminds me of the something [similar I wrote](https://gist.github.com/AndrewSB/9585bcb32fb669d64b73b7a9237eed3b), but way more thought out!

I agree with thinking of your view models as the highly-tested functional core of your app, and you're going to get along _fine_ (as you have been) with a less-specific-that-it-could-be name, but names are important.
[...]

PR #165, opened 2017-05-18
[don't merge] why function vs underlying property for ViewModelInput?

AndrewSB:
I was wondering why you used a function that just wraps a `MutableProperty` in a number of your view models, it seems like totally unnecessary boilerplate.

Is it there just to keep the view controller _simple_, and not leak your Reactive abstraction into the view?

I made an example diff of what the alternative I see could be, and the diff pretty much just removes 3 lines of code. 3 x the number of input functions you have would be a massive negative diff 😉

### DIFF
--- a/Library/ViewModels/ActivitiesViewModel.swift
+++ b/Library/ViewModels/ActivitiesViewModel.swift
@@ -6,7 +6,7 @@ import Result
 
 public protocol ActitiviesViewModelInputs {
   /// Called when the project image in an update activity cell is tapped.
-  func activityUpdateCellTappedProjectImage(activity: Activity)
+  var tappedActivityProjectImageSink: MutableProperty<Activity?> { get }
 
   /// Call when the Find Friends section is dismissed.
   func findFriendsHeaderCellDismissHeader()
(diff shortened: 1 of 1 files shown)


stephencelis:
We wrap our private mutable properties in functions to eliminate a few invalid cases. Consider these two lines:

``` swift
self.viewModel.tappedActivityProjectImageSink.value = nil
// Because there's no default value at initialization, it must be MutableProperty<Activity?>,
// but we never want to null it out again, so we want to prevent this at compile-time.

print(self.viewModel.tappedActivityProjectImageSink.value)
// We also only want our sinks to be setters,
// so we don't want to accidentally use it as a getter.
```

We also may move away from `MutableProperty` in favor of `Signal.pipe` again. Haven't started, yet, though! 😄

There's definitely a lot of boilerplate in exposing our inputs this way, but it gives us the level of expression we want. There are a few things we may explore in the future to eliminate some of this boilerplate (_e.g._ the function approach we mentioned in #164), but for now it's a trade-off we've come to embrace.

AndrewSB:
> Because there's no default value at initialization, it must be an optional MutableProperty<Activity?>, but we never want to null it out again, so we want to prevent this at compile-time.

astute
modeling that problem well is hard

> We also only want our sinks to be setters, so we don't want to accidentally use it as a getter.

why not upcast MutableProperty to something that can only be bound to, not read from? Not sure what it's called in rac, but something like a PublishSubject instead, and then you can use the same MutableProperty as the getter as well, à la

```swift 
class ViewModel {
    private var tappedActivityProjectImage: MutableProperty<Activity?>

    // in place of your function
    internal var tappedActivityProjectImageSink: PublishSubject<Activity> {
        // return new subject that forwards to tappedActivityProjectImage, taking care of problem #1
    }

    // if desired, a getter could also be constructed, it isn't required in this case, but if defined in the inputs:
    internal var tappedActivityProjectImageValue: Activity? {
         return tappedActivityProjectImage.value
    }
}
```

EDIT: after composing this response, I realized that I simply proposed an alternative to using functions that has no actual reduction in complexity. I see why the function wrapping the property makes sense, maybe ⬆️ could be used instead, but there's no pressing reason to swift. I agree with your opinion on wrapping the stream to eliminate the invalid cases

PR #169, opened 2017-05-25
Remove a few deep link routes that we don't support

mbrandonw:
There are a few deep link routes I allowed I #158 that aren't really supported. Removing them will allow the app to redirect back to the website...


ktman:
nice. think the messages route is valid though? am currently able to deep-link properly.

mbrandonw:
> nice. think the messages route is valid though? am currently able to deep-link properly.

oh cool, didn't realize. adding back...

justinswart:
Thanks!

PR #172, opened 2017-05-31
Add more disabled swiftlint rules 

justinswart:
There are a number of rules that we tend to disable whenever we encounter them and it seems to make sense to me that we just add these to the list of disabled rules as they're not really ones that we care about.

If we're in agreement then I will also go ahead and remove all of the places that we have done `//swiftlint:disable <redundant_rule_name>`.

Thoughts on keeping any of these or others we'd like to add? 🤔


theginbin:
totally down with these 4!

PR #175, opened 2017-06-06
Fixing facebook login confirmation top margins

bormind:
Fixed by creating _facebookConfirmationStackViewStyle_ based on _loginRootStackViewStyle_
In _facebookConfirmationStackViewStyle_ added extra logic - if current environment (traitCollection) is not isRegularRegular we decide on margins by checking if traitCollection isVerticallyCompact or not.

I believe that this logic is not very direct and difficult to read. And potentially can cause problems in applying horizontal margins in scenario when we have vertically compact but horizontally regular layout.

We do fix this way because in Prelude we have UITraitCollection extension property _isVerticallyCompact_ but we **don't** have property for _isHorizontallyCompact_. Which forces us to make a decision based on combination of _isRegularRegular_ and _isVerticallyCompact_ properties.

In my opinion  better fix would be to add  _isHorizontallyCompact_ property to UITraitCollection in Prelude and use separate logic to decide on vertical margins based on _isVerticallyCompact_ and on horizontal margins based on  _isHorizontallyCompact_.


.gitignore
```diff
@@ -73,3 +73,7 @@ fastlane/screenshots
 Screenshots/FailureDiffs/*
 
 Frameworks/OpenTok/*
+
+# Ignore Jetbrains local project cash directory
```
stephencelis: You mean "cache"? What app generates this file? If it's not iOS workflow-specific wanna add it to your global `~/.gitignore` file instead?

Library/Styles/LoginStyles.swift
```diff
@@ -100,6 +100,17 @@ public let loginRootStackViewStyle =
         : .init(topBottom: Styles.grid(2), leftRight: Styles.grid(3))
 }
 
+public let facebookConfirmationStackViewStyle =
+  UIStackView.lens.layoutMarginsRelativeArrangement .~ true
+    <> UIStackView.lens.layoutMargins %~~ { _, stack in
```
stephencelis: May not be worth the layered logic, but we could reuse `loginRootStackViewStyle` and ensure that the base insets don't go out of sync:

``` swift
public let facebookConfirmationStackViewStyle =
  loginRootStackViewStyle
    <> UIStackView.lens.layoutMargins..UIEdgeInsets.lens.topBottom %~~ { topBottom, stack in
      !stack.traitCollection.isRegularRegular && stack.traitCollection.isVerticallyCompact
        ? Styles.grid(2)
        : topBottom
}
```

bormind: Thank you @stephencelis! I like it much more than new style. I will make the change and will resubmit..

Library/Styles/LoginStyles.swift
```diff
@@ -100,6 +100,17 @@ public let loginRootStackViewStyle =
...
+      stack.traitCollection.isRegularRegular
+        ? .init(topBottom: Styles.grid(10), leftRight: Styles.grid(20))
+        : .init(
+          topBottom: stack.traitCollection.isVerticallyCompact ? Styles.grid(2) : Styles.grid(10),
+          leftRight: Styles.grid(3)
```
theginbin: Hey @bormind we want to keep all the login flow views consistent with margin padding. That means Login, ForgotPassword, SignUp, and Facebook views should all be using the same top margin padding (`Styles.grid(2)`) and thus share the `loginRootStackViewStyle`. Have you looked into the storyboard to see what else might be affecting the padding here? 

Also, if you are going to change landscape layouts for one view in a shared flow, then it is best to alter all views in this flow to be consistent.

bormind: @theginbin tank you for reviewing this! This was my original thought and fix - directly in loginRootStackViewStyle to be consistent between all login views . But then we discussed it with @justinswart and decided to do it only for facebookConfirmation because we thought that we may want different margins for different login views. 

Based on @stephencelis  suggestion (above) at the moment I implemented it as an expansion on loginRootStackViewStyle like this:
```swift
public let facebookConfirmationStackViewStyle = loginRootStackViewStyle
    <> UIStackView.lens.layoutMargins..UIEdgeInsets.lens.topBottom %~~ { (topBottom, stack)  in
      return (stack.traitCollection.isVerticallyCompact ? Styles.grid(2) : topBottom.0, topBottom.1)
}    
```
But if our intension is to be consistent between all login views I probably should go back to my original fix..
Should we have a quick discussion together with @justinswart, and then I will do the fix base on what we decide?

justinswart:
Thanks for the work on this @bormind but as we determined that the padding shouldn't change right now I'm going to close this. Good that you got to dig into this part of the code 👍

PR #176, opened 2017-06-07
One pass availability reporting

bormind:
replaced property _1password_extension_available_ with property _one_password_extension_available_ for _Viewed Login_ tracking event. Kept property _1password_extension_available_ for _User Login_ event intact because it is marked as deprecated and will be removed..  

Reference Trello card: Update 1password property


justinswart:
Nice Boris! Merge away 👍

PR #178, opened 2017-06-09
Rename copyright notice to have markdown extension.

frewsxcv:
Seems to formatted as markdown, so rename to end in '.md'.


frewsxcv:
e.g. https://github.com/kickstarter/ios-oss/blob/frewsxcv/notice-markdown/NOTICE.md

PR #180, opened 2017-06-13
Strip category colors

theginbin:
# What

Stripped out the category colors and gradients and updated all related tests and screenshots. Tons of deletions here so some 🦅  👀  would be handy. Many screenshots updated, so will do a review with @maggieKSR.


Library/ViewModels/DiscoveryNavigationHeaderViewModel.swift
```diff
@@ -172,17 +171,14 @@ public final class DiscoveryNavigationHeaderViewModel: DiscoveryNavigationHeader
     self.showDiscoveryFilters = rowForFilters
       .takeWhen(paramsAndFiltersAreHidden.filter { !$0.filtersAreHidden })
 
-    self.subviewColor = primaryColor
+    self.subviewColor = categoryId.mapConst(discoveryPrimaryColor())
```
justinswart: Could `mapConst` on `viewDidLoad` here to be explicit about it not depending on `categoryId` perhaps.

justinswart:
screen shot 2017-06-13 at 6 05 56 pm 👍

Kickstarter-iOS/Views/Cells/NoRewardCell.swift
```diff
@@ -12,9 +12,7 @@ internal final class NoRewardCell: UITableViewCell, ValueCell {
   @IBOutlet fileprivate weak var rootStackView: UIStackView!
   @IBOutlet fileprivate weak var copyStackView: UIStackView!
 
-  internal func configureWith(value project: Project) {
-    self.contentView.backgroundColor = Library.backgroundColor(forCategoryId: project.category.rootId)
-  }
+  internal func configureWith(value: Void) {}
```
theginbin: yup exactly. we use it in a few other places, but it's not common.

Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewController.swift
```diff
@@ -32,18 +32,15 @@ internal final class DiscoveryFiltersViewController: UIViewController, UITableVi
...
-    self.backgroundGradientView.endPoint = CGPoint(x: 1.0, y: 0.0)
-
     self.closeButton.addTarget(self, action: #selector(closeButtonTapped), for: .touchUpInside)
 
     self.viewModel.inputs.viewDidLoad()
   }
 
-  override func viewWillAppear(_ animated: Bool) {
-    super.viewWillAppear(animated)
+  override func viewDidAppear(_ animated: Bool) {
```
theginbin: yeah, i tweaked a lot of the animation because the nav header was animating some color transitions at the same time. in the end, decided we shouldn't be animating on `viewWillAppear` because it's starting the animation too early. Better to animate after `viewDidAppear` is called.

Library/ViewModels/DiscoveryNavigationHeaderViewModel.swift
theginbin: totally. good call.

theginbin: i decided to get rid of this output altogether!

Kickstarter-iOS/Views/Cells/NoRewardCell.swift
stephencelis: Might be worth revisiting naming here:

``` diff
-configureWith(value:)
+configure(with:)
```

That way, we could define the function with a default:

```diff
-configureWith(value: Void) {}
+configure(with value: Void = ()) {}
```

And call it as such:

```diff
-cell.configureWith(value: ())
+cell.configure()
```

theginbin:
nice suggestion @stephencelis! i could do that in a follow up PR since it will be a large diff.

stephencelis:
:+1: Yep! Didn't mean to suggest it for this PR. Just a fun iteration for the future.

Library/Styles/BaseStyles.swift
```diff
@@ -81,6 +81,18 @@ public func dropShadowStyle <V: UIViewProtocol> (radius: CGFloat = 2.0,
...
+  return dropShadowStyle(radius: 5.0, offset: .init(width: 0, height: 2.0))
+    <> V.lens.layer.shadowOpacity .~ 0.17
+    <> V.lens.layer.shadowColor .~ UIColor.black.cgColor
+}
+
+public func dropShadowStyleLarge <V: UIViewProtocol> () -> ((V) -> V) {
+  return dropShadowStyle(radius: 6.0, offset: .init(width: 0, height: 3.0))
+    <> V.lens.layer.shadowOpacity .~ 0.17
+    <> V.lens.layer.shadowColor .~ UIColor.black.cgColor
+}
```
theginbin: just a note: adding these medium and large styles in for future work. Soon, we'll be updating the base drop shadow style with the black color and lighter opacity so we can remove these last 2 lines from these functions once that's done.

PR #186, opened 2017-06-21
Strings latest 001

theginbin:
# What

Updated to our latest strings and revised screenshots.

# Why

It's been a while and this will help trim down PRs in the works that need new strings, but have to update unrelated screenshot tests along with them.

**one caveat**
We lost some of the French translation in the Discovery "Saved" empty state because it was removed from `en.yml` for another PR. I reverted it, but the French still needs to propagate again. That does mean a couple screenshots will need to update again soon.


theginbin:
@justinswart yes just a `make strings` and recording some new screens. i may fix the iphone4 postcard cell that goes to a second line.

PR #188, opened 2017-06-23
Push notification opt-in / opt-out tracking

bormind:
# What

Implemented tracking of the user opt-in/opt-out selection when presented with Push Notification authorization dialog. **Implemented for iOS 10 and higher**

# Why

We want to know when user is prompted with authorization for Push Notification dialog and what was the users selection 

# How
In iOS 10 there is new API in `UNUserNotificationCenter` that allows us to `UNUserNotificationCenter.current().requestAuthorization` and in the passed closure to obtain user selection by using `UNUserNotificationCenter.current().getNotificationSettings`

# See 👀

[Trello](https://trello.com/c/tM6uHXj2/366-boris-system-push-opt-in-instrumentation)
[Article discussing the new UNUserNotificationCenter usage](https://www.raywenderlich.com/156966/push-notifications-tutorial-getting-started)

# TODO ⏰

Discussion maybe needed: 
Created helper [count function](https://github.com/kickstarter/ios-oss/compare/register_push_notification_tracking?expand=1#diff-518138b58fb37d89eecb3430944d46c8R633) to make test code more explicit but not sure if something like this not already exists in some of our libraries

Discussion maybe needed: 
Not sure that [approach that I used](https://github.com/kickstarter/ios-oss/compare/register_push_notification_tracking?expand=1#diff-7ed943722eedb7b8135b3a8b29460867R556) for signal to track user selection in the authorization dialog is the best one.


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -214,13 +234,36 @@ AppDelegateViewModelOutputs {
...
+          self.notificationAuthorizationCompletedProperty.signal)
+
+      self.registerForRemoteNotifications = self.notificationAuthorizationStatusProperty.signal
+        .skipNil()
+        .filter { $0.isAuthorized }
+        .ignoreValues()
+
+    } else {
+      //Never firing signal in ios 9
+      self.getNotificationAuthorizationStatus = Signal<(), NoError>.pipe().output
```
stephencelis: I think this can be:

``` swift
self.getNotificationAuthorizationStatus = .empty
```

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -561,34 +566,140 @@ final class AppDelegateViewModelTests: TestCase {
...
+      self.vm.inputs.notificationAuthorizationStatusReceived(MockNotificationAutorizationStatus.notDetermined)
+
+      XCTAssertEqual(1, client.events |> count("Triggered Push Permissions Dialog"))
+      XCTAssertEqual(0, client.events |> count("Confirmed Push Opt-In"))
+      XCTAssertEqual(0, client.events |> count("Dismissed Push Opt-In"))
+
+      self.vm.inputs.notificationAuthorizationCompleted()
+      self.vm.inputs.notificationAuthorizationStatusReceived(MockNotificationAutorizationStatus.authorized)
+
+      XCTAssertEqual(1, client.events |> count("Triggered Push Permissions Dialog"))
```
stephencelis: This is clever, but I'm thinking we probably wanna be more explicit in client event ordering, which we lose here.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -146,6 +165,7 @@ public protocol AppDelegateViewModelOutputs {
 
   /// Emits a config value that should be updated in the environment.
   var updateConfigInEnvironment: Signal<Config, NoError> { get }
+
```
stephencelis: Lil space snuck in.

stephencelis:
Looks like some test failures to check out as well!

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -15,6 +15,31 @@ import ReactiveSwift
...
+    }
+  }
+
+  public var isNotDetermined: Bool {
+    switch self {
+    case .notDetermined: return true
+    case .denied, .authorized: return false
+    }
+  }
+}
```
theginbin: Can you explain why you need to make an extension of `UNAuthorizationStatus` when the enum cases `authorized`, `denied`, and `notDetermined` already exist? Can you just check for these cases instead of working off a `Bool` that wraps these built-in cases?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -21,6 +22,12 @@ public func == (lhs: HockeyConfigData, rhs: HockeyConfigData) -> Bool {
     && lhs.userName == rhs.userName
 }
 
+public protocol NotificationAuthorizationStatusType {
+  var isAuthorized: Bool { get }
+  var isDenied: Bool { get }
+  var isNotDetermined: Bool { get }
+}
```
theginbin: Still not sure I understand why we don't use the built-in enum cases. Please elaborate why this is necessary?
And if so, can you make this an `enum` instead of a `protocol`?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -510,6 +553,18 @@ AppDelegateViewModelOutputs {
...
 
+    self.notificationAuthorizationStatusProperty.signal
+      .skipNil()
+      .skip(until: self.notificationAuthorizationCompletedProperty.signal)
+      .take(first: 1)
+      .observeValues {
+        if $0.isAuthorized {
+          AppEnvironment.current.koala.trackPushPermissionOptIn()
+        } else if $0.isDenied {
+          AppEnvironment.current.koala.trackPushPermissionOptOut()
```
theginbin: Again, if we use the `UNAuthorizationStatus` enum then you can use a `switch/case` here and better ensure we cover all the cases.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1378,6 +1489,39 @@ final class AppDelegateViewModelTests: TestCase {
   }
 }
 
+fileprivate enum MockNotificationAutorizationStatus {
+  case notDetermined
+  case denied
+  case authorized
+}
```
theginbin: Not sure this mock version would be necessary?

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
theginbin: Agree with @stephencelis. The order of events is equally important to the count. Yes, it is a little more tedious for the developer, but breaking tests is what helps us better understand the changes we've added to the code.

Kickstarter-iOS/AppDelegate.swift
bormind: This is work around of the issue that `UNAuthorizationStatus` was introduced in iOS 10.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
bormind: @stephencelis, @theginbin 
First of all, thank you for looking in to it - it helps me a lot!
Second, FYI we head similar discussion with @justinswart and I was trying to convince him why I did it this particular way :). And we decided that I will keep it and submit it for the review and then we would have discussion together :) 

If I understand  the issue - you would prefer to see something like this: 
`XCTAssertEqual(["Triggered Push Permissions Dialog"], client.events))` 
instead of this:
`XCTAssertEqual(1, client.events |> count("Triggered Push Permissions Dialog"))`

My argument against `XCTAssertEqual(["Triggered Push Permissions Dialog"], client.events))` is this:

We pass tracking client to our environment before we start "manipulating" our AppDelegateViewModel. During sequence of manipulations on our viewModel it can trigger some tracking events that are not related to what we are trying to test. For example if someone will decide to track `userSessionStarted` events. Than `client.events` will contain events that are not part of our testing logic. If we would try to accommodate this "extra" events in our test cases it will make our test not **autonomous**.
By using cont of appearances of specific events, I avoid testing for any side effects that are not part of our testing goal - make test case  **autonomous**.

justinswart: Thanks for explaining @bormind, we chatted about this a bit on our call but yes as @theginbin and @stephencelis are saying we probably want to be more explicit here about what will be emitted and in which order. I hear what you're saying about the autonomy of the test case but that isn't really a problem. If a change was made later on and more tracking events were added to `userSessionStarted` we would want to know about them and we would want our test to fail.

In fact in such a situation, where more tracking was performed on the session starting, we could actually rename the the test from `testTrackingPushAutorizationOptOut` to `testUserSessionStartedTrackingEvents` for example and test all of the various emissions that we expect to happen at that time. For now this is not necessary.

Lastly, and @stephencelis @theginbin correct me if I'm wrong, but introducing a helper function into a test generally requires that we _also_ test that function (however simple it may be?) otherwise it _can_ introduce another element of uncertainty in the test? Or is this not really an issue?

bormind: For the reference - if we would be comparing arrays directly this is how part of the passing test would looks like: 
```XCTAssertEqual(["App Close", 
"Closed App", 
"App Open", 
"Opened App"], client.events)

self.vm.inputs.notificationAuthorizationCompleted()
self.vm.inputs.notificationAuthorizationStatusReceived(MockNotificationAutorizationStatus.authorized)

XCTAssertEqual(["App Close", 
"Closed App",
 "App Open",
 "Opened App", 
"Confirmed Push Opt-In"], client.events)

File: Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift (hunk 5 above)
Comment (bormind) [inline Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift]: It all boils down to the property 
```
fileprivate let notificationAuthorizationStatusProperty =
    MutableProperty<NotificationAuthorizationStatusType?>(nil)
  public func notificationAuthorizationStatusReceived(
    _ notificationSetting: NotificationAuthorizationStatusType
  ) {
    return self.notificationAuthorizationStatusProperty.value = notificationSetting
  }
```
of
 `AppDelegateViewModel` being specified only for iOS 10. I would prefer having it as single  enum too but couldn't come up with the way of getting away with direct usage of UNAuthorizationStatus enum

File: Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift (hunk 6 above)
Comment (bormind) [inline Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift]: Gina, this is related to your comment above and I agree it is much nice with the exhaustive enum but could't couldn't figure out how to do it.

File: Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift (hunk 2 above)
Comment (stephencelis) [inline Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift]: @bormind VMs account for a lot of complete, composed logic, so I think of them as autonomous units themselves. This means we care about all changes.

I do understand the concern around verboseness and noise vs. signal per test, but there are some other approaches we could consider to alleviate this pain without sacrificing the information lost by the counting approach. We could, for example, assert on just new events after a certain point:

``` swift
let endIndex = client.events.endIndex

self.vm.inputs.notificationAuthorizationCompleted()
self.vm.inputs.notificationAuthorizationStatusReceived(MockNotificationAutorizationStatus.authorized)

XCTAssertEqual(["Confirmed Push Opt-In"],
               client.events.suffix(from: endIndex)
```

It's cleaner, but we run the risk of not testing certain events.

Meanwhile, [Jest](http://facebook.github.io/jest/) has the ability to snapshot data structures, which—were we to build something like it—could provide that additional layer of checking with less boilerplate. In the end, that time investment hasn't seemed worth it (yet) and we generally embrace boilerplate if it gives us more confidence in the correctness of our code.

PR #189, opened 2017-06-23
Modernize backing info

theginbin:
# What

Redesign of the View Pledge screen with some relevant UX improvements.

# Why

We wanted to modernize this page to better match the design of the rest of the app and make backing info a bit clearer.

# Some stuff

- Added empty state messaging on the Messages page.
- Reply button is disabled on an empty Messages page.
- "View Messages" button returns you back to the Messages screen if you came from Messages instead of taking you on a nested nav stack adventure.
- On the Message Thread screen, the "Backing info" button is hidden if you came from the Backing info screen instead of taking you on a nested nav stack adventure.
- In Creator Activity, the "Reply" and "Pledge Info" buttons are hidden on a comment if you are the creator (aka not the backer).
- Revised some logic around collaborator vs creator - if you are a collaborator, you should still be able to message the creator (since you can be a backer too!) so I've kept the "Message Creator" button in that scenario.
- Refactored the `fetchMessageThread` api method for a `backing` request because an empty message thread was not decoding properly without modifying `MessageThreadEnvelope`. Thanks @stephencelis for the pair sesh on this.


Library/ViewModels/BackingViewModel.swift
```diff
@@ -238,45 +250,118 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
-    case .errored:
-      return Strings.project_view_pledge_status_errored()
-    case .pledged:
-      return Strings.project_view_pledge_status_pledged()
-    case .preauth:
-      fatalError()
+private func statusDescString(for backing: Backing, project: Project, backerIsCurrentUser: Bool)
+  -> NSAttributedString {
+
+  var string = ""
```
justinswart: I think if the `switch` is exhaustive this can be a `let string: String` and then it can be instantiated in the `switch` just for fun!

Library/ViewModels/BackingViewModel.swift
```diff
@@ -124,66 +111,91 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
-    self.estimatedDeliveryDateLabelText = reward.map {
-      $0.estimatedDeliveryOn.map {
-        Format.date(secondsInUTC: $0, dateFormat: "MMMM yyyy", timeZone: UTCTimeZone) }
+    self.rewardTitleWithAmount = Signal.merge(
+      emptyStringOnLoad,
+      projectAndBackingAndBackerIsCurrentUser
+        .map { project, backing, _ in
+          let currency = Format.currency(backing.reward?.minimum ?? 0, country: project.country)
+          if let rewardTitle = backing.reward?.title {
+            return currency + " - " + rewardTitle
```
stephencelis: We get a lil more type-safety with `+` since it requires that the left- and righthand sides be strings. Interpolation can lead to some formatting gotchas.

Kickstarter-iOS/Views/Controllers/BackingViewController.swift
```diff
@@ -178,6 +217,13 @@ internal final class BackingViewController: UIViewController {
   }
 
   fileprivate func goToMessages(project: Project, backing: Backing) {
+    if let nav = self.navigationController {
```
stephencelis: I got a bit lost here with the `for` loop and `return` and all. What if we did something like:

``` swift
guard let nav = self.navigationController else { return }

if nav.childViewControllers.contains(where: { $0 is MessagesViewController }) {
  nav.popViewController(animated: true)
  return
}

let vc = MessagesViewController.configuredWith(project: project, backing: backing)
nav.pushViewController(vc, animated: true)
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -124,66 +111,91 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+    self.backerName = basicBacker.map { $0.name }
 
-    self.backerAvatarURL = backer.map { URL(string: $0.avatar.small) }
+    self.backerAvatarURL = basicBacker.map { URL(string: $0.avatar.small) }
 
-    self.backerPledgeStatus = backing
-      .map { Strings.backer_modal_status_backing_status( backing_status: statusString($0.status)) }
-    self.backerPledgeStatusAccessibilityLabel = self.backerPledgeStatus
+    self.pledgeSectionTitle = Signal.merge(
+      emptyStringOnLoad.map { NSAttributedString(string: $0) },
```
stephencelis: Could point-free :D

``` swift
emptyStringOnLoad.map(NSAttributedString.init(string:))
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -124,66 +111,91 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+      projectAndBackingAndBackerIsCurrentUser.map { project, backing, backerIsCurrentUser in
```
stephencelis: Can this be point-free?

``` swift
projectAndBackingAndBackerIsCurrentUser
  .map(pledgeTitle(for:project:backerIsCurrentUser:))
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -124,66 +111,91 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+      }
+    )
 
-    self.backerShippingDescription = reward.map { $0.shipping.summary  ?? "N/A" }
-    self.backerShippingDescriptionAccessibilityLabel = self.backerShippingDescription
+    self.statusDescription = Signal.merge(
+      emptyStringOnLoad.mapConst(NSAttributedString(string: "")),
+      projectAndBackingAndBackerIsCurrentUser
+      .map { project, backing, backerIsCurrentUser in
+        return statusDescString(for: backing, project: project, backerIsCurrentUser: backerIsCurrentUser)
```
stephencelis: Don't need this `return`, right?

Library/ViewModels/BackingViewModel.swift
stephencelis: Could do some fun functional stuff here :D

``` swift
backing.reward?.title.map { currency + " - " + $0 } ?? currency
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -238,45 +250,118 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+  return titleString.simpleHtmlAttributedString(
+    base: [
+      NSFontAttributeName: UIFont.ksr_subhead(size: 13),
+      NSForegroundColorAttributeName: UIColor.black
+    ],
+    bold: [
+      NSFontAttributeName: UIFont.ksr_headline(size: 15),
+      NSForegroundColorAttributeName: UIColor.black
+    ])
+    ?? NSAttributedString(string: "")
```
stephencelis: Think we can do a bare init here:

``` swift
?? .init()
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -238,45 +250,118 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+}
+
+private func rewardTitle(for reward: Reward?, project: Project, backerIsCurrentUser: Bool)
+  -> NSAttributedString {
+
+  guard let reward = reward else { return NSAttributedString(string: "") }
```
stephencelis: We only seem to be using `reward` to grab the delivery estimate. Can we eliminate the variable?

``` swift
guard let estimatedDate = reward?.estimatedDeliveryOn else { return .init() }
```

Library/ViewModels/BackingViewModel.swift
```diff
@@ -238,45 +250,118 @@ public final class BackingViewModel: BackingViewModelType, BackingViewModelInput
...
+  return titleString.simpleHtmlAttributedString(
+    base: [
+      NSFontAttributeName: UIFont.ksr_subhead(size: 13),
+      NSForegroundColorAttributeName: UIColor.black
+    ],
+    bold: [
+      NSFontAttributeName: UIFont.ksr_headline(size: 15),
+      NSForegroundColorAttributeName: UIColor.black
+    ])
+    ?? NSAttributedString(string: "")
```
stephencelis: `.init()` :)

Library/ViewModels/MessagesViewModel.swift
```diff
@@ -82,45 +91,73 @@ MessagesViewModelOutputs {
...
       )
-      .switchMap { backingOrThread -> SignalProducer<MessageThreadEnvelope, NoError> in
-
+      .switchMap { backingOrThread -> SignalProducer<Event<MessageThreadEnvelope?, ErrorEnvelope>, NoError> in
         switch backingOrThread {
         case let .left(backing):
           return AppEnvironment.current.apiService.fetchMessageThread(backing: backing)
-            .demoteErrors()
+            .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+            .materialize()
```
stephencelis: Yey!

Library/ViewModels/BackingViewModel.swift
justinswart: Ah good point about the type-safety!

KsApi/Service.swift
```diff
@@ -536,4 +536,29 @@ public struct Service: ServiceType {
         )
         .flatMap(decodeModels)
   }
+
+  private func request<M: Decodable>(_ route: Route)
+    -> SignalProducer<M?, ErrorEnvelope> where M == M.DecodedType {
```
theginbin: @justinswart yes, the response is actually entirely empty for a `MessageThreadEnvelope` that doesn't contain messages. The other option was to modify every field in the `MessageThreadEnvelope` to be Optional, but we decided to add the `request` overload instead to avoid modifying the model.

KsApi/MockService.swift
```diff
@@ -278,7 +278,7 @@ internal struct MockService: ServiceType {
 
     self.publishUpdateError = publishUpdateError
 
-    self.fetchMessageThreadResponse = fetchMessageThreadResponse ?? .template
+    self.fetchMessageThreadResponse = fetchMessageThreadResponse
```
theginbin: Ah yes, this case is a little different because the `fetchMessageThreadResponse` is actually of type `MessageThread` and not the entire `MessageThreadEnvelope?`. It certainly could be the entire thing, I'm not sure why it was originally set up this way. You have pointed out that the error is not being tested here so I can add that in. Right now, the "error" that's being tested is the empty messages, which is determined by the nil `fetchMessageThreadResponse`.

Library/ViewModels/BackingViewModel.swift
```diff
@@ -25,68 +26,53 @@ public protocol BackingViewModelOutputs {
...
-
-  /// Emits the accessibility label for backer reward description.
-  var backerRewardDescriptionAccessibilityLabel: Signal<String, NoError> { get }
-
   /// Emits the backer sequence to be displayed.
   var backerSequence: Signal<String, NoError> { get }
 
-  /// Emits the accessibility label for backer sequence.
-  var backerSequenceAccessibilityLabel: Signal<String, NoError> { get }
+  /// Emits with the project when should go to message creator screen.
```
theginbin: yikes hehe

Library/ViewModels/BackingViewModel.swift
theginbin: thanks ya'll!

Kickstarter-iOS/Views/Cells/BackingCell.swift
```diff
@@ -28,8 +28,8 @@ internal final class BackingCell: UITableViewCell, ValueCell {
                                      for: .touchUpInside)
   }
 
-  internal func configureWith(value: (Backing, Project)) {
-    self.viewModel.inputs.configureWith(backing: value.0, project: value.1)
+  internal func configureWith(value: (Backing, Project, Bool)) {
```
theginbin: oh yeah, could do.

Kickstarter-iOS/Views/Controllers/BackingViewController.swift
theginbin: yeah works for me. good to reuse the unwrapped `nav`

KsApi/Service.swift
justinswart: ah i see! ok cool 👍

Library/ViewModels/BackingViewModel.swift
theginbin: it can be if i swap the order of `Project` and `Backing` arguments.

Library/ViewModels/BackingViewModel.swift
theginbin: oh yeah sloppy me :)

PR #190, opened 2017-06-26
Generated new strings file and images for EmptyStateViewController test

bormind:
# What

New generated Strings.swift file and new Screenshots for `EmptyStateViewController` to reflect new translations to french 

# Why

Test was failing after new Strings file generation
# How

run `make strings`. Then generated new images by specifying `self.recordMode = true` in `EmptyStateViewControllerTest`


bormind:
Thank You Gina!

PR #193, opened 2017-06-30
Saving/Hearting in discovery card

cdolm92:
# What

We are allowing users to save projects from discovery. With this change the Backing team did an A/B/C/D test with different icons & copy around saving, and decided to move forward with hearts instead of stars.

# Why

Users can save projects they are interested in from the discovery page and will be reminded to pledge 48 hours before the project ends.

# How

On the `DiscoveryPostCardCell` we implemented  `cache(project: Project, shouldToggle: Bool) -> Bool ` this caches the project and returns a bool that determines if the save/heart button should be toggled. 

We also added a new Notification Center named `ksr_projectSaved` that notifies the `DiscoveryPostCardCell` when the user toggles the heart/save button in the `ProjectNavBarViewController`.

#See 👀

projsave

# TODO ⏰

- [x] Adjust tests in `DiscoveryPostcardViewModelTests`
- [ ] Bug fix 
- [x] Instrumentation update change event name from Unstarred/Starred to Unsaved/Saved Project


Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -273,6 +395,33 @@ public final class DiscoveryPostcardViewModel: DiscoveryPostcardViewModelType,
...
+private func cached(project: Project) -> Project {
+  if let projectCache = AppEnvironment.current.cache[KSCache.ksr_projectSaved] as? [Int: Bool] {
+    let isSaved = projectCache[project.id] ?? project.personalization.isStarred
+    return project |> Project.lens.personalization.isStarred .~ isSaved
+  } else {
+    return project
+  }
+}
+
+private func cache(project: Project, shouldToggle: Bool) -> Bool {
```
christopherwright: I think there's some value in adding a little comment explaining what's happening in this function

Library/Koala/Koala.swift
```diff
@@ -494,6 +507,12 @@ public final class Koala {
       ])
   }
 
+  public func trackSaveProject(context: SaveContext) {
+    self.track(event: "Star", properties: [ // check this
```
cdolm92: Thanks! this is not being used we are now tracking with `trackProjectSave(_ project: Project, context: SaveContext)`.

stephencelis:
@cdolm92 Looks like some of the icons were imported into assets as `png`s. Can we make sure they're `pdf`s instead?

cdolm92:
@stephencelis they have to be pngs because we had sizing issues when we tried importing as pdfs. @theginbin

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -133,10 +150,20 @@ public protocol DiscoveryPostcardViewModelType {
...
-    let project = self.projectProperty.signal.skipNil()
+  // swiftlint:disable function_body_length
+  public init() {
+    let configuredProject = self.projectProperty.signal.skipNil()
+      .map(cached(project:))
+
+    let currentUser = Signal.merge([
+      self.userSessionStartedProperty.signal,
+      self.userSessionEndedProperty.signal,
+      configuredProject.ignoreValues()
```
cdolm92: When we do not have a current user when the project emits we prompt the user to login when they tap on the heart/save button. Otherwise, if we have a user once projects emit we do not have to show the login prompt when they tap the heart/save button. @theginbin

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -303,7 +307,31 @@ extension DiscoveryPageViewController: DiscoveryPostcardCellDelegate {
...
-}
+
+  internal func discoveryPostcardCellProjectSaveAlert() {
+    let alertController = UIAlertController(
+      title: Strings.Project_saved(),
+      message: Strings.Well_remind_you_forty_eight_hours_before_this_project_ends(),
+      preferredStyle: .alert)
+    alertController.addAction(
+      UIAlertAction(
+        title: Strings.discovery_favorite_categories_alert_buttons_got_it(),
```
theginbin: good call. the `Got_it()` string has been added already!

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
```diff
@@ -123,13 +125,14 @@ public final class ProjectNavBarViewController: UIViewController {
       .observeValues { [weak self] in self?.categoryButton.setTitleColor($0, for: .normal) }
     self.categoryButton.rac.title = self.viewModel.outputs.categoryButtonText
     self.projectNameLabel.rac.text = self.viewModel.outputs.projectName
-    self.starButton.rac.accessibilityHint = self.viewModel.outputs.starButtonAccessibilityHint
-    self.starButton.rac.selected = self.viewModel.outputs.starButtonSelected
+    self.saveButton.rac.accessibilityHint = self.viewModel.outputs.starButtonAccessibilityHint // fix this
```
theginbin: I don't think we need a hint here? also you wrote "fix this" for some reason :)

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
```diff
@@ -171,14 +174,33 @@ public final class ProjectNavBarViewController: UIViewController {
...
-  fileprivate func showProjectStarredPrompt(message: String) {
-    let alert = UIAlertController.alert(nil, message: message, handler: nil)
+  fileprivate func showProjectStarredPrompt() {
+    let alert = UIAlertController(
+      title: Strings.Project_saved(),
+      message: Strings.Well_remind_you_forty_eight_hours_before_this_project_ends(),
+      preferredStyle: .alert)
+    alert.addAction(
+      UIAlertAction(
+        title: Strings.discovery_favorite_categories_alert_buttons_got_it(),
```
theginbin: yup, you can use the same "Got_it()" string.

KsApi/MockService.swift
```diff
@@ -361,6 +363,7 @@ internal struct MockService: ServiceType {
     self.submitApplePayResponse = submitApplePayResponse
 
     self.toggleStarResponse = toggleStarResponse
+    self.toggleStarError = toggleStarError
```
theginbin: It looks like you haven't tied this error to the function where it is being called. If you look at the `toggleStar(_ project:)` method, you can see that if `toggleStarResponse` is nil, it returns the same error every time: `.init(error: .couldNotParseJSON)`. Instead, you'll want to update the method to use your `toggleStarError`.

Library/Strings.swift
```diff
@@ -4136,17 +4136,17 @@ catch your eye?"
...
    - **de**: "%{friend_name} nicht mehr folgen"
    - **fr**: "Ne plus suivre %{friend_name}"
-   - **en**: "Unfollow %{friend_name}"
+   - **en**: "Unfollow %{friend_name}."
   */
   public static func Unfollow_friend_name(friend_name: String) -> String {
     return localizedString(
       key: "Unfollow_friend_name",
-      defaultValue: "Unfollow %{friend_name}",
+      defaultValue: "Unfollow %{friend_name}.",
```
theginbin: Just another reminder to merge in master so this file doesn't change. This is old.

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -85,6 +90,12 @@ public protocol DiscoveryPostcardViewModelOutputs {
   /// Emits when we should notify the delegate that the share button was tapped.
   var notifyDelegateShareButtonTapped: Signal<ShareContext, NoError> { get }
 
+  /// Emits when we should notify the delegate that the heart button was tapped.
+  var notifyDelegateShowSaveAlert: Signal<Void, NoError> { get }
+
+  // var notifyDelegateShowLoginTout<LoginIntent>
```
theginbin: Woopsie :)

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -123,7 +140,7 @@ public protocol DiscoveryPostcardViewModelOutputs {
 
   /// Emits a boolean that determines if the social view should be hidden.
   var socialStackViewHidden: Signal<Bool, NoError> { get }
-}
+ }
```
theginbin: I think this space was a typo?

Library/ViewModels/DiscoveryPostcardViewModel.swift
theginbin: Yeah @christopherwright, as Christella is saying, the `currentUser` is referenced later when a logged out user tries to save a project. The reason to put it up here is just for the convenience of getting the current user in one place and filtering on it later when the button is tapped.

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -155,93 +182,184 @@ public final class DiscoveryPostcardViewModel: DiscoveryPostcardViewModelType,
...
+      .takeWhen(self.saveButtonTappedProperty.signal)
+      .filter(isNotNil)
+      .ignoreValues()
+
+    let userLoginAfterTappingSaveButton = Signal.combineLatest(
+      self.userSessionStartedProperty.signal,
+      loggedOutUserTappedSaveButton
+      )
+      .ignoreValues()
+      .take(first: 1)
```
theginbin: Curious why there is a `.take(first: 1)` here? I believe if the user triggered this signal, then logged out, then logged in, and then tried to save a project, this would never emit again. Weird use case, I know, but still...

Library/ViewModels/ProjectNavBarViewModel.swift
```diff
@@ -37,20 +37,26 @@ public protocol ProjectNavBarViewModelOutputs {
   /// Emits when the login tout should be shown to the user.
   var goToLoginTout: Signal<(), NoError> { get }
 
+  /// Emits a project
+  var project: Signal<Project, NoError> { get }
```
theginbin: I think this could have a better output name since it's not clear what the project is emitting for. How about `postNotificationWithProject`?

Library/ViewModels/ProjectNavBarViewModel.swift
```diff
@@ -37,20 +37,26 @@ public protocol ProjectNavBarViewModelOutputs {
...
+  /// Emits a boolean that determines if the save button is selected.
+  var saveButtonSelected: Signal<Bool, NoError> { get }
+
+  /// Emits when the project has been successfully saved and a prompt should be shown to the user.
+  var showProjectSavedPrompt: Signal<Void, NoError> { get }
 
-  /// Emits a boolean that determines if the star button is selected.
-  var starButtonSelected: Signal<Bool, NoError> { get }
+  /// Emits the accessibility hint for the star button.
+  var starButtonAccessibilityHint: Signal<String, NoError> { get } // check this after new strings
```
theginbin: Not sure we need a hint; the label should be enough.

Library/ViewModels/ProjectNavBarViewModel.swift
```diff
@@ -37,20 +37,26 @@ public protocol ProjectNavBarViewModelOutputs {
...
 
   var titleHiddenAndAnimate: Signal<(hidden: Bool, animate: Bool), NoError> { get }
-}
+  }
```
theginbin: this this was a typo?

Library/ViewModels/ProjectNavBarViewModel.swift
```diff
@@ -129,18 +144,27 @@ ProjectNavBarViewModelInputs, ProjectNavBarViewModelOutputs {
...
+      })
+      .ignoreValues()
 
-    self.starButtonSelected = project
+    self.saveButtonSelected = project
       .map { $0.personalization.isStarred == true }
       .skipRepeats()
 
-    self.starButtonAccessibilityHint = self.starButtonSelected
+    self.starButtonAccessibilityHint = self.saveButtonSelected
```
theginbin: I think the hint is redundant since the label is "Save project". I think we can remove the hint here.

Kickstarter-iOS/Assets.xcassets/icons/heart-small-icon.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "icon-heart-filled.pdf",
```
cdolm92: I need this icon for the `BackerDashboardProjectCell` and `BackerDashboardEmptyStateCell`.

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
```diff
@@ -108,9 +108,11 @@ public final class ProjectNavBarViewController: UIViewController {
 
     _ = self.shareButton
       |> shareButtonStyle
+      |> UIButton.lens.accessibilityLabel %~ { _ in Strings.dashboard_accessibility_label_share_project() }
 
-    _ = self.starButton
+    _ = self.saveButton
       |> saveButtonStyle
+      |> UIButton.lens.accessibilityLabel %~ { _ in Strings.Save_this_project() }
```
theginbin: So I think that for both the nav bar and the postcard cell, we want to have the `accessibilityLabel` read:
"Toggle saving this project" and then add an `accessibilityValue` that reads "Saved" or "Unsaved". 

Right now this reads "Save this project. Saves this project." which I know is old code, but we may as well fix it now.

KsApi/MockService.swift
```diff
@@ -892,8 +895,15 @@ internal struct MockService: ServiceType {
...
-    return .init(value: toggleStarResponse)
+//    guard let toggleStarResponse = toggleStarResponse else { return .init(error: .couldNotParseJSON) }
+//    return .init(value: toggleStarResponse)
+
+      if let error = self.toggleStarError {
+        return SignalProducer(error: error)
+      } else if let toggleStar = self.toggleStarResponse {
+        return SignalProducer(value: toggleStar)
+      }
+      return .empty
```
theginbin: you can return `SignalProducer(value: .template)` instead of `.empty`.

KsApi/MockService.swift
```diff
@@ -892,8 +895,15 @@ internal struct MockService: ServiceType {
   }
 
   internal func toggleStar(_ project: Project) -> SignalProducer<StarEnvelope, ErrorEnvelope> {
-    guard let toggleStarResponse = toggleStarResponse else { return .init(error: .couldNotParseJSON) }
-    return .init(value: toggleStarResponse)
+//    guard let toggleStarResponse = toggleStarResponse else { return .init(error: .couldNotParseJSON) }
+//    return .init(value: toggleStarResponse)
```
theginbin: you can remove these comments.

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
stephencelis: @cdolm92 Wanna make this change?

stephencelis:
Somehow my review was pending :/

They're all minor things, though!

Library/ViewModels/ProjectNavBarViewModel.swift
theginbin: this is still a thing, when you get a chance :)

PR #194, opened 2017-07-05
Bind project to no reward cell

stephencelis:
TIL a bit more about what our `ValueCell`s do behind the scenes! Namely: their associated data type is bound directly to the data source's index paths. We recently removed the project type from our "No Reward" cell configuration (we weren't using it in the function directly), but this broke the connection for no-reward pledging! Here's a quick lil fix and quick lil test.


stephencelis:
[No Rewards pledging fix](https://trello.com/c/OhpXDxEw/633-no-rewards-pledging-fix)

PR #195, opened 2017-07-06
revise push handling

theginbin:
# What 

This is a PR on top of @bormind's PR for push notification opt-in and out. I thought this would be the easiest way to explain a refactor. I will make comments inline to explain the changes, which are mainly to clean up the flow of input/outputs based on the [Apple dev guide](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/index.html#//apple_ref/doc/uid/TP40008194-CH3-SW1) and avoid the use of the protocol/enum wrapper for `UNAuthorizationStatus`.  Let me know what you think @bormind @justinswart @stephencelis!

**The revised flow:**
_For iOS 10:_

On app launch, we should call `requestAuthorization(options: UNAuthorizationOptions, completionHandler:)` for new users or users with undetermined notification authorization status. When the completion handler is called, we should check if the authorization has been granted and only then register for remote notifications.

On app launch, we should also be calling `UNUserNotificationCenter.current().getNotificationSettings` for returning users who have set or updated notification settings. In addition to tracking the setting, we should call `requestAuthorization(options: UNAuthorizationOptions, completionHandler:)` again if the setting is `.notDetermined`.

### todo
Should add tests for iOS9, for authorization not granted, and for denied and authorized status.

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -135,10 +135,11 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         self.viewModel.outputs.authorizeForRemoteNotifications
           .observeForUI()
           .observeValues { [weak self] in
-            UNUserNotificationCenter.current().requestAuthorization(options: [.alert, .sound, .badge]) { _ in
-              self?.viewModel.inputs.notificationAuthorizationCompleted()
+            UNUserNotificationCenter.current().requestAuthorization(options: [.alert]) {
+              (isGranted, _) in
+              self?.viewModel.inputs.notificationAuthorizationCompleted(isGranted: isGranted)
+              }
             }
-          }
       }
 
     self.viewModel.outputs.unregisterForRemoteNotifications
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -22,34 +22,10 @@ public func == (lhs: HockeyConfigData, rhs: HockeyConfigData) -> Bool {
     && lhs.userName == rhs.userName
 }
 
-public protocol NotificationAuthorizationStatusType {
-  var isAuthorized: Bool { get }
-  var isDenied: Bool { get }
-  var isNotDetermined: Bool { get }
-}
-
-@available(iOS 10.0, *)
-extension UNAuthorizationStatus: NotificationAuthorizationStatusType {
-  public var isAuthorized: Bool {
-    switch self {
-    case .authorized: return true
-    case .denied, .notDetermined: return false
-    }
-  }
-
-  public var isDenied: Bool {
-    switch self {
-    case .denied: return true
-    case .authorized, .notDetermined: return false
-    }
-  }
-
-  public var isNotDetermined: Bool {
-    switch self {
-    case .notDetermined: return true
-    case .denied, .authorized: return false
-    }
-  }
+public enum NotificationAuthorizationStatusType {
+  case authorized
+  case denied
+  case notDetermined
 }
 
 public protocol AppDelegateViewModelInputs {
@@ -90,10 +66,11 @@ public protocol AppDelegateViewModelInputs {
   func foundRedirectUrl(_ url: URL)
 
   /// Call when notification authorization is completed
-  func notificationAuthorizationCompleted()
+  func notificationAuthorizationCompleted(isGranted: Bool)
 
   /// Call when notification authorization status received
-  func notificationAuthorizationStatusReceived(_ authorizationStatus: NotificationAuthorizationStatusType)
+  @available(iOS 10.0, *)
+  func notificationAuthorizationStatusReceived(_ authorizationStatus: UNAuthorizationStatus)
 
   /// Call when the user taps "OK" from the notification alert.
   func openRemoteNotificationTappedOk()
@@ -265,14 +242,10 @@ AppDelegateViewModelOutputs {
       .filter { AppEnvironment.current.currentUser != nil }
 
     if #available(iOS 10.0, *) {
-      self.getNotificationAuthorizationStatus =
-        Signal.merge(
-          applicationLaunchedWithUser,
-          self.notificationAuthorizationCompletedProperty.signal)
+      self.getNotificationAuthorizationStatus = applicationLaunchedWithUser
 
-      self.registerForRemoteNotifications = self.notificationAuthorizationStatusProperty.signal
-        .skipNil()
-        .filter { $0.isAuthorized }
+      self.registerForRemoteNotifications = self.notificationAuthorizationCompletedProperty.signal
+        .filter(isTrue)
         .ignoreValues()
 
     } else {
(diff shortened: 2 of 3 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -135,10 +135,11 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         self.viewModel.outputs.authorizeForRemoteNotifications
           .observeForUI()
           .observeValues { [weak self] in
-            UNUserNotificationCenter.current().requestAuthorization(options: [.alert, .sound, .badge]) { _ in
-              self?.viewModel.inputs.notificationAuthorizationCompleted()
+            UNUserNotificationCenter.current().requestAuthorization(options: [.alert]) {
+              (isGranted, _) in
+              self?.viewModel.inputs.notificationAuthorizationCompleted(isGranted: isGranted)
```
theginbin: We want to pass in whether authorization was granted or not so that we don't register a user remotely if they don't want to be. 

Note, we don't use `.sound` or `.badge` for iOS 9, so I assume these settings should be the same for iOS 10.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -22,34 +22,10 @@ public func == (lhs: HockeyConfigData, rhs: HockeyConfigData) -> Bool {
...
-    switch self {
-    case .notDetermined: return true
-    case .denied, .authorized: return false
-    }
-  }
+public enum NotificationAuthorizationStatusType {
+  case authorized
+  case denied
+  case notDetermined
 }
```
theginbin: We can just model this enum after the `UNAuthorizationStatus` enum and convert it with a function later. See below.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -265,14 +242,10 @@ AppDelegateViewModelOutputs {
       .filter { AppEnvironment.current.currentUser != nil }
 
     if #available(iOS 10.0, *) {
-      self.getNotificationAuthorizationStatus =
-        Signal.merge(
-          applicationLaunchedWithUser,
-          self.notificationAuthorizationCompletedProperty.signal)
+      self.getNotificationAuthorizationStatus = applicationLaunchedWithUser
```
theginbin: I believe we want to get the notification settings only on app launch (etc), not after authorization has been requested. I didn't see this pattern documented anywhere, but open for debate.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -265,14 +242,10 @@ AppDelegateViewModelOutputs {
...
 
-      self.registerForRemoteNotifications = self.notificationAuthorizationStatusProperty.signal
-        .skipNil()
-        .filter { $0.isAuthorized }
+      self.registerForRemoteNotifications = self.notificationAuthorizationCompletedProperty.signal
+        .filter(isTrue)
         .ignoreValues()
```
theginbin: We want to register if the result of `requestAuthorization()` is true, not check the status of `getNotificationSettings()`. A user with an `.authorized` status has *already* registered for remote notifications.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -282,10 +255,13 @@ AppDelegateViewModelOutputs {
...
-      .skipNil()
-      .filter { $0.isNotDetermined }
-      .ignoreValues()
+    self.authorizeForRemoteNotifications = Signal.merge(
+      applicationLaunchedWithUser,
+      self.notificationAuthorizationStatusProperty.signal
+        .skipNil()
+        .filter { $0 == .notDetermined }
+        .ignoreValues()
+    )
```
theginbin: As noted in the main comment, we want to request authorization both on app launch and if the result of `getNotificationSettings` is `.notDetermined`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -739,17 +717,16 @@ AppDelegateViewModelOutputs {
...
 
   fileprivate let notificationAuthorizationStatusProperty =
     MutableProperty<NotificationAuthorizationStatusType?>(nil)
-  public func notificationAuthorizationStatusReceived(
-    _ notificationSetting: NotificationAuthorizationStatusType
-  ) {
-    return self.notificationAuthorizationStatusProperty.value = notificationSetting
+  @available(iOS 10.0, *)
+  public func notificationAuthorizationStatusReceived(_ authorizationStatus: UNAuthorizationStatus) {
+    return self.notificationAuthorizationStatusProperty.value = authStatusType(for: authorizationStatus)
```
theginbin: Here is my solution to avoid the protocol extension: convert the `UNAuthorizationStatus` right here with a lil function. How do you all feel about this?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
theginbin: thinking about this some more... It seems like `registerForRemoteNotifications` should also be called as you have it @bormind  (for the use case that the user originally denies notifications, but then allows them in their device Settings). It was unclear to me, but it seems like a device can be registered, but they won't see the alert again after the first time.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: Yip I think this approach is fine, perhaps we should drop `Type` in the name as this is the concrete type and not a protocol anymore?

Considered also aligning the two enums exactly and then in the function just returning this type instantiated with the other's `rawValue` but that produces an optional and also having the exhaustive switch is better 👍

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: Yeah, prompt is only shown once and then again if the app is deleted and reinstalled. From looking at how the code was written previously it appears that we try to cover all of the scenarios:

- App launched with options
- User session started
- Will enter foreground 

In the case of entering foreground a scenario could be: user denies permission, backgrounds the app, navigates to settings and grants permission and then when the app moves back to foreground we want to try to register again. I think in this particular scenario the permission would be `.granted` and not `.notDetermined`?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: I like this, definitely prefer using enum to the custom protocol.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: May be this is not the best  name for this signal but our `applicationLaunchedWithUser` is fired not only when application launches but also when application enters foreground. One thing I was trying to avoid in my implementation is to call `UNUserNotificationCenter.current().requestAuthorization` every time application about to enter foreground. Not sure if this is important because ios just ignores redundant calls, but thought that this call belongs to the actual  application startup (or user registration).

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: I don't think anything wrong with using the `authStatusType` function, but I have general question:
If I would be implementing this I would probably think "'NotificationAuthorizationStatus' is the replacement type for the 'UNAuthorizationStatus' and for that I should have a factory static method:"
```
extension NotificationAuthorizationStatus {
    static func createFrom(uiAuthStatus: UNAuthorizationStatus) -> NotificationAuthorizationStatus
}
```
I there specific opinion/approach in our code where we would prefer one way to another?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: @justinswart you are correct - .notDetermined only happen before the Notification prompt dialog shown to the user.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
theginbin: Boris and I discussed this on a call and we're just gonna keep the free-floating function, but this implementation would be fine too.

PR #202, opened 2017-07-12
Xcode9

stephencelis:
This is a backwards-compatible PR that'll allow us to build the project in Xcode 9 and target iOS 11.

### DIFF
--- a/KsApi/Service.swift
+++ b/KsApi/Service.swift
@@ -460,7 +460,7 @@ public struct Service: ServiceType {
     return request(.updateUserSelf(user))
   }
 
-  private func decodeModel<M: Decodable>(_ json: Any) ->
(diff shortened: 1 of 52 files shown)


KsApi/Service.swift
```diff
@@ -460,7 +460,7 @@ public struct Service: ServiceType {
     return request(.updateUserSelf(user))
   }
 
-  private func decodeModel<M: Decodable>(_ json: Any) ->
+  private func decodeModel<M: Argo.Decodable>(_ json: Any) ->
```
stephencelis: Most of this diff was a quick find/replace to address an Argo conflict with Swift's new Codable stuff! (See https://github.com/thoughtbot/Argo/issues/466)

PR #203, opened 2017-07-12
Fix reuseIdentifier for empty message cells

theginbin:
# What

Was testing something for @tiegz  and found a bug where we didn't have the correct reuse identifier for empty Messages cells! Woops. This fixes that as well as points us to `master` on `ReactiveExtensions` with @stephencelis's  latest update.

screen shot 2017-07-12 at 5 37 03 pm

### DIFF
--- a/Kickstarter-iOS/DataSources/MessageThreadsDataSource.swift
+++ b/Kickstarter-iOS/DataSources/MessageThreadsDataSource.swift
@@ -7,6 +7,8 @@ internal final class MessageThreadsDataSource: ValueCellDataSource {
     case messageThreads
   }
 
+  internal let emptyStateCellIdentifier = String(describing: MessageThreadEmptyStateCell.self)
+
   internal func load(messageThreads: [MessageThread]) {
     self.set(values: messageThreads,
              cellClass: MessageThreadCell.self,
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/DataSources/MessageThreadsDataSource.swift
```diff
@@ -14,7 +14,7 @@ internal final class MessageThreadsDataSource: ValueCellDataSource {
   }
 
   internal func emptyState(isVisible: Bool) {
-    self.set(cellIdentifiers: isVisible ? ["MessageThreadsEmptyState"] : [],
+    self.set(cellIdentifiers: isVisible ? ["MessageThreadEmptyStateCell"] : [],
```
stephencelis: Possible to write a test? Also would be nice to throw this in an enum or constant.

theginbin: yeah i could write a test. i hate loose strings and would love to have an enum for all of our cell names, but it seems like a separate task?

Kickstarter-iOS/DataSources/MessageThreadsDataSource.swift
```diff
@@ -7,14 +7,16 @@ internal final class MessageThreadsDataSource: ValueCellDataSource {
     case messageThreads
   }
 
+  internal let emptyStateCellIdentifier = String(describing: MessageThreadEmptyStateCell.self)
```
theginbin: hey @stephencelis in lieu of a more standardized way of accessing these loose cell names across the codebase, I opted for this constant. Let me know if this works.

stephencelis: Sounds good! Wanna make it a static?

theginbin: will do!

PR #204, opened 2017-07-13
Use SKStoreReviewController

stephencelis:
10.3 introduced [SKStoreReviewController](https://developer.apple.com/documentation/storekit/skstorereviewcontroller), which will be the required way to prompt users to leave a review. On the plus side, it allows users to rate inline without leaving the app!

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ThanksViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ThanksViewController.swift
@@ -4,6 +4,7 @@ import Prelude
 import ReactiveSwift
 import ReactiveExtensions
 import Social
+import StoreKit
 import UIKit
 
 internal final class ThanksViewController: UIViewController, UICollectionViewDelegate {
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/Controllers/ThanksViewController.swift
```diff
@@ -185,18 +186,22 @@ internal final class ThanksViewController: UIViewController, UICollectionViewDel
...
-        }, remindHandler: { [weak self] _ in
-          self?.viewModel.inputs.rateRemindLaterButtonTapped()
-        }, noHandler: { [weak self] _ in
-          self?.viewModel.inputs.rateNoThanksButtonTapped()
-      }),
-      animated: true,
-      completion: nil
-    )
+    if #available(iOS 10.3, *) {
+      SKStoreReviewController.requestReview()
```
stephencelis: There's no input to feed back into our view model. There's no mechanism to know if the alert was displayed, if the user reviewed or canceled, etc.

theginbin: oh that's a bummer

justinswart:
@theginbin I think we should, currently it just takes us to the app store.

stephencelis:
> looks good! should we use this method in the SettingsViewController for sending people to the store?

Unfortunately we can't! Per Apple's guidelines this function shouldn't be called as a direct response to a user action. Apple determines whether or not to do anything at all when the function is called based on these requirements:

- Has the user disabled app review notifications in settings?
- Has the user seen an app review notification this version?

PR #205, opened 2017-07-13
Proj of the day label

bormind:
# What

Removing heart icon form the Project of the day label.

# Why
Quotation trello card:
"When we add hearts to project cards, we'll have two hearts on a single discovery card.
After discussing with Maggie, we'll just remove the heart from the PotD label for iOS."

# How

Set `iconImage` property while initializing `PostcardMetadataData` to nil for the `PostcardMetadataDataType.potd` 
After the change realized that we don't have tests for `DiscoveryPostCard` for featured and potD projects. Implemented `DiscoveryPageViewControllerTests.testView_Card_Project_TodaySpecial` test. Generated reference images for featured and potD post cards

# See 👀

[Trello](https://trello.com/c/5ZOF7MHi/589-ios-potd-treatment)


Kickstarter-iOS/Views/Controllers/DiscoveryPageViewControllerTests.swift
```diff
@@ -70,9 +70,43 @@ internal final class DiscoveryPageViewControllerTests: TestCase {
...
+
+          FBSnapshotVerifyView(parent.view,
+                               identifier: "\(labeledProj.0)_lang_\(language)_device_\(device)")
+        }
+    }
+  }
+
   func testView_Card_Project() {
     let projectTemplate = anomalisaNoPhoto
-      |> Project.lens.dates.deadline .~ (self.dateType.init().timeIntervalSince1970 + 60 * 60 * 24 * 6)
```
theginbin: is this just deleted because it is unused?

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewControllerTests.swift
```diff
@@ -70,9 +70,43 @@ internal final class DiscoveryPageViewControllerTests: TestCase {
...
+          controller.change(filter: magicParams)
+
+          self.scheduler.run()
+
+          controller.tableView.layoutIfNeeded()
+          controller.tableView.reloadData()
+
+          FBSnapshotVerifyView(parent.view,
+                               identifier: "\(labeledProj.0)_lang_\(language)_device_\(device)")
+        }
```
theginbin: thanks for adding these tests!

bormind:
Removed non used "metadata-podt" icon from assets

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewControllerTests.swift
bormind: Yes, this is not used in the test and doesn't change the output screen

bormind:
Gina, great catch with shifted labels! 
I fixed it by hiding image ctrl if icon is not present as you suggested.

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -189,6 +189,7 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
     self.viewModel.outputs.metadataData
       .observeForUI()
       .observeValues { [weak self] data in
+        self?.metadataIconImageView.isHidden = data.iconImage == nil
```
justinswart: @theginbin any idea why so many this was originally done with so many side-effects in a single `observeValues`?. I think if we did this today we'd have had an output for each of those and been more explicit about the logic in the view model.

bormind: @justinswart, @theginbin,  I noticed it too. I don't see problem with this approach but it is different from our "regular way" of doing things. I would be very interested to have a short discussion on why having only one side-effect per signal is a better approach..

theginbin: @justinswart you're totally right. I have no idea why this wasn't broken out into several outputs. It may have been built very early or done out of time constraint at the time before shipping. @bormind having only one side effect per signal makes it easier to test and understand exactly what is being changed (for example, label text versus just knowing that an entire piece of data changed)

theginbin:
approving this! but if you feel up for it, it would be good to add a new output for that hidden imageView and a test with it. It's good practice anyway.

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
bormind: @theginbin this is why I wanted to have conversation about this. I actually disagree that It will be easier to test. I understand that change on any one of the data properties causes signal of entire data change. But because the data unit consist of *contextually dependent properties* I believe it will be simpler to describe in the unit test and easier to understand where actual problem is in case of bug happened.

bormind:
Thank you @theginbin! I believe that if we decide to create output signal for `imageView.isHidden` property then should create output signals for all other meta properties:
```
self?.metadataIconImageView.image = data.iconImage
self?.metadataLabel.text = data.labelText
self?.metadataIconImageView.tintColor = data.iconAndTextColor
self?.metadataLabel.textColor = data.iconAndTextColor
```

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
theginbin: I guess by "easer to test and understand", I mean that a unique output tells us exactly what we're expecting to update on the UI specifically. That's why we typically don't have one output to update the entire view because that doesn't tell us what we're doing with individual UI components.

theginbin:
@bormind so yeah I think you should go ahead and create outputs for all of them. By having separate outputs, we give the view model context about what is happening with the data. With a single output, we lose this context.

justinswart:
@bormind if you're feeling up to it it might be a useful exercise to break it up into separate outputs and write the corresponding tests! I realize it adds a fair bit of code to what was otherwise a pretty small change but it would be good for consistency with the rest of our code 👍

justinswart:
@bormind hey since your last commit separating out the `metadataIconHidden` into its own signal, did you not want to remove `self?.metadataIconImageView.isHidden = data.iconImage == nil` from the `observeValues` on `self.viewModel.outputs.metadataData`?

Also, I think the idea was that we would refactor this to be separate outputs for each metadata property.

justinswart:
@bormind hey just checking if you saw my last comment? Also I think we were still going to split this out into multiple outputs?

You said before on the PR:

> Thank you @theginbin! I believe that if we decide to create output signal for imageView.isHidden property then should create output signals for all other meta properties

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -76,8 +76,20 @@ public protocol DiscoveryPostcardViewModelOutputs {
   /// Emits a boolean to determine whether or not to display funding progress container view.
   var fundingProgressContainerViewHidden: Signal<Bool, NoError> { get }
 
-  /// Emits the disparate data to be displayed on the metadata view label.
-  var metadataData: Signal<PostcardMetadataData, NoError> { get }
+  /// Emits matadata label text
```
justinswart: *metadata

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -76,8 +76,20 @@ public protocol DiscoveryPostcardViewModelOutputs {
...
+  var metadataLabelText: Signal<String, NoError> { get }
+
+  /// Emits metadata icon name
```
justinswart: I think this should say:

`/// Emits metadata icon image`

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -76,8 +76,20 @@ public protocol DiscoveryPostcardViewModelOutputs {
...
+  var metadataIcon: Signal<UIImage?, NoError> { get }
+
+  /// Emits metadata icon and text color
+  var metadataIconAndTextColor: Signal<UIColor, NoError> { get }
```
justinswart: Just checking, this only _seems_ to be used for the `textColor`, is `metadataIconAndTextColor` the right name?

bormind:
Thank you @justinswart!

bormind:
Thank you @justinswart, @theginbin !!

PR #206, opened 2017-07-16
Use MockDate when constructing HTTPCookie

justinswart:
# What

Updated code in `ProjectPamphletViewModel` that constructs an `HTTPCookie` to use our `DateProtocol` type on `AppEnvironment`.

# Why

`ProjectPamphletViewModelTests.testTracksRefTag()` keeps randomly failing on Circle and I think it has to do with equality of `HTTPCookie`s being set in the `Set<HTTPCookie>` on `MockCookieStorage`. Usually rebuilding on Circle fixes it but then it subsequently fails again.

# How

I noticed that the code which we were using to create `HTTPCookie`s was not using our `DateProtocol` type on our `AppEnvironment` and therefore in tests it would not be using `MockDate` so results in tests could be unpredictable. Perhaps the two `HTTPCookie`s that we are constructing in this test might have a different `created` date property and are seen as unique when they're added to the `Set<HTTPCookie>`? However, `HTTPCookie` is a class so equality should be on memory address, shouldn't it? 🤔 

Still, I'd like to merge this change to see if it resolves this test occasionally failing on Circle.

# See 👀

screen shot 2017-07-16 at 12 45 22 pm

### DIFF
--- a/Library/ViewModels/ProjectPamphletViewModel.swift
+++ b/Library/ViewModels/ProjectPamphletViewModel.swift
@@ -177,7 +177,7 @@ private func cleanUp(refTagString: String) -> String {
 // Constructs a cookie from a ref tag and project.
 private func cookieFrom(refTag: RefTag, project: Project) -> HTTPCookie? {
 
-  let timestamp = Int(Date().timeIntervalSince1970)
+  let timestamp = Int(AppEnvironment.current.dateType.init().timeIntervalSince1970)
 
   var properties: [HTTPCookiePropertyKey:Any] = [:]
   properties[.name]    = cookieName(project)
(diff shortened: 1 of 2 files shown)


justinswart:
@stephencelis well the test that this hopefully fixes should be the test that catches it, I think? It's essentially testing that even though two view models were configured with the same project, and the tracking was triggered twice, only 1 referral cookie is set for that ref tag.

justinswart:
@stephencelis thought about it some more and it was indeed testable! I've added two tests and proven that separate schedulers would cause the cookies to be created with different `timestamp` values and would therefore not be equal and would both be added to the `Set<HTTPCookie>`.

I think this is what is occasionally happening on Circle and causing the test to fail.

PR #207, opened 2017-07-17
Background color update

cdolm92:
# What

Changed background color of Discovery, Activity, Search, Dashboard, Profile, Project page, Reward screens to white.

# See

See new screenshots for context.


Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -62,7 +62,7 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
 
     _ = self
       |> baseTableViewCellStyle()
-      |> DiscoveryPostcardCell.lens.backgroundColor .~ .clear
+      //|> DiscoveryPostcardCell.lens.backgroundColor .~ .clear
```
theginbin: think you can delete this commented line.

Library/Styles/BaseStyles.swift
```diff
@@ -81,6 +81,12 @@ public func dropShadowStyle <V: UIViewProtocol> (radius: CGFloat = 2.0,
       <> V.lens.layer.shadowOffset .~ offset
 }
 
+public func rewardDropShadowStyle <V: UIViewProtocol> () -> ((V) -> V) {
+  return dropShadowStyle(radius: 5.0, offset: .init(width: 0, height: 2.0))
+    <> V.lens.layer.shadowOpacity .~ 0.17
+    <> V.lens.layer.shadowColor .~ UIColor.ksr_dropShadow.cgColor
```
theginbin: This looks exactly like `dropShadowStyleMedium` except for the shadow color. Is there a reason the black color wasn't working? According to the style guide, there should only be the 3 types of shadow, all with black at 0.17 opacity. I think we should be updating the base `dropShadowStyle` to black with 0.17 opacity. Once this is updated, you can drop the lines for `shadowOpacity` and `shadowColor` in the medium and large styles.  Let me know if you need more clarification. cc @maggieKSR

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -204,7 +204,9 @@ RewardCellViewModelOutputs {
 
     let tappable = Signal.zip(project, reward, youreABacker)
       .map { project, reward, youreABacker in
-        (project.state == .live && reward.remaining != 0) || youreABacker
+        (project.state == .live && reward.remaining != 0)
+        || (project.state != .live && reward.remaining != 0)
+        || youreABacker
```
theginbin: curious why was this line added for tapping on a non-live project reward?

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -215,7 +217,7 @@ RewardCellViewModelOutputs {
 
     self.cardViewBackgroundColor = Signal.combineLatest(allGoneAndNotABacker, self.boundStylesProperty.signal)
       .map(first)
-      .map { $0 ? .ksr_grey_100 : .white }
+      .map { _ in .white }
```
theginbin: you can kill both `maps` and just use `mapConst(.white)`. Or, you may not even need this output now?

Library/ViewModels/RewardCellViewModelTests.swift
```diff
@@ -104,7 +104,7 @@ final class RewardCellViewModelTests: TestCase {
 
     self.vm.inputs.configureWith(project: .template,
                                  rewardOrBacking: .left(.template |> Reward.lens.remaining .~ 0))
-    self.cardViewBackgroundColor.assertValues([UIColor.white, UIColor.ksr_grey_100])
+    self.cardViewBackgroundColor.assertValues([UIColor.white, UIColor.white])
```
theginbin: yeah I think you can kill this output and test and just set the background color in `bindStyles` since it never changes.

theginbin:
Can you confirm that the background should remain light grey on the `RewardPledge` page @maggieKSR ?

theginbin:
@cdolm92 not sure if you're still working on this, but the `ProjectActivity` cells still need to be updated. Thanks.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -219,7 +219,15 @@ internal final class RewardCell: UITableViewCell, ValueCell {
...
-        self?.cardView.layer.shadowOpacity = hidden ? 0 : 1
+        self?.cardView.layer.shadowOpacity = hidden ? 0 : 0.17
+    }
+
+    self.viewModel.outputs.cardViewBorder
+      .observeForUI()
+      .observeValues { [weak self] _ in
+        self?.cardView.layer.borderColor = UIColor.ksr_grey_400.cgColor
+        self?.cardView.layer.borderWidth = 1.0
+        self?.cardView.layer.cornerRadius = 2.0
```
theginbin: So I think we need to consider what happens if a reward goes from all-gone to not all-gone (which could happen if a backer cancels a pledge - I _believe_). This output should really be toggling the visibility of the border because if the reward becomes available, it won't change the view. Also, I think this border could remain on reusing the cell...? The easiest resolution is to set the `borderWidth` to `0.0` when this output emits to hide the border (assuming you will rename this output to `cardViewBorderIsHidden` or `cardViewBorderIsVisible` ;).

Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
```diff
@@ -122,6 +122,7 @@
 "Funding_successful" = "Funding successful";
 "Get_notified_about_this_creators_live_streams" = "Get notified about this creator’s live streams";
 "Go_to_project_page" = "Go to project page";
+"Got_it" = "Got it.";
```
theginbin: is this Strings stuff for a different PR?

cdolm92: yes, for saving

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -213,9 +214,8 @@ RewardCellViewModelOutputs {
       )
       .map(first)
 
-    self.cardViewBackgroundColor = Signal.combineLatest(allGoneAndNotABacker, self.boundStylesProperty.signal)
-      .map(first)
-      .map { $0 ? .ksr_grey_100 : .white }
+    self.cardViewBorderIsVisible = Signal.zip(project, reward)
```
stephencelis: Any reason to prefer `zip` over `combineLatest` here? It's probably safe given how these kinds of outputs shouldn't update independently, but I think we generally have moved to use `combineLatest` by default.

PR #208, opened 2017-07-20
Fix tests in  Xcode 9

justinswart:
# What

Related to https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/66, this fixes `TestObserver`'s `assertValues()` in Swift 3.2 and Xcode 9.

This isn't quite working yet, Circle doesn't have the 10.3.1 simulator which we currently have specified in our `Makefile` to run tests. I've reached out to them to find out if we can install it for Xcode 9. Alternatively we'd need to re-record all of our screenshots for 11.0 but I don't think we should do that while it's still in beta.


justinswart:
For posterity: turns out the changes in ReactiveExtensions weren't necessary and instead we just need to assert on an array of `String` values with `assertValues()` when testing equality rather than an assertion on a single `String` value with `assertValue()`. This clears up the ambiguity.

PR #211, opened 2017-07-28
Search card update

cdolm92:
# What

New `SearchProjectCell` update.

# How

Reused `BackerDashboardProjectCell`. 

# See 👀
before:

image uploaded from ios

after:

simulator screen shot jul 28 2017 2 59 48 pm


Kickstarter-iOS/DataSources/SearchDataSource.swift
```diff
@@ -32,7 +32,9 @@ internal final class SearchDataSource: ValueCellDataSource {
 
     if !projects.isEmpty {
       projects.suffix(from: 1).forEach {
```
justinswart: I believe this could also be written as:

`projects.dropFirst().forEach { ... }`, the `Sequence` type gives us a `dropFirst` function which returns everything but the first element.

Kickstarter-iOS/Views/Cells/SearchProjectCell.swift
```diff
@@ -36,61 +33,44 @@ internal final class SearchProjectCell: UITableViewCell, ValueCell {
...
     self.projectImageView.rac.imageUrl = self.viewModel.outputs.projectImageUrlMed
-    self.projectLabel.rac.text = self.viewModel.outputs.projectNameLabelText
+    self.projectLabel.rac.attributedText = self.viewModel.outputs.projectNameLabelText
+
+    self.viewModel.outputs.progress
+      .observeForUI()
+      .observeValues { [weak element = progressBarView] progress in
+        let anchorX = progress == 0 ? 0 : 0.5 / progress
+        element?.layer.anchorPoint = CGPoint(x: CGFloat(max(anchorX, 0.5)), y: 0.5)
+        element?.transform = CGAffineTransform(scaleX: CGFloat(min(progress, 1.0)), y: 1.0)
```
justinswart: Wondering if these should be separate outputs so that we can write tests for them? 🤔

Also, capturing of `progressBarView` in this way is fine but capturing `[weak self]` directly calling `self?.progressBarView` in the closure also works and I think is more in line with the rest of our code I think. Either is fine.

Library/ViewModels/SearchProjectCellViewModel.swift
```diff
@@ -64,7 +69,18 @@ SearchProjectCellViewModelInputs, SearchProjectCellViewModelOutputs {
 
     self.projectImageUrlFull = project.map { URL(string: $0.photo.full) }
 
-    self.projectNameLabelText = project.map { $0.name }
+    self.projectNameLabelText = project.map(titleString(for:))
```
justinswart: Nice, these split out into `private` functions reads well 👍

Kickstarter-iOS/Views/Cells/MostPopularSearchProjectCell.swift
```diff
@@ -77,15 +86,28 @@ internal final class MostPopularSearchProjectCell: UITableViewCell, ValueCell {
...
+    self.metadataLabel.rac.text = self.viewModel.outputs.metadataText
+    self.percentFundedLabel.rac.attributedText = self.viewModel.outputs.percentFundedText
+    self.progressBarView.rac.backgroundColor = self.viewModel.outputs.progressBarColor
     self.projectImageView.rac.imageUrl = self.viewModel.outputs.projectImageUrlFull
+    self.projectNameLabel.rac.text = self.viewModel.outputs.projectName
+
+    self.viewModel.outputs.progress
+      .observeForUI()
+      .observeValues { [weak element = progressBarView] progress in
+        let anchorX = progress == 0 ? 0 : 0.5 / progress
```
justinswart: Same as other comment about separate outputs.

Kickstarter-iOS/Views/Storyboards/Search.storyboard
```diff
@@ -180,97 +205,122 @@
...
-                                    <outlet property="statsStackView" destination="a95-YE-evX" id="tTO-Cr-5is"/>
+                                    <outlet property="cardView" destination="hna-1r-IBt" id="cDG-bc-Si8"/>
+                                    <outlet property="metadataBackgroundView" destination="LC3-rr-wed" id="bcc-Yq-xXW"/>
+                                    <outlet property="metadataIconImageView" destination="d6r-M1-e1G" id="GJD-jz-SpE"/>
+                                    <outlet property="metadataLabel" destination="2Vq-Ze-Jnt" id="hR7-8W-OB5"/>
+                                    <outlet property="percentFundedLabel" destination="n2e-oy-cd3" id="bds-d3-G0B"/>
+                                    <outlet property="progressBarView" destination="nzr-VY-8lM" id="uVM-4n-HHt"/>
+                                    <outlet property="progressStaticView" destination="ZM0-2K-7i3" id="MVx-x6-cNj"/>
+                                    <outlet property="projectImageView" destination="jOZ-gh-bkN" id="Jy2-XA-mxq"/>
+                                    <outlet property="projectLabel" destination="Z2K-XW-AnR" id="OZd-P5-G64"/>
```
theginbin: It looks like the `SearchProjectCell` is still in the Storyboard. Can you remove it?

Library/ViewModels/SearchProjectCellViewModel.swift
```diff
@@ -10,11 +10,16 @@ public protocol SearchProjectCellViewModelInputs {
...
+  var fundingLargeLabelText: Signal<NSAttributedString, NoError> { get }
   var projectImageUrlFull: Signal<URL?, NoError> { get }
-  var projectNameLabelText: Signal<String, NoError> { get }
+  var projectImageUrlMed: Signal<URL?, NoError> { get }
+  var projectName: Signal<String, NoError> { get }
+  var projectNameLabelText: Signal<NSAttributedString, NoError> { get }
+  var progress: Signal<Float, NoError> { get }
+  var progressBarColor: Signal<UIColor, NoError> { get }
+  var percentFundedText: Signal<NSAttributedString, NoError> { get }
+  var metadataText: Signal<String, NoError> { get }
```
theginbin: You can also delete this view model as you're not using it anymore.

Kickstarter-iOS/Views/Cells/MostPopularSearchProjectCell.swift
justinswart: Ok cool :) actually even a tuple value as the output could also work,  no need for separate outputs.

Library/ViewModels/SearchProjectCellViewModel.swift
cdolm92: I'm using this in the `MostPopularSearchProjectCell`. I can rename this to `MostPopularSearchProjectCellViewModel`.

theginbin: Oh yes! I see. Yeah a rename would be good.

Library/ViewModels/MostPopularSearchProjectCellViewModel.swift
```diff
@@ -0,0 +1,173 @@
...
+  var fundingLargeLabelText: Signal<NSAttributedString, NoError> { get }
+  var projectImageUrlFull: Signal<URL?, NoError> { get }
+  var projectImageUrlMed: Signal<URL?, NoError> { get }
+  var projectName: Signal<String, NoError> { get }
+  var projectNameLabelText: Signal<NSAttributedString, NoError> { get }
+  var progress: Signal<Float, NoError> { get }
+  var progressBarColor: Signal<UIColor, NoError> { get }
+  var percentFundedText: Signal<NSAttributedString, NoError> { get }
+  var metadataText: Signal<String, NoError> { get }
+}
```
theginbin: would you mind adding comments for these outputs?

Library/ViewModels/MostPopularSearchProjectCellViewModel.swift
```diff
@@ -0,0 +1,173 @@
...
+    self.fundingSmallLabelText = project.map {
+      let string = Strings.percentage_funded(
+        percentage: "<b>\(Format.percentage($0.stats.percentFunded))</b>")
+      return string.simpleHtmlAttributedString(base: [
+        NSFontAttributeName: UIFont.ksr_subhead(size: 13.0),
+        NSForegroundColorAttributeName: UIColor.ksr_text_navy_500
+        ],
+                                               bold: [
+                                                NSFontAttributeName: UIFont.ksr_headline(size: 13.0),
+                                                NSForegroundColorAttributeName: UIColor.ksr_text_green_700
```
theginbin: can you fix the indentation here?

Kickstarter-iOS/Views/Cells/BackerDashboardProjectCell.swift
```diff
@@ -50,6 +50,11 @@ internal final class BackerDashboardProjectCell: UITableViewCell, ValueCell {
       |> UITableViewCell.lens.isAccessibilityElement .~ true
       |> UITableViewCell.lens.accessibilityHint %~ { _ in Strings.Opens_project() }
       |> UITableViewCell.lens.accessibilityTraits .~ UIAccessibilityTraitButton
+      |> UITableViewCell.lens.contentView.layoutMargins %~~ { _, cell in
+        cell.traitCollection.isRegularRegular
+          ? .init(topBottom: Styles.grid(2), leftRight: Styles.grid(20))
+          : .init(topBottom: Styles.gridHalf(3), leftRight: Styles.grid(2))
+    }
```
theginbin: Can you update the spacing to be `.init(topBottom: Styles.grid(1), leftRight: Styles.grid(2))` for iPhone size? I realize the spacing between cells is not consistent for a lot of pages, but I think it's based on the size of the cards and this was the spacing in the original design.

PR #215, opened 2017-08-02
Creator message notification

bormind:
# What

Clicking on new message push notification for creator navigates to the corresponding message thread

# How

Created new `case creatorMessages(Param, messageThreadId: Int)` navigation node
Created new input and output in `DashboardViewModel` to control the navigation to the message thread
Altered signal combination logic in the `DashboardViewModel` to accommodate the new navigation behaviour.

# See 👀
messagepush

[Trello](https://trello.com/c/eTFZ84SC/367-update-creator-message-notification-path)

# TODO ⏰

Was not able to test the functionality on real device. Do we have a facility for pushing notifications using development profile?

New navigation node was created but no corresponding navigation rout. Not sure if we need the rout for the push target.

No extra Koala tracking was implemented for this functionality. Will ask Katie about expected level of tracking to the message push


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -308,6 +316,11 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.rootTabBarController?.switchToMessageThread(messageThread)
   }
 
+  fileprivate func goToCreatorMessageThread(_ projectId: Param, _ messageThread: MessageThread) {
```
justinswart: This could also be `private`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -401,6 +404,20 @@ AppDelegateViewModelOutputs {
...
+    self.goToCreatorMessageThread = deepLink
+      .map { navigation -> (Param, Int)? in
+        guard case let .creatorMessages(projectId, messageThreadId) = navigation else { return nil }
+        return .some((projectId, messageThreadId: messageThreadId))
+      }
+      .skipNil()
+      .switchMap { projectId, messageThreadId in
+        AppEnvironment.current.apiService.fetchMessageThread(messageThreadId: messageThreadId)
+          .demoteErrors()
+          .map { (env: MessageThreadEnvelope) in
```
justinswart: Was this causing ambiguity as just as ` .map { env in ... }`? Just curious why the type is specified.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -811,7 +831,11 @@ private func navigation(fromPushEnvelope envelope: PushEnvelope) -> Navigation?
   }
 
   if let message = envelope.message {
-    return .messages(messageThreadId: message.messageThreadId)
+    if envelope.forCreator == true {
```
justinswart: This can just be `if envelope.forCreator { ... }`

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -158,7 +176,7 @@ internal final class DashboardViewController: UITableViewController {
     }
   }
 
-  fileprivate func goToMessages(_ project: Project) {
+  fileprivate func goToMessages(project: Project) {
```
justinswart: Can this be `private`?

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -214,15 +232,26 @@ internal final class DashboardViewController: UITableViewController {
   @objc fileprivate func shareButtonTapped() {
     self.shareViewModel.inputs.shareButtonTapped()
   }
+
+  fileprivate func goToMessageThread(project: Project, messageThread: MessageThread) {
```
justinswart: Can this be `private`?

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -427,4 +467,5 @@ internal final class DashboardViewModelTests: TestCase {
                      self.trackingClient.properties(forKey: "project_pid", as: Int.self))
     }
   }
+
```
justinswart: Extra new line crept in here.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,62 @@ public protocol DashboardViewModelType {
...
+      self.messageThreadNavigatedProperty.signal
+    )
+
+    self.goToMessageThread = projectsAndSelected
+      .map { $0.1 }
+      .switchMap { [messageThreadReceived = self.messageThreadReceived.producer] project in
+        messageThreadReceived
+          .skipNil()
+          .filter { $0.0 == .id(project.id) }
+          .map { (measgaThreadPair: (Param, MessageThread)) -> (Project, MessageThread) in
```
justinswart: Just a spell-fix here: `measgaThreadPair` -> `messageThreadPair`.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,62 @@ public protocol DashboardViewModelType {
 public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewModelOutputs,
   DashboardViewModelType {
 
-    public init() {
+  private let selectedProjectProperty: SignalProducer<Param?, NoError>
```
justinswart: It might be good to add comments to these detailing what they do, particularly because this is a slight deviation from our other viewModels.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -32,6 +32,12 @@ public protocol DashboardViewModelInputs {
   /// Call when the projects drawer has animated out.
   func dashboardProjectsDrawerDidAnimateOut()
 
+  /// Call to open message thread for specific project
+  func messageThreadNavigated(projectId: Param, messageThread: MessageThread)
+
+  /// Call to open project messages thread
+  func openMessageThreadRequested()
```
justinswart: `messagesCellTapped` might be a better name for this.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: I kept the type spec because it was this way in the original `self.goToMessageThread`. It did help me a bit while reading the code just because I am not familiar with the API but I agree - it is not necessary.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: I used this style because of the original code on lines 793,  872 uses it. I prefer the style that you sagest but wanted to keep consistent with the existing code.

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
bormind: @justinswart I am all for using private instead of fileprivate but we have fileprivate class members everywhere in this file. Is is ok if I will switch all of them to private?

Library/ViewModels/DashboardViewModel.swift
bormind: May be.. I try to treat view model inputs similarly to Command Pattern - You don't know who triggered the command you only know what the command should do. But may be my approach is incorrect? We can chat briefly about it.

justinswart: I think that the cell being tapped is the event that we're responding to here. Perhaps there is another alternative name, I just think that _open_ and _requested_ make this seem like an output rather than an input.

Also if there were another way of triggering this we would represent that as a separate input so being specific about exactly what the user did is fine I would say.

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
justinswart: I think that at some point we can refactor all of our `fileprivate`s to `private`s where it makes sense. If you'd like to change all of them here that's cool but I think ideally we just don't want to create _more_ `fileprivate`s.

bormind:
@justinswart Thank you for the review!!
I pushed some fixes. Didn't changed fileprivate to private yet - want to understand if we want to do it for all the members in the file we touch?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: Oh, from the diff it looks like it didn't have the type before?

screen shot 2017-08-03 at 3 46 47 pm

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: Sure no problem, I think either's fine.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
bormind: Sorry my bad - I probably created it while tying analyze the code. I will remove it

justinswart:
@bormind up to you regarding `fileprivate`, my feeling is not to add more of them and over time remove them wherever possible.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -97,9 +97,17 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         self?.goToLiveStream(project: $0, liveStreamEvent: $1, refTag: $2)
     }
 
+    self.viewModel.outputs.goToCreatorMessageThread
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.goToCreatorMessageThread($0.0, $0.1)
```
stephencelis: Can we use `$0, $1` instead of the additional tuple indexing?

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -97,9 +97,17 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+      }
+
     self.viewModel.outputs.goToMessageThread
       .observeForUI()
-      .observeValues { [weak self] in self?.goToMessageThread($0) }
+      .observeValues { [weak self] in
```
stephencelis: Any reason this isn't still a single line? (Compare with above the fold.)

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -304,10 +312,15 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     }
   }
 
-  fileprivate func goToMessageThread(_ messageThread: MessageThread) {
+  private func goToMessageThread(_ messageThread: MessageThread) {
```
stephencelis: 👍

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -410,8 +427,10 @@ AppDelegateViewModelOutputs {
       .switchMap {
         AppEnvironment.current.apiService.fetchMessageThread(messageThreadId: $0)
           .demoteErrors()
-          .map { env in env.messageThread }
-    }
+          .map { (env: MessageThreadEnvelope) in
```
stephencelis: Can clean this one up, too.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
stephencelis: I'd say change both if ya can! Probably a mistake. Generally we try to do:

``` swift
if booleanValue { /* … */ }
// or
if optionalBooleanValue == .some(true) { /* … */ }
```

Kickstarter-iOS/Views/Cells/DashboardActionCell.swift
```diff
@@ -74,8 +74,8 @@ internal final class DashboardActionCell: UITableViewCell, ValueCell {
 
     self.viewModel.outputs.goToMessages
       .observeForUI()
-      .observeValues { [weak self] project in
-        self?.delegate?.goToMessages(self, project: project)
+      .observeValues { [weak self] _ in
```
stephencelis: Should the output stop emitting a project here or is the output reused elsewhere?

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -114,6 +120,12 @@ internal final class DashboardViewController: UITableViewController {
         element?.updateData(data)
     }
 
+    self.viewModel.outputs.goToMessages
+      .observeForControllerAction()
+      .observeValues { [weak self] project in
+        self?.goToMessages(project: project)
+      }
```
stephencelis: lil indent

Library/ViewModels/DashboardViewModel.swift
stephencelis: Our commands are linked to direct inputs, so unless we're reusing this command (which is the exception more than the rule for us, I think), let's be as descriptive as possible per the actual action/input.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
 public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewModelOutputs,
   DashboardViewModelType {
 
-    public init() {
+  //We use producers ot this properties during view appearance and projects retreaval process
```
stephencelis: Would you please proof this comment? There are some typos and I'm not quite following.

We don't typically add extra producers/properties like this that aren't associated with direct inputs. Are there other ways we could write this logic? Thoughts @theginbin @justinswart?

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
...
+    .skipRepeats { lhs, rhs in lhs == rhs }
+
     let projectsAndSelected = projects
-      .switchMap { [switchToProject = self.switchToProjectProperty.producer] projects in
+      .switchMap { [switchToProject = self.selectProjectPropertyOrFirst.producer] projects in
         switchToProject
-          .map { param in
-            find(projectForParam: param, in: projects) ?? projects.first
+          .map { param -> Project? in
+            guard let param = param else {
```
stephencelis: Can use optional composition here over `guard let` and avoid the extra `projects.first` return:

``` swift
param.flatMap { find(projectForParam: $0, in: projects) } ?? projects.first
```

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
...
         AppEnvironment.current.apiService.fetchProjects(member: true)
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
           .demoteErrors()
           .map { $0.projects }
+          .prefix(value: [])
     }
 
+    self.selectedProjectProperty = SignalProducer.merge(
+      self.switchToProjectProperty.producer,
+      self.messageThreadNavigatedProperty.producer.skipNil().map { $0.0 }
```
stephencelis: Whenever you see `.map { $0.0 }` you should be able to call `.map(first)`.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
...
+    )
+
+    self.selectProjectPropertyOrFirst <~ SignalProducer.combineLatest(
+      self.selectedProjectProperty,
+      self.viewWillAppearAnimatedProperty.producer.ignoreValues()
+    )
+    .map { $0.0 }
```
stephencelis:
``` swift
.map(first)
```

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
...
           .map { (projects, $0) }
     }
 
+    self.messageThreadReceived <~ Signal.merge(
+      self.viewWillDisappearProperty.signal.mapConst(nil),
+      self.messageThreadNavigatedProperty.signal
+    )
+
+    self.goToMessageThread = projectsAndSelected
+      .map { $0.1 }
```
stephencelis:
``` swift
.map(second)
```

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,25 +102,65 @@ public protocol DashboardViewModelType {
...
+      .switchMap { [messageThreadReceived = self.messageThreadReceived.producer] project in
+        messageThreadReceived
+          .skipNil()
+          .filter { $0.0 == .id(project.id) }
+          .map { (messgeThreadPair: (Param, MessageThread)) -> (Project, MessageThread) in
```
stephencelis: - [ ] lil typo: `messageThreadPair`
- [ ] don't need the explicit `return`

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -136,11 +191,11 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
     }
 
     let drawerStateProjectsAndSelectedProject = Signal.merge(
-      projectsAndSelected.map { ($0, $1, false) },
-      projectsAndSelected.takeWhen(self.showHideProjectsDrawerProperty.signal).map { ($0, $1, true) }
-      )
+      projectsAndSelected.map { ($0.0, $0.1, false) },
```
stephencelis: Any reason why this is using tuple restructuring now? We generally try to avoid that as it's less readable.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -136,11 +191,11 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
...
+      projectsAndSelected
+        .map { ($0.0, $0.1) }
```
stephencelis: This is just `id`, isn't it? Can we get rid of this `map`?

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -8,6 +8,12 @@ import XCTest
 @testable import ReactiveExtensions
 @testable import ReactiveExtensions_TestHelpers
 
+typealias ProjectMessageThread = (Project, MessageThread)
+
+private func == (lhs: ProjectMessageThread, rhs: ProjectMessageThread) -> Bool {
+  return lhs.0 == rhs.0 && lhs.1 == rhs.1
```
stephencelis: P sure we already have equality operators for tuples for free. Can we delete this?

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -8,6 +8,12 @@ import XCTest
 @testable import ReactiveExtensions
 @testable import ReactiveExtensions_TestHelpers
 
+typealias ProjectMessageThread = (Project, MessageThread)
```
stephencelis: Think we can drop this type alias, too.

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -323,6 +331,38 @@ internal final class DashboardViewModelTests: TestCase {
...
+      self.scheduler.advance()
+
+      self.vm.inputs.viewWillAppear(animated: false)
+      self.scheduler.advance()
+
+      self.goToMessageThread.assertValues([threadProj],
+                                          "Go to message thread not emitted again when view appears")
+
+      self.project.assertValues([threadProj, threadProj],
+        "Keep previeiosly selected project when view Appers")
```
stephencelis: Coupla typos.

Kickstarter-iOS/AppDelegate.swift
bormind: Thank you Stephen. I will try. I think I actually tried but because of the nullable XCode was complaining. I will double-check.

bormind: It worked! Fixed it.

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
bormind: Sorry @stephencelis, I didn't understand this one :)

stephencelis: Ah, just indentation doesn't match the others. Can cut and paste to let Xcode reindent :)

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -413,9 +413,7 @@ AppDelegateViewModelOutputs {
       .switchMap { projectId, messageThreadId in
         AppEnvironment.current.apiService.fetchMessageThread(messageThreadId: messageThreadId)
           .demoteErrors()
-          .map { (env: MessageThreadEnvelope) in
-            return (projectId, env.messageThread)
-          }
+          .map { env in return (projectId, env.messageThread) }
```
stephencelis: Can get rid of the `return` here (and probably use `$0`).

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -427,9 +425,7 @@ AppDelegateViewModelOutputs {
       .switchMap {
         AppEnvironment.current.apiService.fetchMessageThread(messageThreadId: $0)
           .demoteErrors()
-          .map { (env: MessageThreadEnvelope) in
-            return env.messageThread
-          }
+          .map { env in return env.messageThread }
```
stephencelis: Can get rid of the `return` here (and probably use `$0`).

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -151,13 +147,13 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
...
     self.goToMessageThread = projectsAndSelected
-      .map { $0.1 }
+      .map(second)
       .switchMap { [messageThreadReceived = self.messageThreadReceived.producer] project in
         messageThreadReceived
           .skipNil()
           .filter { $0.0 == .id(project.id) }
-          .map { (messgeThreadPair: (Param, MessageThread)) -> (Project, MessageThread) in
-            return (project, messgeThreadPair.1)
+          .map { (messageThreadPair: (Param, MessageThread)) -> (Project, MessageThread) in
```
stephencelis: Can this whole thing just be:

``` swift
.map { (project, $1) }
```

Library/ViewModels/DashboardViewModel.swift
justinswart: Hey @stephencelis so there are some race conditions here in the way projects are refreshed each time the view appears. This is the solution @bormind and I came up with - to use these 'interim' `MutableProperty`s to have a way of clearing the last stored value or basically injecting a `nil` into the stream when necessary.

What we were seeing for example was:

1. Navigate to message thread (`messageThreadNavigatedProperty` emits the `Project` and `MessageThread`).
2. If we navigate away and then back, `projects` will `switchMap` over this again and `self.messageThreadReceived` will emit the same values.

So by doing the following:
```
self.messageThreadReceived <~ Signal.merge(
  self.viewWillDisappearProperty.signal.mapConst(nil),
  self.messageThreadNavigatedProperty.signal
)
```
We can inject a `nil` and ignore it when we need to.

There may certainly be another solution but this one worked for us for now! Happy to hear alternative/better approaches.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,34 +102,69 @@ public protocol DashboardViewModelType {
 public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewModelOutputs,
   DashboardViewModelType {
 
-    public init() {
+  //Project selected by user in the drawer or by push for the project message thread
+  private let selectedProjectProperty: SignalProducer<Param?, NoError>
```
stephencelis: These private consts could be instantiated in the init and not live on the view model.

stephencelis: Also, this one isn't a property, right? Should it be called `selectedProjectProducer`?

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -87,34 +102,69 @@ public protocol DashboardViewModelType {
...
           .map { $0.projects }
+          .prefix(value: [])
     }
 
+    self.selectedProjectProperty = SignalProducer.merge(
+      self.switchToProjectProperty.producer,
+      self.messageThreadNavigatedProperty.producer.skipNil().map(first)
+    )
+
+    self.selectProjectPropertyOrFirst <~ SignalProducer.combineLatest(
```
stephencelis: Gonna try to explore some alternate ways of handling this tomorrow. We have a few isolated cases of using one-off producers for things, but it's definitely the exception and not the rule. Curious if we can find a more direct way of handling signals here.

Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -105,6 +105,16 @@ public final class RootTabBarViewController: UITabBarController {
     profileNav.setViewControllers([profileVC, threadsVC, messageThreadVC], animated: true)
   }
 
+  public func switchToCreatorMessageThread(projectId: Param, messageThread: MessageThread) {
+    self.switchToDashboard(project: nil)
```
bormind: @stephencelis, it has first to navigate to the dashboard `self.switchToDashboard(project: nil)` and then try to obtain the first viewcontroller in the visual stack and if it is a DashboardViewController (navigation to the dashboard was successful) then continue to the message thread

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -102,9 +102,6 @@ public protocol DashboardViewModelType {
 public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewModelOutputs,
   DashboardViewModelType {
 
-  //Project selected by user in the drawer or by push for the project message thread
-  private let selectedProjectProperty: SignalProducer<Param?, NoError>
```
stephencelis: Sorry if I wasn't clear! I think all three `const`s can be local to the function (which includes the two mutable properties below this one).

justinswart:
@stephencelis are we good to merge this or did you want to play with some more ideas? Thinking we could also merge this and PR improvements later on?

bormind:
Removed check for the view appearance animation fro retrieval of the projects.
This will fix the issue that if push happens before we looked at the dashboard for the first time the dashboard appears animated and projects are not fetched.
This will cause refresh of the projects when we navigate back to dashboard in visual stack (from activities or messages). But it will make it more consistent in the  project retrieval logic: view appeared - project refetched.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -103,7 +103,7 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
   DashboardViewModelType {
 
   public init() {
-    let projects = self.viewWillAppearAnimatedProperty.signal.filter(isFalse).ignoreValues()
+    let projects = self.viewWillAppearAnimatedProperty.signal.ignoreValues()
```
bormind: Removed check for the view appearance animation fro retrieval of the projects.
This will fix the issue that if push happens before we looked at the dashboard for the first time the dashboard appears animated and projects are not fetched.
This will cause refresh of the projects when we navigate back to dashboard in visual stack (from activities or messages). But it will make it more consistent in the project retrieval logic: view appeared - project refetched.

bormind:
Thank you @justinswart and @stephencelis

PR #224, opened 2017-08-22
Remove placeholders from Storyboards

stephencelis:
This PR removes our placeholder content from Storyboards. We've had a few reports of placeholder content showing up briefly when screens first appear (especially for cells), so removing placeholder images and text should prevent these occasional glitches.

I've done a sweep to remove content from our storyboards. It makes storyboards slightly more cumbersome to navigate (something we can explore further in the future), but it also revealed a few minor issues in the process.


Kickstarter-iOS/Views/Controllers/FacebookConfirmationViewController.swift
```diff
@@ -67,17 +67,17 @@ internal final class FacebookConfirmationViewController: UIViewController,
     _ = self.rootStackView |> loginRootStackViewStyle
   }
 
-    override func bindViewModel() {
+  override func bindViewModel() {
     super.bindViewModel()
 
     self.viewModel.outputs.displayEmail
-      .observeForControllerAction()
+      .observeForUI()
```
stephencelis: Subtleness here prevented the email from rendering immediately. Screenshot tests were rendering the placeholder!

Kickstarter-iOS/Views/Controllers/RewardPledgeViewController.swift
```diff
@@ -305,6 +305,7 @@ internal final class RewardPledgeViewController: UIViewController {
     _ = self.orLabel
       |> UILabel.lens.font .~ .ksr_footnote()
       |> UILabel.lens.textColor .~ .ksr_dark_grey_500
+      |> UILabel.lens.text %~ { _ in "– or –" }
```
stephencelis: This should be a FIXME. We haven't translated this.

Kickstarter-iOS/Views/Controllers/RewardPledgeViewControllerTests.swift
```diff
@@ -13,6 +13,7 @@ internal final class RewardPledgeViewControllerTests: TestCase {
     |> Project.lens.state .~ .live
   fileprivate let cosmicReward = Project.cosmicSurgery.rewards.last!
     |> Reward.lens.shipping.enabled .~ true
+    |> Reward.lens.shipping.summary .~ "Wherever"
```
stephencelis: A good example of where multiple optional fields on a struct/product type leads to gotchas! Screenshots rendered storyboard template data since the data came back `nil`. Unlikely event in production, but possible, especially with a backend change!

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -139,6 +139,7 @@ internal final class SettingsViewController: UIViewController {
       |> UIButton.lens.titleColor(forState: .normal) .~ .ksr_text_dark_grey_900
       |> UIButton.lens.titleLabel.font .~ .ksr_body()
       |> UIButton.lens.contentHorizontalAlignment .~ .left
+      |> UIButton.lens.title(forState: .normal) .~ "Debug push notifications"
```
stephencelis: This is beta-only content, but we were relying on the storyboard for it.

Kickstarter-iOS/Views/Controllers/TwoFactorViewController.swift
```diff
@@ -48,13 +48,16 @@ internal final class TwoFactorViewController: UIViewController {
 
     _ = self.resendButton
       |> borderButtonStyle
+      |> UIButton.lens.title(forState: .normal) %~ { _ in Strings.two_factor_buttons_resend() }
```
stephencelis: Turns out none of the two-factor auth strings were being used!

Library/ViewModels/RewardPledgeViewModel.swift
```diff
@@ -362,8 +362,7 @@ RewardPledgeViewModelOutputs {
     self.expandRewardDescription = self.expandDescriptionTappedProperty.signal
 
     self.shippingLocationsLabelText = reward
-      .map { $0.shipping.summary }
-      .skipNil()
+      .map { $0.shipping.summary ?? "" }
```
stephencelis: The shipping summary bug: if shipping data exists but the summary is returned as `nil` from the server, we'd display the storyboard content.

Library/ViewModels/FindFriendsFriendFollowCellViewModel.swift
```diff
@@ -78,10 +78,10 @@ public final class FindFriendsFriendFollowCellViewModel: FindFriendsFriendFollow
       .map { $0.stats.backedProjectsCount ?? 0 }
       .map(Strings.social_following_friend_projects_count_backed(backed_count:))
 
-    self.hideProjectsCreated = friend.map { $0.stats.createdProjectsCount == 0 }
```
stephencelis: We might show the storyboard "1 created" string if the `createdProjectsCount` is `nil`! This is because we're only doing a strict equality check against `0` and not accounting for `nil`.

Kickstarter-iOS/Views/Controllers/FacebookConfirmationViewController.swift
justinswart: Good catch!

Kickstarter-iOS/Views/Controllers/RewardPledgeViewController.swift
justinswart: Not worth adding a `localizedString(...)` here in the mean time? I usually search the project for occurrences of `localizedString` from time to time to check that we are using the `Strings` functions.

Kickstarter-iOS/Views/Controllers/TwoFactorViewController.swift
justinswart: 😳

Kickstarter-iOS/Views/Controllers/RewardPledgeViewController.swift
stephencelis: Done!

PR #225, opened 2017-08-23
Dashboard data fix

stephencelis:
Referrer data comes back as a decimal, but we were converting to `Int`, losing precision along the way.

### DIFF
--- a/Library/ViewModels/DashboardReferrerRowStackViewViewModel.swift
+++ b/Library/ViewModels/DashboardReferrerRowStackViewViewModel.swift
@@ -38,7 +38,7 @@ public final class DashboardReferrerRowStackViewViewModel: DashboardReferrerRowS
 
     self.pledgedText = countryReferrer
       .map { country, referrer in
-        Format.currency(referrer.pledged, country: country) + " ("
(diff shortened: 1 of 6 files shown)


Library/ViewModels/DashboardReferrerRowStackViewViewModel.swift
```diff
@@ -38,15 +38,15 @@ public final class DashboardReferrerRowStackViewViewModel: DashboardReferrerRowS
 
     self.pledgedText = countryReferrer
       .map { country, referrer in
-        Format.currency(referrer.pledged, country: country) + " ("
+        Format.currency(Int(referrer.pledged), country: country) + " ("
```
justinswart: So we're losing precision here again, no? Is the change just to keep precision through to Koala tracking?

stephencelis: This is fine. It's a way of `floor`ing before rendering. The main issue was here, where we were summing the numbers up, which could result in a much lower number than expected: https://github.com/kickstarter/ios-oss/pull/225/files#diff-e9f9d9570eaf3e04d6369eafbe3c991cR116

PR #226, opened 2017-08-23
Fix build for Xcode 9 beta 6

stephencelis:
The latest couple Xcode betas don't compile.

> Command failed due to signal: Illegal instruction: 4

This commit should allow us to work in Xcode 9 again.

### DIFF
--- a/Library/ViewModels/DiscoveryFiltersViewModel.swift
+++ b/Library/ViewModels/DiscoveryFiltersViewModel.swift
@@ -214,17 +214,11 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
 private func expandableRows(selectedRow: SelectableRow,
                             categories: [KsApi.Category]) -> [ExpandableRow] {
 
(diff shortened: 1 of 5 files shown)


Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -215,9 +214,17 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
 private func expandableRows(selectedRow: SelectableRow,
                             categories: [KsApi.Category]) -> [ExpandableRow] {
 
-  let expandableRows = categories
+  var grouped: [KsApi.Category:[KsApi.Category]] = [:]
```
justinswart: I think we could satisfy the compiler here and lose the `var` and `forEach`:

```swift
let expandableRows = categories
    .sorted { lhs, _ in !lhs.isRoot }
    .groupedBy { category -> KsApi.Category in
      guard let parent = category.parent else { return category }
      return parent
    }
```

Seems to build for me 👍

PR #230, opened 2017-08-29
Screenshot correction

cdolm92:
# What

New screenshots.

# Why

To correct screenshot errors on master.

# See 👀

See screenshots below.


stephencelis:
Master's screen shots are up-to-date now, so we can punt on this. Should still figure out why Xcode 9 has caused Xcode 8 to render screen shots differently, though.

PR #231, opened 2017-08-29
Generalize currency conversion

stephencelis:
Adds support for currency rates from our API.


stephencelis:
(Test fixes comin.)

KsApi/models/lenses/Project.StatsLenses.swift
```diff
@@ -4,44 +4,67 @@ extension Project.Stats {
...
+                   goal: $1.goal, pledged: $1.pledged, staticUsdRate: $1.staticUsdRate, updatesCount: $0) }
     )
 
     public static let fundingProgress = Lens<Project.Stats, Float>(
       view: { $0.fundingProgress },
-      set: { .init(backersCount: $1.backersCount, commentsCount: $1.commentsCount, goal: $1.goal,
-        pledged: Int($0 * Float($1.goal)), staticUsdRate: $1.staticUsdRate, updatesCount: $1.updatesCount) }
+      set: { .init(backersCount: $1.backersCount, commentsCount: $1.commentsCount,
+                   currentCurrency: $1.currentCurrency, currentCurrencyRate: $1.currentCurrencyRate,
+                   goal: $1.goal, pledged: Int($0 * Float($1.goal)), staticUsdRate: $1.staticUsdRate,
```
justinswart: Feels a little strange that this bit `Int($0 * Float($1.goal))` is in the lens, should we move it to the model? 🤔

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -57,21 +57,31 @@ RewardCellViewModelOutputs {
...
       .filter { p, _ in
-        needsConversion(projectCountry: p.country, userCountry: AppEnvironment.current.config?.countryCode)
+        p.stats.currentCurrency
+          .map { needsConversion(projectCountry: p.country, currentCurrency: $0) }
+          ?? needsConversion(projectCountry: p.country,
+                             userCountry: AppEnvironment.current.config?.countryCode)
       }
       .map { project, rewardOrBacking in
+        let (country, rate) = zip(
+          project.stats.currentCurrency.flatMap(Project.Country.init(currencyCode:)),
```
justinswart: Compiler clever enough here for `.flatMap(.init(currencyCode:))`?

stephencelis: I don't think so. Since function isn't its own type and this is point-free.

KsApi/models/lenses/Project.StatsLenses.swift
stephencelis: This is a lens we already had on `fundingProgress`, which is a computed property on our model. Gotta have the inverse logic here to update the type accordingly.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -211,7 +211,6 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     return self.viewModel.outputs.applicationDidFinishLaunchingReturnValue
   }
-  // swiftlint:enable function_body_length
```
stephencelis: While this PR was goin on swiftlint 0.22.0 dropped, which now checks for stale linter rules. It doesn't find these trailing ones, so I did a quick manual check based on our disabled rules.

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -254,31 +253,31 @@ private func tabData(forUser user: User?) -> TabBarItemsData {
                          isMember: isMember)
 }
 
-extension TabBarItemsData: Equatable {}
-func == (lhs: TabBarItemsData, rhs: TabBarItemsData) -> Bool {
-  return lhs.items == rhs.items &&
-         lhs.isLoggedIn == rhs.isLoggedIn &&
-         lhs.isMember == rhs.isMember
+extension TabBarItemsData: Equatable {
+  static func == (lhs: TabBarItemsData, rhs: TabBarItemsData) -> Bool {
```
stephencelis: Quick lil static promotion. Bunch more of these we could update in the future. Cleans things mostly up when generic types are involved (since you can omit the generic params).

Kickstarter-iOS/ViewModels/UpdatePreviewViewModelTests.swift
```diff
@@ -38,7 +37,7 @@ final class UpdatePreviewViewModelTests: TestCase {
     self.vm.inputs.viewDidLoad()
 
     let previewUrl = "https://\(Secrets.Api.Endpoint.production)/projects/2/updates/1/preview"
-    let query = "client_id=\(self.apiService.serverConfig.apiClientAuth.clientId)"
+    let query = "client_id=\(self.apiService.serverConfig.apiClientAuth.clientId)&currency=USD"
```
stephencelis: Right now our way of sending device currency (from the region) is to include it as a param. This means web views should get currency for free. Alternatively we could have synced currency via config, but this could have slowed things during load.

Kickstarter-iOS/Views/Controllers/ProjectActivityViewControllerTests.swift
```diff
@@ -54,7 +54,7 @@ internal final class ProjectActivityViewControllerTests: TestCase {
   }
 
   func testVoiceOverRunning() {
-    withEnvironment(isVoiceOverRunning: { true }) {
+    withEnvironment(isVoiceOverRunning: const(true)) {
```
stephencelis: New swiftlint rule requires that functions with multiple function parameters avoids trailing closure syntax. For things like `either(ifLeft: { … }, ifRight: { … })` I like this, so keeping it enabled and switching these to use `const`.

Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
```diff
@@ -213,8 +213,6 @@ public final class ProjectPamphletContentViewController: UITableViewController {
   fileprivate func scrollingIsAllowed(_ scrollView: UIScrollView) -> Bool {
     return self.presentingViewController?.presentedViewController?.isBeingDismissed != .some(true)
       && (!scrollView.isTracking || scrollView.contentOffset.y >= 0)
-    // swiftlint:disable:previous force_unwrapping
-    // NB: this ^ shouldn't be necessary, looks like a bug in swiftlint.
```
stephencelis: :wave:

KsApi/MockService.swift
```diff
@@ -9,6 +9,7 @@ internal struct MockService: ServiceType {
   internal let serverConfig: ServerConfigType
   internal let oauthToken: OauthTokenAuthType?
   internal let language: String
+  internal let currency: String
```
stephencelis: Our service now holds onto the currency code to send to the server.

KsApi/ServiceType.swift
```diff
@@ -391,6 +393,7 @@ extension ServiceType {
   fileprivate var defaultQueryParams: [String:String] {
     var query: [String:String] = [:]
     query["client_id"] = self.serverConfig.apiClientAuth.clientId
+    query["currency"] = self.currency
```
stephencelis: Using a query param for now.

KsApi/models/Project.Country.swift
```diff
@@ -36,6 +36,19 @@ extension Project {
...
     // swiftlint:enable comma
+
+    public static let all: [Country] = [
+      .AU, .AT, .BE, .CA, .CH, .DE, .DK, .ES, .FR, .GB, .HK, .IE, .IT, .LU, .MX, .NL, .NO, .NZ, .SE, .SG, .US
+    ]
+  }
+}
+
+extension Project.Country {
+  public init?(currencyCode: String) {
```
stephencelis: Quick lil initializer to return a valid project country by currency code. Used in the reward view model logic.

KsApi/models/Project.Country.swift
```diff
@@ -67,10 +80,10 @@ extension Project.Country: EncodableType {
 
 extension Project.Country: Equatable {}
 public func == (lhs: Project.Country, rhs: Project.Country) -> Bool {
-  return lhs.countryCode == rhs.countryCode &&
-    lhs.currencySymbol == rhs.currencySymbol &&
-    lhs.currencyCode == rhs.currencyCode &&
-    lhs.trailingCode == rhs.trailingCode
+  return lhs.countryCode == rhs.countryCode
+    && lhs.currencySymbol == rhs.currencySymbol
```
stephencelis: We do leading operators everywhere else. Just a lil cleanup 😅

KsApi/models/Project.swift
```diff
@@ -50,6 +50,8 @@ public struct Project {
   public struct Stats {
     public let backersCount: Int
     public let commentsCount: Int?
+    public let currentCurrency: String?
+    public let currentCurrencyRate: Float?
```
stephencelis: Our API now returns the "current" currency of the response (which should be pass-through from the request), and the exchange rate against the project currency.

Library/AppEnvironment.swift
```diff
@@ -240,7 +240,8 @@ public struct AppEnvironment {
           basicHTTPAuth: service.serverConfig.basicHTTPAuth
         ),
         oauthToken: service.oauthToken,
-        language: current.language.rawValue
+        language: current.language.rawValue,
+        currency: current.locale.currencyCode ?? "USD"
```
stephencelis: We could do a more graceful regional fallback here...but not sure how to manage that.

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -57,21 +57,20 @@ RewardCellViewModelOutputs {
     }
     let projectAndReward = Signal.zip(project, reward)
 
-    self.conversionLabelHidden = project.map {
-      !needsConversion(projectCountry: $0.country, userCountry: AppEnvironment.current.config?.countryCode)
-    }
-
+    self.conversionLabelHidden = project.map(needsConversion(project:) >>> negate)
```
stephencelis: This is the meat of the new logic! We generalize the old and new paths (with or without the API returning `current_currency` to a `needsConversion(project:)`.

Cleaned things up by composing this against `negate`, and composing it in the filter below against `first`.

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -57,21 +57,20 @@ RewardCellViewModelOutputs {
...
     self.conversionLabelText = projectAndRewardOrBacking
-      .filter { p, _ in
-        needsConversion(projectCountry: p.country, userCountry: AppEnvironment.current.config?.countryCode)
-      }
+      .filter(first >>> needsConversion(project:))
       .map { project, rewardOrBacking in
+        let (country, rate) = zip(
```
stephencelis: Fun bit of FP: Optional zip coalescing to a default pair.

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -304,10 +303,23 @@ private func minimumRewardAmountTextColor(project: Project, reward: Reward) -> U
   }
 }
 
+private func needsConversion(project: Project) -> Bool {
+  guard let currentCurrency = project.stats.currentCurrency else {
+    return needsConversion(projectCountry: project.country,
+                           userCountry: AppEnvironment.current.config?.countryCode)
+  }
+  return currentCurrency == AppEnvironment.current.apiService.currency
```
stephencelis: If our API ever returns a different currency than the one we send along, we fall back to USD here. We may wanna reconsider that logic, though?

Kickstarter-iOS/AppDelegate.swift
justinswart: Nice, thanks!

Library/ViewModels/RewardCellViewModel.swift
justinswart: Pretty cool!

Library/ViewModels/RewardCellViewModel.swift
justinswart: Could we not keep these defaults on the `Config` and lens them in the first time?

stephencelis: Ah, I mistyped. I mean we don't show any conversion info.

PR #233, opened 2017-08-31
Creator dashboard project activity loading indicator

cdolm92:
# What

Added a loading indicator when navigating to dashboard tab.

# Why

Before adding the loading indicator the user would encounter a blank screen as the dashboard was loading.  The loading indicator gives the user the feedback that the dashboard is loading.

# See 👀

loader


Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -23,6 +25,10 @@ internal final class DashboardViewController: UITableViewController {
   internal override func viewDidLoad() {
     super.viewDidLoad()
 
+    self.tableView.backgroundView = self.backgroundView
```
justinswart: This approach feels like we're fighting `UIKit`. I know `UITableViewController`s can be pretty stubborn when it comes to subviews, but I'm wondering if it wouldn't be better to do this using an empty state or a cell type that we can display while loading.

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -52,9 +65,16 @@ internal final class DashboardViewController: UITableViewController {
     self.viewModel.inputs.viewWillDisappear()
   }
 
+  internal override func viewDidLayoutSubviews() {
+    super.viewDidLayoutSubviews()
+
+    self.loadingIndicatorView.center = self.backgroundView.center
```
justinswart: I think we should center using layout constraints where possible.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -148,6 +154,11 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
 
     self.project = projectsAndSelected.map(second)
 
+    self.loaderIsAnimating = Signal.merge(
+    //  self.viewDidLoadProperty.signal.map(const(true)),
```
justinswart: Stray comment here.

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
```diff
@@ -23,6 +25,10 @@ internal final class DashboardViewController: UITableViewController {
   internal override func viewDidLoad() {
     super.viewDidLoad()
 
+    self.tableView.backgroundView = self.backgroundView
+
+    self.tableView.insertSubview(self.loadingIndicatorView, aboveSubview: self.backgroundView)
```
stephencelis: Not sure if we do this elsewhere, but we gotta stop :(

From the iOS 11 release notes:

> Adding custom subviews directly to `UITableView` is not supported and might break in a future iOS release. (29094519)

We should start using container views to manage this.

Library/ViewModels/DashboardViewModel.swift
stephencelis: Beyond that, if we don't need it, we can drop the `Signal.merge` ;)

Kickstarter-iOS/Views/Controllers/DashboardViewController.swift
stephencelis: (Alternatively we could use a table header view or cell to manage this.)

justinswart:
@cdolm92 header's def a better approach 👍 let me know when you're done and I'll review again 🙂

cdolm92:
@justinswart thanks! going to do some testing will let you know after 😄

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -10,38 +10,41 @@ import XCTest
 
 internal final class DashboardViewModelTests: TestCase {
   internal let vm: DashboardViewModelType = DashboardViewModel()
+
+  internal let animateOutProjectsDrawer = TestObserver<(), NoError>()
```
justinswart: Thanks for alphabetizing 🙂

PR #234, opened 2017-09-05
Localize date format using template

stephencelis:
This PR updates our date formatter from using a general format string (`dateFormat`) to a localized template (using `setLocalizedDateFormatFromTemplate`). This ensures that Japanese dates render appropriately, with the year appearing before the month, and with the appropriate local characters.


luoser:
super sweet that you can do this in swift!

PR #237, opened 2017-09-08
Add licenses for more third-party dependencies

christopherwright:
# What

Add a few licenses for third-party dependencies. If the dependency is contained within this repository, or it's included as a submodule under the Kickstarter org, I've included its license under `NOTICE.md`.

# Why

Ensuring we comply with licenses from our dependencies.


justinswart:
Cool, cool.

So, just confirming - it's not necessary to include OpenTok's license?

https://tokbox.com/support/sdk_license

christopherwright:
@justinswart That's our understanding. We're not distributing OpenTok as part of this repository, it's just installed as a dependency from their servers as part of the bootstrap process.

PR #238, opened 2017-09-08
Project update activity indicator

Scollaco:
# What

Add a ActivityIndicator on ProjectUpdatesViewController.

# Why

Improves user experience by indicating to the user that the page is loading, instead of showing a blank page.

# See 👀

screen


Kickstarter-iOS/Views/Controllers/ProjectUpdatesViewController.swift
```diff
@@ -5,7 +5,7 @@ import SafariServices
 
 internal final class ProjectUpdatesViewController: WebViewController {
   fileprivate let viewModel: ProjectUpdatesViewModelType = ProjectUpdatesViewModel()
-
+  fileprivate let activityIndicator = UIActivityIndicatorView()
```
justinswart: Oh one thing - I think at some point we said we'd change `fileprivate`s back to `private` as we encounter them in the codebase but no big deal!

Scollaco:
Thanks, @justinswart. It will be great to talk about the screenshots!

PR #240, opened 2017-09-12
Fixed bug that makes  spinner persist on reward drill-in

Scollaco:
# What

Fixed bug on which activity indicator wouldn't disappear from RewardPledgeViewController.

# Why

https://trello.com/c/iwWkq5Gd/117-spinner-persists-on-reward-drill-in.

# See 👀

screenbug


justinswart:
@Scollaco merging this in so we can get the bugfixes out 👍

PR #242, opened 2017-09-15
Use private(set) var over let

stephencelis:
Needed for key path lenses to work (produce writable key paths).


luoser:
thx fr evrythng stphn

PR #243, opened 2017-09-15
One last PR!

stephencelis:
Here's an example of some key path lenses! 🎁

Now y'all get to delete lots of code 😄


Scollaco:
It looks good! Thank you again, @stephencelis !

PR #244, opened 2017-09-15
Add graphQL to ksapi

Scollaco:
# What

• Implement graphQL.
• Start consuming Categories using graphQL 

# Why

Retrieve only the necessary data from the server in order to make the app faster and more efficient.

# TODO ⏰

- [x] Finish the transition of Categories api to graphQL
- [x] Add more tests
- [x] Code cleanup


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -4163,7 +4166,7 @@
 					};
 					A7D1F9441C850B7C000D41D5 = {
 						CreatedOnToolsVersion = 7.2.1;
-						DevelopmentTeam = 48YBP49Y5N;
+						DevelopmentTeam = 5DAN4UM3NC;
```
justinswart: Hey let's not commit these changes, it might confuse CircleCI 👍 I'm guessing this was to get it to build locally on device?

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -6382,7 +6386,7 @@
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = 48YBP49Y5N;
+				DEVELOPMENT_TEAM = 5DAN4UM3NC;
```
justinswart: Revert this line too please.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -6399,8 +6403,8 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.kickstarter.kickstarter.alpha;
 				PRODUCT_MODULE_NAME = Kickstarter_iOS;
 				PRODUCT_NAME = KickDebug;
-				PROVISIONING_PROFILE = "";
-				PROVISIONING_PROFILE_SPECIFIER = "Alpha Development";
+				PROVISIONING_PROFILE = "186a3515-ae79-4f05-8f2a-cd8af78d80c2";
+				PROVISIONING_PROFILE_SPECIFIER = "Brandon Development";
```
justinswart: And these two lines 🙂

KsApi/GraphSchema.swift
```diff
@@ -0,0 +1,398 @@
...
+public enum Edges<Q: QueryType> {
+  case cursor
+  case node(NonEmptySet<Q>)
+}
+
+public enum Nodes<Q: QueryType> {
+  case nodes(NonEmptySet<Q>)
+}
+
+//public struct GraphError: Error {
```
justinswart: Still needed?

KsApi/Service.swift
```diff
@@ -184,6 +184,10 @@ public struct Service: ServiceType {
     return request(.friendStats)
   }
 
+  public func fetchGraph<A: Swift.Decodable>(query: NonEmptySet<Query>) -> SignalProducer<A, GraphError> {
```
justinswart: Any reason to not just expose `fetch(_:)` directly?

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -316,10 +304,47 @@ private func favorites(selectedRow: SelectableRow, categories: [KsApi.Category])
...
+  let rootCategories: [Category]
+
+   struct Category: Swift.Decodable {
+    let id: String
+    let name: String
+    let subcategories: SubcategoryConnection
+
+     struct SubcategoryConnection: Swift.Decodable {
+      let totalCount: Int
+      let nodes: [Node]
```
justinswart: Might be able to generalize `Node` to `Node<T>` for better readability when using nodes in other places like `Node<Category>` or `Node<Project>`. I believe it'll still work with `Swift.Decodable`?

Kickstarter.xcodeproj/project.pbxproj
Scollaco: Reverting now. These changes were commited by accident :)

KsApi/GraphSchema.swift
Scollaco: The GraphError struct is no longer needed, so I will delete it

KsApi/Service.swift
Scollaco: In this case I was just following the structure of the current Service.swift file. It may be a good idea to move the graph related functions to a separated file (I think that's how it is in our private repo). What do you think?

Library/ViewModels/DiscoveryFiltersViewModel.swift
Scollaco: It makes sense. Yes, it will work with Swift.Decodable if we declare node as Node<T: Swift.Decodable>

KsApi/Service.swift
justinswart: Yeah I had separated them out into two different services before and kept the shared functionality in `ServiceType`, I felt like that made it easier to work with but it's up to you! The last stuff that I worked on related to that is in this branch https://github.com/kickstarter/ios-private/tree/subs-pledge

justinswart: Yeah I had separated them out into two different services before and kept the shared functionality in ServiceType, I felt like that made it easier to work with but it's up to you! Will send you the link on Slack 👍

KsApi/MockService.swift
```diff
@@ -514,6 +515,19 @@ internal struct MockService: ServiceType {
     )
   }
 
+  func fetchGraphCategories(query: NonEmptySet<Query>) -> SignalProducer<RootCategoriesEnvelope, GraphError> {
+    return SignalProducer(value: self.fetchGraphCategoriesResponse!)
```
justinswart: I think it'd be better to do an `if let response = self.fetchGraphCategoriesResponse { ... }` and if it's `nil` just return `.empty`. Better to have a failing test than tests crashing altogether because of potentially unwrapping a `nil` value.

KsApi/MockService.swift
```diff
@@ -514,6 +515,19 @@ internal struct MockService: ServiceType {
...
+  }
+
+  func fetchGraphCategory(query: NonEmptySet<Query>)
+    -> SignalProducer<RootCategoriesEnvelope.Category, GraphError> {
+    return SignalProducer(value: .template |> RootCategoriesEnvelope.Category.lens.id .~ "\(query.head)")
```
justinswart: This should be a property on the `MockService` so that we can control it from tests.

KsApi/MockService.swift
```diff
@@ -514,6 +515,19 @@ internal struct MockService: ServiceType {
...
+  }
+
+  internal func fetchGraph<A>(query: NonEmptySet<Query>) -> SignalProducer<A, GraphError> where A: Decodable {
+    return .empty
```
justinswart: If we never need to mock this then we could perhaps remove it from the interface and always create query specific functions?

KsApi/models/DiscoveryParams.swift
```diff
@@ -147,3 +148,30 @@ private func stringIntToBool(_ string: String?) -> Decoded<Bool?> {
     .map { .success($0 == 0 ? nil : $0 == 1) }
     .coalesceWith(.failure(.custom("Could not parse string into bool.")))
 }
+
+private func decodeToGraphCategory(_ json: JSON?) -> Decoded<RootCategoriesEnvelope.Category> {
+
+  guard let jsonObj = json else {
+    return .success(RootCategoriesEnvelope.Category(id: "-1", name: "Unknown Category"))
```
justinswart: Not sure I understand the `-1` bit?

KsApi/models/DiscoveryParams.swift
```diff
@@ -147,3 +148,30 @@ private func stringIntToBool(_ string: String?) -> Decoded<Bool?> {
...
+  case .object(let dic):
+    let category = RootCategoriesEnvelope.Category(id: categoryInfo(dic).0,
+                                                   name: categoryInfo(dic).1)
+    return .success(category)
+  default:
+    return .failure(DecodeError.custom("JSON should be object type"))
+  }
+}
+
+private func categoryInfo(_ json: [String: JSON]) -> (String, String) {
```
justinswart: Might be better to have the return type be `(String, String)?` and handle the `nil` case at the call site?

KsApi/models/GraphCategory.swift
```diff
@@ -0,0 +1,132 @@
...
+public struct RootCategoriesEnvelope: Swift.Decodable {
+  public let rootCategories: [Category]
+
+  public struct Category: Swift.Decodable {
+    public let id: String
+    public var intID: Int? {
+        return decompose(id: id)
+    }
+
+    public static let gamesId: Int = 12
```
justinswart: Why this one hard-coded value?

KsApi/models/GraphCategory.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    public let name: String
+    public let parentId: String?
+    internal let parentCategory: ParentCategory?
+    public var _parent: RootCategoriesEnvelope.Category? {
```
justinswart: Was this previously exposed as `_parent`? We could make the `_parent` `private` and expose `parent`, we'd just need to use `CodingKeys` to decode it to the correct property. Of course that would mean implementing the `Decodable` initializer but might be worth not having to refer to `_parent` where it's use in the app... 🤔

KsApi/models/lenses/GraphCategoryLenses.swift
```diff
@@ -0,0 +1,67 @@
+import Prelude
+
+extension RootCategoriesEnvelope.Category {
+
+  public enum lens {
```
justinswart: Couldn't we do these lenses the way they were done in #243?

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -140,6 +143,10 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
       )
       .map { (rows: $1, categoryId: $0.0, selectedRowId: $0.1) }
 
+    self.loadCategoryRows.signal.observeValues { values in
+      print(values)
```
justinswart: You probably don't need this `print` anymore.

Kickstarter-iOS/Views/Cells/DiscoveryExpandedSelectableRow.swift
```diff
@@ -12,7 +12,7 @@ internal final class DiscoveryExpandedSelectableRowCell: UITableViewCell, ValueC
 
   internal func configureWith(value: (row: SelectableRow, categoryId: Int?)) {
     if let category = value.row.params.category, category.isRoot {
-      self.filterTitleLabel.text = RootCategory(categoryId: category.id).allProjectsString()
+      self.filterTitleLabel.text = RootCategory(categoryId: category.intID ?? -1).allProjectsString()
```
justinswart: `-1`?

KsApi/ServiceType.swift
```diff
@@ -106,6 +100,16 @@ public protocol ServiceType {
   /// Fetch friend stats.
   func fetchFriendStats() -> SignalProducer<FriendStatsEnvelope, ErrorEnvelope>
 
+  /// Fetch generic objects using graphQL.
+  func fetchGraph<A: Swift.Decodable>(query: NonEmptySet<Query>) -> SignalProducer<A, GraphError>
```
justinswart: We could get around the `fetchGraph` prefixes by having a `nested` protocol type, similar to how we do `inputs` and `outputs` on view models. Not a big deal but might make it read better as `apiService.graph.fetchCategories`, makes it seems less as if `GraphCategory` is a thing 😄

justinswart:
Also there are a couple failing tests on circle that I'm sure you're still planning on getting to 👍

KsApi/MockService.swift
Scollaco: In this case I'm unwrapping the optional because there is no way for it to be nil, since it will returns a template if the value of `fetchGraphCategoriesResponse` is nil. But I agree that it will make the test code more reliable adding a `if` statement.

KsApi/MockService.swift
Scollaco: Oh right, I will remove it. Should have done this in the past. Good catch!

KsApi/ServiceType.swift
Scollaco: This function actually will be removed since after some refactoring and code cleanup, it's not necessary anymore, because we can simply call the private `fetch` function like:  
`public func fetchGraphCategories(query: NonEmptySet<Query>) -> SignalProducer<RootCategoriesEnvelope, GraphError> {
    return fetch(query: query)
  }`

KsApi/models/DiscoveryParams.swift
Scollaco: -1 is the id of a unrecognized category. a need to pass a value to the `RootCategory` class initializer, so it can be uesd on the expandable/selectable cells configuration and `DiscoveryNavigationHeaderViewController`

KsApi/models/GraphCategory.swift
Scollaco: This code i copied from the old `Category` class. it's related to some logic to show or not a Rating Alert depending if the category is `Game`.

KsApi/models/DiscoveryParams.swift
Scollaco: That's how it was before. I don't know why I changed it :)

KsApi/models/GraphCategory.swift
Scollaco: I agree!

KsApi/models/lenses/GraphCategoryLenses.swift
Scollaco: Oh yes! Will do.

Library/ViewModels/DiscoveryFiltersViewModel.swift
Scollaco: good catch. this line should have been deleted.

Scollaco:
Thank you for the the time to review, @justinswart! I will make some changes and I'm also aware of the failing tests. I had to fix a bug and will need to fix some tests since I had to change some code a little bit.

Kickstarter-iOS/Views/Cells/DiscoveryExpandedSelectableRow.swift
Scollaco: -1 is the id for unrecognized category. I've explained with more detail some comments below.

KsApi/models/DiscoveryParams.swift
```diff
@@ -147,3 +148,30 @@ private func stringIntToBool(_ string: String?) -> Decoded<Bool?> {
...
+  default:
+    return .failure(DecodeError.custom("JSON should be object type"))
+  }
+}
+
+private func categoryInfo(_ json: [String: JSON]) -> (String, String)? {
+  guard let name = json["name"], let id = json["id"] else {
+    return nil
+  }
+  switch (id, name) {
```
manojmahapatra: you could perhaps use `if case` instead and return nil otherwise.

PR #245, opened 2017-09-15
Fix live stream event info page color

justinswart:
# What

At some point the background color of the info page changed for live stream events, probably during #199. This PR just puts that back to the correct grey.


cdolm92:
looks good 👍

PR #247, opened 2017-09-20
Fix signup translation in Japanese

PeterTeng:
# What

`ログインかサインイン` replace with `ログインかサインアップ`

# Why

The Japanese of signup should be `サインアップ` instead of `サインイン`.

|Japanese|English|
|:---:|:---:|
|`ログインかサインイン`|`login or sign in`|
|ログインかサインアップ|`login or sign up`|

This is my first time to make a pull request here. I have an issue that I can't run `make test-all`
It shows below when i run `make test-all`

```
   .
   .
   .
Testing failed:
	Use of undeclared type 'WritableKeyPath'
	Use of undeclared type 'WritableKeyPath'
	Use of undeclared type 'WritableKeyPath'
	Use of undeclared type 'WritableKeyPath'
	Use of undeclared type 'WritableKeyPath'
** TEST FAILED **
The following build commands failed:
	CompileSwift normal x86_64 /Users/PeterTeng/workspace/ios-oss/Frameworks/Prelude/Prelude/Lens.swift
	CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(2 failures)
make[1]: *** [test] Error 65
make: *** [test-all] Error 2
DONGSHIH:ios-oss[peter/fix-jp-local]🦁
```

### DIFF
--- a/Kickstarter-iOS/Locales/ja.lproj/Localizable.strings
+++ b/Kickstarter-iOS/Locales/ja.lproj/Localizable.strings
@@ -154,8 +154,8 @@
 "Live_streaming_now" = "ライブ配信中";
 "Live_with_creator_name" = "<b>%{creator_name}</b>のライブ配信";
 "Loading" = "読み込み中";
-"Log_in_or_sign_up_to_save_this_project_and_we_ll_remind_you" = "ログインかサインインしてこのプロジェクトを保存してください。終了４８時間前にリマインドします。";
-"Log_in_or_sign_up_to_subscribe" = "ログインかサインインしてください";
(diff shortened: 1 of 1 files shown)


justinswart:
@PeterTeng hey thanks for finding that and for letting us know! We won't be able to merge this PR unfortunately due to the way that we generate our `Strings` [file](https://github.com/kickstarter/ios-oss/blob/master/Library/Strings.swift) using [code generation](https://github.com/kickstarter/ios-oss/blob/master/bin/strings.swift). This is a part of our internal l10n process.

Our internationalization team is aware of the translation that you reported and you'll see a fix in an upcoming version of the app.

But please do open an issue on the repo if you find any more translation errors!

PR #248, opened 2017-09-21
Unavailable reward placement (native)

Scollaco:
# What

Re-order rewards as follow:
• no reward tier
• available rewards, sorted from $ to $$$
• unavailable rewards, sorted from $ to $$$

# Why

https://trello.com/c/EBxaeKOA/66-unavailable-reward-placement-native

# See 👀

screen shot 2017-09-21 at 4 11 23 pm


justinswart:
Hey @Scollaco I think this looks cool - Could I suggest: 

- let's move the sorting into a testable function and add a test for it in `ProjectPamphletContentDataSourceTests`? (we can create that if it doesn't exist yet).
- let's refer to `availableRewards` and `unavailableRewards` for consistency I think, unless _allGone_ is something we refer to elsewhere?

Kickstarter-iOS/DataSources/ProjectPamphletContentDataSource.swift
```diff
@@ -55,9 +56,28 @@ internal final class ProjectPamphletContentDataSource: ValueCellDataSource {
       .sorted()
       .map { (project, Either<Reward, Backing>.left($0)) }
 
+    let availableRewards = project.rewards
+      .filter { isMainReward(reward: $0, project: project) }
+      .filter { $0.remaining != 0 }
```
luoser: i think we should also add a check for if remaining is `nil` here, since it's an optional value and if it is nil the rewards are unlimited. this filter can then be more like:

```swift
.filter { $0.remaining == nil || $0.remaining != 0 }
```

Kickstarter-iOS/DataSources/ProjectPamphletContentDataSourceTests.swift
```diff
@@ -124,4 +124,34 @@ final class ProjectPamphletContentDataSourceTests: TestCase {
...
   }
+
+  func testAvailableRewardsSection_ShowsCorrectValues() {
+    let availableSection = ProjectPamphletContentDataSource.Section.availableRewards.rawValue
+    let unavailableSection = ProjectPamphletContentDataSource.Section.unavailableRewards.rawValue
+
+    let reward = Reward.template
+      |> Reward.lens.remaining .~ 1
+    let project = Project.template
+                  |> Project.lens.rewards .~ [reward]
```
justinswart: Alignment got a little off here.

Scollaco: Good catch!

PR #249, opened 2017-09-25
Update README to point to Xcode 9 download

stringerbell:
# What

Update to the xcode link in the readme

# Why

So no one else downloads Xcode 8.3.3 when they're getting started. 😄


justinswart:
Thanks!

PR #250, opened 2017-09-26
Put cancel button back on MessageDialogViewController

justinswart:
# What

Quick fix to bring back the cancel button on `MessageDialogViewController`.

# Why

It's needed to dismiss the dialog if one does not want to send a message.

# How

I think it might have been removed during #224, so if "Cancel" was the default in the Storyboard then the button would have no text not not appear.

# See 👀

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/MessageDialogViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/MessageDialogViewController.swift
@@ -32,10 +32,8 @@ internal final class MessageDialogViewController: UIViewController {
 
   internal override func viewDidLoad() {
     super.viewDidLoad()
+    
     self.viewModel.inputs.viewDidLoad()
-
-    _ = self.cancelButton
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Views/Controllers/MessageDialogViewController.swift
```diff
@@ -32,6 +33,9 @@ internal final class MessageDialogViewController: UIViewController {
   internal override func viewDidLoad() {
     super.viewDidLoad()
     self.viewModel.inputs.viewDidLoad()
+
+    _ = self.cancelButton
+      |> UIBarButtonItem.lens.title %~ { _ in Strings.general_navigation_buttons_cancel() }
```
Scollaco: Is there any reason why we are not setting the title within the <bindStyles()> function?

justinswart: Oh true! I didn't really think about it... I can move it to `bindStyles()`. I guess to me a text isn't really a style but better to keep it consistent I agree 👍

PR #252, opened 2017-09-29
CircleCI Xcode 9 updates

justinswart:
# What

Some changes needed to build our app using Xcode 9 on CircleCI.

Changes are:

- Set CircleCI to use Xcode 9
- Allow fastlane's `gym` command to timeout after 30 mins

`gym` seems to take nearly 20 mins to successfully compile our app now and this was failing with the default 10 min timeout option. Increasing this to 30 mins gets us successful deployments. I would suggest that we keep an eye on this and if we see these times dropping in the future we can try removing this option.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -37,20 +37,20 @@ deployment:
     commands:
       - agvtool new-version -all $(($(date +%s)/100))
       - fastlane alpha_gym --verbose:
-        timeout: 1800
+          timeout: 1800
       - fastlane alpha_hockey --verbose
   ios_beta:
     branch: beta-dist
     commands:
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -12,7 +12,6 @@ dependencies:
     - security find-identity -p codesigning
     - instruments -s devices
     - xcodebuild -showsdks
-    - gem install fastlane -NVv 2.37.0
```
Scollaco: why do we need to delete this line?

justinswart: Circle fastlane now by default and the version that they provide is good for what we need 👍

PR #253, opened 2017-09-29
Fix double paste bug

justinswart:
# What

Fixed a bug that caused our shared links to copy twice to the paste board.

# Why

iOS 11 introduced a bug in that our shared links would be concatenated and therefore not share correctly.

This seems to be caused by putting a `URL` object directly into the `activityItems` array that we pass to instantiate a `UIActivityViewController`. iOS 11 appears to incorrectly interpret this in some way.

However we _do_ need to pass a URL to `activityItems` so that it is passed to our `SafariActivity` so that we can present the _Open in Safari_ item.

# How

Proposed solution here is to wrap the `URL` in a `SafariURL` so that iOS 11 does not try to internally interpret it but still passes it to our `SafariActivity` so that we might.

Not sure if this is the best way to go about it but it does seem to fix the issue, alternatives welcome!

cc @tiegz

### DIFF
--- a/Library/SafariActivity.swift
+++ b/Library/SafariActivity.swift
@@ -3,14 +3,21 @@ import Foundation
 import UIKit
 
 public let SafariActivityType = UIActivityType("com.kickstarter.kickstarter.safari")
+public final class SafariURL {
+  public let url: URL
+
+  public init(_ url: URL) {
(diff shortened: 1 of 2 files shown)


tiegz:
excellent

PR #255, opened 2017-10-05
IPhone X project layout fix

Scollaco:
# What

The top of the project detail page was being "covered" by the iPhone X notch.

# How

Changing the topConstraints of the view according to its parent's safeAreaInsets. (This change will occur dynamically as well if the device rotates).

# See

iphonex_statusbar


justinswart:
@Scollaco try merging in `master`, should fix most of the swiftlint issues.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -24,11 +27,11 @@ public final class ProjectPamphletViewController: UIViewController {
   }
 
   public override var prefersStatusBarHidden: Bool {
-    return self.viewModel.outputs.prefersStatusBarHidden
+    return UIApplication.shared.statusBarOrientation.isLandscape
   }
 
-  public override var preferredStatusBarUpdateAnimation: UIStatusBarAnimation {
-    return .fade
+  public override var preferredStatusBarStyle: UIStatusBarStyle {
```
justinswart: How does this change affect iOS 10?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -74,6 +83,20 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    self.projectPamphletTopConstraint.rac.constant = self.viewModel.outputs.topLayoutConstraintConstant
+  }
+
+  public override func willTransition(to newCollection: UITraitCollection,
+                                      with coordinator: UIViewControllerTransitionCoordinator) {
+    self.viewModel.inputs.willTransitionToNewCollection(to: newCollection)
+  }
+
+  private func update(constraints: [NSLayoutConstraint?], constant: CGFloat?) {
+    _ = constraints.map {
```
justinswart: This could just be a `forEach`.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -8,13 +8,19 @@ public protocol ProjectPamphletViewModelInputs {
...
   /// Call when the view loads.
   func viewDidLoad()
 
   func viewDidAppear(animated: Bool)
 
   /// Call when the view will appear, and pass the animated parameter.
   func viewWillAppear(animated: Bool)
+
+  /// Call when the view will transition to a new trait collection.
+  func willTransitionToNewCollection(to newCollection: UITraitCollection)
```
justinswart: Think this should be `willTransition(toNewCollection collection: UITraitCollection)`

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -60,13 +69,32 @@ ProjectPamphletViewModelOutputs {
...
+      self.viewWillAppearAnimated.signal.ignoreValues(),
+      self.willTransitionToCollectionProperty.signal.ignoreValues()
+    )
 
     self.setNavigationBarHiddenAnimated = Signal.merge(
       self.viewDidLoadProperty.signal.mapConst((true, false)),
       self.viewWillAppearAnimated.signal.skip(first: 1).map { (true, $0) }
     )
 
+    var initialTopConstraint: CGFloat = 0.0
```
justinswart: I need to look into this a little closer but I think there's probably a different, more 'FRP', approach to capturing this local variable.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -99,6 +127,11 @@ ProjectPamphletViewModelOutputs {
     self.configDataProperty.value = (projectOrParam, refTag)
   }
 
+  fileprivate let parentViewControllerProperty = MutableProperty<UIViewController?>(nil)
+  public func parent(viewController: UIViewController?) {
```
justinswart: Just wondering if we need to capture the whole parent VC here, what we're really interested in is the `safeAreaInsets.top` later on. Could we not update the VM by overriding `viewSafeAreaInsetsDidChange()` in the places that we care about?

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -114,14 +147,22 @@ ProjectPamphletViewModelOutputs {
     self.viewWillAppearAnimated.value = animated
   }
 
+  fileprivate let willTransitionToCollectionProperty =
+    MutableProperty<UITraitCollection>(UITraitCollection())
```
justinswart: The type here should be optional and instantiated with `nil`, then in the signal where we use it `skipNil()`. Alternatively we can consider using a `pipe`. I started doing this to avoid the `skipNil()`'s, I'll slack you an example.

Library/ViewModels/ProjectPamphletViewModel.swift
Scollaco: Good catch!

Library/ViewModels/ProjectPamphletViewModel.swift
Scollaco: For now I will use the `skipNil()` because all the other functions on the file uses this pattern. But will consider to use the `pipe()` alternative on future projects.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
Scollaco: I will change that

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
Scollaco: It should work fine for both iOS 10 and 11.

PR #261, opened 2017-10-16
Added shipping metadata on page rewards

Scollaco:
# What

Added Shipping summary on page rewards.

# Why

https://trello.com/c/sJ5bI434/95-display-shipping-metadata-on-project-page-rewards

# See 👀

screen shot 2017-10-16 at 12 18 55

### DIFF
--- a/Library/ViewModels/RewardCellViewModel.swift
+++ b/Library/ViewModels/RewardCellViewModel.swift
@@ -208,7 +208,7 @@ RewardCellViewModelOutputs {
     }
 
     self.shippingLocationsSummaryLabelText = reward.map {
-      " " + ($0.shipping.summary ?? "")
+      $0.shipping.summary ?? ""
     }
 
     let tappable = Signal.zip(project, reward, youreABacker)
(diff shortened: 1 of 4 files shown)


Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -201,6 +203,14 @@ RewardCellViewModelOutputs {
         : Strings.Select_this_reward()
     }
 
+    self.shippingLocationsStackViewHidden = reward.map {
+      $0.shipping.summary == nil
+    }
+
+    self.shippingLocationsSummaryLabelText = reward.map {
+      " " + ($0.shipping.summary ?? "")
```
cdolm92: Instead of adding a space here you can set up the spacing something like this `|> UIStackView.lens.spacing .~ Styles.gridHalf(1)` in the `RewardCell`.

Scollaco: Thank you, @cdolm92 . I will do that.

PR #262, opened 2017-10-17
Ios11 compatibility tweaks

Scollaco:
# What

Fixes Discovery header height discrepancy.

# Why

https://trello.com/c/lXSjlfPa/195-discovery-header-height-discrepancy
This discrepancy occurred due the new safeArea on XCode 9. So while with iOS 10 the layout would look normal (header keeping the same height as Search, Activity, etc), with iOS 11 an extra space(safeArea) is added to the top, making the Discovery header look taller than the other tabs.

# See 👀

# Before:
screen shot 2017-10-17 at 15 50 14

# After:
screen shot 2017-10-17 at 15 50 45

### DIFF
--- a/Library/ViewModels/DiscoveryViewModel.swift
+++ b/Library/ViewModels/DiscoveryViewModel.swift
@@ -22,6 +22,9 @@ public protocol DiscoveryViewModelInputs {
   /// Call from the controller's viewDidLoad.
   func viewDidLoad()
 
+  /// Call from the controller's viewSafeAreaInsetsDidChange.
+  func viewSafeAreaInsetsDidChange()
+
   /// Call from the controller's viewWillAppear.
   func viewWillAppear(animated: Bool)
 
@@ -50,6 +53,9 @@ public protocol DiscoveryViewModelOutputs {
 
   /// Emits when we should manually navigate to a sort's page.
   var navigateToSort: Signal<(DiscoveryParams.Sort, UIPageViewControllerNavigationDirection), NoError> { get }
+
+  /// Emits a CGFloat that should be used to configure the NavigationHeaderViewController height
+  var navigationHeaderHeightLayoutConstant: Signal<CGFloat, NoError> { get }
 
   /// Emits a sort that should be passed on to the sort pager view controller.
   var selectSortPage: Signal<DiscoveryParams.Sort, NoError> { get }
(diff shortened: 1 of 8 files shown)


Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -170,6 +180,12 @@ DiscoveryViewModelOutputs {
   public func viewDidLoad() {
     self.viewDidLoadProperty.value = ()
   }
+
+  private let navigationHeaderTopInsetProperty = MutableProperty<CGFloat?>(nil)
+  public func viewSafeAreaInsetsDidChange() {
+    self.navigationHeaderTopInsetProperty.value = 20
```
justinswart: Not sure if we want to use a constant like this - what is the `20`, the status bar height?

Scollaco: Yeah, I tried using the statusBar.frame.height, but the problem is that the statusBar height is different on iPhone X.

justinswart: I think what you want to do is change your input to be:

```swift
  private let viewSafeAreaInsetsDidChangeProperty = MutableProperty<UIEdgeInsets?>(nil)
  public func viewSafeAreaInsetsDidChange(to insets: UIEdgeInsets) {
    self.viewSafeAreaInsetsDidChangeProperty.value = insets
  }
```

And then your output:

```swift
self.navigationHeaderTopInset = self.viewSafeAreaInsetsDidChangeProperty.signal.skipNil()
      .map { $0.top }
      .skipRepeats()
```

As far as I can tell that should give the same desired result and not base it on a constant value.

Scollaco: Actually the `insets.top` will return the same value as the statusBar. So it would have the same result as if we used the statusBar height directly

justinswart:
Thanks @Scollaco I think it's better without the constant height 👍.

Just a question for you and @ktman, wasn't there some discussion at one point about keeping the status bar black on iOS 10?

If not then this should be good to merge.

ktman:
@justinswart after looking at the options with Ryan, we decided to stick with a white status bar for now and will revisit / address iPhone X implications with Danny. Happy to provide more context over Slack.

PR #269, opened 2017-11-16
Change Argo's url from http -> https

karanjthakkar:
# What

Update the url for Argo to use https instead of http

# Why

While running `make bootstrap` on my machine (macOS Sierra with git 2.7.2), I got this error:

screen shot 2017-11-16 at 7 38 16 pm

Apparently http is no longer supported by Github ¯\_(ツ)_/¯ So I thought maybe this will help someone who tries to set thing up locally.

# How

N/A

# See 👀

N/A

# TODO ⏰

N/A


justinswart:
Thanks!

PR #274, opened 2017-11-20
Privacy violation crash fix

Scollaco:
# What
Added new iOS 11 `NSPhotoLibraryAddUsageDescription` privacy key.

# Why
Some of our users are experiencing crashes (`__CRASHING_DUE_TO_PRIVACY_VIOLATION__`) and adding this key to the info.plist could solve this problem.

GIF

### DIFF
--- a/Kickstarter-iOS/Info.plist
+++ b/Kickstarter-iOS/Info.plist
@@ -96,6 +96,8 @@
 	</dict>
 	<key>NSCameraUsageDescription</key>
 	<string>Add photos and video to your project updates.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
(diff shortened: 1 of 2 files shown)


Scollaco:
[Investigate & fix this crash](https://trello.com/c/GTad4xr6/304-investigate-fix-this-crash)

Scollaco:
@justinswart Yeah, I saw that we already have the 'NSPhotoLibraryUsageDescription`. But what made me believe that the problem is with the other Photos key missing was this line of the stack trace: ` 0x000000018ebc99c4 __79-[PLPrivacy _isPhotosAccessAllowedWithScope:forceHandler:accessAllowedHandler:]_block_invoke_2 + 412`
We will see.

PR #275, opened 2017-11-21
Koala Push events bug fix

Scollaco:
# What
Koala is registering about 10x more PushNotifications permission events than it should. Since the Alert is showing during signup/login (if status is `.notDetermined`), the number of PushNotifications Permission events should be about the same as the number of times the Alert is shown.

# Why
https://trello.com/c/7oqpw5Vn/163-push-opt-in-events-not-firing-properly

GIF


Library/UIPageViewController+Extensions.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  UIPageViewController+Extensions.swift
```
justinswart: Doesn't look like we need this extension?

Scollaco: Actually this extension was added after I merged the branch with the master.

justinswart: Ok, but I guess we can remove it - seems like an empty extension

PR #279, opened 2017-11-29
Bug fix - Recommended filter

Scollaco:
# What
Fixes a bug that shows the EmptyState screen when user selects any personal filter (Recommended, Saved...).

# Why
https://trello.com/c/cQlCKsYS/275-p2-empty-states-appearing-for-personalized-discovery-filters

# How
The `showEmptyState` signal was emitting because `paginatedProjects` sometimes sends false empty arrays (first sending an empty array and sending an array of Projects right after). This behaviour was making the app show the EmptyState screen when we actually had projects to be shown. 

# See
simulator screen shot - iphone 8 - 2017-11-29 at 15 28 12

GIF

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
@@ -240,10 +240,10 @@ internal final class DiscoveryPageViewController: UITableViewController {
     self.navigationController?.pushViewController(vc, animated: true)
   }
 
-  fileprivate func showEmptyState(_ emptyState: EmptyState?) {
-    guard let emptyVC = self.emptyStatesController, let state = emptyState else { return }
+  fileprivate func showEmptyState(_ emptyState: EmptyState) {
+    guard let emptyVC = self.emptyStatesController else { return }
 
-    emptyVC.setEmptyState(state)
+    emptyVC.setEmptyState(emptyState)
     emptyVC.view.isHidden = false
     self.view.bringSubview(toFront: emptyVC.view)
     UIView.animate(withDuration: 0.3,
--- a/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
@@ -240,15 +240,13 @@ internal final class DiscoveryPageViewController: UITableViewController {
     self.navigationController?.pushViewController(vc, animated: true)
   }
 
-  fileprivate func showEmptyState(_ emptyState: EmptyState) {
-    guard let emptyVC = self.emptyStatesController else { return }
+  fileprivate func showEmptyState(_ emptyState: EmptyState?) {
+    guard let emptyVC = self.emptyStatesController, let state = emptyState else { return }
 
-    emptyVC.setEmptyState(emptyState)
+    emptyVC.setEmptyState(state)
     emptyVC.view.isHidden = false
     self.view.bringSubview(toFront: emptyVC.view)
     UIView.animate(withDuration: 0.3,
-                   delay: 2,
-                   options: UIViewAnimationOptions(),
                    animations: {
       self.emptyStatesController?.view.alpha = 1.0
     }, completion: nil)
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -240,16 +240,16 @@ internal final class DiscoveryPageViewController: UITableViewController {
     self.navigationController?.pushViewController(vc, animated: true)
   }
 
-  fileprivate func showEmptyState(_ emptyState: EmptyState) {
-    guard let emptyVC = self.emptyStatesController else { return }
+  fileprivate func showEmptyState(_ emptyState: EmptyState?) {
```
justinswart: Why is `EmptyState` optional now?

Library/UIPageViewController+Extensions.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+extension UIPageViewController {
+
+  public func isScrollEnabled(_ enabled: Bool) {
```
justinswart: Is this to disable the bounce on the page view controller? It's annoying that there isn't just a simple flag to do this, and doesn't feel that nice to iterate through subviews but I understand it's not such an easy one to solve.

But an alternative that I've used in the past is to return `nil` from the `UIPageViewControllerDataSource` methods `pageViewController(_:viewControllerBefore:)` and `pageViewController(_:viewControllerAfter:)`. So if that isn't too difficult to implement in this case it might be a nicer way to do it? Up to you.

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -149,14 +149,16 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
 
     self.asyncReloadData = self.projects.take(first: 1).ignoreValues()
 
-    self.showEmptyState = paramsChanged
-      .takeWhen(paginatedProjects.filter { $0.isEmpty })
-      .map(emptyState(forParams:))
-      .skipNil()
-      .skipRepeats()
+    self.showEmptyState = Signal.combineLatest(paramsChanged, self.projectsAreLoading, paginatedProjects)
```
justinswart: `combineLatest` can be a tricky beast, just be sure it's what you want here!

Perhaps another way to do this to avoid the optional `EmptyState?` could be to use `filter`? Something like:
```
Signal.combineLatest(paramsChanged, self.projectsAreLoading, paginatedProjects)
  .filter { _, projectsAreLoading, projects in projectsAreLoading == false, projects.isEmpty else }
  .map { ... }
```

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
Scollaco: One of the problems that was causing the bug is that the `showEmptyState` signal was emitting when it shouldn't (Like for example emitting empty array of projects while the projects were not done loading). So when it emits in any situation where is doesn't match the requirement (projectsAreLoading == false AND projects.isEmpty), it returns nil. So the ViewController receives a nil value, it won't show the EmptyState view.

Scollaco: Actually your suggestion of using filter after combineLatest was a good one and EmptyState is no longer optional.

Library/ViewModels/DiscoveryPageViewModel.swift
Scollaco: Good suggestion, I don't know why I didn't think of this before!

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -149,14 +149,18 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
 
     self.asyncReloadData = self.projects.take(first: 1).ignoreValues()
 
-    self.showEmptyState = paramsChanged
-      .takeWhen(paginatedProjects.filter { $0.isEmpty })
-      .map(emptyState(forParams:))
+    self.showEmptyState = Signal.combineLatest(paramsChanged, self.projectsAreLoading, paginatedProjects)
```
justinswart: Still just wondering about the `combineLatest` here. Remember that means that `self.showEmptyState` will emit when any of these three signals emit and also not emit until all three have emitted at least once.

Perhaps another way to think about it is - what is the event that should trigger this change and based on what criteria?

In that case it would be:

```
eventThatShouldTrigger
  .withLatestFrom(criteria)
```

`criteria` could then be a `combineLatest` of what we want to base the empty state on?

PR #282, opened 2017-12-06
Third party sharing bug

Scollaco:
# What
A fix for a bug on which didn't allow users to share projects with 3rd-party social networks.

# Why
https://trello.com/c/UPQLLZdj/323-p2-sharing-still-not-working-for-some-contexts

# How
Our code were return Project name as default for 3rd-party social networks. Now the default is name + url.


Scollaco:
[P2: sharing still not working for some contexts](https://trello.com/c/UPQLLZdj/323-p2-sharing-still-not-working-for-some-contexts)

PR #286, opened 2017-12-19
Remove old design assets

dannyalright:
# What

Removing old `Design` directly, which contained .sketch assets and a `colors.json` file.

# Why

Original Sketch files are maintained elsewhere. Colors are referenced from elsewhere

# TODO ⏰

- [x] A check on the assumption that `colors.json` doesn't have any references


justinswart:
Hey @dannyalright, I didn't work too much on maintaining the colors in the past, but as I understand it our `Colors.swift` file is generated by a script which references `Colors.json`, the expectation, I guess, being that `Colors.json` is kept in sync with the Sketch file.

I am not 100% certain how `Colors.json` was updated though in the past. The last major change to it seems to be in #199 where but I'm not sure if those changes were generated from Sketch files or if the file was manually updated.

I guess it just a decision for the team going forward. Do we need the interim step of maintaining the json file from which `Colors.swift` is generated or do we just modify `Colors.swift` directly?

Note, the script that does the work is in `bin/colors`.

dannyalright:
Cheers @justinswart. Just checked out both the `.json` and `.swift`. I think keeping `Colors.json` is a smart move because it's:
- Simpler and therefore more legible
- More portable and therefore easier to keep in sync with web colours

Should I update this PR with restoring the `Colors.json`? I still feel inclined to remove Sketch as the source of truth for colours.

justinswart:
@dannyalright yeah I think let's keep the `.json` file and continue to make updates to that and generate `Colors.swift` from that.

If you're maintaining the Sketch files elsewhere then I don't see any reason to keep them in the repo unless you plan to update them there too. I think that they're kept there mostly just as reference and for exporting assets.

dannyalright:
Cool. I agree. Restored `Design/Colors.json` and kept the Sketch files killed. Ready!

PR #287, opened 2017-12-19
Header title bug fix

Scollaco:
# What
Fixed bug that didn't show category name if user goes to Discovery from Thanks page.

# See 👀
# Before
simulator screen shot - iphone 8 - 2017-12-19 at 15 12 45

# After
simulator screen shot - iphone 8 - 2017-12-19 at 15 00 30

GIF

### DIFF
--- a/Kickstarter-iOS/Views/Cells/DiscoveryExpandedSelectableRow.swift
+++ b/Kickstarter-iOS/Views/Cells/DiscoveryExpandedSelectableRow.swift
@@ -12,7 +12,7 @@ internal final class DiscoveryExpandedSelectableRowCell: UITableViewCell, ValueC
 
   internal func configureWith(value: (row: SelectableRow, categoryId: Int?)) {
     if let category = value.row.params.category, category.isRoot {
-      self.filterTitleLabel.text = RootCategory(categoryId: category.intID ?? -1).allProjectsString()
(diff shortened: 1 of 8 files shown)


justinswart:
Hey @Scollaco looks cool, do you mind explaining what caused the bug?

Scollaco:
Hey @justinswart , the root of the problem was the same as the one preventing recommended projects to appear on the Thanks page. Because the `category_id` sent by the server (fetching discovery projects) differs from the one used on graphQL requests, we had to work around to make the types match: 
```
private func decodeToGraphCategory(_ json: JSON?) -> Decoded<Category> { 
   guard let jsonObj = json else {
     return .success(Category(id: "-1", name: "Unknown Category"))
   }
 
    switch jsonObj {
    case .object(let dic):
      let category = Category(id: categoryInfo(dic).0,
                                               name: categoryInfo(dic).1)
                                               parentId: categoryInfo(dic).2)
      return .success(category)
    default:
      return .failure(DecodeError.custom("JSON should be object type"))
    }
  }
```

The problem now was that we were not setting the `parentId` what affected the `self.showRecommendations` signal. And also, for the header text, we were using `intID` in the logic. But because the ids, again, are different, the logic was failing here. That's why I opted to use the base64 id instead of doing some hack solution. Let me know if you have any questions.

justinswart:
@Scollaco cool I think that makes sense! Thanks 👍

PR #290, opened 2017-12-21
Project image tap event bug

Scollaco:
# What
Inside a project: tapping on Creator's name > Project cover, a webView is presented instead of taking the user back to the project.

# Why
https://trello.com/c/AMqoLZxB/227-inside-a-project-tapping-on-creators-name-project-cover-thing-launch-of-web-view

# How
Adding a logic that compares the request URL(when tapping the WKWebView) and project URL. If they are the same, we should go back to the project page.

### DIFF
--- a/Library/ViewModels/ProjectCreatorViewModel.swift
+++ b/Library/ViewModels/ProjectCreatorViewModel.swift
@@ -17,6 +17,9 @@ public protocol ProjectCreatorViewModelInputs {
 }
 
 public protocol ProjectCreatorViewModelOutputs {
+  /// Emits when we should return to project page.
+  var goBackToProject: Signal<(), NoError> { get }
+
   /// Emits when the LoginToutViewController should be presented.
   var goToLoginTout: Signal<LoginIntent, NoError> { get }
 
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ProjectCreatorViewModel.swift
```diff
@@ -66,10 +69,22 @@ ProjectCreatorViewModelOutputs {
       .filter { _ in AppEnvironment.current.currentUser != nil }
       .map { (MessageSubject.project($0), .projectPage) }
 
-    self.goToSafariBrowser = navigationAction
-      .filter { $0.navigationType == .linkActivated }
-      .filter { !isMessageCreator(request: $0.request) }
-      .map { $0.request.url }
+    self.goBackToProject = Signal.combineLatest(project, navigationAction)
+      .filter { $1.navigationType == .linkActivated }
+      .filter { (arg: (Project, WKNavigationActionData)) -> Bool in
```
justinswart: Sometimes Xcode complains about a tuple not supporting destructuring and then one has to do this but you'll probably find that once it can infer your closure's return type correctly then you can probably remove this and just do:

```
.filter { project, navigation in
  project.urls.web.project != navigation.request.url?.absoluteString
}
```

Either is fine - just mentioning for interest.

PR #295, opened 2018-01-09
Uses specific XCTest matchers over XCTAssertEqual

ornithocoder:
# What

Hi there! I'm working on a [SwiftLint](https://github.com/realm/SwiftLint) rule which enforces specific matchers over `XCTAssertEqual` and `XCTAssertNotEqual`. For instance,  `XCTAssertEqual(true, foo)` throws a violations and suggests `XCTAssertTrue` instead.

SwiftLint's build system runs against popular open source software - which includes this one - and some violations were detected.

# Why

This pull request is just a suggestion. Feel free to reject it if it's relevant for the project.

### DIFF
--- a/KsApi/models/CategoryTests.swift
+++ b/KsApi/models/CategoryTests.swift
@@ -63,7 +63,7 @@ class CategoryTests: XCTestCase {
 
   func testParent() {
     XCTAssertEqual(Category.illustration.parent, Category.art)
-    XCTAssertEqual(Category.art.parent, nil)
+    XCTAssertNil(Category.art.parent)
   }
 
   func testParentCategoryType() {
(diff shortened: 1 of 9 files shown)


justinswart:
@ornithocoder hey thanks for taking the time to do this - do you have a PR open for the rule that you are working on?

ornithocoder:
I do, it's #1995. Unfortunately there's a limitation that I don't think can be avoided. For instance, in `XCTAssertEqual(foo, true)`, `foo` can optional there there's no way to retrieve that information.

From the rule point of view, it doesn't know foo can be `nil` and suggest `XCTAssertTrue(foo)` which will cause the compiler to complain.

For this reason the rule is `Opt-In`.

justinswart:
Ah right, that's an unfortunate limitation. I guess one could also do something like `XCTAssertTrue(foo == .some(true))` but you'd still need to be able to determine whether `foo` was optional to begin with 🤔.

I'll keep an eye to see if your rule is merged and then we can look at merging this PR 👍

justinswart:
Closing this for now as the rule hasn't merged into SwiftLint, but please reopen if it does 👍

PR #298, opened 2018-01-16
Ab test creator name discovery

Scollaco:
# What
A/B test to show (or not) the creator's name on Discovery cell.

# Why
This test will give us a feedback of how the infrastructure is working, so we can add more tests in the future.


Library/ViewModels/ProjectPamphletMainCellViewModel.swift
```diff
@@ -105,6 +103,9 @@ public protocol ProjectPamphletMainCellViewModelOutputs {
   /// Emits a boolean that determines if the project state label should be hidden.
   var stateLabelHidden: Signal<Bool, NoError> { get }
 
+  /// Emits a string to use for the stats stack view accessibility value.
+  var statsStackViewAccessibilityLabel: Signal<String, NoError> { get }
```
cdolm92: 👍, thanks for alphabetizing!

KsApi/models/Config.swift
```diff
@@ -2,6 +2,18 @@ import Argo
 import Curry
 import Runes
 
+public enum Experiment {
+
+  public enum Name: String {
+    case iosTest = "ios_test_test"
```
justinswart: Is `ios_test_test` the name being returned from production at this stage?

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -251,6 +258,7 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
 
   internal func configureWith(value: Project) {
     self.viewModel.inputs.configureWith(project: value)
+    self.viewModel.inputs.current(configuration: AppEnvironment.current.config)
```
justinswart: We don't need this input, `AppEnvironment` is globally accessible so we can just access it where needed in the view model.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -176,6 +180,16 @@ DiscoveryViewModelOutputs {
     self.viewWillAppearProperty.value = animated
   }
 
+  private static func toString(_ dictionary: [String: String]?) -> String {
```
justinswart: `abExperiments` is optional on `Config` so we don't need to return the default empty string. Also - wouldn't it be better to make `Experiment` conform to `Codable` and then we could just use  `JSONDecoder`'s `KeyedDecodingContainer` to encode/decode this? Maybe I'm missing something.

KsApi/models/Config.swift
Scollaco: Yes, but I think it will be removed as soon as we implement real tests.

PR #304, opened 2018-01-31
Koala event fix

Scollaco:
# What
Change the place where we send `current_variants` property with Koala events.

# Why
Because we were sending `current_variants` with `Viewed Discovery` Koala event, the property was being renamed to `discover_current_variants`, what doesn't exist on the backend.

### DIFF
--- a/Library/Koala/KoalaTests.swift
+++ b/Library/Koala/KoalaTests.swift
@@ -356,7 +356,7 @@ final class KoalaTests: TestCase {
     XCTAssertNil(properties["discover_staff_picks"])
     XCTAssertNil(properties["discover_starred"])
     XCTAssertNil(properties["discover_term"])
-    XCTAssertEqual(false, properties["discover_everything"] as? Bool)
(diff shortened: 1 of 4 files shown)


Library/Koala/KoalaTests.swift
```diff
@@ -356,7 +356,7 @@ final class KoalaTests: TestCase {
     XCTAssertNil(properties["discover_staff_picks"])
     XCTAssertNil(properties["discover_starred"])
     XCTAssertNil(properties["discover_term"])
-    XCTAssertEqual(false, properties["discover_everything"] as? Bool)
+    XCTAssertEqual(true, properties["discover_everything"] as? Bool)
```
justinswart: This test seems unrelated, just curious why the boolean flipped?

Scollaco: The value `true`  means that the array of "discovery properties"  is empty in that context. The value was `false` because we adding the `current_variants` property  to the array, but now that we moved the property to another place, the array is empty again (in this test).

PR #306, opened 2018-02-02
Links to project update fix

cdolm92:
# What

Before deep linking into project campaign faqs and update would only take user to the homescreen. Links work properly now.

# See 👀

UPDATES:
deeplink_updates

FAQ:
faqdescription_faq

# TODO ⏰

For now, the faq links take the user to the campaign description and from there the user can see the faqs by tapping on the faq button. This is because we do not have a native view for faq.


cdolm92:
@Scollaco @justinswart g2g? 👀

PR #309, opened 2018-02-14
Swift4 project update

Scollaco:
# What

Updates our project and internal dependencies (Prelude, ReactiveExtensions) to swift 4  and ReactiveSwift to the latest version.

# TODO ⏰
- [x] Fix unit tests


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -184,15 +184,15 @@ AppDelegateViewModelOutputs {
 
   // swiftlint:disable cyclomatic_complexity
   public init() {
-    let currentUserEvent = Signal
+    let currentUserEvent: Signal<Signal<User?, ErrorEnvelope>.Event, NoError> = Signal
```
justinswart: Might not need the whole type signature here but it's cool if it helps the compiler along, it should be able to infer it from the type in the `switchMap` I think.

Kickstarter-iOS/Views/Controllers/LiveVideoViewController.swift
```diff
@@ -147,7 +147,7 @@ public final class LiveVideoViewController: UIViewController {
     let player = AVPlayer(url: url)
     let controller = AVPlayerViewController()
     controller.player = player
-    controller.videoGravity = AVLayerVideoGravityResizeAspectFill
+    controller.videoGravity = AVLayerVideoGravity.resizeAspectFill.rawValue
```
justinswart: Why the `rawValue` here and not the enum type directly?

Library/TestHelpers/TestCase.swift
```diff
@@ -48,7 +49,7 @@ internal class TestCase: FBSnapshotTestCase {
       debounceInterval: .seconds(0),
       device: MockDevice(),
       facebookAppDelegate: self.facebookAppDelegate,
-      isVoiceOverRunning: const(false),
```
justinswart: What did the compiler not like about this?

Library/UpdateActivityItemProvider.swift
```diff
@@ -13,25 +13,25 @@ public final class UpdateActivityItemProvider: UIActivityItemProvider {
...
-                                              itemForActivityType activityType: UIActivityType) -> Any? {
+                                              itemForActivityType activityType: UIActivityType?) -> Any? {
 
     guard let update = self.update else {
       return self.activityViewControllerPlaceholderItem(activityViewController)
     }
 
     switch activityType {
-    case UIActivityType.mail:
+    case UIActivityType.mail?:
```
justinswart: Maybe another `guard` to be sure the `activityType` is not `nil` else return the default string?

Library/ViewModels/UpdateDraftViewModel.swift
```diff
@@ -162,15 +165,17 @@ UpdateDraftViewModelOutputs {
     self.title = draft.map { $0.update.title }
     self.body = draft.map { $0.update.body ?? "" }
 
-    let wasBackersOnly = draft.map { $0.update.isPublic }.map(negate)
+    let wasBackersOnly: Signal<Bool, NoError> = draft.map { (draft: UpdateDraft) -> Bool in
+      draft.update.isPublic }.map(negate
```
justinswart: Parenthesis and brace here would look better on the next line 👍

Library/ViewModels/UpdateDraftViewModel.swift
```diff
@@ -187,17 +192,21 @@ UpdateDraftViewModelOutputs {
 
     let addAttachmentEvent = draft
       .takePairWhen(self.imagePickedProperty.signal.skipNil().map(first))
-      .switchMap { draft, url in
-        AppEnvironment.current.apiService.addImage(file: url, toDraft: draft)
+      .switchMap { (arg: (UpdateDraft, URL))
+        -> SignalProducer<Signal<UpdateDraft.Image, ErrorEnvelope>.Event, NoError> in
```
justinswart: Usually once you have the compiler happy with the return type then you shouldn't need to deconstruct the tuple.

justinswart:
There are a couple of places where we could probably get away with more type inference but I don't think it's such a big deal.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
Scollaco: Yeah, you are right. Since the compiler was complaining so much and not recognising types, I decided to be more conservative and explicitly define the type signature instead of relying on the type inference. But it worked fine after I removed the signature of line 187.

Kickstarter-iOS/Views/Controllers/LiveVideoViewController.swift
Scollaco: Because `videoGravity` is a String, while `resizeAspectFill` is a AVLayerVideoGravity (that is a typedef for NSString). So we have to use rawValue here to match types and avoid the `Cannot assign value of type 'AVLayerVideoGravity' to type 'String'` error.

Library/TestHelpers/TestCase.swift
Scollaco: If I use `const(false)`, the tests will fail with the message `Extraneous argument label 'isVoiceOverRunning:' in call`.

Library/UpdateActivityItemProvider.swift
Scollaco: Good suggestion!

Library/ViewModels/UpdateDraftViewModel.swift
Scollaco: 👍

PR #315, opened 2018-03-02
Discovery header refresh

Scollaco:
# What
Updated Discovery top navigation bar treatment.

# Why
To make the access to filters more intuitive.
https://trello.com/c/RBcW27zv/469-update-top-nav-treatment

# See 👀
# Before 
simulator screen shot - iphone 8 - 2018-03-02 at 14 28 58

# After
simulator screen shot - iphone 8 - 2018-03-02 at 14 25 33

GIF


dannyalright:
looking pretty schmick! Waiting for the internet to work to properly pull and test this out before ✔️

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -17,13 +17,19 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
...
   @IBOutlet fileprivate weak var favoriteContainerView: UIView!
   @IBOutlet fileprivate weak var primaryLabel: UILabel!
   @IBOutlet fileprivate weak var secondaryLabel: UILabel!
   @IBOutlet fileprivate weak var titleButton: UIButton!
-  @IBOutlet fileprivate weak var titleStackView: UIStackView!
+  @IBOutlet fileprivate weak var containerStackView: UIStackView!
   @IBOutlet fileprivate weak var outerStackViewTopConstraint: NSLayoutConstraint!
+  @IBOutlet fileprivate weak var titleStackView: UIStackView! {
+    didSet {
+      titleStackView.pinBackground()
```
justinswart: 🤔 interesting!

I like the approach of using `didSet` to apply one-time configurations on init. It's kind've like using `lazy var` for `@IBOutlet`s

Scollaco: Yes, I would use `lazy var` for a variable that won't necessarily be allocated. Why do you like the `lazy var` approach for IBOutlets?

justinswart: No I'm saying I like overriding the setter and `didSet` because it keeps configuration of the UI element in one place and at the point at which it's defined.

Scollaco: Ahh I understand. Yes, I like this approach to, so we don't have configuration code floating around

PR #317, opened 2018-03-07
Update FBSDK from 4.24.0 to 4.31.1

justinswart:
# What

Update our FBSDK to 4.31.1.

# Why

This version has important fixes for iOS 11, Facebook claims they have seen a "14% better log in success rate on iOS 11 for apps using our latest SDK upgrade".

# How

Download SDK, overwrite files, removed some redundant tvOS-specific files.

# See 👀

https://developers.facebook.com/docs/ios/change-log-4x/


justinswart:
@Scollaco no but I would recommend that we each test fb login before merging, worked for me

PR #318, opened 2018-03-07
Discovery header updates

Scollaco:
# What
- Make small changes such as bring UI elements of discovery header 8pt. down, changed logic to show `Explore` label only if `All projects` is the selected filter.

# Why
https://trello.com/c/0YAOMbBV/496-visual-updates-to-discover-header-refresh

# See 👀
simulator screen shot - iphone 8 - 2018-03-07 at 17 39 58


dannyalright:
bonito!

PR #320, opened 2018-03-15
Implemented Default to 'Recommended For You' Experiment 

cdolm92:
# What

Implemented a/b experiment to show recommended projects in Discovery.

# How

In the `DiscoveryViewModel` I added a function called `initialParam()` which presents projects recommended to the user if the user is in the experimental bucket based on the configuration. 

# To Do

- [ ]  Check snapshot tests 😕


Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -69,9 +69,23 @@ public protocol DiscoveryViewModelType {
   var outputs: DiscoveryViewModelOutputs { get }
 }
 
+private func initialParam() -> DiscoveryParams {
+  let recExperiment = AppEnvironment.current.config?.abExperiments.filter {
+    $0.key == Experiment.Name.defaultToRecs.rawValue
+  }
+
+  if recExperiment?.first?.value == Experiment.Variant.experimental.rawValue {
```
Scollaco: Here we will also need to check if the currentUser is nil, because if he/she logs out, we will have to show `All Projects` instead. The if statements will be something like: if AppEnvironment.current.currentUser != nil && recExperiment?.first?.value == Experiment.Variant.experimental.rawValue { ...

cdolm92:
@nino g2g?

PR #326, opened 2018-04-04
sign up name copy

eoji:
# What
Updating sign up name copy string and generated new snapshots.

# See 👀
 

# cc
@miki-nakata


eoji:
[iOS app asks for user's full name on sign up page](https://trello.com/c/RwpnwGTe/540-ios-app-asks-for-users-full-name-on-sign-up-page)

PR #330, opened 2018-04-11
User agent fix

Scollaco:
# What
Fixes an issue where user-agent header was being overwritten with a default value,  making some `Completed Checkout` events be shown  as `client_type = web`.

# Why
https://trello.com/c/pl9yHIQ0/543-native-clienttype-detection

### DIFF
--- a/KsApi/ServiceType.swift
+++ b/KsApi/ServiceType.swift
@@ -407,6 +407,12 @@ extension ServiceType {
     headers["Authorization"] = self.authorizationHeader
     headers["Kickstarter-App-Id"] = self.appId
     headers["Kickstarter-iOS-App"] = self.buildVersion
+    headers["User-Agent"] = Self.userAgent
+
+    return headers
+  }
+
+  public static var userAgent: String {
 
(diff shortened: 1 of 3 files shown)


Scollaco:
[Native client_type detection](https://trello.com/c/pl9yHIQ0/543-native-clienttype-detection)

KsApi/ServiceType.swift
```diff
@@ -407,6 +407,12 @@ extension ServiceType {
     headers["Authorization"] = self.authorizationHeader
     headers["Kickstarter-App-Id"] = self.appId
     headers["Kickstarter-iOS-App"] = self.buildVersion
+    headers["User-Agent"] = Self.userAgent
```
eoji: what's the difference between `self` and `Self`? 🕵🏾‍♀️

Scollaco: `Self` (capitalized) refers to the type that conforms to the protocol. (In this case, `ServiceType`). While `self` is equivalent to `this` in Java.

eoji: amazing

PR #331, opened 2018-04-17
Swift 4.1 updates, CircleCI 2.0 migration

justinswart:
# What

This PR updates our code to Swift 4.1 (essentially just renaming a bunch of `flatMap`s to `compactMap`s now that they are deprecated in 4.1. This however opened up a can of worms on CircleCI as Xcode 9.3 appears to only be supported by CircleCI 2.0. Considering that CircleCI will sunset 1.0 later this year it makes sense that we get this migration over with.

# How

Migrating to CircleCI 2.0 is detailed [here](https://circleci.com/docs/2.0/ios-migrating-from-1-2/).

Basically it required rewriting our `circle.yml` file to work with Circle's new concepts of workflows and jobs as well as moving our codesigning to use `fastlane match`. This also required re-recording a ton of screenshots for subtle changes in Xcode 9.3.

We will all need to sync up using `fastlance match` once this is merged, which I'll chat about on Slack.

I'll put some more comments in the code on this PR.


Makefile
```diff
@@ -4,13 +4,14 @@ BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION)
...
-OS ?= 11.2
 RELEASE ?= beta
+IOS_VERSION ?= 11.3
+IPHONE_NAME ?= iPhone 8
 BRANCH ?= master
 DIST_BRANCH = $(RELEASE)-dist
 
 ifeq ($(PLATFORM),iOS)
-	DESTINATION ?= 'platform=iOS Simulator,name=iPhone 8,OS=$(OS)'
+	DESTINATION ?= 'platform=iOS Simulator,name=$(IPHONE_NAME),OS=$(IOS_VERSION)'
```
justinswart: These values are now environment variables that are set in `circle.yml`

Makefile
```diff
@@ -113,7 +114,6 @@ secrets:
 		|| true; \
 	fi
 
-OPENTOK_VERSION = 2.10.2
```
justinswart: Caching the Opentok version is handled by `Opentok.lock` now.

.fastlane/Fastfile
```diff
@@ -4,14 +4,74 @@ default_platform :ios
 
 platform :ios do
   before_all do
-    xcversion(version: "9.2")
+    setup_circle_ci
```
justinswart: Turns out this line is important: https://github.com/circleci/circleci-docs/pull/2258

.fastlane/Fastfile
```diff
@@ -4,14 +4,74 @@ default_platform :ios
...
+    xcversion(version: ENV["XCODE_VERSION"])
+  end
+
+  ### MATCH
+
+  desc "Fastlane Match"
+  lane :match_all do
```
justinswart: This syncs up our certificates and provisioning profiles for Drip and Kickstarter.

circle.yml
```diff
@@ -1,56 +1,384 @@
...
+
+      - store_artifacts:
+          path: output
+
+# Workflows
+workflows:
+  version: 2
+  build:
+    jobs:
+      - swiftlint-and-cache
```
justinswart: Tests are run in parallel now.

circle.yml
```diff
@@ -1,56 +1,384 @@
...
+          command: bundle check || bundle install
+          environment:
+            BUNDLE_JOBS: 4
+            BUNDLE_RETRY: 3
+      - save_cache:
+          key: v1-gems-{{ checksum "Gemfile.lock" }}
+          paths:
+            - vendor/bundle
+
+      - attach_workspace:
```
justinswart: I still need to confirm that this actually works 🤔

PR #332, opened 2018-04-20
Show pie chart

cdolm92:
# What

Pie chart is back and the info is correct!

# Why

The pie chart was showing inaccurate information when there were 30 or more referrers.

# How

We use  values from`ReferralAggregateStats`. 

# See 👀

Snapshots below.


Kickstarter-iOS/Info.plist
```diff
@@ -62,6 +62,8 @@
 	<true/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
+		<key>NSAllowsArbitraryLoads</key>
```
ifbarrera: Should this be added?

KsApi/models/ProjectStatsEnvelopeTests.swift
```diff
@@ -23,6 +23,11 @@ final class ProjectStatsEnvelopeTests: XCTestCase {
       ["date": 555444335]
     ]
     let json: [String: Any] = [
+      "referral_aggregates": [
+        "custom": 1.0,
+        "external": "15.0",
```
ifbarrera: Just curious, how come `external` is the only string?

Kickstarter-iOS/Info.plist
justinswart: Woops, no we don't wanna add that please 🙂

Kickstarter-iOS/Views/Cells/DashboardReferrersCell.swift
```diff
@@ -42,6 +42,8 @@ internal final class DashboardReferrersCell: UITableViewCell, ValueCell {
   internal override func awakeFromNib() {
     super.awakeFromNib()
 
+    self.viewModel.inputs.awakeFromNib()
```
justinswart: I think it's best to keep this at the bottom of the function so that we know it's full executed before calling the viewModel, unless it's necessary for functionality.

KsApi/models/ProjectStatsEnvelopeTests.swift
cdolm92: we get a string value from the server. there is a helper function in the `ProjectStatsEnvelope` called `stringToDouble` that we use to convert it into a `Double`

Kickstarter-iOS/Info.plist
cdolm92: Thanks!

KsApi/models/ProjectStatsEnvelopeTests.swift
ifbarrera: Yeah I just meant its strange that the API returns it in this way. Ideally they would be consistent, but that's not part of this ticket 👍

cdolm92:
👍🏾?

PR #334, opened 2018-04-21
Updated to methods changes

amadeu01:
# What

Update of the playground to the newer `UIControlProtocol` API

# Why

The playground was throwing errors.

Some method was changed, however, the playground was not updated.
For instance: 
- ` title(_ for: _)`  -> `title(_ forState: _)`


amadeu01:
Thanks for accepting it !

Issue #75, opened 2017-02-04
make bootstrap error


StoneCX:
Xcode is 8.2.1 and i'm clone it  execute "make bootstrap"then always error and i don't know why
`curl -s -N -L https://tokbox.com/downloads/opentok-ios-sdk-2.10.0 \
		| tar -xz --strip 1 --directory Frameworks/OpenTok OpenTok-iOS-2.10.0/OpenTok.framework
tar: Error opening archive: (Empty error message)
make: *** [opentok] Error 1`

justinswart:
Hi @StoneCX, I'm not sure why that's happening, perhaps it's failing to download the OpenTok library's tar ball. We use this library for live streaming.

You can work around this by doing the following:
1. Comment out the `opentok` recipe steps in `Makefile` (lines 108 - 110).
1. Download OpenTok manually [here](https://tokbox.com/downloads/opentok-ios-sdk-2.10.0).
1. Extract `OpenTok.framework` from the archive to `{PROJECT_DIR}/Frameworks/OpenTok/`.
1. Run `make bootstrap` again.

StoneCX:
@justinswart thanks that's work

waseefakhtar:
Sorry I don’t know which lines are the `opentok` recipe steps. I comment out 108 - 110 like:

```
VERSION_FILE = Frameworks/OpenTok/version
CURRENT_OPENTOK_VERSION = $(shell cat $(VERSION_FILE))
ifeq ($(CURRENT_OPENTOK_VERSION),)
```

And I get `Makefile:112: *** extraneous `endif'.  Stop.`. When I include endif with the commenting out (108 - 112) like:

```
VERSION_FILE = Frameworks/OpenTok/version
CURRENT_OPENTOK_VERSION = $(shell cat $(VERSION_FILE))
ifeq ($(CURRENT_OPENTOK_VERSION),)
	CURRENT_OPENTOK_VERSION = first
endif
```

I get `make: *** No rule to make target `bootsrap'.  Stop.`

Am I commenting out the wrong lines?

justinswart:
@waseefakhtar are you sure you're typing `make bootstrap`? the error you've pasted has the incorrect spelling.

PR #335, opened 2018-04-23
Setting `OPENTOK_VERSION` in `Makefile`

priteshshah1983:
Fixes https://github.com/kickstarter/ios-oss/issues/75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ IOS_VERSION ?= 11.3
 IPHONE_NAME ?= iPhone 8
 BRANCH ?= master
 DIST_BRANCH = $(RELEASE)-dist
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ IOS_VERSION ?= 11.3
 IPHONE_NAME ?= iPhone 8
 BRANCH ?= master
 DIST_BRANCH = $(RELEASE)-dist
+OPENTOK_VERSION = 2.10.2
```
justinswart: Could we change this to `OPENTOK_VERSION ?= 2.10.2` That way the value that we pass through as an environment variable will be used if it's set.

priteshshah1983:
Thanks for the feedback! I updated the PR

PR #342, opened 2018-04-26
Deep link crash fix

justinswart:
# What

Implement `UIApplicationDelegate`'s newer method for opening URL resources.

# Why

We've noticed a significant uptick in crashes on our last build. It appears that deep-links from Safari (FB logins or links from our known associations file) are calling [this](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623112-application) newer delegate method which we aren't yet implementing.

# How

Implement the delegate method.

# See

https://stackoverflow.com/questions/39069361/exc-bad-access-in-facebook-sdk-login-xcode-8-beta/39260549#39260549

giphy-downsized-large


Scollaco:
[Deep link crash fix](https://trello.com/c/PN55qCs4/595-deep-link-crash-fix)

PR #343, opened 2018-04-30
Cross Dissolve Animation spike

cdolm92:
# What

Exploring cross-dissolve between tabs.

# See 👀
crossdissolve

# TODO ⏰

- [ ] Approval to merge?


Kickstarter-iOS/Views/Transitions/ScrollingTransitionAnimator.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
+
+internal final class ScrollingTransitionAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+  weak var transitionContext: UIViewControllerContextTransitioning?
+  var tabBarController: UITabBarController!
```
ifbarrera: Is there any reason you need to hold the references to `tabBarController`, `lastIndex` and `transitionContext`?

Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -211,6 +212,13 @@ extension RootTabBarViewController: UITabBarControllerDelegate {
                                didSelect viewController: UIViewController) {
     self.viewModel.inputs.didSelectIndex(tabBarController.selectedIndex)
   }
+
+  public func tabBarController(_ tabBarController: UITabBarController,
+                               animationControllerForTransitionFrom fromVC: UIViewController,
+                               to toVC: UIViewController) -> UIViewControllerAnimatedTransitioning? {
+
+    return self.transitionCrossDissolve
```
justinswart: I _think_ seeing as there's not real state held in the animator you could just do `return ScrollingTransitionAnimator()` here. Although I'm also not sure if it makes more sense to create it once and just refer to it seeing as it will be used for every tab switch. I think animators are pretty inexpensive to create.

PR #344, opened 2018-05-02
Contextual permission dialog

Scollaco:
# What
Adding context-aware primer dialogs before serving up the default iOS notification allowance dialog.

# Why
GIF

But if you don't:
see: https://trello.com/c/uTpAqTCP/589-notification-primer-dialog

# TODO ⏰

- [x] Fix unit tests
- [x] Will add more tests
- [x] Need final copies


Kickstarter-iOS/Views/Controllers/MessageDialogViewController.swift
```diff
@@ -93,6 +94,17 @@ internal final class MessageDialogViewController: UIViewController {
                                animated: true,
                                completion: nil)
   }
+
+  private func postNotificationIfNeeded() {
+    guard shouldSendNotification() else { return }
```
ifbarrera: This is already checked in the `AppDelegateViewModel`'s `showAlert` signal right? Any reason to check it again?

Library/PushNotificationDialog.swift
```diff
@@ -0,0 +1,59 @@
...
+    case login
+    case message
+    case pledge
+    case save
+  }
+
+  static public func canShowDialog(`for` context: PushNotificationDialog.Context) -> Bool {
+
+    return (!AppEnvironment.current.userDefaults.deniedNotificationContexts.contains(context.rawValue) &&
+            AppEnvironment.current.userDefaults.deniedNotificationContexts.count < 3) &&
```
ifbarrera: Why do you need to check the count is < 3?

Library/PushNotificationdialogTests.swift
```diff
@@ -0,0 +1,85 @@
+import XCTest
+@testable import Library
+@testable import KsApi
+
+class PushNotificationdialogTests: XCTestCase {
```
ifbarrera: Tiny comment: should be `PushNotificationDialogTests`?

Scollaco: Absolutely! :)

Library/PushNotificationDialog.swift
Scollaco: This is one of the rules to show the PushNotification dialog. We can  show the alerts only 3x (In case the user tapped `Not now` on the 2 first times).

PR #352, opened 2018-05-10
hiding recommendations for opted out users

eoji:
image

# What
No longer showing recommendations as a filter for users who are opted out.

# Why
G
D
P
R

# How
Checking if `User.opted_out_of_recommendations` is `true`/`false`/`nil` and updating the Discovery filters. This value is optional and `false` by default.

# See 👀
Wax on, wax off

### DIFF
--- a/KsApi/models/User.swift
+++ b/KsApi/models/User.swift
@@ -5,7 +5,6 @@ import Runes
 public struct User {
   public let avatar: Avatar
   public let facebookConnected: Bool?
-  public let generateRecommendations: Bool?
   public let id: Int
   public let isFriend: Bool?
   public let liveAuthToken: String?
@@ -13,6 +12,7 @@ public struct User {
   public let name: String
   public let newsletters: NewsletterSubscriptions
   public let notifications: Notifications
+  public let optedOutOfRecommendations: Bool?
   public let social: Bool?
   public let stats: Stats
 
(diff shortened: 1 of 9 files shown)


KsApi/models/User.swift
```diff
@@ -12,6 +12,7 @@ public struct User {
   public let name: String
   public let newsletters: NewsletterSubscriptions
   public let notifications: Notifications
+  public let optedOutOfRecommendations: Bool?
```
ifbarrera: It might be better to have this be non-optional with a default value of `false`. That way you can always assume the value is non-nil.

KsApi/models/User.swift
```diff
@@ -86,6 +87,7 @@ extension User: Argo.Decodable {
       <*> json <| "name"
       <*> User.NewsletterSubscriptions.decode(json)
       <*> User.Notifications.decode(json)
+      <*> json <|? "opted_out_of_recommendations"
```
ifbarrera: You can do:
`<*> json <| "opted_out_of_recommendations" <|> .failure(false)` to set `false` as a default value if parsing the key fails. Credit to @Scollaco for helping me figure out if Argo could do this 😄

Scollaco: Just one thing... it should be `.success(false)`, since we want the decoder to consider it a successfully decoded property.

PR #357, opened 2018-05-18
Updating "no saved projects" state

ifbarrera:
# What

Updating the assets, tweaking the font size, updating copy. Also updating screenshots for newly translated disclaimer copy for signup.

# Why

Improving UI, GDPR.

# See 👀
simulator screen shot - iphone 8 - 2018-05-18 at 12 31 21
simulator screen shot - iphone 8 - 2018-05-17 at 12 02 43


Kickstarter-iOS/Views/Controllers/DashboardViewControllerTests.swift
```diff
@@ -47,7 +47,7 @@ internal final class DashboardViewControllerTests: TestCase {
 
         self.scheduler.run()
 
-        FBSnapshotVerifyView(parent.view, identifier: "lang_\(language)_device_\(device)")
+        FBSnapshotVerifyView(parent.view, identifier: "lang_\(language)_device_\(device)", tolerance: 0.05)
```
Scollaco: I wonder if the tolerance is too high here considering DashboardViewController is a view controller with such complex UI. As an example, we set the tolerance on RewardPledgeViewControllerTests to 0.0001. I would be happy to pair on this to see if we can find or not a better solution.

PR #360, opened 2018-05-22
Gdpr following ui

Scollaco:
## Note
 (This PR branches out from one of  @cdolm92's  branch. So to review it, you just need to review the commits assigned to my name).

# What
Following opt out. Gives the user the ability to stop following other users activities. 

# Why
GDPR.

# See 👀

# TODO ⏰
- [ ] Final Strings

GIF


dannyalright:
does this include the fix on the string-break in the alert? The screenshot might be old. Just checking

Scollaco:
@dannyalright What you saw yesterday was happening because the string was hard coded. Once we get the strings from the server, we shouldn't have problems. The prompt will adjust the text dynamically.

KsApi/lib/Route.swift
```diff
@@ -354,7 +354,6 @@ internal enum Route {
 
     case let .user(userId):
       return (.GET, "/v1/users/\(userId)", [:], nil)
-
```
eoji: looks good

PR #361, opened 2018-05-23
Alpha build setup

ifbarrera:
# What

Adds necessary jobs/steps to allow for automatic KickAlpha build uploads.

# Why

Preparation steps for automatic alpha builds.

# How

Merges into master will trigger the creation of a new branch `alpha-dist-<COMMIT>` on the private remote. Any builds to `alpha-dist-.*` branches will run a job that runs the same steps as the `beta` job, but calls `fastlane alpha_match_gym_hockey` instead of `fastlane beta_match_gym_hockey`. The `alpha_match_gym_hockey` lane will build the KickAlpha app and upload it to Hockey.


ifbarrera:
[Automatic Alpha Build](https://trello.com/c/AQeq2Bvk/686-automatic-alpha-build)

circle.yml
```diff
@@ -184,7 +184,71 @@ jobs:
       - store_artifacts:
           path: /tmp/xcode_raw.log
 
-  # Hockey
+  master:
```
justinswart: What's the `master` job for?

Makefile
```diff
@@ -92,12 +92,21 @@ deploy:
...
 	@echo "Deploying private/alpha-dist..."
 
-	@git branch -f alpha-dist private/alpha-dist
+	@git branch -f alpha-dist private/$(BRANCH)
 	@git push -f private alpha-dist
 	@git branch -d alpha-dist
 
 	@echo "Deploy has been kicked off to CircleCI!"
 
+sync:
```
justinswart: Also possibly not needed?

circle.yml
```diff
@@ -292,6 +356,26 @@ workflows:
       - library-tests
       - livestream-tests
       - ksapi-tests
+      # - master:
```
justinswart: Guessing we won't need this if `master`'s not needed?

circle.yml
justinswart: Oh I see now - your intention is that every time we push to `master` it should also release an alpha build 🤔

Ok, we can try it this way - there may be another way to inject the environment variable though if that doesn't work.

Makefile
justinswart: Answered my question second time around 😊

circle.yml
justinswart: Answered my question second time around 😊

justinswart:
@ifbarrera this seems good to me, received an alpha build. Guess all that's left is to rename the branch back to `master`?

PR #368, opened 2018-05-30
Password reveal functionality

ifbarrera:
# What

Show/hide password on login. 👁 

# See 👀

simulator screen shot - iphone 8 - 2018-05-30 at 15 08 34
simulator screen shot - iphone 8 - 2018-05-30 at 15 08 40


ifbarrera:
[Native: Password reveal option to make entering passwords easier](https://trello.com/c/YC53CLk7/630-native-password-reveal-option-to-make-entering-passwords-easier)

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -91,6 +106,26 @@ internal final class LoginViewController: UIViewController {
...
     self.onePasswordButton.rac.hidden = self.viewModel.outputs.onePasswordButtonHidden
 
+    self.viewModel.outputs.showHidePasswordButtonToggled
+    .observeForUI()
+    .observeValues { [weak self] shouldShow in
+      guard let sself = self else { return }
+      let tintColor: UIColor = shouldShow ? .ksr_green_500 : .ksr_grey_400
+      let accessibilityValue = shouldShow ? Strings.Hide_password() : Strings.Show_password()
+
+      _ = sself.showHidePasswordButton
```
Scollaco: `self` here has a double `s`

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -91,6 +106,26 @@ internal final class LoginViewController: UIViewController {
...
+        |> UIButton.lens.tintColor .~ tintColor
+        |> UIButton.lens.accessibilityValue .~ accessibilityValue
+
+      // Note: workaround for cursor whitespace render bug
+      let currentText = sself.passwordTextField ^* UITextField.lens.text
+
+      _ = sself.passwordTextField
```
Scollaco: And here too

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -91,6 +106,26 @@ internal final class LoginViewController: UIViewController {
     self.passwordTextField.rac.text = self.viewModel.outputs.passwordText
     self.onePasswordButton.rac.hidden = self.viewModel.outputs.onePasswordButtonHidden
 
+    self.viewModel.outputs.showHidePasswordButtonToggled
+    .observeForUI()
+    .observeValues { [weak self] shouldShow in
+      guard let sself = self else { return }
```
Scollaco: Here I would rename `sself` to `_self` since that's the name we usually use to check if `self` is not nil.

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -91,6 +106,26 @@ internal final class LoginViewController: UIViewController {
...
+        |> UIButton.lens.tintColor .~ tintColor
+        |> UIButton.lens.accessibilityValue .~ accessibilityValue
+
+      // Note: workaround for cursor whitespace render bug
+      let currentText = sself.passwordTextField ^* UITextField.lens.text
+
+      _ = sself.passwordTextField
+        |> UITextField.lens.secureTextEntry .~ !shouldShow
+        |> UITextField.lens.text .~ " "
+        |> UITextField.lens.text .~ currentText
```
Scollaco: Is there a reason why we are setting the passwordTextField text first to an empty string and the to `currentText`?

ifbarrera: Yeah, I put a note on line 20. This is a workaround for a bug where if you just update `secureTextEntry`, the cursor on the text field does not visually update according to the size of the content in the text field. Setting it to empty, then resetting it to the previous value forces a recalculation of the cursor's position and fixes the issue. Before this I also tried manually triggering `layoutIfNeeded()`, `setNeedsDisplay()` - neither of these worked.

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
justinswart: For readability I'd move all of this to a private function in the file too 👍

Scollaco:
GIF

PR #369, opened 2018-05-30
Updated max pledge value per country

Scollaco:
# What
Updated the max pledge value allowed for launched countries.

# Why
The values on our codebase are outdated and no longer align with the values on the Rails app. This causes a bug where if a user wants to back a project on which the pledge value is higher than the max, the app will throw an error and the pledge won't be completed.

For more context: 
https://trello.com/c/KLOYqmT6/4505-zd-unable-to-make-a-pledge-on-highest-tier-on-mobile-app

GIF


justinswart:
@Scollaco looks cool and I'm sure if @frewsxcv says it's good then it's good 😄 - but where did you reference the amounts?

frewsxcv:
Here's where it's stored for the Rails app:

- https://github.com/kickstarter/kickstarter/blob/master/config/countries.yml

Up to y'all, but might be good to add a link to the file in a comment in this native file

PR #370, opened 2018-06-01
Environment switcher improvement

Scollaco:
# What
Improved Environment Switcher.

# Why
The environment switcher tool had a limitation on which it was only available for logged in users/testers and didn't show the Environment name on its initial state.

Now the tool is available for logged out testers as well and the environment name bug was fixed.

# See 👀


Library/AppEnvironment.swift
```diff
@@ -54,6 +54,7 @@ public struct AppEnvironment {
     replaceCurrentEnvironment(
       apiService: service
     )
+    AppEnvironment.logout()
```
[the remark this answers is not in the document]
Scollaco: Good call. This is actually a line I forgot to delete.

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -44,6 +47,10 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+    self.environmentSwitcherButton.addTarget(self,
+                                       action: #selector(environmentSwitcherTapped),
```
justinswart: Alignment here's a little 🤪

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -295,13 +306,53 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
...
+    alert.addAction(
+      UIAlertAction(title: "Local", style: .default) { [weak self] _ in
+        self?.viewModel.inputs.environmentSwitcherButtonTapped(environment: ServerConfig.local)
+      }
+    )
+
+    alert.addAction(
+      UIAlertAction(title: "Staging", style: .default) { [weak self] _ in
+        self?.viewModel.inputs.environmentSwitcherButtonTapped(environment: ServerConfig.staging)
+
```
justinswart: Some whitespace

Library/AppEnvironment.swift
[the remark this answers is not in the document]
justinswart: @Scollaco does that mean that it doesn't need to logout or will you handle logout elsewhere?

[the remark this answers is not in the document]
Scollaco: @justinswart The logout will be handled wherever we call `updateServerConfig`.

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -295,13 +312,64 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
...
     self.viewModel.inputs.titleButtonTapped()
   }
 
   @objc fileprivate func favoriteButtonTapped() {
     self.viewModel.inputs.favoriteButtonTapped()
   }
+
+  private func showEnvironmentActionSheet() {
+
+    let config = ServerConfig.environmentName(config: AppEnvironment.current.apiService.serverConfig)
```
ifbarrera: I wonder if we could simplify this a bit by changing up how we keep track of the environment types? If we added an enum:
```
public enum EnvironmentType: String {
  case production = "Production"
  case staging = "Staging"
  case local = "Local"
  
  static var allValues: [EnvironmentType] = [.production, .staging, .local]
}
```
and had it as a variable of the `ServerConfig` struct, then we could do:

```
EnvironmentType.allValues.forEach { environment in
      alert.addAction(
        UIAlertAction(title: environment.rawValue, style: .default) { [weak self] _ in
           // action handler
        }
      )
}
```
And we could access the current environment with: `AppEnvironment.current.apiService.serverConfig.environmentType.rawValue`. What do you think?

justinswart: I think this is cool and would avoid the need for the _unknown_ case [here](https://github.com/kickstarter/ios-oss/pull/370/files#diff-2cab5a0cd36c9da79573767a85181bf1R83), Also, Swift 4.2 has `allCases` [baked in](https://github.com/apple/swift-evolution/blob/master/proposals/0194-derived-collection-of-enum-cases.md) so we would be able to drop the array later on, in fact perhaps we want to name that `allCases` in anticipation of just using the built-in static var later on.

Scollaco: I'd added a property called `environmentName` previously that following a similar idea, but for having it as a property, was causing a bug on which the environment wasn't being shown before we changed the environment for the first time (what is not a big deal, since this is an internal build). One solution for this could be to re-add a property and set its default value to `.production`.

Scollaco:
[Improve environment switcher](https://trello.com/c/AqTAQZAs/753-improve-environment-switcher)

PR #372, opened 2018-06-05
Language switcher

ifbarrera:
# What

Switch languages in non-release builds for easy visual QA!

# How

Switching languages from the settings screen posts a notification that `RootTabBarViewController` listens for. Since a language change requires a redraw of basically every screen currently in memory, `RootTabBarViewController` will regenerate all view tab bar view controllers in response to the notification.

# See 👀

acxngzkv6y


Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -117,8 +118,28 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
...
+        let loginProfileViewController: UIViewController = user.isLoggedIn
+          ? profileController() : LoginToutViewController.configuredWith(loginIntent: .generic)
+
+        return [
+          DiscoveryViewController.instantiate(),
+          ActivitiesViewController.instantiate(),
+          SearchViewController.instantiate(),
+          dashboardViewController,
+          loginProfileViewController
+        ]
```
justinswart: I'm guessing we need to recreate these view controllers in order to ensure that all of the labels etc will be updated with new strings?

Wondering if we could combine it somehow with the `viewControllers` `Signal` above which takes into account standard and personalized 🤔

ifbarrera: Yeah it's a bit tricky, I tried that before but `standardViewControllers` and `personalizedViewControllers` listen to different signals (namely `viewDidLoad` and changes to session & user state). We need both to refresh, which was difficult to do without triggering two separate signals (one for `standardViewControllers` and one for `personalizedViewControllers`). This worked, but it resulted in the view controllers being refreshed twice which didn't seem like an optimal solution.

justinswart: Cool cool. Don't want to overcomplicate but wondering if we could introduce something like, I dunno, a `ContentRefreshable` (must be a better name 😄) protocol that VC's could conform to. This could have a `refresh()` function we could call and then in each VC call the respective function to perform that?

Of course there may be issues with some view models that may have unpredictable behaviour if this happens after `configData` is set. I would have to take a closer look to be sure 🤔

justinswart: Reading your comment again I think you're saying we'd need an input on those view models to specifically refresh them? Yeah might be a little more complicated.

ifbarrera: Sorry, my comment above was responding to your suggestion about "combine it somehow with the viewControllers Signal above which takes into account standard and personalized".

In terms of the approach, its more complicated than simply refreshing the view models, because every VC has various things that can happen in `viewDidLoad`, `bindStyles`, etc. I wanted to stay away from manually triggering those things in every VC, and the easiest way to do that is just to recreate it. The easiest way to to do that for *all* VCs currently in memory is to re-initialize the root view controllers in the tab bar.

justinswart: yeah i understand the refreshing isn’t so simple and agree that we should recreate the VC's, I think all i’d like to achieve is that the creation and logic around personal/standard is all in one place.

so we could just have a private function called `createStandardViewControllers` that returns the array:

```
[
 DiscoveryViewController.instantiate(),
 ActivitiesViewController.instantiate(),
 SearchViewController.instantiate()
]
```

Another func `createPersonalizedViewControllers(for user: User) -> [ViewController] {...}`:
```
// logic for personalized
```

And then in both places that we use this combine these two arrays

ifbarrera: That's a great suggestion! 👍 will do

eoji:
Great job on this!!

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -239,6 +245,24 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
   internal var outputs: RootViewModelOutputs { return self }
 }
 
+private func generateStandardViewControllers() -> [UIViewController] {
```
justinswart: 🙌

PR #373, opened 2018-06-07
Private profile functionality

ifbarrera:
# What

Make your profile private or public.

# See 👀

a6vo2zakal


Library/ViewModels/SettingsViewModel.swift
```diff
@@ -533,6 +540,12 @@ SettingsViewModelOutputs {
   public func postLikesTapped(selected: Bool) {
     self.postLikesTappedProperty.value = selected
   }
+
+  fileprivate let showPublicProfileProperty = MutableProperty(false)
+  public func privateProfileSwitchDidChange(isOn: Bool) {
+    self.showPublicProfileProperty.value = !isOn
```
justinswart: I think we should only perform transformation on the boolean in the output signal. These inputs are meant to have no implementation detail and that way we also know where to look for any transformation that we may be performing on the value.

justinswart: Not sure what's up with the double review 😄

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -675,22 +690,23 @@ private enum Notification {
...
-
   case following
   case recommendations
+  case showPublicProfile
 
   fileprivate var trackingString: String {
     switch self {
     case .following: return Strings.Following()
     case .recommendations: return Strings.Recommendations()
+    case .showPublicProfile: return "Show public profile"
```
justinswart: Something that's handy to do here while waiting for a translation is to write it out as `localizedString(key: "Show_public_profile", defaultValue: "Show public profile")`, that way once the translation comes through you can just do a search on the project for instances of `localizedString` to be sure you've caught them all.

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -675,22 +690,23 @@ private enum Notification {
     case .creatorTips:                              return "Creator tips"
     case .follower, .mobileFollower:                return "New followers"
     case .friendActivity, .mobileFriendActivity:    return "Friend backs a project"
-    case .messages, .mobileMessages:                 return "New messages"
+    case .messages, .mobileMessages:                return "New messages"
```
justinswart: Assuming these don't need translations?

Library/ViewModels/SettingsViewModel.swift
ifbarrera: Ok so I wasn't sure about this one - do we actually want the localized string since its a tracking string? 🤔I thought we might want the english version always...

justinswart: Oh sure, I just noticed that the others were localized, not 💯sure about the use case here.

ifbarrera: You know what I just double checked with Nneka and we don't need to track this anyway. I've removed this tracking string and added a default instead.

Library/ViewModels/SettingsViewModel.swift
ifbarrera: I'm assuming they don't, since these are used for tracking.

Library/ViewModels/SettingsViewModel.swift
ifbarrera: I moved the transformation to line 186:

```
self.privateProfileEnabledProperty.signal.negate().map {
    (UserAttribute.privacy(Privacy.showPublicProfile), $0)
}
```

The issue is that it needs to be transformed before it gets stored on the `User` object since variable name has the opposite meaning: `showPublicProfile` (on the User object) vs. `privateProfileEnabled` (user facing text). Does it make more sense to have it there?

justinswart: Yip that's good 👍 We know that all of our transformations are happening in our output signals so that's where we'll look for it in the future.

I have one more question, I'll ask it on the relevant line.

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -541,9 +541,9 @@ SettingsViewModelOutputs {
     self.postLikesTappedProperty.value = selected
   }
 
-  fileprivate let showPublicProfileProperty = MutableProperty(false)
+  fileprivate let privateProfileEnabledProperty = MutableProperty(false)
```
justinswart: Since this has flipped, is the default still `false`?

ifbarrera: Duh yes you're right, it should be `true`. Thanks!

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -827,6 +833,10 @@ internal final class SettingsViewController: UIViewController {
     self.helpViewModel.inputs.helpTypeButtonTapped(.privacy)
   }
 
+  @IBAction func privateProfileSwitchDidChange(_ sender: UISwitch) {
```
eoji: does this need to be `fileprivate` also?

KsApi/lib/Route.swift
```diff
@@ -339,11 +339,7 @@ internal enum Route {
       return (.PUT, "/v1/users/self/notifications/\(notification.id)", params, nil)
 
     case let .updateUserSelf(user):
-
-      var params: [String: Any] = [:]
-      params = user.notifications.encode().withAllValuesFrom(user.newsletters.encode())
-      params["social"] = user.social
-      params["opted_out_of_recommendations"] = user.optedOutOfRecommendations
+      let params = user.encode()
```
eoji: Nice!

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -675,22 +690,23 @@ private enum Notification {
...
-
   case following
   case recommendations
+  case showPublicProfile
 
   fileprivate var trackingString: String {
     switch self {
     case .following: return Strings.Following()
     case .recommendations: return Strings.Recommendations()
+    default: return ""
```
eoji: should we add one for `showPublicProfile`? (I'm not sure)

ifbarrera: I checked with Nneka and we won't be tracking privacy switch changes, so no need to add one.

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
ifbarrera: Added!

PR #347, opened 2018-05-04
Discovery cell project category experiment

ifbarrera:
# What

Adding project category and `Projects We Love` tag to the discovery postcard cells, as an A/B experiment.

simulator screen shot - iphone 8 - 2018-05-07 at 17 03 52

simulator screen shot - iphone 5s - 2018-05-07 at 17 03 18

PR #374, opened 2018-06-11
Thankyou card bug fix

ifbarrera:
# What

Bug fix for unresponsive thank you cards in the thank you page. I introduced this when I added project categories into the cards, and missed a spot that needed to be updated to reflect the new data source type 😭We actually had a test for this scenario that would have failed, but the test file wasn't added to the list of sources, so it wasn't run. This PR fixes the bug and adds the `ThanksProjectDataSourceTests` into the framework test target.

I've also removed the `DiscoveryPostcardCell` view from the `DiscoveryPage` storyboard. We don't use this anymore because we have a `DiscoveryPostcardCell.xib` that we reuse in the thank you page. Tested on both the regular discover view and the thank you screen.

# Why

https://trello.com/c/oRcheQCq/763-ios-post-pledge-thank-you-page-project-recommendation-cards-are-unresponsive

# See 👀

Original PR: https://github.com/kickstarter/ios-oss/pull/347

Video:

zx94c7tgnc


justinswart:
Could you perhaps add a link to the PR where the type was added?

ifbarrera:
@justinswart added the PR link to the description!

ifbarrera:
[iOS post-pledge Thank you page: project recommendation cards are unresponsive](https://trello.com/c/oRcheQCq/763-ios-post-pledge-thank-you-page-project-recommendation-cards-are-unresponsive)

PR #379, opened 2018-06-20
Move beta tools into their own screen

ifbarrera:
# What

Move all of our beta tools out of the settings screen and into their own screen, accessible by tapping on a "bug" icon on the navigation bar in the discovery screen (logged in or logged out).

Other bugs fixed:
- environment name wasn't being saved/restored when you relaunched the app
- discovery pages weren't being refreshed when you changed environments in a logged out state

# Why

It makes sense to collect all of our beta tooling functionality into a single view model & view controller and centralize the logic for showing/displaying the beta tools in one place. This also helps us clean up some of the complexity of the `SettingsViewController` in anticipation of a bigger refactor of this screen in the coming weeks.

# How

*New*
`BetaToolsViewController`
`BetaToolsViewModel`
`BetaToolsViewModelTests`
`BetaToolsViewControllerTests`

All new files & tests for this view, including screenshot tests in only one screensize.

# See 👀

vspq5zqtwk


Library/ViewModels/SettingsViewModel.swift
```diff
@@ -264,8 +255,7 @@ SettingsViewModelOutputs {
       .ignoreValues()
 
     self.logoutWithParams = Signal.merge (
-      self.logoutConfirmedProperty.signal,
-      self.environmentSwitcherButtonTappedProperty.signal.skipNil().ignoreValues()
+      self.logoutConfirmedProperty.signal
```
Scollaco: I don't think we need `merge` here anymore.

Kickstarter-iOS/ViewModels/BetaToolsViewModel.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    self.currentLanguage = self.currentLanguageProperty.signal
+      .skipRepeats()
+      .filter { AppEnvironment.current.language != $0 }
+
+    let updateEnvironment = self.environmentSwitcherButtonTappedProperty.signal.skipNil()
+      .filter { AppEnvironment.current.apiService.serverConfig.environment != $0 }
+
+    _ = updateEnvironment
+      .map(ServerConfig.config(for:))
```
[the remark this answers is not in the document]
ifbarrera: The `updateEnvironment` signal is used for both the `environmentSwitcherButtonTitle` and `logoutWithParams` outputs. `observeValues` transforms the signal into `Disposable` which the above outputs can't use. Is there a better way to approach this?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -177,7 +177,10 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
       .takePairWhen(selectedTabAgain)
       .map { vcs, idx in vcs[idx] }
 
-    self.tabBarItemsData = Signal.combineLatest(currentUser, self.viewDidLoadProperty.signal)
+    self.tabBarItemsData = Signal.combineLatest(currentUser, .merge(
```
justinswart: So now that we want this signal to emit when either of these events occur, we're essentially saying "emit the tab bar items, based on the current user when the view loads or when the current language changes". I find in these situations, `takeWhen` sometimes reads better and _can_ be less error prone than `combineLatest`:
```
self.tabBarItemsData = currentUser
  .takeWhen(.merge(
    self.currentLanguageProperty.signal)
  )
  .map(tabData(forUser:))
```

justinswart: Ah, scratch that - it's not really applicable here because we _also_ want the `currentUser` changing to cause this to emit. I've left the comment here just for consideration should you want to use `takeWhen` in other places.

PR #381, opened 2018-06-27
Export data load status

cdolm92:
# What

Added animation indicator for when export data is processing.
# Why

User feedback

# See 👀

jun-27-2018 18-10-29

# TODO ⏰

- [ ] Correct the expired time shown


Library/ViewModels/SettingsViewModel.swift
```diff
@@ -576,6 +611,34 @@ SettingsViewModelOutputs {
...
+private func dateFormatter(for dateString: String?, state: ExportDataEnvelope.State) -> String {
+  let dateFormatter = DateFormatter()
+  dateFormatter.dateFormat = "yyyy-MM-dd"
+  let newDateFormatter = DateFormatter()
+  newDateFormatter.dateFormat = "MMM d"
+
+  let timeFormatter = DateFormatter()
+  timeFormatter.dateFormat = "HH-mm-sZ"
+  let newTimeFormatter = DateFormatter()
+  newTimeFormatter.timeZone = TimeZone(secondsFromGMT: TimeZone.current.secondsFromGMT())
```
cdolm92: time shown was previously off by an hour. I thought this was the way to fix that. 🤔

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -576,6 +611,34 @@ SettingsViewModelOutputs {
...
+  newTimeFormatter.dateFormat = "h:mm a"
```
ifbarrera: It looks like we might already have a helper function that does a lot of the heavy lifting with date formatting, and caches the formatters whenever possible (this is good for performance). `Format.date(seconds: secondsFromUTC, template: "h: mm a", timeZone: TimeZone.current)` could work, but first you'd have to convert the `dateString` into a `Date` object. It looks like those helper functions (in `Format.swift`) don't currently have a helper that goes from `String -> Date`, but we could always add one. I think converting the `dateString` to a `Date` first could be a better approach here because then you don't have to do all the string parsing manually, which could be error-prone. Happy to pair on this with you if you like!

cdolm92:
@Scollaco @ifbarrera  g2g?

ifbarrera:
Hey @cdolm92 according to the requirements on the [card](https://trello.com/c/kB6Mmg2y/718-export-data-loading) the text should change to "Download my personal data" once the download state is `completed`, and tapping "download my personal data" should open the download link. Is the plan to submit this work in an additional PR?

PR #383, opened 2018-06-28
Add Slack messages for alpha, beta, production

justinswart:
# What

Posts messages to Slack whenever Alpha, Beta or Production builds are uploaded to HockeyApp or App Store Connect respectively.

# Why

For transparency and easy reference.

# How

Updated our webhook on CircleCI and moved it to the channel that @nnekab suggested.

# See 👀

Will look something like:

screen shot 2018-06-28 at 4 43 42 pm

### DIFF
--- a/.fastlane/Fastfile
+++ b/.fastlane/Fastfile
@@ -144,7 +144,7 @@ platform :ios do
         "production",
         get_version_number(xcodeproj: "Kickstarter.xcodeproj", target: "Kickstarter-iOS"),
         get_build_number(xcodeproj: "Kickstarter.xcodeproj"),
-        "iTunes Connect 🚀"
+        "App Store Connect 🚀"
       )
     )
   end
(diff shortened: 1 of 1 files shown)


.fastlane/Fastfile
```diff
@@ -130,10 +140,12 @@ platform :ios do
 
     slack(
       slack_url: ENV["SLACK_WEBHOOK"],
-      message: "Successfully uploaded" +
-        " v" + get_version_number(xcodeproj: "Kickstarter.xcodeproj", target: "Kickstarter-iOS") +
-        " (" + get_build_number(xcodeproj: "Kickstarter.xcodeproj") + ")" +
-        " of the Kickstarter iOS app to iTunes Connect 🚀"
+      message: slack_message(
+        "production",
```
ifbarrera: Hmm I wonder if this would be clearer as "App Store" or "iTunes Connect"?

justinswart: Oh sure, what do you suggest: `App Store`, `Alpha`, `Beta`?

Seems like Apple has _finally_ renamed _iTunes Connect_ to _App Store_ 🍾

ifbarrera:
I think App Store makes sense!

PR #384, opened 2018-06-29
Remove old ProfileViewController.

Scollaco:
# What
Removed old ProfileViewController.

# Why
The ViewController was no longer being used, but the codebase still had some logic applied to it, what caused a bug where, due a `Config` decoding error, the old profile (blank screen) was being shown instead of the `BackerDashboardViewController`.

# How
delete, delete, delete

GIF


Scollaco:
[Support DEV Bug: P1: [iOS] Profile section on app is empty.](https://trello.com/c/nCreysfa/825-support-dev-bug-p1-ios-profile-section-on-app-is-empty)

Scollaco:
[Remove ProfileViewController feature flag (and remove the unused VC)](https://trello.com/c/o0tTqM9Z/831-remove-profileviewcontroller-feature-flag-and-remove-the-unused-vc)

PR #385, opened 2018-07-02
Updating debug bundleId and adding alpha Hockey appId

ifbarrera:
# What

- Removed the bundleId `com.kickstarter.kickstarter.alpha` which was being used for debug builds in favor of `come.kickstarter.kickstarter.debug`. Simpler now.
- fixed a bug where the hockey build updates were displaying incorrectly for `KickAlpha`
- fixed a bug where the debug tools were not accessible on alpha builds


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -578,15 +578,14 @@ AppDelegateViewModelOutputs {
       self.userSessionStartedProperty.signal,
       self.userSessionEndedProperty.signal
       )
+      .filter { !AppEnvironment.current.mainBundle.isDebug }
```
justinswart: So just to confirm - this line would mean HockeyApp doesn't run on debug builds at all (the ones we're testing locally on device) which I think is fine.

I would just say that we probably don't need the additional check on line `588` as we should trust the `filter` to prevent that `mainBundle.isDebug` would ever be `true` here.

Library/ViewModels/DiscoveryNavigationHeaderViewModel.swift
```diff
@@ -122,7 +122,10 @@ DiscoveryNavigationHeaderViewModelInputs, DiscoveryNavigationHeaderViewModelOutp
     }
 
     self.debugContainerViewIsHidden = self.viewDidLoadProperty.signal
-      .map { !AppEnvironment.current.mainBundle.isAlpha && !AppEnvironment.current.mainBundle.isBeta }
+      .map {
```
justinswart: I find this a little strange to read, we're saying 'hide the debug container _if_ we're not this and not this and not this. Would it be easier to read if it was something like:

"if we're any of these things, don't hide the container"
```
.map { AppEnvironment.current.mainBundle.isDebug || AppEnvironment.current.mainBundle.isAlpha || AppEnvironment.current.mainBundle.isBeta }
.map { !$0 }
```

ifbarrera: Good point, or maybe even just `.map { AppEnvironment.current.mainBundl.isRelease }` aka "hide the debug menu if the build is release"

justinswart: Ah indeed!

Library/ViewModels/DiscoveryNavigationHeaderViewModelTests.swift
```diff
@@ -510,13 +510,4 @@ internal final class DiscoveryNavigationHeaderViewModelTests: TestCase {
       self.debugContainerViewIsHidden.assertValue(true)
     }
   }
-
-  func testEnvironmentButtonIsHidden_Unknown() {
```
ifbarrera: I removed this because I don't think this test is necessary.

Library/NSBundleType.swift
```diff
@@ -27,6 +30,14 @@ extension NSBundleType {
     return self.infoDictionary?["CFBundleVersion"] as? String ?? "0"
   }
 
+  public var isTest: Bool {
```
justinswart: Ah interesting, what's this used for?

ifbarrera: When I changed this line: https://github.com/kickstarter/ios-oss/pull/385/files#diff-63e4e32ea3a28f8587151aa70d4525fbR125 we had a snapshot test that started failing because the debug tools button was being shown on test builds. I added this so we can hide the debug menu on test bundleIds.

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewControllerTests.swift
```diff
@@ -19,12 +19,11 @@ internal final class DiscoveryNavigationHeaderViewControllerTests: TestCase {
 
   override func setUp() {
     super.setUp()
-    AppEnvironment.pushEnvironment(mainBundle: Bundle.framework)
```
justinswart: Just want to understand this change and the one on line 27. I think the pushing and popping of the environment may have been intentional in this test. I put these lines back and ran the tests locally and they fail because the debug menu icon appears on the header. I'm wondering if we'd rather want to fix that by not showing the debug icon unless we're running the debug build specifically?

One option is this line could change to `AppEnvironment.pushEnvironment(mainBundle: MockBundle(bundleIdentifier: KickstarterBundleIdentifier.test.rawValue))` now that we have the test bundle identifier and we can bring back the pushing and popping of environments?

But I think ideally we don't want to be making too many decisions based on whether we're in a test or not, so this might point to our booleans that we changed in `DiscoveryNavigationHeaderViewModel.swift`'s `debugContainerViewIsHidden` being the wrong approach.

Perhaps we should revert back to explicitly checking those booleans to show/hide the menu. Here's another suggestion:

1. Add something like this to `NSBundleType` (naming could be better):
```
public var isUserTestableBuild: Bool {
  return self.isBeta || self.isAlpha || self.isDebug
}
```
2. Change whether to show/hide the debug menu to:
```
self.debugContainerViewIsHidden = self.viewDidLoadProperty.signal
  .map { !AppEnvironment.current.mainBundle.isUserTestableBuild }
```
3. Remove `isTest` because we shouldn't need it in this case.

Lemme know what you think!

ifbarrera: Sounds good to me! Added a helper func to `DiscoveryNavigationHeaderViewModel` 👍

PR #395, opened 2018-07-31
Adding Fabric & Crashlytics

ifbarrera:
# What

Adding support for Fabric crash reporting.

# Why

Better crash reporting 🐞 

# How

Added the `fabric` recipe to the makefile that downloads the Fabric SDK to the directory `Frameworks/Fabric`. A build phase script then conditionally configures the SDK in CI builds by checking for the `FABRIC_API_KEY` environment variable. This `ENV` variable is only available on builds in the private remote.


bin/fabric.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ ! -z "$CI" ] && [ ! -z "$FABRIC_API_KEY" ]; then
+	echo "Configuring Fabric..."
+	../Frameworks/Fabric/Fabric.framework/run "$FABRIC_API_KEY"
```
justinswart: Xcode's environment variables should also be available here from the run phase, so perhaps this could be:

`"${PROJECT_DIR}"/Frameworks/Fabric/Fabric.framework/run "$FABRIC_API_KEY"`

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -169,6 +171,14 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         self?.presentRemoteNotificationAlert($0)
       }
 
+    self.viewModel.outputs.configureFabric
```
justinswart: Could wrap lines 174 - 180 in the conditional that way nothing is subscribed to in `DEBUG`.

ifbarrera:
[Crashlytics Integration](https://trello.com/c/1YnwPaX6/816-crashlytics-integration)

PR #396, opened 2018-07-31
Discover card bug fix

Scollaco:
# What
Disable `Discover` card for rewards of projects outside US and non-US users.

# Why
We are having problems to complete transactions when projects were funded on these conditions, and to avoid the transactions to fail again, as it's happening now, we are disabling Discover cards for projects outside US and non-US users.

For more details: 
https://trello.com/c/4heDPDAs/274-spike-apple-pay-legacy-issues

GIF


Scollaco:
[Block JCB and Discover cards for non-US backers or any backer pledging to a non-US account](https://trello.com/c/tGoedciG/833-block-jcb-and-discover-cards-for-non-us-backers-or-any-backer-pledging-to-a-non-us-account)

PR #401, opened 2018-08-06
Conditional creator digest notifications

ifbarrera:
# What

When pledge activity email notifications are enabled, show a "Email frequency" button that allows a user to select how frequently they should receive notifications. The option presents a picker with the two frequency options. Selecting an option should update the user's notification setting.

If pledge activity email notifications are disabled, the "Email frequency" cell should not show.

# Why

A more intuitive way to select email notification frequency for pledge activity emails.

# See 👀

vn2h3jwo0q

# TODO ⏰

- [ ] delete `CreatorDigestSettingsViewController` and associated view model, tests, etc
- [ ] implement custom UIPickerView & additional design requirements
- [x] write tests for `insert` and `deleteRows` functions in `ValueCellDataSource`


ifbarrera:
[Conditional changes depending on whether the user (a creator) has chosen to receive Pledge activity by email or not](https://trello.com/c/LiCNctRS/895-conditional-changes-depending-on-whether-the-user-a-creator-has-chosen-to-receive-pledge-activity-by-email-or-not)

Kickstarter-iOS/DataSources/SettingsNotificationsDataSource.swift
```diff
@@ -8,7 +8,7 @@ public struct SettingsNotificationCellValue {
 }
 
 final class SettingsNotificationsDataSource: ValueCellDataSource {
-  weak var cellDelegate: SettingsNotificationCellDelegate?
+  weak var cellDelegate: SettingsNotificationsViewController?
```
justinswart: Why the type rename?

Kickstarter-iOS/EmailFrequency.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
+
+public enum EmailFrequency: Int {
+  case daily
+  case individualEmails
+
+  public static let allCases: [EmailFrequency] = [.daily, .individualEmails]
+
+  public static let rowHeight: CGFloat = 45.0
```
justinswart: Curious - I know in the past we used grid values for things, not really applicable here? Not sure if designs still come in grids?

Kickstarter-iOS/ViewModels/SettingsNotificationPickerViewModel.swift
```diff
@@ -0,0 +1,62 @@
+import KsApi
+import Library
+import Prelude
+import ReactiveSwift
+import Result
+
+protocol SettingsNotificationPickerViewModelOutputs {
+  var didTapFrequencyPickerButton: Signal<Void, NoError> { get }
```
justinswart: I think this reads like an input rather than an output. The output should emit _what_ side-effect should happen in the VC, not necessarily the event that triggered it.

So actually in this case it would be something like `notifyDelegateDidTapFrequencyButton`.

Kickstarter-iOS/ViewModels/SettingsNotificationPickerViewModel.swift
```diff
@@ -0,0 +1,62 @@
...
+  var frequencyValueText: Signal<String, NoError> { get }
+}
+
+protocol SettingsNotificationPickerViewModelInputs {
+  func configure(with cellValue: SettingsNotificationCellValue)
```
justinswart: Yeah I think this way of naming the `configure(with:)` methods is better 👍

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
```diff
@@ -21,7 +25,11 @@ internal final class SettingsNotificationsViewController: UIViewController {
     tableView.dataSource = dataSource
     tableView.delegate = self
 
+    emailFrequencyPickerView.delegate = self
```
justinswart: Think we should prefix all of these property references with `self`

Kickstarter-iOS/Views/SettingsNotificationPickerCell.swift
```diff
@@ -0,0 +1,57 @@
...
+
+  weak var delegate: SettingsNotificationPickerCellDelegate?
+
+  private let viewModel: SettingsNotificationPickerViewModelType = SettingsNotificationPickerViewModel()
+
+  func configureWith(value: SettingsNotificationCellValue) {
+    self.viewModel.inputs.configure(with: value)
+
+    _ = titleLabel
+    |> UILabel.lens.text .~ value.cellType.title
```
justinswart: Think this needs an indent.

Kickstarter-iOS/Views/SettingsNotificationPickerCell.swift
```diff
@@ -0,0 +1,57 @@
...
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = titleLabel
+    |> settingsTitleLabelStyle
+
+    _ = selectFrequencyButton
+    |> UIButton.lens.titleLabel.font .~ .ksr_body()
+    |> UIButton.lens.titleColor(for: .normal) .~ .ksr_text_dark_grey_400
```
justinswart: Few indents in this function, just helps to see that these styles are applying to the button above. Also think we want to prefix all the property refs with `self`.

Kickstarter-iOS/Views/SettingsNotificationPickerCell.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import KsApi
+import Library
+import Prelude
+
+protocol SettingsNotificationPickerCellDelegate: class {
+  func didFailToSaveChange(errorMessage: String)
```
justinswart: I made a few comments in #403 too on the delegate method naming. It's up to you guys really but I tend to think the delegate method should include the object that triggered it. I know it can get a little verbose sometimes but it helps to stick to a convention. The [Google Swift style guide](https://google.github.io/swift/#delegate-methods) also explains it well.

So for this for example: 
```swift
func settingsNotificationPickerCell(_ cell: SettingsNotificationPickerCell, didFailToSaveChange errorMessage: String)
```

Library/ViewModels/SettingsNotificationsViewModel.swift
```diff
@@ -1,21 +1,25 @@
...
 import Prelude
 import ReactiveSwift
 import ReactiveExtensions
 import Result
 
 public protocol SettingsNotificationsViewModelInputs {
   func didSelectRow(cellType: SettingsNotificationCellType)
+  func didSelectEmailFrequency(frequency: EmailFrequency)
   func failedToUpdateUser(error: String)
+  func showPickerView(show: Bool)
```
justinswart: This also reads like an output, so it should probably be more like `func didTapFrequencyPickerButton()` which is really the event that occurred, also seeing as that always passes in true then this can just be mapped to `true`/`false` on [line 86](https://github.com/kickstarter/ios-oss/pull/401/files#diff-5201ce86be2c9258634f949a836fff31R86).

Kickstarter-iOS/DataSources/SettingsNotificationsDataSource.swift
ifbarrera: `cellDelegate` is now the delegate for both `SettingsNotificationPickerCell` and `SettingsNotificationCell`. `SettingsNotificationsViewController` conforms to both of these delegate protocols.

Kickstarter-iOS/EmailFrequency.swift
ifbarrera: Ah ok I didn't know about the grid values, now I do! Will update

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
ifbarrera: Just curious, how come? I thought this wasn't necessary unless you're referencing `self` in a closure

ifbarrera: Unless it's just for consistency since we do it everywhere else in the repo?

Kickstarter-iOS/Views/SettingsNotificationPickerCell.swift
ifbarrera: Oh actually this delegate function isn't even needed. Will remove, but I hear you about delegate function naming.

ifbarrera: I'll rename the others

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
justinswart: Oh yeah, if for no other reason it'd be for consistency with the rest of the codebase. I personally think it's handy to know if something is global or specific to the instance just by reading it but I know it's also something of personal preference.

Kickstarter-iOS/DataSources/SettingsNotificationsDataSource.swift
justinswart: oh cool, seeing as conformance to the protocols is still the thing we care about here, can I suggest using Swift's protocol composition - either:

```swift
typealias SettingsNotificationCellDelegates = SettingsNotificationPickerCellDelegate & SettingsNotificationCellDelegate

weak var cellDelegate: SettingsNotificationCellDelegates?
```
or:

```swift
weak var cellDelegate: (SettingsNotificationPickerCellDelegate & SettingsNotificationCellDelegate)?
```

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
```diff
@@ -88,15 +104,30 @@ internal final class SettingsNotificationsViewController: UIViewController {
...
     self.navigationController?.pushViewController(vc, animated: true)
   }
+
+  private func animatePickerView(isHidden: Bool) {
+    UIView.animate(withDuration: 0.25) {
+      self.emailPickerViewTopConstraint.constant = isHidden
+        ? 0 : -SettingsNotificationsViewController.emailPickerViewHeight
+
+      self.view.setNeedsLayout()
+      self.view.layoutIfNeeded()
```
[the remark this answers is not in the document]
ifbarrera: No, you have to call `layoutIfNeeded` to force a layout otherwise it doesn't animate.

ifbarrera:
@justinswart the picker in this PR is a temporary generic `UIPickerView`. We're planning on replacing it with a custom `UIPickerView` that @dannyalright designed next sprint!

dannyalright:
@justinswart's feedback is great. An inline picker is an obviously better solution than rolling a custom bottom-picker. Is that ok for this ticket @ifbarrera or should I make a new one?

Here's my rendering of that (the picker background should be `#ffffff` white):

367f7e6a-4bcc-4d63-be0d-24995275fbce

[Abstract link](https://share.goabstract.com/81fbe2c5-7747-4e74-8252-5a8704721b3d).

RE feedback #1: the entire cell should be tappable already.

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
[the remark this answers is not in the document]
justinswart: Oh true but just was wondering if both calls were needed?

justinswart:
@dannyalright cool! Up to you and @ifbarrera if you want to do it as part of this PR or iterate in another 👍

PR #404, opened 2018-08-09
Move & rename circle.yml to .circleci/config.yml

yatriks:
### What
As CircleCI sunsets version 1 configurations, we get a warning for this repo even though it's on version 2. It has to do with the filenames/locations.

### Why
So we can stop seeing that pesky yellow banner:
image

### CC
@kickstarter/infrastructure @eoji


yatriks:
🎉 ty!

PR #406, opened 2018-08-13
Update Facebook reconnect copy

ifbarrera:
# What

Updates the facebook reconnect copy.

# Why

Need to include the disclaimer of the info we collect when logging in with FB.

# See 👀

testview_showfacebookreconnect_lang_en_device_phone4_7inch 2x


ifbarrera:
[Re-Connect with facebook legal copy](https://trello.com/c/SmTLfa93/910-re-connect-with-facebook-legal-copy)

PR #407, opened 2018-08-13
Update debug push notification

Scollaco:
# What
Updated function that allows the user to see push notifications and fixed a bug where the notification isn't being shown if the app is active.

# Why
This feature is very important to test the app behaviour when it receives push notifications.

**note**: The notification debug tool creates a `LocalNotification` to simulate a `Remote Notification`. So the notification you see after tapping `In-App` or `Delayed` on the Debug Screen is not a remote push notification, but just a simulation.

# How
For iOS 10+, the function that creates the LocalNotification was updated to use `UNMutableNotificationContent`, `UNNotificationRequest`, and ` UNUserNotificationCenter.current().add()` function. We are still supporting `UILocalNotification` for iOS 9 devices, though.

To fix the alert message bug, the viewcontroller hierarchy was changed, in order to avoid the `Warning: Attempt to present UIAlertController: whose view is not in the window hierarchy!` error.

# See 👀
 

GIF

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -283,7 +283,6 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.viewModel.inputs.didReceive(remoteNotification: userInfo,
                                      applicationIsActive: application.applicationState == .active)
   }
-
   internal func application(_ application: UIApplication,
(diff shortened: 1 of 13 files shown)


Scollaco:
[Delayed App notifications debug tool not working](https://trello.com/c/SAke0X0y/860-delayed-app-notifications-debug-tool-not-working)

PR #409, opened 2018-08-15
Deleting unused CreatorDigestSettingsViewController

ifbarrera:
# What

Deletes `CreatorDigestViewController` because that functionality has been moved to the main `SettingsNotificationViewController`

# Why

Clean up!


Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -324,8 +324,7 @@ internal final class SettingsViewController: UIViewController {
   }
 
   fileprivate func goToEmailFrequency(user: User) {
-    let vc = CreatorDigestSettingsViewController.configureWith(user: user)
```
ifbarrera: Left this empty because `SettingsViewController` will be deleted soon also.

PR #417, opened 2018-08-21
Settings transition

cdolm92:
# What

Fixed the transition to the settings screen.

# See 👀
settings-transition


dannyalright:
this is great! Thanks for sneaking it in

PR #422, opened 2018-08-30
Move "Find Facebook Friends" to main settings screen

ifbarrera:
# What

Moves the "Find Facebook Friends" row from the Notifications screen to the main Settings screen. Conditionally disables the row if the user has disabled "following" (aka social), and adds an informational message below the cell.

# Why

To improve social feature visibility.

# How

Added a notification listener for `.ksr_userUpdated` that refreshes the user in the main settings screen. This notification is fired when the current user is updated from changes to the `following` setting in the privacy screen.

# See 👀

simulator screen shot - iphone 8 - 2018-08-30 at 18 38 00

# TODO ⏰

- [ ] localized copy


Library/ViewModels/SettingsPrivacyViewModel.swift
```diff
@@ -72,10 +73,9 @@ SettingsPrivacyViewModelInputs, SettingsPrivacyViewModelOutputs {
       .takeWhen(self.unableToSaveError)
       .map { previous, _ in previous }
 
-   self.updateCurrentUser = Signal.merge(initialUser, updatedUser, previousUserOnError)
+   self.updateCurrentUser = Signal.merge(updatedFetchedUser, previousUserOnError)
```
ifbarrera: I removed the `initialUser` signal here to reduce the number of notification calls to `.ksr_userUpdated` which actually has a lot of listeners.

Library/ViewModels/SettingsPrivacyViewModel.swift
```diff
@@ -62,6 +61,8 @@ SettingsPrivacyViewModelInputs, SettingsPrivacyViewModelOutputs {
           .materialize()
     }
 
+    let updatedFetchedUser = updateEvent.values()
```
ifbarrera: This fixed a race condition where the notification for `.ksr_userUpdated` was firing *before* the user was actually updated on the server. Spent an infuriating hour looking for this bug.

Kickstarter-iOS/Views/Cells/FindFriendsCell.swift
```diff
@@ -0,0 +1,70 @@
...
+      |> settingsTitleLabelStyle
+      |> UILabel.lens.text .~ SettingsCellType.findFriends.title
+
+    _ = arrowImageView
+      |> settingsArrowViewStyle
+
+    _ = disabledDescriptionLabel
+      |> settingsDescriptionLabelStyle
+      |> UILabel.lens.text
+      .~ "Following must be on to find Facebook friends. Following can be changed in Account > Privacy."
```
justinswart: Does this need to be localized? I've found it useful to use the `localizedString()` function while waiting for translations so that it's easy to do a find/replace on that once the translations are ready. It's also safe to merge because `localizedString()` is what all of the `Strings` functions are using in any case.

Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
```diff
@@ -50,7 +50,7 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
       .observeForUI()
       .observeValues { [weak self] in
         guard let _self = self else { return }
```
justinswart: For interest sake, @dusi and I have started using:
```swift
guard let `self` = self else { return }
```
That way the subsequent line can refer to:
```swift
self.delegate?.settingsFollowCellDidDisableFollowing(self)
```
and the compiler is satisfied. The backticks around `self` just allows one to use a reserved keyword in local scope.

Kickstarter-iOS/Views/Cells/SettingsTableViewCell.swift
```diff
@@ -11,16 +11,18 @@ final class SettingsTableViewCell: UITableViewCell, ValueCell, NibLoading {
     super.init(coder: aDecoder)
   }
 
-  func configureWith(value cellType: SettingsCellTypeProtocol) {
+  func configureWith(value cellValue: SettingsCellValue) {
```
justinswart: Do we want to change these to the `configure(with:)` pattern when we touch them?

Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
ifbarrera: Yeah I've used that in the past too, but for consistency continued using the `_self` pattern since that's what we use everywhere else.

Kickstarter-iOS/Views/Cells/SettingsTableViewCell.swift
ifbarrera: I think this is part of the `ValueCell` protocol, we would have to change it everywhere we use `ValueCell`s. Not sure if this PR is the best place for that?

PR #424, opened 2018-09-07
Re-adding private profile setting

ifbarrera:
# What

Add the ability to make your profile private.

# Why

We had this before, but it accidentally got removed with the settings refactor.

# How

Added a new cell `SettingsPrivacySwitchCell` which is configured with a user object and reads from the `showPublicProfile` setting on the `User` object. Toggling the switch triggers an update on the user object.

# See 👀

xsizlsndgm

# TODO ⏰

- [ ] New string translations


Kickstarter-iOS/ViewModels/SettingsPrivacySwitchCellViewModelTests.swift
```diff
@@ -0,0 +1,46 @@
+import Library
+import Prelude
+import Result
+import XCTest
+@testable import Kickstarter_Framework
+@testable import KsApi
+@testable import ReactiveExtensions_TestHelpers
+
+final class SettingsPrivacySwitchCellViewModelTests: TestCase {
+  private let privacySwitchEnabledObserver = TestObserver<Bool, NoError>()
```
Scollaco: Here I would change to `privacySwitchIsOnObserver`. For the same reason.

Kickstarter-iOS/Views/Cells/SettingsPrivacySwitchCell.swift
```diff
@@ -0,0 +1,79 @@
...
+import Library
+import Prelude
+
+protocol SettingsPrivacySwitchCellDelegate: class {
+  func privacySettingsSwitchCell(_ cell: SettingsPrivacySwitchCell,
+                                 didTogglePrivacySwitch on: Bool)
+}
+
+final class SettingsPrivacySwitchCell: UITableViewCell, ValueCell, NibLoading {
+  @IBOutlet fileprivate weak var lineLayer: UIView!
```
Scollaco: Being a little picky, throughout the app, we name `lineLayer`  by `separatorView`, since it simulates the UITableView separator. I would rename the property for consistency.

Kickstarter-iOS/Views/Cells/SettingsPrivacySwitchCell.swift
```diff
@@ -0,0 +1,79 @@
...
+  @IBOutlet fileprivate weak var primaryDescriptionLabel: UILabel!
+  @IBOutlet fileprivate weak var titleLabel: UILabel!
+  @IBOutlet fileprivate weak var topLineLayer: UIView!
```
Scollaco: Instead of create `lineLayer` and `topLineLayer` IBOutlets, you can ctrl + drag both views and creating a `Outlet Connection`, since they have the same style. You then will have `@IBOutlet fileprivate var separatorViews: [UIView]!` and style them using:
`_ = self.separatorViews
      ||> settingsSeparatorStyle`

PR #426, opened 2018-09-11
Account screen

Scollaco:
# What
The new Account screen.

# Why
This screen is part of the new Settings feature, where among other actions, the user will be able to change account preferences such as email, currency, password...

# See 👀

GIF


Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -0,0 +1,25 @@
+import KsApi
+import Library
+
+final class SettingsAccountDataSource: ValueCellDataSource {
+
+  func configureRows() {
+    _ = SettingsAccountSectionType.allCases.map { section -> Void in
```
ifbarrera: This was a suggestion on one of my PRs from @justinswart , but you might try using `.forEach` instead since we're not actually mutating the values of the array, just performing an action for each of it's items. And then we don't need the discardable variable `_`.

Kickstarter-iOS/Views/Cells/SettingsTableViewCell.swift
```diff
@@ -54,4 +50,24 @@ final class SettingsTableViewCell: UITableViewCell, ValueCell, NibLoading {
     _ = self
       |> UITableViewCell.lens.backgroundColor .~ highlightedColor
   }
+
+  private static func descriptionString(for cellType: SettingsCellTypeProtocol) -> String {
```
ifbarrera: Would it make sense to have these be variables within the `cellType`, similar to how we use `title`? Then it would be called using `cellType.descriptionString`.

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+        self?.tableView.reloadData()
+    }
+  }
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = self
+      |> baseControllerStyle()
+      |> UIViewController.lens.view.backgroundColor .~ .ksr_grey_200
```
ifbarrera: I think we use this styling in enough VCs to warrant it getting its own style property in `SettingsStyles.swift`. Maybe something like `baseSettingsControllerStyle`?

Library/SettingsAccountCellType.swift
```diff
@@ -0,0 +1,80 @@
...
+import UIKit
+
+public enum SettingsAccountSectionType: Int {
+
+  case emailPassword
+  case privacy
+  case payment
+
+  public static var sectionHeaderHeight: CGFloat {
+    return 30.0
```
ifbarrera: How about `Styles.grid(5)`? I'm guilty of not using `Styles.grid` also 😅

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -0,0 +1,59 @@
...
+      .map { SettingsAccountViewModel.viewController(for: $0) }
+      .skipNil()
+  }
+
+  private let viewDidLoadProperty = MutableProperty(())
+  public func viewDidLoad() {
+    self.viewDidLoadProperty.value = ()
+  }
+
+  private var selectedCellTypeProperty = MutableProperty<SettingsAccountCellType?>(nil)
```
justinswart: This should probably be a `let` 😬

Scollaco: Ops! 😬

Scollaco:
[User can see account screen](https://trello.com/c/53wjFugm/959-user-can-see-account-screen)

Scollaco:
[User can see account screen](https://trello.com/c/53wjFugm/959-user-can-see-account-screen)

Scollaco:
[User can see account screen](https://trello.com/c/53wjFugm/959-user-can-see-account-screen)

PR #427, opened 2018-09-12
Hide private profile setting for creator

Scollaco:
# What
Hide `Private profile` setting from creators.

# Why
If the user is a creator, the option to make to profile private shouldn't appear.

# How
On the data source `load` function, we check if the user is a creator before configure the cells, hiding the Private Profile section if that's the case.

# See 👀
### Creator

GIF


Scollaco:
[Hide Private profile from creators](https://trello.com/c/iAbe2ViW/974-hide-private-profile-from-creators)

PR #429, opened 2018-09-14
Swift 4.2

Scollaco:
# What
Made small changes in order to build the app using swift 4.2 and XCode 10.

# Why
Because now it's official and we need it ;)

GIF


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -26,6 +26,7 @@ public enum NotificationAuthorizationStatus {
   case authorized
   case denied
   case notDetermined
+  case provisional
```
dusi: Just looking at the docs shouldn't this be only available on `iOS 12+`?

```
// UNNotificationSettings
@available(iOS 12.0, *)
case provisional
```

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -103,10 +103,9 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
 
     self.backgroundGradientView.startPoint = .zero
     self.backgroundGradientView.endPoint = CGPoint(x: 0, y: 1)
-    self.backgroundGradientView.setGradient([
-      (UIColor(white: 0, alpha: 0.5), 0),
-      (UIColor(white: 0, alpha: 0), 1)
-      ])
+    let gradient: [(UIColor?, Float)] =  [(UIColor.init(white: 0.0, alpha: 0.5), 0.0),
```
dusi: Since we're explicitly declaring `gradient` to be an array of tuples `(UIColor?, Float)` we can simplify
```
[(UIColor.init(white: 0.0, alpha: 0.5), 0.0)
```
to
```
[(UIColor.init(white: 0.0, alpha: 0.5), 0)
```
(unless this is a code style you prefer for the KSR code base)?

Scollaco:
@ifbarrera Yeah. That's the next step

.circleci/config.yml
```diff
@@ -1,17 +1,17 @@
...
 fabric_sdk_version: &fabric_sdk_version 3.10.5
-preload_simulator: &preload_simulator xcrun instruments -w 'iPhone 8 (11.4) [2437FE39-2885-482D-A42F-A169C65AE271]' || true
+preload_simulator: &preload_simulator xcrun instruments -w 'iPhone 8 (12.0) [2437FE39-2885-482D-A42F-A169C65AE271]' || true
 store_fabric_sdk_version: &store_fabric_sdk_version echo "$FABRIC_SDK_VERSION" > fabric_version.txt
 store_opentok_version: &store_opentok_version echo "$OPENTOK_VERSION" > opentok_version.txt
 
 base_job: &base_job
   macos:
-    xcode: "9.4.0"
+    xcode: "10.0"
```
justinswart: Try `"10.0.0"`

.circleci/config.yml
```diff
@@ -1,17 +1,17 @@
...
 
-xcode_version: &xcode_version 9.4
+xcode_version: &xcode_version 10.0
 iphone_name: &iphone_name iPhone 8
-ios_version: &ios_version 11.4
+ios_version: &ios_version 12.0
 opentok_version: &opentok_version 2.10.2
 fabric_sdk_version: &fabric_sdk_version 3.10.5
-preload_simulator: &preload_simulator xcrun instruments -w 'iPhone 8 (11.4) [2437FE39-2885-482D-A42F-A169C65AE271]' || true
+preload_simulator: &preload_simulator xcrun instruments -w 'iPhone 8 (12.0) [2437FE39-2885-482D-A42F-A169C65AE271]' || true
```
justinswart: We don't use this preloading anymore, you can try remove it and see if it works for you. The problem is that the identifier changes a lot so it breaks every time Circle deploys a new version of Xcode.

.circleci/config.yml
```diff
@@ -1,17 +1,16 @@
 version: 2
 
-xcode_version: &xcode_version 9.4
+xcode_version: &xcode_version 10.0
 iphone_name: &iphone_name iPhone 8
-ios_version: &ios_version 11.4
+ios_version: &ios_version 12.0
```
justinswart: Ok, the issue here is that the environment variable `IOS_VERSION` is being interpolated as `12` in the `Makefile` when it executes the command to run tests. This needs to be `12.0`.

We actually only use this env var in the `Makefile` in one place so you have two options:

1. Remove this line from the `.circleci/config.yml` and the reference to it below on line `23`. This will cause it to use the default value set in the `Makefile` and it'll work.
or
2. Try changing this value in the `.circleci/config.yml` to be "12.0" although I think that would cause it to print out in the command as `="12.0"` and that would fail too.

Scollaco:
[Run app using XCode 10](https://trello.com/c/xcRwqkQb/980-run-app-using-xcode-10)

PR #430, opened 2018-09-17
Format decimal values

cdolm92:
# What

Accurate pledge amount now shown for users who pledged a decimal values. 

# How

Helper function `shouldShowDecimal()` determines whether to display the amount pledged as a whole number or decimal value.

# See 👀

When user pledges with decimal value:

simulator screen shot - iphone 8 - 2018-09-17 at 18 33 01

When pledge amount is a whole number:

simulator screen shot - iphone 8 - 2018-09-17 at 18 30 57


Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -372,6 +373,14 @@ private func footerString(project: Project, reward: Reward) -> String {
     .joined(separator: " • ")
 }
 
+private func shouldShowDecimal(_ backing: Backing) -> String {
```
Scollaco: Functions and variables starting with `should` usually return / are of type `Bool`. TableView, for example, has methods like `shouldHighlightRowAt` and `shouldShowMenuForRowAt`. I would rename this function to a name that helps to understand what it's returning exactly. Something like `backingAmount(for backing: Backing)`

Scollaco: Functions and variables starting with should usually return / are of type Bool. TableView, for example, has methods like shouldHighlightRowAt and shouldShowMenuForRowAt. I would rename this function to a name that helps to understand what it's returning exactly. Something like `formattedAmount(for backing: Backing)`

PR #431, opened 2018-09-18
Upgrade Stripe SDK

ifbarrera:
# What

Upgrades to the latest Stripe SDK, and removes the framework from source control.

# Why

So we can have the latest and greatest tools from Stripe!

# How

Added a script to the Makefile that downloads the SDK from Github and puts it in the `Frameworks/Stripe` folder.

## TODOs

- test Apple pay in Production environment


Makefile
```diff
@@ -173,5 +174,20 @@ fabric:
 		rm -rf Frameworks/Fabric; \
 	fi
 
+stripe:
```
justinswart: Now that we're doing this download and unzip dance for 3 frameworks perhaps it makes sense to extract this into a shell script that takes a URL and a path? Pity that the OpenTok one is a tar ball and not a zip though 🤔

ifbarrera: @justinswart noted! I've added a card to our backlog to [refactor the Makefile 3rd party downloads into a bash script](https://trello.com/c/GH1NIT7F/464-update-makefile-for-frameworks-to-use-shared-bash-script)

ifbarrera:
Tested in Production environment for Apple Pay and regular checkout 👍

dusi:
[Upgrade Stripe SDK](https://trello.com/c/hqoF2RZF/955-upgrade-stripe-sdk)

PR #432, opened 2018-09-21
Account hidden release

Scollaco:
# What
Hides the Account section on Production.

# Why
The account screen is not functional yet, so in the mean time it will be hidden on `Release` build but visible on `Alpha`, `Beta` and `Debug` for testing purposes.

# See 👀


Scollaco:
[User can see account screen](https://trello.com/c/53wjFugm/959-user-can-see-account-screen)

PR #437, opened 2018-09-25
Contextual Dialog bug fix

Scollaco:
# What
Removed ubiquitousStore if statements from PushNotificationDialog.

# Why
The value we use to check if we should show the Contextual Dialog (`true` by default and `false` if the user select `Not now` 3x on the alert dialog) is being saved on iCloud (ubiquitousStore) and restored if the app is reinstalled. The problem with that is the fact the Push Notifications settings reset on the iOS level, and because we are not showing the dialog anymore, the user is not able to re-register for push notifications.


Scollaco:
@ifbarrera Just needed to remove `ubiquitousStore` from the environment initialisation (for cleanup). The code adds the values to both UserDefaults and UbiquitousStore (as a backup, case the user deletes the app) but uses only one of them to apply the necessary logic. Therefore, by deleting ubiquitousStore, the result of the tests didn't change.

Scollaco:
[Contextual Dialog not showing up after app reinstallation.](https://trello.com/c/4tRT1GNq/994-contextual-dialog-not-showing-up-after-app-reinstallation)

PR #439, opened 2018-10-01
Reset Dialog context after logout

Scollaco:
# What
Reset the value of `.login` Push notification dialog after logout.

# Why
There's a bug where after logging in, if the users tapped `Not now` on the Contextual Push Notification, they would never see the Dialog prompt again, even after logging out. This could prevent them to, in the future, opt to receive push notifications.

# How
After logging out, we remove `.login` from the `deniedNotificationContexts` array (on UserDefaults).

GIF


Scollaco:
[Contextual Dialog not showing up after app reinstallation.](https://trello.com/c/4tRT1GNq/994-contextual-dialog-not-showing-up-after-app-reinstallation)

Library/PushNotificationdialogTests.swift
```diff
@@ -81,4 +81,18 @@ class PushNotificationDialogTests: XCTestCase {
       XCTAssertEqual(["login", "save"], userDefaults.deniedNotificationContexts )
     }
   }
+
+  func testLoginContextResets_AfterLogout() {
+
+    withEnvironment(userDefaults: userDefaults) {
+
+      PushNotificationDialog.didDenyAccess(for: .login)
+      PushNotificationDialog.didDenyAccess(for: .save)
```
ifbarrera: Shouldn't we also reset the `.save` context? Since "saving" is a logged-in action.

Scollaco: The prompt should appear within the `.save` context only for the first ever project saved. The initial requirement was to show the Contextual Dialog up to 3x to avoid overwhelming our users with alerts popping up on the screen. But this is a good question. I will check w/ @dannyalright and see what would be the best UX.

dannyalright: yeah let's chat in-person @Scollaco (and @ifbarrera if you'd like to join)

PR #440, opened 2018-10-01
Introducing Graph mutations

ifbarrera:
# What

Introducing a new pattern for adding Graph mutations.

Also in this PR:
- cleaned up the file size of `Service.swift` by moving `ServiceType` conformance to a new file and extension. I moved all of the functions required for `ServiceType` to this file, so we now have a dedicated file for adding new `ServiceType` requests.
- fixed a bug where the `ServerConfig` environment variables weren't being properly restored from ubiquitous storage

# Why

Needed to apply mutations for usage with GraphQL.

# How

We now have two new protocols: `GraphMutation` and `GraphMutationInput`.

`GraphMutation` consists of one `input` variable which should conform to `GraphMutationInput`. `GraphMutation` must also conform to `CustomStringConvertible`. If we look at the example `ChangePasswordMutation`, we can see the description of any `GraphMutation`-conforming object contains the actual mutation string. Mutations should always accept an input object conforming to `GraphMutationInput`.

We can also see a new function `applyMutation` that accepts a generic `GraphMutation` and returns a `SignalProducer` generic on `A`, and a `GraphError`. This is similar to how our `fetch` request is structured. To see an example of the usage of `applyMutation`, we can take a look at the `changePassword` function:

```
  public func changePassword(input: ChangePasswordInput) ->
    SignalProducer<GraphMutationEmptyResponseEnvelope, GraphError> {
    return applyMutation(mutation: ChangePasswordMutation(input: input))
  }
```

We pass the `changePassword` function our input, and we create the mutation and pass it to the `applyMutation` function. This function goes on to prepare a graph mutation request, and execute the request.

Lastly, we are now parsing Graph errors being returned. `GraphResponse` has been modified to optionally parse an array of `GraphResponseError`s, which simply contain a message that we can display to the user*

*this is still under discussion pending concerns about error message localization

# TODO ⏰

- [ ] confirm handling of server-generated errors


KsApi/Service+ServiceType.swift
```diff
@@ -0,0 +1,452 @@
+import Foundation
```
Scollaco: I like the idea of making the `Service` class cleaner by creating an extension. I would move the generic functions to the extension instead of the api request functions, though. This would keep the `Service` and `MockService` classes with an identical structure, making the navigation through the code more intuitive.

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -127,6 +121,24 @@ final class ChangePasswordViewController: UIViewController {
...
 
+    self.viewModel.outputs.testPasswordInput
+      .observeValues {
+        print("CHANGE PASSWORD SUCCESS")
+    }
+
+    self.viewModel.outputs.changePasswordFailure
+      .observeForControllerAction()
+      .observeValues { [weak self] errorMessage in
+        let alert = UIAlertController(title: "Error",
```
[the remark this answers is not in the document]
ifbarrera: This is temporary, it'll be removed in the main `change-password` PR in favor of the message banner

KsApi/Service+ServiceType.swift
ifbarrera: Good point! Done

KsApi/ServiceTypeTests.swift
```diff
@@ -211,6 +211,26 @@ final class ServiceTypeTests: XCTestCase {
...
+                                                         queryString: "mutation",
+                                                         input: ["mutation_input": 123])
+
+    let jsonBody = try? JSONSerialization.jsonObject(with: request?.httpBody ?? Data(capacity: 1),
+    options: [])
+
+    XCTAssertNotNil(request)
+    XCTAssertEqual(request?.httpMethod, "POST")
+    XCTAssertEqual(request?.allHTTPHeaderFields?["Content-Type"], "application/json; charset=utf-8")
+    XCTAssertNotNil(jsonBody)
```
justinswart: Should we test the json body as a string?

KsApi/mutations/Mutations.swift
```diff
@@ -0,0 +1,19 @@
...
+public struct ChangePasswordMutation: GraphMutation {
+  var input: ChangePasswordInput
+
+  public init(input: ChangePasswordInput) {
+    self.input = input
+  }
+
+  public var description: String {
+    return """
+    mutation UpdateUserPassword($input: UpdateUserAccountInput!) {\
```
justinswart: I think mutations and queries are lowercase in the schema, so `updateUserPassword` right?

ifbarrera:
@justinswart @Scollaco addressed your comments in the new PR against master here: https://github.com/kickstarter/ios-oss/pull/442

PR #443, opened 2018-10-09
Custom message banner

ifbarrera:
# What

A message banner view controller that can be embedded into any VC and configured as a success, failure, or info banner with a custom message.

# Why

Messaging to the user when something has happened.

# How

`MessageBannerViewController` can be embedded as a child view controller to any VC. Here it's been embedded into `ChangePasswordViewController` and `ChangeEmailViewController`.  The container VC's view should be positioned on the bottom of the view, since the banner is configured to appear from below the frame. To show the banner, simply configure it using the `setBannerType(type: message:)` and then call `showBannerView`. The banner view will automatically hide itself after 4 seconds, or it will dismiss if the user interacts with it either with a tap, or a panning gesture.

# See 👀

Two ways to dismiss: either by waiting 4 seconds, or by tapping or dragging the view.

xw4ukppsqu

7xdh9ytk4n


Kickstarter-iOS/ViewModels/MessageBannerViewModel.swift
```diff
@@ -0,0 +1,103 @@
+import Foundation
+import Prelude
+import ReactiveSwift
+import Result
+
+protocol MessageBannerViewModelOutputs {
```
Scollaco: We usually place the `inputs` above the `outputs`.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+         |> UILabel.lens.textAlignment .~ textAlignment
+    }
+  }
+
+  func setBannerType(type: MessageBannerType, message: String) {
+    self.viewModel.inputs.setBannerType(type: type)
+    self.viewModel.inputs.setBannerMessage(message: message)
+  }
+
+  func showBannerView() {
```
Scollaco: Since we always call `showBannerView()` after setting its type, what if we create a function 
```
func showBannerType(_ type: MessageBannerType, message: String) {
    self.viewModel.inputs.setBannerType(type: type)
    self.viewModel.inputs.setBannerMessage(message: message)
    self.viewModel.inputs.showBannerView(shouldShow: true)
}
``` 
?

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+    self.viewModel.inputs.showBannerView(shouldShow: true)
+  }
+
+  private func showViewAndAnimate(_ isHidden: Bool) {
+    let duration = isHidden ? AnimationConstants.hideDuration : AnimationConstants.showDuration
+
+    if !isHidden {
```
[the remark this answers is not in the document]
ifbarrera: No because when `isHidden` is true, we don't want to hide it before the animation has completed. Only in the scenario where `isHidden` is false, do we want the view's visibility to be updated before the animation starts.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
ifbarrera: Updated to `showBanner(with type: MessageBannerType, message: String)` 😃

PR #445, opened 2018-10-12
Emoji logging ✨

eoji:
# What
**_Xcode_** displays a giant white blog of text for logs...They'd be a lot nicer if there was some color indication of what's going on.

# Why
To help with debugging 🕵🏾‍♀️

# How
EMOJI EVERYWHERE 🙏🏾🤓⚡️
⚪️Request is starting
🔵Request was successful
🔴Request was unsuccessful
🐨Koala

# See
screen shot 2018-10-12 at 5 14 16 pm


dusi:
It's `Xcode` not `XCode` :trollface:

KsApi/Service+RequestHelpers.swift
```diff
@@ -51,10 +51,12 @@ extension Service {
       let task = URLSession.shared.dataTask(with: request) {  data, response, error in
         if let error = error {
           observer.send(error: .requestError(error, response))
+          print("🔴 [KsApi] Failure \(error.localizedDescription)")
```
dusi: I know @justinswart has mentioned this in the Slack room but though I'll point it out again - there's this thing called 
`os_log` which defines various levels of logging which makes it more flexible to filter through in `Xcode.app` or `Console.app`. Logging network calls seems like a great example of `Debug` level logging.
 
https://medium.com/back-market-engineering/discovering-using-oslog-output-console-with-swiftybeaver-on-ios-54d7bfbe3b97

If you're not considering various level at this moment I'm sure `print` will do just fine!

KsApi/Service+RequestHelpers.swift
```diff
@@ -51,10 +51,12 @@ extension Service {
...
           return
         }
 
         guard let data = data else {
+          print("🔴 [KsApi] Failure emptyResponse")
```
dusi: Should we make this more human readable `🔴 [KsApi] Failure - Empty response`?

KsApi/Service+RequestHelpers.swift
```diff
@@ -63,10 +65,13 @@ extension Service {
           let decodedObject = try JSONDecoder().decode(GraphResponse<A>.self, from: data)
           if let errors = decodedObject.errors, let error = errors.first {
             observer.send(error: .decodeError(error))
+            print("🔴 [KsApi] Failure decodeError \(error.message)")
```
dusi: and separate hardcoded / dynamic values like `🔴 [KsApi] Failure - Decoding error: \(error.message)`?

KsApi/Service+RequestHelpers.swift
eoji: I was asking @Scollaco about this and he let me know `os_log` is for iOS 10 and above, so if I wanted to add it, I'd have to do `if #available` every place I log. So he recommended we use `print` for now and `os_log` when we update our app minimum to iOS 10. But I'm happy to add it if you feel strongly!

justinswart: We're still supporting iOS 9?

dusi: I'm sorry Izzy..I had no idea we still support iOS 9

KsApi/Service+RequestHelpers.swift
```diff
@@ -51,10 +51,12 @@ extension Service {
       let task = URLSession.shared.dataTask(with: request) {  data, response, error in
         if let error = error {
           observer.send(error: .requestError(error, response))
+          print("🔴 [KsApi] Failure - Request error \(error.localizedDescription)")
```
dusi: Is this missing colon? `🔴 [KsApi] Failure - Request error: \(error.localizedDescription)`

KsApi/Service+RequestHelpers.swift
eoji: @justinswart yeah 😬 

@dusi Nino said it's on the backlog to update our minimum OS 🚀

eoji:
@dusi Nino approved but I wondered if you had any other comments!

PR #449, opened 2018-10-16
Currency picker alert

cdolm92:
# What

Now presenting alert that confirms a currency change by tapping Yes, change currency".

# How

This only sets up the currency picker alert. It is presented by the `SettingsAccountViewController` using the  `settingsCurrencyPickerCellDidChangeCurrency(_ currency: Currency)`  delegate method. The SettingsCurrencyCell updates a the currency text with a notification. 

# See 👀

currency-picker-alert

# TODO ⏰

- [ ] PR Feedback
- [x] Graph Mutation


cdolm92:
This is ready for review. There is a test failing for change password. If ya'll have any insight or feedback as to why please share, thank you!

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
```diff
@@ -0,0 +1,54 @@
...
+  @IBOutlet fileprivate weak var titleLabel: UILabel!
+  @IBOutlet fileprivate weak var currentCurrencyLabel: UILabel!
+  @IBOutlet fileprivate weak var separatorView: UIView!
+
+  private let viewModel: SettingsCurrencyCellViewModelType = SettingsCurrencyCellViewModel()
+
+  func configureWith(value: SettingsCellValue) {
+    switch value.cellType {
+    case SettingsAccountCellType.currency:
+      NotificationCenter.default
```
ifbarrera: I think there might be a way to do this without notifications. What if the `dataSource` passed the value of the currency in as part of the `SettingsValueCellValue`?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -54,6 +62,14 @@ final class SettingsAccountViewController: UIViewController {
       .observeValues { [weak self] (viewController) in
         self?.navigationController?.pushViewController(viewController, animated: true)
     }
+
+    self.viewModel.outputs.updateCurrency
+      .observeForUI()
+      .observeValues { selectedCurrency in
+        NotificationCenter.default.post(name: Notification.Name.ksr_updatedCurrencyCellDetailText,
```
ifbarrera: Here instead of posting a notification you could call `self?.dataSource.configureCurrencyRow(currency: selectedCurrency)` and either reload the table view or just the currency cell.

KsApi/models/lenses/UserCurrencyLenses.swift
```diff
@@ -0,0 +1,10 @@
+import Prelude
+
+extension UserCurrency {
+  public enum lens {
```
ifbarrera: Can we not just use `keyPath` lenses instead? All you have to do is update `let chosenCurrency: String` to `public private(set) var chosenCurrency: String`

KsApi/mutations/UpdateUserProfileMutation .swift
```diff
@@ -0,0 +1,19 @@
...
+public struct UpdateUserProfileMutation<T: GraphMutationInput>: GraphMutation {
+  var input: T
+
+  public init(input: T) {
+    self.input = input
+  }
+
+  public var description: String {
+    return """
+    mutation updateUserCurrency($input: UpdateUserProfileInput!) {
```
ifbarrera: Let's rename this mutation to `updateUserProfile` for consistency with the name of the swift `struct`

KsApi/mutations/mutation inputs/ChangeCurrencyInput.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
```
ifbarrera: Is this file a duplicate from the one located at `KsApi/models/ChangeCurrencyInput.swift`?

Library/Currency.swift
```diff
@@ -1,54 +1,71 @@
...
-  case poundSterling
-  case hongKongDollar
-  case yen
-  case mexicanPeso
-  case norwegianKrone
-  case newZealandDollar
-  case swedishKrona
-  case singaporeDollar
-  case usDollar
+public enum Currency: String, Encodable {
```
ifbarrera: Why remove the `CaseIterable` conformance?

Library/ViewModels/SettingsCurrencyCellViewModel.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    let fetchedCurrency = initialUser.signal
+      .switchMap { _ in
+        return AppEnvironment.current.apiService
+          .fetchGraphCurrency(query: UserQueries.chosenCurrency.query)
+          .materialize()
+      }
+
+    let chosenCurrency = fetchedCurrency.values().map {
+      Currency(rawValue: $0.me.chosenCurrency ?? "")?.descriptionText ?? "" }
```
ifbarrera: Do we want to default it to `USD` if the user doesn't have a currency set? 🤔 I think that's what the web does.

KsApi/mutations/mutation inputs/ChangeCurrencyInput.swift
cdolm92: It shouldn't be. I moved it to `KsApi/mutations/mutation inputs` because it looks like that's where we're keeping the inputs.

Kickstarter-iOS/ViewModels/ChangePasswordViewModelTests.swift
```diff
@@ -151,7 +151,7 @@ final class ChangePasswordViewModelTests: TestCase {
 
       scheduler.advance()
 
-      self.changePasswordFailureObserver.assertValues(["Error changing password"])
+      self.changePasswordFailureObserver.assertValueCount(1)
```
cdolm92: made this change because the test was not passing on my branch but on master it was. would appreciate feedback on this. thank you!

cdolm92:
@kickstarter/native-squad g2g?

Kickstarter-iOS/ViewModels/SettingsCurrencyPickerCellViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+  var outputs: SettingsCurrencyPickerCellViewModelOutputs { get }
+}
+
+final class SettingsCurrencyPickerCellViewModel: SettingsCurrencyPickerCellViewModelOutputs,
+SettingsCurrencyPickerCellViewModelInputs, SettingsCurrencyPickerCellViewModelType {
+
+  public init() {
+    self.notifyCurrencyPickerCellRemoved = self.selectedCurrencyProperty.signal.ignoreValues()
+
+    self.showCurrencyChangeAlert = self.selectedCurrencyProperty.signal.skipNil()
```
ifbarrera: Can we combine these two outputs into just one? It seems that `notifyCurrencyPickerCellRemoved` is probably not needed anymore.

Kickstarter-iOS/ViewModels/SettingsCurrencyPickerCellViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  fileprivate let selectedCurrencyProperty = MutableProperty<Currency?>(nil)
+  public func didSelectCurrency(currency: Currency) {
+    self.selectedCurrencyProperty.value = currency
+  }
+
+  fileprivate let cellTypeProperty = MutableProperty<SettingsAccountCellType?>(nil)
```
ifbarrera: It also seems like this input is no longer needed since we don't read from `cellTypeProperty` anywhere anymore.

Kickstarter-iOS/Views/Cells/SettingsCurrencyPickerCell.swift
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
+    return Currency.allCases.count
+  }
+}
+
+extension SettingsCurrencyPickerCell: UIPickerViewDelegate {
+  func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+    return Currency.allCases[row].descriptionText //Currency(rawValue: row)?.descriptionText
```
ifbarrera: Do we need this comment?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -53,18 +80,71 @@ final class SettingsAccountViewController: UIViewController {
     _ = tableView
       |> settingsTableViewStyle
   }
+
+  func showCurrencyPickerCell(_ show: Bool) {
+    if show {
```
ifbarrera: Should we also add the `show == false` case here? Or remove the `show` parameter if it's not really needed.

Library/Currency.swift
```diff
@@ -0,0 +1,56 @@
...
+  case HKD
+  case JPY
+  case MXN
+  case NOK
+  case NZD
+  case SEK
+  case SGD
+  case USD
+
+  public static var rowHeight: CGFloat {
```
ifbarrera: Hm, this seems like more of a `View`-related property, can we move it to where it's used instead? Just to keep this `Currency` enum as more of a clean model object.

Library/DataSource/ValueCellDataSource.swift
```diff
@@ -136,6 +136,18 @@ open class ValueCellDataSource: NSObject, UICollectionViewDataSource, UITableVie
       return IndexPath(row: index, section: section)
   }
 
+  public final func deleteRow<Cell: ValueCell, Value: Any>(value: Value,
```
ifbarrera: Can we add a test for this in `ValueCellDataSourceTests`?

Library/SettingsAccountCellType.swift
```diff
@@ -75,12 +84,14 @@ public enum SettingsAccountCellType: SettingsCellTypeProtocol {
...
       return Strings.Currency()
+    case .currencyPicker:
+     return ""
     }
   }
 
   public var description: String? {
     switch self {
-    case .currency: return "$ Dollar (USD)"
+    case .currency: return Strings.Currency_USD()
```
ifbarrera: I think we can remove this since we're reading the currency from the `SettingsCurrencyCellValue` now

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -1,23 +1,28 @@
...
 
 public protocol SettingsAccountViewModelOutputs {
-
-  var reloadData: Signal<Void, NoError> { get }
+  var dismissCurrencyPicker: Signal<Void, NoError> { get }
+  var presentCurrencyPicker: Signal<Bool, NoError> { get }
+  var reloadData: Signal<(User, Currency), NoError> { get }
+  var showAlert: Signal<(), NoError> { get }
   var transitionToViewController: Signal<UIViewController, NoError> { get }
+  var updateCurrency: Signal<Currency, NoError> { get }
```
ifbarrera: I think we can remove this output since we're using `reloadData` to signal when the currency has been updated.

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -53,18 +80,67 @@ final class SettingsAccountViewController: UIViewController {
     _ = tableView
       |> settingsTableViewStyle
   }
+
+  func showCurrencyPickerCell() {
```
ifbarrera: Can we make all these helper functions `private`?

Library/SettingsAccountCellType.swift
ifbarrera: Can we remove this case?

Library/ViewModels/SettingsAccountViewModelTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    self.vm.outputs.updateCurrencyFailure.observe(self.updateCurrencyFailure.observer)
+  }
+
+  func testReloadData() {
+    self.vm.inputs.viewDidLoad()
+    self.reloadDataUser.assertValueCount(1)
+    self.reloadDataCurrency.assertValueCount(1)
+    self.vm.inputs.showChangeCurrencyAlert(for: Currency.CHF)
+    self.vm.inputs.didConfirmChangeCurrency()
+    self.scheduler.advance()
```
ifbarrera: Can we assert here that `reloadDataUser` and `reloadDataCurrency` have a value count of 2?

PR #452, opened 2018-10-22
New card screen ui

cdolm92:
# What

UI for new credit card.

# Why

Users will be able to add a new credit card on mobile.

# See 👀

oct-22-2018 19-03-31

# TODO ⏰

- [ ] Strings


cdolm92:
@ifbarrera @Scollaco @justinswart @dusi added the placeholder strings, would appreciate feedback on this. thanks!🙂

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+  @IBOutlet fileprivate weak var cardholderNameLabel: UILabel!
+  @IBOutlet fileprivate weak var cardholderNameTextField: UITextField!
+  @IBOutlet fileprivate weak var expirationLabel: UILabel!
+  @IBOutlet fileprivate weak var expirationTextField: UITextField!
+  @IBOutlet fileprivate weak var securityCodeLabel: UILabel!
+  @IBOutlet fileprivate weak var securityCodeTextField: UITextField!
+  @IBOutlet fileprivate weak var zipCodeLabel: UILabel!
+  @IBOutlet fileprivate weak var zipCodeTextField: UITextField!
+
+  private weak var saveButtonView: LoadingBarButtonItemView!
```
justinswart: Is there a reason we're holding a reference to this view?

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    let cancelButton = UIBarButtonItem(title: Strings.Cancel(),
+                                       style: .plain,
+                                       target: self,
+                                       action: #selector(cancelButtonTapped))
+    cancelButton.tintColor = UIColor.ksr_green_700
```
justinswart: Can just infer the type here: `cancelButton.tintColor = .ksr_green_700`

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -14,6 +14,7 @@ public protocol SettingsAccountViewModelOutputs {
 
   var reloadData: Signal<Void, NoError> { get }
   var transitionToViewController: Signal<UIViewController, NoError> { get }
+  var goToAddCard: Signal<Void, NoError> { get }
```
justinswart: Did we want to add a test for this output?

cdolm92: Not yet, this output is actually going to be removed  and go in the payments method screen when we get to that. It's only here to get to the `Add New Card` screen. We can actually remove this?

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
cdolm92: We need a save button on this screen and I implemented similarly to `ChangePasswordViewController` and `ChangeEmailViewController`. I can also remove this and revisit the functionality when I start on the implementation(View Model).

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    self.saveButtonView = LoadingBarButtonItemView.instantiate()
+    self.saveButtonView.setTitle(title: Strings.Save())
+    self.saveButtonView.addTarget(self, action: #selector(saveButtonTapped))
+    let navigationBarButton = UIBarButtonItem(customView: self.saveButtonView)
+    self.navigationItem.setRightBarButton(navigationBarButton, animated: false)
+  }
+
+  override func bindStyles() {
+    super.bindStyles()
```
Scollaco: I think here is a good opportunity to adopt keyPath instead of lenses. So, for example: 
`||> UITextField.lens.textAlignment .~ .right` would become `||> \UITextField.textAlignment .~ .right`, or the shorthand: `||> \.textAlignment .~ .right` and so on...

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+import Prelude
+import ReactiveSwift
+import UIKit
+
+internal final class AddNewCardViewController: UIViewController {
+  @IBOutlet fileprivate weak var cardNumberLabel: UILabel!
+  @IBOutlet fileprivate weak var cardNumberTextField: UITextField!
+  @IBOutlet fileprivate weak var cardholderNameLabel: UILabel!
+  @IBOutlet fileprivate weak var cardholderNameTextField: UITextField!
+  @IBOutlet fileprivate weak var expirationLabel: UILabel!
```
Scollaco: I know it's not part of the scope, but i wonder if it would be a good idea to limit the number of characters in `5` or `7` for this textfield and create a mask for it. So after adding the first 2 digits, let's say `11`, the textField would automatically append a `\` character, and adding the year `23` (or `2023`), the textfield would show `11/23` or `11/2023`. We can't talk about it with @dannyalright and I can create an UITextField extension to handle that.

PR #453, opened 2018-10-23
Clean up target membership

dusi:
# What

Looks like couple cells were checked in under wrong target and also found an unused piece of code.

screen shot 2018-10-23 at 5 49 42 pm

# Why

Part of my on-boarding to `iOS OSS` codebase plus it's always good to clean up & remove unused code.

# How

Simply check cells under correct target.

# Acceptance criteria
- [ ] On iOS 11 enabling `Darken Colors` under `General > Accessibility > Increase Contrast` and opening project detail from the feed will NOT crash the app

#wcpgw
giphy


Library/UITableViewCell+SeparatorCatch.swift
```diff
@@ -1,8 +0,0 @@
-import UIKit
-
-@available(iOS 11.0, *)
-extension UITableViewCell {
-  func _setSeparatorDrawsInVibrantLightMode(_ value: Bool) {
```
dusi: Doesn't seem like this is used anywhere else in the codebase. Hope I'm not missing something?

justinswart: Ah this was a totally weird bug @Scollaco and I caught a while back - related PR is #272. We think that this might also have been solved by removing a dangling `IBOutlet`. I'd say let's try reproduce and see if it's good to remove this 👍

dusi: So it looks like #272 haven't really fixed the crashing issue completely and it was #273 that actually helped. Hence removing this extension should not reintroduce the crash.

Ways that I've tried reproducing this crash was to build the following

Environment:  Xcode 9
iOS: 11.4

1. Checkout #272 
- Build & Run
- Enable `Darken Colors` under `General > Accessibility > Increase Contrast`
- Crashes

2. Checkout #273
- Comment out this extension
```
//@available(iOS 11.0, *)
//extension UITableViewCell {
//  func _setSeparatorDrawsInVibrantLightMode(_ value: Bool) {
//
//  }
//}
```
- Build & Run
- Enable `Darken Colors` under `General > Accessibility > Increase Contrast`
- Does NOT crash 😎 & proves that #273 has fixed the issue

dusi:
@Scollaco would you please be able to run this on device with iOS 11.4? (we currently don't have any in YVR)

Library/UITableViewCell+SeparatorCatch.swift
dusi: I've ran another round of validation and did following:

1. Checkout #272 
2. Comment out the extension
```
//@available(iOS 11.0, *)
//extension UITableViewCell {
//  func _setSeparatorDrawsInVibrantLightMode(_ value: Bool) {
//
//  }
//}
```
3. Remove separator view from `ProjectPamphlet.storyboard` (single line change as demonstrated in #273 - https://github.com/kickstarter/ios-oss/pull/273/files#diff-7bd3076ab44e90789ac8a37211e24aa2L520)
>> Does NOT crash

justinswart:
@dusi nice! I'm sure it's the separator that was crashing it and the extension workaround just intercepted the erroneous message sending.

justinswart:
[Clean up target membership](https://trello.com/c/yOvRjpGG/1038-clean-up-target-membership)

PR #454, opened 2018-10-24
User currency displays on project page

ifbarrera:
# What

The project page now displays currency values in the user's preferred currency. 💲 

# Why

Support for user preferred currency.

# How

Two things to note:
- we are now taking into account the user's preferred currency when determining whether a value needs conversion or not. If the user's preferred currency is different from the project's currency, the value will be converted (`project.stats.currency != project.stats.currentCurrency`)
- if the user's preferred currency is `USD` (or undefined, which defaults to `USD`) *and* the user's location is in the US, then we only display the `$` symbol instead of `US$`
- we are now receiving the user's location from `config.countryCode`, and the `Environment`'s `countryCode` now reads from that value instead of being hard-coded to `US`

# See 👀

For screenshots of all possible combinations, please see images in the `Kickstarter_Framework_iOSTests.ProjectPamphletContentViewControllerConversionTests`.

test_usproject_nonususer_nonuslocation_projectlocation_us_usercurrency_sek_userlocation_se 2x


ifbarrera:
[Currency Preferences on Project Pages](https://trello.com/c/rtycrU1H/997-currency-preferences-on-project-pages)

Library/SharedFunctions.swift
```diff
@@ -83,17 +83,21 @@ internal func minPledgeAmount(forProject project: Project, reward: Reward?) -> D
...
+                           env: Environment = AppEnvironment.current) -> String {
 
-  guard AppEnvironment.current.launchedCountries.currencyNeedsCode(country.currencySymbol) else {
+  guard env.launchedCountries.currencyNeedsCode(country.currencySymbol) else {
     // Currencies that dont have ambigious currencies can just use their symbol.
     return country.currencySymbol
   }
 
-  if country == .us && AppEnvironment.current.countryCode == "US" {
+  if country == .us && env.countryCode == "US" && omitCurrencyCode {
```
justinswart: Is this comparison preferred over `env.countryCode == Project.Country.us.countryCode`?

Library/ViewModels/ProjectPamphletMainCellViewModel.swift
```diff
@@ -354,6 +340,51 @@ private func fundingStatus(forProject project: Project) -> String {
   }
  }
 
+private func pledgeAmountAndGoalAndCountry(forProject project: Project,
+                                           needsConversion: Bool) -> (Int, Int, Project.Country) {
```
justinswart: Would a `typealias` be useful here? Perhaps something like:

```swift
typealias PledgedAmountData = (pledged: Int, goal: Int, country: Project.Country)
```

This would improve readability when referring to the tuple members later on:
`pledgedAmountData.goal` vs `pledgedGoalCountry.1`

KsApi/models/Project.swift
```diff
@@ -90,6 +95,29 @@ public struct Project {
...
+      guard let currentCurrency = self.currentCurrency else {
+        return nil
+      }
+
+      return Project.Country(currencyCode: currentCurrency)
+    }
+
+    /// Omit US currency code
+    public var omitUSCurrencyCode: Bool {
+      let currentCurrency = self.currentCurrency ?? "USD"
```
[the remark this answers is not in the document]
ifbarrera: No reason, I'll update 👍

Library/SharedFunctions.swift
ifbarrera: We can do that no problem 👍

PR #451, opened 2018-10-22
Silence warnings in Xcode 10

justinswart:
# What

Just silencing a bunch of warnings in the app.

# Why

There were a few: 
screen shot 2018-10-22 at 2 54 31 pm

Now there are fewer:
screen shot 2018-10-22 at 2 52 44 pm

I wasn't able to silence all of them - any ideas? Not sure if we want to remove the compile time warning 🤔, that's a Swift flag.

screen shot 2018-10-22 at 2 53 03 pm

# How

Updated dependencies:
- Argo
- Curry
- Prelude
- ReactiveExtensions
  - ReactiveSwift
- FBSnapshotTestCase

Removed a couple deprecations.

# TODO ⏰
- [x] Open related PR's for Prelude and ReactiveExtensions and reference them here.
- [x] Merge https://github.com/kickstarter/Kickstarter-Prelude/pull/89
- [x] Merge https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/70

PR #455, opened 2018-10-25
Fix provisioning regression

justinswart:
# What

Fixed a build regression introduced in #451 and tidied up some provisioning settings.

# Why

I removed the value `$(SYMROOT)/Release$(EFFECTIVE_PLATFORM_NAME)` from `Framework search paths` thinking it was not being used for anything. The app also built just fine without it. However, it [turns out](https://stackoverflow.com/questions/26073088/swift-frameworks-do-not-work-with-build-configurations-named-other-than-debug#35627389) that this _is_ necessary if your scheme names are anything other than `Debug` or `Release` which ours of course are for `Hockey Alpha` and `Hockey Beta`.

Also set all of our framework provisioning settings to be the same (none/none) because there were some inconsistencies:

screen shot 2018-10-25 at 2 51 30 pm

# How

Added `$(SYMROOT)/Release$(EFFECTIVE_PLATFORM_NAME)` back to `Framework search paths`, updated provisioning settings.


justinswart:
[Fix build regression](https://trello.com/c/MmZl18Cd/1040-fix-build-regression)

justinswart:
Tested on CircleCI:
screen shot 2018-10-25 at 4 22 22 pm

PR #457, opened 2018-10-26
Payment methods

Scollaco:
# What
Payment methods screen. It currently lists all saved credit cards, but the `Edit` functionality is not implemented yet..

# Why
This is part of the Settings v.2

# See 👀

# TODO ⏰
- [x] Final translations

GIF


dannyalright:
nice! Margins look a bit off. Want to pair on Monday quickly to fix that?

Scollaco:
> nice! Margins look a bit off. Want to pair on Monday quickly to fix that?

It sounds good.

Scollaco:
[User can see Payment Methods screen](https://trello.com/c/xkrMMlGY/952-user-can-see-payment-methods-screen)

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+}
+
+extension PaymentMethodsViewController: UITableViewDelegate {
+
+  func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+    return 0.1
+  }
+
+  func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+    return 80
```
ifbarrera: Same here, lets use `Styles.grid` for consistent sizing.

KsApi/GraphSchema.swift
```diff
@@ -193,6 +193,12 @@ public enum Query {
     }
   }
 
+  public enum ProjectsConnection {
```
ifbarrera: What's this for? 👀

KsApi/GraphSchema.swift
```diff
@@ -400,13 +414,25 @@ extension Query.User: QueryType {
...
+    case let .storedCards(args, fields):        return "storedCards\(connection(args, fields))"
     case .slug:                                 return "slug"
     case .url:                                  return "url"
     case .userId:                               return "uid"
     }
   }
 }
 
+extension Query.User.CreditCard: QueryType {
+  public var description: String {
```
ifbarrera: If you make `CreditCard` be of type `String`, then here you can just do `self.rawValue`

KsApi/models/UserCreditCard.swift
```diff
@@ -0,0 +1,19 @@
...
+  public private(set) var storedCards: CreditCardConnection
+
+  public struct CreditCard: Swift.Decodable, Equatable {
+    public private(set) var expirationDate: String
+    public private(set) var id: String
+    public private(set) var lastFour: String
+    public private(set) var type: String
+
+    public var formattedExpirationDate: String {
+      return String(expirationDate.dropLast(3))
```
ifbarrera: Hmm I'm not sure we should be doing date formatting in this way, if something changes with the order of how the date is returned from the API, this could break. The safest way to do this would be to convert the String we get into a Date, and then use a `DateFormatter` to format the date into a correctly formatted string.

ifbarrera: Oh now I see it's formatted in the view model. But is it necessary to truncate the string?

Library/ViewModels/CreditCardCellViewModel.swift
```diff
@@ -0,0 +1,87 @@
...
+private func toDate(dateString: String) -> Date {
+
+  guard let date = dateFormatter.date(from: dateString) else {
+    fatalError("Unable to parse date format")
+  }
+  return date
+}
+
+private let dateFormatter: DateFormatter = {
+  let dateFormatter = DateFormatter()
```
ifbarrera: It might be worthwhile to cache this date formatter using `DateFormatter.cachedFormatter(forConfig:)` so that this formatter is not re-created for every credit card cell.

justinswart:
@Scollaco I think you may just need to merge master or rebase this - there seem to be changes to submodules etc that are probably not part of this PR 👍

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+  public static func instantiate() -> PaymentMethodsViewController {
+    return Storyboard.Settings.instantiate(PaymentMethodsViewController.self)
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.viewModel.inputs.viewDidLoad()
+    self.navigationItem.rightBarButtonItem =
+      UIBarButtonItem(title: Strings.discovery_favorite_categories_buttons_edit(),
```
justinswart: I think a better way to get this to fit is:

```swift
self.navigationItem.rightBarButtonItem = UIBarButtonItem(
  title: Strings.discovery_favorite_categories_buttons_edit(),
  style: .plain,
  target: self,
  action: nil
)
```

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+import UIKit
+
+internal final class PaymentMethodsViewController: UIViewController {
+
+  private let dataSource = PaymentMethodsDataSource()
+  private let viewModel: PaymentMethodsViewModelType = PaymentMethodsViewModel()
+
+  @IBOutlet private weak var headerLabel: UILabel!
+  @IBOutlet private weak var tableView: UITableView! {
+    didSet {
```
justinswart: I kinda like this way of in-lining the configuration of things. Just wondering if we've started doing this anywhere else? It just kind've break consistency.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    self.viewModel.outputs.goToAddCardScreen
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.goToAddCardScreen()
+    }
+  }
+
+  private func goToAddCardScreen() {
+    // Push AddNewCardScreen from here.
```
justinswart: If we're adding this in a separate PR then I'd suggest a `//fixme:` so that it's easy to find and to also base this PR off a feature branch.

Kickstarter-iOS/Views/PaymentMethodsFooterView.swift
```diff
@@ -0,0 +1,39 @@
+import Library
+import Prelude
+import UIKit
+
+internal protocol PaymentMethodsFooterViewDelegate: class {
+  func didTapAddNewCardButton()
```
justinswart: I think we should stick a convention for our delegate methods. They should always describe the object that's triggering the action as well as pass in that object and any additional values, something like:

```swift
func paymentMethodsFooterViewDidTapAddNewCardButton(_ footerView: PaymentMethodsFooterView)
```

More here: https://google.github.io/swift/#delegate-methods

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -0,0 +1,55 @@
...
+import Result
+
+public protocol PaymentMethodsViewModelInputs {
+  func viewDidLoad()
+  func didTapAddNewCardButton()
+}
+
+public protocol PaymentMethodsViewModelOutputs {
+  /// Emits the user's stored cards
+  var didFetchPaymentMethods: Signal<[GraphUserCreditCard.CreditCard], NoError> { get }
```
justinswart: This output should just be called `paymentMethods`, I think from an interface perspective we aren't concerned with whether they were fetched, just that this emits values.

KsApi/GraphSchema.swift
Scollaco: Oh, this is a result of a rabbit hole I was in. 😅 Will remove this.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
Scollaco: Yeah, I like this too way of configuring views too. But will change it to keep consistency.

KsApi/models/UserCreditCard.swift
Scollaco: Before implementing this, I confirmed with the API that the format wouldn't change. The `dropLast` function is just getting rid of the day returned by graphQL. The cards expiration date usually has the format `MM/yyyy` and the day returned is always `01`, making the expiration date wrong.

Scollaco:
@ifbarrera , @justinswart g2g?

KsApi/models/UserCreditCard.swift
```diff
@@ -0,0 +1,19 @@
+public struct GraphUserCreditCard: Swift.Decodable {
+
+  public private(set) var storedCards: CreditCardConnection
+
+  public struct CreditCard: Swift.Decodable, Equatable {
```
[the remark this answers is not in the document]
Scollaco: Actually the `expirationDate` returned by graphQL is a `String` with the format `yyyy-MM-dd`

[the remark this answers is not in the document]
justinswart: Oh really? That's weird - it has the type `Date!` 🤔 I was sure those were scalar `Double`s. So much for standardization! 😄Ok, well thoughts on the `enum`?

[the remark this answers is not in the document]
Scollaco: I just pushed a code with the `type` converted to `enum`. It was actually a good idea. 
Now about make it a global model, can we talk about it?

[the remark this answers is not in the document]
justinswart: Oh, we chatted on slack - but let's do that work as a separate PR when we have a requirement for `CreditCardType` outside of this model 👍

PR #458, opened 2018-10-26
Removed "Most funded" projects from Discovery

Scollaco:
# What
Removed "Most funded" projects from Discovery.

# Why
The projects listed within that sort rarely change, so we decided to remove it from the app.

# How
By deleting if from `DiscoverParams` and fixing the errors and tests.


Scollaco:
[Remove Most Funded from sorts](https://trello.com/c/PYwWNQ0W/1028-remove-most-funded-from-sorts)

PR #459, opened 2018-10-26
Ensure generated strings ordering consistency

justinswart:
# What

Ensures that each time we run `make strings` we maintain the ordering of the locales in the function documentation.

# Why

I've noticed a few PR's lately with huge amounts of changes (just like this one), e.g.:

screen shot 2018-10-26 at 12 33 27 pm

And upon closer inspection, noticed that this is due to changes to `Strings.swift`:

screen shot 2018-10-26 at 12 33 51 pm

This appears to be because of Swift 4.2 (correctly) offering no guarantee for the ordering of dictionary keys, which we use to construct the function documentation in `Strings.swift`.

# How

Updated `bin/swift` to remove deprecations and added the required ordering. Unfortunately this PR also contains a huge amount of changes but this should be last time these will be re-shuffled. I would then encourage everyone with open PR's to rebase on `master` once this is merged.

# TODO ⏰

I'd like to use an investment day to write unit tests for this script - for now I just wanted to mitigate the large PR's 😅


justinswart:
[Fix ordering in strings generation](https://trello.com/c/UB0Oc4A2/1044-fix-ordering-in-strings-generation)

PR #460, opened 2018-10-26
Save project optimistically

justinswart:
# What

Enables an optimistic UI for saving projects. When we save projects the heart will be toggled immediately and only flipped back if the call fails.

# Why

This creates a better user experience. Backers aren't interested in knowing that a call is taking a few seconds and the majority of these calls are likely to succeed.

# How

This was mostly achievable simply be removing the signals that cared about whether a network call was in flight. In addition I've debounced the toggling so that if a user taps the button on/off frequently it will wait 1 second before firing off the call to the back-end.

# See 👀

| Before (non-optimistic)  | After (optimistic) |
|  |  |


justinswart:
[Delay when saving a project](https://trello.com/c/RHDv4MtB/1037-delay-when-saving-a-project)

justinswart:
@Scollaco good question - the current error handling pattern hasn't changed so it looks like the current behaviour doesn't show an error on failure, it just toggles back to the previous state.

@ifbarrera did some testing and found that sometimes the state wasn't correct after tapping so I'm just looking into that a bit more. The debouncing might be causing it.

justinswart:
I've added a WIP label to this - it's a little trickier than I'd initially anticipated because we don't actually send a value through to the backend for saving, we only tell it to toggle. I'm going to spend a little more time looking into whether we can solve this without changes to the backend but we might always have inconsistencies trying to do that.

PR #461, opened 2018-10-29
Log custom events to Crashlytics

justinswart:
# What

Logs our tracking events to Crashlytics.

# Why

It should be useful for debugging purposes to have a log of events that occurred leading up to a crash. We tried to do this before in #217 but it never seemed to work (events just never appeared in Hockey crash logs). I've used Crashlytics for this before on previous projects and it worked pretty well.

# How

Replace Hockey log call with Crashlytics call.

# See 👀

https://docs.fabric.io/apple/answers/answers-events.html#custom-event

### DIFF
--- a/Library/Koala/KoalaTrackingClient.swift
+++ b/Library/Koala/KoalaTrackingClient.swift
@@ -57,7 +57,7 @@ public final class KoalaTrackingClient: TrackingClientType {
     print("🐨 [Koala Track]: \(event), properties: \(properties)")
 
     self.queue.async {
-      Answers.logCustomEvent(withName: event, customAttributes: properties)
+      Answers.logCustomEvent(withName: event, customAttributes: nil)
       self.buffer.append(["event": event, "properties": properties])
     }
   }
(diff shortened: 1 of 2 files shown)


Library/Koala/KoalaTrackingClient.swift
```diff
@@ -57,7 +57,7 @@ public final class KoalaTrackingClient: TrackingClientType {
     print("🐨 [Koala Track]: \(event), properties: \(properties)")
 
     self.queue.async {
-      BITHockeyManager.shared().metricsManager.trackEvent(withName: event)
+      Answers.logCustomEvent(withName: event, customAttributes: properties)
```
justinswart: While we _can_ pass `properties` through here, it might not be necessary and we might want to reduce the payload?

justinswart:
[Log custom events to Crashlytics](https://trello.com/c/GsJcIupa/1043-log-custom-events-to-crashlytics)

Library/Koala/KoalaTrackingClient.swift
Scollaco: Nice, @justinswart ! Yeah, we can see if the logs are enough without the properties and simply add them in the future in case we need it.

justinswart: Ok cool, will pass `nil` for properties for now.

PR #463, opened 2018-10-29
iOS haptic feedback

cdolm92:
# What

Haptic feedback when successfully backed a project and when user saves/unsaves a project.

# Why

User knows that backing and saving happened successfully. 

# TODO ⏰

- [ ] PR Feedback


dusi:
Would be happy to pair on this with you @cdolm92 !

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -320,6 +320,17 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
   }
 
   @objc fileprivate func saveButtonTapped() {
+    if self.saveButton.isSelected {
```
Scollaco: I think since we are using the same code for a UIButton to generate feedback in multiple places, we could extend `UIButton` and move the FeedbackGenerator code to a `generateFeedback()` function.
So whenever we need to use it, we would do it by simply using 
```
if #available(iOS 10.0, *) {
      self.myButton.generateFeedback()
    }
```

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -320,6 +320,15 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
...
+      if #available(iOS 10.0, *) {
+        self.saveButton.generateSelectionFeedback()
+      }
+    } else {
+      if #available(iOS 10.0, *) {
+        self.saveButton.generateSuccessFeedback()
+      }
+    }
     self.viewModel.inputs.saveButtonTapped()
```
dusi:
```

1. self.viewModel.inputs.saveButtonTapped(self.saveButton.isSelected)

2. saveButtonTapped
// On VM
!isSelected -> self.generateSelectionFeedback
isSelected -> self.generateSuccessFeedback

// On VC

self.viewModel.outputs.generateSelectionFeedback.subscribe {
  self.button.generateSelectionFeedback() // extension implementation
}

self.viewModel.outputs.generateSelectionFeedback.subscribe {
  self.button.generateSuccessFeedback() // extension implementation
}

```

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -320,6 +320,15 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
   }
 
   @objc fileprivate func saveButtonTapped() {
+    if self.saveButton.isSelected {
+      if #available(iOS 10.0, *) {
+        self.saveButton.generateSelectionFeedback()
+      }
+    } else {
+      if #available(iOS 10.0, *) {
```
justinswart: We could consider moving the `if #available(iOS 10.0, *) { ... }` checks to inside the extension, that way we're only checking in one place for whether haptic feedback is available and calling this on pre iOS 10 will have no effect.

Library/UIButton+HapticFeedback.swift
```diff
@@ -0,0 +1,14 @@
...
+extension UIButton {
+
+  @available(iOS 10.0, *)
+  public func generateSelectionFeedback() {
+    UISelectionFeedbackGenerator().selectionChanged()
+  }
+
+  @available(iOS 10.0, *)
+  public func generateSuccessFeedback() {
+    UINotificationFeedbackGenerator().notificationOccurred(.success)
```
dusi: This could be replaced with `UIFeedbackGenerator().ksr_successFeedback()` in order to not repeat code

Library/UIButton+HapticFeedback.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
+extension UIButton {
+
+  @available(iOS 10.0, *)
+  public func generateSelectionFeedback() {
+    UISelectionFeedbackGenerator().selectionChanged()
```
dusi: This should be moved to `UIFeedbackGenerator` extension and called like `UIFeedbackGenerator(). ksr_selectionFeedback()`

Library/UIFeedbackGenerator+Extensions.swift
```diff
@@ -0,0 +1,8 @@
+import UIKit
+
+@available(iOS 10.0, *)
+extension UIFeedbackGenerator {
+  public func ksr_successFeedbackGenerator() {
```
dusi: Might make sense to omit `Generator` in the method naming because it will always be called like `UIFeedbackGenerator().ksr_success()` (which would duplicate both Feedback and Generator)...so how about `ksr_success()` & `ksr_selection()`

Library/ViewModels/DiscoveryPostcardViewModelTests.swift
```diff
@@ -159,7 +183,7 @@ internal final class DiscoveryPostcardViewModelTests: TestCase {
         self.saveButtonSelected.assertValues([true], "Save button is selected at first.")
         self.saveButtonEnabled.assertValueCount(0)
 
-        self.vm.inputs.saveButtonTapped()
+        self.vm.inputs.saveButtonTapped(selected: true)
```
dusi: If you want to omit these explicit parameters passed into the function you can make the default value `true`? So that you don't have to change the call site here and it could stay `self.vm.inputs.saveButtonTapped()`

so the function would be
```
func saveButtonTapped(selected: Bool = true) { ... }
```

justinswart:
@dusi it's called PTO! We've got this 😄👍

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -71,6 +71,12 @@ public protocol DiscoveryPostcardViewModelOutputs {
   /// Emits a boolean to determine whether or not to display funding progress container view.
   var fundingProgressContainerViewHidden: Signal<Bool, NoError> { get }
 
+  /// Emits a boolean to determine whether to create haptics
+  var generateSelectionFeedback: Signal<Void, NoError> { get }
```
justinswart: Think we should make these types `()` just for consistency.

Library/ViewModels/DiscoveryPostcardViewModel.swift
```diff
@@ -422,6 +431,8 @@ public final class DiscoveryPostcardViewModel: DiscoveryPostcardViewModelType,
   public let deadlineTitleLabelText: Signal<String, NoError>
   public let fundingProgressBarViewHidden: Signal<Bool, NoError>
   public let fundingProgressContainerViewHidden: Signal<Bool, NoError>
+  public let generateSuccessFeedback: Signal<Void, NoError>
```
justinswart: Think we should make these types `()` just for consistency.

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -315,6 +327,6 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
   }
 
   @objc fileprivate func saveButtonTapped() {
-    self.viewModel.inputs.saveButtonTapped()
+    self.viewModel.inputs.saveButtonTapped(selected: self.saveButton.isSelected)
```
justinswart: Seeing as we need the button here, how about we change the point where we add the target to:

```swift
self.saveButton.addTarget(self, action: #selector(saveButtonTapped(_:)), for: .touchUpInside)
```

That way the button will pass itself through to the selector and we could access it like this:

```swift
@objc fileprivate func saveButtonTapped(_ button: UIButton) {
  self.viewModel.inputs.saveButtonTapped(selected: button.isSelected)
}
```

justinswart:
@ifbarrera yeah @cdolm92 and I paired on this yesterday and she mentioned that but you're right it goes hand-in-hand with making the UI act optimistically so it should be resolved soon when I get the other work merged in #460 👍

PR #465, opened 2018-10-30
Reference scroll views properly per controller basis

dusi:
# What

Fixes tab bar's tap again gesture to properly scroll top root level controllers.

# Why

Due to the fact our main controllers got more complex this functionality got lost in the process.

# How

Explicitly conform to `TabBarScrollable` protocol and properly reference scroll view. The default protocol implementation works just fine for regular view controllers whose top view is `UIScrollView` and we're using custom override for more complex hierarchies.

# See 👀

| Before | After |
| --- | --- |
| before-logged-out | after-logged-out |
| before-logged-in | eee |

### DIFF
--- a/Kickstarter-iOS/ViewModels/RootViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModel.swift
@@ -26,8 +26,11 @@ internal protocol RootViewModelInputs {
   /// Call when the language selection has changed
   func currentLanguageChanged()
 
+  /// Call before selected tab bar index changes.
+  func shouldSelect(index: Int?)
+
   /// Call when selected tab bar index changes.
-  func didSelectIndex(_ index: Int)
+  func didSelect(index: Int)
 
   /// Call when we should switch to the activities tab.
   func switchToActivities()
@@ -157,6 +160,7 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
     self.selectedIndex =
       Signal.combineLatest(
         .merge(
+          self.viewDidLoadProperty.signal.mapConst(0),
           self.didSelectIndexProperty.signal,
           self.switchToActivitiesProperty.signal.mapConst(1),
           self.switchToDiscoveryProperty.signal.mapConst(0),
@@ -169,13 +173,21 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
         self.viewDidLoadProperty.signal)
         .map { idx, vcs, _ in clamp(0, vcs.count - 1)(idx) }
 
-    let selectedTabAgain = self.selectedIndex.combinePrevious()
-      .map { prev, next -> Int? in prev == next ? next : nil }
+    let shouldSelectIndex = self.shouldSelectIndexProperty.signal
       .skipNil()
 
-    self.scrollToTop = self.setViewControllers
-      .takePairWhen(selectedTabAgain)
+    let selectedTabAgain = self.selectedIndex
+      .takePairWhen(shouldSelectIndex)
+      .filter { prev, next in prev == next }
+      .map { $1 }
+
+    self.scrollToTop = Signal.combineLatest(
+      self.setViewControllers,
+      selectedTabAgain
+      )
       .map { vcs, idx in vcs[idx] }
+      .map(extractViewController)
+      .skipNil()
 
     self.tabBarItemsData = Signal.combineLatest(currentUser, .merge(
       self.viewDidLoadProperty.signal,
@@ -195,8 +207,13 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
     self.currentLanguageProperty.value = ()
   }
 
+  fileprivate let shouldSelectIndexProperty = MutableProperty<Int?>(nil)
+  internal func shouldSelect(index: Int?) {
+    self.shouldSelectIndexProperty.value = index
+  }
+
   fileprivate let didSelectIndexProperty = MutableProperty(0)
-  internal func didSelectIndex(_ index: Int) {
+  internal func didSelect(index: Int) {
     self.didSelectIndexProperty.value = index
   }
   fileprivate let switchToActivitiesProperty = MutableProperty(())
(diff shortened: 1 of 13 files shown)


Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -157,6 +160,7 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
     self.selectedIndex =
       Signal.combineLatest(
         .merge(
+          self.viewDidLoadProperty.signal.mapConst(0),
```
dusi: This is so that by default it is tapped at least once (thanks to `viewDidLoad`). This should be a proper behavior.

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -319,3 +336,11 @@ private func profileController() -> UIViewController {
 
   return BackerDashboardViewController.instantiate()
 }
+
+private func extractViewController(_ viewController: UIViewController) -> UIViewController? {
+  if let navigationController = viewController as? UINavigationController {
+    return navigationController.viewControllers.count == 1 ? navigationController.viewControllers.first : nil
```
dusi: Only extract embedded view controller if we're popped all the way to the top of the stack..we don't want to scroll view controllers on the stack if they are overlapped with other controllers.

Kickstarter-iOS/Views/Controllers/BackerDashboardViewController.swift
```diff
@@ -367,3 +367,11 @@ extension BackerDashboardViewController: UIGestureRecognizerDelegate {
       return true
   }
 }
+
+extension BackerDashboardViewController: TabBarControllerScrollable {
+  func scrollToTop() {
```
dusi: Unfortunately due to the nature and setup of `BackerDashboardViewController` I didn't find a good way to test this.

Kickstarter-iOS/Views/Controllers/DiscoveryViewController.swift
```diff
@@ -175,3 +175,19 @@ extension DiscoveryViewController: DiscoveryNavigationHeaderViewDelegate {
     self.filter(with: params)
   }
 }
+
+extension DiscoveryViewController: TabBarControllerScrollable {
+  func scrollToTop() {
```
dusi: Unfortunately due to the nature and setup of `DiscoveryViewController` I didn't find a good way to test this.

Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -219,9 +231,16 @@ public final class RootTabBarViewController: UITabBarController {
 }
 
 extension RootTabBarViewController: UITabBarControllerDelegate {
+  public func tabBarController(_ tabBarController: UITabBarController,
```
dusi: In addition to `didSelect` we had to utilize `shouldSelect` since this is the only place we can check the navigation stack (we simply don't have that information in `didSelect` because it's already reflecting popped state). We need navigation stack in order to prevent from scrolling to top for non-top level controllers.

i.e. scenario:
1. Go to Profile (stack: Profile)
2. Tap Inbox (stack: Profile > Inbox - inbox is on top)
- now tapping scroll to top would scroll Profile while Inbox is being popped of the stack (not desired)

dusi:
[Fix tab bar's scroll to top functionality](https://trello.com/c/fixjG1tz/1039-fix-tab-bars-scroll-to-top-functionality)

justinswart:
Marked as WIP until Pavel's back from PTO 🏖

Kickstarter-iOS/Views/Controllers/BackerDashboardViewControllerTests.swift
```diff
@@ -0,0 +1,23 @@
+@testable import Kickstarter_Framework
+import XCTest
+
+internal final class BackerDashboardViewControllerTests: TestCase {
+  func testScrolToTop() {
+    let timeout = 0.1
```
dusi: Arbitrary (small enough) timeout so that our tests are not slowed down too much

Kickstarter-iOS/Views/Controllers/BackerDashboardViewControllerTests.swift
```diff
@@ -0,0 +1,23 @@
...
+    let vc = BackerDashboardViewController.instantiate()
+    _ = vc.view
```
dusi: Causes `viewDidLoad` to be called

Kickstarter-iOS/Views/Controllers/BackerDashboardViewControllerTests.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    let configureExpectation = expectation(description: "Waiting for view controllers to be configured")
+
+    DispatchQueue.main.asyncAfter(deadline: .now() + timeout) {
```
dusi: We have to use `asyncAfter` in order for the controllers (of the page view controller) to be configured (this is due to the fact that we're A) using `.observeForControllerAction()` and B) using `DispatchQueue.main.async`

Kickstarter-iOS/Views/Controllers/DiscoveryViewControllerTests.swift
```diff
@@ -0,0 +1,32 @@
+@testable import Kickstarter_Framework
+import XCTest
+
+internal final class DiscoveryViewControllerTests: TestCase {
+  func testScrollToTop_DiscoveryPages() {
```
justinswart: Can I suggest this change:

```swift
func testScrollToTop_DiscoveryPages() {
  let vc = DiscoveryViewController.instantiate()
  _ = vc.view

  let configureExpectation = expectation(description: "Waiting for view controllers to be configured")
  let pageViewController = vc.children

  DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
    while pageViewController?.viewControllers?.first?.view as? UIScrollView != nil {
      configureExpectation.fulfill()
      break
    }
  }

  waitForExpectations(timeout: 10, handler: nil)

  XCTAssertNotNil(pageViewController?.viewControllers?.first?.view as? UIScrollView)
}
```

The main difference is that the test will take as little as 0.1 seconds but not more than 10 seconds. I think it _might_ be less flakey but in most cases will pass quickly.

You can repurpose this for the other async test too.

dusi: Still won't pass :/

dusi:
I'd suggest we simply leave the tests out for now and revisit this later...they're too flaky at the moment and I'm out of ideas how to make this more robust :/

PR #456, opened 2018-10-26
Access preferredContentSizeCategory on the main thread

justinswart:
# What

Ensures that we access `preferredContentSizeCategory` on the main thread.

# Why

Xcode's main thread sanitizer was flagging the fact that we're sometimes accessing `preferredContentSizeCategory` outside of the main thread during Koala tracking.

screen shot 2018-10-25 at 5 01 27 pm
screen shot 2018-10-25 at 5 01 48 pm

# How

- Set the value on instantiation and hold a reference to it for tracking later on.
- Observe changes to it and update the value on change (on the main thread).

PR #467, opened 2018-10-31
Ensure preferredContentSizeCategory is accessed on the main thread

justinswart:
# What

Once again ensures that this property is only accessed on the main thread.

# Why

Turns out #456 didn't completely solve this property being accessed on background threads because `Koala` is sometimes instantiated from a background thread on environment changes. Xcode's main thread sanitizer has been flagging it occasionally.

# How

Implement the following logic:
1. If on main thread at instantiation, set this var.
2. Else on instantiation set it asynchronously on main thread.
3. Continue to observe any changes to it on main thread.

### DIFF
--- a/Library/Koala/Koala.swift
+++ b/Library/Koala/Koala.swift
@@ -341,7 +341,9 @@ public final class Koala {
   }
 
   private func updateAndObservePreferredContentSizeCategory() {
-    let update = { self.preferredContentSizeCategory = UIApplication.shared.preferredContentSizeCategory }
+    let update = { [weak self] in
+      self?.preferredContentSizeCategory = UIApplication.shared.preferredContentSizeCategory
+    }
 
     self.preferredContentSizeCategoryObserver = NotificationCenter.default.addObserver(
       forName: UIContentSizeCategory.didChangeNotification,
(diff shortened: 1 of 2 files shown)


Library/Koala/Koala.swift
```diff
@@ -328,26 +328,32 @@ public final class Koala {
...
       self.distinctId = distinctId
-      self.preferredContentSizeCategory = preferredContentSizeCategory
 
-      self.observePreferredContentSizeCategory()
+      self.updateAndObservePreferredContentSizeCategory()
   }
 
-  private func observePreferredContentSizeCategory() {
+  private func updateAndObservePreferredContentSizeCategory() {
+    let update = { self.preferredContentSizeCategory = UIApplication.shared.preferredContentSizeCategory }
```
ifbarrera: Should we use `[weak self]` here just in case?

justinswart: Hmm yeah good question - I _think_ the observer will capture the `update` closure here but that in turn _shouldn't_ capture `self`.

I'll do a couple tests!

justinswart: Yip that did cause a retain cycle, good catch!

PR #468, opened 2018-11-02
Fix view model target memberships

justinswart:
# What

This moves our view models into their correct target memberships.

# Why

I noticed that a bunch of view models and types had become part of the app framework. These are meant to be in the `Library` framework. There are only a handful of view models that should be members of the app framework such as the `AppDelegateViewModel`, `RootViewModel` and one or two others. And, in fact, perhaps even those could be refactored to not be.

| ❌| ✅|
| screen shot 2018-11-01 at 3 28 48 pm | screen shot 2018-11-01 at 6 22 59 pm |

This helps in a few ways:
- Consistency
- View models should generally not be aware of UI element types
- Better control over visibility of methods in app classes and in tests ensuring that we only reach into view model `internal` types via `@testable`

# How

Moved memberships and relocated types so that they could still be accessed where necessary.

Note: There may be more types that should move but I focused on view models for now.


Kickstarter-iOS/MessageBannerType.swift
```diff
@@ -9,24 +8,20 @@ public enum MessageBannerType {
...
-      return UIColor.ksr_apricot_600
+      return .ksr_apricot_600
     case .info:
-      return UIColor.ksr_cobalt_500
+      return .ksr_cobalt_500
     }
   }
 
-  var iconImage: UIImage? {
+  var iconImageName: String? {
```
justinswart: Returning the string name here so that this doesn't need access to the `image(named:)` function.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -41,8 +41,9 @@ final class MessageBannerViewController: UIViewController {
         self?.showViewAndAnimate(isHidden)
     }
 
-    self.viewModel.outputs.iconImage
+    self.viewModel.outputs.iconImageName
       .observeForUI()
+      .map { image(named: $0) }
```
justinswart: Moved the `image(named:)` function call here.

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -9,7 +9,9 @@ final class SettingsAccountViewController: UIViewController {
   @IBOutlet private weak var tableView: UITableView!
 
   private let dataSource = SettingsAccountDataSource()
-  fileprivate let viewModel: SettingsAccountViewModelType = SettingsAccountViewModel()
+  fileprivate let viewModel: SettingsAccountViewModelType = SettingsAccountViewModel(
+    SettingsAccountViewController.viewController(for:)
```
justinswart: In order to avoid needing access to `UIViewController` types in the view model I'm passing this function in here which was previously contained inside the view model. I think we could potentially improve the way this logic is handled but just did this for now to get it building with minimal changes. Let's discuss 👍

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -9,7 +9,9 @@ final class SettingsViewController: UIViewController {
 
   private let dataSource = SettingsDataSource()
   private var userUpdatedObserver: Any?
-  private let viewModel: SettingsViewModelType = SettingsViewModel()
+  private let viewModel: SettingsViewModelType = SettingsViewModel(
+    SettingsViewController.viewController(for:)
```
justinswart: Same pattern here.

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -35,12 +34,12 @@ protocol ChangePasswordViewModelOutputs {
...
 }
 
-protocol ChangePasswordViewModelType {
+public protocol ChangePasswordViewModelType {
   var inputs: ChangePasswordViewModelInputs { get }
   var outputs: ChangePasswordViewModelOutputs { get }
 }
 
-struct ChangePasswordViewModel: ChangePasswordViewModelType,
+public class ChangePasswordViewModel: ChangePasswordViewModelType,
```
justinswart: View models should always be `class`es

justinswart: In fact this should also be `final` - will add.

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -78,7 +77,7 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
     self.dismissCurrencyPicker = self.dismissPickerTapProperty.signal
 
     self.transitionToViewController = self.selectedCellTypeProperty.signal.skipNil()
-      .map { SettingsAccountViewModel.viewController(for: $0) }
+      .map(viewControllerFactory)
```
ifbarrera: We should maybe consider just having the `cellType` returned in the output and keeping `viewController(for cellType:)` in the view controller. The benefit to keeping it in the view model though is that you can test that tapping a particular cell will transition to a VC of a specific type.

PR #470, opened 2018-11-05
added currency picker cell to frameworks

cdolm92:
# What

Added Currency Picker Cell to Frameworks.

# Why

App was crashing when currency cell was tapped.


justinswart:
@cdolm92 oh good that you caught that! Which framework was it missing from?

PR #472, opened 2018-11-06
Refactor project saving

justinswart:
# What

This is the continuation of work that began in #460. What initially seemed like a fairly trivial change turned into this 😄. But one positive outcome is that it's allowed us to replace one of our v1 api endpoints with GraphQL mutations.  I've made this a separate PR based off that branch so that it's easier to review the changes. 

This PR:
- adds `watchProject` and `unwatchProject` GraphQL mutations.
- refactors tons of duplicated 'project starring' code to be in a shared view model called `WatchProjectViewModel`.
- refactors caching and error handling for optimistic UI project watching (saving)

# Why

See #460 for more info about this.

# How

- Refactored `ProjectNavBarViewController` and `DiscoveryPostcardCell` to use the `WatchProjectViewModel` for all actions related to watching and unwatching projects
  - previous nomenclature had this as project starring/saving. I believe it's still called _saving_ to the user.
- Implemented caching of the toggle state for both views whereas it was previously only being used for the `DiscoveryPostcardCell`.
- Handle error cases and toggle the state back if the call fails.

# What this isn't

I've intentionally not made changes to the list of saved projects in the profile view. This list refreshes on the view appearing and updates the list that way. If it's necessary to make changes to that view from a product/UX perspective, I'd like to address that as a separate piece of work. cc @nnekab @dannyalright


KsApi/GraphIDBridging.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public protocol GraphIDBridging {
```
justinswart: I've added this as a simple protocol that our models can conform to if they're able to return a GraphQL relay ID which is the concatenation of a model name, hyphen and the integer ID, base64-encoded. Essentially all the model needs to do is to conform to this protocol and return its model name, e.g. `"Project"`.

KsApi/GraphIDBridging.swift
```diff
@@ -0,0 +1,14 @@
...
+  static var modelName: String { get }
+  var graphID: String { get }
+  var id: Int { get }
+}
+
+extension GraphIDBridging {
+  public var graphID: String {
+    //swiftlint:disable:next force_unwrapping
+    return "\(type(of: self).modelName)-\(self.id)".data(using: .utf8)!.base64EncodedString()
```
justinswart: I think I can actually refactor this to not use a force unwrap.

KsApi/MockService.swift
```diff
@@ -142,6 +139,11 @@ internal struct MockService: ServiceType {
 
   fileprivate let updateUserSelfError: ErrorEnvelope?
 
+  fileprivate let unwatchProjectMutationResult: Result<GraphMutationUnwatchProjectResponseEnvelope,
+  GraphError>?
+
+  fileprivate let watchProjectMutationResult: Result<GraphMutationWatchProjectResponseEnvelope, GraphError>?
```
justinswart: I'm planning to spend some time to change all of these to `Result<T, Error>` types.

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -0,0 +1,267 @@
...
+
+  public init() {
+    let viewReady = Signal.merge(self.viewDidLoadProperty.signal, self.awakeFromNibProperty.signal)
+
+    let configuredProject = Signal.combineLatest(
+      self.projectProperty.signal.skipNil(),
+      viewReady
+      )
+      .map(first)
+      .map(cached(project:))
```
justinswart: This is key for the optimistic UI approach to work correctly between discovery and drilled-in view. As soon as the project is configured it checks the cache for its 'starred' value.

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -0,0 +1,267 @@
...
+      .takePairWhen(saveButtonTapped)
+
+    let watchProjectToggle = projectOnSaveButtonToggle
+      .map { project, selected in (project, !selected) }
+      // immediately cache and return the project with the correct watch value
+      .map(watchAndCacheProject(_:shouldWatch:))
+
+    // make the mutation request to the watch/unwatch and flip the result only if the call fails
+    let watchProjectResult = watchProjectToggle
+      .ksr_debounce(.milliseconds(500), on: AppEnvironment.current.scheduler)
```
justinswart: I've changed debouncing to be 500ms, 1 second felt too long.

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -0,0 +1,267 @@
...
+
+    self.postNotificationWithProject = project
+      .takeWhen(saveButtonTapped)
+
+    projectOnSaveButtonToggleSuccess
+      .observeValues { AppEnvironment.current.koala.trackProjectSave($0, context: .project) }
+  }
+
+  private let awakeFromNibProperty = MutableProperty(())
+  public func awakeFromNib() {
```
justinswart: We have `awakeFromNib()` as well as `viewDidLoad()` given that this view model is sometimes used in a cell and sometimes in a view controller.

justinswart:
[Delay when saving a project](https://trello.com/c/RHDv4MtB/1037-delay-when-saving-a-project)

KsApi/GraphIDBridging.swift
justinswart: tl;dr - I did

KsApi/models/GraphMutationUnwatchProjectResponseEnvelope.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public struct GraphMutationUnwatchProjectResponseEnvelope: Decodable {
+  public private(set) var unwatchProject: UnwatchProject
+
+  public struct UnwatchProject: Decodable {
```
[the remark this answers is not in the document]
justinswart: Hmm yeah good question... I guess I figured we want a 1 response per query/mutation and to try to be consistent around that? 

But, true, we could use an alias on the root key in the mutation for the model to be reusable (we'd need `watchProject` and `unwatchProject` to both be the same).

I'm happy to combine them! What do you think?

KsApi/mutations/inputs/WatchProjectInput.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public struct WatchProjectInput: GraphMutationInput {
```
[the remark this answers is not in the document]
justinswart: Yeah, see my other comment but sure, we could do that!

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -0,0 +1,268 @@
...
+
+    self.showProjectSavedAlert = project
+      .takeWhen(saveButtonTapped)
+      .filter { !$0.endsIn48Hours(today: AppEnvironment.current.dateType.init().date ) }
+      .filter { _ in
+        !AppEnvironment.current.ubiquitousStore.hasSeenSaveProjectAlert ||
+          !AppEnvironment.current.userDefaults.hasSeenSaveProjectAlert
+      }
+      .on(value: { _ in
+        AppEnvironment.current.ubiquitousStore.hasSeenSaveProjectAlert = true
```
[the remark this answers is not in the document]
justinswart: This is just code that has always been in the (formerly) starring/unstarring view models so I decided to not change it at this stage and prove that all the tests still pass. Isn't `ubiquitousStore` referring to iCloud? I haven't really dug into it. But I think I'd like to understand that better and then potentially change this in a separate PR if necessary.

PR #473, opened 2018-11-07
Stripe element add new card

cdolm92:
# What

Credit card field in "Add Card" screen should use `STPPaymentCardTextField` Stripe element, and it is configured to look like the designs.

# See 👀
simulator screen shot - iphone 8 - 2018-11-12 at 14 26 24

simulator screen shot - iphone 8 - 2018-11-12 at 14 26 11

# TODO ⏰

- [x] CC Image Color is different from design @dannyalright 
- [x] Snapshot Tests
- [x] Should 'Use Camera' be included in this? @nnekab 
- [ ] Feedback


Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -45,42 +41,20 @@ internal final class AddNewCardViewController: UIViewController {
...
-    _ = self.expirationTextField
-      |> \.placeholder %~ { _ in Strings.MMYY() }
-
-    _ = self.securityCodeTextField
-      |> \.placeholder %~ { _ in Strings.CVC() }
+    _ = self.paymentField
+      |> \.borderColor .~ nil
+      |> \.font .~ .ksr_body()
+      |> \.cursorColor .~ .ksr_green_700
+      |> \.placeholderColor .~  UIColor(red: 0.0, green: 0.0, blue: 0.0, alpha: 0.17)
```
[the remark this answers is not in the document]
dannyalright: is that a design question? I'm not really sure I understand it. Do you mean the blinking cursor? The placeholder text is greyscale

dannyalright:
>  CC Image Color is different from design

It's different because you're using the one from Stripe. Right? That's cool by me.

dannyalright:
Looking good visually. Any way we can get the `CVC` text to line up to the same right-margin as the `Name` in the field above? I'm guessing not, as it probably gets its positing from the card number and expiry date.

Could you please tell me the color for the placeholder? It's the default iOS color for placeholder text, right?

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,69 @@
...
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = self
+      |> settingsViewControllerStyle
+
+    _ = self.cardholderNameLabel
+      |> \.text %~ { _ in Strings.Cardholder_name() }
+      |> \.textColor .~ .ksr_text_dark_grey_900
```
ifbarrera: Can we use `settingsTitleLabelStyle` so it's consistent with our other form fields in Settings?

cdolm92:
The color is `UIColor.lightGrayColor` or `#AAAAAA`. It's not the default iOS Color for placeholder text. This is the default for the stripe element placeholder text.

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
cdolm92: going to use `settingsSectionLabelStyle` because we want the text color to be  `.ksr_text_dark_grey_900`

cdolm92:
@ifbarrera  @kickstarter/native-squad  g2g?

dannyalright:
Could we hold off until our after our color call?

ifbarrera:
@cdolm92 can you make this PR against `feature-payment-methods`? Lgtm after that 👍

PR #474, opened 2018-11-07
Resend email verification

Scollaco:
# What
Shows a view with a `Send Verification Email` (backers) or `Re-send Verification Email`(creators) button on ChangeEmailViewController.

# Why
After changing the email, we should ask the user to verify the new email. (Note that this is optional for backers).
For more details: https://trello.com/c/Zi7vGBjD/1047-email-update-resending-verification-ios

# See 👀
change-email


Scollaco:
[Email update: resending verification iOS](https://trello.com/c/Zi7vGBjD/1047-email-update-resending-verification-ios)

KsApi/mutations/inputs/EmptyInput.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+public struct EmptyInput: GraphMutationInput {
+
+  public init() {}
```
ifbarrera: Is this initializer necessary?

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -176,6 +186,61 @@ final class ChangeEmailViewModelTests: TestCase {
...
+    withEnvironment(apiService: MockService(sendEmailVerificationError: error)) {
+
+      self.vm.inputs.resendVerificationEmailButtonTapped()
+      self.scheduler.advance()
+
+      self.didFailToSendVerificationEmail.assertValue(GraphError.invalidInput.localizedDescription)
+    }
+  }
+
+  func testdidSendVerificationEmailEmits_OnSuccess() {
```
ifbarrera: Small camel-case typo here 😅. Should be `testDidSend...`

KsApi/mutations/inputs/EmptyInput.swift
Scollaco: Yeah, otherwise the compiler will interpret the initializer as `internal` and the ViewModel won't be able to access it.

PR #479, opened 2018-11-09
Delete payment methods

justinswart:
# What

Adds the ability to delete payment methods using standard iOS deletion gestures.

# Why

This is a continuation of the work done in #457 to allow users to delete payment methods. This branch is also based off the `feature-payment-methods` branch.

# How

`UITableView` gives us most of this functionality for free so this was fairly straightforward to implement. I've also taken an optimistic UI approach here so the credit card will be removed immediately upon deletion and we will only show it again if an error occurs.

Some ways that we may want to iterate on this:

-  I think we might want to confirm deletion in some way, either through an alert _after_ swiping, but _before_ performing the deletion, or potentially displaying an undo button after animating it away but not committing the deletion.
- I've come up with some copy for an error alert (see below) which refers to the thing that failed to be removed as a _credit card_, do we want to stick to _payment method_ here?

cc @nnekab @dannyalright for the above.

# See 👀

| Tap to delete | Swipe to delete | Error case |
| ------------- | ------------- | ------------- |
|  |  |  |

### DIFF
--- a/Kickstarter-iOS/DataSources/PaymentMethodsDataSourceTests.swift
+++ b/Kickstarter-iOS/DataSources/PaymentMethodsDataSourceTests.swift
@@ -10,12 +10,35 @@ internal final class PaymentMethodsDataSourceTests: XCTestCase {
   let tableView = UITableView(frame: .zero)
 
   func testDataSource() {
-
     let cards = GraphUserCreditCard.template.storedCards.nodes
     self.dataSource.load(creditCards: cards)
 
     XCTAssertEqual(7, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
 
     XCTAssertEqual("CreditCardCell", self.dataSource.reusableId(item: 0, section: 0))
+  }
+
+  func testCardDeletion() {
+    let cards = GraphUserCreditCard.template.storedCards.nodes
+    self.dataSource.load(creditCards: cards)
+
+    XCTAssertEqual(7, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
+
+    XCTAssertEqual("CreditCardCell", self.dataSource.reusableId(item: 0, section: 0))
+
+    guard let card = cards.first else {
+      XCTFail("Card should exist")
+      return
+    }
+
+    var deletedCard: GraphUserCreditCard.CreditCard?
+    let deletionHandler = { card in
+      deletedCard = card
+    }
+
+    self.dataSource.deletionHandler = deletionHandler
+    self.dataSource.tableView(self.tableView, commit: .delete, forRowAt: .init(row: 0, section: 0))
+
+    XCTAssertEqual(deletedCard, card)
   }
 }
--- a/Kickstarter-iOS/DataSources/PaymentMethodsDataSourceTests.swift
+++ b/Kickstarter-iOS/DataSources/PaymentMethodsDataSourceTests.swift
@@ -22,10 +22,6 @@ internal final class PaymentMethodsDataSourceTests: XCTestCase {
     let cards = GraphUserCreditCard.template.storedCards.nodes
     self.dataSource.load(creditCards: cards)
 
-    XCTAssertEqual(7, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
-
-    XCTAssertEqual("CreditCardCell", self.dataSource.reusableId(item: 0, section: 0))
-
     guard let card = cards.first else {
       XCTFail("Card should exist")
       return
--- a/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
@@ -77,7 +77,8 @@ internal final class PaymentMethodsViewController: UIViewController {
     self.viewModel.outputs.tableViewIsEditing
       .observeForUI()
       .observeValues { [weak self] isEditing in
-        self?.tableView.isEditing = isEditing
+        _ = self?.tableView
+          ?|> \.isEditing .~ isEditing
     }
 
     self.viewModel.outputs.showAlert
--- a/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
@@ -28,8 +28,12 @@ internal final class PaymentMethodsViewController: UIViewController {
       title: Strings.discovery_favorite_categories_buttons_edit(),
       style: .plain,
       target: self,
-      action: nil
+      action: #selector(edit)
     )
+
+    self.dataSource.deletionHandler = { [weak self] creditCard in
+      self?.viewModel.inputs.didDelete(creditCard)
+    }
   }
 
   override func bindStyles() {
@@ -69,6 +73,24 @@ internal final class PaymentMethodsViewController: UIViewController {
       .observeValues { [weak self] in
         self?.goToAddCardScreen()
     }
+
+    self.viewModel.outputs.tableViewIsEditing
+      .observeForUI()
+      .observeValues { [weak self] isEditing in
+        self?.tableView.isEditing = isEditing
+    }
+
+    self.viewModel.outputs.showAlert
+      .observeForControllerAction()
+      .observeValues { [weak self] message in
+        self?.present(UIAlertController.genericError(message), animated: true)
+    }
+  }
+
+  // MARK: - Actions
+
+  @objc private func edit() {
+    self.viewModel.inputs.editButtonTapped()
   }
 
   private func goToAddCardScreen() {
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -156,7 +156,6 @@
 		77C5E252214182CA002E1670 /* SettingsPrivacySwitchCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77C5E251214182CA002E1670 /* SettingsPrivacySwitchCell.swift */; };
 		77CD894921791B05003066DA /* ProjectStatsTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77CD894821791B05003066DA /* ProjectStatsTemplate.swift */; };
 		77CD8981217FA01B003066DA /* ProjectPamphletContentViewControllerConversionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77CD8980217FA01B003066DA /* ProjectPamphletContentViewControllerConversionTests.swift */; };
-		77CD89F2218215CB003066DA /* GraphError+LocalizedDescription.swift in Sources */ = {isa = PBXBuildFile; fileRef = 775DFAD621627D9C00620CED /* GraphError+LocalizedDescription.swift */; };
 		77D7A739214187A9003F258C /* SettingsSwitchCellType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77D7A738214187A9003F258C /* SettingsSwitchCellType.swift */; };
 		77E44B7421823A56006446B8 /* ChangePasswordInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77E44B7321823A56006446B8 /* ChangePasswordInput.swift */; };
 		77E6440120F64F0B005F6B38 /* HelpDataSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77E6440020F64F0B005F6B38 /* HelpDataSource.swift */; };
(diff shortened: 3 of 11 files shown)


justinswart:
[Delete Payment Methods MVP](https://trello.com/c/qVBiR4qk/1050-delete-payment-methods-mvp)

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -28,8 +28,12 @@ internal final class PaymentMethodsViewController: UIViewController {
       title: Strings.discovery_favorite_categories_buttons_edit(),
       style: .plain,
       target: self,
-      action: nil
+      action: #selector(edit)
     )
+
+    self.dataSource.deletionHandler = { [weak self] creditCard in
+      self?.viewModel.inputs.didDelete(creditCard)
```
justinswart: This flow is a little awkward but we are only told about the deletion via the `UITableViewDataSourceDelegate` 🤔

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -1600,13 +1600,6 @@
 			remoteGlobalIDString = A7DB1D5A1C9F4FD7008244DA;
 			remoteInfo = "ReactiveExtensions-TestHelpers-tvOS";
 		};
-		D04AABCB218BB25E00CF713E /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = 80762DE71D071BCF0074189D /* AlamofireImage.xcodeproj */;
```
justinswart: This keeps coming and going 🤦‍♂️

KsApi/mutations/inputs/PaymentSourceDeleteInput.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public struct PaymentSourceDeleteInput: GraphMutationInput {
+  let paymentSourceId: String
```
justinswart: Not sure why the input for this is `paymentSourceId` and not just `id`. We could create a reusable input type like `IdOnlyInput` for this sort of thing if it were.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -24,17 +28,52 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
...
+          defaultValue: "Something went wrong and we were unable to remove your credit card, please try again."
+        )
+    }
 
-    self.paymentMethods = paymentMethodsEvent.values().map { $0.me.storedCards.nodes }
+    let paymentMethodsValues = paymentMethodsEvent.values().map { $0.me.storedCards.nodes }
+
+    self.paymentMethods = Signal.merge(
+      paymentMethodsValues,
+      paymentMethodsValues.takeWhen(deletePaymentMethodEventsErrors)
```
justinswart: We repeat the last fetched values if an error occurs.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -24,17 +28,52 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
...
+
+    let deletePaymentMethodEventsErrors = deletePaymentMethodEvents.errors()
+
+    self.showAlert = deletePaymentMethodEventsErrors
+      .ignoreValues()
+      .map {
+        localizedString(
+          key: "Something_went_wrong_and_we_were_unable_to_remove_your_credit_card_please_try_again",
+          //swiftlint:disable:next line_length
+          defaultValue: "Something went wrong and we were unable to remove your credit card, please try again."
```
justinswart: Copy subject to review!

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -24,17 +28,52 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
...
+    self.tableViewIsEditing = self.editButtonTappedSignal.scan(false) { current, _ in !current }
+  }
+
+  let (didDeleteCreditCardSignal, didDeleteCreditCardObserver) = Signal<GraphUserCreditCard.CreditCard,
+    NoError>.pipe()
+  public func didDelete(_ creditCard: GraphUserCreditCard.CreditCard) {
+    self.didDeleteCreditCardObserver.send(value: creditCard)
+  }
+
+  let (editButtonTappedSignal, editButtonTappedObserver) = Signal<(), NoError>.pipe()
```
justinswart: Would really like to start using `pipe`s instead of `MutableProperty`s. `pipe()` has the following signature:

```swift
func pipe(disposable: Disposable? = nil) -> (output: Signal, input: Observer)
```

This allows us to `send` values to the `input` which is an `Observer`. The main benefit over a `MutableProperty` is that it does not require an initial value, and therefore doesn't require an initial `nil` value as we tend to provide. That then in turn means that we don't need to use `.signal.skipNil()` everywhere because it's already a `Signal<T, E>`.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -24,17 +28,52 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
...
+    let paymentMethodsValues = paymentMethodsEvent.values().map { $0.me.storedCards.nodes }
+
+    self.paymentMethods = Signal.merge(
+      paymentMethodsValues,
+      paymentMethodsValues.takeWhen(deletePaymentMethodEventsErrors)
+    )
 
     self.goToAddCardScreen = self.didTapAddCardButtonProperty.signal
+
+    self.tableViewIsEditing = self.editButtonTappedSignal.scan(false) { current, _ in !current }
```
justinswart: Just using scan to toggle between `true` and `false` here with the initial value being `false`.

nnekab:
@justinswart nice -- love the error case. Looks great to me. Makes me wonder if we should implement something similar for adding new cards as well

Kickstarter-iOS/DataSources/PaymentMethodsDataSourceTests.swift
```diff
@@ -10,12 +10,35 @@ internal final class PaymentMethodsDataSourceTests: XCTestCase {
...
+    XCTAssertEqual("CreditCardCell", self.dataSource.reusableId(item: 0, section: 0))
+  }
+
+  func testCardDeletion() {
     let cards = GraphUserCreditCard.template.storedCards.nodes
     self.dataSource.load(creditCards: cards)
 
     XCTAssertEqual(7, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
 
     XCTAssertEqual("CreditCardCell", self.dataSource.reusableId(item: 0, section: 0))
```
Scollaco: I don't think we need to assert the Cell value again, since it's already being testes on the `testDataSource` function. Actually, since this code is identical to the test above, we could either put everything in one `testDataSource` test or delete the first assertions.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
Scollaco: Yeah,  another option would be to create a delegate, but I think it would be weird.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -69,6 +73,24 @@ internal final class PaymentMethodsViewController: UIViewController {
       .observeValues { [weak self] in
         self?.goToAddCardScreen()
     }
+
+    self.viewModel.outputs.tableViewIsEditing
+      .observeForUI()
+      .observeValues { [weak self] isEditing in
+        self?.tableView.isEditing = isEditing
```
Scollaco: Maybe here we could use 
```
_ = self?.tableView
          ?|> \.isEditing .~ isEditing
```
🤔

KsApi/mutations/inputs/PaymentSourceDeleteInput.swift
Scollaco: That's a good idea. it's similar to the `EmptyInput` we have.

Library/ViewModels/PaymentMethodsViewModel.swift
Scollaco: This approach is actually more used for what I've seen. I don't see a problem with it, just think the team should talk about it to avoid using to different approaches to do the same thing. Maybe it's a good opportunity for a rfc? 😄

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -10,18 +10,21 @@ import Prelude
...
+  let paymentMethods = TestObserver<[GraphUserCreditCard.CreditCard], NoError>()
+  let showAlert = TestObserver<String, NoError>()
+  let tableViewIsEditing = TestObserver<Bool, NoError>()
 
   internal override func setUp() {
     super.setUp()
 
-    self.vm.outputs.paymentMethods.observe(paymentMethods.observer)
     self.vm.outputs.goToAddCardScreen.observe(goToAddCardScreen.observer)
+    self.vm.outputs.paymentMethods.observe(paymentMethods.observer)
```
Scollaco: a reference to `self` is missing here. (I think I was the one who forgot it 😅  )

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -43,4 +46,78 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
+    withEnvironment(apiService: apiService) {
+
+      self.vm.inputs.viewDidLoad()
+      self.scheduler.advance()
+
+      self.tableViewIsEditing.assertValues([])
+      self.showAlert.assertValues([])
+      self.paymentMethods.assertValues([GraphUserCreditCard.template.storedCards.nodes])
+
+      self.paymentMethods.assertValues([GraphUserCreditCard.template.storedCards.nodes])
```
Scollaco: There's a duplicated line here.

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -43,4 +46,78 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
```
Scollaco: And here

KsApi/mutations/inputs/PaymentSourceDeleteInput.swift
ifbarrera: So I think whenever we use `id` in graph it refers to the relay ID, but in the case of payment methods the `id` we care about is actually the Rosie id because I believe the type `CreditCard` doesn't conform to the `Node` interface. So its better that it's explicitly named `paymentSourceId` to avoid ambiguity about what the id really is.

dannyalright:
## In general

such a nice PR.

> I've also taken an optimistic UI approach here so the credit card will be removed immediately upon deletion and we will only show it again if an error occurs.

Love it. I think we should do this with most things.

## Error case UI
This could be a nice opportunity to flog the error state of the [Snackbar/Message Banner component](https://homebase.ksr.io/kickstarter/components#snackbar). I think @ifbarrera built it in a way to be re-usable. If that's correct, is this PR the best place to implement it? Or should we make a separate ticket? 

If you disagree or its not possible: the standard iOS alert is fine.

## Error case copy

> I've come up with some copy for an error alert (see below) which refers to the thing that failed to be removed as a credit card, do we want to stick to payment method here?

I like your copy, although I think we should use _payment method_ instead of _credit card_. We allow both bank account and credit card in Germany, with possibly more countries to come.


## Undoing a deletion

> ...or potentially displaying an undo button after animating it away but not committing the deletion.

I'm not sure about this. I feel like it could be unnecessary as it's pretty hard to accidentally delete a card and not too difficult to reenter one. I could be convinced though.

[I made a Snackbar/Message Banner pattern](https://share.goabstract.com/c771e92a-c41c-4b17-bac2-d2f9e3db509f) **with button** a few months ago. I can't remember exactly why, but we decided as a team to have Snackbars only show static text. If you or someone else feels strongly about undo, I think this would be the way to do it. 

If I'm wrong, and it is indeed easy to accidentally delete a card (presumably by the _swipe to delete_ method), we could remove the _swipe to delete_ method.

KsApi/mutations/inputs/PaymentSourceDeleteInput.swift
justinswart: Ah right, thanks for explaining!

Library/ViewModels/PaymentMethodsViewModel.swift
justinswart: I don't think it's a big enough change for an RFC, will just discuss at next guild 👍

justinswart:
@dannyalright Thanks for the comments! I'll update to use _payment methods_. I think the alert vs snackbar we should investigate and iterate on in a separate PR 👍

Library/ViewModels/PaymentMethodsViewModel.swift
dannyalright: Does this needed to get updated to the new YAML key (line 50) and value (line 52) from 
https://github.com/kickstarter/kickstarter/pull/13237?

PR #480, opened 2018-11-09
Push registration on app launch

ifbarrera:
# What

Ensures that we register for push notifications on app launch, so that we can have the most up-to-date push tokens.

# Why

Previously, we only registered for push notifications once: when the user first sees the push registration dialog and hits "Ok". This means that the push tokens stored on the server could be invalid, since APNs issues new tokens under certain conditions: https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/HandlingRemoteNotifications.html#//apple_ref/doc/uid/TP40008194-CH6-SW3

We are new fetching an updated push token on every app launch if the user has authorized the app for push notifications. 

Also added some additional logging so we can easily see when push registration is happening in the logs 🤓


ifbarrera:
@justinswart thanks for thoroughly testing!

PR #481, opened 2018-11-14
Use correct function to log events in crash logs

justinswart:
# What

Uses the correct Crashlytics function to log "bread crumb" events in crash logs.

# Why

We were trying to do this using Answers and determined that it wasn't working. When I dug into investigating why that was I also found that that's not actually what we wanted 😄. Crashlytics provides a logging function for this very purpose, described [here](https://docs.fabric.io/apple/crashlytics/enhanced-reports.html#custom-logs).

# How

Replaced the `Answers` line with the Crashlytics `CLSLogv` function.

# See 👀

Tested it by forcing a crash and it looks like this on Crashlytics:

screen shot 2018-11-13 at 4 29 22 pm

### DIFF
--- a/Kickstarter-iOS/Info.plist
+++ b/Kickstarter-iOS/Info.plist
@@ -52,12 +52,14 @@
 		<key>APIKey</key>
 		<string>$(FABRIC_API_KEY)</string>
 		<key>Kits</key>
-		<array>
-			<dict>
-				<key>KitInfo</key>
-				<string>Crashlytics</string>
-			</dict>
-		</array>
+    <array>
+      <dict>
+        <key>KitInfo</key>
+        <dict/>
+        <key>KitName</key>
+        <string>Crashlytics</string>
+      </dict>
+    </array>
 	</dict>
 	<key>FacebookAppID</key>
 	<string>69103156693</string>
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/Info.plist
```diff
@@ -55,6 +55,8 @@
 		<array>
 			<dict>
 				<key>KitInfo</key>
+				<dict/>
+				<key>KitName</key>
```
justinswart: According to the [docs](https://www.fabric.io/kits/ios/crashlytics/install) this was missing.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -5470,6 +5470,7 @@
 			files = (
 			);
 			inputPaths = (
+				"$(BUILT_PRODUCTS_DIR)/$(INFOPLIST_PATH)",
```
justinswart: As @ifbarrera pointed out, this is now needed in Xcode 10.

justinswart:
[Log custom events to Crashlytics](https://trello.com/c/GsJcIupa/1043-log-custom-events-to-crashlytics)

PR #482, opened 2018-11-15
Account email verified/undeliverable alert icon

ifbarrera:
# What

Show a little alert icon in the account screen if a user's email is unverified or undeliverable.

Also in this PR:
- removed the `SettingsAccountStoryboard` and instead moved the scene to the `Settings` storyboard
- removed a request for the `User` object that wasn't being used
- added error-handling for when the request for the user account fields fails; now we show an error message banner
- added tests for the `SettingsAccountDataSource` file that were missing

# Why

So they know to correct the issue in the "Change Email" screen.

# See 👀

testaccountview_whenuseremailisunverified_device_phone4inch 2x

gxwbtglnji


ifbarrera:
[Change email error on Account page](https://trello.com/c/PLhMfph3/1065-change-email-error-on-account-page)

dannyalright:
I think we should use an Activity Indicator for any cell that has loading (or never going to load) content. I thought of this independently of this work, so sorry for the goose-chase on this. Let's chat when we're both back.

Scollaco:
I understand that the `SettingsAccountViewController` is part of the Settings, but IMO, I think that by having many ViewControllers in one storyboard, merge conflicts can happen more frequently with the time, mainly now that our team is growing. Instead, I would move to the opposite direction and keep view controllers together only if they are part of the same screen (containers / child view controllers)...

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -46,13 +56,17 @@ final class SettingsAccountDataSource: ValueCellDataSource {
       return value.cellType as? SettingsAccountCellType
     } else if let currencyValue = self[indexPath] as? SettingsCurrencyCellValue {
       return currencyValue.cellType as? SettingsAccountCellType
+    } else if self[indexPath] is Bool {
```
Scollaco: Could we follow the same pattern `else if let...` here? I think checking if `self[indexPath] is Bool` get's a little harder to read.

ifbarrera:
@justinswart in this case we have two cells that depend on content from the server - the chosen currency cell and now the `Change Email` cell with this new alert icon. IMO we should minimize the number of requests made on load of a single view controller whenever possible to reduce networking hardware overhead. For this reason I combined the request for the user's chosen currency with that of the `isEmailDeliverable` and `isEmailVerified` flags. You could make the argument that both those requests (chosen currency and `isEmailDeliverable`) would happen around the same time (on `viewDidLoad`) in which case the networking hardware would probably still be powered up to reduce the networking overhead. It's probably difficult to measure the actual impact of making multiple requests within a single view, but this was my thinking around this particular scenario.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -43,7 +43,7 @@ final class MessageBannerViewController: UIViewController {
 
     self.viewModel.outputs.iconImageName
       .observeForUI()
-      .map { image(named: $0) }
+      .map { image(named: $0, inBundle: Bundle.framework) }
```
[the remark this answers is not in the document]
ifbarrera: Yeah the asset wasn't showing up properly in the snapshots otherwise

ifbarrera:
> I understand that the `SettingsAccountViewController` is part of the Settings, but IMO, I think that by having many ViewControllers in one storyboard, merge conflicts can happen more frequently with the time, mainly now that our team is growing. Instead, I would move to the opposite direction and keep view controllers together only if they are part of the same screen (containers / child view controllers)...

@Scollaco I hear you on this one. The main reason for moving the `SettingsAccountViewController` to the `Settings` storyboard is that it now uses the message banner, which is set up as a child view controller in the `Settings` storyboard and also used by the `Change Email Scene` and the `Change Password Scene`. I realize that this means that the message banner is probably too coupled to the `Settings` storyboard, so I intend to remove it from there in a separate item of work in the next sprint (see my comment in https://github.com/kickstarter/ios-oss/pull/488).

Here is the Trello card for that work: https://trello.com/c/J7mBSotV/580-remove-messagebannerviewcontroller-from-the-settings-storyboard

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
ifbarrera: Swiftlint was complaining about unused optional binding but I can silence

Scollaco:
@ifbarrera  Oh cool, I missed your comment on the other PR. Yeah, I think that by setting the position programatically will allow us to use the banner wherever we want, making it more reusable 👍

justinswart:
> @justinswart in this case we have two cells that depend on content from the server - the chosen currency cell and now the `Change Email` cell with this new alert icon. IMO we should minimize the number of requests made on load of a single view controller whenever possible to reduce networking hardware overhead.

Totally makes sense here, just wanted to understand where and why we'd done it, thanks for clarifying 👍

PR #484, opened 2018-11-15
Change email updates

ifbarrera:
# What

- fixes a bug where the `Current email` label would get truncated with v long emails
- makes the change email & change password screens bounce
- updates the keyboard type of the `New email` field to `emailAddress`

# See 👀

simulator screen shot - iphone 8 - 2018-11-15 at 15 53 56

5uhrgtbham


ifbarrera:
[Email update: unverified email iOS](https://trello.com/c/f8sPHqRS/1056-email-update-unverified-email-ios)

justinswart:
Lgtm, just thought I'd mention that I think we can also have the password field work with iOS 12's password management feature as part of that work so that users can generate new passwords.

dusi:
@ifbarrera I have approved this PR but please consider looking at #488 prior to merging this to `master`. There still seems to be an issue with scrolling on both `change email` & `change password` screens.

PR #486, opened 2018-11-16
Bug fixes - QA feedback

Scollaco:
# What
Fixed bugs found during our QA process:
1- Change email request being made 2x,
2- Go button enabled even if inputs were not valid,
3- Email allowing emails with invalid formats

# How
The `Go` return type was replaced with the `Done` button. This will make the keyboard dismiss, simplifying the logic, and by tapping the `Save` button will be the only way to make the request.

I also replaced the regex used for the email validation with one on this [article](http://www.cocoawithlove.com/2009/06/verifying-that-string-is-email-address.html).
According to the article, `This regular expression is adapted from a version at regular-expressions.info and is a complete verification of RFC 2822.`. This regex validates a much larger number of email formats.

GIF


Library/ViewModels/ChangeEmailViewModel.swift
```diff
@@ -257,6 +241,5 @@ ChangeEmailViewModelOutputs {
 }
 
 private func shouldEnableSaveButton(email: String?, newEmail: String, password: String) -> Bool {
```
justinswart: Do we allow space-only passwords? Otherwise we might want to trim these too:

```suggestion
private func shouldEnableSaveButton(email: String?, newEmail: String, password: String) -> Bool {
  guard email != newEmail else { return false }

  return ![newEmail, password]
    .map { !$0.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }
    .contains(false)
}
```

Library/ViewModels/ChangeEmailViewModel.swift
```diff
@@ -207,14 +200,6 @@ ChangeEmailViewModelOutputs {
...
-  private let changePasswordProperty = MutableProperty<(String, String)?>(nil)
-  public func submitForm(newEmail: String?, password: String?) {
-
-    if let newEmail = newEmail, let password = password {
-      self.changePasswordProperty.value = (newEmail, password)
-    }
-  }
-
   private let saveButtonTappedProperty = MutableProperty<(String, String)?>(nil)
   public func saveButtonTapped(newEmail: String?, password: String?) {
```
justinswart: We shouldn't have any logic in the input function itself, it should be handled via signals.

Also - I don't think these values should be passed in this way, we're setting them on every keypress so we should be able to just have combine those signals and grab the values on our save button press using `takeWhen`.

Scollaco:
Yeah, about the 1Password button, it's beyond the scope of this PR, but it's a good idea to bring this up on the next sprints and implement it on a separate PR. Thanks for your feedback, @justinswart !

Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
```diff
@@ -251,9 +241,17 @@ internal final class ChangeEmailViewController: UIViewController {
   }
 
   private func resetFields(string: String) {
-    _ = [self.passwordTextField, self.newEmailTextField]
-      ||> UITextField.lens.text .~ string
```
justinswart: Why this change?

Scollaco: I just got rid of lenses here, and I remember I saw a warning once saying that, to set texts, it's preferred to use the `%~` operator over `.~`.

justinswart: Ah right - so in this case that isn't necessary because the string is not being evaluated for localization. In other places we do this when referring to `Strings.*` to be sure that we're evaluating the closure against any changes to the environment.

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -19,14 +19,13 @@ final class ChangeEmailViewModelTests: TestCase {
...
-  private let passwordFieldBecomeFirstResponder = TestObserver<Void, NoError>()
-  private let resetFields = TestObserver<String, NoError>()
-  private let shouldSubmitForm = TestObserver<Void, NoError>()
   private let messageLabelViewHiddenObserver = TestObserver<Bool, NoError>()
+  private let passwordFieldBecomeFirstResponder = TestObserver<Void, NoError>()
   private let passwordText = TestObserver<String, NoError>()
   private let resendVerificationEmailViewIsHiddenObserver = TestObserver<Bool, NoError>()
+  private let resetFields = TestObserver<String, NoError>()
   private let saveButtonIsEnabled = TestObserver<Bool, NoError>()
+  private let textFieldAreEnabled = TestObserver<Bool, NoError>()
```
justinswart: This property name is missing an `s`

Scollaco:
[Change Email Implementation](https://trello.com/c/gaJRDQz4/869-change-email-implementation)

PR #488, opened 2018-11-20
Hide message banner superview properly

dusi:
# What

Addresses a small issue with the message banner's container view preventing it from scrolling (or any user interaction) on the `change email` & `change password` screens.

# Why

#484 has properly added elasticity to both `change email` & `change password` screens. Unfortunately due to the fact that the banner container view is presented above the scroll view it can intercept touches and prevent the scroll view to scroll.

This can be very easily reproduced by trying to swipe in the bottom area (where the container is overlaying the scroll view). Since the container is invisible it might appear like nothing should block the touch events, but it actually is.

# How

In order to fix this we can simply set the `isHidden` property on the container. To follow iOS standards this can be done using a `delegate` call from it's child. We can simply do this in the child controller (`MessageBannerViewController`) but to be a good citizen and follow best practices we need to do this in the parent `ChangeEmailViewController` & `ChangePasswordViewController` that has a control of the container view.

# See 👀

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
@@ -7,6 +7,7 @@ import UIKit
 internal final class ChangeEmailViewController: UIViewController {
   @IBOutlet fileprivate weak var currentEmailLabel: UILabel!
   @IBOutlet fileprivate weak var currentEmail: UILabel!
+  @IBOutlet fileprivate weak var messageBannerContainer: UIView!
   @IBOutlet fileprivate weak var messageLabelView: UIView!
   @IBOutlet fileprivate weak var newEmailLabel: UILabel!
   @IBOutlet fileprivate weak var newEmailTextField: UITextField!
@@ -35,6 +36,7 @@ internal final class ChangeEmailViewController: UIViewController {
       fatalError("Couldn't instantiate MessageBannerViewController")
     }
     self.messageBannerViewController = messageBannerViewController
+    self.messageBannerViewController.delegate = self
 
     self.saveButtonView = LoadingBarButtonItemView.instantiate()
     self.saveButtonView.setTitle(title: Strings.Save())
@@ -264,5 +266,11 @@ extension ChangeEmailViewController: UITextFieldDelegate {
 
     self.viewModel.inputs.textFieldShouldReturn(with: textField.returnKeyType)
     return textField.resignFirstResponder()
+  }
+}
+
+extension ChangeEmailViewController: MessageBannerViewControllerDelegate {
+  func messageBannerViewControllerIsHidden(_ isHidden: Bool) {
+    self.messageBannerContainer.isHidden = isHidden
   }
 }
--- a/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
@@ -20,7 +20,7 @@ internal final class ChangeEmailViewController: UIViewController {
   @IBOutlet fileprivate weak var warningMessageLabel: UILabel!
 
   private let viewModel: ChangeEmailViewModelType = ChangeEmailViewModel()
-  private var messageBannerView: MessageBannerViewController!
+  private var messageBannerViewController: MessageBannerViewController!
 
   private weak var saveButtonView: LoadingBarButtonItemView!
 
@@ -31,11 +31,10 @@ internal final class ChangeEmailViewController: UIViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    guard let messageBannerView = self.children.first as? MessageBannerViewController else {
+    guard let messageBannerViewController = self.children.first as? MessageBannerViewController else {
       fatalError("Couldn't instantiate MessageBannerViewController")
     }
-
-    self.messageBannerView = messageBannerView
+    self.messageBannerViewController = messageBannerViewController
 
     self.saveButtonView = LoadingBarButtonItemView.instantiate()
     self.saveButtonView.setTitle(title: Strings.Save())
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
```diff
@@ -20,7 +21,7 @@ internal final class ChangeEmailViewController: UIViewController {
   @IBOutlet fileprivate weak var warningMessageLabel: UILabel!
 
   private let viewModel: ChangeEmailViewModelType = ChangeEmailViewModel()
-  private var messageBannerView: MessageBannerViewController!
+  private var messageBannerViewController: MessageBannerViewController!
```
dusi: Renaming to `...Controller` so that it's not confused with a `view`

Kickstarter-iOS/Views/Controllers/ChangeEmailViewControllerTests.swift
```diff
@@ -80,4 +81,38 @@ final class ChangeEmailViewControllerTests: TestCase {
       }
     }
   }
+
+  func testMessageBannerContainerIsHiddenByDefault() {
+    let controller = ChangeEmailViewController.instantiate()
+    _ = controller.view
```
dusi: Causes `viewDidLoad` to be called

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -370,7 +370,7 @@
                                     <constraint firstItem="VyE-lE-aLQ" firstAttribute="leading" secondItem="XfD-wg-Nx4" secondAttribute="leading" id="pUW-P0-AFT"/>
                                 </constraints>
                             </view>
-                            <containerView opaque="NO" contentMode="scaleToFill" ambiguous="YES" translatesAutoresizingMaskIntoConstraints="NO" id="Z5v-98-lpx">
+                            <containerView hidden="YES" opaque="NO" contentMode="scaleToFill" ambiguous="YES" translatesAutoresizingMaskIntoConstraints="NO" id="Z5v-98-lpx">
```
dusi: Hidden by default

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -10,13 +10,14 @@ final class ChangePasswordViewController: UIViewController {
   @IBOutlet fileprivate weak var currentPasswordLabel: UILabel!
   @IBOutlet fileprivate weak var currentPasswordTextField: UITextField!
   @IBOutlet fileprivate weak var validationErrorMessageLabel: UILabel!
+  @IBOutlet fileprivate weak var messageBannerContainer: UIView!
```
justinswart: Maybe `messageBannerViewControllerContainer`?

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -621,7 +622,7 @@
                                     <constraint firstAttribute="trailing" secondItem="Jfc-kg-fPd" secondAttribute="trailing" id="ydu-rL-nNN"/>
                                 </constraints>
                             </scrollView>
-                            <containerView opaque="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Orx-BR-Kot">
+                            <containerView hidden="YES" opaque="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Orx-BR-Kot">
```
justinswart: Potentially this one too?

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
dusi: I've actually decided to use protocol extension to make the code more DRY and with that I think it's more appropriate for the protocol to have its variable be called `messageBannerViewControllerContainer`. In order to prevent from naming collisions I've decided to keep `messageBannerContainer` the same as it's now (mostly for clarity and simplicity) and will be wrapping it in the protocol like:
```
extension ChangeEmailViewController: MessageBannerViewControllerDelegate {
  var messageBannerViewControllerContainer: UIView {
    return self.messageBannerContainer
  }
}
```

justinswart: But why wouldn't the `ChangeEmailViewController` in this case just call its container `messageBannerViewControllerContainer` and in that way it's already conforming to the protocol?

dusi: I couldn't make the compiler happy...it's probably something to do with the fact that `messageBannerViewControllerContainer` is `IBOutlet weak` 🤷‍♂️

Kickstarter-iOS/Views/Controllers/ChangeEmailViewControllerTests.swift
```diff
@@ -80,4 +81,41 @@ final class ChangeEmailViewControllerTests: TestCase {
       }
     }
   }
+
+  func testMessageBannerContainerIsSetToHiddenByDefault() {
+    let controller = ChangeEmailViewController.instantiate()
+    controller.beginAppearanceTransition(true, animated: false)
```
dusi: Causes `viewWillAppear:` to be called which is where we're binding styles

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
dusi: The only way I was able to do this was to declare protocol as `@objc`:

```
@objc internal protocol MessageBannerViewControllerDelegate {
  var messageBannerContainer: UIView { get }
}
```
and then declare the `IBOutlet` as:
```
@IBOutlet internal var messageBannerContainer: UIView!
```
or by default
```
@IBOutlet var messageBannerContainer: UIView!
```

which causes SwiftLint to produce `Private Outlets Violation` warning 🤷‍♂️ 

This doesn't seem very safe to me and it seems more like fighting the framework so I'd prefer to keep it this way (for now).

dusi:
[Email update: unverified email iOS](https://trello.com/c/f8sPHqRS/1056-email-update-unverified-email-ios)

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
justinswart: Sounds good!

PR #489, opened 2018-11-21
Settings tracking vol 2

dusi:
# What

Adds tracking for the following events:
* Viewed Account
* Viewed Change Email
* Viewed Change Password
* Changed Email
* Changed Password
* Resent Verification Email 

# Why

Tracking the new settings screens #YOLO

# How

Viewed events are triggered on `viewDidAppear` in order for us to being able to observe navigating to the screen repeatedly on the navigation stack.
Action events are pretty straightforward

# Todo

- [ ] Double check events are being sent to `Looker`

# Acceptance criteria

- [x] Navigating to `Account` from Settings tracks `Viewed Account` in Looker
- [x] Navigating back to `Account` from `Change email` / `Change password` screens tracks `Viewed Account` in Looker
- [x] Navigating to `Change email` from `Account` tracks `Viewed Change Email` tracks `Viewed Change Email` in Looker
- [x] Navigating to `Change password` from `Account` tracks `Viewed Change Password` tracks `Viewed Change Password` in Looker
- [ ] Successfully changing email on the `Change email` screen tracks `Changed Email` in Looker
- [ ] Successfully changing password on the `Change password` screen tracks `Changed Password` in Looker
- [ ] Taping the re-send or send verification email button tracks `Resent Verification Email ` in Looker

### DIFF
--- a/Library/Koala/Koala.swift
+++ b/Library/Koala/Koala.swift
@@ -438,7 +438,7 @@ public final class Koala {
     self.track(event: "Triggered 1Password")
   }
 
-  // MARK: Discovery Events
+  // MARK: - Discovery Events
 
   /**
    Call when a discovery search is made, including pagination.
@@ -534,7 +534,7 @@ public final class Koala {
       ])
   }
 
-  // MARK: Checkout Events
+  // MARK: - Checkout Events
   public func trackCheckoutCancel(project: Project,
                                   reward: Reward,
                                   pledgeContext: PledgeContext) {
@@ -630,7 +630,7 @@ public final class Koala {
     self.track(event: "Closed Reward", properties: props)
   }
 
-  // MARK: Login Events
+  // MARK: - Login Events
   public func trackLoginTout(intent: LoginIntent) {
     // Deprecated event
     self.track(event: "Application Login or Signup",
@@ -712,7 +712,7 @@ public final class Koala {
     self.track(event: "Resent Two-Factor Code")
   }
 
-  // MARK: Signup
+  // MARK: - Signup
 
   // Call when an error is returned after attempting to signup.
   public func trackSignupError(authType: AuthType) {
@@ -738,7 +738,7 @@ public final class Koala {
     self.track(event: "Viewed Signup")
   }
 
-  // MARK: Comments Events
+  // MARK: - Comments Events
   public func trackLoadNewerComments(project: Project, update: Update?, context: CommentsContext) {
     let props = properties(project: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(update.map { properties(update: $0) } ?? [:])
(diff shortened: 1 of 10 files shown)


Library/Koala/Koala.swift
```diff
@@ -438,7 +438,7 @@ public final class Koala {
     self.track(event: "Triggered 1Password")
   }
 
-  // MARK: Discovery Events
+  // MARK: - Discovery Events
```
dusi: It was almost impossible to navigate `Koala.swift`'s jump bar in Xcode so I've decided to jump the gun and update our pragma marks. This adds a separator between each section.

dusi:
@Scollaco I still have to test this being send to `Looker` but you can start reviewing at this stage if you want!

Library/Koala/KoalaTests.swift
```diff
@@ -614,4 +614,49 @@ final class KoalaTests: TestCase {
     XCTAssertEqual("Apple", client.properties.last?["manufacturer"] as? String)
     XCTAssertEqual("Apple", callBackProperties?["manufacturer"] as? String)
   }
+
+  func testTrackViewedAccount() {
+    let client = MockTrackingClient()
```
Scollaco: Since `KoalaTests` inherits from `TestCase` class, I wonder if we need to initialize a client for every test instead of using its parent's `trackingClient`. 🤔

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -367,4 +367,56 @@ final class ChangeEmailViewModelTests: TestCase {
     self.vm.inputs.saveButtonTapped(newEmail: "", password: "")
     self.shouldSubmitForm.assertValueCount(2)
   }
+
+  func testTrackViewedChangeEmail() {
+    let client = MockTrackingClient()
```
Scollaco: Same thing here. Maybe we can just use the existing `trackingClient`.

Library/Koala/KoalaTests.swift
justinswart: I think because it's a class you'll likely find the `client.events` overlapping each other and assertions may fail.

dusi: Looks like we can use `trackingClient` from TestCase and then re-use it in each test like:
```
let koala = Koala(client: self.trackingClient)
```
and it would still work (since the client is being instantiated for each test separately.

However I think for now we should be good to go with this solution which is consistent with the rest of the test suite in `KoalaTests` where we instantiate the client for each test. 

I'm happy to re-factor this but currently it represents 37 instances of us instantiating client in each test like `let client = MockTrackingClient()` which could add up to this PR and make the codebase even more inconsistent with other tests.

dusi:
[Settings Rollout 2 Event Tracking](https://trello.com/c/sW631378/1077-settings-rollout-2-event-tracking)

PR #495, opened 2018-11-24
Refactor push registration

justinswart:
# 📲 What

This PR hopes to significantly simplify the way that we register for push notifications in the app.

# 🤔 Why

When I started working on #490 I was just really confused by how things were working. Also there were two scenarios that I think weren't working correctly:

- on iOS 9 I think users were being prompted to allow push notifications on app launch (based on reading the code).
- on iOS 10+ users that had previously allowed push notifications would never see badges because we would not request those permissions again.

So in trying to solve these two issues I decided to try and refactor things.

# 🛠 How

The main simplification is that by using `SignalProducer`s and wrapping the push registration code inside these we can deal with iOS 9 registration and iOS 10+ registration at the callsite without caring about which iOS version we're on. We can do all of the `@available` checks inside these `SignalProducer`s and as well as handle the fact that sometimes these calls are synchronous and that sometimes they're asynchronous. So this PR:

- introduces `PushNotificationType`
  - an `enum` to type-erase `UIUserNotificationType` and `UNAuthorizationOptions`.
- `PushRegistrationType`, a `protocol` providing two functions that return `SignalProducer`s:
  - `register(for types: [PushNotificationType])` to wrap the registration process and return `true` or `false` for iOS 10+ and `nil` for iOS 9 as we do not know the user action on iOS 9.
  - `currentAuthorization()` to return whether or not we are currently authorized for push notifications. This is used to decide whether to re-register our permissions at app launch.
- strips out all of the various inputs/outputs from `AppDelegateViewModel` that made things so confusing and instead bases registration on one of two scenarios:
  - the user said _yes_ to the prompt asking them whether they want to receive pushes.
  - the user previously allowed push notifications and we refresh that registration on app launch.
- **NB:** This PR is based off #490 to make it easier to review. This should merge before that PR.

# ✅ Acceptance criteria

- [x] All of the scenarios described in #490.
- [x] Test that Koala tracks the same events.

And now for my favourite refactoring gif:


justinswart:
@ifbarrera @Scollaco There may be edge cases I haven't considered that you've encountered before so please let me know if I've missed things!

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -127,46 +127,18 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
       self.viewModel.outputs.pushTokenSuccessfullyRegistered
       .observeForUI()
-      .observeValues {
-        print("📲 [Push Registration] Push token successfully registered ✨")
+      .observeValues { token in
+        print("📲 [Push Registration] Push token successfully registered (\(token)) ✨")
     }
 
-      if #available(iOS 10.0, *) {
```
justinswart: Moved to `PushRegistration`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -73,13 +73,6 @@ public protocol AppDelegateViewModelInputs {
   /// Call when the redirect URL has been found, see `findRedirectUrl` for more information.
   func foundRedirectUrl(_ url: URL)
 
-  /// Call when notification authorization is completed with user permission or denial.
```
justinswart: All of these removed inputs/outputs are in `SignalProducer`s now.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -256,51 +243,24 @@ AppDelegateViewModelOutputs {
...
-        self.didAcceptReceivingRemoteNotificationsProperty.signal,
-        userHasAuthorizedRemoteNotifications.ignoreValues()
-      )
-    } else {
-      //Never firing signal in ios 9
-      self.getNotificationAuthorizationStatus = .empty
-      self.authorizeForRemoteNotifications = .empty
-
-      self.registerForRemoteNotifications = applicationIsReadyForRegisteringNotifications
+    let pushTokenRegistrationStartedEvents = Signal.merge(
```
justinswart: The only two scenarios that we trigger push registration on.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -313,11 +273,11 @@ AppDelegateViewModelOutputs {
     self.pushTokenSuccessfullyRegistered = self.deviceTokenDataProperty.signal
       .map(deviceToken(fromData:))
       .ksr_debounce(.seconds(5), on: AppEnvironment.current.scheduler)
-      .switchMap {
-        AppEnvironment.current.apiService.register(pushToken: $0)
+      .switchMap { token in
+        AppEnvironment.current.apiService.register(pushToken: token)
           .demoteErrors()
+          .map { _ in token }
```
justinswart: Returning the token here for easy debugging in the console.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -626,19 +586,21 @@ AppDelegateViewModelOutputs {
...
+    )
+    .filter { _, previouslyAuthorized in !previouslyAuthorized }
+    .map { isGranted, _ in isGranted }
+    .skipNil()
+    .take(first: 1)
+    .observeValues { isGranted in
+      if isGranted {
+        AppEnvironment.current.koala.trackPushPermissionOptIn()
+      } else {
+        AppEnvironment.current.koala.trackPushPermissionOptOut()
```
justinswart: These Koala events should track the same as before, no need to do the `@available` check anymore as that's handled in `PushRegistration`.

Library/AppEnvironment.swift
```diff
@@ -124,15 +124,14 @@ public struct AppEnvironment {
     debounceInterval: DispatchTimeInterval = AppEnvironment.current.debounceInterval,
     device: UIDeviceType = AppEnvironment.current.device,
     facebookAppDelegate: FacebookAppDelegateProtocol = AppEnvironment.current.facebookAppDelegate,
-    isRegisteredForPushNotifications: SignalProducer<Bool, NoError> =
```
justinswart: Replaced by `PushRegistrationType`.

Library/MockPushRegistration.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+@testable import Library
+import ReactiveSwift
+import Result
+
+public struct MockPushRegistration: PushRegistrationType {
```
justinswart: Pretty easy to mock!

Library/PushRegistrationType.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import ReactiveSwift
+import Result
+import UIKit
+import UserNotifications
+
+public enum PushNotificationType: CaseIterable {
```
justinswart: An `enum` that allows us to not care about `UIUserNotificationType` or `UNAuthorizationOptions` at the callsite.

justinswart:
[iOS Notification Badges](https://trello.com/c/tWdsuKpU/827-ios-notification-badges)

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
ifbarrera: Can we return this token even if the registration API request fails? (It tends to fail all the time when running on a debug build)/

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -687,9 +649,9 @@ AppDelegateViewModelOutputs {
     deepLinkFromNotification
       .observeValues { _ in AppEnvironment.current.koala.trackNotificationOpened() }
 
-    self.applicationIconBadgeNumber = self.applicationLaunchOptionsProperty.signal
+    self.applicationIconBadgeNumber = self.applicationWillEnterForegroundProperty.signal
```
ifbarrera: Can we merge `applicationLaunchOptionsProperty` with `applicationWillEnterForegroundProperty` so the badge is cleared both from fresh launches and background launches?

Library/AppEnvironment.swift
```diff
@@ -124,15 +124,14 @@ public struct AppEnvironment {
...
-      AppEnvironment.current.isRegisteredForPushNotifications,
     isVoiceOverRunning: @escaping (() -> Bool) = AppEnvironment.current.isVoiceOverRunning,
     koala: Koala = AppEnvironment.current.koala,
     language: Language = AppEnvironment.current.language,
     launchedCountries: LaunchedCountries = AppEnvironment.current.launchedCountries,
     liveStreamService: LiveStreamServiceProtocol = AppEnvironment.current.liveStreamService,
     locale: Locale = AppEnvironment.current.locale,
     mainBundle: NSBundleType = AppEnvironment.current.mainBundle,
+    pushRegistrationType: PushRegistrationType.Type = AppEnvironment.current.pushRegistrationType,
```
ifbarrera: Does it make more sense to declare this as `pushRegistration: PushRegistrationType.Type`? The object is a `pushRegistration` object of type `PushRegistrationType`. Then we can call it like `AppEnvironment.current.pushRegistration.register()`

Library/PushRegistration.swift
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   Returns a signal producer that emits a `Bool` value representing whether the user has allowed push
+   notification permissions in the past.
+   The returned producer emits once and completes.
+
+   - returns: A signal producer.
+   */
+  public static func currentAuthorization() -> SignalProducer<Bool, NoError> {
```
ifbarrera: Hmm I feel like this naming is a bit confusing because the Producer emits a value that corresponds to whether `isAuthorized == true`, not the current authorization type. Could this be more clear as `hasAuthorized()` or `hasAuthorizedNotifications()`?

Library/AppEnvironment.swift
justinswart: Well I chose to use the type because there doesn't really need to be any instance/state retained for this, they're just functions that execute and end. The type is just to group the functions and make it easier to mock behind the protocol.

Library/PushRegistration.swift
justinswart: Yeah that makes sense, I'll rename!

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -312,12 +272,13 @@ AppDelegateViewModelOutputs {
 
     self.pushTokenSuccessfullyRegistered = self.deviceTokenDataProperty.signal
       .map(deviceToken(fromData:))
+      .on(value: { print("📲 [Push Registration] Push token generated: (\($0))") })
```
justinswart: I've added this as a side-effect so this will print regardless of the call to the back-end success/failure

Library/AppEnvironment.swift
ifbarrera: Fair enough 👍

PR #496, opened 2018-11-26
Payment methods event tracking

Scollaco:
# 📲 What
- Payment Methods and chosen currency events tracking.

# 🤔 Why
- To understand our users' behaviour in this particular part of the app and see if the features are relevant or can be improved.

# 🛠 How
For events that trigger if user saw the views will be send on `viewDidLoad`. For events triggered after an action, events will be sent accordingly.

# ✅ Acceptance criteria

Acceptance criteria
~~- [ ]  Successfully changing currency on `Account` tracks `Selected Chosen Currency`~~  (Part of this [PR](https://github.com/kickstarter/ios-oss/pull/509))
- [x]  Navigating to `Account > Payment methods` tracks `Viewed Payment Methods` in Looker
- [x]  Successfully deleting payment methods tracks `Deleted Payment Method` in Looker
- [x]  If deleting payment methods fails, we track `Errored Delete Payment Method` in Looker
- [x]  Navigating to `Account > Payment methods > Add New Card` tracks `Viewed Add New Card` in Looker
- [x]  Successfully adding a new payment method tracks `Saved Payment Method` in Looker
- [x]  If adding a new payment method fails, we track `Failed Payment Method Creation` in Looker

GIF


Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -63,6 +63,24 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.goToAddCardScreen = self.didTapAddCardButtonProperty.signal
 
     self.tableViewIsEditing = self.editButtonTappedSignal.scan(false) { current, _ in !current }
+
+    // Koala:
+    self.viewDidLoadProperty.signal
```
justinswart: Wondering if this event should be on `viewDidAppear` 🤔

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -115,4 +115,47 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
+    XCTAssertEqual(["Viewed Payment Methods"], self.trackingClient.events)
+  }
+
+  func testTrackDeletePaymentMethods() {
+
+    guard let card = GraphUserCreditCard.template.storedCards.nodes.first else {
+      XCTFail("Card should exist")
+      return
+    }
+      let apiService = MockService(deletePaymentMethodResult: .success(GraphMutationEmptyResponseEnvelope()))
```
justinswart: Slight mis-alignment here

Library/ViewModels/PaymentMethodsViewModel.swift
Scollaco: Many of our `Viewed` tracking events are triggered on `viewDidLoad`. I could confirm on Looker that the events are being sent, though. 
The difference is that on `viewDidLoad` the event will only be triggered once, while on `viewDidAppear` it can be triggered more than once, for example once when we open the view and again case we dismiss a child view controller, if any. I wonder which case we would consider as ideal 🤔

justinswart: From conversation with @terminalbae the expectation is that we track this each time the user navigates to the view. So `viewDidAppear` is the correct event. I think that there are some occasions where we use `viewWillAppear` and we actually filter based on `animated` being `false` - [example](https://github.com/kickstarter/ios-oss/blob/master/Library/ViewModels/ActivitiesViewModel.swift#L270). I don't know for certain as there doesn't seem to be documentation on this, but I would guess that this is being done because of view controller containment and that when a VC appears in a child container then `animated` is `false` and that we're tracking that specifically.

I should probably just test this to be sure - so let me see if I can do that real quick.

Scollaco: Thank you for your investigation. So I will change to `viewWillAppear` 👍

Library/Koala/Koala.swift
```diff
@@ -1824,6 +1824,35 @@ public final class Koala {
     }
   }
 
+  public func trackChangedCurrency(_ currency: Currency) {
```
ifbarrera: Just an idea, but this `Koala.swift` file is a 2000 line beast of a file. Should we consider breaking it up into extensions by feature?

Scollaco: Yeah, I like this idea! I would do it in other PR, though, but would be happy to refactor `Koala.swift`

Scollaco:
[Settings Rollout 3 Event Tracking](https://trello.com/c/Lm9qOiZK/1086-settings-rollout-3-event-tracking)

Kickstarter-iOS/Views/Cells/SettingsCurrencyPickerCell.swift
```diff
@@ -54,7 +54,7 @@ extension SettingsCurrencyPickerCell: UIPickerViewDataSource {
 
 extension SettingsCurrencyPickerCell: UIPickerViewDelegate {
   func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
-    return Currency.allCases[row].descriptionText //Currency(rawValue: row)?.descriptionText
+    return Currency.allCases[row].descriptionText
```
dusi: Good job killing this...I removed this on the `master` branch too :) 
giphy

Library/Koala/Koala.swift
```diff
@@ -1824,6 +1824,30 @@ public final class Koala {
     }
   }
 
+  public func trackViewedPaymentMethods() {
```
dusi: I almost want to say we should rebase `master` so that these could be properly placed under `Settings` group. This would require rebasing both this branch and the parent feature branch.
What do you think?

Library/Koala/Koala.swift
```diff
@@ -1824,6 +1824,30 @@ public final class Koala {
...
+
+  public func trackViewedAddNewCard() {
+    self.track(event: "Viewed Add New Card")
+  }
+
+  public func trackDeletedPaymentMethod() {
+    self.track(event: "Deleted Payment Method")
+  }
+
+  public func trackErroredDeletePaymentMethod() {
```
dusi: Looks like the only other function in `KoalaSwift` that uses `trackErrorred....` format is `trackStripeTokenErroredForApplePay` everything else prefers `trackDeletePaymentMethodError` format. Should we unify this and rename this to `trackDeletePaymentMethodError`?

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -92,6 +93,22 @@ AddNewCardViewModelOutputs {
       self.addNewCardSuccess.mapConst(false),
       self.addNewCardFailure.mapConst(false)
     )
+
+    // Koala
+    self.viewWillAppearProperty.signal
+      .observeValues {
+        AppEnvironment.current.koala.trackViewedAddNewCard()
```
dusi: Is there a reason for these 3 tracking events to be on a separate line? Should fit on 1 line..

```
self.viewWillAppearProperty.signal
      .observeValues { AppEnvironment.current.koala.trackViewedAddNewCard() }

    self.addNewCardSuccess
      .observeValues { _ in AppEnvironment.current.koala.trackSavedPaymentMethod() }

    self.addNewCardFailure
      .observeValues { _ in AppEnvironment.current.koala.trackFailedPaymentMethodCreation() }
```

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -220,4 +220,33 @@ internal final class AddNewCardViewModelTests: TestCase {
     self.cardExpYear.assertValues([99])
     self.cardCVC.assertValues(["123"])
   }
+
+  func testTrackViewedAddNewCard() {
+
```
dusi: Do we need these empty lines in these tests?

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -68,6 +68,24 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.presentBanner = self.presentMessageBannerProperty.signal
 
     self.tableViewIsEditing = self.editButtonTappedSignal.scan(false) { current, _ in !current }
+
+    // Koala:
+    self.viewWillAppearProperty.signal
+      .observeValues { _ in
```
dusi: Same 1 liner opportunity here

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -126,4 +126,47 @@ internal final class PaymentMethodsViewModelTests: TestCase {
       )
     }
   }
+
+  func testTrackPaymentMethodsView() {
+
```
dusi: Same empty lines

Library/ViewModels/PaymentMethodsViewModel.swift
dusi: If we track this in `viewWillAppear` do we want to filter on `animated` being `false`?

Scollaco: For events being tracked on `viewWillAppear` will only be filtered on `animated` being false is they are `RootTabBarViewController`'s child (such as Discovery, Activity, Dashboard...), this is because we are setting the value manually to false while configuring the tabBar's viewcontrollers (you can see it on `RootTabBarViewController.swift`)

Library/ViewModels/AddNewCardViewModel.swift
Scollaco: No reason :) Wil change it

Library/ViewModels/PaymentMethodsViewModel.swift
dusi: Looks like `setViewControllers($0, animated: false)` doesn't affect `viewWillAppear` at all.

Library/Koala/Koala.swift
dusi: I've already added some pragma marks to `Koala.swift` for better navigation so it should be slightly easier (those changes are on `master` branch

Library/Koala/KoalaTests.swift
```diff
@@ -590,4 +590,52 @@ final class KoalaTests: TestCase {
...
+
+  func testTrackDeletedPaymentMethod() {
+    let client = MockTrackingClient()
+    let koala = Koala(client: client)
+
+    koala.trackDeletedPaymentMethod()
+    XCTAssertEqual(["Deleted Payment Method"], client.events)
+  }
+
+  func testTrackErroredDeletePaymentMethod() {
```
dusi: We should also rename the test here. Feel free to simply accept the suggestion through Github

```suggestion
  func testTrackDeletePaymentMethodError() {
```

Scollaco: Ops. missed that one

PR #497, opened 2018-11-26
Replace screenshots with wordmark

justinswart:
# 📲 What

Just a suggestion - that we remove the (outdated) screenshots in the `README.md` and add our Kickstarter wordmark for some colour.

# 🤔 Why

The current screenshots are outdated (and pretty pixelated). Until we have our screenshots automated I think we should rather remove them for now. I think the wordmark looks good at the top of the readme but I'm open to other options!

# 🛠 How

Uploaded an SVG of the wordmark, set it to full width.

# 👀 See

Preview how it looks [here](https://github.com/kickstarter/ios-oss/tree/add-wordmark).

# ✅ Acceptance criteria

- [ ] Everyone agrees 😄


dusi:
👍

PR #499, opened 2018-11-28
Refactor test observer naming for consistency

dusi:
# 📲 What

Refactor test observer naming

# 🤔 Why

We should name test observers consistently based on VM's properties

# 🛠 How

Mostly just removes `...Observer` suffix, but there was already 1 minor change to a VM who was using this suffix for no real reason.
I've made sure that test observer instances reflect their counterparts in VMs (so the test observers should now have the same name as VM signal)

# ✅ Acceptance criteria

- [x] All tests pass

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/VideoViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/VideoViewController.swift
@@ -88,10 +88,10 @@ public final class VideoViewController: UIViewController {
         }, completion: nil)
     }
 
-    self.viewModel.outputs.addCompletion
+    self.viewModel.outputs.addCompletionObserver
       .observeForUI()
       .observeValues { [weak self] time in
-        self?.addCompletion(atTime: time)
+        self?.addCompletionObserver(atTime: time)
     }
 
     self.viewModel.outputs.configurePlayerWithURL
@@ -152,7 +152,7 @@ public final class VideoViewController: UIViewController {
     }
   }
 
-  func addCompletion(atTime time: CMTime) {
+  func addCompletionObserver(atTime time: CMTime) {
     guard let player = self.playerController.player else { return }
 
     self.timeObserver = player.addBoundaryTimeObserver(
--- a/Kickstarter-iOS/Views/Controllers/VideoViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/VideoViewController.swift
@@ -88,10 +88,10 @@ public final class VideoViewController: UIViewController {
         }, completion: nil)
     }
 
-    self.viewModel.outputs.addCompletionObserver
+    self.viewModel.outputs.addCompletion
       .observeForUI()
       .observeValues { [weak self] time in
-        self?.addCompletionObserver(atTime: time)
+        self?.addCompletion(atTime: time)
     }
 
     self.viewModel.outputs.configurePlayerWithURL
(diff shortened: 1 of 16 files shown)


Kickstarter-iOS/Views/Controllers/VideoViewController.swift
```diff
@@ -88,10 +88,10 @@ public final class VideoViewController: UIViewController {
         }, completion: nil)
     }
 
-    self.viewModel.outputs.addCompletionObserver
```
dusi: Remove `..Observer` suffix from the VM to stay consistent with the rest of the codebase

Kickstarter-iOS/Views/Controllers/VideoViewController.swift
```diff
@@ -152,7 +152,7 @@ public final class VideoViewController: UIViewController {
     }
   }
 
-  func addCompletionObserver(atTime time: CMTime) {
+  func addCompletion(atTime time: CMTime) {
```
justinswart: I would say that this should still have `Observer` in the name, the function is adding the completion observer.

Library/ViewModels/ChangePasswordViewModelTests.swift
```diff
@@ -10,38 +10,38 @@ import XCTest
...
+  private let onePasswordButtonIsHidden = TestObserver<Bool, NoError>()
+  private let onePasswordFindPasswordForURLString = TestObserver<String, NoError>()
+  private let saveButtonIsEnabled = TestObserver<Bool, NoError>()
+  private let validationErrorLabelIsHidden = TestObserver<Bool, NoError>()
+  private let validationErrorLabelMessage = TestObserver<String, NoError>()
 
   override func setUp() {
     super.setUp()
 
-    self.vm.outputs.activityIndicatorShouldShow.observe(activityIndicatorShouldShowObserver.observer)
```
justinswart: I know it didn't have before but maybe you could add `self` to all of these?

Library/ViewModels/FindFriendsCellViewModelTests.swift
```diff
@@ -12,32 +12,32 @@ import Foundation
...
+  private let disabledDescriptionLabelIsHidden = TestObserver<Bool, NoError>()
+  private let isDisabled = TestObserver<Bool, NoError>()
 
   override func setUp() {
     super.setUp()
 
     self.vm.outputs.disabledDescriptionLabelIsHidden
-      .observe(disabledDescriptionLabelIsHiddenObserver.observer)
-    self.vm.outputs.isDisabled.observe(isDisabledObserver.observer)
+      .observe(disabledDescriptionLabelIsHidden.observer)
```
justinswart: `self`'s here too. SwiftFormat 😅

Library/ViewModels/FindFriendsFaceookConnectCellViewModelTests.swift
```diff
@@ -32,7 +32,7 @@ import Prelude
     vm.outputs.attemptFacebookLogin.observe(attemptFacebookLogin.observer)
     vm.outputs.facebookConnectCellTitle.observe(title.observer)
     vm.outputs.facebookConnectCellSubtitle.observe(subtitle.observer)
-    vm.outputs.hideCloseButton.observe(hideCloseButtonObserver.observer)
+    vm.outputs.hideCloseButton.observe(hideCloseButton.observer)
```
justinswart: Here too but now I'm wondering if it should be a separate PR.

Kickstarter-iOS/Views/Controllers/VideoViewController.swift
justinswart: Yeah this should still have `Observer` in the name too.

PR #500, opened 2018-11-29
Fix banner icon position, tint, and height

ifbarrera:
# 📲 What

The initial designs for the banner view had the icon on the right, but the final designs had it on the left. Moved it over, and fixed the tint that was missing when the image got changed from default to template. Also increased the height to be consistent with the designs.

# 🤔 Why

✨ Design Excellence Amirite ✨ 

# 👀 See

simulator screen shot - iphone x - 2018-11-28 at 17 40 39

# ✅ Acceptance criteria

- [x] You can see the banner by going to change password, enter an incorrect password, hit "save". Check that icon is on the left, and the banner is the right height.


dannyalright:
thanks for the attention to detail!

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -14,6 +14,7 @@ public protocol MessageBannerViewModelOutputs {
   var bannerBackgroundColor: Signal<UIColor, NoError> { get }
   var bannerMessage: Signal<String, NoError> { get }
   var iconIsHidden: Signal<Bool, NoError> { get }
+  var iconTintColor: Signal<UIColor, NoError> { get }
```
Scollaco: Could you alphabetise here?

Library/ViewModels/MessageBannerViewModelTests.swift
```diff
@@ -12,6 +12,7 @@ internal final class MessageBannerViewModelTests: TestCase {
   let bannerBackgroundColorObserver = TestObserver<UIColor, NoError>()
   let bannerMessageObserver = TestObserver<String, NoError>()
   let iconIsHiddenObserver = TestObserver<Bool, NoError>()
+  let iconImageTintObserver = TestObserver<UIColor, NoError>()
```
Scollaco: And here

Scollaco:
Good job, @ifbarrera . I just left a few comments ⭐️

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -54,6 +54,14 @@ final class MessageBannerViewController: UIViewController {
         self?.showViewAndAnimate(isHidden)
     }
 
+    self.viewModel.outputs.iconTintColor
+      .observeForUI()
+      .observeValues { [weak self] color in
+        guard let `self` = self else { return }
```
justinswart: I think I once suggested capturing `self` weakly in this way but apparently is _incorrect_ as it exploits a bug in Swift. Apologies if I caused confusion!

https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160118/007425.html

Good news is, since Swift 4.2 you should be able to do:

```swift
```

https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md

Library/MessageBannerType.swift
```diff
@@ -27,6 +27,15 @@ public enum MessageBannerType {
     }
   }
 
+  var iconImageTintColor: UIColor? {
+    switch self {
+    case .error:
+      return UIColor.black
```
justinswart: Could be just `.black` if we want to use inference.

ifbarrera:
@Scollaco this is ready for re-review! 😅

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -57,9 +57,8 @@ final class MessageBannerViewController: UIViewController {
     self.viewModel.outputs.iconTintColor
       .observeForUI()
       .observeValues { [weak self] color in
-        guard let `self` = self else { return }
-        _ = self.iconImageView
-          |> UIImageView.lens.tintColor .~ color
+        _ = self?.iconImageView
+          ?|> UIImageView.lens.tintColor .~ color
```
Scollaco: Here I would use keyPaths, since the plan is to deprecate lenses in the future.

ifbarrera: Whoops I missed this. Will do!

PR #501, opened 2018-11-29
Music newsletters

Scollaco:
# 📲 What
Added Music Newsletter.

# 🤔 Why
https://trello.com/c/4fT8O2CI/1096-add-music-newsletter-to-our-settings

# 🛠 How
To add `Music Newsletter` I just needed to add a new value to the `Newsletter` enum and update the related files.
I also had to update `NewsletterSubscriptionsLenses`, because, in this case, replacing it by keyPath would increase considerably the scope of the PR.

# 👀 See

# ✅ Acceptance criteria

- [x] Navigating to `Settings > Newsletters` we should see `Kickstarter Music` between `Kickstarter Films` and `Kickstarter Reads`
- [x] Music newsletter should be disabled by default
- [x] Music newsletter should have title `Kickstarter Music` and the description `It's like the radio but nothing sucks and also it's a newsletter.`
- [x] After enabling `Kickstarter Music`, closing and reopening the app, the `Kickstarter Music` switch should be enabled.

# ⏰ TODO
- [ ] Translated copies

GIF


Library/Newsletter.swift
```diff
@@ -62,6 +55,8 @@ public enum Newsletter {
       return Strings.Welcome_to_our_library_Peruse_the_stacks_with_us()
     case .alumni:
       return Strings.profile_settings_newsletter_alumni_newsletter()
+    case .music:
+      return "It's like the radio but nothing sucks and also it’s a newsletter."
```
justinswart: I'd recommend putting this in a `localizedString()` function and use the expected `key` and `defaultValue`.  That way we can even merge this PR before the translations are done and later just run `make strings` and merge that in. Also, we can easily search the codebase for `localizedString` to know where it needs to be replaced in the event that we forget where we didn't use localization.

tl;dr I think it's a safer practice to use `localizedString(key:defaultValue:)` than just a raw string 👍

justinswart: lol I probably should've put the tl;dr at the top 😬

justinswart:
Thanks for adding the acceptance criteria!

Library/Newsletter.swift
Scollaco: lol I will change that

PR #502, opened 2018-11-30
Deprecate iOS 9

justinswart:
# 📲 What

This deprecates iOS 9 support.

# 🤔 Why

We have a tiny amount of users that are still on iOS 9, currently 0.6% of daily actives. The additional overhead of supporting iOS 9 simply isn't worth it so let's move with the times.

screen shot 2018-11-30 at 2 29 16 pm

# 🛠 How

- Set project target's minimum deployment to iOS 10.3 (latest version of minimum supported major version).
- Remove all `@available` and `#available` checks for iOS 10.0.
- Remove code deprecated in iOS 10.
- Update our notification handling to use `UNUserNotificationCenter`.
- Remove old code that did manual alert dialogues for app store ratings, we get this for free since iOS 10.
- Ensure that notifications are only handled via the `UNUserNotificationCenterDelegate` methods.
  - Notifications are no longer delivered via `application(_:didFinishLaunchingWithOptions:)` so I've removed the related `inputs` from `AppDelegateViewModel` and updated tests. - [reference](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/SchedulingandHandlingLocalNotifications.html#//apple_ref/doc/uid/TP40008194-CH5-SW14).

🚀 screen shot 2018-11-30 at 5 27 16 pm

# 👀 See

Something that's cool that we get for free is this new behaviour for push notifications while the app is open. Instead of the plain old alert dialog we now have the OS push banner:

| < iOS 10  | iOS 10+ |
|  | ezgif com-gif-maker |

# ✅ Acceptance criteria

- [x] Confirm that I haven't deleted code that we still need (happy to pair-review that).
- [x] Can receive and deep-link via remote notifications (app in background).
- [x] Can receive and deep-link via remote notifications (app in foreground).
- [x] Can rate on the app store at the expected time in `ThanksViewController`. This appears to be based on a bunch of criteria defined by the `shouldShowGamesAlert` signal in `ThanksViewModel`.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -407,3 +372,24 @@ extension AppDelegate: URLSessionTaskDelegate {
...
+    completionHandler(.alert)
+  }
+
+  public func userNotificationCenter(
+    _: UNUserNotificationCenter,
+    didReceive response: UNNotificationResponse,
+    withCompletionHandler completion: @escaping () -> Void
+    ) {
+    self.viewModel.inputs.didReceive(remoteNotification: response.notification.request.content.userInfo)
+    completion()
```
justinswart: I don't think it matters when we call this completion in our case 🤔

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -280,44 +274,10 @@ AppDelegateViewModelOutputs {
...
-    let remoteNotificationFromLaunch = self.applicationLaunchOptionsProperty.signal.skipNil()
-      .map { _, options in
-        options?[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any] }
-      .skipNil()
-
-    let localNotificationFromLaunch = self.applicationLaunchOptionsProperty.signal.skipNil()
-      .map { _, options in
-        options?[UIApplication.LaunchOptionsKey.localNotification] as? UILocalNotification }
-      .map { $0?.userInfo }
+    let deepLinkFromNotification = self.remoteNotificationProperty.signal.skipNil()
```
justinswart: This is a lot simpler now that we don't care about whether the app state is active.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -804,66 +802,12 @@ final class AppDelegateViewModelTests: TestCase {
 
   }
 
-  func testOpenPushNotification_WhileInForeground() {
```
justinswart: As far as I could tell we don't really need these tests anymore?

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1071,26 +985,14 @@ final class AppDelegateViewModelTests: TestCase {
...
+      self.vm.inputs.didReceive(remoteNotification: pushData)
 
       self.goToDashboard.assertValueCount(0)
       self.goToDiscovery.assertValueCount(0)
       self.presentViewController.assertValueCount(0)
     }
   }
 
-  func testOpenNotification_LocalNotification_FromLaunch() {
-    let localNotification = UILocalNotification()
```
justinswart: `UILocalNotification` is deprecated.

Kickstarter-iOS/Views/Controllers/LiveVideoViewController.swift
```diff
@@ -148,15 +148,7 @@ public final class LiveVideoViewController: UIViewController {
...
-    do {
-      if #available(iOS 10.0, *) {
-        try AVAudioSession.sharedInstance().setCategory(.playback,
-                                                        mode: .default,
-                                                        options: [])
-      } else {
-        // Apple removed the deprecated method!. Since iOS 12 was release, an idea is to end supporting iOS 9
-      }
-    } catch {}
+    try? AVAudioSession.sharedInstance().setCategory(.playback, mode: .default, options: [])
```
justinswart: Just doing `try?` here seeing as we weren't handling the error in `catch { }`.

Kickstarter-iOS/Views/Controllers/ThanksViewController.swift
```diff
@@ -214,22 +208,7 @@ internal final class ThanksViewController: UIViewController, UITableViewDelegate
...
-          }, remindHandler: { [weak self] _ in
-            self?.viewModel.inputs.rateRemindLaterButtonTapped()
-          }, noHandler: { [weak self] _ in
-            self?.viewModel.inputs.rateNoThanksButtonTapped()
-        }),
-        animated: true,
-        completion: nil
-      )
-    }
+    SKStoreReviewController.requestReview()
```
justinswart: > When you call this method in your shipping app and a rating/review request view is displayed, the system handles the entire process for you.

Apple docs: https://developer.apple.com/documentation/storekit/skstorereviewcontroller/2851536-requestreview

Library/PushRegistrationType.swift
```diff
@@ -4,49 +4,7 @@ import Result
 import UIKit
 import UserNotifications
 
-public enum PushNotificationType: CaseIterable {
```
justinswart: Removing this type required no changes at the callsite 🎉

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -167,19 +152,6 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
     self.showGamesNewsletterAlert
       .observeValues { AppEnvironment.current.userDefaults.hasSeenGamesNewsletterPrompt = true }
 
-    project
```
justinswart: Seeing as all of this is handled by `SKStoreReviewController.requestReview()` none of this tracking should be needed (nor can it be tracked as far as I can tell).

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
ifbarrera: Actually maybe we want to keep one of these as a test for the `Notification Opened` and `Opened Notification` events when the app is in the foreground?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -65,17 +65,14 @@ public protocol AppDelegateViewModelInputs {
...
+  func didReceive(remoteNotification notification: [AnyHashable: Any])
 
   /// Call when the app delegate gets notice of a successful notification registration.
   func didRegisterForRemoteNotifications(withDeviceTokenData data: Data)
 
   /// Call when the redirect URL has been found, see `findRedirectUrl` for more information.
   func foundRedirectUrl(_ url: URL)
 
-  /// Call when the user taps "OK" from the notification alert.
-  func openRemoteNotificationTappedOk()
```
dusi: Has been removed from the protocol but still looks like we haven't removed this from `AppDelegateViewModel`

```
fileprivate let openRemoteNotificationTappedOkProperty = MutableProperty(())
public func openRemoteNotificationTappedOk() {
  self.openRemoteNotificationTappedOkProperty.value = ()
}
```

justinswart: Good catch thanks!

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
justinswart: Hmm good question - let me take another look to see if it makes sense with the new APIs.

justinswart:
> Do you think we should hold on merging this until 3.10.0 goes out? It might be a good idea to ship this as a separate release to minimize changes on 3.10.0.

@ifbarrera Happy to do that, should be fairly easy to rebase this on that once we've released.

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -110,7 +110,7 @@ final class SettingsViewController: UIViewController {
 
   private func goToAppStore(link: String) {
     guard let url = URL(string: link) else { return }
-    UIApplication.shared.openURL(url)
+    UIApplication.shared.open(url)
```
dusi: I've tried searching the code base for `openURL(` and found `UIApplication.shared.openURL(url)` in `Library > SafariActivity.swift:perform()` which now causes compiler warning. Wanna take a look at it?

Library/UIAlertController.swift
```diff
@@ -84,39 +84,6 @@ public extension UIAlertController {
...
-                            remindHandler: @escaping ((UIAlertAction) -> Void),
-                            noHandler: @escaping ((UIAlertAction) -> Void)) -> UIAlertController {
-    let alertController = UIAlertController(
-      title: Strings.profile_settings_rating_title() ,
-      message: Strings.profile_settings_rating_message() ,
-      preferredStyle: .alert
-    )
-    alertController.addAction(
-      UIAlertAction(
-        title: Strings.profile_settings_rating_option_rate_now() ,
```
dusi: Looks like these 3 strings are no longer being used. Wanna get rid of those?

justinswart: Ah thanks, I'll have to open a PR in our strings for that 👍

PR #505, opened 2018-12-03
[Bug] Request data button

Scollaco:
# 📲 What
Fixed bug on which the `Request my personal data` cell on `Settings > Privacy` appeared blank with a visible arrow if the user opened that screen being offline.

# 🤔 Why
Because showing blank cells is not good, and also because of this:

https://trello.com/c/efHhPiz1/1090-p1-request-my-personal-data-cell-is-not-appearing-on-iphone-xs-max-unclear-on-other-phones

# 🛠 How
Instead of showing a blank cell while the API returns an User, I set the initial value to `Request my personal data` and kept the arrow hidden until the app is ready to download the data.

# 👀 See
| Before | After |
| --- | --- |
|||

# ✅ Acceptance criteria

## With wifi on
- [x] I go to `Settings > Account > Privacy` , I should see `Request my personal data` and the arrow (chevron) should be hidden.
- [x] Tap the cell, I should see the prompt to download data
- [x] After confirming the request, I should see an activity indicator and the text `Preparing your personal data...`
- [x] If i leave the screen and go back to it, I should see `Download your personal data` and the arrow visible

## With wifi off
- [x] I go to `Settings > Account`, I should see `Request my personal data` and the arrow (chevron) should be hidden.
- [x] Tap the cell, but nothing will happen. (This part could be improved once we implement the online offline error on the app).

GIF


dannyalright:
> kept the arrow hidden until the values are updated.

Shouldn't the arrow _never_ show? Maybe this is what you mean.

> Tap the cell, but nothing will happen. (This part could be improved once we implement the online offline error on the app).

Nice call. I think an activity indicator in place of the arrow would work. But yeah, different task.

Scollaco:
> > kept the arrow hidden until the values are updated.
> 
> Shouldn't the arrow _never_ show? Maybe this is what you mean.
> 
The arrow is been shown once the app is ready to download data. I couldn't find the screen on Abstract, but this is how it looks now: 


By tapping the cell we are redirected to safari. So should I just delete the arrow or keep it how it is?

Library/ViewModels/SettingsRequestDataCellViewModel.swift
```diff
@@ -68,18 +68,26 @@ public final class SettingsRequestDataCellViewModel: SettingsRequestDataCellView
...
     self.requestedDataExpirationDate = exportEnvelope.map {
         dateFormatter(for: $0.expiresAt, state: $0.state)
       }
 
-    self.dataExpirationAndChevronHidden = exportEnvelope
-      .map { $0.state == .expired || $0.expiresAt == nil || $0.dataUrl == nil ? true : false }
+    self.dataExpirationAndChevronHidden = Signal.merge(
+      self.awakeFromNibProperty.signal.mapConst(true),
+      exportEnvelope
+      .map { $0.state == .expired || $0.expiresAt == nil || $0.dataUrl == nil ? true : false })
```
justinswart: The ternary operator here: `? true : false` seems unnecessary because the result of the `||` tests will return a `Bool`, right?

Scollaco: That's true. I just moved that code and forgot to refactor it. Good catch!

Scollaco:
[P1: 'Request my personal data' cell is not appearing on iPhone XS Max, unclear on other phones](https://trello.com/c/efHhPiz1/1090-p1-request-my-personal-data-cell-is-not-appearing-on-iphone-xs-max-unclear-on-other-phones)

PR #506, opened 2018-12-04
Message banner refactor

ifbarrera:
# 📲 What

Moved the message banner out of the `Settings.storyboard` and into it's own `.xib` file, and added a protocol `MessageBannerPresenting` for easy configuration on any view controller. Also cleaned up the view model a bit by combining a couple of the inputs into one `update(with:)` input that takes a typealias-ed tuple.

# 🤔 Why

Before, the message banner lived inside the settings storyboard as a child view controller of several VCs. It could not be used outside the settings features. Now, it can be used by any view controller. This also gets rid of the additional `containerView` and instead embeds the view controller's `view` directly, so we don't need a delegate anymore.

# 🛠 How

Added `MessageBannerViewController.xib`. The default implementation of `configureMessageBannerViewController` loads the nib, then installs it as a child view controller of the parent view controller. The parent can the call `self.messageBannerViewController.show` as needed.

# 👀 See

4prwowrufy

# ✅ Acceptance criteria

- [x] go to `Change Email` and change your email. You should see the banner appear and you should be able to tap it and swipe to dismiss.
- [x] go to `Change Password` and try to change your password with the wrong `current password`. This should show an error message banner (invalid password). You should be able to tap it and swipe to dismiss.
- [x] "invisible container" issue where the view overlaps the current view's content, preventing you from tapping that content


justinswart:
I had another look at that `layoutMargins` issue, it seems related to that [doc](https://developer.apple.com/documentation/uikit/uiview/1622566-layoutmargins) we were looking at and the fact that we can't manipulate the `layoutMargins` of the VC's root view 🤔.

It really seems that setting `viewRespectsSystemMinimumLayoutMargins` to `false` is what we're looking for (https://stackoverflow.com/questions/27421469/setting-layoutmargins-of-uiview-doesnt-work) but I wasn't successful getting that to work.

I'll look a bit more tomorrow, otherwise we could just pin this to the view's edges and set our padding manually 👍 🤷‍♂️

ifbarrera:
@justinswart I ended up doing what that stack overflow article recommended and using a container view to set the `layoutMargins`.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -2,28 +2,22 @@ import Foundation
 import Library
 import Prelude
 
-internal protocol MessageBannerViewControllerDelegate: class {
-  var messageBannerViewControllerContainer: UIView { get }
-  func messageBannerViewControllerIsHidden(_ isHidden: Bool)
+protocol MessageBannerViewControllerPresenting {
```
justinswart: What if we added `messageBannerViewController` as a `var` in the protocol too to be sure it is always a property on view controllers that are presenting it?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -22,15 +22,11 @@ final class SettingsAccountViewController: UIViewController {
...
-    guard let messageBannerView = self.children.first as? MessageBannerViewController else {
-      fatalError("Couldn't instantiate MessageBannerViewController")
-    }
-
-    self.messageBannerView = messageBannerView
-
     self.tableView.dataSource = dataSource
     self.tableView.delegate = self
 
+    self.messageBannerViewController = configureMessageBannerViewController(on: self)
```
justinswart: Would we ever not pass `self` to this function? 🤔

If we added the `var` to the protocol then this could just be a call like:

```swift
self.configureMessageBannerViewController()
```

And then within that default implementation we could set the property.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -91,34 +92,32 @@ final class MessageBannerViewController: UIViewController {
...
+
     UIView.animate(withDuration: duration, delay: 0.0,
                    options: UIView.AnimationOptions.curveEaseInOut,
                    animations: { [weak self] in
-                    guard let `self` = self else { return }
-                    let frameHeight = self.view.frame.size.height
-                    self.backgroundViewBottomConstraint.constant = isHidden
-                      ? frameHeight : self.bottomMarginConstraintConstant
-                    self.view.layoutIfNeeded()
+                    guard let strongSelf = self else { return }
```
dusi: Can also use `guard let `\``self`\`` = self else { return }`

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
ifbarrera: It felt a little side-effect-y when I initially did it this way; I thought it might be better to allow the caller to do what it wants with the configured `messageBannerViewController`. Setting it in the `configure` function would also require we make the function `mutating`. It seemed slightly more functional to have it take the parent view controller as an argument, and the subsequently configured `messageBannerViewController` as the return value.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
ifbarrera: Ah yes - thanks!

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
justinswart: Cool, sounds good!

Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
```diff
@@ -32,11 +31,7 @@ internal final class ChangeEmailViewController: UIViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    guard let messageBannerViewController = self.children.first as? MessageBannerViewController else {
-      fatalError("Couldn't instantiate MessageBannerViewController")
-    }
-    self.messageBannerViewController = messageBannerViewController
-    self.messageBannerViewController.delegate = self
+    self.messageBannerViewController = configureMessageBannerViewController(on: self)
```
justinswart: `configureMessageBannerViewController` should probably be `self.configureMessageBannerViewController` in the places where it's used 😬

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -3,11 +3,12 @@ import Prelude
...
 import Result
 
+public typealias MessageBannerConfiguration = (type: MessageBannerType, message: String)
+
 public protocol MessageBannerViewModelInputs {
-  func setBannerType(type: MessageBannerType)
-  func setBannerMessage(message: String)
-  func showBannerView(shouldShow: Bool)
   func bannerViewAnimationFinished(isHidden: Bool)
+  func showBannerView(shouldShow: Bool)
```
justinswart: Can I suggest another name for this input to be `bannerViewVisibilityWillChange(toShow show: Bool)` just to keep with conventions of outputs generally being _what_ the effect of the input is.

dusi:
[Remove MessageBannerViewController from the Settings storyboard](https://trello.com/c/J7mBSotV/1089-remove-messagebannerviewcontroller-from-the-settings-storyboard)

PR #508, opened 2018-12-05
[BUGFIX] Logout does not dismiss the second time

ifbarrera:
# 📲 What

Bugfix for a bug where the settings screen does not dismiss when you attempt to log out a second time.

# 🤔 Why

Because 🐛 s are 👎 

# 🛠 How

Previously, although we were transitioning views to the discovery page, and logging out with the server, we were not dismissing the `SettingsViewController` that had previously been presented (see screenshot below). This was causing an issue where the second attempt to log out would not appear as though the view actually transitioned.

screen shot 2018-12-05 at 11 51 11 am

Now we're actually dismissing the `SettingsViewController` when the user logs out.

# 👀 See

isjw2nbkxm

# ✅ Acceptance criteria

- [x] Log in once, then log out. Then log in again, and then log out again. You should be transitioned to the discovery screen every time you log out.


Kickstarter-iOS/Views/Controllers/BackerDashboardViewController.swift
```diff
@@ -370,6 +370,8 @@ extension BackerDashboardViewController: UIGestureRecognizerDelegate {
 
 extension BackerDashboardViewController: TabBarControllerScrollable {
   func scrollToTop() {
+    //swiftlint:disable todo
+    //FIXME: crash because self.pageViewController is nil
```
dusi: I've added this to `[RELEASE] Settings v2` card so we keep track of it.

dusi: I also believe I know how to reproduce this bug..it happens when you're toggling between an account that is a backer and that isn't.

Steps to reproduce:
1. Login as backer
2. Logout
3. Login as non-backer
>> should crash because `RootTabBarController` is trying to call `scrollToTop()` on `BackerDashboardViewController`

PR #510, opened 2018-12-05
Hide payment methods from release builds

ifbarrera:
# 📲 What

Only shows the "Payment Methods" cell on debug builds. This PR also re-adds the `Account` cell to all builds, and removes the `Privacy` cell from the main settings screen (it's now in the Account screen).

# 🤔 Why

Because we're not ready to release settings v3, which includes payment methods.

# 🛠 How

Adds a preprocessor check `#if DEBUG` to add or remove the `paymentMethods` cell type.

# 👀 See

Debug build:

simulator screen shot - iphone 8 - 2018-12-05 at 16 59 07

Release build:

simulator screen shot - iphone 8 - 2018-12-05 at 16 55 29

# ✅ Acceptance criteria

- [x] edit the schema's build configuration to `Release`, run the app, go to Settings -> Account, check that the payment methods cell is *not* there
- [x] edit the schema's build configuration to `Debug`, run the app, go to Settings -> Account, check that the payment methods cell is there
- [x] Privacy cell should now be in the `Account` screen, tapping it should go to the Privacy screen
- [x] issues to watch out for: crashes when tapping the currency cell to hide/show the currency picker, on both `DEBUG` and `RELEASE` builds


Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -28,10 +28,9 @@ final class SettingsAccountDataSource: ValueCellDataSource {
   func insertCurrencyCell(currency: Currency?) -> IndexPath {
     let cellValue = SettingsCurrencyCellValue(cellType: SettingsAccountCellType.currency, currency: currency )
 
-    return self.insertRow(value: cellValue,
+    return self.appendRow(value: cellValue,
```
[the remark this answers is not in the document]
justinswart: This should be fine given that this cell is always expected to be last unless the picker is shown and given that that is also being appended it will be after this one. The only thing that could cause an incorrect index situation would be the removal of the picker which depends on the index being 2. In this case it'll be fine and seeing as we're refactoring this soon I don't think it's a problem but I think generally I'd suggest searching for it in the array and removing it at the index that we find it at to prevent regression bugs.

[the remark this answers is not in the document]
ifbarrera: Good catch, this was actually a bug.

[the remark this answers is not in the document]
justinswart: Oh good then, it seemed to me like it would have worked too 😄 🤷‍♂️

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -122,8 +122,12 @@ final class SettingsAccountViewController: UIViewController, MessageBannerViewCo
   }
 
   private func dismissCurrencyPickerCell() {
+    guard let pickerCell = self.dataSource.removeCurrencyPickerRow() else {
```
justinswart: Can I suggest that we call this `pickerCellIndexPath` just so it's clearer what's being returned.
```suggestion
    guard let pickerCellIndexPath = self.dataSource.removeCurrencyPickerRow() else {
```

Library/DataSource/ValueCellDataSource.swift
```diff
@@ -258,6 +258,13 @@ open class ValueCellDataSource: NSObject, UICollectionViewDataSource, UITableVie
     return self.values.reduce(0) { accum, section in accum + section.count }
   }
 
+  /**
+   - returns: The total number of items in given section, in the data source.
+   */
+  public final func numberOfItems(in section: Int) -> Int {
+    return self.values[section].reduce(0) { accum, _ in accum + 1 }
```
justinswart: Not just?
```suggestion
    return self.values[section].count
```

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -122,8 +122,12 @@ final class SettingsAccountViewController: UIViewController, MessageBannerViewCo
...
+      return
+    }
+
     tableView.beginUpdates()
-    self.tableView.deleteRows(at: [self.dataSource.removeCurrencyPickerRow()], with: .top)
+    self.tableView.deleteRows(at: [pickerCell], with: .top)
```
justinswart: If you make use of the other suggestion 😄 
```suggestion
    self.tableView.deleteRows(at: [pickerCellIndexPath], with: .top)
```

Library/DataSource/ValueCellDataSource.swift
ifbarrera: 🤦‍♀️  yep that works too haha

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -122,8 +122,12 @@ final class SettingsAccountViewController: UIViewController, MessageBannerViewCo
   }
 
   private func dismissCurrencyPickerCell() {
+    guard let pickerCellIndexPath = self.dataSource.removeCurrencyPickerRow() else {
+      return
+    }
+
     tableView.beginUpdates()
-    self.tableView.deleteRows(at: [self.dataSource.removeCurrencyPickerRow()], with: .top)
+    self.tableView.deleteRows(at: [pickerCell], with: .top)
```
ifbarrera:
```suggestion
    self.tableView.deleteRows(at: [pickerCellIndexPath], with: .top)
```

PR #511, opened 2018-12-05
Prevent currency picker interaction during requests

dusi:
# 📲 What

Fixes the crash we were seeing related to the currency picker. This crash would happen if a user tried to show currency picker while the currency change was in progress causing the application to crash.

# 🤔 Why

We really, really, really should ship Settings.

# 🛠 How

This workaround waits till the currency change requests is done before allowing presenting the currency picker again. This solution would ideally show an activity indicator but since we're planning on completely changing this we should be good for now.

# 👀 See

| Before | After |
| --- | --- |
| before | after |

# ✅ Acceptance criteria

For easy testing, please download and install Network Link Conditioner (NLC) - https://developer.apple.com/downloads/ (Additional_Tools_for_Xcode_{YOUR_VERSION})

Enable NLC and use any option that slows down your Internet connection in order for you to being able to select the currency row while the progress is happening.

Go to `Account Settings`..

- [x] Tapping the currency picker opens the currency picker with proper currency selected (reflecting the last successfully updated currency)
- [x] Changing currency using the currency picker should work
- [x] Tapping the currency row will present the picker if no request (currency change) is happening
- [x] Tapping the currency row will NOt present the picker if a request (currency change) is happening

Don't forget to disable NLC in order to regain your Internet speedzzZZZzz

### DIFF
--- a/Kickstarter-iOS/DataSources/HelpDataSource.swift
+++ b/Kickstarter-iOS/DataSources/HelpDataSource.swift
@@ -4,7 +4,7 @@ import KsApi
 final class HelpDataSource: ValueCellDataSource {
   func configureRows() {
     _ = HelpSectionType.allCases.map { section -> Void in
-      let values = section.cellRowsForSection.map { SettingsCellValue(user: nil, cellType: $0) }
+      let values = section.cellRowsForSection.map { SettingsCellValue(cellType: $0) }
       self.set(values: values,
                cellClass: SettingsTableViewCell.self,
                inSection: section.rawValue)
--- a/Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
+++ b/Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
@@ -6,7 +6,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
     clearValues()
     SettingsAccountSectionType.allCases
       .forEach { section -> Void in
-      let values = section.cellRowsForSection.map { SettingsCellValue(user: nil, cellType: $0) }
+      let values = section.cellRowsForSection.map { SettingsCellValue(cellType: $0) }
 
       self.set(values: values,
                cellClass: SettingsTableViewCell.self,
@@ -34,8 +34,8 @@ final class SettingsAccountDataSource: ValueCellDataSource {
                           inSection: SettingsAccountSectionType.payment.rawValue)
   }
 
-  func insertCurrencyPickerRow() -> IndexPath {
-    let cellValue = SettingsCellValue(user: nil, cellType: SettingsAccountCellType.currencyPicker)
+  func insertCurrencyPickerRow(currency: Currency) -> IndexPath {
+    let cellValue = SettingsCellValue(cellType: SettingsAccountCellType.currencyPicker, currency: currency)
 
     return self.appendRow(value: cellValue,
                           cellClass: SettingsCurrencyPickerCell.self,
@@ -43,7 +43,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
   }
 
   func removeCurrencyPickerRow() -> IndexPath {
-    let cellValue = SettingsCellValue(user: nil, cellType: SettingsAccountCellType.currencyPicker)
+    let cellValue = SettingsCellValue(cellType: SettingsAccountCellType.currencyPicker)
 
     return self.deleteRow(value: cellValue,
                           cellClass: SettingsCurrencyPickerCell.self,
--- a/Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
+++ b/Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
@@ -15,7 +15,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
 
     self.insertChangeEmailCell(shouldHideEmailWarning)
 
-    _ = self.insertCurrencyCell(currency: currency)
+    _ = self.insertCurrencyCell(with: currency)
   }
 
   func insertChangeEmailCell(_ shouldHideEmailWarning: Bool) {
@@ -25,7 +25,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
                    inSection: SettingsAccountSectionType.emailPassword.rawValue)
   }
 
-  func insertCurrencyCell(currency: Currency?) -> IndexPath {
+  func insertCurrencyCell(with currency: Currency?) -> IndexPath {
     let cellValue = SettingsCurrencyCellValue(cellType: SettingsAccountCellType.currency, currency: currency )
 
     return self.insertRow(value: cellValue,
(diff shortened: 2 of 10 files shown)


dusi:
[Changing currency preference using picker](https://trello.com/c/Vn9ntXpQ/984-changing-currency-preference-using-picker)

Kickstarter-iOS/DataSources/HelpDataSource.swift
```diff
@@ -4,7 +4,7 @@ import KsApi
 final class HelpDataSource: ValueCellDataSource {
   func configureRows() {
     _ = HelpSectionType.allCases.map { section -> Void in
-      let values = section.cellRowsForSection.map { SettingsCellValue(user: nil, cellType: $0) }
```
dusi: Made this optional so that we don't have to explicitly pass `nil`s all over the place

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -34,16 +34,16 @@ final class SettingsAccountDataSource: ValueCellDataSource {
                           inSection: SettingsAccountSectionType.payment.rawValue)
   }
 
-  func insertCurrencyPickerRow() -> IndexPath {
-    let cellValue = SettingsCellValue(user: nil, cellType: SettingsAccountCellType.currencyPicker)
+  func insertCurrencyPickerRow(currency: Currency) -> IndexPath {
```
dusi: I was thinking about most of the method signatures that take `currency` to take unnamed parameter but last minute decided to be more explicit. Happy to change this to `methodName(_ currnecy: Currency)` 😎

Kickstarter-iOS/Views/Cells/SettingsCurrencyPickerCell.swift
```diff
@@ -54,7 +59,7 @@ extension SettingsCurrencyPickerCell: UIPickerViewDataSource {
 
 extension SettingsCurrencyPickerCell: UIPickerViewDelegate {
   func pickerView(_ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
-    return Currency.allCases[row].descriptionText //Currency(rawValue: row)?.descriptionText
+    return Currency.allCases[row].descriptionText
```
dusi: Good bye useless comment

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -111,11 +113,14 @@ final class SettingsAccountViewController: UIViewController {
       |> settingsTableViewStyle
   }
 
-  private func showCurrencyPickerCell() {
+  private func showCurrencyPickerCell(currency: Currency) {
+    let tapRecognizer = UITapGestureRecognizer(
```
dusi: Minor reorganization so that the animation block is more clear

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -125,10 +130,10 @@ final class SettingsAccountViewController: UIViewController {
   }
 
   private func dismissCurrencyPickerCell() {
-    tableView.beginUpdates()
+    self.tableView.beginUpdates()
```
dusi: Added missing `self`s

Library/SettingsCellType.swift
```diff
@@ -6,18 +6,20 @@ public struct SettingsCurrencyCellValue {
...
-    self.currency = currency
     self.cellType = cellType
+    self.currency = currency
   }
 }
 
 public struct SettingsCellValue {
-  public let user: User?
   public let cellType: SettingsCellTypeProtocol
+  public let currency: Currency?
```
dusi: Sorted alphabetically

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -76,7 +79,17 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
...
-    self.presentCurrencyPicker = currencyCellSelected.signal.mapConst(true).ignoreValues()
+    let updateCurrencyInProgress = Signal.merge(
+      self.viewDidLoadProperty.signal.mapConst(false),
+      didConfirmChangeCurrency.signal.mapConst(true),
+      updateCurrencyEvent.filter { $0.isTerminating }.mapConst(false)
+    )
+
+    self.presentCurrencyPicker = Signal.combineLatest(currency, updateCurrencyInProgress)
+      .takePairWhen(currencyCellSelected.signal)
+      .map { ($0.0, $0.1, $1) }
```
dusi: I'd be happy to make this more readable if we find a better way

Kickstarter-iOS/DataSources/HelpDataSource.swift
```diff
@@ -4,7 +4,7 @@ import KsApi
...
+      let values = section.cellRowsForSection.map { SettingsCellValue(cellType: $0) }
```
justinswart: If you want to be point-free:
```suggestion
      let values = section.cellRowsForSection.map(SettingsCellValue(cellType:))
```

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -6,7 +6,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
     clearValues()
     SettingsAccountSectionType.allCases
       .forEach { section -> Void in
-      let values = section.cellRowsForSection.map { SettingsCellValue(user: nil, cellType: $0) }
+      let values = section.cellRowsForSection.map { SettingsCellValue(cellType: $0) }
```
justinswart: If you're in that point-free mood 🤤

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
justinswart: I think we'd probably want `insertCurrencyPickerRow(with currency: Currency)` but fine for now.

Library/ViewModels/SettingsAccountViewModel.swift
justinswart: Looks like this does compile, Xcode was just struggling before:

```swift
self.presentCurrencyPicker = Signal.combineLatest(currency, updateCurrencyInProgress)
  .takePairWhen(currencyCellSelected.signal)
  .map(unpack)
  .filter(second >>> isFalse)
  .map(first)
```

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
dusi: If there was suggestion here as well I could have batched these on Github...this way I went to Xcode and did these 2 changes in IDE.

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
dusi: Great minds think alike..and then go and change it to something lame...hahaha, used `with` in my first iteration :)

Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
dusi: Ugh, also I believe you can't point free something that has multiple optional params :/

PR #512, opened 2018-12-06
Change password hidden

Scollaco:
# 📲 What
Hide `Change Email / Change Password` section if user doesn't have password.

# 🤔 Why
Because we require a password to change email or password, users who signed up using Facebook won't be able to use this feature, since they, initially won't have password.

# 🛠 How
Before presenting the `SettingsAccountViewController`, we fetch some user's fields, one of them beings `hasPassword`. If the value of `hasPassword` is `false` or `nil`, we hide the section on the Data Source. 

# 👀 See

| No Password | With Password |
| --- | --- |
|||

# ✅ Acceptance criteria

- [ ] I Signup using Facebook, go to `Settings > Account`. The Change Email / Password section  `should not` be visible.
- [ ] I Signup using email and create a password.  The Change Email / Password section  `should` be visible.
- [ ] I go to `Settings > Account` with no wi-fi. The Change Email / Password section `should not` be visible.
- [ ] I login using email and create a password.  The Change Email / Password section `should` visible.


Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -98,6 +102,12 @@ final class SettingsAccountViewController: UIViewController {
       .observeValues { [weak self] _ in
         self?.showChangeCurrencyAlert()
     }
+
+    self.viewModel.outputs.tableViewTopConstraint
```
justinswart: Why was this needed with this change?

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -43,13 +44,26 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
...
 
         return isEmailVerified && isDeliverable
     }
 
+    let initialHiddenState = self.viewWillAppearProperty.signal.mapConst(true)
+
+    let shouldHideEmailPasswordSection = Signal.merge(
+      initialHiddenState,
+      userAccountFields.values()
+        .map { response -> Bool in
```
justinswart: This can be:

```swift
.map { $0.me.hasPassword == .some(false) }
```

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -51,16 +52,19 @@ final class SettingsAccountViewController: UIViewController, MessageBannerViewCo
...
         self?.tableView.reloadData()
     }
 
     self.viewModel.outputs.fetchAccountFieldsError
       .observeForUI()
       .observeValues { [weak self] in
-        self?.dataSource.configureRows(currency: nil, shouldHideEmailWarning: true)
+        self?.dataSource.configureRows(currency: nil,
+                                       shouldHideEmailWarning: true,
+                                       shouldHideEmailPasswordSection: true)
```
ifbarrera: I would say that `shouldHideEmailPasswordSection` should default to `false` under error scenarios. We still want most users to be able to get to Change email and Change password if the settings account fields fail to load; facebook users would get an error at the change email/change password screen anyway.

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
Scollaco: I don't know if that's the best approach, but this will hide the `emailPassword` section header view. Without updating the constraint, we could see an extra header above the privacy section and it looked weird.

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -44,13 +45,26 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
...
-              let isDeliverable = response.me.isDeliverable else {
-          return true
+          let isDeliverable = response.me.isDeliverable else {
+            return true
         }
 
         return isEmailVerified && isDeliverable
     }
 
+    let initialHiddenState = self.viewWillAppearProperty.signal.mapConst(true)
```
ifbarrera: Here I would also say we should default to assuming that users will be able to view that section, and hide it if not. @dannyalright thoughts on this?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
justinswart: Oh I see. I would suggest instead returning a `nil` header view for that section.

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -140,12 +156,17 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
...
   public let showAlert: Signal<(), NoError>
+  public let tableViewTopConstraint: Signal<CGFloat, NoError>
   public let transitionToViewController: Signal<UIViewController, NoError>
   public let updateCurrencyFailure: Signal<String, NoError>
 
   public var inputs: SettingsAccountViewModelInputs { return self }
   public var outputs: SettingsAccountViewModelOutputs { return self }
 }
+
+private func updateConstraintConstant(_ shouldHideEmailPasswordSection: Bool) -> CGFloat {
```
ifbarrera: If you update the number of sections in the data source, you shouldn't need to modify the header height 🤔.

Kickstarter-iOS/DataSources/SettingsAccountDataSourceTests.swift
```diff
@@ -8,42 +9,74 @@ final class SettingsAccountDataSourceTests: XCTestCase {
   private let dataSource = SettingsAccountDataSource()
   private let tableView = UITableView()
 
-  func testConfigureRows() {
-    self.dataSource.configureRows(currency: Currency.USD, shouldHideEmailWarning: true)
+  func testConfigureRows_EmailPasswordRows_Shown() {
+    self.self.dataSource.configureRows(currency: Currency.USD,
+                                  shouldHideEmailWarning: true,
```
ifbarrera: Indentation is a bit off here

Kickstarter-iOS/DataSources/SettingsAccountDataSourceTests.swift
```diff
@@ -8,42 +9,74 @@ final class SettingsAccountDataSourceTests: XCTestCase {
...
+                                  shouldHideEmailPasswordSection: false)
+
+    let currency = Currency.USD
 
     _ = self.dataSource.insertCurrencyPickerRow(with: currency)
 
-    // swiftlint:disable line_length
-    XCTAssertEqual(3, dataSource.tableView(tableView,
+    XCTAssertEqual(3, self.dataSource.tableView(self.tableView,
```
ifbarrera: Same here

Scollaco:
[Hide Change Password and Change email for users with no `hasPassword`](https://trello.com/c/fOUDrJj7/1094-hide-change-password-and-change-email-for-users-with-no-haspassword)

PR #515, opened 2018-12-07
Remove shrug emoji

cdolm92:
# 📲 What

Removed "🤷🏾‍♀️" on currency cell where currency label should be present.
# 🤔 Why

This emoji was appearing whenever the chosen currency text was nil. Ex. no  wifi/ cellular data.

# 🛠 How

If the chosen currency text is nil then a blank string should emit.

# 👀 See

# ✅ Acceptance criteria

- [ ] Use your phone and turn off wifi/cellular data
- [ ] Navigate to  the Account screen and make sure "🤷🏾‍♀️" is no longer there.


cdolm92:
@kickstarter/native-squad This is blocked until we figure out how to best communicate w/ users that currency details fail to load in cases where a user has no service.

PR #517, opened 2018-12-07
Banner animation fixes

ifbarrera:
# 📲 What

Fixes two bugs uncovered during QA:

- banner does not account for safe area insets (iPhone X home indicator)
- stutter during panning
- fix the xib warning

# 🤔 Why

🐛 s are 🚫 

# 🛠 How

Pinning the banner to the layout margins of the superview was causing the height of the view to change during dragging. This was messing up the animation, which depends on a constant height. To fix this, we manually set the view's padding, and make sure the bottom constraint of the view is constrained against `layoutMarginsGuide.bottom`, which accounts for the home indicator.

# 👀 See

mupyljf96d

image

# ✅ Acceptance criteria

- [x] go to change password, enter *the wrong password*, you should see the error banner come up. You should be able to tap to dismiss, swipe to dismiss (no stutter), or it should just dismiss automatically
- [x] banner should not cover the home indicator on iPhone X


dannyalright:
Tested on a iPhone XS via the Currency error message banner. Works well!

PR #521, opened 2018-12-10
Add the ability to mock availability API

dusi:
# 📲 What

Adds the ability to mock availability API

# 🤔 Why

We should be able to mock this in tests.

# 🛠 How

screen shot 2018-12-09 at 11 16 12 pm

# ✅ Acceptance criteria

- [x] 1Password button shows on iOS 11 if `1Password.app` is available
- [x] 1Password button does NOT show on iOS 11 if `1Password.app` is NOT available
- [x] 1Password button does NOT shows on iOS 12 no matter if `1Password.app` is available or NOT


Library/SharedFunctions.swift
```diff
@@ -157,9 +157,20 @@ public func countdownProducer(to date: Date)
...
+  if AppEnvironment.current.isOSVersionAvailable(12.0) {
     return true
   } else {
     return isHidden
   }
 }
+
+public func ksr_isOSVersionAvailable(_ version: Double) -> Bool {
+  switch version {
+  case 12.0...:
```
justinswart: If we needed to test say `12.1` vs `12.1.1` then we'd just make this `switch` more specific, right?

dusi: Yes, that is correct....as long as the order is correct and the higher version is declared before the open range operator..

screen shot 2018-12-10 at 3 18 27 pm

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -94,21 +94,33 @@ final class ChangeEmailViewModelTests: TestCase {
...
   }
 
-  func testOnePasswordButtonHidesIfNotAvailable() {
+  func testOnePasswordButtonHidesProperly_OnIOS11AndEarlier() {
     self.vm.inputs.viewDidLoad()
-    self.vm.inputs.onePassword(isAvailable: false)
 
-    self.onePasswordButtonIsHidden.assertValues([true])
+    let iOS12: (Double) -> Bool = { _ in false }
+    withEnvironment(isOSVersionAvailable: iOS12) {
```
ifbarrera: Is the one password extension dependent on the availability API? It seems that the test here is behaving independently of the OS version.

ifbarrera: Ah now I see, on iOS 11 and under it just returns the value of the `onePasswordIsAvailableProperty` signal.

PR #522, opened 2018-12-10
Fix disappearing email button, fix banner not showing after multiple taps

ifbarrera:
# 📲 What

This PR fixes the following bugs:

- change password in landscape mode is too wide
- error message banner stops appearing after multiple "Save" actions
- send verification email button disappears when moving into landscape mode

# 🤔 Why

Cleanup all the bugs around Settings V2.

# 🛠 How

- Cleaned up all the constraints in Change Email and Change Password, which should have been made relative to the safe area, not the superview.
- the `sendEmailVerification` button was being manually hidden in `bindStyles`, which gets re-called when you move to landscape mode. In order to achieve the same thing, we can merge the `viewDidLoadProperty` signal to always hide the `resendEmailVerificationView` when the view loads
- the banner view had a bunch of logic in the view controller that was based on the `isAnimating` bool. I realized that we don't actually need that logic because the view model already filters repeating signals (via `skipRepeats()`). Removing that unnecessary logic fixes the banner bug.

# 👀 See

vtc5rccjlm
teiswjone5

# ✅ Acceptance criteria

- [x] go to the change password screen, turn your phone to landscape mode and see that it is the correct size
- [x] go to the change email screen, turn your phone to landscape mode: the `Send email verification` button should not disappear
- [x] go to the change password screen, enter the wrong `current password` and then any password in the `new password` fields. Hit Save, then continue to hit save (~ 10 times). The banner should always re-appear after disappearing.


cdolm92:
Awesome 🎉! The 1st and 3rd AC works as expected. Will test the 2nd AC when I have access to an account I can test on.

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -225,16 +225,16 @@
...
+                                                    <constraint firstItem="IjC-cR-RB2" firstAttribute="bottom" secondItem="Cdr-bL-P2W" secondAttribute="bottom" constant="-18" id="7IG-eC-1Nw"/>
                                                     <constraint firstItem="IjC-cR-RB2" firstAttribute="leading" secondItem="Cdr-bL-P2W" secondAttribute="leading" constant="12" id="RcF-Rf-Ixj"/>
                                                     <constraint firstItem="IjC-cR-RB2" firstAttribute="top" secondItem="Cdr-bL-P2W" secondAttribute="top" constant="6" id="lg9-uS-DXy"/>
                                                 </constraints>
                                             </view>
-                                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="ioz-I8-SoZ" userLabel="Resend Verification Email Button View">
+                                            <view hidden="YES" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="ioz-I8-SoZ" userLabel="Resend Verification Email Button View">
```
dusi: We should probably see these explicitly in code ... as discussed in https://github.com/kickstarter/ios-oss/pull/488#discussion_r235212773

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -205,17 +205,17 @@
...
+                                                    <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" translatesAutoresizingMaskIntoConstraints="NO" id="IjC-cR-RB2">
                                                         <rect key="frame" x="12" y="6" width="351" height="0.0"/>
                                                         <subviews>
-                                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="255" verticalHuggingPriority="255" text="Label" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="lz0-3G-6S5">
-                                                                <rect key="frame" x="0.0" y="0.0" width="42" height="0.0"/>
+                                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" text="Label" lineBreakMode="wordWrap" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="lz0-3G-6S5">
```
dusi: I think it would also be better if we cans set these in code? It's just overall simpler to make changes to that later on and easier to find (wrap in a styles)....I think this is something we should all talk about on our team to see if this is the pattern we want to follow or not

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
ifbarrera: This actually is being explicitly set [via the view model](https://github.com/kickstarter/ios-oss/blob/2278546279f5360a350f095adc75a20b39f0419e/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift#L129) - I can remove this flag in IB.

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
ifbarrera: This is also not needed and is being set [via a styles variable](https://github.com/kickstarter/ios-oss/blob/2278546279f5360a350f095adc75a20b39f0419e/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift#L85). Will also remove!

ifbarrera:
@cdolm92 one way to test the second AC is the following:

1. Login on staging
2. Go to change email
3. Change your email to something you can remember (I usually use another one of my personal emails)
4. *You should then see the "Send verification email button"*
5. After you're done testing, just change your email back to your original email (this will also happen automatically when the staging db is re-seeded)

PR #525, opened 2018-12-12
Fixed crash when trying to change environment/language

Scollaco:
# 📲 What
Fixed a crash that was happening on `BetaToolsViewController` when trying to change environment or language.

# 🤔 Why
The app was crashing because we were not setting the modalPresentation style to `.popover`, what's  required for iPad if we are trying to present an ActionSheet.

# 🛠 How
By setting if the device is iPad and setting the presentation style to popover.

# 👀 See

# ✅ Acceptance criteria

- [ ] Using an iPad, go to `Beta Tools` and tap the environment button. A popover should appear with the environment options.
- [ ] Using an iPad, go to `Beta Tools` and tap the language button. A popover should appear with the language options.

GIF

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
@@ -131,6 +131,11 @@ internal final class BetaToolsViewController: UIViewController {
       UIAlertAction.init(title: "Cancel", style: .cancel)
     )
 
+    if UIDevice.current.userInterfaceIdiom == .pad {
+      alert.modalPresentationStyle = .popover
+      alert.popoverPresentationController?.sourceView = self.languageSwitcher
+    }
+
     self.present(alert, animated: true, completion: nil)
   }
 
@@ -148,6 +153,11 @@ internal final class BetaToolsViewController: UIViewController {
     alert.addAction(
       UIAlertAction.init(title: "Cancel", style: .cancel)
     )
+
+    if UIDevice.current.userInterfaceIdiom == .pad {
+      alert.modalPresentationStyle = .popover
+      alert.popoverPresentationController?.sourceView = self.environmentSwitcher
+    }
 
     self.present(alert, animated: true, completion: nil)
   }


Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -131,6 +131,11 @@ internal final class BetaToolsViewController: UIViewController {
       UIAlertAction.init(title: "Cancel", style: .cancel)
     )
 
+    if UIDevice.current.userInterfaceIdiom == .pad {
```
ifbarrera: Just a thought, we could also use `AppEnvironment.current.device`, which would allow us to mock devices and test this logic. Looking at it more carefully, there's probably a case for adding this logic for the view model (right now `showEnvironmentActionSheet` is being called directly from the tap gesture handler) and writing a test for it.

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -149,6 +154,11 @@ internal final class BetaToolsViewController: UIViewController {
...
```
dusi: Would it make sense to extract these checks to a `UIAlertController`’s extension in order to prevent this from happening in the real code base and be more DRY? We can either have a convenience function to present alert or just to create new instance

Scollaco: That's a good idea. But instead of creating a `UIAlertController`'s extension, what about create a `UIViewController`'s extension to present the alert? We can add a function containing that logic like:
```
public func present(_ alert: UIAlertController, sourceView: UIView? = nil) 
```
And, from the ViewController, simply call 
```
self.present(alert, sourceView: self.languageSwitcher)
```

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
Scollaco: After digging a little more, I noticed that the `self.dismiss(animated:)` function is not actually necessary, because the AppDelegate will set the destination ViewController once the `.ksr_languageChanged` notification is fired. And the completion block wasn't being called (I just wonder since when this is happening). 

Also, the `UIDevice.current.userInterfaceIdiom == .pad` is the convention we use on several places of the app just to set the UIAlertController's modal presentation type. In this particular case, I think @dusi 's solution to move the logic to an extension might be quicker and more efficient. What do you think?

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
Scollaco: Nm. I just extracted the logic to the `UIAlertController` extension.

Library/UIAlertController.swift
```diff
@@ -6,6 +6,23 @@ import KsApi
 
 public extension UIAlertController {
 
+  public static func `init`(title: String,
```
ifbarrera: Should we name this consistently with the static function below? Something like:

`public static func alertController(with title: String, message: String?...)`

ifbarrera:
Tested on iPad Air 2 (12.1):
✅beta tools opens environment popover without crashing
✅beta tools opens language popover without crashing
✅beta tools switches environment correctly
✅beta tools switches language correctly

Tested on iPhone 8 (12.1):
✅beta tools opens environment action sheet without crashing
✅beta tools opens language action sheet without crashing
✅beta tools switches environment correctly
✅beta tools switches language correctly

PR #528, opened 2018-12-14
Filters out Chinese translations

Scollaco:
# 📲 What
This [PR](https://github.com/kickstarter/kickstarter/pull/13592) adds Chinese to our locale options, but, because we still don't have a folder for chinese localization, the `make strings` command fails.

# 🤔 Why
`make strings` should work in order to give us the latest translations.

# 🛠 How
For now, chinese is being filtered out on our script.

# ✅ Acceptance criteria
- [x] On master, I run `make strings` and the script fails.
- [x] I check out `strings_script_fix` branch, run `make strings`. The script should work.


bin/strings.swift
```diff
@@ -121,7 +121,7 @@ let stringsByLocale = stringsByLocale1
 }
 
 stringsByLocale?.forEach { locale, strings in
-  guard locale != "it" else { return }
```
ifbarrera: Should we instead do a check that `locale` is contained within a list of our supported locales? That way if kickstarter adds another locale that we don't have a folder for, we don't have to update this again?

Something like:
```
let supportedLocales = ["en", "es", ...]
stringsByLocale?.forEach { locale, strings in
    guard supportedLocales.contains(locale) else { return }
    ...
}
```

PR #533, opened 2018-12-20
Pledge subtitle label textColor should match pledge label textColor

ifbarrera:
# 📲 What

The `textColor` of the pledge subtitle label should always match the `textColor` of the pledge label.

# 🤔 Why

Better UI.

# 🛠 How

Remove the hard-coded `textColor` on the `pledgeSubtitleLabel` which was causing the color to appear grey -> green. Now it just appears immediately as the color defined by the view model.

# 👀 See

*Before*
hrpe4rxlqh

*After*
kgwraiqgcn

# ✅ Acceptance criteria

- [ ] subtitle label text color should always immediately appear the same color as the pledge label text color


ifbarrera:
[Standardize pledge color on Project Page](https://trello.com/c/YUYSZPi6/1101-standardize-pledge-color-on-project-page)

PR #534, opened 2018-12-20
Accessibility - Settings - Rows

dusi:
# 📲 What

Flags common settings cells as a button for Voice Over.

# 🤔 Why

Better indicates that there's an action linked to these cells for visually impaired users.

# 🛠 How

Simply setting `accessibilityTraits = .button`.

# ✅ Acceptance criteria

on all the following: iOS 10, iOS 11 & iOS 12
with Voice Over turned ON

- [x] All cells in `Settings` (except `App version`) now have button trait (i.e. Account - button, Log out - button, etc.)
- [x] All cells in `Settings > Account` now have button trait
- [x] All cells in `Settings > Help` now have button trait
- [x] All cells in `Settings > Help` now show disclosure indicator (except Contact which is presented modally) - based on discussion with @dannyalright


Kickstarter-iOS/Views/Cells/SettingsAccountWarningCell.swift
```diff
@@ -9,8 +9,11 @@ final class SettingsAccountWarningCell: UITableViewCell, ValueCell, NibLoading {
   @IBOutlet fileprivate weak var warningIconImage: UIImageView!
 
   func configureWith(value shouldHideAlertIcon: Bool) {
+    _ = self
+      |> \.accessibilityTraits .~ .button
```
dusi: Since we only use this cell in one place this should be fine. Refactoring the signature would result in a lot more changes that don't seem necessary unless we start using this cell more.

Library/SettingsAccountCellType.swift
```diff
@@ -34,6 +34,10 @@ public enum SettingsAccountCellType: SettingsCellTypeProtocol, CaseIterable {
   case currency
   case currencyPicker
```
dusi: Returning `button` for all rows should also be fine with `currencyPicker` being removed soon.

dusi:
[Audit Settings Accessibility](https://trello.com/c/xkODdosY/1125-audit-settings-accessibility)

PR #536, opened 2018-12-21
[BUG] Regenerate tab VCs when user changes currency

ifbarrera:
# 📲 What

When a user changes their currency, we reload all of the tabs so there is no lag in showing project pledge amounts in the user's selected currency.

# 🤔 Why

When we show a project, we first populate the pledge amount using the current `Project` object from the discovery feed. When a user changes their currency, unless we refresh the discovery feed, the `Project` object from discovery will have a "stale" currency until the "updated" currency is populated on the project page. This appears as a "lag" where the user will momentarily see pledge amounts in the old currency before seem them in their preferred currency.

Now, we refresh all the discovery feeds when users change their currency.

# 🛠 How

Combined the `languageChanged` notification into `userLocalePreferencesChanged`, which will now fire both when a user changes their currency, and when you change your language using the beta tools.

# 👀 See

**Before**
wylse3umss

**After**
nfynecq0yu

# ✅ Acceptance criteria

- [x] navigate to change currency, change your currency, then navigate to any project (from profile, discovery, search, activity...) – there should be no lag in displaying pledge amounts in the correct currency
- [x] change language using the beta tools should work as before

### DIFF
--- a/Library/Notifications.swift
+++ b/Library/Notifications.swift
@@ -1,15 +1,15 @@
 import Foundation
 
 public enum CurrentUserNotifications {
-  public static let showNotificationsDialog = "CurrentUserNotifications.showNotificationsDialog"
-  public static let projectSaved = "CurrentUserNotifications.projectSaved"
   public static let dataRequested = "CurrentUserNotifications.dataRequested"
-  public static let savedProjectEmptyStateTapped = "CurrentUserNotifications.savedProjectEmptyStateTapped"
-  public static let sessionStarted = "CurrentUserNotifications.sessionStarted"
-  public static let sessionEnded = "CurrentUserNotifications.sessionEnded"
-  public static let userUpdated = "CurrentUserNotifications.userUpdated"
-  public static let languageChanged = "CurrentUserNotification.languageChanged"
   public static let environmentChanged = "CurrentUserNotification.environmentChanged"
+  public static let localePreferencesChanged = "CurrentUserNotification.localePreferencesChanged"
+  public static let projectSaved = "CurrentUserNotifications.projectSaved"
+  public static let savedProjectEmptyStateTapped = "CurrentUserNotifications.savedProjectEmptyStateTapped"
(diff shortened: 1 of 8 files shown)


Library/Notifications.swift
```diff
@@ -1,15 +1,15 @@
 import Foundation
 
 public enum CurrentUserNotifications {
-  public static let showNotificationsDialog = "CurrentUserNotifications.showNotificationsDialog"
-  public static let projectSaved = "CurrentUserNotifications.projectSaved"
   public static let dataRequested = "CurrentUserNotifications.dataRequested"
+  public static let environmentChanged = "CurrentUserNotification.environmentChanged"
+  public static let localePreferencesChanged = "CurrentUserNotification.localePreferencesChanged"
```
ifbarrera: `localePreferencesChanged` is the only new notification. I just alphabetized this enum.

Library/Notifications.swift
```diff
@@ -18,19 +18,19 @@ public enum UserInfoKeys {
...
+  public static let ksr_sessionStarted = Notification.Name(rawValue: CurrentUserNotifications.sessionStarted)
+  public static let ksr_sessionEnded = Notification.Name(rawValue: CurrentUserNotifications.sessionEnded)
   public static let ksr_showNotificationsDialog =
     Notification.Name(rawValue: CurrentUserNotifications.showNotificationsDialog)
-  public static let ksr_languageChanged = Notification.Name(rawValue:
-    CurrentUserNotifications.languageChanged
-  )
-  public static let ksr_environmentChanged = Notification.Name(rawValue:
-    CurrentUserNotifications.environmentChanged
+  public static let ksr_userLocalePreferencesChanged = Notification.Name(rawValue:
```
ifbarrera: Same here, `ksr_userLocalePreferencesChanged` is the only new notification. The others are now alphabetized.

ifbarrera:
[Preferred currency is shown with a small delay](https://trello.com/c/vshYCprk/1146-preferred-currency-is-shown-with-a-small-delay)

PR #537, opened 2018-12-21
Accessibility - Settings - Change email - Current email

dusi:
# 📲 What

Groups current email title and value label into one accessible element.

# 🤔 Why

This makes it easier for Voice Over to navigate and more human readable to digest.

# 🛠 How

Simply disabling accessibility on labels and using a container view as the accessible element.

Side note:
I've renamed some of the labels on the controller due to the lack of clarity they provided.

# 👀 See

| Before | After |
| --- | --- |
| screen shot 2018-12-21 at 11 37 49 am | screen shot 2018-12-21 at 11 36 57 am |

# ✅ Acceptance criteria

on all the following: iOS 10, iOS 11 & iOS 12
with Voice Over turned ON

- [x] Current email row is one single focus element and is read by Voice Over like so: `Current email - whateveremailthereis@domain.com` (note: the title is localized)

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
@@ -5,8 +5,9 @@ import ReactiveSwift
 import UIKit
 
 internal final class ChangeEmailViewController: UIViewController, MessageBannerViewControllerPresenting {
-  @IBOutlet fileprivate weak var currentEmailLabel: UILabel!
-  @IBOutlet fileprivate weak var currentEmail: UILabel!
+  @IBOutlet fileprivate weak var currentEmailContainer: UIView!
+  @IBOutlet fileprivate weak var currentEmailTitle: UILabel!
+  @IBOutlet fileprivate weak var currentEmailValue: UILabel!
   @IBOutlet fileprivate weak var messageLabelView: UIView!
   @IBOutlet fileprivate weak var newEmailLabel: UILabel!
   @IBOutlet fileprivate weak var newEmailTextField: UITextField!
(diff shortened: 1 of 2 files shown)


dusi:
[Audit Settings Accessibility](https://trello.com/c/xkODdosY/1125-audit-settings-accessibility)

Kickstarter-iOS/Views/Controllers/ChangeEmailViewController.swift
```diff
@@ -86,13 +87,22 @@ internal final class ChangeEmailViewController: UIViewController, MessageBannerV
       |> \.textColor .~ .ksr_red_400
       |> \.text %~ { _ in Strings.We_ve_been_unable_to_send_email() }
 
-    _ = self.currentEmailLabel
+    _ = self.currentEmailContainer
+      |> \.isAccessibilityElement .~ true
+      |> \.accessibilityLabel %~ { _ in
```
justinswart: I suppose the `guard` isn't necessary if we're returning `nil`? Could just be:
```swift
|> \.accessibilityLabel %~ { _ in self.currentEmailTitle.text }
```

justinswart: Oh and you could try `.~ self.currentEmailTitle.text` because this doesn't need to go through a localization function I think.

dusi: What was I thinking 😄

dusi: Turns out the `accessibilityLabel` has to be localized...the way we're combining these two are `accessibilityLabel` & `accessibilityValue` which reads them as one element.

We set the value on line 135 where we bind it to the VM

cdolm92:
Works as expected - tested on iPhone 7(12.1.1) ✅

PR #538, opened 2018-12-21
Ignore project preview deep-links

justinswart:
# 📲 What

Ignores project preview links as these should be handled by mobile safari and not by the app itself.

# 🤔 Why

We don't currently support project preview links in the app but because their URL structure matches a route in our `Navigation` enum the app attempts to open them. This fails and the user is stuck in the app when they should be redirected back out to mobile safari.

# 🛠 How

- Added a `projectPreview` case to `Navigation` so that we can be sure that a given URL is in fact a preview URL.
- If a match is found, we ignore it by returning `.failure`, thereby using the existing functionality to ignore unrecognized URLs.

# 👀 See

https://trello.com/c/4GS7SNZ8

# ✅ Acceptance criteria

- [x] Project preview links should open the app but then redirect back to mobile safari (will need to be tested in Release mode).

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -287,16 +287,13 @@ AppDelegateViewModelOutputs {
     let continueUserActivityWithNavigation = continueUserActivity
       .filter { $0.activityType == NSUserActivityTypeBrowsingWeb }
       .map { activity in (activity, activity.webpageURL.flatMap(Navigation.match)) }
-      .filter(second >>> isNotNil)
 
-    self.continueUserActivityReturnValue <~ continueUserActivityWithNavigation.mapConst(true)
+    self.continueUserActivityReturnValue <~ continueUserActivityWithNavigation.map(second >>> isNotNil)
 
     let deepLinkUrl = Signal
       .merge(
         openUrl.map { $0.url },
-
         self.foundRedirectUrlProperty.signal.skipNil(),
-
         continueUserActivity
           .filter { $0.activityType == NSUserActivityTypeBrowsingWeb }
           .map { $0.webpageURL }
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1365,6 +1365,39 @@ final class AppDelegateViewModelTests: TestCase {
     self.goToMobileSafari.assertValues([unrecognizedUrl], "Go to mobile safari for the unrecognized url.")
   }
 
+  func testEmailDeepLinking_UnrecognizedUrl_ProjectPreview() {
```
justinswart: This test is essentially the same as the one above it but it explicitly uses a known project preview link (containing `token` query param) to test that it will be ignored.

Library/NavigationTests.swift
```diff
@@ -38,6 +38,8 @@ public final class NavigationTests: XCTestCase {
     KSRAssertMatch(.project(.slug("project"), .root, refTag: nil),
                    "/projects/creator/project")
 
+    KSRAssertMatch(nil, "/projects/creator/project?token=4")
```
justinswart: Just proves that we get a `nil` match on this type of URL.

Library/Navigation.swift
```diff
@@ -401,10 +405,22 @@ private func signup(_: RouteParams) -> Decoded<Navigation> {
 }
 
 private func project(_ params: RouteParams) -> Decoded<Navigation> {
-  return curry(Navigation.project)
+  let projectPreview = curry(Navigation.projectPreview)
     <^> params <| "project_param"
     <*> .success(.root)
     <*> params <|? "ref"
+    <*> params <| "token"
```
justinswart: Treats `token` as required for a project preview link.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -407,8 +404,11 @@ AppDelegateViewModelOutputs {
       .filter { $0 == .tab(.me) }
       .ignoreValues()
 
-    self.goToMobileSafari = self.foundRedirectUrlProperty.signal.skipNil()
-      .filter { Navigation.deepLinkMatch($0) == nil }
+    self.goToMobileSafari = Signal.merge(
```
ifbarrera: I *think* you might be able to just use `deeplinkFromUrl.skipNil()` here. Just wanted to make a note of that in case I forget later 😄

justinswart:
[Support project preview links](https://trello.com/c/4GS7SNZ8/1078-support-project-preview-links)

Library/NavigationTests.swift
dusi: Wonder whether we should also add `KSRAssertMatch(nil, "/projects/creator/project?ref=discovery&token=4")` for exhausted url example.

justinswart:
Thanks for digging in and testing!

PR #539, opened 2018-12-27
♿️ - Settings - Message Banner

dusi:
# 📲 What

Improves message banner accessibility.

# 🤔 Why

Presenting message banner isn't communicated to Voice Over at all at this point. This should improve the accessibility of this feature.

# 🛠 How

Originally I intended to use `UIAccessibility` announcements but it showed too unreliable (fighting with Voice Over's focus system). In order to get around this issue I believe a better solution is to actually focus on the banner when it shows up on screen. This not only allows Voice Over to read the message but also allows user to dismiss the banner by double tapping on it.

# ✅ Acceptance criteria

- [x] Presenting message banner on (error, success info) focuses Voice Over on the banner itself and reads the label
- [x] Voice Over reads the message banner as the `label` value + `double tap to dismiss` hint
- [x] Message banner is dismissed after 10 seconds when Voice Over is ON
- [x] Message banner is dismissed after 4 seconds when Voice Over is OFF


Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -30,6 +30,16 @@ final class MessageBannerViewController: UIViewController, NibLoading {
     static let showDuration: TimeInterval = 0.3
   }
 
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    _ = self.backgroundView
```
dusi: Moving this to `viewDidLoad` to better separate concerns (making components accessible vs setting styles)

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -103,18 +114,35 @@ final class MessageBannerViewController: UIViewController, NibLoading {
...
-                    guard let self = self else { return }
-
-                    self.bottomConstraint?.constant = isHidden ? hiddenConstant : 0
-
-                    self.view.superview?.layoutIfNeeded()
-    }, completion: { [weak self] _ in
-      self?.view.isHidden = isHidden
-
-      self?.viewModel.inputs.bannerViewAnimationFinished(isHidden: isHidden)
+    UIView.animate(
```
dusi: Mostly changing the style of completion blocks for better readability but two important things happen here:
1. When the banner is showed we focus accessibility reader on it
2. When the banner is hidden we focus accessibility on the perilously focused element (which will most likely be `Save` button)

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -53,14 +54,34 @@ MessageBannerViewModelInputs, MessageBannerViewModelOutputs {
     self.messageTextAlignment = bannerType.map { $0.textAlignment }
 
     self.bannerMessage = self.messageBannerConfiguration.signal.skipNil().map(second)
+    self.bannerMessageAccessibilityLabel = self.bannerMessage
+      // todo: Needs to be localized - Strings.banner_tap_to_dismiss
+      .map { "\($0), double tap to dismiss" }
```
cdolm92: should "double tap to dismiss" be a hint?

cdolm92:
Awesome, Pavel! This was tested ACs on iPhone 7, iOS 12.1.1 and works as expected ✨

Library/ViewModels/MessageBannerViewModel.swift
dusi: I tried using `accessibilityHint` but VoiceOver would simply not read it. I've seen this happen before, that VoiceOver would in certain situations not read the `hint` but don't really have a clue as to why.

Thinking about this a bit more led me to Apple's documentation - https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint where you can clearly read: "Avoid beginning the phrase with the imperative form of a verb, because this can make the hint sound like a command. For example, do not create a hint such as “Play the song” or “Purchase the item." which made me extend the `accessibilityLabel` instead.

I think in this situation using an `accessibilityLabel` works slightly better by using comma (there is a small pause) - if we were to use a `hint` (and it in fact worked) the pause would be considerably bigger.

cdolm92: Given that the sentence is using an imperative form verb and that VoiceOver is not reading the hint I think this is a great work around. Thanks for explaining!

PR #540, opened 2019-01-03
Update conformance to Hashable protocol to make Swiftlint happy

dusi:
# 📲 What

Updates conformance to `Hashable` protocol.

# 🤔 Why

This is a required Swiftlint rule which is currently failing branches that are not up to date.

# 🛠 How

Replace `Hashable` conformance from `var hashValue: Int` to `func hash(into hasher: inout Hasher)`

# 👀 See

https://nshipster.com/hashable/
https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md

# ✅ Acceptance criteria

- [x] All tests pass

### DIFF
--- a/Library/Format.swift
+++ b/Library/Format.swift
@@ -323,13 +323,22 @@ private struct DateFormatterConfig {
 }
 
 extension DateFormatterConfig: Hashable {
-  fileprivate var hashValue: Int {
-    return
-      (self.template?.hashValue ?? 0)
-        ^ (self.dateStyle?.hashValue ?? 0)
-        ^ self.locale.hashValue
-        ^ (self.timeStyle?.hashValue ?? 0)
-        ^ self.timeZone.hashValue
+  fileprivate func hash(into hasher: inout Hasher) {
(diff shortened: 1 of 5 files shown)


Library/Format.swift
```diff
@@ -323,13 +323,22 @@ private struct DateFormatterConfig {
...
 extension DateFormatterConfig: Hashable {
-  fileprivate var hashValue: Int {
-    return
-      (self.template?.hashValue ?? 0)
-        ^ (self.dateStyle?.hashValue ?? 0)
-        ^ self.locale.hashValue
-        ^ (self.timeStyle?.hashValue ?? 0)
-        ^ self.timeZone.hashValue
+  fileprivate func hash(into hasher: inout Hasher) {
+    if let template = self.template {
```
dusi: I'm not entirely sure about these...previously we would return `String` or `Int` which the `hasher.combine()` function doesn't like..the type should be known beforehand.

I assumed that adding `0` doesn't really affect the hash value but I might be wrong.

PR #541, opened 2019-01-03
Update readme with more generic getting started instructions

dusi:
# 📲 What

Small README update

# 🤔 Why

We don't require specific version of Xcode (mentioned previously) plus there are multiple channels how to get Xcode (Mac App Store vs direct download from developer.apple.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ and read more about our journey to open source [here](https://kickstarter.engine
 
 ## Getting Started
 
-1. [Download](https://developer.apple.com/xcode/download/) the Xcode 9 release.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,7 @@ and read more about our journey to open source [here](https://kickstarter.engine
 
 ## Getting Started
 
-1. [Download](https://developer.apple.com/xcode/download/) the Xcode 9 release.
+1. Install Xcode
```
justinswart: Just added a full-stop 😬 
```suggestion
1. Install Xcode.
```

PR #544, opened 2019-01-05
♿️ - Settings - Switch Toggle Rows 

dusi:
# 📲 What

Fixes the accessibility of our Settings cells with `UISwitch` control.

This includes:
* Settings > Account > Privacy
* Settings > Notifications > Project notifications
* Settings > Newsletters

# 🤔 Why

The intention is to simplify the amount of information we expose to VoiceOver reader. Our settings cells that have a `UISwitch` control currently expose multiple elements (label, switch control, header/footer). In order to make it easier for visually impaired users to navigate through settings we only want to expose the `UISwitch` control and provide enough context in form of a label or hint.

Since we're using custom cells we don't get the native behaviour for free. Normally (in order to support accessibility) a `UISwitch` control would be placed inside of a cell's `accessoryView`. This would result in the following key functionality (which we have to mitigate):

1. The cell would only expose the `UISwitch` control, making it instantly possible to toggle it
- go to Settings.app > Do Not Disturb to see this in action
- we are able to implement this by only exposing the `UISwitch` control to the VoiceOver reader like so `self.accessibilityElements = [self.mySwitchControl]`

2. The `UISwitch` control focus area would be the whole cell
- we are not able to achieve the above exactly reflecting native behaviour (visually)
- but what we get by exposing the `UISwitch` control is identical behaviour (tapping anywhere on the cell will focus on the `UISwitch` control except visually the focus area is only the `UISwitch` control frame) - this honestly is good enough since the functionality stays the same.

# 🛠 How

Tweaking bunch of accessibility thingies here and there 😄 

# 👀 See

**Settings > Account > Privacy**

| Before | Before | Before | After |
| --- | --- | --- | --- |
| Exposes content view which uses label's accessibility label .. | .. and switch control .. | .. and footer | Only exposes switch control with label & footer information |
| img_0050 | img_0051 | img_0052 | img_0047 |

**Settings > Notifications > Project notifications**

| Before | Before | After |
| --- | --- | --- |
| Exposes content view which uses label's accessibility label .. | .. and switch control | Only exposes switch control with label information |
| img_0053 | img_0054 | img_0048 |

**Settings > Newsletters**

| Before | Before | After |
| --- | --- | --- |
| Exposes content view which uses label's accessibility label .. | .. and switch control | Only exposes switch control with label information |
| img_0055 | img_0056 | img_0049 |

# ✅ Acceptance criteria

Test under the following conditions
👉 iOS 10, iOS 11 & iOS 12
👉Voice Over ON
👉 In order to test localization prefer to QA with one of the supported language set as the primary 📱 language

**Settings > Account > Privacy**

- [x] Swiping right/left navigates between switch controls
- [x] VoiceOver reads switch control as "title + switch button + state + footer + double tap to toggle setting" (all localized)
- [x] Tapping anywhere on the cell focuses on the switch control
- [x] Tapping anywhere on the footer (which does not belong to the cell) does nothing (unfortunately but based on our current implementation)
- [x] Toggling `Following` OFF and `Canceling` the alert re-focuses VoiceOver to the `Following` cell (which is now back to ON state and read by VoiceOver)
- [x] Toggling `Following` OFF and `Accepting` the alert re-focuses VoiceOver to the `Following` cell (which is now in OFF state and read by VoiceOver)

**Settings > Notifications > Project notifications**

- [x] Swiping right/left navigates between switch controls
- [x] VoiceOver reads switch control as "title + switch button + state + double tap to toggle setting" (all localized)
- [x] Tapping anywhere on the cell focuses on the switch control

**Settings > Newsletters**

- [x] Swiping right/left navigates between switch controls
- [x] VoiceOver reads switch control as "title + switch button + state + header or footer + double tap to toggle setting" (all localized)
- [x] Tapping anywhere on the cell focuses on the switch control
- [x] Tapping anywhere on the header/footer focuses on the switch control

Since some of this work has touched the following cell functionality the following is a regression testing:

👉Voice Over OFF
- [x] Toggle following cell OFF > alert will appear > cancel > toggle will switch back ON with animation
- [x] Toggle following cell OFF > alert will appear > confirm > toggle will stay OFF
- [x] Toggle following cell ON > no alert will appear > toggle will switch ON

### DIFF
--- a/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
@@ -20,8 +20,21 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
   @IBOutlet fileprivate weak var followingSwitch: UISwitch!
   @IBOutlet fileprivate var separatorView: [UIView]!
 
-  internal func configureWith(value user: User) {
-    self.viewModel.inputs.configureWith(user: user)
+  override func awakeFromNib() {
+    super.awakeFromNib()
+
+    _ = self
+      |> \.accessibilityElements .~ [self.followingSwitch]
+
+    _ = self.followingSwitch
+      |> \.accessibilityLabel %~ { _ in Strings.Following() }
+  }
+
+  internal func configureWith(value: SettingsPrivacyStaticCellValue) {
+    self.viewModel.inputs.configureWith(user: value.user)
+
+    _ = self.followingSwitch
+      |> \.accessibilityHint %~ { _ in value.cellType.description }
   }
 
   internal override func bindStyles() {
--- a/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
@@ -34,7 +34,7 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
     self.viewModel.inputs.configureWith(user: value.user)
 
     _ = self.followingSwitch
-      |> \.accessibilityHint %~ { _ in value.cellType.description }
+      |> \.accessibilityHint .~ value.cellType.description
   }
 
   internal override func bindStyles() {
--- a/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
@@ -76,6 +76,10 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
     self.followingSwitch.rac.on = self.viewModel.outputs.followingPrivacyOn
   }
 
+  func toggleOn(animated: Bool = true) {
+    self.followingSwitch.setOn(true, animated: animated)
+  }
+
   @IBAction func followingPrivacySwitchTapped(_ followingPrivacySwitch: UISwitch) {
     self.viewModel.inputs.followTapped(on: followingPrivacySwitch.isOn)
   }
--- a/Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
@@ -20,19 +20,28 @@ internal final class SettingsNewslettersCell: UITableViewCell, ValueCell {
 
   public weak var delegate: SettingsNewslettersCellDelegate?
 
-  func configureWith(value: (newsletter: Newsletter, user: User)) {
+  override func awakeFromNib() {
+    super.awakeFromNib()
 
+    _ = self
+      |> \.accessibilityElements .~ [self.newslettersSwitch]
+  }
+
+  func configureWith(value: (newsletter: Newsletter, user: User)) {
     self.viewModel.inputs.configureWith(value: value)
 
+    _ = self.newslettersSwitch
+      |> \.accessibilityLabel %~ { _ in value.newsletter.displayableName }
+      |> \.accessibilityHint %~ { _ in value.newsletter.displayableDescription }
+
     _ = self.newslettersLabel
-      |> UILabel.lens.text %~ { _ in value.newsletter.displayableName }
+      |> \.text %~ { _ in value.newsletter.displayableName }
 
     _ = self.newslettersDescriptionLabel
-      |> UILabel.lens.text %~ { _ in value.newsletter.displayableDescription }
+      |> \.text %~ { _ in value.newsletter.displayableDescription }
   }
 
   override func bindStyles() {
-
     _ = self.separatorViews
       ||> separatorStyle
 
--- a/Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
@@ -31,14 +31,14 @@ internal final class SettingsNewslettersCell: UITableViewCell, ValueCell {
     self.viewModel.inputs.configureWith(value: value)
 
     _ = self.newslettersSwitch
-      |> \.accessibilityLabel %~ { _ in value.newsletter.displayableName }
-      |> \.accessibilityHint %~ { _ in value.newsletter.displayableDescription }
+      |> \.accessibilityLabel .~ value.newsletter.displayableName
+      |> \.accessibilityHint .~ value.newsletter.displayableDescription
 
     _ = self.newslettersLabel
-      |> \.text %~ { _ in value.newsletter.displayableName }
+      |> \.text .~ value.newsletter.displayableName
 
     _ = self.newslettersDescriptionLabel
-      |> \.text %~ { _ in value.newsletter.displayableDescription }
+      |> \.text .~ value.newsletter.displayableDescription
   }
 
   override func bindStyles() {
--- a/Kickstarter-iOS/Views/Cells/SettingsPrivacyRecommendationCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsPrivacyRecommendationCell.swift
@@ -12,8 +12,21 @@ internal final class SettingsPrivacyRecommendationCell: UITableViewCell, ValueCe
   @IBOutlet fileprivate weak var recommendationsSwitch: UISwitch!
   @IBOutlet fileprivate var separatorView: [UIView]!
 
-  internal func configureWith(value user: User) {
-    self.viewModel.inputs.configureWith(user: user)
+  override func awakeFromNib() {
+    super.awakeFromNib()
+
+    _ = self
+      |> \.accessibilityElements .~ [self.recommendationsSwitch]
+
+    _ = self.recommendationsSwitch
+      |> \.accessibilityLabel %~ { _ in Strings.Recommendations() }
+  }
+
+  internal func configureWith(value: SettingsPrivacyStaticCellValue) {
+    self.viewModel.inputs.configureWith(user: value.user)
+
+    _ = self.recommendationsSwitch
+      |> \.accessibilityHint %~ { _ in value.cellType.description }
   }
 
   internal override func bindStyles() {
--- a/Kickstarter-iOS/Views/Cells/SettingsPrivacyRecommendationCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsPrivacyRecommendationCell.swift
@@ -26,7 +26,7 @@ internal final class SettingsPrivacyRecommendationCell: UITableViewCell, ValueCe
     self.viewModel.inputs.configureWith(user: value.user)
 
     _ = self.recommendationsSwitch
-      |> \.accessibilityHint %~ { _ in value.cellType.description }
+      |> \.accessibilityHint .~ value.cellType.description
   }
 
   internal override func bindStyles() {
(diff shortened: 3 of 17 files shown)


cdolm92:
Tested on iPhone 7/iOS 12.1.1, works as expected. Only suggestion is a11y ACs for when the following switch is toggle. Only suggesting this because a confirmation alert is presented when user toggles off and VoiceOver reads the switch as off. If the user decides to cancel on the alert VoiceOver does not read that the switch has reverted back to on. I'm thinking this is happening because there is no focus on that switch after the alert gets dismissed. But it would be great if the ACs could clarify that behavior. Let me know if you need me to explain this better irl.

dusi:
I've made sure that after the user toggles `Account > Privacy > Following` toggle it re-focuses VoiceOver back to the `Following` cell which reads the toggle value, etc.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -124,7 +124,9 @@ final class MessageBannerViewController: UIViewController, NibLoading {
         if isHidden {
           // Tells Voice Over to resign focus of the message banner.
           // This causes the reader to focus on the previously selected element.
-          UIAccessibility.post(notification: UIAccessibility.Notification.layoutChanged, argument: nil)
+          if AppEnvironment.current.isVoiceOverRunning() {
```
dusi: I've placed these under `isVoiceOverRunning()` check in order to make this consistent thorough the code base

Kickstarter-iOS/Views/Controllers/SettingsPrivacyViewController.swift
```diff
@@ -62,8 +62,14 @@ internal final class SettingsPrivacyViewController: UITableViewController {
     self.viewModel.outputs.refreshFollowingSection
       .observeForUI()
       .observeValues { [weak self] _ in
-        let followingSection = IndexSet(integer: Section.following.rawValue)
-        self?.tableView.reloadSections(followingSection, with: .none)
+        let section = Section.following.rawValue
+
+        self?.tableView.reloadSections(IndexSet(integer: section), with: .none)
+
+        if AppEnvironment.current.isVoiceOverRunning() {
```
dusi: Another `isVoiceOverRunning()` check

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -280,6 +280,7 @@ DashboardViewModelType {
       .takeWhen(self.messagesCellTappedProperty.signal)
 
     self.focusScreenReaderOnTitleView = self.viewWillAppearAnimatedProperty.signal.ignoreValues()
+      .filter { AppEnvironment.current.isVoiceOverRunning() }
```
dusi: If VoiceOver is OFF there's no need for this signal to be triggered

Library/ViewModels/SettingsPrivacyViewModel.swift
```diff
@@ -81,7 +81,10 @@ SettingsPrivacyViewModelInputs, SettingsPrivacyViewModelOutputs {
    self.updateCurrentUser = Signal.merge(updatedFetchedUser,
                                          previousUserOnError)
 
-   self.refreshFollowingSection = self.didCancelSocialOptOutProperty.signal
+    self.refreshFollowingSection = Signal.merge(
```
dusi: Both `Cancel` and `Confirm` now triggers `Following` section refresh in order for the VoiceOver to re-focus on the cell

Library/ViewModels/DashboardProjectsDrawerViewModel.swift
```diff
@@ -60,6 +60,7 @@ DashboardProjectsDrawerViewModelInputs, DashboardProjectsDrawerViewModelOutputs
     self.notifyDelegateDidAnimateOut = self.animateOutCompletedProperty.signal
 
     self.focusScreenReaderOnFirstProject = self.animateInCompletedProperty.signal
+      .filter { AppEnvironment.current.isVoiceOverRunning() }
```
dusi: If VoiceOver is OFF there's no need for this signal to be triggered

dusi:
[♿️ Settings - Switch toggle rows](https://trello.com/c/Drlp347o/1155-%E2%99%BF%EF%B8%8F-settings-switch-toggle-rows)

cdolm92:
Tested the new ACs on iPhone Sim and iPhone 7 and I found one issue.

follow-toggle-issues

Toggle should be off when user accepts/confirms

Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
```diff
@@ -63,6 +76,10 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
     self.followingSwitch.rac.on = self.viewModel.outputs.followingPrivacyOn
   }
 
+  func toggleOn(animated: Bool = true) {
```
dusi: I don't feel strongly about the default attribute `animated: Bool = true` but I find it useful to always explicitly state this instead of calling something with `animated: true` - makes it easier to refactor later on

Kickstarter-iOS/Views/Controllers/SettingsPrivacyViewController.swift
```diff
@@ -59,11 +59,18 @@ internal final class SettingsPrivacyViewController: UITableViewController {
...
       .observeForUI()
       .observeValues { [weak self] _ in
-        let followingSection = IndexSet(integer: Section.following.rawValue)
-        self?.tableView.reloadSections(followingSection, with: .none)
+        let indexPath = IndexPath(row: 0, section: Section.following.rawValue)
+        let cell = self?.tableView.cellForRow(at: indexPath) as? SettingsFollowCell
+        cell?.toggleOn()
+    }
+
+    self.viewModel.outputs.focusScreenReaderOnFollowingCell
```
dusi: Explicit signal instead of a side effect previously when we had this in `resetFollowingSection`

Kickstarter-iOS/Views/Controllers/SettingsPrivacyViewController.swift
```diff
@@ -59,11 +59,18 @@ internal final class SettingsPrivacyViewController: UITableViewController {
         self?.present(UIAlertController.genericError(message), animated: true, completion: nil)
     }
 
-    self.viewModel.outputs.refreshFollowingSection
+    self.viewModel.outputs.resetFollowingSection
```
dusi: `reset` seems better than `refresh` in this case

Library/ViewModels/SettingsPrivacyViewModel.swift
```diff
@@ -14,7 +14,8 @@ public protocol SettingsPrivacyViewModelInputs {
 }
 
 public protocol SettingsPrivacyViewModelOutputs {
-  var refreshFollowingSection: Signal<Void, NoError> { get }
+  var focusScreenReaderOnFollowingCell: Signal<Void, NoError> { get }
+  var resetFollowingSection: Signal<Void, NoError> { get }
```
cdolm92: alphabetize

Library/ViewModels/SettingsPrivacyViewModelTests.swift
```diff
@@ -9,20 +9,40 @@ import Prelude
 
 internal final class SettingsPrivacyViewModelTests: TestCase {
   let vm = SettingsPrivacyViewModel()
-  internal let refreshFollowingSection = TestObserver<Void, NoError>()
+  internal let focusScreenReaderOnFollowingCell = TestObserver<Void, NoError>()
+  internal let resetFollowingSection = TestObserver<Void, NoError>()
   internal let reloadData = TestObserver<User, NoError>()
```
cdolm92: here too

Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
```diff
@@ -20,19 +20,28 @@ internal final class SettingsNewslettersCell: UITableViewCell, ValueCell {
...
+
+    _ = self
+      |> \.accessibilityElements .~ [self.newslettersSwitch]
+  }
 
+  func configureWith(value: (newsletter: Newsletter, user: User)) {
     self.viewModel.inputs.configureWith(value: value)
 
+    _ = self.newslettersSwitch
+      |> \.accessibilityLabel %~ { _ in value.newsletter.displayableName }
```
cdolm92: This string does not look like it's being evaluated for localization so we don't need to use %~.
`|> \.accessibilityLabel .~ value.newsletter.displayableName` should be enough.

Kickstarter-iOS/Views/Cells/SettingsPrivacySwitchCell.swift
```diff
@@ -18,17 +18,30 @@ final class SettingsPrivacySwitchCell: UITableViewCell, ValueCell, NibLoading {
...
     _ = self.primaryDescriptionLabel
-    |> UILabel.lens.text .~ value.cellType.primaryDescriptionString
+      |> \.text .~ value.cellType.primaryDescription
 
     _ = self.secondaryDescriptionLabel
-    |> UILabel.lens.text .~ value.cellType.secondaryDescriptionString
+      |> \.text .~ value.cellType.secondaryDescription
+
+    _ = self.switchButton
+      |> \.accessibilityLabel %~ { _ in value.cellType.title }
```
cdolm92: same here

Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
justinswart: This sounds correct to me - my understanding of `.~` vs `%~` is that in the latter the current value is passed in to a function so that one might return a value based on the current. Generally, the way that we use it we tend to ignore the current value but we use `%~` because we can pass it a closure that is evaluated at runtime. We tend to use this for `Strings.*` functions because under the hood those are calling `localizedString()` which means that if we change the language on our `Environment` between calls to `bindStyles()` we will get the updated translations each time as opposed to the original function being captured and executed each time.

In the case of setting this value from another variable as in the case above, it's up to the place where that variable is initially set in order for the localization to take place.

^ all of this is just based on my understanding of how it works, tbh I'm not sure if that's completely correct. I think that a way to test this would be with snapshot tests. I will try to find an example.

dusi:
Addressed your comments @cdolm92 . Please review my last commit (https://github.com/kickstarter/ios-oss/pull/544/commits/a2fd42a2b61b5a145459f332cb38938b7cd0fb00) where I've tried to properly localize strings.

Kickstarter-iOS/Views/Cells/SettingsPrivacyRecommendationCell.swift
```diff
@@ -26,7 +26,7 @@ internal final class SettingsPrivacyRecommendationCell: UITableViewCell, ValueCe
     self.viewModel.inputs.configureWith(user: value.user)
 
     _ = self.recommendationsSwitch
-      |> \.accessibilityHint %~ { _ in value.cellType.description }
+      |> \.accessibilityHint .~ value.cellType.description
```
cdolm92: Thanks for changing these too!

Kickstarter-iOS/Views/Cells/SettingsNewslettersCell.swift
```diff
@@ -31,14 +31,14 @@ internal final class SettingsNewslettersCell: UITableViewCell, ValueCell {
...
+      |> \.accessibilityHint .~ value.newsletter.displayableDescription
 
     _ = self.newslettersLabel
-      |> \.text %~ { _ in value.newsletter.displayableName }
+      |> \.text .~ value.newsletter.displayableName
 
     _ = self.newslettersDescriptionLabel
-      |> \.text %~ { _ in value.newsletter.displayableDescription }
+      |> \.text .~ value.newsletter.displayableDescription
   }
```
cdolm92: And the ones here!

PR #545, opened 2019-01-07
♿️- Settings - Change email / Change password input fields

cdolm92:
# 📲 What

Accessibility improvements to textfields in Change Email and Change Password screens.

# 🤔 Why

Previously VoiceOver read both the title labels and textfields we found that this was not a good pattern because it adds to the complexity of focusable elements and doesn’t provide non-visual value.

# 🛠 How

We disabled accessibility on all title labels. We then set the accessibility label of the textfields to the title labels text.

# 👀 See
# ✅ Acceptance criteria
on all the following: iOS 10, iOS 11 & iOS 12
with Voice Over turned ON

**Test Change Email Screen**

Test Tap on row
- [x] Navigate to `Settings > Account > Change email`
- [x] Tap `New email` row, should read `New email - Email Address - Text Field - double tap to edit`
- [x] Tap `Current password` row, should read `Current password - Password - Text Field - double tap to edit`

Test Double Tap on TextField
- [x] Navigate to `Settings > Account > Change email`
- [x] Tap `New email` row, should read `New email - Email Address - Text Field - double tap to edit`
- [x] Double-tap should now read `New Email - text field - is editing - "Email Address" - character mode - insertion point at start`
- [x] Tap `Current Password - Secure Text Field - is editing - "Password" - character mode - insertion point at start`

Test Tap on row when text is present
- [x] Navigate to `Settings > Account > Change email`
- [x] Tap `New email` row, should read `New email - (email address read) - Text Field - double tap to edit`
- [x] Tap `Current password` row, should read `Current password - (number of characters) - Secure Text Field - double tap to edit`

Test Tap on textfield when text is present
- [x] Navigate to `Settings > Account > Change email`
- [x] Enter some text on each textfield
- [x] Double Tap on one of the textfields so that cursor is present
- [x] For **New email**  row  should read: `New Email - text field - is editing - (character) - character mode - insertion point at end`
- [x] For **Current password**  row should read:  `Current Password - Secure Text Field - is editing - (number of characters) - character mode - insertion point at end`

**Test Change Password Screen**

Test Tap on row
- [x] Navigate to `Settings > Account > Change password`
- [x] Tap `Current password` row, should read `Current password - Password - Text Field - double tap to edit`
- [x] Tap `New password` row, should read `New password - Password - Text Field - double tap to edit`
- [x] Tap `Confirm new password` row, should read `Confirm new password - Password - Text Field - double tap to edit`

Test Tap on row when text is present
- [x] Navigate to `Settings > Account > Change password`
[...]

PR #535, opened 2018-12-21
Accessibility - Settings - Change email status

dusi:
# 📲 What

Adds accessibility hint to the `Change email` cell in case the email is unverified.

# 🤔 Why

Better reflects visual UI to Voice Over users

# 🛠 How

Simply adds `accessibilityHint`. This is so that the order of the reader is as follows: `Change email` - `Button` - `This email address is unverified`. This should be a suitable solution since the information is secondary.

# ✅ Acceptance criteria

on all the following: iOS 10, iOS 11 & iOS 12
with Voice Over turned ON

- [x] Focusing on Change email cell in `Settings > Account` reads as `Change email - Button - This email address is unverified`

PR #532, opened 2018-12-19
Accessibility - Settings - Dismiss button

dusi:
# 📲 What

Adds accessibility label to the dismiss button & fixes the minimum width of the button in order to follow HIG best practices for tappable elements (44x44 pts)

# 🤔 Why

Better accessibility

# 🛠 How

Simply adding accessibility label (this will prevent from using the image name)

# 👀 See

| Before | After |
| --- | --- |
| screen shot 2018-12-19 at 2 32 17 pm | screen shot 2018-12-19 at 2 33 59 pm |
| screen shot 2018-12-19 at 2 32 26 pm | screen shot 2018-12-19 at 2 34 58 pm |

# ✅ Acceptance criteria

- [x] Voice Over reads the dismiss button as "Dismiss - Button" (please not that this is localized)
- [x] Accessibility Inspector doesn't show min-width warning when running the audit on Settings screen

PR #546, opened 2019-01-07
♿️ - Settings

dusi:
# 📲 What

Settings accessibility cleanup

# 🤔 Why

Some of our settings screens needed a bit of a11y love 🙏 

# 🛠 How

# 👀 See

#532 - Dismiss button
#534 - Settings - rows
#535 - Settings - Change email status
#537 - Settings - Change email - Current email
#545 - Settings - Change email / password input fields
#549 - Settings - Account - Privacy - Delete my KSR account button
#550 - Settings - Heading
#552 - Settings - Notifications - Manage project notifications + Email frequency
#547 - Settings - Notifications - Daily Digest picker bug
#544 - Settings - Switch toggle rows
#539 - Settings - Message banner


Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
```diff
@@ -63,6 +76,10 @@ internal final class SettingsFollowCell: UITableViewCell, ValueCell {
     self.followingSwitch.rac.on = self.viewModel.outputs.followingPrivacyOn
   }
 
+  func toggleOn(animated: Bool = true) {
+    self.followingSwitch.setOn(true, animated: animated)
```
justinswart: Why is `true` hard-coded here 🤔

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
```diff
@@ -103,18 +114,39 @@ final class MessageBannerViewController: UIViewController, NibLoading {
...
-
-                    self.view.superview?.layoutIfNeeded()
-    }, completion: { [weak self] _ in
-      self?.view.isHidden = isHidden
-
-      self?.viewModel.inputs.bannerViewAnimationFinished(isHidden: isHidden)
+    UIView.animate(
+      withDuration: duration,
+      delay: 0.0,
+      options: UIView.AnimationOptions.curveEaseInOut,
```
justinswart: Can this infer `.curveEaseInOut` or `[.curveEaseInOut]`?

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
```diff
@@ -89,13 +90,40 @@ internal final class SettingsNotificationsViewController: UIViewController {
...
-
-      self.view.setNeedsLayout()
-      self.view.layoutIfNeeded()
-    }
+    UIView.animate(
+      withDuration: 0.25,
+      animations: { [weak self] in
+        guard let self = self else { return }
+
+        if !isHidden {
```
justinswart: Should we combine the nested `if`s here into one? `if !isHidden && AppEnvironment.current.isVoiceOverRunning()`, also on line `115`.

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -89,6 +84,21 @@ final class SettingsViewController: UIViewController {
       .observeValues { [weak self] link in self?.goToAppStore(link: link) }
   }
 
+  private func leftBarButtonItem() -> UIBarButtonItem {
```
justinswart: This can be:
```suggestion
  private func leftBarButtonItem() -> UIBarButtonItem {
    return UIBarButtonItem(
      image: UIImage(named: "icon--cross"),
      style: .plain,
      target: self,
      action: #selector(closeButtonPressed)
    )
    |> \.accessibilityLabel %~ { _ in Strings.Dismiss() }
    |> \.width .~ 44
  }
```

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -53,14 +54,34 @@ MessageBannerViewModelInputs, MessageBannerViewModelOutputs {
     self.messageTextAlignment = bannerType.map { $0.textAlignment }
 
     self.bannerMessage = self.messageBannerConfiguration.signal.skipNil().map(second)
+    self.bannerMessageAccessibilityLabel = self.bannerMessage
+      // todo: Needs to be localized - Strings.banner_tap_to_dismiss
+      .map { "\($0), double tap to dismiss" }
```
justinswart: Has this been added to our strings yet?

I think it's always a good idea to use the `localizedString()` function directly if we don't yet have it in `Strings.swift` that way if we forget about this it's going to use localization or if we need an easy way to find it later we can just search for uses of `localizedString()`.

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -53,14 +54,34 @@ MessageBannerViewModelInputs, MessageBannerViewModelOutputs {
...
 
     let bannerViewShouldHide = self.showBannerViewProperty.signal.negate()
 
-    let postAnimationBannerViewShouldHide = self.bannerViewIsHiddenProperty.signal
-      .filter { isFalse($0) }
+    let dismissBanner = self.bannerViewIsHiddenProperty.signal
+      .filter(isFalse)
+
+    // Dismisses the banner after 4 seconds when VoiceOver is OFF
```
justinswart: Were we certain about the use of `debounce` here vs `delay`? I remember chatting about it but can't recall 😅

Kickstarter-iOS/Views/Cells/SettingsFollowCell.swift
dusi: Good question. So this method could be more generic (i.e. `toggle(on: Bool, animated: Bool)`) but since we only ever need to toggle back ON we have used more specific function named `toggleOn` which implies that we're switching the toggle ON...hence `setOn(true, ...)`

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
dusi: Can try

Library/ViewModels/MessageBannerViewModel.swift
dusi: Yes, we actually really want to use `debounce`. The banner was showing a weird behaviour when using `delay`.

Kickstarter-iOS/Views/Controllers/MessageBannerViewController.swift
dusi: Hopefully it won't break the compiler :trollface:

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
dusi: Wow, Github really messed up here..had to manually fix it later on :)

Library/ViewModels/MessageBannerViewModel.swift
dusi: I completely forgot about this...good catch Justin.

Created https://github.com/kickstarter/kickstarter/pull/14146 and will block this PR until we have those strings.

dusi: Also I had to update the string to be in form of a sentence to better match server response..

Before
screen shot 2019-01-31 at 11 19 02 am

After
screen shot 2019-01-31 at 11 27 03 am

dusi: Also note that we should remove `project_count_projects_backed` when https://github.com/kickstarter/kickstarter/pull/13979#pullrequestreview-198853575 gets merged

dusi:
Addressed all outstanding feedback

Kickstarter-iOS/Locales/de.lproj/Localizable.strings
```diff
@@ -247,6 +247,7 @@
 "Manage_your_pledge_below" = "Folgenden Beitrag verwalten";
 "Manage_your_reward" = "Belohnung verwalten";
 "Message_backer" = "Nachricht an Unterstützer senden";
+"Message_banner.accessibility.Double_tap_to_dismiss" = "Double tap to dismiss.";
```
dusi: Currently this string is not localized, but based on the conversation we had with @peat it should not delay merging stuff to `master`/:shipit:

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
```diff
@@ -89,13 +90,37 @@ internal final class SettingsNotificationsViewController: UIViewController {
...
+          self.emailFrequencyPickerView.accessibilityViewIsModal = true
+
+          UIAccessibility.post(
+            notification: UIAccessibility.Notification.screenChanged,
+            argument: self.emailFrequencyPickerView
+          )
+        }
+
+        self.emailPickerViewTopConstraint.constant = isHidden ? 0 : self.emailFrequencyPickerView.frame.height
+        self.view.setNeedsLayout()
```
justinswart: Have you tested without `setNeedsLayout()`? I think we only need `layoutIfNeeded()`.

Library/ViewModels/MessageBannerViewModel.swift
```diff
@@ -53,14 +54,33 @@ MessageBannerViewModelInputs, MessageBannerViewModelOutputs {
     self.messageTextAlignment = bannerType.map { $0.textAlignment }
 
     self.bannerMessage = self.messageBannerConfiguration.signal.skipNil().map(second)
+    self.bannerMessageAccessibilityLabel = self.bannerMessage
+      .map { "\($0) \(Strings.Message_banner_accessibility_Double_tap_to_dismiss())" }
```
justinswart: I guess we don't have much choice here but just something to bear in mind is that string concatenation can break translations for languages where the direction isn't left to right.

Library/ViewModels/SettingsPrivacyViewModel.swift
```diff
@@ -78,10 +79,15 @@ SettingsPrivacyViewModelInputs, SettingsPrivacyViewModelOutputs {
...
+    self.updateCurrentUser = Signal.merge(updatedFetchedUser, previousUserOnError)
 
-   self.refreshFollowingSection = self.didCancelSocialOptOutProperty.signal
+    self.resetFollowingSection = self.didCancelSocialOptOutProperty.signal
+
+    self.focusScreenReaderOnFollowingCell = Signal.merge(
+      self.didCancelSocialOptOutProperty.signal,
+      self.didConfirmSocialOptOutProperty.signal
+    )
+      .filter { _ in AppEnvironment.current.isVoiceOverRunning() }
```
justinswart: Could we align the `.` with the last parenthesis?

Kickstarter-iOS/Views/Controllers/SettingsNotificationsViewController.swift
dusi: Not originally my code...but have tried removing and it looks like everything works. Agree that we should only need to call `layoutIfNeeded()`.

PR #552, opened 2019-01-11
♿️ - Settings > Notifications - Manage project notifications + Email frequency

dusi:
# 📲 What

Improves the accessibility of the following rows
* Settings > Notifications > Manage project notifications
* Settings > Notifications > Email frequency

# 🤔 Why

Making a11y great again

# 🛠 How

Using `accessibilityTraits` and `accessibilityLabel` + `accessibilityValue`

# ✅ Acceptance criteria

Test under the following conditions
👉 iOS 10, iOS 11 & iOS 12
👉Voice Over ON
👉 In order to test localization prefer to QA with one of the supported language set as the primary 📱 language

- [x] Settings > Notifications > Manage project notifications row is read by VoiceOver as `Manage project notification, X, button` where X is the number of projects backed so (i.e. `Manage project notifications, twelve, button` (make sure this is localized)

- [x] Settings > Notifications > Email frequency row is read by VoiceOver as `Email frequency, Y, button` where X is the selected option of frequency so (i.e. `Email frequency, Individual emails, button`) (make sure to login as a creator to see this option)

### DIFF
--- a/Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
@@ -100,8 +100,6 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
     self.projectCountLabel.rac.text = viewModel.outputs.projectCountText
     self.pushNotificationsButton.rac.selected = viewModel.outputs.pushNotificationsEnabled
     self.pushNotificationsButton.rac.hidden = viewModel.outputs.pushNotificationButtonIsHidden
-    self.projectCountLabel.rac.accessibilityHint = viewModel.outputs
-      .manageProjectNotificationsButtonAccessibilityHint
 
     viewModel.outputs.enableButtonAnimation
     .observeForUI()
--- a/Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
+++ b/Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
@@ -34,6 +34,10 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
 
     viewModel.inputs.configure(with: cellValue)
 
+    _ = self
+      |> \.accessibilityTraits .~ cellValue.cellType.accessibilityTraits
+      |> \.accessibilityLabel %~ { _ in cellValue.cellType.title }
+
     _ = titleLabel
       |> UILabel.lens.text .~ cellValue.cellType.title
 
@@ -90,6 +94,7 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
   override func bindViewModel() {
     super.bindViewModel()
 
+    self.rac.accessibilityValue = viewModel.outputs.projectCountText
     self.emailNotificationsButton.rac.selected = viewModel.outputs.emailNotificationsEnabled
     self.emailNotificationsButton.rac.hidden = viewModel.outputs.emailNotificationButtonIsHidden
     self.projectCountLabel.rac.text = viewModel.outputs.projectCountText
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -90,13 +94,12 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
   override func bindViewModel() {
     super.bindViewModel()
 
+    self.rac.accessibilityValue = viewModel.outputs.projectCountText
     self.emailNotificationsButton.rac.selected = viewModel.outputs.emailNotificationsEnabled
     self.emailNotificationsButton.rac.hidden = viewModel.outputs.emailNotificationButtonIsHidden
     self.projectCountLabel.rac.text = viewModel.outputs.projectCountText
     self.pushNotificationsButton.rac.selected = viewModel.outputs.pushNotificationsEnabled
     self.pushNotificationsButton.rac.hidden = viewModel.outputs.pushNotificationButtonIsHidden
-    self.projectCountLabel.rac.accessibilityHint = viewModel.outputs
```
dusi: I wasn't able to make VoiceOver read the hint (maybe this needs more testing on an actual device with different OS version). @Scollaco could you please run AppStore version on iOS 10/11 and see if this hint is being spoken?

If not we should also be able to remove `profile_project_count_projects_backed` from the `Strings` file

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -90,13 +94,12 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
   override func bindViewModel() {
     super.bindViewModel()
 
+    self.rac.accessibilityValue = viewModel.outputs.projectCountText
```
Scollaco: After testing out the code, I noticed that tapping any cell on `Settings > Notifications` - let's say Project Updates -will make the VoiceOver say `Project Updates, X`, where X is the project count. I may be happening because the `accessibilityValue` is being set to `viewModel.outputs.projectCountText` in every cell here.  I think that this can be solved but returning `projectCountText` only if the cellType = .projectNotifications

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
Scollaco: I was able to make the VoiceOver read the hint once I changed `self.projectCountLabel.rac.accessibilityHint` to `self.rac.accessibilityHint`, because than the hint would be coupled to the cell and not to the label.

Library/ViewModels/SettingsNotificationCellViewModel.swift
```diff
@@ -14,7 +14,6 @@ public protocol SettingsNotificationCellViewModelOutputs {
   var emailNotificationsEnabled: Signal<Bool, NoError> { get }
   var emailNotificationButtonIsHidden: Signal<Bool, NoError> { get }
   var pushNotificationButtonIsHidden: Signal<Bool, NoError> { get }
-  var manageProjectNotificationsButtonAccessibilityHint: Signal<String, NoError> { get }
```
Scollaco: We would need to re-add this signal after making the suggested change above.

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
dusi: Good eye! Thanks for catching this!

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
dusi: Interesting.

Not really sure why it won't use hint on a `UILabel`. But anyways in order to follow the native Settings.app patterns and the a11y audit, I think we should remove this.

@ifbarrera since you've added this localized hint, do you feel strongly about it?

ifbarrera: @dusi nope don't feel strongly about it, you can go ahead and remove it for consistency with the rest of Settings accessibility work!

dusi:
[♿️ Settings > Notifications - Manage project notifications](https://trello.com/c/d7SIFUzt/1162-%E2%99%BF%EF%B8%8F-settings-notifications-manage-project-notifications)

PR #553, opened 2019-01-11
Match email links based on host rather than paths

justinswart:
# 📲 What

This ignores the paths of email links for deep-linking and instead bases it on the host portion of these URLs.

# 🤔 Why

We found that some of our newsletter email links weren't deep-linking correctly. Upon investigating, this appeared to be because they have a different format to the links we've expected in the past, for example:

### Currently recognized links
`https://click.e.kickstarter.com/?qs=deadbeef`
`https://email.kickstarter.com/mpss/c/2gA/Oiw/t.25j/deadbeef/h1/dead-beef`
### New (unrecognized) recognized link
`https://click.em.kickstarter.com/wf/click?upn=deadbeef`

Our parser handles the detection of these URLs as email links by way of parsing their paths and query params. We don't actually make use of any of these params other than to recognize the URLs and to then attempt to resolve redirects for them which we then again test against the route parser and deep-link if we find a supported route.

# 🛠 How

The solution I'm proposing is that we instead base the recognition on the `HOST` portion of the URL. So if a URL matches any of the _known_ email hosts that we support, we return some empty params so that it matches on the `/` route and returns `Navigation.emailClick` which we then handle as before and find a redirect URL for.

Chatting to @pcreux about this he mentioned that we don't have any control over the construction of these URLs so we might be playing a game of whack-a-mole if we keep adding different paths in order to parse these URLs. Instead we can better-predict that host portions of these URLs and match on that, ignoring the paths and query strings.

# 👀 See

https://trello.com/c/Y033E5v2

# ✅ Acceptance criteria

- [ ] Newsletter email links deep-link correctly
- [ ] Push notifications deep-link correctly
- [ ] Project preview links are ignored correctly (as per #538)

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -324,10 +324,8 @@ AppDelegateViewModelOutputs {
 
     self.findRedirectUrl = deepLinkUrl
       .filter {
-        switch Navigation.match($0) {
-        case .some(.emailClick), .some(.emailLink): return true
-        default:                                    return false
-        }
+        guard case .some(.emailClick) = Navigation.match($0) else { return true }
+        return false
     }
 
     self.goToDiscovery = deepLink
--- a/Library/Navigation.swift
+++ b/Library/Navigation.swift
@@ -594,9 +594,9 @@ private func parsedParams(url: URL, fromTemplate template: String) -> RouteParam
 
   // if we're parsing against the '/' emailClick template and this is a recognized email host
   // return the expected params for that route to be resolved
-//  if templateComponents.isEmpty && isRecognizedEmailHost {
-//    return .object(["qs": .string("deadbeef")])
-//  }
+  if templateComponents.isEmpty && isRecognizedEmailHost {
+    return .object(["qs": .string("deadbeef")])
+  }
 
   guard templateComponents.count == urlComponents.count else { return nil }
 
--- a/Library/Navigation.swift
+++ b/Library/Navigation.swift
@@ -8,7 +8,7 @@ public enum Navigation {
   case checkout(Int, Navigation.Checkout)
   case creatorMessages(Param, messageThreadId: Int)
   case projectActivity(Param)
-  case emailClick(qs: String)
+  case emailClick
   case messages(messageThreadId: Int)
   case signup
   case tab(Tab)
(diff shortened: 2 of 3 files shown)


justinswart:
[Bug: System Email links deep-link to app while Newsletter links go to mobile web](https://trello.com/c/Y033E5v2/1053-bug-system-email-links-deep-link-to-app-while-newsletter-links-go-to-mobile-web)

Library/Navigation.swift
```diff
@@ -8,8 +8,7 @@ public enum Navigation {
   case checkout(Int, Navigation.Checkout)
   case creatorMessages(Param, messageThreadId: Int)
   case projectActivity(Param)
-  case emailClick(qs: String)
```
justinswart: Removed the associated string value from this case as we don't use it anywhere.

Library/Navigation.swift
```diff
@@ -569,24 +559,29 @@ private func userSurvey(_ params: RouteParams) -> Decoded<Navigation> {
 // MARK: Helpers
 
 private func parsedParams(url: URL, fromTemplate template: String) -> RouteParams? {
-
-  // early out on URL's that are not recognized as kickstarter URL's
-
-  let recognizedHosts = [
-    AppEnvironment.current.apiService.serverConfig.apiBaseUrl.host,
-    AppEnvironment.current.apiService.serverConfig.webBaseUrl.host,
+  let recognizedEmailHosts = [
```
justinswart: Split out `recognizedHosts` into that and `recognizedEmailHosts`.

Library/Navigation.swift
```diff
@@ -596,6 +591,12 @@ private func parsedParams(url: URL, fromTemplate template: String) -> RouteParam
     .components(separatedBy: "/")
     .filter { $0 != "" && !$0.hasPrefix("?") }
 
+  // if we're parsing against the '/' emailClick template and this is a recognized email host
+  // return the expected params for that route to be resolved
+  if templateComponents.isEmpty && isRecognizedEmailHost {
```
justinswart: This is the key logic, once we know we're on the empty `/` route template, and this is an email link we can just return an empty `RouteParams.object([:])` so that it's seen as a match and our parser will resolve it against the `emailClick(_ params: RouteParams)` function so that it returns `Navigation.emailClick`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -324,10 +324,8 @@ AppDelegateViewModelOutputs {
 
     self.findRedirectUrl = deepLinkUrl
       .filter {
-        switch Navigation.match($0) {
-        case .some(.emailClick), .some(.emailLink): return true
-        default:                                    return false
-        }
+        guard case .some(.emailClick) = Navigation.match($0) else { return false }
```
ifbarrera: Would `return Navigation.match($0) == Navigation.emailClick` work here?

justinswart: ah yes, well-spotted!

PR #554, opened 2019-01-14
Update "Find Facebook Friends" directive copy

ifbarrera:
# 📲 What

- Updates the copy that tells the user that they must turn `Following` on in order to find facebook friends. This needs to be updated because we changed the location of the `Following` switch: it used to be in `Settings > Privacy` but is now in `Settings > Account > Privacy`.
- fixes a bug that was preventing the `Base` localization (English) from being updated with the contents from the server... tldr: we really need to move this code to SPM so we can refactor & test it
- fixes a visual bug in `Strings.swift` where it was showing unsupported translations in the comments

# 🤔 Why

Clearer instructions for users on how to access the `Find Facebook friends` feature.

# 👀 See

simulator screen shot - iphone 8 - 2019-01-14 at 12 24 18

# ✅ Acceptance criteria

- [x] Go to `Settings > Account > Privacy` and turn `Following` off. Then go back to `Settings`, and below `Find Facebook friends` you should see the correct copy: "Following must be on to find Facebook friends. Change your account privacy settings to enable this feature."


ifbarrera:
[Directive copy for turning on Following is wrong, and uses the wrong pattern](https://trello.com/c/Vu4FSmJJ/1131-directive-copy-for-turning-on-following-is-wrong-and-uses-the-wrong-pattern)

bin/strings.swift
```diff
@@ -120,7 +120,7 @@ let stringsByLocale = stringsByLocale1
     }
 }
 
-let supportedLocales = ["de", "en", "es", "fr", "ja"]
+let supportedLocales = ["Base", "de", "en", "es", "fr", "ja"]
```
ifbarrera: Note the two changes to this file:
- `en` gets mapped to `Base` in line 115, which is why I noticed that the script wasn't writing changes to the english `.strings` file. This is not a great solution, but honestly I don't think we should attempt any more improvements to this file until we can move it to SPM and write some tests for it
- second change only appends the locale comment if the locale is in the list of supported locales (removed `**it**:` etc from the `Strings.swift` file)

justinswart: Ah interesting - so this would have a regression introduced in #528?

ifbarrera: Yep exactly

PR #555, opened 2019-01-14
Adds a11y section to the PR template

dusi:
# 📲 What

See title..

# 🤔 Why

Reminds engineers to be conscious about accessibility support when building user interface.

# 🛠 How

Simply adds additional ♿️ checklist to our PR template.

# ✅ Acceptance criteria

- [x] CI passes 💚


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,7 +6,7 @@ A description of the change.
 
 # 🤔 Why
 
-Why it's needed, background context.
```
dusi: This always sounded weird to me, is it even correct?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,6 +16,12 @@ More in-depth discussion of the change or implementation.
 
 Trello, screenshots, external resources?
 
+# ♿️ Accessibility 
+
+- [ ] Tap targets use minimum of 44x44 pts dimensions
```
dusi: This is a start and we can expand on these later.

Scollaco:
Nice, @dusi ! I just have one question: If i'm working in some feature that already exists but doesn't support correctly Dynamic Type (due the way some cells were created), the refactor should be done on the same PR to meet the AC or another PR should be created, since a refactor can be necessary, making the PR bigger than expected?

dusi:
Good question @Scollaco. For now this should only be considered for new work and more importantly for UI/UX work because if you're fixing let's say network calls, etc. we don't need to worry about accessibility.

Aso for the more important question - I would assume that Dynamic Type is broken right now thorough the app and until we run an audit we simply ignore this point (same as if your PR doesn't have any TODOs - just remove the section).

After we've done audit & possibly fixed/updated the app to support Dynamic Type...all UI/UX work should take this point into consideration.

Does that make sense?

Cc: @justinswart @ifbarrera @cdolm92

PR #562, opened 2019-01-24
Ensure gym output path is consistent

justinswart:
# 📲 What

This ensures that the path that `fastlane gym` outputs to is always the same as the path that we try to read from.

# 🤔 Why

When we run `fastlane gym` on CircleCI, we have an environment variable called `FL_OUTPUT_DIR` which is set to `output`. This causes any output that `fastlane` actions generate to be put under that path. So for example, `fastlane gym` will output to `./output/gym`. However, when we run this locally that environment variable is not set and so, because I initially tested this locally, I set it to look for the `ipa` file at `.` and not `./output/gym`. This has happened a couple of times before so I decided to try and prevent it going forward.
# 🛠 How

If we look at our `setup_circle_ci` command in our `Fastfile` it [constructs an environment var](https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/setup_circle_ci.rb#L22) called `GYM_OUTPUT_DIRECTORY`. I've placed a line at the top of our `Fastfile` which uses the value of this env var if it's present and if not coalesces to the value `.`. We then use the `gym_dir` value in the places where we need to reference the `gym` output dir.

# ✅ Acceptance criteria

- [x] Running `fastlane beta_gym` both locally and on CircleCI should succeed.
- [x] Manipulating the `GYM_OUTPUT_DIRECTORY` locally and running `fastlane beta_gym` should show up in the console for that command (as in screenshot below).

| `GYM_OUTPUT_DIRECTORY` value set to `output`  | `GYM_OUTPUT_DIRECTORY` not set |
| screen shot 2019-01-23 at 5 14 20 pm | screen shot 2019-01-23 at 5 14 29 pm |

### DIFF
--- a/.fastlane/Fastfile
+++ b/.fastlane/Fastfile
@@ -101,8 +101,8 @@ platform :ios do
     sh "./beta-changelog.rb"
 
     upload_s3(
-      ipa: "KickBeta.ipa",
-      dsym: "KickBeta.app.dSYM.zip"
+      ipa: "./output/gym/KickBeta.ipa",
+      dsym: "./output/gym/KickBeta.app.dSYM.zip"
     )
 
     sh "./beta-post-s3.rb"
--- a/.fastlane/Fastfile
+++ b/.fastlane/Fastfile
@@ -3,8 +3,6 @@ fastlane_version "2.28.3"
 default_platform :ios
 
 platform :ios do
-  gym_dir = ENV["GYM_OUTPUT_DIRECTORY"] || "."
-
   ### MATCH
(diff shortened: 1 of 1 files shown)


justinswart:
[Ensure gym output path is consistent](https://trello.com/c/N86T8ftF/1180-ensure-gym-output-path-is-consistent)

justinswart:
Cool thanks @ifbarrera, turns out the issue before was that I was setting `gym_dir` globally before `setup_circle_ci` ran which is the command that sets the environment var that we care about.

PR #563, opened 2019-01-25
Colors tests

Scollaco:
# 📲 What
Adds tests for our Colors script.

# 🤔 Why
The current script has no test and was this was an opportunity to do some refactor and make it more testable.

# 🛠 How
In order to do this, I used [Swift Package Manager](https://swift.org/package-manager/) and generated an Xcode proj to make it easier to run tests . Also, the colors script has been split into `ColorScriptCore` (Library + tests) and `ColorScript` (Executable).

To use the script we just need to  `$ cd bin/ColorScript` folder and run `$ .build/release/ColorScript`.

# ✅ Acceptance criteria

- [ ] All tests should pass
- [ ] By running `.build/release/ColorScript`, a list of all colors should appear on the terminal followed by the message `✨ Done regenerating Colors.swift ✨`

GIF


bin/ColorScript/Sources/ColorScript/main.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+import ColorScriptCore
+
+let colorScript = ColorScript()
+
+do {
+  try colorScript.run()
+} catch {
+  print("Error")
```
ifbarrera: Can we add another case to the `ColorScriptError` enum to capture this error and return a more descriptive error message?

bin/ColorScript/README.md
```diff
@@ -0,0 +1,3 @@
+# ColorScriptExecutable
+
+A description of this package.
```
ifbarrera: Can we add the usage instructions and maybe some details about how to update and re-build this script in here?

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScript.swift
```diff
@@ -0,0 +1,34 @@
...
+
+public enum ColorScriptError: Error {
+  case missingArguments
+}
+
+public final class ColorScript {
+  private let arguments: [String]
+
+  public let inPath = "Resources/Colors.json"
+  public let outPath = "Resources/Colors.swift"
```
ifbarrera: Hm we should probably keep the output path the same as where the `Colors.swift` file currently lives. I believe it's at `/Library/Styles/Colors.swift`.

Design/Colors.swift
```diff
@@ -0,0 +1,230 @@
+//===============================================================
```
ifbarrera: Is there a reason we need the `Colors.swift` file here too? I noticed it's also at `bin/ColorScript/Resources/Colors.swift`

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScript.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    print(inPath)
+    print(outPath)
+
+    let data = try! Data(contentsOf: URL(fileURLWithPath: inPath))
+    let color = Color(data: data)
+
+    print("All colors: \n\(color.prettyColors)")
+
+    try! color.staticStringsLines()
```
ifbarrera: Can we wrap this in a `do catch` and return a descriptive error message if this fails?

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScript.swift
```diff
@@ -0,0 +1,34 @@
...
+  public init(arguments: [String] = CommandLine.arguments) {
+    self.arguments = arguments
+  }
+
+  public func run() throws {
+
+    print(inPath)
+    print(outPath)
+
+    let data = try! Data(contentsOf: URL(fileURLWithPath: inPath))
```
ifbarrera: I think we should avoid using `try!` since we now have a nice way to fail gracefully. We could wrap this in a `do catch` and throw a `ColorScriptError` with a descriptive error message

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
```diff
@@ -0,0 +1,118 @@
...
+
+  public func colors() -> [(key: String, value: String)]? {
+    do {
+      if let json = try (JSONSerialization.jsonObject(with: data, options: []) as? [String: String]) {
+        let colors = json
+          .map { (key: $0, value: $1) }
+          .sorted { $0.key < $1.key }
+        return colors
+      }
+      return nil
```
ifbarrera: I think it might be more useful for us to throw an error here instead of returning `nil`, which could make it harder to figure out what happened that caused the script to crash further down (when we unwrap the output `colors()!`)

bin/ColorScript/Sources/ColorScript/main.swift
Scollaco: I think here we could capture the error from the catch block and print the error message, like:
```
catch (let error) {
  print("Error: \(error.localizedDescription)")
}
```
What do you think?

Design/Colors.swift
Scollaco: No. I forgot to delete it. Good catch :)

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScript.swift
Scollaco: It makes sense. I thought of `Resources/` as a replacement for the `Design/` folder (used by the old script). But I just noticed that we have 2 identical `Colors.swift`. One at `Design/Colors.swift` and the other at `Styles/Colors.swift`. I'm wondering what's the motivation behind it.

bin/ColorScript/.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
```
justinswart: Does this need its own `.gitignore`?

bin/ColorScript/Package.swift
```diff
@@ -0,0 +1,16 @@
+// swift-tools-version:4.2
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
```
justinswart: This file isn't using 2-space indentation.

bin/ColorScript/Sources/ColorScriptCore/.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
```
justinswart: This file also required?

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
```diff
@@ -0,0 +1,118 @@
+// swiftlint:disable force_unwrapping
+import Foundation
+
+extension Dictionary {
+  public func withAllValuesFrom(_ other: Dictionary) -> Dictionary {
```
justinswart: Swift now has a built-in function for this, called `merging(_:uniquingKeysWith:)` you can reproduce the same behaviour as the function in this extension by doing:
```swift
public func withAllValuesFrom(_ otherDictionary: Dictionary) -> Dictionary {
  return self.merging(otherDictionary) { $1 }
}
```
If it's used in more than one place, I'd do it this way, otherwise just use that directly at the callsite.

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
```diff
@@ -0,0 +1,118 @@
...
+        return colors
+      }
+      return nil
+    } catch {
+      return nil
+    }
+  }
+
+  public var prettyColors: String {
+    return colors()!.map { (color, value) in
```
justinswart: Might want to handle the optionality of `colors()` here and also refer to `self`.

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
```diff
@@ -0,0 +1,118 @@
...
+      return "  \(color): #\(value)"
+      }.joined(separator: "\n")
+  }
+
+  public var allColors: [(key: String, value: [(key: Int, value: String)])] {
+    return colors()!
```
justinswart: And here, the force-unwrap.

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
```diff
@@ -0,0 +1,118 @@
...
+      .reduce([String: [Int: String]]()) { accum, pair in
+        let (name, _) = pair
+
+        let components = name.components(separatedBy: "_")
+        guard components.count > 1 else { return accum }
+        let colorWeight: Int? = Int(components.last!)
```
justinswart: Force-unwrap here.

bin/ColorScript/Tests/ColorScriptTests/ColorScriptTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+  func testColorNameFormat() {
+
+    let data = json.data(using: .utf8)!
+    let colors = Color(data: data)
+    colors.allColors.forEach { (arg: (key: String, value: [(key: Int, value: String)])) in
+
+      let (key, value) = arg
+      let colorName = key.lowercased()
+      let weight = value.first!.key
```
justinswart: Preferably want to `guard` against `first` being `nil`.

bin/ColorScript/Tests/ColorScriptTests/XCTestManifests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+
+#if !os(macOS)
+public func allTests() -> [XCTestCaseEntry] {
+    return [
```
justinswart: 2-space indentation.

bin/ColorScript/Tests/ColorScriptTests/ColorScriptTests.swift
```diff
@@ -0,0 +1,99 @@
+// swiftlint:disable force_unwrapping
+import Foundation
+import XCTest
+import ColorScriptCore
+import ColorScript
+
+final class ColorScriptTests: XCTestCase {
```
justinswart: Indentation in this file is 4 spaces.

bin/ColorScript/.gitignore
Scollaco: SPM created it when I ran `git init` (following a tutorial), but since we're not creating a separated repo for this anymore, I will remove it.

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScriptCore.swift
Scollaco: Good to know!

bin/ColorScript/Sources/ColorScript/main.swift
ifbarrera: We could also have `ColorScriptError` conform to `CustomStringConvertible` so we can provide our own description of various error types. In that case all you would have to do is:

```
print("Error: \(error)")
```

Scollaco:
> Great job with SPM! Mostly stylistic suggestions so far. I would also recommend getting this under `swiftlint` to pick up the force-unwraps and maybe some other things.

Thanks for the feedback, @justinswart . Swiftlint didn't allow me to push changes when lint violations occurred, so I suppose it's working fine?

justinswart:
Hey my only other suggestion would be to search for `!` and try to replace it with `guards` or handle the optionality as far as possible. And maybe also dig into why `swiftlint` is not linting this?

bin/ColorScript/Tests/ColorScriptTests/ColorScriptTests.swift
```diff
@@ -0,0 +1,99 @@
+// swiftlint:disable force_unwrapping
```
justinswart: Oh, 😂this might be why it's ignoring the force unwraps

bin/ColorScript/Sources/ColorScriptCore/Sources/ColorScript.swift
```diff
@@ -0,0 +1,42 @@
+// swiftlint:disable force_unwrapping
```
justinswart: And here.

justinswart:
Just a reminder to remove the swiftlint disable annotations so that we can have it run on this, then I'll re-review 👍

justinswart:
As per slack convo:

- Ideally let's remove force unwraps from tests too, we'd prefer them to fail than crash.
- Let's remove the old `bin/colors.swift`.
- Let's figure out why `.build/release/ColorScript` doesn't work for me locally but `./bin/ColorScript/colorscript` does.

PR #564, opened 2019-01-29
Reusable download framework script

justinswart:
# 📲 What

Improves our `Makefile` a wee bit by moving duplicated logic for downloading our Stripe and Fabric SDKs into a shared bash script.

# 🤔 Why

We like to keep things DRY and this wasn't being very DRY at all. In the event that we download more frameworks this way during `make bootstrap` this will hopefully make that cleaner.

# 🛠 How

Extracted the logic that we were using into a shared bash script that receives arguments.

Note that I didn't do this for OpenTok, reasons being that we're likely to remove it soon and the additional complexity of it being a bzip'd tar ball made for a pretty gnarly script to handle that.

# ✅ Acceptance criteria

- [x] Scripts download as before
- [ ] Everything works on CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,11 +44,11 @@ jobs:
 
       - restore_cache:
           keys:
-            - opentok-cache-{{ checksum "opentok_version.txt" }}
+            - opentok-sdk-cache-{{ checksum "opentok_version.txt" }}
 
       - restore_cache:
           keys:
-            - fabric-cache-{{ checksum "fabric_version.txt" }}
+            - fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
 
       - run:
           name: Make bootstrap
@@ -60,12 +60,12 @@ jobs:
             - ".git"
 
       - save_cache:
-          key: opentok-cache-{{ checksum "opentok_version.txt" }}
+          key: opentok-sdk-cache-{{ checksum "opentok_version.txt" }}
           paths:
             - "Frameworks/OpenTok"
 
       - save_cache:
-          key: fabric-cache-{{ checksum "fabric_version.txt" }}
+          key: fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
           paths:
             - "Frameworks/Fabric"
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -161,34 +161,9 @@ opentok:
...
-		unzip fabric.zip -d Frameworks/Fabric || true; \
-		rm fabric.zip; \
-	fi
-	@if [ -e Frameworks/Fabric/Fabric.framework ]; then \
-		echo "Fabric v$(FABRIC_SDK_VERSION) downloaded"; \
-	else \
-		echo "Failed to download Fabric SDK"; \
-		rm -rf Frameworks/Fabric; \
-	fi
+	bin/download_framework.sh Fabric $(FABRIC_SDK_VERSION) https://s3.amazonaws.com/kits-crashlytics-com/ios/com.twitter.crashlytics.ios/INSERT_SDK_VERSION/com.crashlytics.ios-manual.zip; \
```
ifbarrera: Is it possible to store the download url in an environment variable define at the top of this file? 🤔

justinswart: > One thing I noticed is that the Fabric cache on Circle seems to no longer be working. ie. although the `Frameworks/Fabric` path is restored, the script still re-downloads Fabric.
> 
> It might also speed up our builds if we start caching the `Frameworks/Stripe` path, I must have missed adding this when we upgraded the Stripe SDK 😓

Cool will add that and look at moving the URL to an env var at the top 👍

.circleci/config.yml
```diff
@@ -96,13 +112,22 @@ jobs:
...
           keys:
-            - opentok-cache-{{ checksum "opentok_version.txt" }}
+            - opentok-sdk-cache-{{ checksum "opentok_version.txt" }}
 
       - restore_cache:
           keys:
-            - fabric-cache-{{ checksum "fabric_version.txt" }}
+            - fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
+
+      - save_cache:
```
ifbarrera: Should this be `restore_cache`?

justinswart: Aah, yip thanks!

PR #566, opened 2019-02-01
Zipcode field in Add New Card screen

ifbarrera:
# 📲 What

Adds a required zipcode/postal code field to the "Add new card" flow.

# 🛠 How

- created a new generic `SettingsFormFieldView` which contains a label and a text field. This can be used on storyboards and configured as needed, but will have the same consistent styling
- added a `SettingsFormFieldView` instance to the add new card `stackView`, configured `titleLabel` text

Additional work:
- scroll view now increases the `contentInset` when the keyboard displays
- form now submits and keyboard dismisses when you hit "Save" and when you hit "Return" on the last field (zipcode field)
- updated form constraints to `layoutMargins` and set margins on the stack view

Known bugs:
- (iOS 10): zipcode is incorrectly focused after completing the credit card number field. This seems to be a bug with how the Stripe view is transitioning from one internal text field to another before calling paymentCardFieldDidFinishEditing()

Tested on:
- Sade (iOS 10.3.1)
    - Portrait mode, Landscape mode
- Elvis (iOS 11.2.2)
    - Portrait mode, Landscape mode
- iPhone X (iOS 12.1.2)
    - Portrait, Landscape
- Simulator iPhone 5 iOS 10.3.1
- Simulator iPad Pro (11-inch) iOS 12.1

# 👀 See

iakb9hfn5z

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Navigate to "Add card screen", enter name, credit card details, zipcode, user should be able to submit form
- [x] "Save" button should **not** be enabled if no zipcode is entered


Kickstarter-iOS/Views/Storyboards/LoadingBarButtonItemView.xib
```diff
@@ -20,7 +20,7 @@
                     <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
```
ifbarrera: Repositioned the activity indicator to aligned to the trailing edge of the `UIButton`, and right aligned the `titleLabel` within the `UIButton`

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -23,24 +23,54 @@ internal final class AddNewCardViewModelTests: TestCase {
   private let paymentDetailsBecomeFirstResponder = TestObserver<Void, NoError>()
   private let saveButtonIsEnabled = TestObserver<Bool, NoError>()
   private let setStripePublishableKey = TestObserver<String, NoError>()
+  private let zipcode = TestObserver<String, NoError>()
+  private let zipcodeTextFieldBecomeFirstResponder = TestObserver<Void, NoError>()
 
   override func setUp() {
     super.setUp()
-    self.vm.outputs.activityIndicatorShouldShow.observe(activityIndicatorShouldShow.observer)
```
ifbarrera: Added self to all the test observer references.

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -738,7 +738,7 @@
                                 <rect key="frame" x="0.0" y="20" width="375" height="647"/>
                                 <subviews>
                                     <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" translatesAutoresizingMaskIntoConstraints="NO" id="lnb-P0-tav">
-                                        <rect key="frame" x="0.0" y="0.0" width="375" height="176"/>
```
ifbarrera: - Added the `SettingsFormFieldView` to the main stack view
- changed all left & right constraints within stack view elements to be relative to the `layoutMargins`, then set the `layoutMargins` of the stack view
- updated the height of all the separators to be 0.5pts (checked this value with Danny)

ifbarrera:
[Zip code for iOS payment methods](https://trello.com/c/FVCzX3eL/1181-zip-code-for-ios-payment-methods)

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -196,6 +226,15 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
     self.viewModel.inputs.cardholderNameTextFieldReturn()
```
justinswart: The parenthesis on the line just above this seems to have shifted weirdly.

Kickstarter-iOS/Views/SettingsFormFieldView.swift
```diff
@@ -0,0 +1,42 @@
...
+  @IBOutlet weak var textField: UITextField!
+  @IBOutlet weak var titleLabel: UILabel!
+  @IBOutlet fileprivate weak var separatorView: UIView!
+  @IBOutlet fileprivate weak var stackView: UIStackView!
+
+  required init?(coder aDecoder: NSCoder) {
+    super.init(coder: aDecoder)
+
+    guard let view = self.view(fromNib: .SettingsFormFieldView) else {
+        fatalError("Failed to load view")
```
justinswart: `swiftlint` missed this 4-space indent 🤔

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -214,15 +253,19 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
...
   }
 
+  internal func paymentCardTextFieldDidEndEditing(_ textField: STPPaymentCardTextField) {
+    self.viewModel.inputs.paymentCardTextFieldDidEndEditing()
+  }
+
   // MARK: - Private Functions
-  private func createStripeToken(cardholderName: String, cardNumber: String, expirationMonth: Month,
-                                 expirationYear: Year, cvc: String) {
+  private func createStripeToken(paymentDetails: PaymentDetails) {
```
justinswart: can I suggest:

```swift
private func createStripeToken(with paymentDetails: PaymentDetails) { ... }
```
So that the callsite reads:

```swift
self?.createStripeToken(with: paymentDetails)
```

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -23,24 +23,54 @@ internal final class AddNewCardViewModelTests: TestCase {
...
+    self.vm.outputs.zipcodeTextFieldBecomeFirstResponder
+      .observe(self.zipcodeTextFieldBecomeFirstResponder.observer)
+  }
+
+  func testZipcodeTextFieldReturn_submitsPaymentDetails() {
+    self.vm.inputs.viewDidLoad()
+    self.saveButtonIsEnabled.assertDidNotEmitValue()
+
+    self.vm.inputs.cardholderNameChanged("Native Squad")
+    self.vm.inputs.creditCardChanged(cardDetails: ("4242 4242 4242 4242", 11, 99, "123"))
```
justinswart: Do we need to test a scenario where the card has expired?

justinswart:
Could we force capitals in the zip code field? This is useful when typing a Canadian postal code like `V6B 1M1`

justinswart:
Could we also dismiss the keyboard on tapping `Save`? Currently it dismissing on `return` on the zip code field.

Scollaco:
Well done, @ifbarrera! I just left a few suggestions.

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -58,6 +62,12 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
     let navigationBarButton = UIBarButtonItem(customView: self.saveButtonView)
     self.navigationItem.setRightBarButton(navigationBarButton, animated: false)
 
+    self.zipcodeView.textField.addTarget(self, action: #selector(zipcodeTextFieldDoneEditing),
```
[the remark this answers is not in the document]
Scollaco: I was just thinking that would be a better practice to avoid accessing IBOutlets from outside the class, and treat `SettingsFormFieldView` as a TextField representation, instead of an UIView container. Since the zipcodeTextField represents a textField, this approach would let the class implement the `UITextFieldDelegate` messages itself, keeping it in its own context. The tradeoff is the implementation of the `SettingsFormFieldViewDelegate`, in other hand, though, we could get rid of the `addTarget` methods. But this was just a suggestion and a different idea to solve the same problem.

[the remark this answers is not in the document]
ifbarrera: I hear you, and if this view is going to be reused in a bunch of places it would be nice to get rid of the `addTarget` methods. I'd say if we end up using this view again we can re-think where the`UITextFieldDelegate` methods should live.

PR #568, opened 2019-02-01
① [Currency selection] Remove currency picker

justinswart:
# 📲 What

Paired with @Scollaco to remove the currency picker from `SettingsAccountViewController`.

# 🤔 Why

We've decided to replace the picker behaviour with a table view that will be pushed onto the navigation stack.

# 🛠 How

Deleted all references to the picker.

# ✅ Acceptance criteria

- [x] Project builds successfully
- [x] Tests pass
- [x] Currency picker is gone

#569


justinswart:

dusi:
Looks like `case SettingsCurrencyPickerCell` in `Nib.swift` can be removed as well..

dusi:
Also from visual perspective there are 2 questions that come into my mind:

1) Should we add disclosure indicator to the change currency cell to imply that this will push user to a different screen?

2) Should we change the text color to something more subtle?
Using green color makes the text look like a button, but since we want the whole cell to react to taps we don't explicitly need to make this look like a button (the disclosure indicator from first point should make that clear). Cc @dannyalright 

screen shot 2019-02-04 at 11 15 52 am

justinswart:
@dannyalright sure - just to be clear we add a chevron disclosure icon in a later PR, but are you saying the text should not be green?

dannyalright:
that's right. It should be the default cell 'value' colour. [Here's the exact screen](https://share.goabstract.com/25031582-9081-48f0-9e13-3c44f36ab82d) just to be safe. All good RE chevron in later PR

PR #569, opened 2019-02-01
② [Currency selection] Add table view

justinswart:
# 📲 What

Paired with @Scollaco to add the `SelectCurrencyViewController` which is a table view that we push on to our navigation stack in order to select a preferred currency.

# 🤔 Why

This is part of a larger body of work to replace the current picker behaviour with a table view for selection.

# 🛠 How

- Added `UITableView` programmatically (no xib) and used the `Currency` `enum` as a very simple datasource.
  - This is not our convention but we felt that a separate datasource class for this view was overkill and thought we would try this approach. We're very much up for discussion on whether or not this is okay to do!
- Added a selection indicator also using basic `UITableViewCell`s as this sufficient for what we needed (the checkmark is just using the cell's `accessoryType` property).

# ♿️ Accessibility 

- We will address this in a separate PR.

# ✅ Acceptance criteria

- [ ] When tapping the currency cell on the account screen you should be taken to this table view and see your current preferred currency selected.
- [ ] Can select different currencies and the selection indicator moves to those cells.

#568


justinswart:

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
```diff
@@ -3,17 +3,19 @@ import Library
 import Prelude
 
 final class SettingsCurrencyCell: UITableViewCell, NibLoading, ValueCell {
-  @IBOutlet fileprivate weak var titleLabel: UILabel!
+  @IBOutlet fileprivate weak var chevronImageView: UIImageView!
```
dusi: image

:trollface:

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+      |> \.separatorStyle .~ .singleLine
+  }
+
+  // MARK: Subviews
+
+  private lazy var tableView: UITableView = {
+    let tableView = UITableView(frame: .zero, style: .plain)
+
+    tableView.translatesAutoresizingMaskIntoConstraints = false
+    tableView.register(UITableViewCell.self, forCellReuseIdentifier: "cell")
```
ifbarrera: Can we use a more descriptive identifier? Maybe something like `CurrencySelectionCell`

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
```diff
@@ -22,11 +24,17 @@ final class SettingsCurrencyCell: UITableViewCell, NibLoading, ValueCell {
     _ = self.titleLabel
       |> settingsTitleLabelStyle
 
+    _ = self.chevronImageView
```
dusi: Wonder if we should use the same right padding for both the `..WarningCell` and this cell?

screen shot 2019-02-04 at 12 07 31 pm

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
```diff
@@ -3,17 +3,19 @@ import Library
...
+  @IBOutlet fileprivate weak var titleLabel: UILabel!
 
   private let viewModel: SettingsCurrencyCellViewModelType = SettingsCurrencyCellViewModel()
 
   func configureWith(value: SettingsCurrencyCellValue) {
     self.viewModel.inputs.configure(with: value)
 
     _ = titleLabel
-      |> UILabel.lens.text .~ value.cellType.title
+      |> \.text .~ value.cellType.title
```
dusi: Wanna set the `backgroundColor` to `.white` to prevent rendering as blended layer?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
+import Library
+import Prelude
+import Prelude_UIKit
+import UIKit
+
+final class SelectCurrencyViewController: UIViewController, MessageBannerViewControllerPresenting {
```
dusi: Do we need to use `UIViewController` as oppose to `UITableViewController`? It would save us couple lines (23, 24 in `viewDidLoad`)...but I'm unsure if it will work properly with the message banner.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    self.viewModel.inputs.viewDidLoad()
+  }
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = self.tableView
+      |> settingsTableViewStyle
+      |> \.separatorStyle .~ .singleLine
```
dusi: Also, since we're introducing a native table view, we're getting the default separator look and feel (with left padding)...if we want to use edge to edge separators we can use the following (to stay consistent)

```
self.tableView.separatorInset = .zero
```

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // MARK: Subviews
```
dusi: `// MARK: - Text` in order to add separators to the jump bar pls 🙏

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+
+extension SelectCurrencyViewController: UITableViewDelegate {
+  func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+    let currency = Currency.allCases[indexPath.row]
+
+    self.viewModel.inputs.didSelect(currency)
+
+    tableView.deselectRow(at: indexPath, animated: true)
+    tableView.visibleCells.forEach { $0.accessoryType = .none }
+    tableView.cellForRow(at: indexPath)?.accessoryType = .checkmark
```
ifbarrera: Hmm is there any way to test that the visual treatment of this cell is actually changing if its not coupled to an `output` on the view model? 🤔 This treatment feels like business logic to me, which would ideally live in the view model.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    _ = self.tableView
+      |> settingsTableViewStyle
+      |> \.separatorStyle .~ .singleLine
+  }
+
+  // MARK: Subviews
+
+  private lazy var tableView: UITableView = {
+    let tableView = UITableView(frame: .zero, style: .plain)
```
dusi: I know this is super readable but wonder if we can skip the need for declaring local variable by using the pipe

i.e.
```
return UITableView()
  |> translatesAutoresizingMaskIntoConstraints .~ false
  |> ...
```

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
dusi: I think this is traditionally done in `viewDidLoad` but I'm kinda seeing the point here...not sure which one I like better 😄

dusi: How about making the `reuseIdentifier` type safe (by using an enum and more appropriate string `"SelectCurrencyCell"`?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    tableView.translatesAutoresizingMaskIntoConstraints = false
+    tableView.register(UITableViewCell.self, forCellReuseIdentifier: "cell")
+    tableView.tableFooterView = UIView(frame: .zero)
```
dusi: Wouldn't `UIView()` suffice?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+    tableView.dataSource = self
```
dusi: Again, wonder if this should be in `viewDidLoad` or not :/

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+extension SelectCurrencyViewController: UITableViewDelegate {
+  func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+    let currency = Currency.allCases[indexPath.row]
+
+    self.viewModel.inputs.didSelect(currency)
+
+    tableView.deselectRow(at: indexPath, animated: true)
```
dusi: Should deselecting be part of the `output` signal, so that the order is

1. `tableView:didSelectRowAt`
2. self.viewModel.inputs.didSelect
3. self.viewModel.outputs.didSelect (that we would subscribe to) - here we would do UI work (deselecting, updating checkmarks)

dusi: It means we would also have to pass in the cell or index at least

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
ifbarrera: One way to handle this would be to create an `output` for `selectedCurrency`:

```
self.viewModel.outputs.selectedCurrency
    .observeForUI()
    .observeValues { currency in
       // clear all accessory values && set the accessory on the correct cell here
    }
```

Then you can write a test with confidence that if the `selectedCurrency` output fires, then your cells are getting visually updated correctly.

Kickstarter-iOS/UIView+AutoLayout.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
+extension UIView {
+  public func constrainEdges(to view: UIView) {
```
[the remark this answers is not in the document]
justinswart: The function naming is correct 🙂 
Good idea on the constraints, will update 👍

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: Sure! I just went with a simple string seeing as there's only one super simple cell in this table, I can update.

Kickstarter-iOS/UIView+AutoLayout.swift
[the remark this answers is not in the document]
justinswart: Let's chat more about the constraints - one or two edge cases occurred to me. To be honest, ideally we'd solve something like this with a decent AutoLayout DSL.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: I think it's better to call the designated initializer and pass in an empty frame but sure we can look at piping it through keypaths.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: > I think this is traditionally done in `viewDidLoad` but I'm kinda seeing the point here...not sure which one I like better 😄

😄 yeah not sure, i tend to like how `lazy var` let's us group configuration together, but also could go either way.

I've moved it to a local constant - not sure an `enum` makes sense with only on case?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: I think the `UIView(frame:)` designated initializer is better.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: Ah yeah actually this is a good idea - and ties in with what @ifbarrera suggested earlier - will implement that way 👍

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: Going to run with @dusi's [suggestion](https://github.com/kickstarter/ios-oss/pull/569#discussion_r253633306), should be a good middle-ground.

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
dusi: Designated initializer sounds fine...I was just being lazy re-typing everything :) the point was the pipe

Kickstarter-iOS/UIView+AutoLayout.swift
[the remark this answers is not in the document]
justinswart: As per offline chat - we want the table view to extend all the way to the bottom because it's modally displayed. The article is referring specifically to the deprecation of `topLayoutGuide` and `bottomLayoutGuide` on `UIViewController` which we should replace with `safeAreaLayoutGuide` on iOS 11+ 👍

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
Scollaco: 🤦‍♂️

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: Due to the complexity of implementing this and fixing the tests later, I am going to do this refactor on #572 👍

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
justinswart: This cell has since been removed, let's do these changes as a sweep or as the merging of #573

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: I think we should figure this out as part of https://github.com/kickstarter/rfcs/pull/54

Kickstarter-iOS/Views/Cells/SettingsCurrencyCell.swift
```diff
@@ -3,17 +3,19 @@ import Library
 import Prelude
 
 final class SettingsCurrencyCell: UITableViewCell, NibLoading, ValueCell {
```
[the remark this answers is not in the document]
justinswart: Removed `SettingsCurrencyCell`

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: As per [this](https://github.com/kickstarter/ios-oss/pull/569#discussion_r254096862) comment, I'm going to refactor this in #572

dusi:
I've noticed that the screen doesn't have a title (in the navigation bar)...should we add it or has it been added in one of the follow up PRs?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -0,0 +1,79 @@
+import Library
+import Prelude
+import Prelude_UIKit
+import UIKit
+
+private let reuseIdentifier = "CurrencySelectionCell"
```
dusi: Not a suggestion for this PR but we should start extracting these to a singular place (my PRs are going the same direction :-)

justinswart: Yeah let's chat about it!

justinswart:
> I've noticed that the screen doesn't have a title (in the navigation bar)...should we add it or has it been added in one of the follow up PRs?

Ah good point, I'll find out what it should be and add it 👍

Library/SettingsAccountCellType.swift
```diff
@@ -48,8 +43,6 @@ public enum SettingsAccountCellType: SettingsCellTypeProtocol, CaseIterable {
 
   public var detailTextColor: UIColor {
     switch self {
-    case .currency:
```
ifbarrera: Any reason for removing this? I believe we still want the currency text to display in green instead of of the default grey color

justinswart: Not according to [this](https://github.com/kickstarter/ios-oss/pull/568#issuecomment-460772246) comment.

PR #570, opened 2019-02-01
③ [Currency selection] Add table view header

justinswart:
# 📲 What

Paired with @Scollaco to add a header to `SelectCurrencyViewController`.

# 🤔 Why

This is a continuation of the work in #568 and #569 and adds the tableview header as per the designs.

# 🛠 How

- We've added `SelectCurrencyTableViewHeader` programmatically as an experiment which we are open to discussion on.
  - One challenge this presents is that `UITableView` headers and footers don't really participate in auto layout so there is some additional work required to have this work correctly.
  - After any changes to the header such as the text setting that we do in `bindStyles()` it's necessary to set the header view again on the table view.

# ♿️ Accessibility 

- Will be addressed in separate PR.

# ✅ Acceptance criteria

- [ ] Header appears above the list of currencies.
- [ ] Font and text colours are accurate.

#568 
#569


justinswart:

Kickstarter-iOS/Assets.xcassets/icons/icon--currency-header.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "icon--currency-header.pdf"
```
dusi: Since I'm not familiar with our naming conventions, what is the reason for using 2 dashes? `icon--curr...`

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -24,6 +24,7 @@ final class SelectCurrencyViewController: UIViewController, MessageBannerViewCon
 
     self.tableView.register(UITableViewCell.self, forCellReuseIdentifier: reuseIdentifier)
 
+    self.tableView.setConstrained(headerView: self.headerView)
```
dusi: Reading this call makes me think we're making `headerView` fill the `tableView`...would it make sense to use a better function signature?

```
func setConstrainedWidth(for headerView: UIView) { }
func setConstrainedToWidth(headerView: UIView) { }
```

Kickstarter-iOS/Views/SelectCurrencyTableViewHeader.swift
```diff
@@ -0,0 +1,63 @@
+import Library
+import Prelude
+import Prelude_UIKit
+import UIKit
+
+final class SelectCurrencyTableViewHeader: UITableViewHeaderFooterView {
```
dusi: I've had a quick peek at the documentation and am starting to think we should have used `UIView` subclass instead. It looks like `UITableViewHeaderFooterView` is designed only for the `section` headers/footers that take advantage of the reuse of those views (just like cells). On the other hand `UITableView`'s `tableHeaderView`/`tableFooterView` is an accessory view that isn't reused.

https://developer.apple.com/documentation/uikit/uitableviewheaderfooterview

Kickstarter-iOS/Views/SelectCurrencyTableViewHeader.swift
```diff
@@ -0,0 +1,63 @@
...
+  public var text: String? {
+    didSet {
+      self.headerLabel.text = text
+    }
+  }
+
+  // MARK: Subviews
+
+  private lazy var headerStackView: UIStackView = {
+    let stackView = UIStackView(arrangedSubviews: [
```
dusi: Is there a need for defining local variable as oppose to returning `UIStackView` directly just like we do in `headerImageView` and other places?

Kickstarter-iOS/Assets.xcassets/icons/icon--currency-header.imageset/Contents.json
justinswart: Agreed! I'm pretty sure this started as an error and then became convention. @Scollaco tells me otherwise 😂

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
justinswart: Sure - I'll improve it!

Kickstarter-iOS/Views/SelectCurrencyTableViewHeader.swift
justinswart: ok good point - I think we were trying a bunch of things to get it to layout properly and may have got a false positive on this change 👍

Kickstarter-iOS/Views/SelectCurrencyTableViewHeader.swift
justinswart: There is no need!

dusi:
Forgot to ask whether we should set the label background color explicitly to prevent layer blending?

screen shot 2019-02-06 at 2 15 32 pm

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewControllerTests.swift
```diff
@@ -0,0 +1,31 @@
...
+    UIView.setAnimationsEnabled(false)
+  }
+
+  override func tearDown() {
+    UIView.setAnimationsEnabled(true)
+    super.tearDown()
+  }
+
+  func testView() {
+    combos(Language.allLanguages, [Device.phone4_7inch, Device.phone5_8inch, Device.pad])
```
ifbarrera: I think we can do `Devices.allCases`

justinswart: Sure, good idea!

Kickstarter-iOS/Views/SelectCurrencyTableViewHeader.swift
```diff
@@ -0,0 +1,62 @@
...
+      |> settingsDescriptionLabelStyle
+      |> \.textColor .~ .ksr_text_dark_grey_500
+      |> \.backgroundColor .~ .ksr_grey_200
+  }
+
+  // MARK: Accessors
+
+  public var text: String? {
+    didSet {
+      self.headerLabel.text = text
```
ifbarrera: Can we use the key-path setter? `_ = self.headerLabel |> \.text .~ text`

justinswart:
Thanks for the thorough review @ifbarrera!

PR #571, opened 2019-02-01
④ [Currency selection] Add change currency api calls

justinswart:
# 📲 What

Paired with @Scollaco to add the api calls to update the currency along with the save button on the nav bar. This also adds error messaging via the `MessageBannerViewController`.

# 🛠 How

Repurposed the network calls that were previously used for the picker and implemented the existing `LoadingBarButtonItemView` for saving.

# ♿️ Accessibility 

Will address in separate PR.

# ✅ Acceptance criteria

- [x] Save button is disabled when the currently selected currency in the table view matches the one that the user has set as their preferred currency.
- [x] Tapping the enabled save button displays an activity indicator and the newly preferred currency now behaves as the one before.
- [x] Errors are displayed in the message banner.

#568 
#569 
#570


justinswart:

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -62,6 +71,41 @@ final class SelectCurrencyViewController: UIViewController, MessageBannerViewCon
     }
   }
 
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.viewModel.outputs.activityIndicatorShouldShow
+      .observeForUI()
+      .observeValues { shouldShow in
```
ifbarrera: Can we use `[weak self]` here?

Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -62,6 +71,41 @@ final class SelectCurrencyViewController: UIViewController, MessageBannerViewCon
...
+
+    self.viewModel.outputs.saveButtonIsEnabled
+      .observeForUI()
+      .observeValues { [weak self] (isEnabled) in
+        self?.saveButtonView.setIsEnabled(isEnabled: isEnabled)
+    }
+
+    self.viewModel.outputs.updateCurrencyDidFailWithError
+      .observeForUI()
+      .observeValues { error in
```
ifbarrera: Same here, let's use `[weak self]`

Library/ViewModels/SelectCurrencyViewModel.swift
```diff
@@ -33,6 +39,51 @@ SelectCurrencyViewModelOutputs {
       initialChosenCurrency,
       self.didSelectCurrencySignal
     )
+
+    let updateCurrencyEvent = self.selectedCurrencyProperty.signal.skipNil()
+      .takeWhen(self.saveButtonTappedSignal.ignoreValues())
+      .switchMap { input in
+        AppEnvironment.current.apiService
+          .changeCurrency(input: ChangeCurrencyInput(chosenCurrency: input.rawValue))
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
```
ifbarrera: Any reason for this delay? I know we sometimes use it, sometimes we don't.

Library/ViewModels/SelectCurrencyViewModel.swift
```diff
@@ -33,6 +39,51 @@ SelectCurrencyViewModelOutputs {
...
+
+    self.activityIndicatorShouldShow = Signal.merge(
+      self.saveButtonTappedSignal.mapConst(true),
+      updateCurrencyEvent.filter { $0.isTerminating }.mapConst(false)
+    )
+
+    let updatedCurrency = updateCurrencyEvent.values()
+
+    updatedCurrency
+      .observeValues { AppEnvironment.current.koala.trackChangedCurrency($0) }
```
ifbarrera: Do we also need a koala event for the view loading?

Library/ViewModels/SelectCurrencyViewModel.swift
```diff
@@ -33,6 +39,51 @@ SelectCurrencyViewModelOutputs {
...
+
+    let initialAndSelected = Signal.combineLatest(
+      initialChosenCurrency,
+      self.didSelectCurrencySignal
+    )
+
+    let updatedAndSelected = Signal.combineLatest(
```
ifbarrera: How will this behave if there is an error? Should we take the initial currency in case of an error and merge `updatedCurrency` with `initialCurrencyOnError` to produce this `updatedAndSelected`?

Library/ViewModels/SelectCurrencyViewModel.swift
justinswart: Yeah we use `ksr_delay` so that we can do an `self.scheduler.advance()` in the tests. I'm doing this in #572 and it's here because I broke the work up retroactively for the multiple PR's.

Library/ViewModels/SelectCurrencyViewModel.swift
justinswart: Possibly - will find out from @nnekab what the event would be for that - there wasn't one before because it was a picker.

Library/ViewModels/SelectCurrencyViewModel.swift
```diff
@@ -33,6 +39,51 @@ SelectCurrencyViewModelOutputs {
...
       self.didSelectCurrencySignal
     )
+
+    let updateCurrencyEvent = self.selectedCurrencyProperty.signal.skipNil()
+      .takeWhen(self.saveButtonTappedSignal.ignoreValues())
+      .switchMap { input in
+        AppEnvironment.current.apiService
+          .changeCurrency(input: ChangeCurrencyInput(chosenCurrency: input.rawValue))
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .map { _ in input }
```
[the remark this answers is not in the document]
justinswart: It ignores the response from the mutation which is empty and passes back the currency that was selected so that we know it was successfully updated as the selected currency.

Library/ViewModels/SelectCurrencyViewModel.swift
justinswart: Neither of the combined signals can error 🤔

justinswart:
@ifbarrera ok, this PR is a little bigger now but ready for re-review

PR #574, opened 2019-02-04
[Payment methods] CVC bug fix

cdolm92:
# 📲 What

We were presenting a successful message banner when a credit card w/ an incorrect CVC or date was entered. 

# 🤔 Why

Instead we want to inform the user that the card failed to be added as a payment method. 

# 🛠 How

`addNewCreditCard(input: CreatePaymentSourceInput)` was returning `SignalProducer<GraphMutationEmptyResponseEnvelope, GraphError>` instead of  a `SignalProducer<CreatePaymentSourceEnvelope, GraphError>`.  The `CreatePaymentSourceEnvelope` has the field `isSuccessful` that determines if creating the payment source was successful and the field `errorMessage` to present when there is an error. 

# ♿️ Accessibility 
⚠️Found some a11y issues w/ this screen that would be best fixed in another PR ⚠️

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria
Test on Production

**Test w/ correct CVC**
- [x] Go to `Settings` navigate to `Account > Payment Methods` tap `+ Add a new card`
- [x] Enter your name and credit card info w/ **correct** CVC
- [x] Tap `Save` you should see an activity indicator and the view dismisses back to `Payment Methods` screen w/ a success banner. You should see your credit card listed

**Test w/ incorrect CVC**
- [x] Go to `Settings` navigate to `Account > Payment Methods` tap `+ Add a new card`
- [x] Enter your name and credit card info w/ **incorrect** CVC
- [x] Tap `Save` you should see an activity indicator, followed by a error banner. Navigate back to `Payment Methods` screen and you should not see your credit card listed

# ⏰ TODO

- [x] Fix merge issue
- [x] Testing a11y
- [x] Writing acceptance criteria


KsApi/MockService.swift
```diff
@@ -12,10 +12,9 @@ internal struct MockService: ServiceType {
   internal let currency: String
   internal let buildVersion: String
 
-  fileprivate let addNewCreditCardError: GraphError?
+  fileprivate let addNewCreditCardResult: Result<CreatePaymentSourceEnvelope, GraphError>?
```
justinswart: 👍 for converting to `Result<T, Error>`!

KsApi/mutations/CreatePaymentsSourceMutation.swift
```diff
@@ -12,6 +12,8 @@ public struct CreatePaymentSourceMutation<T: GraphMutationInput>: GraphMutation
     mutation createPaymentSource($input: CreatePaymentSourceInput!) {
     createPaymentSource(input: $input) {
       clientMutationId\
```
justinswart: Is the `\` here correct? What does that mean in GraphQL? I would think that there's no `\` and the three properties should align.

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -99,27 +99,36 @@ AddNewCardViewModelOutputs {
...
-    self.addNewCardSuccess = addNewCardEvent.values().ignoreValues()
-      .map { _ in Strings.Got_it_your_changes_have_been_saved() }
-
     let stripeInvalidToken = self.stripeErrorProperty.signal.map {
       $0?.localizedDescription
     }.skipNil()
     let graphError = addNewCardEvent.errors().map {
       $0.localizedDescription
     }
+    let addNewError = addNewCardEvent.map { $0.value?.errorMessage }.skipNil()
```
justinswart: `addNewCardError` for consistency 👍

KsApi/mutations/CreatePaymentsSourceMutation.swift
cdolm92: Thanks! we do not need this here. I've seen us do this in other schemas. It also seems like GraphQL does not require `\` in schemas so I went ahead and took those out.

ifbarrera: This was just to add a line break in a multi-line string literal. Not technically needed but it reads nicer when we print the mutations in the debug logs.

cdolm92: Thanks Isabel! I think we should keep them since it helps w/ readability on the debug log.

cdolm92:
@justinswart g2g? 👀

justinswart:
> @justinswart g2g? 👀

I think you just have a merge conflict to resolve and i still need to test it

PR #577, opened 2019-02-06
🔑 [Create Password] Screen Layout

dusi:
# 📲 What

`Create Password` screen layout (no functionality for now).

Includes
* Generic grouped settings header (reusable in the future)
* Text input cell (reusable)
* Generic grouped settings footer (reusable in the future)

# 🤔 Why

By using `UITableView` we're able to quickly create and reuse common components (header, cell, footer). Since we're using the native look and feel of iOS's grouped table we are using grouped table view instead of re-creating UI by hand. This gives us more flexibility and future-proofs the app should Apple decides to change this style or adds dark mode in the future or anything else. In addition to that we get a lot of a11y features for free.

# 🛠 How

We're using `UITableViewController` to do the heavy lifting of showing the UI (constructed from a simple enum), handling the keyboard and responding to Dynamic type changes.

# 👀 See

Please note that the screenshots have been generated before there was a change to the header padding. The bottom padding is actually `12` (not `30` as seen in the screenshots). So you might experience it to being way smaller on the device or simulator.

**Portrait**

| iPhone X (iOS 12.1) | iPhone 7 (iOS 11.4) | iPhone SE (iOS 10.3.1) |
| --- | --- | --- |
| screen shot 2019-02-05 at 5 05 21 pm | screen shot 2019-02-05 at 5 21 01 pm |  screen shot 2019-02-05 at 5 20 37 pm |

**Landscape**

| iPhone XS (iOS 12.1) |
| --- |
| screen shot 2019-02-05 at 5 21 57 pm | 

| iPhone 7 (iOS 11.4) |
| --- |
| screen shot 2019-02-05 at 5 21 59 pm | 

| iPhone SE (iOS 10.3.1) |
| --- |
| screen shot 2019-02-05 at 5 22 01 pm | 

# ▶️▶️▶️ Note
In order to being able to see the view controller please add the following code snippet to `SettingsAccountViewController.swift` on line `38` in `viewDidLoad`

```
// SettingsAccountViewController.swift
// viewDidLoad
// L: 38

DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
  let vc = CreatePasswordViewController.instantiate()
}
```

Then by going to `Settings > Account` you will be presented with the new `Create password` screen after a 1s delay.

# ♿️ Accessibility

**Needs to be tested on iPhone & iPad devices**
**Needs to be tested on iOS 10, 11 and 12**

- [x] Dynamic type resizes header, cell (switches between horizontal & vertical axis), footer properly - see attachments bellow
- [x] VoiceOver reads header, text fields and footer properly

| Small | Medium | Large | Extra extra large |
| --- | --- | --- | --- |
| screen shot 2019-02-05 at 5 39 46 pm | screen shot 2019-02-05 at 5 39 52 pm | screen shot 2019-02-05 at 5 40 01 pm | screen shot 2019-02-05 at 5 39 54 pm |
| screen shot 2019-02-05 at 5 35 14 pm | screen shot 2019-02-05 at 5 38 38 pm | screen shot 2019-02-05 at 5 06 44 pm | screen shot 2019-02-05 at 5 43 49 pm |

# ✅ Acceptance criteria

**Needs to be tested on iPhone & iPad devices**
**Needs to be tested on iOS 10, 11 and 12**

- [x] Includes and passes Snapshot tests
- [x] Shows `Create password` screen properly in portrait
- [x] Shows `Create password` screen properly in portrait with keyboard on
- [x] Shows `Create password` screen properly in landscape
- [x] Shows `Create password` screen properly in landscape with keyboard on
- [x] Shows disabled `Save` button in the navigation bar


Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,119 @@
+import Prelude
+import UIKit
+
+private struct Margin {
```
dusi: In the future, we can extract this to common constants file, but for now we can keep it here

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,119 @@
...
+    super.init(style: style, reuseIdentifier: reuseIdentifier)
+
+    self.stackView.addArrangedSubview(self.label)
+    self.stackView.addArrangedSubview(self.textField)
+
+    self.contentView.addSubview(self.stackView)
+    self.stackView.constrainEdges(to: self.contentView)
+    self.stackView.heightAnchor.constraint(greaterThanOrEqualToConstant: 44).isActive = true
+
+    self.label.setContentHuggingPriority(.required, for: .horizontal)
```
dusi: Who really knows how these things work ¯\_(ツ)_/¯

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,119 @@
...
+    super.traitCollectionDidChange(previousTraitCollection)
+
+    let traitCollection = self.traitCollection
+
+    if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+      self.configure(with: traitCollection)
+    }
+  }
+
+  private func configure(with traitCollection: UITraitCollection) {
```
dusi: This is the key to the Dynamic Type resizing functionality.

When the category becomes too big (accessible) we switch from horizontal alignment to vertical alignment and vice versa. This allows placing `label` and `text field` above each other or side by side (depending on the size category)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,119 @@
...
+      self.stackView.alignment = .fill
+      self.textField.textAlignment = .right
+    }
+  }
+}
+
+// MARK: - Extensions
+
+private extension UIView {
+  func constrainEdges(to view: UIView) {
```
dusi: Currently there seems to be another category that @justinswart is creating therefore I'm using a `private` extension. We can potentially extract this in the future when all code merges.

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,119 @@
...
+      self.topAnchor.constraint(equalTo: view.topAnchor),
+      self.rightAnchor.constraint(equalTo: view.rightAnchor),
+      self.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+      self.leftAnchor.constraint(equalTo: view.leftAnchor)
+      ])
+  }
+}
+
+private extension UIContentSizeCategory {
+  func ksr_isAccessibilityCategory() -> Bool {
```
dusi: Just a convenience accessor to hide the ugly availability check.

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+      |> \.title %~ { _ in Strings.Create_password() }
+
+    _ = self.navigationItem
+      |> \.rightBarButtonItem .~ self.rightBarButtonItem
+
+    _ = self.tableView
+      |> \.allowsSelection .~ false
+      |> \.separatorInset .~ .zero
+
+    if #available(iOS 11, *) { } else {
```
dusi: Unfortunately necessary for iOS 10. Otherwise it doesn't know how to size the header, the cell nor the footer.

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,121 @@
...
+private struct Margin {
+  static let leftRight: CGFloat = 8
+  static let topBottom: CGFloat = 15
+}
+
+class SettingsTextInputCell: UITableViewCell {
+  // MARK: - Accessors
+
+  private lazy var stackView: UIStackView = {
+    let layoutMargins = UIEdgeInsets.init(topBottom: Margin.topBottom, leftRight: Margin.leftRight)
```
dusi: Compiler couldn't infer simple `.init` on my machine, so I had to use `UIEdgeInsets.init` explicitly

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,121 @@
...
+      |> \.isLayoutMarginsRelativeArrangement .~ true
+      |> \.layoutMargins .~ layoutMargins
+      |> \.spacing .~ 8
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private lazy var label: UILabel = {
+    return UILabel()
+      |> \.adjustsFontForContentSizeCategory .~ true
+      |> \.font .~ UIFont.preferredFont(forTextStyle: .body)
```
dusi: Our settings styles don't use `UIFont` properly in order to support Dynamic Type. Therefore I've opted for basic implementation (the could potentially be extracted to a style, but I didn't see the value at this step)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,121 @@
...
+      |> \.isSecureTextEntry .~ true
+  }()
+
+  public func configure(with title: String, placeholder: String) {
+    _ = self.label
+      |> \.text .~ title
+
+    _ = self.textField
+      |> \.placeholder .~ placeholder
+      |> \.accessibilityLabel .~ self.label.accessibilityLabel
```
dusi: Uses `label`'s accessibility label (`label` is not an accessible element in order to simplify information density)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,121 @@
...
+    self.stackView.constrainEdges(to: self.contentView)
+    self.stackView.heightAnchor.constraint(greaterThanOrEqualToConstant: 44).isActive = true
+
+    self.label.setContentCompressionResistancePriority(.required, for: .horizontal)
+    self.label.setContentHuggingPriority(.required, for: .horizontal)
+
+    self.textField.setContentHuggingPriority(.defaultLow, for: .horizontal)
+  }
+
+  required init?(coder aDecoder: NSCoder) {
```
dusi: The day Apple removes these required initializers for `UIView` subviews, we'll party! 😄

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,121 @@
...
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: - Layout
+
+  override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
dusi: This is where the Dynamic Type magic happens

Kickstarter-iOS/Views/SettingsGroupedFooterView.swift
```diff
@@ -0,0 +1,36 @@
+import Prelude
+import UIKit
+
+private struct Padding {
+  static let horizontal: CGFloat = 8
+  static let vertical: CGFloat = 12
+}
+
+class SettingsGroupedFooterView: UITableViewHeaderFooterView {
```
dusi: We had to customize the `UITableViewHeaderFooterView` because it would unfortunately work well with iOS 10. The auto layout on iOS 10 didn't work properly with this view, so we couldn't use the default `textLabel` or `detailTextLabel` but had to add a custom `UILabel` into the view hierarchy.

dusi:
[🔑 Create password - Create password screen UI only](https://trello.com/c/pvJIFApU/1187-%F0%9F%94%91-create-password-create-password-screen-ui-only)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,151 @@
...
+private struct Margin {
+  static let leftRight: CGFloat = 8
+  static let topBottom: CGFloat = 15
+}
+
+/*
+
+ Preview 1
+ 
+ +---------------------------------------+
```
dusi: Just an example of how we can better document the view hierarchy without the Interface Builder.

This was generated using https://textik.com tool.

dusi: I realize we currently have problems running playgrounds, but one other suggestion I can try is creating a UI only playground (under the same branch as our UI code) and populate it with some values in order for us to being able to preview this without needing to build the app.

dusi:
I wanted to run something by you @dannyalright since I realized using native `UITableView` with `grouped` style uses slightly different background color to what we have been using everywhere else?

Can you check these two screenshots side by side and let me know how inconvenient this is? I'd very much prefer to go with plain `grouped` style without further color changes, but that will introduce inconsistency to the rest of settings (which might be a bit of work to re-style properly).

| Change password (and the rest of settings) | Create password |
| --- | --- |
| screen shot 2019-02-06 at 11 52 21 am | screen shot 2019-02-06 at 11 52 26 am |

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        |> \.estimatedSectionFooterHeight .~ estimatedSectionFooterHeight
+        |> \.estimatedSectionHeaderHeight .~ estimatedSectionHeaderHeight
+        |> \.estimatedRowHeight .~ estimatedRowHeight
+        |> \.sectionFooterHeight .~ height
+        |> \.sectionHeaderHeight .~ height
+        |> \.rowHeight .~ height
+    }
+
+    self.tableView.register(
+      SettingsGroupedHeaderView.self, forHeaderFooterViewReuseIdentifier: ReuseIdentifier.header
```
Scollaco: Since we are moving away from storyboards, I think we could create a UITableView extension with a function similar to the one we have on `Nib.swift` to register headers and footers. Something like:
```
public func registerHeaderFooter(_ headerFooter: UITableViewHeaderFooterView.Type,
                                   inBundle bundle: Bundle = .framework) {
    self.register(
      headerFooter,
      forHeaderFooterViewReuseIdentifier: headerFooter
        .description()
        .components(separatedBy: ".")
        .dropFirst()
        .joined(separator: "."))
  }
```

So we could use `self.tableView.registerHeaderFooter(SettingsGroupedHeaderView.self)` instead. For this, would just need to make sure that the identifier is the class name `"SettingsGroupedHeaderView"`,`"SettingsGroupedFooterView"`, etc. For this, we could go even a little further and create a protocol `Identifiable` (or whatever) with a `identifier` property that gives us the identifier automatically. 

But we can discuss this later on.

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    )
+    self.tableView.register(
+      SettingsGroupedFooterView.self, forHeaderFooterViewReuseIdentifier: ReuseIdentifier.footer
+    )
+    self.tableView.register(SettingsTextInputCell.self, forCellReuseIdentifier: ReuseIdentifier.cell)
```
Scollaco: And the same thing for cells :)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
ifbarrera: I'm also curious, why do you say our styles don't support dynamic type? All of our font functions (`.ksr_body()` etc) under the hood use `UIFont.preferredFont(forTextStyle: .body)`

dannyalright:
> I wanted to run something by you @dannyalright since I realized using native `UITableView` with `grouped` style uses slightly different background color to what we have been using everywhere else?
> 
> Can you check these two screenshots side by side and let me know how inconvenient this is? I'd very much prefer to go with plain `grouped` style without further color changes, but that will introduce inconsistency to the rest of settings (which might be a bit of work to re-style properly).

It's a little disorientating. I see the value in using the `grouped` style, though. Is it not possible to maintained the `grouped` style and just change the background colour? 

> which might be a bit of work to re-style properly

Are you saying the same thing I just described? Or the inverse; changing the rest of Settings to use `grouped` style with the iOS grey background?

dusi:
@dannyalright false alarm, we can style the background appropriately. On the other hand changing the rest of the settings seems pretty trivial as well so we can pick a better matching color and go with it.

| `.ksr_grey_200` (current) | `.groupTableViewBackground` (system) |
| --- | --- |
| screen shot 2019-02-15 at 2 59 30 pm | screen shot 2019-02-15 at 2 58 05 pm |
| screen shot 2019-02-15 at 3 02 33 pm | screen shot 2019-02-15 at 2 58 10 pm |

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
dusi: I have removed these comments, they didn't grow on me.

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
dusi: I was able to make the above work (resize the font when the font size changes) with the following:

```
|> \.font .~ .ksr_body()
```

so it looks like `adjustsFontForContentSizeCategory` is the key for this to properly resize the `UILabel`/`UITextField` on iOS 10+.

I originally suspected the problem was in `Fonts.swift` where we declare a convenience font accessor which calculates font size based on hard coded values

```
static func preferredFont(style: UIFont.TextStyle, size: CGFloat? = nil) -> UIFont { }
```

and is used by majority of our styles (`.ksr_body()`, etc.). If you look at the font that's being returned from the method it's actually constructed using `descriptor` not `preferredFont` so my initial assumption was that this was causing Dynamic Type to not resize properly, but it seems like that's not the case. This method basically uses custom scale so that we can tweak the size of the font to our likings.

@Scollaco do you have (by any chance) any context to why we implemented `preferredFont` this way? Looking at the method the two sole contributors is Brandon and you.

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,151 @@
...
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: - Layout
+
+  override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+    super.traitCollectionDidChange(previousTraitCollection)
+
+    let traitCollection = self.traitCollection
```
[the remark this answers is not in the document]
dusi: It's not only reused twice, but also allows the if condition to fit on one line 😅 

screen shot 2019-02-15 at 3 48 52 pm

dannyalright:
> @dannyalright false alarm, we can style the background appropriately. On the other hand changing the rest of the settings seems pretty trivial as well so we can pick a better matching color and go with it.

@dusi let's stick with `.ksr_grey_200` as our background throughout Settings, please. We'll likely get a color refresh in the next few months which we can nudge towards.

dusi:
Please hold on to the review. I have to fix Dynamic Type related problem.

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,82 @@
...
+    self.contentView.addSubviewConstrainedToMargins(self.stackView)
+  }
+
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: - Styles
+
+  override func bindStyles() {
```
dusi: Moved styling to `bindStyles` for consistency with the rest of the codebase

Kickstarter-iOS/Views/SettingsGroupedFooterView.swift
```diff
@@ -0,0 +1,45 @@
+import Library
+import Prelude
+import UIKit
+
+final class SettingsGroupedFooterView: UITableViewHeaderFooterView {
+  // MARK: - Properties
+  
+  private(set) lazy var label: UILabel = {
+    UILabel(frame: .zero)
+      |> settingsHeaderFooterLabelBaseStyle
```
dusi: The base style has to be available in `init` so that the auto layout & dynamic sizing is established correctly. The first time we're instantiating this lazy property is in `init` when we call `self.contentView.addSubviewConstrainedToMargins(self.label)`  I'm not entirely sure why at this point but spend enough time on this so we should accept this and maybe look into it at some time in the future.

Library/Styles/SettingsStyles.swift
```diff
@@ -79,3 +79,53 @@ public func settingsAttributedPlaceholder(_ string: String) -> NSAttributedStrin
...
+
+public let settingsHeaderContentViewStyle =
+  settingsLayoutMarginsStyle
+    <> UIView.lens.layoutMargins .~ settingsHeaderLayoutMargins
+
+public let settingsFooterContentViewStyle =
+  settingsLayoutMarginsStyle
+
+public let settingsHeaderFooterLabelBaseStyle =
+  UILabel.lens.font %~ { _ in .ksr_footnote() }
```
dusi: Notice that the `font` needs to use `%~` this is due to the fact that we don't want it to cache the old size value. We have paired on this with Justin and is one of the major issues why fonts are not resizing properly in Settings. I'll be fixing those separately.

dusi:
There's been quite a bit of work on this so I was wondering if I could get a sanity check @justinswart @Scollaco ? 🙏

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -0,0 +1,61 @@
...
+    super.bindStyles()
+
+    _ = self.contentView
+      |> settingsContentViewStyle
```
dusi: Extracted common styling properties to `SettingsStyles`

Library/Styles/SettingsStyles.swift
```diff
@@ -79,3 +79,85 @@ public func settingsAttributedPlaceholder(_ string: String) -> NSAttributedStrin
     attributes: [NSAttributedString.Key.foregroundColor: UIColor.ksr_text_dark_grey_400]
   )
 }
+
+public func settingsContentViewStyle(_ view: UIView) -> UIView {
```
dusi: Since we've decided to move away from using lenses I've used key paths and functions.

Also ordered alphabetically.

dusi: (didn't touch the previous styles for now)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
Scollaco: Thanks for digging in. Unfortunately this was implemented way before I started working here. Maybe @justinswart has a little context?

dusi: I noticed you've mostly just updated it for Swift 4.2 so wanted to be sure you're not holding some secrets Nino :) 

For clarification, I've chatted with Justin offline and it appears we're using this constructor in order to customize the font scale. You can easily replace the implementation of this with 
```
fileprivate static func preferredFont(style: UIFont.TextStyle, size: CGFloat? = nil) -> UIFont {
    return UIFont.preferredFont(forTextStyle: style)
}
```
and see the font difference when running snapshot tests.

Kickstarter-iOS/Views/SettingsGroupedFooterView.swift
```diff
@@ -0,0 +1,45 @@
...
+      |> settingsHeaderFooterLabelStyle
+  }
+}
+
+// MARK: - HeaderFooterViewProtocol
+
+extension SettingsGroupedFooterView: HeaderFooterViewProtocol {
+  func configure(with text: String) {
+    _ = self.label
+      |> \.text %~ { _ in text }
```
justinswart: Seeing as we're not using `Strings.*` here it should be sufficient to just do `.~ text`, either is fine. It's always the responsibility of wherever this value originally comes from to ensure that it's evaluated in a closure for translation.

Kickstarter-iOS/Views/SettingsGroupedHeaderView.swift
```diff
@@ -0,0 +1,45 @@
...
+      |> settingsHeaderFooterLabelStyle
+  }
+}
+
+// MARK: - HeaderFooterViewProtocol
+
+extension SettingsGroupedHeaderView: HeaderFooterViewProtocol {
+  func configure(with text: String) {
+    _ = self.label
+      |> \.text %~ { _ in text }
```
justinswart: Same comment, up to you.

Library/DataSource/UITableView-Extensions.swift
```diff
@@ -1,18 +1,42 @@
...
 
 public extension UITableView {
-  public func registerCellClass <CellClass: UITableViewCell> (_ cellClass: CellClass.Type) {
-    register(cellClass, forCellReuseIdentifier: cellClass.description())
+  public func registerHeaderFooterClass(_ headerFooterClass: UITableViewHeaderFooterView.Type) {
+    let className = self.classNameWithoutModule(headerFooterClass)
+    self.register(headerFooterClass, forHeaderFooterViewReuseIdentifier: className)
   }
 
+  public func registerCellClass(_ cellClass: UITableViewCell.Type) {
```
justinswart: Why the change?

Library/Styles/SettingsStyles.swift
```diff
@@ -79,3 +79,85 @@ public func settingsAttributedPlaceholder(_ string: String) -> NSAttributedStrin
...
+  return view
+    |> \.layoutMargins .~ .init(all: Styles.grid(2))
+    |> \.preservesSuperviewLayoutMargins .~ false
+}
+
+public let settingsFooterContentViewStyle =
+  settingsContentViewStyle
```
justinswart: This can fit on one line 👍

Library/UITraitCollectionTests.swift
```diff
@@ -0,0 +1,45 @@
+import XCTest
+@testable import Library
+
+final class UITraitCollectionTests: XCTestCase {
```
justinswart: 👍

Library/UITraitCollection.swift
```diff
@@ -0,0 +1,18 @@
+import UIKit
+
+public extension UITraitCollection {
+  public func ksr_isAccessibilityCategory() -> Bool {
```
justinswart: How about a `switch` and early `return`?

```swift
  public func ksr_isAccessibilityCategory() -> Bool {
    let sizeCategory = self.preferredContentSizeCategory

    if #available(iOS 11, *) {
      return sizeCategory.isAccessibilityCategory
    }

    switch sizeCategory {
      case .accessibilityMedium,
           .accessibilityLarge,
           .accessibilityExtraLarge,
           .accessibilityExtraExtraLarge,
           .accessibilityExtraExtraExtraLarge:
      return true
    default:
      return false
    }
  }
```

Library/DataSource/UITableView-Extensions.swift
dusi: Making this more consistent with the rest of these functions.

Library/UITraitCollection.swift
dusi: My thinking here was simple `2 returns` vs `3 returns` 😄 Happy to change, just let me know

justinswart: I guess I just don't like all the `||`'s 😄 

You know my favourite would be:

```swift
  public func ksr_isAccessibilityCategory() -> Bool {
    let sizeCategory = self.preferredContentSizeCategory

    if #available(iOS 11, *) {
      return sizeCategory.isAccessibilityCategory
    } else {
      return [
        .accessibilityMedium,
        .accessibilityLarge,
        .accessibilityExtraLarge,
        .accessibilityExtraExtraLarge,
        .accessibilityExtraExtraExtraLarge
      ]
      .contains(sizeCategory)
    }
  }
```

Do it for fun :trollface:

dusi: I can do that. Just wondering if the contains wouldn’t be more expensive that OR? 🤔

justinswart: Doubt it would be noticeable, even if it was, it's only iOS 10 users 😄 

By the way we can also lose the `else` and just return 👍

PR #578, opened 2019-02-06
[Payment methods] Bugs & visual fixes

ifbarrera:
# 📲 What

Fixes the following:
- payment methods table view should appear without a delay on `viewDidLoad`
- removes custom "header" view and instead created `SettingsTableViewHeader` and added it as the header view of the table view. This means the header now moves with the table view as you scroll.
- removes manually added "spacer" on the `PaymentMethodsFooterView`. Checked this with Danny, it aligns better with Apple's Wallet design.
- made `PaymentMethodsFooterView` the table view's footer (`tableFooterView`) instead of the section's footer, so that even if there is an error fetching cards, the “Add New Card” button still displays

Tested on:
- Sade iPhone 6s+ (10.3.2)
    - Portrait, landscape
- Elvis iPhone 6s+ (11.2.2)
    - Portrait, landscape
- iPhone X (12.1.2)
    - Portrait, landscape

# 🤔 Why

Visual cleanup and small bug fixes.

# 👀 See

Before and after removing the custom view "header" and instead using the table view's header:

**Before**
mwbc95kehm
**After**
lfszgicfow

Before and After, with 100% connection loss:

**Before**
simulator screen shot - iphone x - 2019-02-05 at 18 20 37

**After**
simulator screen shot - iphone x - 2019-02-05 at 18 17 51

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Payment methods should display and function as normal. Scrolling the list should also scroll the header
- [x] Payment methods should correctly display in landscape & portrait mode on iOS 10, iOS 11 and iOS 12
- [x] With no network connection, payment methods screen should show the "Add Card" button


Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -46,6 +49,23 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
...
   }
 
+  override func viewDidLayoutSubviews() {
+    super.viewDidLayoutSubviews()
+
+    guard let footerView = tableView.tableFooterView, let headerView = tableView.tableHeaderView else {
+      return
+    }
+
+    let footerViewHeight = footerView.systemLayoutSizeFitting(UIView.layoutFittingCompressedSize).height
```
ifbarrera: Table header & footer views require an explicit height. Setting the height in `viewDidLayoutSubviews` guarantees that the height is correct when we move from portrait to landscape mode using `systemLayoutSizeFitting`.

Kickstarter-iOS/Views/Storyboards/PaymentMethodsFooterView.xib
```diff
@@ -16,54 +16,45 @@
...
-                    <rect key="frame" x="0.0" y="37" width="375" height="44"/>
+                <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="QRi-RW-gGm" userLabel="Content View">
+                    <rect key="frame" x="0.0" y="0.0" width="375" height="81"/>
                     <subviews>
                         <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="left" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="ZVF-Hv-L8e">
-                            <rect key="frame" x="0.0" y="0.0" width="375" height="44"/>
+                            <rect key="frame" x="0.0" y="0.0" width="375" height="81"/>
                             <constraints>
-                                <constraint firstAttribute="height" constant="44" id="RY1-Id-ADx"/>
+                                <constraint firstAttribute="height" relation="greaterThanOrEqual" priority="750" constant="44" id="RY1-Id-ADx"/>
```
ifbarrera: Lowered the priority of the height constraint so there's no ambiguity with the constraints the table view gives this view.

Kickstarter-iOS/Views/Storyboards/PaymentMethodsFooterView.xib
```diff
@@ -16,54 +16,45 @@
...
+                            <rect key="frame" x="0.0" y="80" width="375" height="1"/>
+                            <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                            <constraints>
+                                <constraint firstAttribute="height" constant="1" id="YKt-EX-gZm"/>
+                            </constraints>
+                        </view>
                     </subviews>
                     <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
                     <constraints>
-                        <constraint firstAttribute="trailing" secondItem="ZVF-Hv-L8e" secondAttribute="trailing" id="3IW-Bp-AWL"/>
```
ifbarrera: Removed the additional "padding view" we had in this footer.

Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -681,20 +681,8 @@
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
-                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="xml-QE-znh">
```
ifbarrera: Removed this manually added "header view" and instead created an actual table view header. This scene is now just a table view.

ifbarrera:
[Payment Methods 🐛 - Table appears with a delay](https://trello.com/c/tH2aeoo7/1195-payment-methods-%F0%9F%90%9B-table-appears-with-a-delay)

ifbarrera:
[Payment Methods 🐛 - Disable floating header view](https://trello.com/c/RcFKgdTz/1198-payment-methods-%F0%9F%90%9B-disable-floating-header-view)

Kickstarter-iOS/Views/SettingsTableViewHeader.swift
```diff
@@ -0,0 +1,21 @@
+import Library
+import Prelude
+
+final class SettingsTableViewHeader: UITableViewHeaderFooterView {
```
dusi: Looks like there's still some confusion between when to use `UIView` and when to use `UITableViewHeaderFooterView` - https://github.com/kickstarter/ios-oss/pull/570#discussion_r254473087.

As per the documentation the `UITableViewHeaderFooterView` should only be used as a reusable section header / footer - https://developer.apple.com/documentation/uikit/uitableviewheaderfooterview. 

On the other hand `UITableView`'s header / footer is simply an accessory view (non-reusable).

The difference is the reusability (which we don't really need in case of `tableHeaderView`/`tableFooterView` there's always going to be just one instance. The section headers/footers are different though.

In order to use `UIView` subclass as a `tableHeaderView` it looks like we need to do the following..

1. Convert `SettingsTableViewHeader` to a `UIView` subclass
2. Remove `|> \.preferredMaxLayoutWidth .~ self.titleLabel.bounds.width` from `SettingsTableViewHeader` (auto layout will do this for us)
3. Simply assign the header view

```
// We no longer need this since this is mostly for section header/footer views
self.tableView.registerHeaderFooter(nib: .SettingsTableViewHeader)
```

in `viewDidLoad` set the `tableView.tableHeaderView` and add width constraint so that it's properly resizing based on the screen size

```
if let tableHeaderView = SettingsTableViewHeader.fromNib(nib: Nib.SettingsTableViewHeader) {
  tableHeaderView.translatesAutoresizingMaskIntoConstraints = false
  tableHeaderView.configure(with: Strings.Any_payment_methods_you_saved_to_Kickstarter())

  self.tableView.tableHeaderView = tableHeaderView

  NSLayoutConstraint.activate([
    tableHeaderView.widthAnchor.constraint(equalTo: self.tableView.widthAnchor)
    ])
  }
```

Remove `viewDidLayoutSubviews` since it's no longer necessary. I've tried to follow these articles but didn't have much luck of properly resizing without setting the `width` constraint.
https://useyourloaf.com/blog/variable-height-table-view-header/
https://collindonnell.com/2015/09/29/dynamically-sized-table-view-header-or-footer-using-auto-layout/

🤷‍♂️ 

The above should remove a lot of boiler plate and we're also following best practices of how to use `tableHeaderView`/`tableFooterView` so it should be future proof should Apple made any changes to the way headers/footers are presented.

Same applies to the footer view.

ifbarrera: I agree it's not necessary to make these `UITableViewHeaderFooterView`s 👍  However I was never able to get the accessory views to properly size using only autolayout. Maybe we can pair on it on Monday.

dusi: Sounds good...the trick is having the `headerView` size itself (by setting its subview's auto layout - in our case just the label)..and then establishing the width constraint as outlined above.

Lets pair on that when you have a minute.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -27,12 +27,12 @@ internal final class DiscoveryPageViewController: UITableViewController {
   internal override func viewDidLoad() {
     super.viewDidLoad()
 
+    self.tableView.register(nib: Nib.DiscoveryPostcardCell)
+
     self.tableView.addSubview(self.loadingIndicatorView)
 
     self.tableView.dataSource = self.dataSource
 
-    self.tableView.register(nib: Nib.DiscoveryPostcardCell)
```
ifbarrera: Ok so this was weird, I was testing and suddenly I started seeing this crash where this table view was attempting to dequeue a cell of type `DiscoverPostcardCell` that hadn't been registered. It was very strange, I hadn't touched any of this work and I kept getting it every time I ran the app. I moved the `register` call to the top (before the `dataSource` is set) and that fixed the crash. No idea why it started happening now out of nowhere, but I figured its a good idea to register the cell before setting the `dataSource` either way.

dusi:
Tested on:
SIM: iPhone SE - iOS 10.3.1, iOS 11.4, iOS 12.1
Device: iPhone XS - iOS 12.1

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -117,24 +123,49 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
...
+      footer.delegate = self
+
+      let footerContainer = UIView(frame: .zero)
+      footerContainer.addSubview(footer)
+
+      self.tableView.tableFooterView = footerContainer
+
+      footer.constrainEdges(to: footerContainer)
+
+      _ = footer.widthAnchor.constraint(equalTo: self.tableView.widthAnchor)
```
ifbarrera: @justinswart updating the priority here, in the header's constraints and in the xibs fixed all the auto-layout warnings (there were also some happening when moving into landscape mode).

Priorities updated:
- height constraint on the `PaymentFooterView xib` -> `.defaultHigh`
- width constraint on footer and header -> `.defaultHigh`
- top and bottom constraints on the `UILabel` in `SettingsTableViewHeader` -> `.defaultHigh`

Tested on iPhone X (12.1), iPhone 6 (11.4), iPhone 5s (10.3.1)

PR #579, opened 2019-02-06
[Payment methods] Minor bug fixes

ifbarrera:
# 📲 What

Fixes the following:
- table view should end editing when the user navigates to "Add New Card"
- success message banner should only display *after* the "Add New Card" screen has fully dismissed

🔴 NOTE: Also fixes a bug where switching from production -> staging didn't actually change the Stripe publishable key. This meant you sometimes couldn't add a test card in staging (because Stripe thought you were adding the card to production). Now the `StripePublishableKey` comes from the `Secrets` file. 

Tested on:
	- SIM iPhone 5s 10.3.1
	- SIM iPhone 7+ 11.4
	- SIM iPhone X 12.1
	- SIM iPad Pro 12.1

# 👀 See

assetdcnfd
5wkwkuplz2

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Navigate to payment methods, then tap "Edit". Table view should show in editing mode. Then tap "Add New Card", then close the "Add New Card" screen. Table view should not be in editing mode.
- [x] Navigate to "Add New Card" screen and add a new card. On success, the message banner should only display after the "Add New Card" screen has dismissed

### DIFF
--- a/Configs/Secrets.swift.example
+++ b/Configs/Secrets.swift.example
@@ -63,6 +63,11 @@ public enum Secrets {
     public static let endpoint = "https://api.com/stuff"
   }
 
+  public enum StripePublishableKey {
+  	public static let staging = "deadbeef"
+  	public static let production = "beefdead"
+  }
+
   public enum WebEndpoint {
     public static let production = "www.kickstarter.com"
     public static let staging = "staging.com"
--- a/Configs/Secrets.swift.example
+++ b/Configs/Secrets.swift.example
@@ -64,8 +64,8 @@ public enum Secrets {
   }
 
   public enum StripePublishableKey {
-  	public static let staging = "deadbeef"
-  	public static let production = "beefdead"
+    public static let staging = "deadbeef"
+    public static let production = "beefdead"
   }
 
   public enum WebEndpoint {
--- a/Library/Environment.swift
+++ b/Library/Environment.swift
@@ -46,6 +46,11 @@ public struct Environment {
   /// The current device running the app.
   public let device: UIDeviceType
 
+  /// Returns the current environment type
+  public var environmentType: EnvironmentType {
+    return self.apiService.serverConfig.environment
+  }
+
   /// A delegate to handle Facebook initialization and incoming url requests
   public let facebookAppDelegate: FacebookAppDelegateProtocol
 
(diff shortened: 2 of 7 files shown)


Library/Environment.swift
```diff
@@ -46,6 +46,11 @@ public struct Environment {
   /// The current device running the app.
   public let device: UIDeviceType
 
+  /// Returns the current environment type
+  public var environmentType: EnvironmentType {
+    return self.apiService.serverConfig.environment
```
ifbarrera: I got tired of always typing this out when trying to determine the current environment type. Now you can just do `AppEnvironment.current.environmentType` 😸

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -215,13 +215,11 @@ internal final class AddNewCardViewModelTests: TestCase {
   }
 
   func testSetPublishableKey() {
```
Scollaco: Maybe we could update the test (or create a new one) that checks if the `stripePublishableKey` is correct for staging and production?

ifbarrera: This is a bit tricky because the key is different on Circle than on our local machines (because on Circle it skips cloning the `Secrets` repo).

Configs/Secrets.swift.example
```diff
@@ -63,6 +63,11 @@ public enum Secrets {
     public static let endpoint = "https://api.com/stuff"
   }
 
+  public enum StripePublishableKey {
+  	public static let staging = "deadbeef"
```
justinswart: Thanks for adding to the template, let's keep the 2-space indentation though 👍

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -241,4 +240,9 @@ AddNewCardViewModelOutputs {
 
   public var inputs: AddNewCardViewModelInputs { return self }
   public var outputs: AddNewCardViewModelOutputs { return self }
+
+  static func publishableKey(for environment: EnvironmentType) -> String {
```
justinswart: Any reason you chose to use a `static` function here and not just a `private` function in this file?

ifbarrera: Ah yeah it's because otherwise I get a `'self' captured by a closure before all members were initialized` error

justinswart: Oh right but why not make this a global `private` function in the file so outside of the type definition?

PR #580, opened 2019-02-06
Add CircleCI step to pre-load simulator before running tests

justinswart:
# 📲 What

Adds a step to all of our jobs in `circle.yml` that run tests that will pre-load the iOS simulator before running those tests.

# 🤔 Why

We've had a bunch of jobs failing with an error that the simulator can't be found:

screen shot 2019-02-06 at 2 57 52 pm

According to CircleCI it is sometimes useful to [pre-load](https://circleci.com/docs/2.0/testing-ios/#pre-starting-the-simulator) the simulator before running tests to avoid this.

# 🛠 How

- Added step to pre-load simulator in each job that runs tests.

# ✅ Acceptance criteria

- [ ] All test-running jobs complete successfully on CircleCI with this step in place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,7 @@ stripe_sdk_version: &stripe_sdk_version 13.2.0
 store_fabric_sdk_version: &store_fabric_sdk_version echo "$FABRIC_SDK_VERSION" > fabric_version.txt
 store_opentok_version: &store_opentok_version echo "$OPENTOK_VERSION" > opentok_version.txt
 store_stripe_version: &store_stripe_version echo "$STRIPE_SDK_VERSION" > stripe_version.txt
+preload_simulator: &preload_simulator xcrun instruments -w "iPhone 8 (12.1) [" || true
 
 base_job: &base_job
   macos:
@@ -133,6 +134,10 @@ jobs:
           command: make bootstrap
 
       - run:
+          name: Pre-load simulator
+          command: *preload_simulator
+
+      - run:
           name: Kickstarter tests
           command: PLATFORM=iOS make test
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -8,6 +8,7 @@ stripe_sdk_version: &stripe_sdk_version 13.2.0
 store_fabric_sdk_version: &store_fabric_sdk_version echo "$FABRIC_SDK_VERSION" > fabric_version.txt
 store_opentok_version: &store_opentok_version echo "$OPENTOK_VERSION" > opentok_version.txt
 store_stripe_version: &store_stripe_version echo "$STRIPE_SDK_VERSION" > stripe_version.txt
+preload_simulator: &preload_simulator xcrun instruments -w "iPhone 8 (12.1) [" || true
```
justinswart: Added this as a variable so that when we need to increment the simulator version we only need to change it in one place.

.circleci/config.yml
```diff
@@ -132,6 +133,10 @@ jobs:
           name: Make bootstrap
           command: make bootstrap
 
+      - run:
+          name: Pre-load simulator
+          command: *preload_simulator
```
justinswart: When we had this in place before we sometimes followed it with a `sleep 15` command so that it would wait a while before running the next step. I'd like to try it without that first so see if this solves it on its own.

.circleci/config.yml
dusi: Can't we use existing `&iphone_name` and add new constant `&iOS_version`?

justinswart: I don't think yaml works like that unfortch 🤔 https://stackoverflow.com/a/5954748/2271759

PR #581, opened 2019-02-07
[Payment Methods] Update padding and image view size

dusi:
# 📲 What

Fixes the right margin on the credit card cell to follow our conventions of using `12 pts` instead of `6 pts` (which seemed too small).

# 🤔 Why

Better UX.

# 🛠 How

Initially I was simply going to update the right padding of the `label`. But since the `label` is placed inside of a `stackView` it makes sense to rather update the `stackView`'s horizontal padding (so that it's identical on both sides - left & right). Unfortunately what this does is that it adds more padding from the left side of the `imageView` which makes the gap between screen edge and the `image` even bigger). 

After doing some investigation I've found out that we're not sizing the `imageView` properly in the first place. The `PDF` image assets that we use have `72 × 45` dimensions. Previously we have set the `height` of the image view to `36 pts` in order to scale it down (this is fine since we're using `PDF` resources), but we completely forgot to set the `width` constraint, therefore the `imageView` has been stretching a bit and adding white padding to the `image` itself.

We should always use the proper `width`/`height` of the `image` to present it therefore I've also updated the constraints of the `imageView`. This is not only for better reasoning about the UI but also performance reasons.

See attachments for more details.

# 👀 See

| Before (notice the right padding) | After |
| --- | --- |
| screen shot 2019-02-07 at 2 59 58 pm | screen shot 2019-02-07 at 3 11 46 pm |
| screen shot 2019-02-07 at 3 06 14 pm | screen shot 2019-02-07 at 3 14 02 pm |

**Constraints before**
screen shot 2019-02-07 at 3 37 15 pm

**Constraints after**
screen shot 2019-02-07 at 3 15 32 pm

# ✅ Acceptance criteria

Best replicated on `iPhone SE` with `French` set as the language.

- [x] If the title label is too long it uses proper padding on the right side (which is identical to the left padding of the image view)

:trollface: https://twitter.com/iamdevloper/status/1093483640426643456 :trollface:


Kickstarter-iOS/Views/Storyboards/CreditCardCell.xib
```diff
@@ -18,26 +18,27 @@
...
                 <autoresizingMask key="autoresizingMask"/>
                 <subviews>
-                    <stackView opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" alignment="center" spacing="12" translatesAutoresizingMaskIntoConstraints="NO" id="DRF-8a-Nrl">
-                        <rect key="frame" x="6" y="6" width="308" height="53"/>
+                    <stackView opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" alignment="center" spacing="18" translatesAutoresizingMaskIntoConstraints="NO" id="DRF-8a-Nrl">
+                        <rect key="frame" x="12" y="6" width="296" height="53"/>
                         <subviews>
                             <imageView userInteractionEnabled="NO" contentMode="scaleAspectFit" horizontalHuggingPriority="750" verticalHuggingPriority="251" translatesAutoresizingMaskIntoConstraints="NO" id="b6V-DZ-ihX">
-                                <rect key="frame" x="0.0" y="8.5" width="70" height="36"/>
+                                <rect key="frame" x="0.0" y="8.5" width="58" height="36"/>
```
dusi: Updating `width` causes majority of the DIFF in this XIB file (thanks Apple)

Kickstarter-iOS/Views/Storyboards/CreditCardCell.xib
```diff
@@ -55,13 +56,13 @@
                     </view>
                 </subviews>
                 <constraints>
-                    <constraint firstAttribute="trailing" secondItem="DRF-8a-Nrl" secondAttribute="trailing" constant="6" id="7xh-nj-DfR"/>
+                    <constraint firstAttribute="trailing" secondItem="DRF-8a-Nrl" secondAttribute="trailing" constant="12" id="7xh-nj-DfR"/>
```
dusi: This is the right padding of the stack view

Kickstarter-iOS/Views/Storyboards/CreditCardCell.xib
```diff
@@ -55,13 +56,13 @@
...
                     <constraint firstAttribute="trailing" secondItem="qhp-KL-uM9" secondAttribute="trailing" id="IT1-1x-kYH"/>
                     <constraint firstItem="qhp-KL-uM9" firstAttribute="top" secondItem="DRF-8a-Nrl" secondAttribute="bottom" constant="6" id="LlQ-rl-2uz"/>
                     <constraint firstAttribute="bottom" secondItem="qhp-KL-uM9" secondAttribute="bottom" constant="-0.5" id="MKp-2I-ihI"/>
                     <constraint firstItem="DRF-8a-Nrl" firstAttribute="top" secondItem="eaL-oG-LUu" secondAttribute="top" constant="6" id="TkB-SH-Xh3"/>
                     <constraint firstItem="qhp-KL-uM9" firstAttribute="leading" secondItem="eaL-oG-LUu" secondAttribute="leading" id="UzA-dc-x3a"/>
-                    <constraint firstItem="DRF-8a-Nrl" firstAttribute="leading" secondItem="eaL-oG-LUu" secondAttribute="leading" constant="6" id="aUd-nJ-OF6"/>
+                    <constraint firstItem="DRF-8a-Nrl" firstAttribute="leading" secondItem="eaL-oG-LUu" secondAttribute="leading" constant="12" id="aUd-nJ-OF6"/>
```
dusi: This is the left padding of the stack view

Kickstarter-iOS/Views/Storyboards/CreditCardCell.xib
```diff
@@ -18,26 +18,27 @@
                 <rect key="frame" x="0.0" y="0.0" width="320" height="65.5"/>
                 <autoresizingMask key="autoresizingMask"/>
                 <subviews>
-                    <stackView opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" alignment="center" spacing="12" translatesAutoresizingMaskIntoConstraints="NO" id="DRF-8a-Nrl">
-                        <rect key="frame" x="6" y="6" width="308" height="53"/>
+                    <stackView opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" alignment="center" spacing="18" translatesAutoresizingMaskIntoConstraints="NO" id="DRF-8a-Nrl">
```
dusi: We also need to slightly update spacing that was affected by the image view stretching too much

dusi:
[Payment Methods 🐛 - Payment Method Cell right padding](https://trello.com/c/tTbqGRdp/1199-payment-methods-%F0%9F%90%9B-payment-method-cell-right-padding)

Kickstarter-iOS/Views/Storyboards/CreditCardCell.xib
ifbarrera: @dusi I wonder if it might make more sense to pin the edges of the `stackView` to the `layoutMargins` of the `contentView`, and then programmatically set the `layoutMargins` to be the correct padding values?

dusi: We can use `layoutMargins` (on iOS 10) and `directionalLayoutMargins` on `iOS 11+` (see the docs https://developer.apple.com/documentation/uikit/uiview/1622566-layoutmargins.

Is there a specific reason we'd prefer to use `layoutMargins` as oppose to auto layout? If it's only so that we can expose this in code I'd be probably more in favour of going with this solution since the rest of this UI is still defined in the `CreditCardCell.xib`

PR #583, opened 2019-02-09
🏰 Update playgrounds to Swift 4.2

dusi:
# 📲 What

Migrates playground support files to Swift 4.2

# 🤔 Why

It's one of the first steps necessary for fixing playgrounds.

### DIFF
--- a/Kickstarter-iOS.playground/Sources/playgroundController.swift
+++ b/Kickstarter-iOS.playground/Sources/playgroundController.swift
@@ -35,7 +35,7 @@ public func playgroundControllers(device: Device = .phone4_7inch,
   -> (parent: UIViewController, child: UIViewController) {
 
     let parent = UIViewController()
-    parent.addChildViewController(child)
+    parent.addChild(child)
     parent.view.addSubview(child.view)
 
     child.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
@@ -134,7 +134,7 @@ public func playgroundControllers(device: Device = .phone4_7inch,
     child.view.backgroundColor = .white
 
     let allTraits = UITraitCollection.init(traitsFrom: [traits, additionalTraits])
-    parent.setOverrideTraitCollection(allTraits, forChildViewController: child)
+    parent.setOverrideTraitCollection(allTraits, forChild: child)
 
     return (parent, child)
 }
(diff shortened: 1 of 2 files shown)


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -130,7 +130,7 @@
 		774A76E920D841110012A71F /* BetaToolsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 774A76E820D841110012A71F /* BetaToolsViewController.swift */; };
 		774A76EE20D863EF0012A71F /* BetaTools.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 774A76AF20D83F6B0012A71F /* BetaTools.storyboard */; };
 		774A76F520D98EEF0012A71F /* BetaToolsViewControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 774A76F420D98EEF0012A71F /* BetaToolsViewControllerTests.swift */; };
-		77502FC22147D92000620BBC /* Stripe.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 775229A72146805400595846 /* Stripe.framework */; };
+		77502FC22147D92000620BBC /* Stripe.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 775229A72146805400595846 /* Stripe.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
```
dusi: Would love to get some piece of mind from whoever imported `Stripe.framework` in the first place that this is safe..if not we can live without the way we currently use playgrounds.

justinswart: I recall in the past something about having to copy the framework for use in playgrounds, wondering if this might be preferable to linking weakly?

https://m.pardel.net/using-3rd-party-frameworks-in-swift-playgrounds-fd415f6dba10
https://stackoverflow.com/a/43476755

ifbarrera: Yeah I'd say lets keep this as `required` for now. We should merge the rename changes though.

Scollaco: I agree w/ @ifbarrera . I'm still not able to make the fixed Playgrounds work properly on my machine, though.

dusi: Agree, I'd be happy to invest more time into this...I've reverted the framework changes so this PR should only fix playground support files Swift 4.2 migration.

We should approve & close & fix the actual functionality later.

PR #585, opened 2019-02-12
Make the tests pass

dusi:
# 📲 What

Should make the tests pass

# 🤔 Why

We need failing tests .. NAAAAT

borat

# 🛠 How

Please see inline comments..

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
@@ -36,9 +36,14 @@ final class SelectCurrencyViewController: UIViewController, MessageBannerViewCon
 
     self.tableView.tableHeaderView = headerContainerView
     self.headerView.widthAnchor.constraint(equalTo: self.tableView.widthAnchor).isActive = true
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/Controllers/SelectCurrencyViewController.swift
```diff
@@ -36,11 +36,16 @@ final class SelectCurrencyViewController: UIViewController, MessageBannerViewCon
 
     self.tableView.tableHeaderView = headerContainerView
     self.headerView.widthAnchor.constraint(equalTo: self.tableView.widthAnchor).isActive = true
-    self.tableView.ksr_sizeHeaderFooterViewsToFit()
 
     self.viewModel.inputs.viewDidLoad()
   }
 
+  override func viewDidLayoutSubviews() {
```
dusi: I think one of the reasons we don't really need to force the header/footer to resize in `viewDidLoad` is that at that time the superview still doesn't have proper dimensions...so I think the next best place is `viewDidLayoutSubviews` which should be called when the `bounds` change (final layout to screen bounds (parent bounds) OR rotation)

dusi:
Hope this helps make the tests pass? 🤔

PR #587, opened 2019-02-12
Tableview appears without delay fix

dusi:
# 📲 What

"Should" fix the header label issue 😬

# 🤔 Why

The label was still not resizing properly after rotating the device.

# 🛠 How

Please see inline comments

# ✅ Acceptance criteria

Tested on
* iPhone SE (iOS 10.3.1, iOS 11.4, iOS 12.1)
* iPhone XS (iOS 12.1)

- [x] Payment methods screen properly sizes header & footer in portrait
- [x] Payment methods screen properly sizes header & footer in landscape
- [x] Payment methods screen properly sizes header & footer when rotating the device


Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -23,24 +23,21 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
     self.tableView.dataSource = self.dataSource
     self.tableView.delegate = self
     self.tableView.register(nib: .CreditCardCell)
-    self.tableView.registerHeaderFooter(nib: .PaymentMethodsFooterView)
-    self.tableView.registerHeaderFooter(nib: .SettingsTableViewHeader)
 
     self.configureHeaderFooterViews()
 
-    self.viewModel.inputs.viewDidLoad()
```
dusi: Looks like there were duplicates so I've remove those and moved it to the end of this method

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -139,18 +125,31 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
...
+
   private func configureHeaderFooterViews() {
-    let footerView = tableView.dequeueReusableHeaderFooterView(
-  withIdentifier: Nib.PaymentMethodsFooterView.rawValue
-      ) as? PaymentMethodsFooterView
-    footerView?.delegate = self
+    if let header = SettingsTableViewHeader.fromNib(nib: Nib.SettingsTableViewHeader) {
+      header.configure(with: Strings.Any_payment_methods_you_saved_to_Kickstarter())
+
+      let headerContainer = UIView(frame: .zero)
```
dusi: I'm not sure but think this container (that doesn't use auto-layout) was missing and is the key...If only Apple could make this work in the next OS version 🤦‍♂️

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -176,3 +175,37 @@ extension PaymentMethodsViewController: AddNewCardViewControllerDelegate {
     self.viewModel.inputs.cardAddedSuccessfully(message)
   }
 }
+
+private extension UIView {
```
dusi: I have copy pasted these from Justin's other PR and made them `private`. This way when all of our branches meet up we can simply remove these private extensions and use a global (public) extension.

Kickstarter-iOS/Views/SettingsTableViewHeader.swift
```diff
@@ -12,10 +12,11 @@ final class SettingsTableViewHeader: UITableViewHeaderFooterView {
...
     super.bindStyles()
 
-    _ = self |> \.layoutMargins .~ .init(topBottom: Styles.grid(3), leftRight: Styles.grid(2))
+    _ = self
+      |> \.backgroundColor .~ .ksr_grey_200
+      |> \.layoutMargins .~ .init(topBottom: Styles.grid(3), leftRight: Styles.grid(2))
 
     _ = self.titleLabel
       |> settingsDescriptionLabelStyle
-      |> \.preferredMaxLayoutWidth .~ self.titleLabel.bounds.width
```
dusi: This thingy was causing the label to not resize properly for me

Kickstarter-iOS/Views/SettingsTableViewHeader.swift
```diff
@@ -12,10 +12,11 @@ final class SettingsTableViewHeader: UITableViewHeaderFooterView {
   override func bindStyles() {
     super.bindStyles()
 
-    _ = self |> \.layoutMargins .~ .init(topBottom: Styles.grid(3), leftRight: Styles.grid(2))
+    _ = self
+      |> \.backgroundColor .~ .ksr_grey_200
```
dusi: Since we've decided to use `UIView` (instead of `UITableViewHeaderFooterView` it won't inherit the table view's background and so we need to set it explicitly here).

PR #588, opened 2019-02-13
[Payment methods] Refresh payment methods table view properly

dusi:
# 📲 What

Refreshes payment methods properly

# 🤔 Why

Previously if you have added a new card on iPad, the payment methods table would not refresh due to the way we present the modal. It relies on `UIViewController` lifecycle methods that are called differently when we present things using `.formSheet` modal presentation style.

# 🛠 How

Causes the view model to re-fetch on "Add New Card" modal success or dismissal.

# ✅ Acceptance criteria

on iPad..

- [x] Payment methods screen properly refreshes when first navigating to the screen
- [x] Payment methods screen properly refreshes when dismissing the `Add New Card` modal (add or remove payment methods on the web after the modal has been presented and before dismissing)
- [x] Payment methods screen properly refreshes when submitting the `Add New Card` modal (no need to do anything as this will by itself add a new payment method)

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -6,7 +6,8 @@ import Stripe
 import UIKit
 
 internal protocol AddNewCardViewControllerDelegate: class {
-  func presentAddCardSuccessfulBanner(_ message: String)
+  func addNewCardViewControllerSucceeded(with message: String)
+  func addNewCardViewControllerDismissed()
 }
 
 internal final class AddNewCardViewController: UIViewController,
@@ -212,7 +213,7 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
   }
 
   @objc fileprivate func cancelButtonTapped() {
-    self.dismiss(animated: true, completion: nil)
+    self.delegate?.addNewCardViewControllerDismissed()
   }
 
   @objc fileprivate func saveButtonTapped() {
@@ -251,9 +252,7 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
   }
 
   private func dismissAndPresentMessageBanner(with message: String) {
-    self.navigationController?.dismiss(animated: true, completion: { [weak self] in
-      self?.delegate?.presentAddCardSuccessfulBanner(message)
-    })
+    self.delegate?.addNewCardViewControllerSucceeded(with: message)
   }
 
   private func dismissKeyboard() {
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -6,8 +6,9 @@ import Stripe
 import UIKit
 
 internal protocol AddNewCardViewControllerDelegate: class {
-  func addNewCardViewControllerSucceeded(with message: String)
-  func addNewCardViewControllerDismissed()
+  func addNewCardViewController(_ viewController: AddNewCardViewController,
+                                didSucceedWithMessage message: String)
+  func addNewCardViewControllerDismissed(_ viewController: AddNewCardViewController)
 }
 
 internal final class AddNewCardViewController: UIViewController,
@@ -213,7 +214,7 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
   }
 
   @objc fileprivate func cancelButtonTapped() {
-    self.delegate?.addNewCardViewControllerDismissed()
+    self.delegate?.addNewCardViewControllerDismissed(self)
   }
 
   @objc fileprivate func saveButtonTapped() {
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -6,7 +6,8 @@ import Stripe
 import UIKit
 
 internal protocol AddNewCardViewControllerDelegate: class {
-  func presentAddCardSuccessfulBanner(_ message: String)
+  func addNewCardViewControllerSucceeded(with message: String)
```
dusi: I'm happy to discuss the naming on these

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -251,9 +252,7 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
   }
 
   private func dismissAndPresentMessageBanner(with message: String) {
-    self.navigationController?.dismiss(animated: true, completion: { [weak self] in
-      self?.delegate?.presentAddCardSuccessfulBanner(message)
-    })
+    self.delegate?.addNewCardViewControllerSucceeded(with: message)
```
dusi: Delegate should be responsible for the rest (no need to dismiss anything from here as it adds more responsibility to the controller which it shouldn't be aware of)

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -182,8 +182,18 @@ extension PaymentMethodsViewController: PaymentMethodsFooterViewDelegate {
 }
 
 extension PaymentMethodsViewController: AddNewCardViewControllerDelegate {
-  internal func presentAddCardSuccessfulBanner(_ message: String) {
-    self.viewModel.inputs.cardAddedSuccessfully(message)
+  func addNewCardViewControllerSucceeded(with message: String) {
+    self.dismiss(animated: true) { [weak self] in
+      guard let self = self else { return }
```
dusi: `guard` might not be needed, we can just optionally chain here...either should be fine

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -31,10 +31,9 @@ internal final class PaymentMethodsViewModelTests: TestCase {
   }
 
   func testPaymentMethodsFetch_OnViewDidLoad() {
-    let response = UserEnvelope<GraphUserCreditCard>(
-      me: GraphUserCreditCard.template
-    )
+    let response = UserEnvelope<GraphUserCreditCard>(me: GraphUserCreditCard.template)
```
dusi: 1 liner

dusi:
[Payment Methods 🐛 - Refresh payment methods properly on iPad when Add payment method is dismissed (cancel or success)](https://trello.com/c/qPnaZNJf/1200-payment-methods-%F0%9F%90%9B-refresh-payment-methods-properly-on-ipad-when-add-payment-method-is-dismissed-cancel-or-success)

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
justinswart: Hmm I would even consider testing this without the `weak` reference as I'm not sure this kind of closure would cause a retain cycle 🤔  but I guess it's good practice to just always do it.

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
dusi: Sounds good. At some point I was passing vc back but then decided not to for obvious reasons...I'll change this back.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
dusi: You're right! We're not holding a reference to this block so there should not be a retain cycle.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -182,8 +182,17 @@ extension PaymentMethodsViewController: PaymentMethodsFooterViewDelegate {
...
-  internal func presentAddCardSuccessfulBanner(_ message: String) {
-    self.viewModel.inputs.cardAddedSuccessfully(message)
+  func addNewCardViewController(_ viewController: AddNewCardViewController,
+                                didSucceedWithMessage message: String) {
+    self.dismiss(animated: true) {
+      self.viewModel.inputs.cardAddedSuccessfully(message)
+    }
+  }
+
+  func addNewCardViewControllerDismissed(_ viewController: AddNewCardViewController) {
```
ifbarrera: Just wondering, why do we need to refresh the table view when the "Add Card" screen is dismissed but no new card has been added? Would it not be better to only refresh the table view on on 1) `viewDidLoad` 2) when a new card has been added (via the `cardAddedSuccessfully` input).

dusi: My reasoning was that since we don't have pull to refresh or any other technique to refresh to payment methods screen (other than presenting it)...it's more useful for now to refresh both when dismissing and submitting the modal. This will for now help in situations where you'd modified your cards (on the web / other mobile client) while the modal was on and then dismissed the screen.

PR #589, opened 2019-02-13
[Payment methods] Optimistically disable edit button on card deletion

justinswart:
# 📲 What

Disables the edit button optimistically once we've deleted our last payment method, before the request has completed.

# 🤔 Why

We've used an optimistic UI approach for the rest of the behaviour on deleting payment methods and we'd like to use it for the disabling of the edit button too.

# 🛠 How

- Keeping track of cards as we delete them using `scan`.
- Filtering the users stored cards at the time of reload against the deleted cards in memory.
- Once the count of the filtered cards reaches zero, we disabled editing.
- If any error or success occurs we re-enable/keep the button disabled based on the new array of stored cards.

# ✅ Acceptance criteria

- [x] Use link conditioner to slow down your connection, delete all your cards, the button should be disabled immediately as the last card is deleted.
- [x] If the call errors, the cards should reappear and the button should be re-enabled.
- [x] Navigating to/from this view should cause the cards to refresh.


Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -67,9 +67,30 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
       paymentMethodsValues.takeWhen(deletePaymentMethodEventsErrors)
     )
 
+    let cardsDeletedSinceReload = Signal.merge(
+      self.didDeleteCreditCardSignal.map { card -> GraphUserCreditCard.CreditCard in card },
```
justinswart: I had to help the compiler along with the types here, it wasn't able to infer that this can sometimes be a `raphUserCreditCard.CreditCard` and sometimes `nil`. We use `nil` just to reset the accumulator in the `scan`.

justinswart:
[Payment methods - Optimistically disable edit button on deletion.](https://trello.com/c/jwWqVHxz/1217-payment-methods-optimistically-disable-edit-button-on-deletion)

Scollaco:
I tested using my device on flight mode, but the button doesn't re-enable if there's an error deleting the last card:

![file]

justinswart:
@Scollaco oh thanks i'll check that out 👍

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -197,6 +197,70 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
+        self?.vm.didDelete(card)
+      }
+
+      self.editButtonIsEnabled.assertValues([true, false], "Editing mode disabled, all cards deleted")
+
+      // advance scheduler to perform delete call
+      self.scheduler.advance()
+
+      self.editButtonIsEnabled.assertValues(
+        [true, false, true], "Editing mode enabled, delete response has count of 2"
```
ifbarrera: So in this scenario although all cards were initially deleted, a few calls "failed" and thats why the edit button is re-enabled, correct? How is that different from the test below?

ifbarrera: Can we add a test where the button stays disabled when all cards are successfully deleted?

justinswart:
I've pushed a new solution where:

- We pass in the visible cell count for the optimistic button enabling/disabling.
- We use the stored cards in the result from deletion so that if a deletion succeeds and a subsequent one fails, we don't show the cards from the initial load on view.

PR #590, opened 2019-02-13
Dismiss email frequency picker

cdolm92:
# 📲 What

When the Email Frequency picker is opened and then Project Activity email notifications are disabled, the `Email Frequency` cell disappears, but the picker stays open. This PR makes sure the picker dismisses when the `Email Frequency` cell disappears after Project Activity email notifications are disabled.

# 🤔 Why

Better UX, user should not be able to set the email frequency if they disabled Project Activity email notifications.

# 🛠 How

We check if user disabled the project activity email frequency and dismiss if they did. 

# 👀 See
| Before | After |
| --- | --- |
| dismiss-picker-fix |dismiss-picker-fix-after |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Navigate to `Settings > Notifications` enable email notifications for `Project Activity`, `Email Frequency` cell should appear
- [ ] Tap on `Email Frequency` cell a picker should appear
- [ ] Disable email notifications for `Project Activity` picker to dismiss

### DIFF
--- a/Library/ViewModels/SettingsNotificationsViewModel.swift
+++ b/Library/ViewModels/SettingsNotificationsViewModel.swift
@@ -96,7 +96,7 @@ SettingsNotificationsViewModelInputs, SettingsNotificationsViewModelOutputs {
 
     let projectActivityEmailFrequencyDisabled = projectActivityNotificationChanged.signal
       .map { $0.1 }
-      .filter { $0 == false }
+      .filter(isFalse)
 
     self.pickerViewIsHidden = Signal.merge(
       emailFrequencyCellSelected.signal.mapConst(false),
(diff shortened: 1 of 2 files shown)


Library/ViewModels/SettingsNotificationsViewModel.swift
```diff
@@ -94,10 +94,15 @@ SettingsNotificationsViewModelInputs, SettingsNotificationsViewModelOutputs {
       .skipNil()
       .filter { $0 == .emailFrequency }
 
+    let projectActivityEmailFrequencyDisabled = projectActivityNotificationChanged.signal
```
ifbarrera: I think you can do:
```
    let projectActivityEmailFrequencyDisabled = projectActivityNotificationChanged.signal
      .map(second)
      .filter(isFalse)
```

to be point-free :)

PR #591, opened 2019-02-13
[Payment methods] Refetch payment methods on viewDidLoad or explicitly (by delegate)

dusi:
# 📲 What

See the title

# 🤔 Why

Fixing iPad behaviour, caused duplicate calls on iPhone devices.

# 🛠 How

We've recently added the ability to re-fetch payment methods when the user returns from the `Add new payment method` modal (by submitting or dismissing). In order to consistently perform the fetch action we're using a delegate pattern (mostly because on iPad the view controller lifecycle events are called slightly different when presenting settings with `.pageSheet` presentation style). So where we previously relied on `viewWillAppear` these events worked on iPhones but not on iPads and in order to prevent from duplicate calls we now only cause fetching new data in `viewDidLoad` or explicitly by submitting & dismissing the `Add new payment method` modal or canceling the `Add new payment method` modal.

# ✅ Acceptance criteria

on iPhone & iPad

- [x] Navigate to payment methods and verify that your payment methods have been loaded
- [x] Verify that adding a new payment method (in the app) refreshes the list after modal is dismissed
- [x] Verify that removing or adding new payment method (on the web) while presenting the `Add new payment method` card refreshes the list when the modal is `canceled`

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
@@ -185,13 +185,13 @@ extension PaymentMethodsViewController: AddNewCardViewControllerDelegate {
   func addNewCardViewController(_ viewController: AddNewCardViewController,
                                 didSucceedWithMessage message: String) {
     self.dismiss(animated: true) {
-      self.viewModel.inputs.cardAddedSuccessfully(message)
+      self.viewModel.inputs.addNewCardSucceeded(with: message)
     }
   }
 
   func addNewCardViewControllerDismissed(_ viewController: AddNewCardViewController) {
     self.dismiss(animated: true) {
-      self.viewModel.inputs.refresh()
+      self.viewModel.inputs.addNewCardDismissed()
     }
   }
 }
--- a/Library/ViewModels/PaymentMethodsViewModel.swift
+++ b/Library/ViewModels/PaymentMethodsViewModel.swift
@@ -86,7 +86,7 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
 
     self.goToAddCardScreen = self.didTapAddCardButtonProperty.signal
 
-    self.presentBanner = self.addNewCardSucceededProperty.signal
+    self.presentBanner = self.addNewCardSucceededProperty.signal.skipNil()
 
     self.tableViewIsEditing = Signal.merge(
       self.editButtonTappedSignal.scan(false) { current, _ in !current },
@@ -132,7 +132,7 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.didTapAddCardButtonProperty.value = ()
   }
 
-  fileprivate let addNewCardSucceededProperty = MutableProperty("")
+  fileprivate let addNewCardSucceededProperty = MutableProperty<String?>(nil)
   public func addNewCardSucceeded(with message: String) {
     self.addNewCardSucceededProperty.value = message
   }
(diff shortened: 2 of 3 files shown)


Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -185,13 +185,13 @@ extension PaymentMethodsViewController: AddNewCardViewControllerDelegate {
   func addNewCardViewController(_ viewController: AddNewCardViewController,
                                 didSucceedWithMessage message: String) {
     self.dismiss(animated: true) {
-      self.viewModel.inputs.cardAddedSuccessfully(message)
+      self.viewModel.inputs.addNewCardSucceeded(with: message)
```
dusi: Renaming the signals to better describe their nature.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -132,14 +132,14 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.didTapAddCardButtonProperty.value = ()
   }
 
-  fileprivate let cardAddedSuccessfullyProperty = MutableProperty("")
-  public func cardAddedSuccessfully(_ message: String) {
-    self.cardAddedSuccessfullyProperty.value = message
+  fileprivate let addNewCardSucceededProperty = MutableProperty("")
```
justinswart: 🤔 I see it was this way before but usually we'd not set a default string in the `MutableProperty` initializer, these defaults make it harder to track down where values are coming from.

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -243,7 +255,8 @@ internal final class PaymentMethodsViewModelTests: TestCase {
     let apiService = MockService(deletePaymentMethodResult: .failure(.invalidInput))
     withEnvironment(apiService: apiService) {
 
-      self.vm.inputs.viewWillAppear()
+      self.vm.inputs.viewDidLoad()
```
justinswart: Can we keep the `viewWillAppear` calls in these tests? As we found out while pairing, it's often a good idea to ping the view model with the correct/expected lifecycle events so that we know they won't interfere with each other based on the outputs.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -37,9 +37,9 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.reloadData = self.viewDidLoadProperty.signal
 
     let paymentMethodsEvent = Signal.merge(
-      self.viewWillAppearProperty.signal,
```
ifbarrera: I'm wondering if we should remove the `viewWillAppear` input signal completely and switch the tracking call to `viewDidLoad` or if tracking is intentionally done on `viewWillAppear` 🤔

dusi: Good catch!

We actually talked about this yesterday with Justin (should have maybe reposted on Slack - sry about that)

Currently the tracking behaviour acts differently on iPhone and iPad devices (this is due to the way iPad lifecycle events are called for certain modal presentation styles - specifically the one we use `.pageSheet`?). It honestly should be fine (this is how the iOS was intended to work).

We can definitely switch to `viewDidLoad` or think about different ways to track these events but I’d suggest we do this thorough the whole app since this is not the only place that’s affected.

dusi: ☝️☝️☝️ Cc @Scollaco @cdolm92 for more transparency

PR #594, opened 2019-02-14
[Payment Methods] Add New Card Design Fixes

cdolm92:
# 📲 What

Fixes:

- Left Inset on divider lines
- Inconsistent margin line
- Zip/postal code return key was changed to `Done`

# ⏰ TODO

- [x] Need feedback on a design for error message and zipcode placeholder


Kickstarter-iOS/Views/Storyboards/Settings.storyboard
```diff
@@ -799,11 +799,11 @@
...
                                                     <constraint firstItem="gOu-68-PJO" firstAttribute="top" secondItem="0rS-r2-TWS" secondAttribute="bottom" id="jjv-ZY-867"/>
                                                     <constraint firstAttribute="trailing" secondItem="gOu-68-PJO" secondAttribute="trailing" id="lRG-Pt-sFA"/>
+                                                    <constraint firstItem="gOu-68-PJO" firstAttribute="leading" secondItem="BMB-LZ-IRH" secondAttribute="leading" constant="12" id="tNK-ou-V5B"/>
                                                     <constraint firstItem="ygp-ig-bdL" firstAttribute="top" secondItem="BMB-LZ-IRH" secondAttribute="top" id="ufK-tN-Ge0"/>
-                                                    <constraint firstItem="ygp-ig-bdL" firstAttribute="leading" secondItem="BMB-LZ-IRH" secondAttribute="leadingMargin" id="yXd-us-5i8"/>
+                                                    <constraint firstItem="ygp-ig-bdL" firstAttribute="leading" secondItem="BMB-LZ-IRH" secondAttribute="leading" constant="12" id="wE1-ah-iRT"/>
```
ifbarrera: Hmm can we keep this pinned to the `layoutMargin` and just adjust the `layoutMargins` on the stack view in `AddNewCardViewController`? That way it's all in code instead of IB.

ifbarrera: I think this could prob be part of a greater sweep around using `layoutMargins` in settings though, since I think we're using them inconsistently in various places.

cdolm92: I agree, this is going to have to be a part of a greater sweep. I did try to make the suggested adjustments but was not getting the desired functionality.

PR #595, opened 2019-02-14
[Payment methods] Expiration date bug fix

Scollaco:
# 📲 What
Fixes a bug that shows the wrong expiration date on the `CreditCardCell`.

# 🤔 Why
Because we should show the correct expiration date and avoid problems 😬 

# 🛠 How
The formatter was using the `UTCTimeZone` to convert the date string to `Date`, but was using the Environment's timezone to format and convert back the Date to String. By using `UTCTimeZone` in both `Format.date(from: dateFormat: timeZone:)` functions solved the problem. (Good catch, @ifbarrera ).

# ✅ Acceptance criteria

- [x] I add a new card a chosen expiration date.
- [x] The Credit Card cell on `Payment Methods` should reflect the expiration date of the added card.


Scollaco:
[Bug: Credit Card Expiration Month in Payment Methods Screen shows up incorrectly](https://trello.com/c/N25spnuU/1222-bug-credit-card-expiration-month-in-payment-methods-screen-shows-up-incorrectly)

PR #596, opened 2019-02-14
[Payment Methods] Support optional card type and error handling for payment methods

ifbarrera:
# 📲 What

- Makes `type` in `GraphUserCreditCard` optional, to support unknown credit card types.
- Adds error handling for fetching payment methods.

# 👀 See

gar3tlapxi

### DIFF
--- a/KsApi/models/DeletePaymentMethodEnvelopeTests.swift
+++ b/KsApi/models/DeletePaymentMethodEnvelopeTests.swift
@@ -19,6 +19,7 @@ class DeletePaymentMethodEnvelopeTests: XCTestCase {
     let data = jsonString.data(using: .utf8)
 
     do {
+      //swiftlint:disable force_unwrapping
       let envelope = try JSONDecoder().decode(DeletePaymentMethodEnvelope.self, from: data!)
         XCTAssertEqual(envelope.totalCount, 0)
     } catch {
--- a/KsApi/models/GraphUserCreditCard.swift
+++ b/KsApi/models/GraphUserCreditCard.swift
@@ -13,11 +13,15 @@ public struct GraphUserCreditCard: Swift.Decodable {
     }
 
     public var imageName: String {
-      switch self.type {
+      guard let cardType = self.type else {
+        return "icon--generic"
+      }
+
+      switch cardType {
       case .generic:
         return "icon--generic"
       default:
-        return "icon--\(self.type.rawValue.lowercased())"
+        return "icon--\(cardType.rawValue.lowercased())"
       }
     }
   }
--- a/KsApi/models/GraphUserCreditCard.swift
+++ b/KsApi/models/GraphUserCreditCard.swift
@@ -6,7 +6,7 @@ public struct GraphUserCreditCard: Swift.Decodable {
     public private(set) var expirationDate: String
     public private(set) var id: String
     public private(set) var lastFour: String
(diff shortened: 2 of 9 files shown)


KsApi/models/GraphUserCreditCard.swift
```diff
@@ -6,18 +6,22 @@ public struct GraphUserCreditCard: Swift.Decodable {
...
-    public private(set) var type: CreditCardType
+    public private(set) var type: CreditCardType?
 
     public var formattedExpirationDate: String {
       return String(expirationDate.dropLast(3))
     }
 
     public var imageName: String {
-      switch self.type {
+      guard let cardType = self.type else {
```
dusi: We should be able to use Swift's pattern matching in order to get rid of the `guard`.

I think this should work without the need to repeat `icon--generic` twice.

```
switch self.type {
  case nil, .some(.generic):
    return "icon--generic"
  case .some(let type):
    return "icon--\(type.rawValue.lowercased())"
}
```

KsApi/models/DeletePaymentMethodEnvelopeTests.swift
```diff
@@ -19,6 +19,7 @@ class DeletePaymentMethodEnvelopeTests: XCTestCase {
     let data = jsonString.data(using: .utf8)
 
     do {
+      //swiftlint:disable force_unwrapping
```
dusi: Are you getting a Swiftlint warning here? Is this by any chance a newer version of Swiftlint than what we're using (I think we hardcoded the build script to use `0.29.2`).

Library/ViewModels/CreditCardCellViewModel.swift
```diff
@@ -37,8 +37,14 @@ CreditCardCellViewModelOutputs, CreditCardCellViewModelType {
       .map(cardImage(with:))
 
     self.cardNumberAccessibilityLabel = self.cardProperty.signal.skipNil()
-      .map { [$0.type.description, Strings.Card_ending_in_last_four(last_four: $0.lastFour)]
-        .compactMap{ $0 }.joined(separator: ", ")
+      .map { card -> String in
+        guard let cardType = card.type else {
```
dusi: The `guard` here should not be necessary...the `compactMap` makes sure to only concatenate (join) strings that are non optional.

So, unless I'm overlooking something this should be fine..

```
return [$0.type?.description, Strings.Card_ending_in_last_four(last_four: $0.lastFour)]
          .compactMap { $0 }
          .joined(separator: ", ")
```

dusi:
Tested on
iPhone XS (iOS 12.1) ✅ 
iPhone SE SIM (iOS 10.3.1) ✅ 

(tested by adding Union Pay - card that is currently unsupported - to my account through the web app [staging] environment) ... the `Payment methods` screen shows an alert. After removing the unsupported card (again through web) cards load properly.

KsApi/models/DeletePaymentMethodEnvelopeTests.swift
ifbarrera: I'm not, but that's because swiftlint is currently broken 😬 once its fixed, this line is needed

Library/ViewModels/CreditCardCellViewModel.swift
ifbarrera: Yes you're right!

PR #597, opened 2019-02-14
Payment methods

justinswart:
# 📲 What

This PR represents a larger body of work to bring payment methods to settings.

# 🤔 Why

This allows users to add/edit/delete payment methods in the app.

# 🛠 How

https://github.com/kickstarter/ios-oss/pulls?utf8=%E2%9C%93&q=is%3Apr+payment-methods+


Kickstarter-iOS/DataSources/PaymentMethodsDataSource.swift
```diff
@@ -0,0 +1,35 @@
...
+             cellClass: CreditCardCell.self,
+             inSection: 0)
+  }
+
+  override func configureCell(tableCell cell: UITableViewCell, withValue value: Any) {
+    switch (cell, value) {
+    case let (cell as CreditCardCell, value as GraphUserCreditCard.CreditCard):
+      cell.configureWith(value: value)
+    default:
+      assertionFailure("Unrecognized (cell, viewModel) combo.")
```
dusi: We sure we want to crash if something goes wrong?

dusi: Can someone double check that we're not currently testing this behaviour?

justinswart: This is in all of our data sources, there isn't really an alternative 🤷‍♂️ . Will `assertionFailure` crash in production though? Thought asserts are ignored if not running in debug.

Kickstarter-iOS/Views/Cells/CreditCardCell.swift
```diff
@@ -0,0 +1,53 @@
+import KsApi
+import Library
+import Prelude
+import ReactiveSwift
```
dusi: Looks like this import is not necessary. Maybe we can remove it in the future (should not block the release)

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -0,0 +1,299 @@
+import KsApi
+import Library
+import Prelude
+import ReactiveSwift
```
dusi: Same as above

KsApi/models/GraphUserCreditCardTests.swift
```diff
@@ -0,0 +1,107 @@
...
+                  "lastFour": "1111",
+                  "id": "2768",
+                }
+            ],
+            "totalCount": 0
+          }
+        }
+        """
+
+    let data = jsonString.data(using: .utf8)
```
justinswart: In future we should try to replace these with the non-failable initializer:
```swift
let data = Data(bytes: jsonString.utf8)
```
Then we won't need the force unwraps.

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  private let stripeErrorProperty = MutableProperty(Error?.none)
+  public func stripeError(_ error: Error?) {
+    self.stripeErrorProperty.value = error
+  }
+
+  private let stripeTokenProperty = MutableProperty<(String, String)?>(nil)
+  public func stripeCreated(_ token: String?, stripeID: String?) {
+    if let token = token, let stripeID = stripeID {
```
justinswart: Our inputs should not contain any logic.

KsApi/models/DeletePaymentMethodEnvelopeTests.swift
```diff
@@ -0,0 +1,50 @@
...
+                ]
+              }
+            }
+        }
+      }
+      """
+    let data = jsonString.data(using: .utf8)
+
+    do {
+      //swiftlint:disable force_unwrapping
```
dusi: With a lot of these disables we do not re-enable them after the interested scope. This applies to some of the other rules in this PR. Should we use `<disable><enable>` or `<disable-next-line>`?

PR #599, opened 2019-02-15
Swiftlint fix

Scollaco:
# 📲 What
- Removes the line of code that breaks swiftlint on the root project.

# 🤔 Why
- Broken swiftlint is no good.

# 🛠 How
- Removing the `included` section from `.swiftlint.yml`. This section explicitly included `bin/ColorScript`, but it's unecessary, since swiftlint already includes subprojects, but as a side effect, took precedence over the files of our root project, so swiftlint wouldn't warn us about violations.


justinswart:
And with this change we're sure it's still linting `bin/ColorScript`?

Scollaco:
> And with this change we're sure it's still linting `bin/ColorScript`?

Yeah... Swiftlint should include subprojects. I added a commit with a violation just to make sure (and to serve as a proof) that circle would fail.

ifbarrera:
@Scollaco looks like swiftlint is failing on Circle, can we fix the violations so we can merge this?

PR #602, opened 2019-02-16
Edit button title and enabled state handling

justinswart:
# 📲 What

- Improves the way that we handle the edit button and table view's enabled state
- Toggled the button title between `Edit` and `Done`.

# 🤔 Why

The enabled state of the button and the tableview's `isEditing` property were handled by a merged signal:
```swift
    self.tableViewIsEditing = Signal.merge(
      self.editButtonIsEnabled.filter(isFalse),
      self.editButtonTappedSignal.scan(false) { current, _ in !current },
      self.didTapAddCardButtonProperty.signal.mapConst(false)
    )
```
This was problematic because the `scan` on `editButtonTappedSignal`  would maintain its own booelan toggle independent of the signals that it was merged with causing them to walk over each other. One effect of this (shown below) is that if we tap the _add card_ button while the table view is in editing, upon returning this screen we will need to tap `Edit` twice before it re-enables editing.

The button's title remained `Edit` despite its state and we'd prefer that it was `Done` providing a cue to the user that they can tap the button again to end editing.

# 🛠 How

- Improved the signal to enable/disable `isEditing` by storing the last value in a `MutableProperty<Bool>`. Tapping edit will toggle whatever the latest value is in this property and store the new value there.
- Added a new output to emit the button title.

# 👀 See

| Befre | After |
|  |  |

# ♿️ Accessibility 

- [x] Works with VoiceOver

# ✅ Acceptance criteria

- [x] Tap _edit_, title should change to Done, tap again, it should change back to Edit.
- [x] Tap _add card_ during editing, tableview state should end editing and title should change back to edit.
- [x] Repeat the last step and tap edit after navigating back from _add card_, only one tap should be required to enter edit mode again.


justinswart:
[[Payment methods] Edit button should only require one tap after adding card](https://trello.com/c/66QPxiPJ/751-payment-methods-edit-button-should-only-require-one-tap-after-adding-card)

justinswart:
[[Payment methods] Edit button should toggle between Edit and Done](https://trello.com/c/28UpSQMi/745-payment-methods-edit-button-should-toggle-between-edit-and-done)

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -122,6 +122,12 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
       .observeValues { [weak self] message in
         self?.present(UIAlertController.genericError(message), animated: true)
     }
+
+    self.viewModel.outputs.editButtonTitle
+      .observeForUI()
+      .observeValues { [weak self] title in
+        self?.editButton.title = title
```
Scollaco: For consistency, could we use
```
_ = self?.editButton
          ?|> \.title %~ { _ in title }
```
here?

dusi: The documentation mentioned kinda vague fact that the `title` should be set before adding the item to the bar - https://developer.apple.com/documentation/uikit/uibaritem/1616412-title which makes me wonder if we should actually re-create the `UIBarButtonItem` but with a different title?

Something like:
```
private func editButton(_ title: String) -> UIBarButtonItem {
    return UIBarButtonItem(
      title: title,
      style: .plain,
      target: self,
      action: #selector(edit)
    )
  }
```

dusi: Not sure if this is related to the bug that I've reported on the Trello card 🤷‍♂️

justinswart: Ah interesting, I could try that, thanks 👍

justinswart: I wasn't able to reproduce the truncation issue but I've tried a property called `possibleTitles`, first I've heard of it - https://developer.apple.com/documentation/uikit/uibarbuttonitem/1617132-possibletitles

Let me know if you're still seeing it.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -37,6 +37,10 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
     self.configureHeaderFooterViews()
 
     self.navigationItem.rightBarButtonItem = self.editButton
+    self.editButton.possibleTitles = [
```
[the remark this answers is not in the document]
justinswart: lol same

dannyalright:
> This is very minor, but is it possible to debounce the signal that triggers `stopEditing` when you tap the "Add New Card" button so it stops editing some time after the `Add Card` screen has been presented? I find it visually disorienting to see the table stop editing right when you tap `Add new card` 🤔 cc @dannyalright

Just so we're on the same page, we're talking about behaviour like in the [iOS Alarm app](https://drive.google.com/file/d/12AFfHQLCKqiHfR-9U-He4-Wsz7Rmr5ry/view?usp=sharing), right?

Doesn't bother me too much, but would appreciate it if it's not too hard

justinswart:
@dannyalright @ifbarrera made that change:

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewController.swift
```diff
@@ -147,7 +147,7 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
     let nav = UINavigationController(rootViewController: vc)
     nav.modalPresentationStyle = .formSheet
 
-    self.present(nav, animated: true, completion: nil)
+    self.present(nav, animated: true) { self.viewModel.inputs.addNewCardPresented() }
```
justinswart: Doesn't cause a retain cycle because we're not holding a reference to the closure 👍

PR #603, opened 2019-02-19
Makefile cleanup

ifbarrera:
# 📲 What

- fixes `make sync` command to sync private and oss repos
- defaults the release to `itunes` since we no longer manually upload betas. This enables us to just run `make deploy` when doing App Store releases


Makefile
```diff
@@ -4,7 +4,7 @@ BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION)
 SCHEME ?= $(TARGET)-$(PLATFORM)
 TARGET ?= Kickstarter-Framework
 PLATFORM ?= iOS
-RELEASE ?= beta
+RELEASE ?= itunes
```
justinswart: In our `deploy:` task we're doing checks against `beta` wanna keep those?

ifbarrera: Will remove 👍

PR #605, opened 2019-02-20
♿️ [Dynamic Type] Navigation bar title

dusi:
# 📲 What

Uses default (system) font for the navigation bar title in order to prevent from allowing resizing when the font size changes.

# 🤔 Why

More native behaviour.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| screen shot 2019-02-19 at 9 00 55 pm | screen shot 2019-02-19 at 8 59 09 pm |

# ♿️ Accessibility 

- [ ] Changing font size does not affect navigation bar title anywhere on the primary navigation stack (anything under tab bar or presented modally)
- [ ] Changing font size does not affect navigation bar title on KSR Live

### DIFF
--- a/Library/Styles/BaseStyles.swift
+++ b/Library/Styles/BaseStyles.swift
@@ -144,8 +144,7 @@ private let navBarLens: Lens<UINavigationController?, UINavigationBar?> = Lens(
 
 private let baseNavigationBarStyle =
   UINavigationBar.lens.titleTextAttributes .~ [
-    NSAttributedString.Key.foregroundColor: UIColor.black,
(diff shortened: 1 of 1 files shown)


ifbarrera:
@dusi should we also disable dynamic type for the `Explore` custom navigation title and the search bar placeholder?

image

image

dusi:
Yes, I was thinking about those as a separate piece of work since that's a custom component, not a navigation bar style.

If we'd prefer to have that work as part of this PR I would have to properly create a card and add it to our sprint (this fix came out of trying to figure out what's wrong and due to its triviality I've opened a PR hoping this will make it easier in the future to take on that work).

PR #606, opened 2019-02-20
🔑 [Create Password] Account Screen

Scollaco:
# 📲 What
Adds a `Create Password` row on `Settings > Account` if user hasn't created a password yet.

# 🤔 Why
In order to change email / password, users have to confirm these actions using their password. Accounts created with Facebook don't have password initially, thus users are not able to change email. With this new feature, we are allowing Facebook accounts to change email/password in the future.

# 🛠 How
Instead of simple hiding the `Change Email / Password` section if `hasPassword == false`, we now replace it with a `Create Password` section.
A footer was also created with a label explaining why one should create a password. 

In order to show the user's email on the footer, I needed to add `.email` to the UserAccountFields query, and passed this value to the footer's `configure(with email:)` function.

# 👀 See

| With Password | No Password |
|---|---|
|  |  | 
## Dynamic Type (Footer)

# ♿️ Accessibility 
- [x] By tapping the Create Password cell, I should hear VoiceOver say `Create Password, button`
- [x] By tapping the footer, I should hear VoiceOver read the text (including my email)

# ✅ Acceptance criteria

- [x] I'm a user with password and go to `Settings > Account`. I should see the `Change Email / Password` section followed by `Privacy`, `Payment methods` and `Currency`
- [x] I'm a user with NO password and go to `Settings > Account`. I should see the `Create password` section followed by `Privacy`, `Payment methods` and `Currency`
- [x] The footer's text should say: `You're connected via Facebook (your_email). Create a password for this account to edit your Kickstarter credentials.`


Kickstarter-iOS/DataSources/SettingsAccountDataSource.swift
```diff
@@ -11,7 +11,7 @@ final class SettingsAccountDataSource: ValueCellDataSource {
 
     self.filteredSections = shouldHideEmailPasswordSection
       ? SettingsAccountSectionType.allCases.filter { $0 != .emailPassword }
```
dusi: Wonder if we should change `emailPassword` case to something more descriptive (just like `createPassword`)..i.e. `changeEmailPassword`?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -199,13 +200,31 @@ extension SettingsAccountViewController: UITableViewDelegate {
   }
 
   func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
-    return 0.1
+
```
dusi: Has there been a consensus if we prefer to open functions, etc. with empty line or is this mostly a personal preference?

Kickstarter-iOS/Views/CreatePasswordFooterView.swift
```diff
@@ -0,0 +1,27 @@
+import Prelude
+import Library
+
+internal final class CreatePasswordFooterView: UITableViewHeaderFooterView, NibLoading {
+
+  @IBOutlet private weak var descriptionLabel: UILabel!
+
+  func configure(with email: String?) {
+    guard let email = email else {
```
dusi: Should this check happen somewhere else maybe? Wouldn't it make reasoning about this UI component a bit simpler if all it took was non-optional string and its sole purpose would be to display it? I'm not sure if a UI component should fail if it's being passed wrong data (it should instead not allow wrong data from being passed to it).

So instead I'd suggest we do `func configure(with email: String) {...}` and handle the optional unwrapping somewhere else.

What do you think Nino?

Library/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -147,6 +151,11 @@ SettingsAccountViewModelOutputs, SettingsAccountViewModelType {
     self.viewDidLoadProperty.value = ()
   }
 
+  fileprivate let userHasPasswordAndEmailProperty = MutableProperty<(Bool, String?)>((true, nil))
```
dusi: Should we test this?

dusi: Also instead of using default value `(true, nil)`, would it be safer to use optional tuple that would not be initialized like this?

So instead of
```
fileprivate let userHasPasswordAndEmailProperty = MutableProperty<(Bool, String?)>((true, nil))
```
use
```
fileprivate let userHasPasswordAndEmailProperty = MutableProperty<(Bool, String)?>(nil)
```

dusi:
@dannyalright what are your thoughts on putting each sentence on a new line (maybe adding a one line space)? This is how it renders on iPhone X/XS devices with default font size which makes the text look zig-zaggy (if you know what I mean 😄 

img_7319f54ae010-1

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -199,13 +200,31 @@ extension SettingsAccountViewController: UITableViewDelegate {
...
   }
 
+  func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
+
+    let (userHasPassword, email) = self.viewModel.outputs.userHasPasswordAndEmail
+    guard section == SettingsAccountSectionType.createPassword.rawValue && !userHasPassword else {
+      return nil
+    }
+
+    let footerView = tableView.dequeueReusableHeaderFooterView(
```
dusi: So for the optional comment I had bellow, this is where we would not display a section if email from the signal was `nil`

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
Scollaco: It's used to add a line between the function declaration and its implementation to make the reading easier, but I've been avoiding this on our codebase. So I would say that these lines are there by accident :)

Kickstarter-iOS/Views/CreatePasswordFooterView.swift
Scollaco: That's how it was before, and I decided to make the string optional so I wouldn't need to provide a default value on the ViewModel 🤦‍♂️ . But what you are saying makes sense and now thinking about it, I guess you're right.

Library/ViewModels/SettingsAccountViewModel.swift
Scollaco: In this particular case, we are `returning` a MutableProperty value and not `assigning` a value, like we usually do for functions. So we would have to set a default value anyway, since the variable is non-Optional. Does it make sense? 
Based on your previous comment though, I would change to <(Bool, String)> instead, making the String non-optional.

What do you think?

dannyalright:
> @dannyalright what are your thoughts on putting each sentence on a new line (maybe adding a one line space)? This is how it renders on iPhone X/XS devices with default font size which makes the text look zig-zaggy (if you know what I mean 😄
> 
> img_7319f54ae010-1

Good shout. @vickisiolos are you ok with a line break between `...@gmail.com).` and `Create a password...`?

vickisiolos:
@dannyalright @dusi Would this copy rearrangement solve the alignment issue?
"Create a password for this account to edit your Kickstarter credentials. You’re currently logged in with pavel.dustatko@gmail.com via Facebook."
(Hoping to avoid a line break!)

Kickstarter-iOS/Views/CreatePasswordFooterView.swift
```diff
@@ -0,0 +1,23 @@
+import Prelude
+import Library
+
+internal final class CreatePasswordFooterView: UITableViewHeaderFooterView, NibLoading {
```
dusi: I forgot to mention that since this is a pretty generic view I might be able to re-use it very very soon. Would you mind renaming this to `SettingsHeaderView` and injecting the `descriptionLabel` text instead (through `configure(with email: String, description: String)`)? 🙏

Scollaco: It sounds good! 👍

dannyalright:
@dusi: @vickisiolos and I played around with line-breaks and formatting. We've decided to keep the copy as-is. Thanks for suggesting, though!

Library/ViewModels/SettingsAccountViewModel.swift
dusi: Thanks for the explanation Nino! It makes perfect sense now 😄 

giphy

Kickstarter-iOS/Library/Nib.swift
```diff
@@ -3,6 +3,7 @@ import UIKit
 public enum Nib: String {
   case BackerDashboardEmptyStateCell
   case BackerDashboardProjectCell
+  case CreatePasswordFooterView
```
dusi: Do we still need this class if we're going to use `SettingsGroupedFooterView`?

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -106,12 +106,32 @@ extension SettingsAccountViewController: UITableViewDelegate {
   }
 
   func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
-    return 0.1
+    let (userHasPassword, _) = self.viewModel.outputs.userHasPasswordAndEmail
+    guard section == SettingsAccountSectionType.createPassword.rawValue, !userHasPassword else {
+      return 0.1
+    }
+    return Styles.grid(10)
```
dusi: Looks like this breaks Dynamic Type support. If we set this directly we won't allow the footer to resize itself. I believe a better approach would be to do the following:

```suggestion
    return UITableView.automaticDimension
```

Which will allow the height to be dynamic..and in addition to that we also should set `estimatedSectionFooterHeight` in order to give it at least a bit of information. Isabel previously suggested this is a `static var` on the footer class itself so we could do:

```
final class SettingsGroupedFooterView: UITableViewHeaderFooterView {
  static var defaultHeight: CGFloat = 44
  ...
}
```

Kickstarter-iOS/Views/Controllers/SettingsAccountViewController.swift
```diff
@@ -30,7 +30,7 @@ final class SettingsAccountViewController: UIViewController, MessageBannerViewCo
     self.tableView.register(nib: .SettingsTableViewCell)
     self.tableView.register(nib: .SettingsAccountWarningCell)
     self.tableView.registerHeaderFooter(nib: .SettingsHeaderView)
-
+    self.tableView.registerHeaderFooterClass(SettingsGroupedFooterView.self)
```
dusi: Can we keep the space between registering reusable views and calling `viewDidLoad` on the view model?

Kickstarter-iOS/Library/Nib.swift
Scollaco: Actually I forgot to do the cleanup. Will delete the unused files 👍

Library/Styles/SettingsStyles.swift
```diff
@@ -86,7 +86,15 @@ public func settingsContentViewStyle(_ view: UIView) -> UIView {
     |> \.preservesSuperviewLayoutMargins .~ false
 }
 
-public let settingsFooterContentViewStyle = settingsContentViewStyle
+public func settingsFooterContentViewStyle(_ view: UIView) -> UIView {
+  return view
+  |> settingsContentViewStyle
```
dusi: Can we adjust the spacing here? CMD+X > CMD+V

Scollaco: Or Ctrl+I 😄

dusi: Wuuuuuuuuuuuuuuuut :trollface:

PR #608, opened 2019-02-20
♿️ [Voice Over] Settings > Help Accessibility Traits

dusi:
# 📲 What

Uses link accessibility trait for each row that presents a website.

# 🤔 Why

This should give visually impaired users more context to where is this action going to take them.

# 🛠 How

Simply decide on the row type and return either `.link` for actions that navigate the user to a website or `.button` for presenting the mail composer to contact support.

# ♿️ Accessibility 

navigating to `Settings > Help`, VoiceOver reads the following cells like so

- [ ] `Help Center, link`, 
- [ ] `Contact, button`
- [ ] `Terms of Use, link`
- [ ] `Privacy Policy, link`
- [ ] `Cookie Policy, link`

### DIFF
--- a/Library/HelpTypeTests.swift
+++ b/Library/HelpTypeTests.swift
@@ -1,0 +1,14 @@
+import XCTest
+import Library
+
+final class HelpTypeTests: XCTestCase {
+  func testAccessibilityTraits() {
+    XCTAssertEqual(HelpType.helpCenter.accessibilityTraits, UIAccessibilityTraits.link)
+    XCTAssertEqual(HelpType.contact.accessibilityTraits, UIAccessibilityTraits.button)
+    XCTAssertEqual(HelpType.howItWorks.accessibilityTraits, UIAccessibilityTraits.link)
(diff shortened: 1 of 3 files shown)


Library/HelpTypeTests.swift
```diff
@@ -0,0 +1,14 @@
+import XCTest
+import Library
+
+final class HelpTypeTests: XCTestCase {
+  func testAccessibilityTraits() {
+    XCTAssertEqual(HelpType.helpCenter.accessibilityTraits, UIAccessibilityTraits.link)
```
dusi: I've provided tests for all the cases even though we don't currently show `howItWorks` and `trust` which I believe show a web view as well.

dusi:
[♿️ Settings > Links](https://trello.com/c/fNIY56qH/1242-%E2%99%BF%EF%B8%8F-settings-links)

Library/HelpTypeTests.swift
cdolm92: Nice, thanks for adding!

PR #609, opened 2019-02-21
Change password keyboard dismissal

Scollaco:
# 📲 What
This PR fixes a bug on which, while on `Change Password` screen, the keyboard would cover the MessageBanner depending on what textField was the first responder.

# 🛠 How
Our current logic only dismisses the `confirmNewPasswordTextField`. This works only if this TextField is the firstResponder. 
The new logic identifies which textField is editing and then dismiss it. 

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
| dismiss_keyboard_bug_before | dismiss_keyboard_bug|

# ✅ Acceptance criteria
-  I go to `Settings Account > Change Password` and enter information that will cause an error (e.g. incorrect current password).

- [x] Tapping `Save` while `Current Password` is first responder should dismiss keyboard before showing Message Banner.
- [x] Tapping `Save` while `New Password` is first responder should dismiss keyboard before showing Message Banner.
- [x] Tapping `Save` while `Confirm new Password` is first responder should dismiss keyboard before showing Message Banner.


Scollaco:
[Keyboard does not dismiss on _Change password_ `Save`](https://trello.com/c/MIff5ZKv/1239-keyboard-does-not-dismiss-on-change-password-save)

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -211,6 +211,13 @@ final class ChangePasswordViewController: UIViewController, MessageBannerViewCon
     }
   }
 
+  private func dismissKeyboard() {
+    _ = [self.newPasswordTextField, self.confirmNewPasswordTextField, self.currentPasswordTextField]
```
justinswart: Would it be a problem to just blanket ping all of the textFields to resign?

```swift
[self.newPasswordTextField, self.confirmNewPasswordTextField, self.currentPasswordTextField]
  .forEach { $0.resignFirstResponder() }
```
Also `resignFirstResponder()` returns `Void` so no need for the `_ =` to discard the return value.

Scollaco: Not at all. I can change that.

PR #610, opened 2019-02-21
♿️Change Password Validation Message Focus

ifbarrera:
# 📲 What

Brings voice-over focus to the validation error message when it changes.

Also in this PR:
- cleaned up constraints on the Change Password IB scene, converted all leading, trailing, top, bottom constraints to `margin` and set the margins on the containing `UIStackView`
- keyboard now dismisses when you hit the `return` key on the last field, regardless of whether the form is valid or not

# 👀 See

| Before 🐛 | After 🦋 |
| -- | --  |
|simulator screen shot - iphone 8 - 2019-02-21 at 15 31 42|simulator screen shot - iphone x - 2019-02-21 at 15 23 18|

**Landscape iPhone 7 (11.4)**
simulator screen shot - iphone 7 - 2019-02-21 at 15 20 34

**iPhone 5 (10.3.1)**
simulator screen shot - iphone 5s - 2019-02-21 at 15 21 43

# ✅ Acceptance criteria

- [x] Navigate to Change Password. Turn voice over on. Enter current password, then enter new password (less than 6 chars long), then enter different text in "Confirm New Password". Voice over should focus on error message with the correct message ("new password must be > 6 chars long")
- [x] Navigate to Change Password. Turn voice over on. Enter current password, then enter new password (more than 6 chars long), then enter different text in "Confirm New Password". Voice over should focus on error message with the correct message ("password must match")

Issues to look out for:
- error message should only come into focus when the message is different from the previous message
- error message should not come into focus when it's hidden


ifbarrera:
[♿️ Settings > Account > Change password - Passwords matching](https://trello.com/c/WcI7lEdY/1158-%E2%99%BF%EF%B8%8F-settings-account-change-password-passwords-matching)

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -53,6 +54,9 @@ final class ChangePasswordViewController: UIViewController, MessageBannerViewCon
     _ = self.scrollView
       |> \.alwaysBounceVertical .~ true
 
+    _ = self.stackView
```
dusi: Should we also set `preservesSuperviewLayoutMargins` to `false`?

Looking at this in SIMulator, the horizontal padding looks smaller than `12pts`. Using visual inspector actually shows `8pts`.

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -174,6 +178,13 @@ final class ChangePasswordViewController: UIViewController, MessageBannerViewCon
         self?.logoutAndDismiss()
     }
 
+    self.viewModel.outputs.accessibilityFocusValidationErrorLabel
+      .observeForUI()
+      .observeValues { _ in
```
dusi: `[weak self]` to retain cycle?

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
```diff
@@ -174,6 +178,13 @@ final class ChangePasswordViewController: UIViewController, MessageBannerViewCon
...
+        UIAccessibility.post(notification: .layoutChanged,
+                             argument: self.validationErrorMessageLabel)
```
dusi: Also would this not fit on one line?

Kickstarter-iOS/Views/Controllers/ChangePasswordViewController.swift
ifbarrera: Hmm I think the default is false. I'll investigate 🕵️‍♀️

ifbarrera: Ok the issue is that `preservesSuperviewLayoutMargins` needs to be set to `true` on all the `stackView`'s arranged subviews. This is because they're all container views which hold the label and the text field.

dusi: That's pretty unfortunate. I guess this is where the concept of using `UITableViewCell` and its `contentView` would be helpful (in a sense that we'd only need to set the margins for the content view).

Good find Isabel!

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -128,8 +129,13 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
...
         }
-    }.skipNil()
-    .skipRepeats()
+    }
+
+    self.validationErrorLabelMessage = errorLabelMessage
+      .skipNil()
+      .skipRepeats()
+
+    self.accessibilityFocusValidationErrorLabel = self.validationErrorLabelMessage.ignoreValues()
```
dusi: I totally forgot to mention that we should only trigger this signal if `VoiceOver` is turned ON (for consistency with the rest of the code base)

```
.filter { AppEnvironment.current.isVoiceOverRunning() }
```

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -128,8 +131,19 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
...
+      .skipNil()
+      .skipRepeats()
+
+    let entryChanged = Signal.merge(self.newPasswordProperty.signal, self.confirmNewPasswordProperty.signal)
+
+    self.accessibilityFocusValidationErrorLabel = requirementsMet
+      .takeWhen(entryChanged)
+      .filter(isFalse)
+      .ignoreValues()
+      .filter { _ in AppEnvironment.current.isVoiceOverRunning() }
```
dusi: Would it make sense to `filter` whether VoiceOver is running before any other operation is done on the signal? Because if VoiceOver is running we're still doing calculations just to discard it later.

So instead of
```
self.accessibilityFocusValidationErrorLabel = requirementsMet
  .takeWhen(entryChanged)     
  .filter(isFalse)
  .ignoreValues()
  .filter { _ in AppEnvironment.current.isVoiceOverRunning() }
```

do
```
self.accessibilityFocusValidationErrorLabel = requirementsMet
  .takeWhen(entryChanged)
  .filter { _ in AppEnvironment.current.isVoiceOverRunning() }
  .filter(isFalse)
  .ignoreValues()
```

PR #504, opened 2018-12-02
FunSwift2018: Reactive view models, simplified

stephencelis:
:wave: Hey squad!

I gave a talk yesterday at Functional Swift Conference in which I live-refactored a Kickstarter view model and controller. @ifbarrera attended and suggested that I open a PR with the result. We've been tweaking the reactive view model approach a bit the past coupla years and have been wanting to share.

Here's what the end result looks like:

https://github.com/stephencelis/ios-oss/blob/40cfb4da8afd23ea4d4cea637fbab161c238839d/Library/ViewModels/SignupViewModel.swift

Main benefits we've seen:

- Addresses the big boilerplate problem. Instead of containing view model logic using a bunch of different constructs (including multiple protocols, a class/struct, an initializer, a bunch of properties, a bunch of input functions that delegate directly to private properties), we can use a single construct: a single function. Half of the implementation boilerplate goes away and things become a bit easier to digest as a result

- Another benefit of this style is that we get unused warnings for outputs not used in the view controller (one of which I found during the refactor).

Note that this PR adds ReactiveCocoa as a dependency for less verbose binding in the view controller (avoids the extra target-action boilerplate). This isn't necessary, so take it or leave it, but after using RxSwift+RxCocoa on a project, it's been pretty nice to avoid the extra work and it made my talk go by a bit quicker.

Not in this PR: updated tests! Some (but definitely not all) of the boilerplate lost in the main view model reappears in the tests since you need to manually create mutable properties in the test case, but it seems like a nicer trade-off to hide this boilerplate away in tests. For example:

``` diff
 internal final class SignupViewModelTests: TestCase {
-  fileprivate let vm: SignupViewModelType = SignupViewModel()
+  fileprivate let emailChangedProperty = MutableProperty("")
+  fileprivate let emailTextFieldReturnProperty = MutableProperty(())
+  fileprivate let environmentLoggedInProperty = MutableProperty(())
+  fileprivate let nameChangedProperty = MutableProperty("")
+  fileprivate let nameTextFieldReturnProperty = MutableProperty(())
+  fileprivate let passwordChangedProperty = MutableProperty("")
+  fileprivate let passwordTextFieldReturnProperty = MutableProperty(())
+  fileprivate let signupButtonPressedProperty = MutableProperty(())
+  fileprivate let viewDidLoadProperty = MutableProperty(())
+  fileprivate let weeklyNewsletterChangedProperty = MutableProperty<Bool>(nil)
+
[...]

PR #612, opened 2019-02-22
Signup View Model Refactor

ifbarrera:
# 📲 What

Following a great talk at the Swift Functional Conf by @stephencelis on refactoring our view models, I decided to take a crack at this by taking some elements of the refactor from Stephen’s talk, while also attempting to reduce the impact on view model tests. I also wanted to see how it might look without bringing in `ReactiveCocoa` as a dependency (something we should evaluate separately).

# 🛠 How

View model input signal/observer pairs are stored as properties within an `Inputs` struct in the view model. They are immutable, and are created using `Signal.pipe()`. They are publicly accessible, which means observers can be sent values using `observer.send(value: value)` from view controllers as needed. The inputs are fed as an argument into the `func outputs(from inputs: Inputs)` function, which returns a type-aliased `Outputs` tuple. All the logic happens within this function.

With this approach, the tests don’t really change much (no additional boilerplate) which is nice, and the usage in the ViewController is also very similar. In `bindViewModel`, you call the `outputs(from:)` function and then use the outputs as needed within `bindViewModel`.

Most of the benefits of this overall approach are described in [Stephen’s PR](https://github.com/kickstarter/ios-oss/pull/504), but some additional benefits in *this* PR include:
- Doesn’t require Reactive Cocoa dependency
- Don’t need to store reactive properties in the view controller, they stay in the view model
- The tests don’t get any more verbose: can reuse the inputs from the view model in the tests
- gets rid of `MutableProperty`, which saves us from having to initialize them with initial values, plus cuts down on the verboseness of having to call `property.signal`

Drawbacks:
- To refactor the view model tests, one must change all calls to input functions to instead call `send(value: value)`. This is a bit tedious because all the old input functions have different names so a simple search & replace won’t work here.

All in all… this refactor nets approx. -48 lines of code! Not bad.

# 👀 See

Signup works as usual:
tqitputjdn

# ✅ Acceptance criteria

- [ ] Sign-up form works as expected


Library/ViewModels/SignupViewModel.swift
```diff
@@ -4,108 +4,85 @@ import Prelude
...
+    public let (passwordTextFieldDidReturnSignal, passwordTextFieldDidReturnObserver) =
+      Signal<(), NoError>.pipe()
+    /// Call when the user taps signup.
+    public let (signupButtonPressedSignal, signupButtonPressedObserver) = Signal<(), NoError>.pipe()
+    /// Call when the view did load.
+    public let (viewDidLoadSignal, viewDidLoadObserver) = Signal<(), NoError>.pipe()
+    /// Call when the user toggles weekly newsletter.
+    public let (weeklyNewsletterChangedSignal, weeklyNewsletterChangedObserver) = Signal<Bool, NoError>.pipe()
+
+    init() {}
```
justinswart: Strange that Swift wouldn't let us get away without this when we use the `pipe()`s!

Library/ViewModels/SignupViewModel.swift
```diff
@@ -4,108 +4,85 @@ import Prelude
...
+    postNotification: Signal<Notification, NoError>,
+    /// Sets whether the name text field is the first responder.
+    nameTextFieldBecomeFirstResponder: Signal<(), NoError>,
+    /// Emits the value for the weekly newsletter.
+    setWeeklyNewsletterState: Signal<Bool, NoError>,
+    /// Emits when a signup error has occurred and a message should be displayed.
+    showError: Signal<String, NoError>
+  )
+
+  public let inputs = Inputs()
```
justinswart: Why do we have this local `let` again?

ifbarrera: We store a reference to the inputs in order to send them values (via self.vm.inputs.input.send(value: value)` in both the view controller and the view model tests, but also in order to pass inputs into the `outputs(for inputs:)` function to generate our outputs.

justinswart: Oh right... why do we need to pass it in as an argument if it's a local var? Couldn't `outputs()` take no arguments and return `self.inputs`?

ifbarrera: Yep, that could also work.

justinswart: Cool, yeah I think if there's no real reason to pass it in then that makes sense? It doesn't help us with tests as far as I can tell.

haaakon:
Interesting approach. We at our company love the way you guys set up your `ViewModels`, and i was trying this new refactored way of doing them. Only caveat i found was when configuring `ViewControllers` one has to make sure the outputs function is called on the `ViewController` init, and not on `viewDidLoad`, since there can be input signals firing before a `ViewController` will get its `viewDidLoad` function run. Dunno if thats a problem in your app, but if anyone else tries this new approach its just something to have in mind :)

justinswart:
@haaakon what we’ve done in the past to work around this sort of race condition where an output may emit before a necessary lifecycle event occurs is to combine the output with that event’s input signal. So for example if we need an output to emit _after_ `viewDidLoad` is called then we would create a combined signal of the output we’re interested in and `viewDidLoad`’s input signal.

However am I misunderstanding your concern? Maybe you can share some example code that demonstrates the problem?

haaakon:
We often do 

```
let newViewController = NewViewController.init()
newViewController.inputs.configureData(myData)
present(newViewController)
```
and with this pattern, the `bindViewModel` is called in the `ViewController` `viewDidLoad()`
``` 
override func viewDidLoad() { 
      bindViewModel()
      bindStyles()
      vm.inputs.viewDidLoadObserver.send(value:())
} 

func bindViewModel() {
   let (variable1, variable2) = self.outputs() 
}
```

means 
```
func outputs() -> Outputs {
    let variable1 = Signal.combineLatest(viewDidLoadSignal,  configureDataSignal)    
    }

```
`ConfigureData` is called before this `combineLatest` is set up, since the outputs are generated at `bindViewModel()`. Then the `combineLatest` will "miss" a firing of configureData, if configureData is called immedeately when a viewController is created

justinswart:
@haaakon aah right. Yes that's a good catch, thanks. This may actually be a reason to keep using `MutableProperty<T>` in such cases 🤔 If we set the `value` on a property rather than just send values through pipes then the `MutableProperty<T>` should retain its last value and emit that when it's converted to a `SignalProducer<T, Error>`. I _believe_ you may actually want to use `.producer` and not `.signal` in this case (sorry, I'm not currently able to test it), see this for reference: https://github.com/ReactiveCocoa/ReactiveSwift/issues/620.

We like `pipe()` because it's simpler than using `MutableProperty` in that we don't need an initial `nil` value and then aren't later required have to `skipNil()` - but it does appear that in this case it could be problematic.

@ifbarrera perhaps we should pair on a few more of these edge cases when I'm back 😅

haaakon:
A quick & dirty way of fixing this is doing
```
class MyViewController() : UIViewController {
      let vm = MyViewModel()
      var outputs: MyViewModel.Outputs!
 
    override public init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
        super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
        self.outputs = vm.outputs()
    }
    
    required public init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
        self.outputs = vm.outputs()
    }
     
    func  bindViewModel() { 
        outputs.myVariable.observeValues { [weak self] value in 
               print(value)
        }
    }
```
That solves the problem for `ViewControllers` at least, but i havent gotten to `UIView` subclasses yet. We also do alot of inputs with `Property<T>` since we "stream" data in our app, so will have to do more research if that still works.

I would hope one does not need to go back to MutableProperty<T> though, its so verbose and i don't like the skipNil() on everything. This new pattern saves time, is easier to understand and cleaner in my opinion, so i will definately follow this PR to see if you have and suggestions!

ifbarrera:
After discussing with the rest of the team, we've decided that the slight reduction boilerplate code for view models with this refactor is probably not worth the inconsistency it will introduce into the codebase. However, we are still investigating ways to ensure that `bindViewModel` is called as early as possible - something we're exploring is using a protocol with a default implementation of an `instantiate()` function that calls `bindViewModel()` immediately after initialization on programmatic view controllers. Combined with a little bit of extra logic in our swizzled `ksr_viewDidLoad()` function to make sure `bindViewModel()` isn't called multiple times, we can ensure that view models are bound as early as possible to prevent possible issues with missed signals in the future.

PR #614, opened 2019-02-22
👩‍🔬 [Investment Day] Disable tracking for non-production builds

dusi:
# 📲 What

Disables tracking in `DEBUG` mode.

This means tracking will still work for AppStore, Alpha and Beta builds but won't be actively logging when run by engineers in test mode, simulator, etc.

# 🤔 Why

Currently we're using `Koala` to track events for non-production builds. As much as this hasn't been a problem for us in the past I think it has couple disadvantages for engineers:

1. Congested Xcode console logs
- These make logging / debugging much harder than it needs to be.
- Often makes us miss broken auto layout constraints simply due to the amount of data in the console.

2. Skew data
- This might not be a big deal but let's face it...we currently have X number of iOS engineers, building and running the app 8 hours a day, 5 days a week.
- In addition to that CircleCI runs those same builds (unless those builds are run in `Release` mode)

3. Looker
- As far as I've been able to dig into this, I'm not aware of the fact that we would differentiate between data that's been recorded in `DEBUG` mode (iOS simulator, CircleCI, etc.) and production (AppStore, Beta, etc.)
- I've compared requests under `DEBUG` and `RELEASE` configurations and didn't find any additional variations (which should cover my previous assumption)

# 🛠 How

Simply put, we only call & log the tracking event if the app was build in `RELEASE` mode.

# ✅ Acceptance criteria

Switching between `DEBUG` and `RELEASE` mode can be done in `Product > Scheme > Edit Scheme > Run > Build configuration`

screen shot 2019-02-22 at 11 44 27 am

Enabling/disabling environment variable can be done in `Product > Scheme > Edit Scheme > Run > Arguments`

screen shot 2019-03-05 at 1 47 52 pm

- [x] `KOALA_TRACKING` environment variable is OFF by default (☐)
- [x] `KOALA_TRACKING` environment variable's value is `true` (`String`)
- [x] Koala events are being tracked & logged in the console when in `RELEASE` mode && `KOALA_TRACKING` variable is ON (✅ )
- [x] Koala events are being tracked & logged in the console when in `RELEASE` mode && `KOALA_TRACKING` variable is OFF (☐)
- [x] Koala events are being tracked & logged in the console when in `DEBUG` mode && `KOALA_TRACKING` variable is ON (✅ )
- [x] Koala events are NOT being tracked & NOT being logged in the console when in `DEBUG` mode && `KOALA_TRACKING` variable is OFF (☐)


dusi:
Pretty good article to explain the various levels of logging http://www.avanderlee.com/debugging/oslog-unified-logging/

ifbarrera:
@dusi I'm generally onboard with not tracking Koala events in `DEBUG` mode, but I do worry that someone working on tracking won't know where to look to re-enable tracking while testing. Could we use a combination of `DEBUG` and an scheme environment variable to make this easier?

Maybe something like this:
image

Combined with:
```
    var enableKoalaTracking: Bool = true

    #if DEBUG
    guard ProcessInfo.processInfo.environment["TRACK_KOALA"] != nil else {
      enableKoalaTracking = false

      return
    }
    #endif

    if enableKoalaTracking {
        // Track
    }
```

That would allow us to easily toggle ON/OFF Koala tracking in debug mode without having to dig through the code to comment something out.

dusi:
I'd personally also prefer environment variable...which is why it was one of the `TODO` discussions.

I'll see what I can do...ideally this is something we would be able to inject into the `AppEnvironment` struct for consistency and test purposes.

Library/EnvironmentVariables.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct EnvironmentVariables {
+  private let processInfo: ProcessInfoType
+
+  public init(processInfo: ProcessInfoType = ProcessInfo.processInfo) {
+    self.processInfo = processInfo
+  }
+}
+
+extension EnvironmentVariables {
+  public var kaoalaTrackingIsOn: Bool {
```
dusi: Another option for naming these could be `isKoalaTrackingEnabled`, `isNetworkLoggingEnabled`, etc.

ifbarrera: I like `isKoalaTrackingEnabled` 👍

Library/Environment.swift
```diff
@@ -121,7 +124,8 @@ public struct Environment {
     reachability: SignalProducer<Reachability, NoError> = Reachability.signalProducer,
     scheduler: DateScheduler = QueueScheduler.main,
     ubiquitousStore: KeyValueStoreType = NSUbiquitousKeyValueStore.default,
-    userDefaults: KeyValueStoreType = UserDefaults.standard) {
+    userDefaults: KeyValueStoreType = UserDefaults.standard,
+    variables: EnvironmentVariables = EnvironmentVariables()) {
```
ifbarrera: Could `envVariables` or `environmentVariables` be a more descriptive name for this argument?

Library/EnvironmentVariables.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct EnvironmentVariables {
+  private let processInfo: ProcessInfoType
+
+  public init(processInfo: ProcessInfoType = ProcessInfo.processInfo) {
+    self.processInfo = processInfo
+  }
+}
+
+extension EnvironmentVariables {
+  public var isKaoalaTrackingEnabled: Bool {
```
ifbarrera: Small typo here, should be `isKoalaTrackingEnabled`

dusi: 😄

Library/Environment.swift
dusi: My initial thinking was that since this is on an `Environment` object it already has a context that these are environment variables `Environment.variables`. The way this is called is `AppEnvironment.current.variables` so maybe naming this `environmentVariables` doesn't sound so bad after all! I can make the change!

PR #615, opened 2019-02-22
👩‍🔬 [Investment Day] Track VoiceOver

dusi:
# 📲 What

Originally this PR intended to add tracking for both Dynamic Type font size & VoiceOver state, but since we already track the font size, we're only adding whether or not VoiceOver is running.

# 🤔 Why

Tracking a11y features helps us make more informed decisions around supporting Dynamic Type, VoiceOver, etc.

# 🛠 How

Simply by adding one extra property to default properties.

# ✅ Acceptance criteria

Launching the app (or performing any other event that is being tracked)..

- [x] Logs that event with properties that include `"is_voiceover_running": false` if VoiceOver is NOT running
- [x] Logs that event with properties that include `"is_voiceover_running": true` if VoiceOver is running

# ⏰ TODO

- [x] Discuss with @nnekab


Library/Koala/KoalaTests.swift
```diff
@@ -46,6 +46,31 @@ final class KoalaTests: TestCase {
...
+    let koala = Koala(client: client)
+
+    var isVoiceOverRunning = { true }
+
+    withEnvironment(isVoiceOverRunning: isVoiceOverRunning) {
+      koala.trackAppOpen()
+
+      let properties = client.properties.last
+
+      XCTAssertTrue(properties?["is_voiceover_running"] as! Bool)
```
justinswart: The force cast necessary?

dusi: I always considered force unwrapping in tests an OK approach since tests are not shipping to production and you actually really want the value to be what you tell it to be so if it's not you have a problem.

Anyways, I'm fine with using optionals...I've also removed any other force unwraps in this test suite for consistency. It looks like `XCTAssertEqual` is a better API to use when dealing with optionals since `XCTAssertTrue`/`False` expects non-optionals.

justinswart: Thanks, yeah I think it's better for tests to fail rather than potentially crash. I didn't mean for you to go and change all of the other force unwraps but thanks for doing that! 😄 I was just thinking that the force _cast_ in this instance maybe wasn't necessary 👍

PR #575, opened 2019-02-04
Currency Selection Feature

justinswart:
# 📲 What

Replaces the currency picker behaviour with selection from a table view.

| Before  | After |
| ------------- | ------------- |
|  |  |

# 🤔 Why

The previous behaviour was a little tricky to use and switching to a list felt like a better user experience and more consistent with the rest of our settings implementations.

# 🛠 How

- Removed the currency picker, replaced with a `UITableView`.
- Changed the transition to push the table view onto the navigation stack.

# 👀 See

[Trello](https://trello.com/c/YCF8GxEv)

# ♿️ Accessibility 

- [ ] Tap targets use minimum of 44x44 pts dimensions
- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

TBD

# ⏰ TODO

- [x] Merge #568 
- [x] Merge #569 
- [x] Merge #570 
- [x] Merge #571 
- [x] Merge #572

PR #616, opened 2019-02-25
Fix scrollToTop again

dusi:
# 📲 What

Use optionals to prevent from possible crash caused by accessing implicitly unwrapped optional.

# 🤔 Why

We're still seeing couple crashes related to the `scrollToTop` issue. Due to the fact that our crash reporting tools indicated the problem was only related to `BackerDashboardViewController` it was the only controller fixed at the time. Since the same implementation is also used in `DiscoveryViewController` it's only adequate to also address it here.

# 🛠 How

Using optionals when extracting a reference to a view that might be on a controller that might not be allocated due to being declared as implicitly unwrapped optional.

# 👀 See

https://github.com/kickstarter/ios-oss/pull/516

# ✅ Acceptance criteria

- [ ] `Explore` screen still scrolls to top when tapping the status bar
- [ ] Same behaviour works for different filter (Projects We Love, Popular, Newest, etc.)


dusi:
[Scroll to top crash](https://trello.com/c/guggng35/778-scroll-to-top-crash)

ifbarrera:
@dusi from what I recall, our previous hypothesis on this crash had to do with a potential memory leak on VCs storing weak references to `pageViewController`s. I re-opened the trello card with a reference to the crash hoping we could take a closer look at the leak in a future sprint. This temporary solution looks good for now, but maybe we should make sure we address the root cause in an upcoming sprint.

justinswart:
Thanks for looking into this @dusi.

Just to summarize from our discussion on Slack:

- The likely cause of this appears to be that some of our VC's are leaking.
- We should try to reproduce using the breadcrumb logs that we have in the crash reports, they all appear to crash when the user selects the currency cell after logging out and in again (and doing a bunch of stuff in between).
- This seems related to us previously propagating the `ksr_userLocalePreferencesChanged` notification when the user updates their currency. This was removed in #575 but should be considered a regression and be brought back.
- If we can reproduce it this way, we should try to prove that it is or isn't caused by the stuff that we're doing in `RootViewModel` with signals emitting our root view controllers: https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/ViewModels/RootViewModel.swift#L108. I believe that this could be problematic and we should look at ways to change this so that signals are not retaining view controllers. Perhaps instead we should emit `enum` cases representing the view controllers to create rather than the view controllers themselves.

PR #617, opened 2019-02-26
♿️ Settings notifications multiple toggles

cdolm92:
# 📲 What

Provides user with more context as to what notifications they’re toggling. 

# 🤔 Why

The toggles are also not explicitly stating their purpose since these are custom components we’re using.

# 🛠 How

Disabling the title helped us structure the information in a better manner for visually impaired users.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|unnamed 1| unnamed|

# ♿️ Accessibility 

Navigate to `Settings > Notifications`. Then turn VoiceOver ON.

- [x] Tap on the email notification of each notification. 
          If selected: VoiceOver should read: `[Notification Name], email notification on - Double tap to toggle setting`
   If deselected: VoiceOver should read: `[Notification Name], email notification off - Double tap to toggle setting`
- [x] Toggle email notification:
         If toggled on VoiceOver should read: `[Notification Name], email notification on`
         If toggled off VoiceOver should read: `[Notification Name], email notification off`
- [x] Tap on the push notification of each notification. 
          If selected: VoiceOver should read: `[Notification Name], push notification on - Double tap to toggle setting`
   If deselected: VoiceOver should read: `[Notification Name], push notification off - Double tap to toggle setting`
- [x] Toggle push notification:
         If toggled on VoiceOver should read: `[Notification Name], push notification on`
         If toggled off VoiceOver should read: `[Notification Name], push notification off`


Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -71,10 +74,11 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
       |> UIButton.lens.image(for: .selected) .~ Library.image(named: "email-icon",
                                                       tintColor: .ksr_green_700,
                                                       inBundle: Bundle.framework)
-      |> UIButton.lens.accessibilityLabel %~ { _ in Strings.Email_notifications() }
 
     _ = self.pushNotificationsButton
       |> notificationButtonStyle
+      |> \.accessibilityTraits .~ .none
+      |> UIButton.lens.accessibilityHint %~ { _ in Strings.Double_tap_to_toggle_setting() }
```
Scollaco: Since the `accessibilityHint` string is the same for both push and email buttons, I would consider moving this line to the `notificationButtonStyle`, avoiding to repeat ourselves (The same would apply for `\.accessibilityTraits .~ .none`).

justinswart: Do we include strings in any of our other style functions?

Scollaco: The `settingsLogoutButtonStyle` includes strings

cdolm92: @justinswart We do!

Kickstarter-iOS/Locales/ja.lproj/Localizable.strings
```diff
@@ -371,7 +376,7 @@
 "Rewards_count_rewards_colon.one" = "%{rewards_count} 種類のリワード";
 "Rewards_count_rewards_colon.two" = "%{rewards_count} 種類のリワード";
 "Rewards_count_rewards_colon.zero" = "%{rewards_count} 種類のリワード";
-"Save" = "救う";
+"Save" = "保存";
```
dusi: 🤔

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -38,6 +38,7 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
 
     _ = titleLabel
       |> UILabel.lens.text .~ cellValue.cellType.title
+      |> \.accessibilityElementsHidden .~ cellValue.cellType.accessibilityElementsHidden
```
dusi: Is there a reason we're using `accessibilityElementsHidden` and not `isAccessibilityElement`? I'm not sure that `titleLabel` can have any subviews?

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -92,9 +91,12 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
 
     self.emailNotificationsButton.rac.selected = viewModel.outputs.emailNotificationsEnabled
     self.emailNotificationsButton.rac.hidden = viewModel.outputs.emailNotificationButtonIsHidden
+    self.emailNotificationsButton.rac.accessibilityLabel =
+      viewModel.outputs.emailNotificationAccessibilityLabel
```
dusi: Can we change the names of these 2 signals?

So instead of `emailNotificationAccessibilityLabel` do `emailNotificationsButtonAccessibilityLabel` and also instead of `pushNotificationAccessibilityLabel` do `pushNotificationsButtonAccessibilityLabel`.

This should follow the pattern of `component+property`.

Library/Strings.swift
```diff
@@ -20182,7 +20267,7 @@ projets enregistrés"
    - **de**: "%{created_count} erstellt"
    - **es**: "%{created_count} creados"
    - **fr**: "%{created_count} projets créés"
-   - **ja**: "%{created_count} プロジェクト"
+   - **ja**: "%{created_count}プロジェクト"
```
dusi: I wonder if we should get rid of these strings that are unrelated to this PR and maybe review them before they slip to our code base? It looks weird that `プ` would replace a space character to me (even though I don't speak Japanese).

What do you think Christella?

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
cdolm92: Because these buttons behave more like switches I was not sure that referring to it as a button . was accurate. Let me know what you think.

Library/Strings.swift
cdolm92: Ideally but maybe we should have a team discussion on handling strings? I think we can change this approach later.

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
dusi: I realize we're not really using them as buttons, but technically the component (`UIButton`) and the name of the component (`emailNotificationsButton`) refer to a button so we should probably stay consistent. The behaviour should be considered a side effect IMO

Library/Strings.swift
dusi: Agree, we should bring this up with Nneka & Peat (either parking lot or slack)?

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
```diff
@@ -83,7 +83,6 @@ final class SettingsNotificationCell: UITableViewCell, NibLoading, ValueCell {
       |> UIButton.lens.image(for: .selected) .~ Library.image(named: "mobile-icon",
                                                       tintColor: .ksr_green_700,
                                                       inBundle: Bundle.framework)
-      |> UIButton.lens.accessibilityLabel %~ { _ in Strings.Push_notifications() }
```
dusi: Noticed that there was previously an empty line here, can we possibly remove it?

Kickstarter-iOS/Views/Cells/SettingsNotificationCell.swift
cdolm92: When `isAccessibilityElement` is set to `false` for the notification cells w/  email/push icons and `true` for `Manage project notifications` button, VoiceOver as expected, would avoid focusing on the notification cell titles, however, for `Manage project notifications` button  VoiceOver was reading the button title and number of notifications separately. If the user tapped over the cell VoiceOver was only reading the number of notifications, the user would have to tap over the title for it to read.  With `accessibilityElementsHidden` VoiceOver avoids focusing on reading the titles of the notification cells w/ email/push icons while still reading the `Manage Project notifications` title and number of notifications together.

dusi: So the only way I was able to have this work properly was to use your original approach + fix the order of the buttons like so:

```
func configureWith(value cellValue: SettingsNotificationCellValue) {
    self.notificationType = cellValue.cellType

    viewModel.inputs.configure(with: cellValue)

    let accessibilityElementsHidden = cellValue.cellType.accessibilityElementsHidden

    _ = self
      |> \.accessibilityTraits .~ cellValue.cellType.accessibilityTraits

    _ = self.stackView
      |> \.accessibilityElements .~ (
        accessibilityElementsHidden ? [self.emailNotificationsButton, self.pushNotificationsButton] : nil
    )

    _ = self.titleLabel
      |> UILabel.lens.text .~ cellValue.cellType.title
      |> \.accessibilityElementsHidden .~ accessibilityElementsHidden

    _ = self.arrowImageView
      |> UIImageView.lens.isHidden .~ cellValue.cellType.shouldHideArrowView
      |> UIImageView.lens.tintColor .~ .ksr_dark_grey_400

    _ = self.projectCountLabel
      |> UILabel.lens.isHidden .~ cellValue.cellType.projectCountLabelHidden
      |> \.accessibilityElementsHidden .~ accessibilityElementsHidden
  }
```

please notice that:
* I've added `self.` to reference `IBOutlet`s consistently
* Also set `accessibilityElementsHidden` on the `projectCountLabel` (it worked without it but might be confusing as to why since we don't won't either of the labels to be accessible and let the cell take over instead)

Library/SettingsNotificationCellType.swift
```diff
@@ -81,6 +81,15 @@ public enum SettingsNotificationCellType {
     }
   }
 
+  public var accessibilityElementsHidden: Bool {
+    switch self {
+    case .projectNotifications, .emailFrequency:
```
dusi: Do we need `.emailFrequency` here at all? The cell we're using for that is `SettingsNotificationPickerCell` which doesn't seem to use this value, yet it still works.

PR #618, opened 2019-02-26
"Find Friends" cell styling & refactor

ifbarrera:
# 📲 What

Updates the styling of the "Find Friends" cell. Also refactored to make the "disabled text" a footer on the `.findFriends` section, so that it's styling is also consistent with other settings table view section footers.

Other work in this PR:
- removed most custom separators from settings table view cells in favor of `UITableView`'s default separators

# 🤔 Why

The "Find Friends" cell styling was a little weird. It looks much better now. Also the footer description is now much more consistent with the rest of settings.

# 🛠 How

When `following` is disabled on the User's settings, the "Find Friends" cell should be disabled and display a footer with a description as to why it's disabled. This description text was previously part of the `FindFriendsCell`, which broke the way separators displayed. By moving it out into a section footer, we can now use `UITableView`'s default separators instead of rolling our own.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| simulator screen shot - iphone 8 - 2019-02-26 at 12 19 04 | simulator screen shot - iphone 8 - 2019-02-25 at 17 04 04|

Other Settings screens now using `UITableView` separators:

**Help Screen**
simulator screen shot - iphone x - 2019-02-26 at 11 14 39

**Account Screen**
simulator screen shot - iphone x - 2019-02-26 at 11 14 44

**Notifications Screen**
simulator screen shot - iphone 8 - 2019-02-26 at 12 28 57

**Regression Test**
ifmikivqps

**Dynamic Type**
simulator screen shot - iphone x - 2019-02-26 at 14 21 20

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Navigate to the Settings screen. If Settings -> Account -> Privacy -> Following is turned on, the "Find Facebook Friends" cell should be enabled, with the same styling as the other cells in the table. 
- [x] Navigate to the Settings screen. If Settings -> Account -> Privacy -> Following is turned **OFF**, the "Find Facebook Friends" cell should be disabled, and the description should appear below the cell as a section footer, with the same styling as the "App Version" footer.
- [x] With Voice Over **ON** and Following **OFF**, tapping the "Find Facebook Friends" cell should say "Find Facebook Friends, dimmed" and it should speak the footer description as a hint.
- [x] With Voice Over **ON** and Following **ON**, tapping the "Find Facebook Friends" cell should say "Find Facebook Friends" and it should **not** speak the footer description.


ifbarrera:
[⚙️ Find Facebook friends TableViewCell: change 'disabled' styling to match that of 'Current email' TableViewCell](https://trello.com/c/BDF1QNDd/1193-%E2%9A%99%EF%B8%8F-find-facebook-friends-tableviewcell-change-disabled-styling-to-match-that-of-current-email-tableviewcell)

Kickstarter-iOS/Views/Controllers/HelpViewController.swift
```diff
@@ -38,6 +38,7 @@ final class HelpViewController: UIViewController {
 
     _ = tableView
       |> settingsTableViewStyle
+      |> settingsTableViewSeparatorStyle
```
Scollaco: Could we add the separator style to the tableView style instead?
```
public let settingsTableViewStyle = UITableView.lens.backgroundColor .~ .ksr_grey_200
  <> settingsTableViewSeparatorStyle
```

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
```diff
@@ -169,15 +175,18 @@ extension SettingsViewController: UITableViewDelegate {
   }
 
   func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
-    guard section == SettingsSectionType.ratingAppVersion.rawValue else { return nil }
+    guard let section = SettingsSectionType(rawValue: section), section.hasSectionFooter else { return nil }
```
Scollaco: Since the logic is identical to the `heightForFooterInSection` method, we could create a private helper function  that returns a Bool and use the return value to return the height / view. 
```
private func shouldHideFooterHeaderView(for section: Int) -> Bool {
guard let section = SettingsSectionType(rawValue: section), section.hasSectionFooter else { return true }
    if section == SettingsSectionType.findFriends
      && !self.viewModel.outputs.findFriendsDisabledProperty.value {
      return true
    }
return false
}
```
and do `return shouldHideFooterHeaderView(for: section) ? 0.1 : UITableView.automaticDimension`, etc...

Library/ViewModels/SettingsViewModel.swift
```diff
@@ -42,7 +43,13 @@ SettingsViewModelOutputs, SettingsViewModelType {
       }
       .skipNil()
 
-    self.reloadDataWithUser = user
+    let isFollowingEnabled = user
+      .map { $0 |> User.lens.social.view }
```
Scollaco: We have a `view` function that supports keyPaths: `$0 |> (\User.social).view`

Kickstarter-iOS/Views/Controllers/HelpViewController.swift
ifbarrera: We can't do that quite yet because the settings newsletters screen is using the `settingsTableViewStyle` but still using custom separators. I thought it might get a little complicated to remove them on that screen in this PR, but once we do we can merge the two styles.

Kickstarter-iOS/Views/Controllers/SettingsViewController.swift
ifbarrera: Good idea, I'll work on something like this, maybe in the data source though so it's more easily testable 👍

PR #620, opened 2019-02-28
Disable COMPILER_INDEX_STORE_ENABLE on CI

justinswart:
# 📲 What

Adds a `CircleCi.xcconfig` file which disables `COMPILER_INDEX_STORE_ENABLE` for a small performance gain on CI.

# 🤔 Why

@dusi mentioned a tweet that suggested this would offer a small performance gain so we figured we'd give it a try.

Results that I've seen on an initial run are:

| Test suite | [Before](https://circleci.com/workflow-run/fa2c6a79-ea07-453e-9693-cf7d23bf9054) 🐛 | [After](https://circleci.com/workflow-run/1e23e453-1a18-4282-8ba0-1a130c138d3e) 🦋 |
| --- | --- | --- |
| Library | 5:51 | 5:37 |
| Kickstarter-Framework | 9:20 | 8:31 |
| KsApi | 1:22 | 1:35 |
| LiveStream | 1:03 | 1:09 |

Curiously, on KsApi and LiveStream we saw a slight decline in speed, so it's hard to say whether the gain of nearly a minute for `Kickstarter-Framework` is worth it?

# 🛠 How

- Added a `CircleCI.xcconfig` file which inherits from `Base.xcconfig`.
- Implemented the use of `CircleCI.xcconfig` in our `Makefile` for tests and `Fastfile` for `gym`.

# 👀 See

https://twitter.com/steipete/status/1100404024224804871

# ✅ Acceptance criteria

- [ ] There is a tangible performance gain

### DIFF
--- a/Configs/CircleCI.xcconfig
+++ b/Configs/CircleCI.xcconfig
@@ -1,0 +1,3 @@
+#include "Base.xcconfig"
+
+COMPILER_INDEX_STORE_ENABLE=NO
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 XCODEBUILD := xcodebuild
-BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION)
+BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION) -xcconfig Configs/CircleCI.xcconfig
 
 SCHEME ?= $(TARGET)-$(PLATFORM)
 TARGET ?= Kickstarter-Framework
--- a/.fastlane/Fastfile
+++ b/.fastlane/Fastfile
@@ -233,7 +233,8 @@ platform :ios do
       export_method: options[:export_method],
       scheme: "Kickstarter-iOS",
       configuration: options[:configuration],
-      export_options: export_options
+      export_options: export_options,
+      xcconfig: "Configs/CircleCI.xcconfig"
     )
   end
 


Configs/CircleCI.xcconfig
```diff
@@ -0,0 +1,3 @@
+#include "Base.xcconfig"
```
justinswart: I've included `Base.xcconfig` here because it exists but to be honest I'm not sure if we actually use this. Does Xcode use it by default if it's called `Base.xcconfig`?

justinswart:
[Disable compiler indexing flag on CircleCI](https://trello.com/c/tYybUcDf/1253-disable-compiler-indexing-flag-on-circleci)

Configs/CircleCI.xcconfig
dusi: Looks like it is...it's easily spotted in build settings.

screen shot 2019-02-28 at 1 50 48 pm

Makefile
```diff
@@ -1,5 +1,5 @@
 XCODEBUILD := xcodebuild
-BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION)
+BUILD_FLAGS = -scheme $(SCHEME) -destination $(DESTINATION) -xcconfig Configs/CircleCI.xcconfig
```
dusi: Good idea

dusi:
Wonder whether we should try running this couple times (10-50 times) and try collecting the metrics? Might be an overkill but would be interesting to see if these times are consistent or not.

Configs/CircleCI.xcconfig
justinswart: Ah thanks

justinswart:
Yeah that makes sense @dusi, I'll re-run the workflow a couple times, and let's see if we can get some averages 👍

justinswart:
Yeah that makes sense @dusi, I'll re-run the workflow a couple times, and let's see if we can get some averages 👍

justinswart:
Closing this for now - I couldn't reliably reproduce significant improvements 🤷‍♂️

PR #621, opened 2019-02-28
Comments keyboard inset

ifbarrera:
# 📲 What

Fixes the inset on the `UITextView`'s `bottom` constraint to properly account for the height of the keyboard.

# 🤔 Why

Previously, users could not see what they were typing because it would fall underneath the keyboard's frame.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|3iz8ktwjhf| l2p2nkjgbm|

# ♿️ Accessibility 

- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Navigate to a backed project (Settings -> Backed -> Project) then navigate to a new comment (Comments -> Comment). Type a bunch of text in the text view until it reaches the top of the keyboard, and continue typing. The text should not fall underneath the frame of the keyboard.


Kickstarter-iOS/Views/Controllers/CommentDialogViewController.swift
```diff
@@ -111,7 +111,8 @@ internal final class CommentDialogViewController: UIViewController {
 
   fileprivate func animateTextViewConstraint(_ change: Keyboard.Change) {
     UIView.animate(withDuration: change.duration, delay: 0.0, options: change.options, animations: {
-      self.bottomConstraint.constant = self.view.frame.height - change.frame.minY
+      let keyboardOffset = self.view.frame.height - change.frame.height
+      self.bottomConstraint.constant = keyboardOffset + Styles.grid(2)
```
ifbarrera: The additional `Styles.grid(2)` is just to give it a tiny bit of padding beyond the frame of the keyboard.

ifbarrera:
[[ZD] [iOS] Bonjour, Assistance de l'application Kickstarter (FR - EN)](https://trello.com/c/o3HM0ghS/1244-zd-ios-bonjour-assistance-de-lapplication-kickstarter-fr-en)

justinswart:
Hmm, any idea why the gap would be so large here?

Kickstarter-iOS/Views/Storyboards/Comments.storyboard
```diff
@@ -262,16 +262,12 @@
         <scene sceneID="BCA-NB-R56">
             <objects>
                 <viewController storyboardIdentifier="CommentDialogViewController" id="99a-92-DMl" customClass="CommentDialogViewController" customModule="Kickstarter_Framework" customModuleProvider="target" sceneMemberID="viewController">
-                    <layoutGuides>
```
ifbarrera: The `textView` and the `LoadingView` were previously constrained to the bottom layout guide. This caused an issue because the height of the keyboard always includes space that would technically be outside the "safe area".

Instead, pinned the bottom of these views to the superview's `bottom`, and manually adjusted the inset of the `bottom` constraint to account for the total height of the keyboard.

PR #622, opened 2019-03-01
Unknown card type Decodable fix

ifbarrera:
# 📲 What

[This PR](https://github.com/kickstarter/ios-oss/commit/796d58a824c514ee4bee7e1eded650202c3273a6) was meant to support unknown strings coming from the card `type` in GraphQL, but unfortunately only actually supported `null` `types`'s.

# 🤔 Why 

It turns out that Swift's standard `Decodable` conformance for optional `enum`s surprisingly causes deserialization to fail if the raw value does not match a `case` (instead of setting the enum to `nil`). Why? We'll probably never know because we don't know how Swift's default `Decodable` implementation works under the hood, which is frustrating. A theory is that since `init(from decoder)` is *not* a failable initializer, it can't actually (validly) return `nil`, so it's only option is to `throw` when it can't generate a case from the associated value. The solution is to override `init(from decoder)` for the `CreditCardType` enum to manually handle decoding, setting the enum to `.generic` when an unknown string value is used.

**Update** Swift is open source. Duh. So we *can* know what's going on under the hood, and here it is:
https://github.com/apple/swift/blob/master/stdlib/public/core/Codable.swift.gyb#L1686

Summary is:
- if the value being decoded is `Optional` it first tries `container.decodeNil()` which simply checks whether a value is present in the container for the given key. If it finds a value, it then calls `try container.decode(Wrapped.self)` where `Wrapped` is the enum type
- `try container.decode(Wrapped.self)` will hit the `init(from decoder: Decoder)` implementation of `RawRepresentable`, which is linked above. The implementation shows that if the value of `Self(rawValue: decoded)` is nil, a `DecodingError.dataCorrupted` error is thrown, which is why it doesn't "fail gracefully" into `nil`

# 🛠 How

Override `CreditCardType`s `init(from decoder)` and return `.generic` if initialization with the associated string fails. Also added the missing test that would have caught this.

# 👀 See

dwgmug1ofl

# ✅ Acceptance criteria

- [ ] Add a union pay card on staging web. Navigate to payment methods screen on iOS: stored cards should all display correctly, with the union pay card showing the "generic" treatment


ifbarrera:
[Payment methods screen should handle unknown cards correctly](https://trello.com/c/132VRnA3/1255-payment-methods-screen-should-handle-unknown-cards-correctly)

KsApi/models/GraphUserCreditCardTests.swift
```diff
@@ -104,4 +104,41 @@ final class GraphUserCreditCardTests: XCTestCase {
...
+            ],
+            "totalCount": 0
+          }
+        }
+        """
+
+    let data = jsonString.data(using: .utf8)
+
+    do {
+      //swiftlint:disable force_unwrapping
```
dusi: Can we use `//swiflint:disable:next`? That should only disable the rule on the next line...otherwise it will disable it till end of file.

PR #625, opened 2019-03-02
🔑 [Create Password] Form validation

dusi:
# 📲 What

Adds local validation to the `Create password` form.

Additionally (based on @Scollaco 's feedback) this PR addresses and improves UX when validating input fields.

# 🤔 Why

This allows us to do the following:
* enables/disables the `Save` button in the navigation bar
* shows/hides the `Validation label`

# 🛠 How

Identical to the `Change password` functionality. In order to stay DRY this PR attempts extracting common logic from both of these VMs (`CreatePasswordViewModel` and `ChangePasswordViewModel`).

# ➡️ 

Prerequisites: Login to the app with a Facebook account that you have not yet created a password for

or

copy-paste this code snippet in `SettingsAccountViewController.swift`'s `viewDidLoad`

```
DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
  let vc = CreatePasswordViewController.instantiate()
}
```

# 👀 See

| Length validation | Matching passwords validation |
| --- | --- |
| validation1 | validation2 |

# ♿️ Accessibility 

For both create password & change password forms

- [x] Tapping either `New password` or `Confirm new password` cell (anywhere) focuses on the text field and reads its value + state
- [x] While typing in `Confirm new password` field and releasing holding the keyboard, VoiceOver will focus on the `Validation label` and read its value (either `short password` or `passwords do not match`).
- [x] The form responds to font size changes

# ✅ Acceptance criteria

For create password form

- [x] Default state of the `Save` button is disabled
- [x] Default state of the `Validation label` is hidden
- [x] Tapping either `New password` or `Confirm new password` cell (anywhere) makes its text field first responder (and shows the keyboard)
- [x] `New password` text field uses `Next` key as its return key
- [x] Tapping `New password` text field's `Next` button will make `Confirm new password` first responder
- [x] `Confirm new password` text field uses `Done` key as its return key
- [x] Tapping `Confirm new password` text field's `Done` button will dismiss the keyboard
- [x] Entering value into `New password` field that is shorter than 6 characters will result in showing the `Validation label` with a message `Your password must be at least 6 characters long.`
- [x] Entering value into `New password` field that is longer than 6 characters will result in hiding the `Validation label`
- [x] Entering value into `Confirm new password` field that does not match `New password` field results in showing the `Validation label` with a message `New passwords must match.`
- [x] The validation considers `New password` field validation first, `Confirm new password` field later (i.e. `New password` = `test` & `Confirm new password` = `testtest`=> `Your password must be at least 6 characters long.`
- [x] Both validation messages (`short password` and `passwords not matching`) are localized
- [x] `Save` button is only enabled if both passwords are valid (6+ characters & matching)

For change password form

- [x] Entering value into `New password` field that is shorter than 6 characters will result in showing the `Validation label` with a message `Your password must be at least 6 characters long.`
- [x] Entering value into `New password` field that is longer than 6 characters will result in hiding the `Validation label`
- [x] Entering value into `Confirm new password` field that does not match `New password` field results in showing the `Validation label` with a message `New passwords must match.`
- [x] The validation considers `New password` field validation first, `Confirm new password` field later (i.e. `New password` = `test` & `Confirm new password` = `testtest`=> `Your password must be at least 6 characters long.`


Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -2,20 +2,29 @@ import Library
...
 
+public typealias TextFieldTargetAction = (target: Any?, action: Selector, event: UIControl.Event)
+
 final class SettingsTextInputCell: UITableViewCell {
   // MARK: - Properties
 
   private lazy var stackView: UIStackView = { UIStackView(frame: .zero) }()
   private lazy var label: UILabel = { UILabel(frame: .zero) }()
-  private lazy var textField: UITextField = { UITextField(frame: .zero) }()
+  private(set) lazy var textField: UITextField = { UITextField(frame: .zero) }()
```
dusi: We need to access it publicly in order to assign it to local variable on the VC

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -2,20 +2,29 @@ import Library
...
     _ = self.label
       |> \.text .~ title
 
     _ = self.textField
       |> \.placeholder .~ placeholder
       |> \.accessibilityLabel .~ self.label.accessibilityLabel
+      |> \.returnKeyType .~ returnKeyType
+  }
+
+  public func configure(with targetActions: [TextFieldTargetAction]) {
```
dusi: I've decided to use separate function that configures user interactions (as oppose to passing both display values & user interaction actions)

Kickstarter-iOS/Views/Cells/SettingsTextInputCell.swift
```diff
@@ -43,6 +52,9 @@ final class SettingsTextInputCell: UITableViewCell {
   override func bindStyles() {
     super.bindStyles()
 
+    _ = self
```
dusi: This PR adds the ability to tap the cell anywhere which makes the text field first responder (the side effect was that it also highlighted the cell)...this is a standard way of disabling selection mode on the cell.

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -1,27 +1,48 @@
...
+  @objc optional func newPasswordConfirmationTextFieldChanged(_ sender: UITextField)
+  @objc optional func newPasswordConfirmationTextFieldDidReturn(_ sender: UITextField)
 }
 
 final class CreatePasswordViewController: UITableViewController {
-  // MARK: - Accessors
+  // MARK: - Properties
+
+  private let viewModel: CreatePasswordViewModelType = CreatePasswordViewModel()
+  private weak var newPasswordTextField: UITextField?
```
dusi: These `weak var`iables hold references to the UI (just like we previously used to when using XIBs)

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -48,13 +69,77 @@ final class CreatePasswordViewController: UITableViewController {
...
+  }
+
   // MARK: - Styles
 
   override func bindStyles() {
     super.bindStyles()
 
     _ = self.tableView
       |> settingsGroupedTableViewStyle
+      |> \.allowsSelection .~ true
```
dusi: In order for the table to allow cell selection (that allows us changing first responder on tap and focusing on the text field even if user taps on the label)

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -48,13 +69,77 @@ final class CreatePasswordViewController: UITableViewController {
...
+    self.viewModel.outputs.validationLabelIsHidden
+      .observeForUI()
+      .observeValues { [weak self] isHidden in
+        self?.footerView?.isHidden = isHidden
+    }
+
+    self.viewModel.outputs.validationLabelText
+      .observeForUI()
+      .observeValues { [weak self] text in
+        UIView.performWithoutAnimation {
```
dusi: It turned out that this was pretty tricky since simply calling `self?.footerView?.configure(with: text)` would not properly resize the footer (if it was initialized without text) - the `label` height would be 0.

To properly reload a footer we have to put the update between `beginUpdates()` & `endUpdates()` which unfortunately had another side effect of animating the footer in and out (this looked quite odd)...turns out we can disable animations for this operation (not globally - so should be safe) by wrapping it inside of `performWithoutAnimation` block.

Oh UIKit sometimes I love you, sometimes not really 😄

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -93,8 +187,57 @@ final class CreatePasswordViewController: UITableViewController {
     guard let footerView = tableView.dequeueReusableHeaderFooterView(withClass: className)
       as? SettingsGroupedFooterView else { return nil }
 
-    footerView.configure(with: Strings.Password_min_length_message())
+    footerView.configure(with: self.viewModel.outputs.currentValidationLabelText())
```
dusi: In case that the font size gets changed the `tableView` will be reloaded...this is why we have to update footer with the latest value from the `VM` that we can access from a mutable property using `currentValidationLabelText`

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -93,8 +187,57 @@ final class CreatePasswordViewController: UITableViewController {
...
 
     return footerView
   }
+
+  // swiftlint:disable line_length
```
dusi: Not a huge fan of wrapping `UITableView` + `Delegate` or `DataSource` methods on multiple lines so I've disabled Swiftlint...happy to change this if it's not ideal

Kickstarter-iOS/Views/HeaderFooterViewProtocol.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
 
 protocol HeaderFooterViewProtocol {
-  func configure(with text: String)
+  func configure(with text: String?)
```
dusi: So that we can pass `nil` and hide the label (in addition to kidding the wrapper) ... this should be proper behaviour.

Hiding = label with no text

Kickstarter-iOS/Views/SettingsGroupedFooterView.swift
```diff
@@ -7,7 +7,7 @@ final class SettingsGroupedFooterView: UITableViewHeaderFooterView {
 
   static var defaultHeight: CGFloat = 44
 
-  private(set) lazy var label: UILabel = {
+  private lazy var label: UILabel = {
```
dusi: No need for exposing this apparently

Library/PasswordValidation.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+public typealias Password = String
+
+public func passwordsMatch(_ pwds: (first: Password, second: Password)) -> Bool {
```
dusi: Couple convenience logic around matching passwords (reused within `CreatePasswordVM` & `ChangePasswordVM`.

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -48,30 +48,22 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
...
 
-    let fieldsNotEmpty = Signal
-      .combineLatest(
-        currentPasswordSignal,
-        self.newPasswordProperty.signal,
-        self.confirmNewPasswordProperty.signal)
-      .map { valuesInFields in
-        return !valuesInFields.0.isEmpty && !valuesInFields.1.isEmpty && !valuesInFields.2.isEmpty
-    }
+    let fieldsNotEmpty = Signal.combineLatest(
```
dusi: Common logic has been extracted in order for it to being shared between `CreatePasswordVM` & `ChangePasswordVM`

Library/PasswordValidation.swift
dusi: Wonder if we can simply do a `String` comparison...then the call site would look as simple as `let validationMatch = combinedPasswords.map(=)`

Scollaco:
Nice @dusi . I started testing it locally, but I wonder if it would be a better UX if the length validation occurred once the users start editing the `New Password` textfield, so they know right away that the password should have at least 6 characters.

dusi:
@Scollaco yeah, that's a great idea...our validation isn't perfect...I can see what I can do about it! 😄

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -48,13 +69,66 @@ final class CreatePasswordViewController: UITableViewController {
...
+      .observeValues { [weak self] isHidden in
+        self?.groupedFooterView?.isHidden = isHidden
+    }
+
+    self.viewModel.outputs.validationLabelText
+      .observeForUI()
+      .observeValues { [weak self] text in
+        UIView.performWithoutAnimation {
+          self?.tableView.beginUpdates()
+          self?.groupedFooterView?.label.text = text
```
Scollaco: I wonder if it would be a better practice to create a public text var on the GroupedFooterView class and set the label text there, this way we wouldn't need to modify a label's property from outside its class (same principle of private IBOutlets). Something like: 
```
public var text: String? {
    didSet {
      _ = self.label
    |> \.text .~ self.text
    }
  }
```

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -64,27 +138,35 @@ final class CreatePasswordViewController: UITableViewController {
...
 
     guard let headerView = tableView.dequeueReusableHeaderFooterView(withClass: className)
       as? SettingsGroupedHeaderView else { return nil }
 
-    headerView.configure(
-      with: Strings.Well_ask_you_to_sign_back_into_the_Kickstarter_app_once_youve_changed_your_password()
-    )
+    let text = Strings.Well_ask_you_to_sign_back_into_the_Kickstarter_app_once_youve_changed_your_password()
 
+    headerView.label.text = text
```
Scollaco: The same idea mentions on my previous comment would apply for the headerView as well.

Library/ViewModels/CreatePasswordViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    let combinedPasswords = Signal.combineLatest(
+      self.newPasswordChangedProperty.signal.skipNil(),
+      self.newPasswordConfirmationChangedProperty.signal.skipNil()
+    )
+
+    let validationMatch = combinedPasswords.map(passwordsMatch)
+    let validationLength = self.newPasswordChangedProperty.signal.skipNil().map(passwordLengthValid)
+
+    self.validationLabelText = Signal.combineLatest(validationMatch, validationLength)
```
Scollaco: I think here the `combineLatest` is preventing the NewPasswordTextField to show the length validation label, since it will emit only after `validationMatch` and validationLength` emit.

Library/ViewModels/CreatePasswordViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+      .map(passwordValidationText)
+      .skipRepeats()
+
+    self.currentValidationLabelTextProperty <~ self.validationLabelText
+
+    let validationFields = Signal.combineLatest(
```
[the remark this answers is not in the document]
dusi: Good eye! Thank you~

Library/ViewModels/CreatePasswordViewModel.swift
dusi: I simply copied change password behavior. I've looked into it more and have a better way of doing things...stay tuned! 😎

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
dusi: I was initially considering this approach but ended up going with allowing accessing the `label` directly in order to follow UIKit patterns that apply similarly for `UITableViewCell` (which exposes `textLabel` and `detailTextLabel`, or even `UITableViewHeaderFooterView` that exposes `textLabel` and `detailTextLabel` (unfortunately these didn't work well with custom auto layout so they're just there unused).

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -64,26 +152,36 @@ final class CreatePasswordViewController: UITableViewController {
...
+
     return textInputCell
   }
 
   // MARK: - UITableViewDelegate
 
+  override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+    guard let textInputCell = tableView.cellForRow(at: indexPath) as? SettingsTextInputCell else { return }
+
+    self.viewModel.inputs.textFieldShouldBecomeFirstResponder(textInputCell.textField)
```
ifbarrera: I think we generally try to keep view models free of any knowledge of `UIKit` things. Could we pass the index path's row here instead to determine which text field needs to become the first responder?

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -117,30 +109,35 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
...
 
-    self.validationErrorLabelMessage = errorLabelMessage
-      .skipNil()
-      .skipRepeats()
+    self.validationErrorLabelMessage = validationLabelText
+      .map { $0 ?? String() }
+
+    let validationLabelTextIsEmpty = self.validationErrorLabelMessage
+      .map { $0.isEmpty }
+    .logEvents(identifier: "*** validationLabelTextIsEmpty")
```
ifbarrera: We should remove logging here

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -117,30 +109,35 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
...
+
+    self.validationErrorLabelIsHidden = validationLabelTextIsEmpty
 
-    let entryChanged = Signal.merge(self.newPasswordProperty.signal, self.confirmNewPasswordProperty.signal)
+    let inputsChanged = Signal.merge(
+      self.newPasswordProperty.signal, self.confirmNewPasswordProperty.signal
+    ).logEvents(identifier: "*** inputsChanged")
```
ifbarrera: And here

Library/ViewModels/ChangePasswordViewModel.swift
```diff
@@ -230,3 +227,9 @@ ChangePasswordViewModelInputs, ChangePasswordViewModelOutputs {
     return self
   }
 }
+
+// MARK: - Functions
+
+private func formFieldsNot(_ pwds: (first: String, second: String, third: String)) -> Bool {
```
ifbarrera: Should this be `formFieldsNotEmpty`? 🤔

Library/ViewModels/ChangePasswordViewModelTests.swift
```diff
@@ -145,73 +141,118 @@ final class ChangePasswordViewModelTests: TestCase {
...
 
-      self.validationErrorLabelIsHidden.assertValues([false])
-      self.validationErrorLabelMessage
-        .assertValues(["Your password must be at least 6 characters long."])
-      self.accessibilityFocusValidationErrorLabel.assertValueCount(1)
-      self.saveButtonIsEnabled.assertValues([false])
+      self.vm.inputs.currentPasswordFieldTextChanged(text: "password")
+      self.accessibilityFocusValidationErrorLabel.assertValueCount(0)
+      self.saveButtonIsEnabled.assertValueCount(0)
+      self.validationErrorLabelIsHidden.assertLastValue(true)
```
ifbarrera: I find that using `assertValues` can help uncover instances where the signal is firing more often than it should be.

Library/ViewModels/CreatePasswordViewModel.swift
```diff
@@ -0,0 +1,151 @@
...
+    self.validationLabelText = Signal.merge(
+      self.viewDidAppearProperty.signal.mapConst(nil),
+      newPasswordValidationText,
+      newPasswordAndConfirmationValidationText
+    )
+
+    self.currentValidationLabelTextProperty <~ self.validationLabelText
+
+    let validationLabelTextIsNil = self.validationLabelText
+      .map { $0 == nil }
```
ifbarrera: Could also use `.map(isNil)` here

Library/ViewModels/ChangePasswordViewModel.swift
dusi: giphy

Library/ViewModels/ChangePasswordViewModelTests.swift
dusi: In this particular example that's exactly what I was going to hide since the signals will be triggered multiple times but the final outcome is that the label will be hidden. But I can see the value of having historical values...will change that.

Library/ViewModels/CreatePasswordViewModel.swift
dusi: Oh nice!

PR #626, opened 2019-03-04
Settings Dynamic Type Fixes

ifbarrera:
# 📲 What

Fixes a bug that prevented fonts in settings from resizing correctly with dynamic type. Also turns off font resizing on the `LoadingBarButtonItemView` since it's a navigation item.

# 🛠 How

A deep-dive done by @dusi into dynamic type within Settings identified a bug where several labels whose fonts were set using our custom `.ksr_()` functions were not resizing correctly when the font size was changed using the accessibility inspector. An investigation showed that this was due to the helper styling functions in `SettingsStyles.swift` capturing the current context which essentially cached the initial `preferredFont` instead of re-evaluating it every time `bindStyles` was called.

The solution was to convert the `SettingsStyles` helpers that are setting a font into a closure that takes a label and applies the styles within the closure.

# 👀 See

Before:
 - notice that although the section footers resize, the title labels for the settings cells do not
 - notice that the "Save" button resizes even though it shouldn't

After:
 - notice that all cell titles now resize
 - notice that the "Save" button no longer resizes

| Before 🐛 | After 🦋 |
| --- | --- |
| y4xlkmeme4 | mhnrvh6nhy |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Run the app with the accessibility inspector. Then use the accessibility inspector to increase the font size. Navigate to Settings. Cell titles in all settings cells should now resize.
- [x] Run the app with the accessibility inspector. Then use the accessibility inspector to increase the font size. Navigate to Settings -> Account -> Change Password. The Save button on the top right *should not* resize.
- [x] Run the app with the accessibility inspector. Then use the accessibility inspector to increase the font size. Navigate to Settings -> Account -> Change Email. The Save button on the top right *should not* resize.

### DIFF
--- a/Library/Styles/SettingsStyles.swift
+++ b/Library/Styles/SettingsStyles.swift
@@ -8,24 +8,46 @@ public let settingsSectionButtonStyle =
 
 public let settingsArrowViewStyle = UIImageView.lens.tintColor .~ .ksr_dark_grey_400
 
-public let settingsSectionLabelStyle =
-  UILabel.lens.textColor .~ .ksr_soft_black
-    <> UILabel.lens.font .~ .ksr_subhead()
-    <> UILabel.lens.numberOfLines .~ 2
+public let settingsSectionLabelStyle = { (label: UILabel) in
+  label
+    |> \.textColor .~ .ksr_soft_black
+    |> \.font .~ .ksr_subhead()
+    |> \.numberOfLines .~ 2
+}
 
-public let settingsTitleLabelStyle =
-  UILabel.lens.textColor .~ .ksr_soft_black
-    <> UILabel.lens.font .~ .ksr_body()
+public let settingsTitleLabelStyle = { (label: UILabel) in
+  label
+    |> \.textColor .~ .ksr_soft_black
+    |> \.font .~ .ksr_body()
+}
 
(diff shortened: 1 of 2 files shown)


ifbarrera:
[[Dynamic Type Fixes] Settings cell titles should resize](https://trello.com/c/bUQH1Ies/1260-dynamic-type-fixes-settings-cell-titles-should-resize)

ifbarrera:
[[Dynamic Type Fixes] Save Button should not resize](https://trello.com/c/OcLTSb05/1259-dynamic-type-fixes-save-button-should-not-resize)

Library/Styles/SettingsStyles.swift
```diff
@@ -85,15 +103,3 @@ public func settingsAttributedPlaceholder(_ string: String) -> NSAttributedStrin
     attributes: [NSAttributedString.Key.foregroundColor: UIColor.ksr_text_dark_grey_400]
   )
 }
-
-public func settingsHeaderFooterLabelBaseStyle(_ label: UILabel) -> UILabel {
```
dusi: Moving these 2 will probably cause some merge conflicts on the `Create password` branch but now that we know about it we should be able to resolve it just fine!

PR #628, opened 2019-03-04
Strings script tests

Scollaco:
# 📲 What
- Adds tests to our Strings script.

# 🤔 Why
- To avoid headaches in the future 😬 .

# 🛠 How
- Similarly to our `ColorsScript`, a subproject was created using Swift Package Manager, then refactored to make the code testable.
- In order to run the subproject in release mode, first I had to add `Secrets.swift` file to gitignore and then copied the file to the subproject. Note that the gitignore will ignore only the added Secrets, and not the one on the main ksr project.
- Also, the script on the `Makefile` needed to be tweaked, so we can run the `strings` script from the correct directory, in this case, `root_proj_path/bin/StringsScript`

# ✅ Acceptance criteria
- [ ] run `make strings` and the script should work just fine.

# ⏰ TODO
- [ ] Test cloning the project using a non-ksr github account and see if the main project builds.


Scollaco:
[Test strings.swift file](https://trello.com/c/esuvjSdZ/1241-test-stringsswift-file)

bin/StringsScript/Tests/StringsScriptTests/StringsScriptTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  func testDeserialize() {
+    //swiftlint:disable:next todo
+    //TODO
+  }
+
+  func testStaticStringsFileContents() {
+    let stringsByLocale = ["Base": ["Are_you_sure_you_wish_to_remove_this_card": "Are you sure you wish to remove this card from your payment method options?"]]
```
justinswart: Should we include a non-Base locale test in the generation here too?

bin/StringsScript/Tests/StringsScriptTests/StringsScriptTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    if let (locale, content) = strings.localePathsAndContents(with: localePath,
+                                                              stringsByLocale: stringsByLocale).first {
+      XCTAssertEqual(locale, "Kickstarter-iOS/Locales/fr.lproj/Localizable.strings")
+      XCTAssertEqual(content, "\"Kickstarter_is_not_a_store\" = \"Kickstarter n\'est pas un magasin.\";")
+    } else {
+      XCTFail("Locale and content cannot be nil")
+    }
+  }
+
+  func testDeserialize() {
```
justinswart: Still going to add this?

bin/StringsScript/Tests/StringsScriptTests/StringsScriptTests.swift
Scollaco: The `staticStringsFileContents` generates exclusively strings for the Base locale. If we test the function using a non-Base locale, it will simply generate the header comments and an empty `Strings` enum. Do you think it's worth to test it regardless?

justinswart: Oh, I see this line in that function: `try stringsByLocale["Base"]?.keys`. But the same function is used to generate the other locales too. I would want to test this function with a more complete dictionary. Maybe what we really want to do is add some dependency injection to `fetchStringsByLocale()` and pass in some stubbed JSON to test with?

bin/StringsScript/README.md
```diff
@@ -0,0 +1,14 @@
+# StringsScript
+
+## How to contribute to this script:
+
+Most of the fuctionality lives in StringsScriptCore. Any new functionality should also be added to this framework and tested.
+
+## Good things to know:
+
+* You can build this script from the command line by navigating to the root /StringsScript folder and running `swift build`
```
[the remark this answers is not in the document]
justinswart: Might be worth mentioning the step here as well about having to copy `Secrets.swift` over from the main project.

bin/StringsScript/README.md
```diff
@@ -0,0 +1,14 @@
...
+	* you can run the debug binary from the command line using `./.build/debug/StringsScript` (from the root StringsScript folder)
```
justinswart: Which arguments must be passed to the command?
```
$ ./.build/debug/StringsScript
❌ Error: Insufficient arguments
```

Scollaco: `./Library/Strings.swift` and `./Kickstarter-iOS/Locales`. These are the paths where the files will be created. I updated the README.

PR #629, opened 2019-03-05
Friend avatar corrected in Explorer

cdolm92:
# 📲 What

Friend avatar was showing up as a rectangle. This PR corrects that and shows it as a circle.

# 🛠 How

Changed class to `CircleAvatarImageView` which makes the image circular.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|img_ffb507d530d2-1 jpeg|simulator screen shot - iphone 8 - 2019-03-05 at 19 30 38|

# ✅ Acceptance criteria

- [ ] Scroll through `Explorer` screen and check if friend avatar image appears as a circle and not a rectangle

# ⏰ TODO

- [x] Design feedback

### DIFF
--- a/Kickstarter-iOS/Library/UIImageView+URL.swift
+++ b/Kickstarter-iOS/Library/UIImageView+URL.swift
@@ -17,6 +17,17 @@ extension UIImageView {
                      runImageTransitionIfCached: false,
                      completion: nil)
   }
+
+  public func ksr_roundedImageWith(_ url: URL, rounded: Bool = false)  {
+    self.af_setImage(withURL: url,
+                     placeholderImage: nil,
+                     filter: CircleFilter(),
+                     progress: nil,
+                     progressQueue: DispatchQueue.main,
+                     imageTransition: .crossDissolve(0.3),
+                     runImageTransitionIfCached: false,
+                     completion: nil)
+  }
 }
 
 private enum Associations {
--- a/Kickstarter-iOS/Library/UIImageView+URL.swift
+++ b/Kickstarter-iOS/Library/UIImageView+URL.swift
@@ -18,7 +18,7 @@ extension UIImageView {
                      completion: nil)
   }
 
-  public func ksr_roundedImageWith(_ url: URL) {
+  public func ksr_setRoundedImageWith(_ url: URL) {
     self.af_setImage(withURL: url,
                      placeholderImage: nil,
                      filter: CircleFilter(),
--- a/Kickstarter-iOS/Library/UIImageView+URL.swift
+++ b/Kickstarter-iOS/Library/UIImageView+URL.swift
@@ -18,7 +18,7 @@ extension UIImageView {
                      completion: nil)
   }
 
-  public func ksr_roundedImageWith(_ url: URL, rounded: Bool = false)  {
+  public func ksr_roundedImageWith(_ url: URL, rounded: Bool = false) {
     self.af_setImage(withURL: url,
                      placeholderImage: nil,
                      filter: CircleFilter(),
--- a/Kickstarter-iOS/Library/UIImageView+URL.swift
+++ b/Kickstarter-iOS/Library/UIImageView+URL.swift
@@ -18,7 +18,7 @@ extension UIImageView {
                      completion: nil)
   }
 
-  public func ksr_roundedImageWith(_ url: URL, rounded: Bool = false) {
+  public func ksr_roundedImageWith(_ url: URL) {
     self.af_setImage(withURL: url,
                      placeholderImage: nil,
                      filter: CircleFilter(),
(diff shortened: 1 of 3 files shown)


delmabrocales:
Hi! Just a small feedback around padding - can we make friend avatar's left padding match with the content above it so that it's aligned? And also make the top padding uniform to the spacing between the % funded line and category line?
 - hope that helps! Thanks!

cdolm92:
@delmabrocales updated screenshot, please let me know if this is better. Thanks!

dusi:
I've noticed that this avatar appears in a feed (i.e. something that's scrollable)...should we try to optimize this in order to avoid performance hit? Using `cornerRadius` and `masksToBounds` causes offscreen rendering and can cause frame drops.

There are multiple techniques this could be done. One example would be to round the image (https://stackoverflow.com/a/17593733) not the presenting container. We might need to consider caching it properly as well.

Any other ideas?

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -348,4 +349,12 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
   @objc fileprivate func saveButtonTapped(_ button: UIButton) {
     self.watchProjectViewModel.inputs.saveButtonTapped(selected: button.isSelected)
   }
+
+  func roundImage(_ url: URL) -> UIImage? {
```
dusi: `private`?

dusi: Can we also use more descriptive signature for this function?

I.e. `roundImage(for url: URL) { .. }`

delmabrocales:
Spacing looks good to me. Thanks for updating! 😄

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -348,4 +349,12 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
...
+    guard let imageData = try? Data(contentsOf: url) else { return nil }
+
+    let avatar = UIImage(data: imageData, scale: UIScreen.main.scale)?
+      .af_imageRoundedIntoCircle()
```
dusi: Since `AF` uses `rounded` naming would it make sense to also name our function `roundedImage` instead of `roundImage`?

cdolm92: I agree, going to correct! thank you!

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
cdolm92: Yup going to correct this too

Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
ifbarrera: If we look at our convenience func `ksr_setImageWithURL`, it calls `af_setImage` which accepts a `filter` argument. I believe in here, we could pass `CircleFilter()` which comes for free out of the box with Alamofire and wouldn't require us to convert the image url into `Data` before rounding it. One thing we could do is extend the functionality of `ksr_setImageWithURL` to accept an additional optional argument, maybe something like: `ksr_setImageWithURL(_ url: URL, rounded: Bool = false)` and passing the `CircleFilter()` into Alamofire's `filter` argument if `rounded` is `true`.

Another thing to note is that calling `Data(contentsOf: url)` is not recommended for remote url's because it's synchronous, which would [block the main thread](https://developer.apple.com/documentation/foundation/nsdata/1413892-init).

Kickstarter-iOS/Library/UIImageView+URL.swift
```diff
@@ -17,6 +17,17 @@ extension UIImageView {
                      runImageTransitionIfCached: false,
                      completion: nil)
   }
+
+  public func ksr_roundedImageWith(_ url: URL, rounded: Bool = false) {
```
dusi: Is the `rounded` parameter being used?

cdolm92: removed it

Kickstarter-iOS/Library/UIImageView+URL.swift
```diff
@@ -17,6 +17,17 @@ extension UIImageView {
                      runImageTransitionIfCached: false,
                      completion: nil)
   }
+
+  public func ksr_roundedImageWith(_ url: URL) {
```
dusi: One final thing I overlooked (apologies)...since this is a setter on `UIImageView` would it make sense to call this method `ksr_setRoundedImageWith(_ url: URL) { }` to imply that?

cdolm92: fixed that 👍🏾

PR #630, opened 2019-03-05
📨 [Change email] Saving with Done Return Key

cdolm92:
# 📲 What

Changing email and pressing `Done` on the keyboard only dismisses the keyboard. The `Done` on the keyboard now dismisses the keyboard and saves the changes.

# 🤔 Why

This is the desired functionality.

# ✅ Acceptance criteria

Recommended Environment: Your kickstarter account on staging
Recommended device: iPhone 7, 8, X, XS on iOS 11 or 12

Testing w/ **correct** password:
- [x] Navigate to `Settings > Account > Change email`
- [x] Enter and email address tap `Next` on keyboard. Cursor should be on the password textfield.
- [x] `Done` return key on password textfield should be disabled. Start typing your password.
- [x] Tap `Done` keyboard should dismiss and success alert that reads `Got it! Your changes have been saved.` should present.

Testing w/ **incorrect** password:
- [x] Navigate to `Settings > Account > Change email`
- [x] Enter and email address tap `Next` on keyboard. Cursor should be on the password textfield.
- [x] `Done` return key on password textfield should be disabled. Start typing an incorrect password.
- [x] Tap `Done` keyboard should dismiss and error alert that reads `Validation failed: Current password is incorrect.` should present.


Library/ViewModels/ChangeEmailViewModel.swift
```diff
@@ -46,11 +46,17 @@ public protocol ChangeEmailViewModelType {
 public final class ChangeEmailViewModel: ChangeEmailViewModelType, ChangeEmailViewModelInputs,
 ChangeEmailViewModelOutputs {
   public init() {
+    self.dismissKeyboard = Signal.merge(
```
cdolm92: Issue w/ this is that it will update the email even if the save button is disabled. The save button is disabled when the user enters their current email address which prevents them from "updating" their email w/ their current email. The user can still "update" their email through the return key. currently working on fixing this bug but would be happy to get feedback on this.

Library/ViewModels/ChangeEmailViewModel.swift
```diff
@@ -46,11 +46,17 @@ public protocol ChangeEmailViewModelType {
...
+        .filter { $0 == .done }
+        .ignoreValues(),
+      self.saveButtonTappedProperty.signal.ignoreValues())
+
     let changeEmailEvent = Signal.combineLatest(
       self.newEmailProperty.signal.skipNil(),
       self.passwordProperty.signal.skipNil()
       )
-      .takeWhen(self.saveButtonTappedProperty.signal)
+      .takeWhen(self.dismissKeyboard)
```
ifbarrera: Hmm, I think the desired behavior is that the `changeEmailEvent` should fire if *either* the save button is tapped, *or* the current password text field returns, but *only* if the save button is enabled, otherwise it's sending a request to the server without going through validation first.

PR #631, opened 2019-03-06
Add UnionPay support

Scollaco:
# 📲 What
We now support China UnionPay cards!

# 🛠 How
Created a `.unionPay` type to the  GraphUserCreditCard enum and added it to the `supportedCardBrands` array on AddNewCardViewController.

# 👀 See
union_pay_support

# ✅ Acceptance criteria
## On staging
- [x] Go to `Account > Payment Methods > Add New Card` and add a UnionPay card test (number `6200 0000 0000 0005`)
- [x] You should be brought back to the Payment Methods screen and see the success banner.
- [x] You should see the UnionPay card on the list.


Library/PKPaymentAuthorizationViewController.swift
```diff
@@ -27,13 +27,14 @@ extension PKPaymentAuthorizationViewController {
 
     if countryCode == "US" && project.country != Project.Country.us ||
       countryCode != "US" {
-      return PKPaymentAuthorizationViewController.supportedNetworks.filter { $0 != .discover }
+      return PKPaymentAuthorizationViewController.supportedNetworks
+        .filter { $0 != .discover && $0 != .chinaUnionPay }
```
Scollaco: According to this PR https://github.com/kickstarter/kickstarter/blob/master/browser/javascripts/stripe-elements/credit-cards/utils.ts , UnionPay should not be supported for non-US projects.

Scollaco:
[Allow backers to add a UnionPay card in Account Settings](https://trello.com/c/DhaRDOUd/1264-allow-backers-to-add-a-unionpay-card-in-account-settings)

Scollaco:
[Allow backers to add a UnionPay card in Account Settings](https://trello.com/c/DhaRDOUd/1264-allow-backers-to-add-a-unionpay-card-in-account-settings)

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -25,7 +25,7 @@ STPPaymentCardTextFieldDelegate, MessageBannerViewControllerPresenting {
   @IBOutlet private weak var zipcodeView: SettingsFormFieldView!
 
   private let supportedCardBrands: [STPCardBrand] = [.visa, .masterCard, .amex, .dinersClub,
-                                                     .discover, .JCB]
+                                                     .discover, .JCB, .unionPay]
```
dusi: Could we order these alphabetically and maybe put each on a separate line for better readability?

KsApi/models/GraphUserCreditCard.swift
```diff
@@ -24,17 +24,22 @@ public struct GraphUserCreditCard: Swift.Decodable {
...
     case visa = "VISA"
-    case diners = "DINERS"
     case generic = "----"
 
     public var description: String? {
       switch self {
       case .amex, .discover, .jcb, .mastercard, .visa, .diners:
         return self.rawValue.capitalized
+      case .unionPay:
+        return self.rawValue
```
dusi: Do you think it would make sense to introduce a helper function on a String object something like

```
extension String {
  func snakeToCamel() {
    return self.replacingOccurences(of: "_", with: " ")
  }
}
```

which could be testable..so that then we can simply use `return self.rawValue.snakeToCamel().capitalized`?

Library/PKPaymentAuthorizationViewController.swift
dusi: This is starting to look really confusing since we do have 

`static func supportedNetworks(for project: Project) ...` 

and then

`static func supportedNetworks`

I wonder if it would be clearer with something like this:

```
import KsApi
import PassKit

extension PKPaymentAuthorizationViewController {

  public static var merchantIdentifier: String {
    return "merchant.com.kickstarter"
  }

  public static func applePayCapable() -> Bool {
    return PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: PaymentNetworks.all)
  }

  public static func applePayDevice() -> Bool {
    return PKPaymentAuthorizationViewController.canMakePayments()
  }

  public static func applePayCapable(for project: Project) -> Bool {
    return PKPaymentAuthorizationViewController.canMakePayments(
      usingNetworks: PKPaymentAuthorizationViewController.supportedNetworks(for: project)
    )
  }

  public static func supportedNetworks(for project: Project) -> [PKPaymentNetwork] {
    let countryCode = AppEnvironment.current.countryCode

    if (countryCode == "US" && project.country != Project.Country.us) || countryCode != "US" {
      return PaymentNetworks.US.supported
    }

    return PaymentNetworks.all
  }
}

// MARK: - Payment networks

// swiftlint:disable private_over_fileprivate
fileprivate struct PaymentNetworks {
  fileprivate static var all: [PKPaymentNetwork] {
    return [.amex, .masterCard, .visa, .discover, .chinaUnionPay]
  }

  fileprivate struct US {
    fileprivate static var unsupported: [PKPaymentNetwork] {
      return [.chinaUnionPay, .discover]
    }

    fileprivate static var supported: [PKPaymentNetwork] {
      return PaymentNetworks.all.filter { US.unsupported.contains($0) }
    }
  }
}
// swiftlint:enable private_over_fileprivate
```

What do you think Nino? (this is just a suggestion, I'd be OK with this approach for now)

dusi: Also I would consider putting `countryCode == "US" && project.country != Project.Country.us` in parenthesis for better readability.

KsApi/models/GraphUserCreditCard.swift
Scollaco: I think that for now, since that's the only place where this transformation needs to be made, we don't need to extract the code to a helper function. It's a good idea though, in case we use the same logic somewhere else.

Library/PKPaymentAuthorizationViewController.swift
Scollaco: This is so much better! Thanks, @dusi !

PR #633, opened 2019-03-08
👩‍🔬 [Investment Day] Swiftlint Cleanup

dusi:
# 📲 What

* Removes unused rules
* Prefers to use `next` where possible

# 🔨 How

Where possible I've replaced `// swiftlint:disable` with `// swiftlint:disable:next` or disable block. If a specific file disabled a rule for the whole file (at the beginning of the file), I would have replaced this with `next` only if the rule wasn't broken more than 5 times. If it was broken more times it was an indicator that the file was assuming / working differently than we would like which is a work for some other time.

# 🤔 Why

Best practices

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| 0 Swiftlint warnings after `Clean Build` | 0 Swiftlint warnings after `Clean Build`  |

# ✅ Acceptance criteria

- [x] All tests pass
- [ ] No `Swiftlint` warnings

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -191,7 +191,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .observeForUI()
       .observeValues { [weak self] in self?.findRedirectUrl($0) }
 
-    //swiftlint:disable discarded_notification_center_observer
+    // swiftlint:disable discarded_notification_center_observer
     NotificationCenter.default
       .addObserver(forName: Notification.Name.ksr_sessionStarted, object: nil, queue: nil) { [weak self] _ in
         self?.viewModel.inputs.userSessionStarted()
@@ -207,7 +207,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .addObserver(forName: Notification.Name.ksr_sessionEnded, object: nil, queue: nil) { [weak self] _ in
         self?.viewModel.inputs.userSessionEnded()
     }
-    //swiftlint:enable discarded_notification_center_observer
+    // swiftlint:enable discarded_notification_center_observer
 
     self.window?.tintColor = .ksr_green_700
 
--- a/Kickstarter-iOS/DataSources/SettingsAccountDataSourceTests.swift
+++ b/Kickstarter-iOS/DataSources/SettingsAccountDataSourceTests.swift
@@ -1,4 +1,3 @@
-// swiftlint:disable line_length
 import Prelude
 import XCTest
 import Library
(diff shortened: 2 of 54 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -191,7 +191,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .observeForUI()
       .observeValues { [weak self] in self?.findRedirectUrl($0) }
 
-    //swiftlint:disable discarded_notification_center_observer
+    // swiftlint:disable discarded_notification_center_observer
```
dusi: For consistency with the rest of the codebase `// swiftlint` is preferred to `//swiftlint`

Kickstarter-iOS/DataSources/SettingsAccountDataSourceTests.swift
```diff
@@ -41,10 +40,18 @@ final class SettingsAccountDataSourceTests: XCTestCase {
                                   shouldHideEmailWarning: true,
                                   shouldHideEmailPasswordSection: false)
 
-    XCTAssertEqual(SettingsAccountCellType.changeEmail, self.dataSource.cellTypeForIndexPath(indexPath: indexPath1))
-    XCTAssertEqual(SettingsAccountCellType.changePassword, self.dataSource.cellTypeForIndexPath(indexPath: indexPath2))
-    XCTAssertEqual(SettingsAccountCellType.privacy, self.dataSource.cellTypeForIndexPath(indexPath: indexPath3))
-    XCTAssertEqual(SettingsAccountCellType.paymentMethods, self.dataSource.cellTypeForIndexPath(indexPath: indexPath4))
+    XCTAssertEqual(
```
dusi: Splitting up the lines didn't make the readability worse and allowed us to get rid of the rule altogether.

KsApi/MockService.swift
```diff
@@ -1455,7 +1455,6 @@ private extension MockService {
       }
     )
   }
-  // swiftlint:enable type_name
```
dusi: Unused rule...there were couple others in this PR.

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -62,11 +62,7 @@ PaymentMethodsViewModelInputs, PaymentMethodsViewModelOutputs {
     self.showAlert = deletePaymentMethodEventsErrors
       .ignoreValues()
       .map {
-        localizedString(
-          key: "Something_went_wrong_and_we_were_unable_to_remove_your_payment_method_please_try_again",
-          //swiftlint:disable:next line_length
-          defaultValue: "Something went wrong and we were unable to remove your payment method, please try again."
-        )
+        Strings.Something_went_wrong_and_we_were_unable_to_remove_your_payment_method_please_try_again()
```
dusi: Looks like localized string was available but we didn't use it 🤷‍♂️

Scollaco: Actually it was being used. The `localizedString(key:defaultValue:)` function is exactly what `Something_went_wrong...` string function returns. If I'm not wrong, this was implemented that way so we could merge the PR before having the strings translated. But good catch though. 😄

PR #634, opened 2019-03-08
👩‍🔬 [Investment Day] Clean up compiler warnings

dusi:
# 📲 What

Removes type check warnings by helping the compiler by explicitly setting types.

# 🤔 Why

It's considered best practice to keep warnings down to minimum in order to prevent from warnings bankruptcy which usually causes more warnings to slip into the codebase unnoticed.

# 🛠 How

Our project build settings uses experimental build flag `-warn-long-function-bodies=1000` (that's 1000ms) to set a type evaluation threshold for the compiler. Every time compiler tries to infer Swift types its evaluation time is considered against this threshold. If it takes the compiler longer than 1000ms, Xcode will show a warning. More about the topic [here](http://khanlou.com/2016/12/guarding-against-long-compiles).

Screen Shot 2019-03-08 at 3 38 38 PM

In general this is a good practice. It helps us identify code that could take longer to infer. Unfortunately it appears that sometimes it's necessary to help the compiler a little. This is pretty simple and can be done by stating types explicitly.

It also appears that the most problematic area of our code base (that was causing the most warnings were the UI styles). As part of this PR I've tried to explicitly set types on these styles in order to help the compiler. This successfully suppressed compiler warnings.

I would like to propose that in the future we try to define our styles with using less of type inference and explicitly stating the types. This might add up to the code but ideally this would only happen in the styles sheet. In addition it might also help making it more readable.

I'd love to have a discussion around this topic and hear everyone's thoughts if possible before we decide what to do: @ifbarrera @Scollaco @cdolm92. This will affect the speed with which Xcode can infer types and therefore auto complete code (which is currently pretty slow).

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| 13 warnings 😞  | 5 warnings 👍  |
| Screen Shot 2019-03-08 at 3 40 15 PM | Screen Shot 2019-03-08 at 3 37 58 PM |

# ✅ Acceptance criteria

- [ ] All tests pass

### DIFF
--- a/Library/Styles/BaseStyles.swift
+++ b/Library/Styles/BaseStyles.swift
@@ -111,19 +111,28 @@ public let feedTableViewCellStyle = baseTableViewCellStyle()
       : .init(topBottom: Styles.gridHalf(3), leftRight: Styles.grid(2))
 }
 
-public let formFieldStyle =
-  UITextField.lens.font .~ .ksr_body()
-    <> UITextField.lens.textColor .~ .ksr_soft_black
-    <> UITextField.lens.backgroundColor .~ .clear
-    <> UITextField.lens.borderStyle .~ .none
-    <> UITextField.lens.autocapitalizationType .~ .none
-    <> UITextField.lens.autocorrectionType .~ .no
-    <> UITextField.lens.spellCheckingType .~ .no
-    <> UITextField.lens.tintColor .~ .ksr_green_700
+public let formTextInputStyle: (UITextField) -> UITextField = { (textField: UITextField) in
+  textField
+    |> \.autocapitalizationType .~ UITextAutocapitalizationType.none
+    |> \.autocorrectionType .~ UITextAutocorrectionType.no
+    |> \.spellCheckingType .~ UITextSpellCheckingType.no
+}
 
-public let separatorStyle =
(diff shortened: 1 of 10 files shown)


dusi:
Also look into this option https://lapcatsoftware.com/articles/validate-project-settings-never.html

Library/Styles/BaseStyles.swift
```diff
@@ -111,19 +111,28 @@ public let feedTableViewCellStyle = baseTableViewCellStyle()
...
-    <> UITextField.lens.borderStyle .~ .none
-    <> UITextField.lens.autocapitalizationType .~ .none
-    <> UITextField.lens.autocorrectionType .~ .no
-    <> UITextField.lens.spellCheckingType .~ .no
-    <> UITextField.lens.tintColor .~ .ksr_green_700
-
-public let separatorStyle =
-  UIView.lens.backgroundColor .~ .ksr_grey_400
-    <> UIView.lens.accessibilityElementsHidden .~ true
+public let formTextInputStyle: (UITextField) -> UITextField = { (textField: UITextField) in
```
dusi: Use alias for.

dusi:
TODO: Try running with build time measurement

PR #636, opened 2019-03-13
Pull to refresh

Scollaco:
# 📲 What
Allows users to refresh Discovery Page by dragging the top of the list downward. (UIRefreshControl).

# 🤔 Why
🌟 Product Excellence
(Details here: https://trello.com/c/2sWEdiyF/1265-pull-down-on-the-feed-to-refresh-the-feed)

# 🛠 How
- Creating an instance of `UIRefreshControl` and implementing a function to handle its action.
- The value of `clearOnNewRequest` had to be set to `false` to avoid the list to be blank while pulling to refresh projects.
- Some logic was added to handle the states of UIRefreshControl and ActivityIndicator, since the second shouldn't appear while loading projects through `pull to refresh` action.

# 👀 See
pull_to_refresh

# ✅ Acceptance criteria

- [ ] I go to Discovery and I see the `Activity Indicator` while loading projects.
- [ ] I drag the list downward and I should see the `UIRefreshControl` spinning (and no `Activity Indicator`)
- [ ] The projects rarely change, so to see if the feature is working, go to the `Ending soon` tab and look for a project the `x min to go`. Wait one minute and pull to refresh. The ending time should update.


dusi:
So my first impression (looking at this improvement) is whether we should simply use just one spinner (the pull to refresh control's spinner to be more precise).

We currently show 2 loading indicators:
* one I believe is either a view or cell that's displayed when the content is being loaded
* the other is the pull to refresh

Hopefully this explains it better than my words..

| Step 1 | Step 2 | Step 3 |
| --- | --- | --- |
| Screen Shot 2019-03-14 at 12 19 00 PM | Screen Shot 2019-03-14 at 12 18 35 PM | Screen Shot 2019-03-14 at 1 07 19 PM |


How hard would it be to remove the original spinner and keep just the `UIRefreshControl` one?

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -87,26 +92,21 @@ internal final class DiscoveryPageViewController: UITableViewController {
...
   }
 
     internal override func bindViewModel() {
     super.bindViewModel()
 
-    self.loadingIndicatorView.rac.animating = self.viewModel.outputs.projectsAreLoading
+    self.viewModel.outputs.projectsAreLoading
+      .observeForUI()
+      .observeValues { [weak self] in
+        $0 ? self?.refreshControl?.beginRefreshing() : self?.refreshControl?.endRefreshing()
```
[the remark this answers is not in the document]
dusi: I think we'll also have to pass back one more attribute (or use a different signal) for when this is manually triggered refresh.

manual = user pulled down to refresh
automatic = first time load of the page

Otherwise there's a weird bounce animation when the loading is done and the page control is being hidden. 

On the left - old behaviour | On the right new behaviour
Untitled

There is a workaround for this:

```
self.viewModel.outputs.projectsAreLoading
  .observeForUI()
  .observeValues { [weak self] (isLoading, animated) in
    if isLoading {
      self?.performClosureAnimated(animated, closure: { self?.refreshControl?.endRefreshing() })
    } else {
      self?.performClosureAnimated(animated, closure: { self?.refreshControl?.beginRefreshing() })
    }
}
```

where we have a private function

```
private func performClosureAnimated(_ animated: Bool, closure: () -> Void) {
  if animated {
    closure()
  } else {
    UIView.performWithoutAnimation { closure() }
  }
}
```

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -7,13 +7,14 @@ import UIKit
...
   fileprivate var emptyStatesController: EmptyStatesViewController?
   fileprivate let dataSource = DiscoveryProjectsDataSource()
-  fileprivate let loadingIndicatorView = UIActivityIndicatorView()
   private var sessionEndedObserver: Any?
   private var sessionStartedObserver: Any?
   private var currentEnvironmentChangedObserver: Any?
   fileprivate let viewModel: DiscoveryPageViewModelType = DiscoveryPageViewModel()
   fileprivate let shareViewModel: ShareViewModelType = ShareViewModel()
 
+  private let discoveryRefreshControl = UIRefreshControl()
```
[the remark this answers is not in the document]
Scollaco: I tried that before, but the refresher didn't appear 🤔

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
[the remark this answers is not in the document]
Scollaco: This one is a little trickier because is involves multiple signals. Will have to take some time to work on this one.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -277,10 +279,22 @@ internal final class DiscoveryPageViewController: UITableViewController {
     guard let navigator = self.presentedViewController as? ProjectNavigatorViewController else { return }
     navigator.updatePlaylist(playlist)
   }
+
+  @objc private func pulledToRefresh() {
+    self.viewModel.inputs.pulledToRefresh()
+  }
+
+  private func performClosureAnimated(_ animated: Bool, closure: () -> Void) {
```
dusi: This has no dependency on the instance of `DiscoveryPageViewController` .. could we extract it outside of the class (at the bottom of this file)..then instead of calling it with `self?.` like

```
self?.performClosureAnimated(...)
```
we can simply do
```
performClosureAnimated(...)
```

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -142,36 +145,58 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
+      isLoading,
+      isLoading.takeWhen(isRefreshing).mapConst(false)
+    ).skipRepeats()
+
+    self.projectsAreLoadingAnimated = Signal.merge(
+      isRefreshing.map { ($0, true) },
+      projectsLoadingNoRefresh.map { ($0, false) },
+      self.viewWillAppearProperty.signal.take(first: 1).mapConst((true, false))
+    )
+    .skipRepeats(==)
```
dusi: Can you explain this line and in particular `(==)`?

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -142,36 +145,58 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
 
-    self.showEmptyState = Signal.combineLatest(paramsChanged, self.projectsAreLoading, paginatedProjects)
+    let isRefreshing = isLoading
+      .combineLatest(with: self.pulledToRefreshProperty.signal)
+      .map { $0.0 }
+      .skipRepeats()
+
+    let projectsLoadingNoRefresh = Signal.merge(
+      isLoading,
+      isLoading.takeWhen(isRefreshing).mapConst(false)
```
dusi: Isn't this `isLoading` duplicate signal? Can you explain this?

Library/ViewModels/DiscoveryPageViewModel.swift
Scollaco: In some cases, we need to pass a function that determines that values from the signal are equivalent. Without the `==`, the compiler will show the `missing argument` error.

Library/ViewModels/DiscoveryPageViewModel.swift
Scollaco: While `isLoading` will get updated every time the projects are loading or stop loading, the `isLoading.takeWhen(isRefreshing).mapConst(false)` signal will only emit once the user pulls to refresh. Without the `isLoading`, the `projectsLoadingNoRefresh` won't emit until the user pulls to refresh and the UIRefreshControl won't spin. Does it make sense?

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewControllerTests.swift
```diff
@@ -34,6 +34,7 @@ internal final class DiscoveryPageViewControllerTests: TestCase {
                       userDefaults: MockKeyValueStore()) {
 
           let controller = DiscoveryPageViewController.configuredWith(sort: .magic)
+            controller.tableView.refreshControl = nil
```
Scollaco: Hides the refreshControl so the screenshot is not changed.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -87,26 +90,25 @@ internal final class DiscoveryPageViewController: UITableViewController {
...
-    internal override func bindViewModel() {
+  internal override func bindViewModel() {
     super.bindViewModel()
 
-    self.loadingIndicatorView.rac.animating = self.viewModel.outputs.projectsAreLoading
+    self.viewModel.outputs.projectsAreLoadingAnimated
+      .observeForUI()
+      .observeValues { [weak self] (isLoading, animated) in
+        if isLoading {
+          self?.performClosureAnimated(true, closure: { self?.refreshControl?.beginRefreshing() })
```
ifbarrera: Is it possible to name this function more specifically? Something like 
```
func updateRefreshControl(animated: Bool, _ closure: () -> Void)
```

If we remove the `closure` parameter's argument label, it could be called like this:

```
          self?.updateRefreshControl(animated: animated, {
            self?.refreshControl?.beginRefreshing()
          })
```

Which is less verbose.

ifbarrera: @Scollaco how come the `performClosureAnimated` function is called with `animated: true` when `isLoading` is `true`, but with `animated: animated` when `isLoading` is false? Is this logic something we could put in the view model so it's not hard-coded on the VC side?

Scollaco: If the refreshControl is beginning to refresh, the `animate` property has to be `true`, otherwise it will appear on the screen without animation.

PR #637, opened 2019-03-15
🔑 [Create Password] GraphMutation call on Save 

cdolm92:
# 📲 What

Uses `Create password` graph mutation when `Save` button is tapped in order to call the endpoint. If everything is successful, logs the user out.

# 🤔 Why

Facebook users w/o a password can now create a password.

# 🛠 How

`createPasswordEvent` in the `CreatePasswordViewModel` takes the input from the textfields and inputs it to `CreatePasswordInput`.

# 👀 See

create-password2

# ✅ Acceptance criteria

Please make sure to delete your password before testing. 
**Test Environment** : Staging

**Create Password w/ Keyboard ReturnKey**
- [x] Navigate to `Settings > Account > Create Password`
- [x] Keyboard should display and cursor should be in `New Password` textfield. Enter a password and press `Next` return key 
- [x] Keyboard should still display and cursor should now appear in the `Confirm new password` textfield. Enter the same password as above and press the `Done` return key.   `Done` return key should also enable when there is text in the field
- [x] Loader is shown where the save button was, Keyboard should dismiss and login screen is displayed.
- [x] Login with the newly created password. 
- [x] Navigate to `Settings > Account` you should see the `Change Email` and `Change Password` buttons on screen.

**Create Password w/ Save Button**
- [x] Navigate to `Settings > Account > Create Password`
- [x] Keyboard should display and cursor should be in `New Password` textfield. Enter a password and press `Next` return key 
- [x] Keyboard should still display and cursor should now appear in the `Confirm new password` textfield. Enter the same password as above and press the `Save` button.   `Save` button should be enabled when you enter a valid/matching password.
- [x] Loader is shown where the save button was, Keyboard should dismiss and login screen is displayed.
- [x] Login with the newly created password. 
- [x] Navigate to `Settings > Account` you should see the `Change Email` and `Change Password` buttons on screen.


Library/ViewModels/CreatePasswordViewModelTests.swift
```diff
@@ -39,19 +47,47 @@ final class CreatePasswordViewModelTests: TestCase {
   }
 
   func testCreatePassword() {
+    let service = MockService()
```
dusi: Do we need a local variable since we're not reusing it later on?

Library/ViewModels/CreatePasswordViewModel.swift
```diff
@@ -82,6 +87,32 @@ CreatePasswordViewModelInputs, CreatePasswordViewModelOutputs {
...
+    }
+
+    self.createPasswordSuccess = createPasswordEvent.values().ignoreValues()
+
+    self.activityIndicatorShouldShow = Signal.merge(
+      triggerSaveAction.signal.mapConst(true),
+      self.createPasswordSuccess.mapConst(false)
+    )
+
+    self.dismissKeyboard = Signal.merge(self.saveButtonTappedProperty.signal,
```
dusi: Looks like there are couple events that are repeated here...wonder if we can simplify this a bit.

```
// This is being called when user taps 'Save' or `Done'
let submitFormEvent = Signal.merge(
  self.newPasswordConfirmationDidReturnProperty.signal,
  self.saveButtonTappedProperty.signal
)

// So the following action isn't only concerned about 'Done' key but both 'Done' & 'Save'
// 'autoSaveSignal' becomes 'saveAction'
// allows us to remove 'triggerSaveAction' altogether
let saveAction = formIsValid
  .takeWhen(submitFormEvent)
  .filter(isTrue)
  .ignoreValues()

let createPasswordEvent = combinedPasswords
  .takeWhen(saveAction)
  .map { CreatePasswordInput(password: $0.0, passwordConfirmation: $0.1) }
  .ignoreValues()
  ... // call graph here //

self.createPasswordSuccess = createPasswordEvent

self.activityIndicatorShouldShow = Signal.merge(
  saveAction.signal.mapConst(true),
  self.createPasswordSuccess.mapConst(false)
)

self.dismissKeyboard = submitFormEvent
```

Also I haven't really seen us naming variables with as `somethingSignal` but rather `somethingEvent` or `somethingAction` so wonder if for consistency we should not use `autoSaveSignal`?

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -143,9 +160,33 @@ final class CreatePasswordViewController: UITableViewController {
...
     }
   }
 
+  private func logoutAndDismiss() {
+    AppEnvironment.logout()
+    PushNotificationDialog.resetAllContexts()
+
+    NotificationCenter.default.post(.init(name: .ksr_sessionEnded))
+
+    self.dismiss(animated: true, completion: nil)
```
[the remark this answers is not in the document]
cdolm92: To animate the transition.

Library/ViewModels/CreatePasswordViewModelTests.swift
cdolm92: Nope, removed it!

Library/ViewModels/CreatePasswordViewModel.swift
cdolm92: I agree, renamed it.

cdolm92: Also not going to `ignoreValues()` of the `createPasswordEvent` because I'll need the error description later.

Library/SharedFunctions.swift
```diff
@@ -174,3 +174,13 @@ public func ksr_isOSVersionAvailable(_ version: Double) -> Bool {
 
   return false
 }
+
+public func logoutAndDismiss(_ viewController: UIViewController) {
```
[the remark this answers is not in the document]
dusi: Also, I wasn't really able to think about better ways to test this than the following

First we need to declare protocols that `AppEnvironment` and `PushNotificationDialogType` would conform to (these should follow our project convention just like the `ServiceType` protocol - so I'm thinking these should be two separate files `AppEnvironmentProtocol` & `PushNotificationDialogType` side by side their representative implementations..

```
// AppEnvironmentType.swift
public protocol AppEnvironmentType {
  static func logout()
}
```

```
// PushNotificationDialogType.swift
public protocol PushNotificationDialogType {
  static func resetAllContexts()
}
```

then we need to tweak our `logoutAndDismiss` function to allow injection of types that conform to the above. This is so that we can provide mock structs like so..

```
public func logoutAndDismiss(viewController: UIViewController,
                             appEnvironment: AppEnvironmentType.Type = AppEnvironment.self,
                             pushNotificationDialog: PushNotificationDialogType.Type = PushNotificationDialog.self) {
  appEnvironment.logout()

  pushNotificationDialog.resetAllContexts()

  NotificationCenter.default.post(.init(name: .ksr_sessionEnded))

  viewController.dismiss(animated: true, completion: nil)
}
```

The above will allow us to mock `AppEnvironment` and `PushNotificationDialog` in order to write a real test for those. I'm leaving out `Notification Center` for now coz it's a little bit more work. It should also not change the calcite since we're passing default values.

So that for our testing purposes in `SharedFunctionsTests.swift` we can declare our mocks...

```
struct MockAppEnvironment: AppEnvironmentType {
  static var logoutWasCalled = false

  static func logout() {
    self.logoutWasCalled = true
  }
}

struct MockPushNotificationDialog: PushNotificationDialogType {
  static var resetAllContextsWasCalled = false

  static func resetAllContexts() {
    self.resetAllContextsWasCalled = true
  }
}

class MockViewController: UIViewController {
  var dismissAnimatedWasCalled = false

  override func dismiss(animated flag: Bool, completion: (() -> Void)? = nil) {
    self.dismissAnimatedWasCalled = true
  }
}
```

and use them to test the function like so. This integration test makes sure that each component does what we want it to do...but we don't ask any further questions about its internal behaviour.

```
func testLogoutAndDismiss() {
    let mockAppEnvironment = MockAppEnvironment.self
    let mockPushNotificationDialog = MockPushNotificationDialog.self
    let mockViewController = MockViewController()

    XCTAssertFalse(mockAppEnvironment.logoutWasCalled)
    XCTAssertFalse(mockPushNotificationDialog.resetAllContextsWasCalled)
    XCTAssertFalse(mockViewController.dismissAnimatedWasCalled)
    mockAppEnvironment.logout()
    XCTAssertTrue(mockAppEnvironment.logoutWasCalled)
    XCTAssertTrue(mockPushNotificationDialog.resetAllContextsWasCalled)
    XCTAssertTrue(mockViewController.dismissAnimatedWasCalled)
  }
```

I'd be happy to walk you through it.

What do you think?

Kickstarter-iOS/AppEnvironmentType.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import Library
+
+public protocol AppEnvironmentType {
+  static func logout()
+}
+
+extension AppEnvironment: AppEnvironmentType {
```
dusi: What would you say about moving this to the `AppEnvironment` file instead? It will be more clear if the struct itself is defined as conforming to this protocol, instead of hiding it here in the protocol file?

So if we do this we can remove the `import Library` here.

and move the conformance to

```
// AppEnvironment.swift
public struct AppEnvironment: AppEnvironmentType { .. }
```

This will then follow the same conventions as `Service` & `ServiceType`

Kickstarter-iOS/AppEnvironmentType.swift
```diff
@@ -0,0 +1,12 @@
...
+  public var appEnvironment: AppEnvironment {
```
dusi: Is this necessary?

dusi:
I've noticed that these files are not in the right folder, can we add them to the right folder and sort them alphabetically?

* `SharedFunctions` + `SharedFunctionsSwift`
* `AppEnvironmentType`
* `PushNotificationDialogType`

Kickstarter-iOS/AppEnvironmentType.swift
dusi: Looks like it compiles even if I remove these variables

Kickstarter-iOS/SharedFunctionsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+    let mockAppEnvironment = MockAppEnvironment.self
+    let mockPushNotificationDialog = MockPushNotificationDialog.self
+    let mockViewController = MockViewController()
+
+    XCTAssertFalse(mockAppEnvironment.logoutWasCalled)
+    XCTAssertFalse(mockPushNotificationDialog.resetAllContextsWasCalled)
+    XCTAssertFalse(mockViewController.dismissAnimatedWasCalled)
+    mockAppEnvironment.logout()
+    XCTAssertTrue(mockAppEnvironment.logoutWasCalled)
+    mockPushNotificationDialog.resetAllContexts()
```
dusi: Since we want to test this function

```
func logoutAndDismiss(
  viewController: UIViewController,
  appEnvironment: AppEnvironmentType.Type = AppEnvironment.self,
  pushNotificationDialog: PushNotificationDialogType.Type = PushNotificationDialog.self) { .. }
```

we should simply call it with the mock objects like so

```
logoutAndDismiss(viewController: mockViewController, appEnvironment: mockAppEnvironment, pushNotificationDialog: mockPushNotificationDialog)
```

this will make sure that `AppEnvironment` calls `logout`, `PushNotificationDialog` calls `resetAlContexts()` and `viewController` will call `dismiss(animated: true, completion: nil)`.

So the final test should probably look like this
```
func testLogoutAndDismiss() {
    let mockAppEnvironment = MockAppEnvironment.self
    let mockPushNotificationDialog = MockPushNotificationDialog.self
    let mockViewController = MockViewController()

    XCTAssertFalse(mockAppEnvironment.logoutWasCalled)
    XCTAssertFalse(mockPushNotificationDialog.resetAllContextsWasCalled)
    XCTAssertFalse(mockViewController.dismissAnimatedWasCalled)
    logoutAndDismiss(viewController: mockViewController, appEnvironment: mockAppEnvironment, pushNotificationDialog: mockPushNotificationDialog)
    XCTAssertTrue(mockAppEnvironment.logoutWasCalled)
    XCTAssertTrue(mockPushNotificationDialog.resetAllContextsWasCalled)
    XCTAssertTrue(mockViewController.dismissAnimatedWasCalled)
  }
```

Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -43,10 +43,10 @@ final class CreatePasswordViewController: UITableViewController {
   private weak var newPasswordTextField: UITextField?
   private weak var newPasswordConfirmationTextField: UITextField?
   private weak var groupedFooterView: SettingsGroupedFooterView?
-
-  private lazy var rightBarButtonItem: UIBarButtonItem = {
-    UIBarButtonItem(barButtonSystemItem: .save, target: nil, action: nil)
-      |> \.isEnabled .~ false
+  private lazy var saveButtonView: LoadingBarButtonItemView = {
+   let buttonView = LoadingBarButtonItemView.instantiate()
```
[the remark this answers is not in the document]
dusi: Since we don't have a linter, can we also fix the spacing?

Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+    XCTAssertFalse(mockViewController.dismissAnimatedWasCalled)
+    logoutAndDismiss(viewController: mockViewController, appEnvironment: mockAppEnvironment,
+                     pushNotificationDialog: mockPushNotificationDialog)
+    XCTAssertTrue(mockAppEnvironment.logoutWasCalled)
+    XCTAssertTrue(mockPushNotificationDialog.resetAllContextsWasCalled)
+    XCTAssertTrue(mockViewController.dismissAnimatedWasCalled)
+  }
+}
+
+struct MockAppEnvironment: AppEnvironmentType {
```
dusi: These mock objects should probably be `private` since we don't need to reuse them anywhere else in our test suite.

PR #642, opened 2019-03-21
Request personal data copy change

cdolm92:
# 📲 What

New copy informs user of specific email account to check when requesting their data.

# 🛠 How

Graph call to fetch user email in `SettingsRequestDataCellViewModel`

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|Simulator Screen Shot - iPhone 8 - 2019-03-21 at 15 36 40|Simulator Screen Shot - iPhone 8 - 2019-04-01 at 13 33 10|

# ✅ Acceptance criteria

- [x] Navigate to `Settings > Account > Privacy > Request my Personal Data`
- [x] Alert should present w/ copy that reads "You'll receive an email at [account email] when your download is ready."
- [x] CTA on alert should read "Request my data"


Library/ViewModels/SettingsRequestDataCellViewModel.swift
```diff
@@ -39,16 +39,28 @@ public final class SettingsRequestDataCellViewModel: SettingsRequestDataCellView
...
+
     let exportEnvelope = initialUser
       .switchMap { _ in
         AppEnvironment.current.apiService.exportDataState()
           .demoteErrors()
     }
 
-    self.showRequestDataPrompt = exportEnvelope
+    self.showRequestDataPrompt = Signal.combineLatest(exportEnvelope, requestDataAlertText)
+      .filter { $0.0.dataUrl == nil || $0.0.state == .expired || $0.0.expiresAt == nil }
```
Scollaco: Since you touched this code, I think we can improve the readability of it by extracting the logic to a new function
```
private func canRequestData(_ envelope: ExportDataEnvelope) -> Bool {
  return envelope.dataUrl == nil || envelope.state == .expired || envelope.expiresAt == nil
}
```
and simply filter using `.filter { canRequestData(0.0) }`

PR #643, opened 2019-03-26
🔑 [Create Password] Error handling

dusi:
# 📲 What

Shows errors in the message banner.

# 🤔 Why

Should something go wrong with the password creation we have to show alert to the user indicating something went wrong.

# 🛠 How

Unfortunately, it wasn't possible to present the alert on `UITableView` as we've previously done with `UIView` containers, so this PR had to embed `CreatePasswordTableViewController` (previously `CreatePasswordViewController`) in (now) `CreatePasswordViewController` container. This container is now responsible for presenting the screen title and save button (which is interacted with through a delegate functions).

# 👀 See

| 🚨Shows an alert if something goes wrong |
| --- |
| Screen Shot 2019-03-26 at 12 14 41 PM
 |

# ♿️ Accessibility 

- [x] VoiceOver properly focuses on the alert banner after it's been presented

# ✅ Acceptance criteria

- [x] Performing `Create password` operation while offline (Airplane mode, etc.) results in an alert showing in the message banner
- [x] Performing `Create password` operation under normal conditions properly creates password and logs user out


Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -35,7 +35,7 @@ private struct MockPushNotificationDialog: PushNotificationDialogType {
   }
 }
 
-class MockViewController: UIViewController {
+private class MockViewController: UIViewController {
```
dusi: Does not need to be exposed

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
+import Library
```
dusi: Mostly just renaming the VC and slight internal changes

Kickstarter-iOS/Views/Controllers/CreatePasswordTableControllerTests.swift
```diff
@@ -3,7 +3,7 @@ import Prelude
 @testable import Library
 import XCTest
 
-final class CreatePasswordViewControllerTests: TestCase {
+final class CreatePasswordTableViewControllerTests: TestCase {
```
dusi: Only renaming

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+}
+
+@objc protocol CreatePasswordTableViewControllerType {
+  @objc optional func newPasswordTextFieldChanged(_ sender: UITextField)
+  @objc optional func newPasswordTextFieldDidReturn(_ sender: UITextField)
+  @objc optional func newPasswordConfirmationTextFieldChanged(_ sender: UITextField)
+  @objc optional func newPasswordConfirmationTextFieldDidReturn(_ sender: UITextField)
+}
+
+final class CreatePasswordTableViewController: UITableViewController {
```
dusi: `CreatePasswordTableViewController` now solely exists to display the data (create password UI). Since it's embedded in `CreatePasswordViewController` we had to extract configuring `title` and `saveButton` on that level.

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+  // MARK: - Properties
+
+  private let viewModel: CreatePasswordViewModelType = CreatePasswordViewModel()
+  weak var delegate: CreatePasswordViewControllerDelegate?
```
dusi: And we're using a `delegate` to interact with the container

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+  // MARK: - Binding
+
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.viewModel.outputs.activityIndicatorShouldShow
+      .observeForUI()
+      .observeValues { [weak self] shouldShow in
+        if shouldShow {
+          self?.delegate?.saveButtonStartAnimating()
```
dusi: Like here..

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+      .observeForControllerAction()
+      .observeValues { [weak self] errorMessage in
+        self?.delegate?.showError(with: errorMessage)
+    }
+
+    self.viewModel.outputs.createPasswordSuccess
+      .observeForControllerAction()
+      .observeValues { [weak self] in
+        guard let self = self else { return }
+        logoutAndDismiss(viewController: self)
```
dusi: (this probably can stay here)

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+    self.viewModel.outputs.newPasswordConfirmationTextFieldDidResignFirstResponder
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.newPasswordConfirmationTextField?.resignFirstResponder()
+    }
+
+    self.viewModel.outputs.saveButtonIsEnabled
+      .observeForUI()
+      .observeValues { [weak self] isEnabled in
+        self?.delegate?.saveButtonSetIsEnabled(isEnabled)
```
dusi: Also using delegate here..

Kickstarter-iOS/Views/Controllers/CreatePasswordTableViewController.swift
ifbarrera: I wonder if it might make more sense for the delegate type to be `CreatePasswordTableViewControllerDelegate` since it's the class of type `CreatePasswordTableViewController` that is *delegating* work to another object 🤔

dusi: right

PR #644, opened 2019-03-26
Uitests page model object

Scollaco:
# 📲 What
- This is a proof of concept showing how to write UITests using the Page Object Model pattern.
Here are some links to help understand a little better about the pattern:

*Note* - The tests currently live in the `BaseTest` file. This file should contain simply the functions that will be shared with all UITestCase classes (in this case `setup` and `tearDown`), and the tests should live on their own UITestCase files.

https://alexilyenko.github.io/xcuitest-page-object/
https://medium.com/@danielcarlosce/some-good-practices-for-xcuitest-807bfe6b720d
https://medium.com/@SunXiaoShan/uitest-page-object-module-703564351f88 (You may need to use the Chrome's translation tool for this one :) )

# 👀 See
## Search test (Searching and tapping  the Cancel button):
uitest_search


Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -27,6 +27,7 @@ public final class RootTabBarViewController: UITabBarController {
   override public func viewDidLoad() {
     super.viewDidLoad()
     self.delegate = self
+    self.tabBar.accessibilityIdentifier = AccessibilityIdentifier.RootTabBar.tabBar.rawValue
```
ifbarrera: I thought what they did in [this blog post](https://medium.com/inside-bux/simplifying-ios-ui-testing-with-page-object-model-765a23e8ebd4) was neat and it could help to do something similar here. First, they define a protocol `UIElement` that acts as a wrapper around `accessibilityIdentifier`:

```
protocol UIElement {
    var accessibilityIdentifier: String { get }
}

extension UIElement where Self: RawRepresentable {
    var accessibilityIdentifier: RawValue {
        return self.rawValue
    }
}
```
Then a second protocol `UITestablePage` can be used to make any `UIViewController` testable. A view controller can conform to `UITestablePage` and then define a `makeTestable` function that adds the accessibility identifiers to the views/elements that we want to expose to a UITest:

```
protocol UITestablePage {
    associatedtype UIElementType: UIElement

    func makeViewTestable(_ view: UIView, using element: UIElementType)
}

extension UITestablePage {
    
    func makeViewTestable(_ view: UIView, using element: UIElementType) {
        view.accessibilityIdentifier = element.accessibilityIdentifier
    }
}
```
```
extension RootTabBarViewController: UITestablePage {
    typealias UIElementType = AccessibilityIdentifier.RootTabBar
    func makeTestable() {
        self.makeViewTestable(self.tabBar, using: .tabBar)
    }
}
```

Scollaco: Yeah, I like the fact we can set directly the views to their respective UIElement. It can probably (not sure) make the tests run faster, since it won't need to iterate through each array of subviews until find the right one based on its accessibilityIdentifier. 

The only thing I didn't like so much is how we have to set the value manually for each view. I wonder if we could make write something to make this step automatic.

dusi:
This work was as a spike to POM and it served its purpose...closing for now.

PR #645, opened 2019-03-26
👩‍🔬 [Investment Day] Add Smart Invert Support

dusi:
# 📲 What

Adds Smart Invert support

# 🤔 Why

Smart invert allows inverting the "chrome" of our app while keeping images untouched. This is only supported on iOS 11 so it's behind `availability` check.

# 🛠 How

"Simply" applied style to all `UIImageView`s that qualify as `displaying user generated content`.

# 👀 See

| Before | Smart Invert ⚫️ | Normal ⚪️ |
| --- | --- | --- |
| Screen Shot 2019-03-26 at 4 05 50 PM | Screen Shot 2019-03-26 at 3 35 54 PM | Screen Shot 2019-03-26 at 3 36 03 PM | Screen Shot 2019-03-26 at 4 06 06 PM | Screen Shot 2019-03-26 at 3 36 53 PM | Screen Shot 2019-03-26 at 3 37 00 PM |
| Screen Shot 2019-03-26 at 4 06 29 PM | Screen Shot 2019-03-26 at 3 39 06 PM | Screen Shot 2019-03-26 at 3 39 18 PM |
| Screen Shot 2019-03-26 at 4 06 46 PM | Screen Shot 2019-03-26 at 3 40 17 PM | Screen Shot 2019-03-26 at 3 40 24 PM |
| Screen Shot 2019-03-26 at 4 07 07 PM | Screen Shot 2019-03-26 at 3 41 21 PM | Screen Shot 2019-03-26 at 3 41 28 PM |
| Screen Shot 2019-03-26 at 4 07 16 PM | Screen Shot 2019-03-26 at 3 41 58 PM |Screen Shot 2019-03-26 at 3 42 04 PM |
| Screen Shot 2019-03-26 at 4 07 29 PM | Screen Shot 2019-03-26 at 3 42 13 PM | Screen Shot 2019-03-26 at 3 42 19 PM |
| Screen Shot 2019-03-26 at 4 07 36 PM | Screen Shot 2019-03-26 at 3 42 37 PM | Screen Shot 2019-03-26 at 3 42 45 PM |
| Screen Shot 2019-03-26 at 4 07 43 PM | Screen Shot 2019-03-26 at 3 43 47 PM | Screen Shot 2019-03-26 at 3 43 55 PM |
| Screen Shot 2019-03-26 at 4 08 42 PM | Screen Shot 2019-03-26 at 3 44 34 PM | Screen Shot 2019-03-26 at 3 44 41 PM |

[...]

PR #648, opened 2019-03-28
👩‍🔬 [Investment Day] Remove unused code

dusi:
# 📲 What

Removes unused files.

# 🤔 Why

During my work on #645 I've noticed that there are couple files that are unused in our project so I've decided to remove these.

# ✅ Acceptance criteria

- [x] CI passes
- [x] Create password screen works correctly in portrait & landscape mode


Kickstarter-iOS/Views/Controllers/CreatePasswordViewController.swift
```diff
@@ -35,8 +35,6 @@ final class CreatePasswordViewController: UIViewController, MessageBannerViewCon
       |> \.rightBarButtonItem .~ UIBarButtonItem(customView: self.saveButtonView)
 
     if let childView = self.createPasswordTableViewController.tableView {
-      childView.translatesAutoresizingMaskIntoConstraints = false
```
dusi: This is done internally inside of `childView.constrainEdges(to: self.view)`

Kickstarter-iOS/Views/Controllers/CreatePasswordViewControllerTests.swift
```diff
@@ -0,0 +1,11 @@
+@testable import Kickstarter_Framework
+import XCTest
+
+final class CreatePasswordViewControllerTests: XCTestCase {
```
dusi: We should probably test this just to be sure controller containment doesn't break in the future.

PR #654, opened 2019-04-02
💲 [Native Checkout] Plumbing

dusi:
# 📲 What

Basic plumbing for the native checkout feature work

Includes:
- VC + Tests
- VM + Tests
- DataSource + Tests

# 🤔 Why

This should be a starting ground for the follow up work that should make it easy for us to work on different pieces in parallel.

# 🛠 How

`Simply` creates a `PledgeViewController` that manages its child `PledgeTableViewController` which displays content in form of a table. The reason for wrapping up `PledgeTableViewController` in `PledgeViewController` is so that we can present alert banners or other fancy things. The reason for using `PledgeTableViewController` being a subclass of `UITableViewController` is in order for getting all the keyboard handling functionality & layout resizing for free.

The architecture closely follows `ProjectPamphletContentViewController`.

# 📍 Note

In order to being able to see the screen you can navigate to `ProjectPamphletContentViewController` and change the behaviour of `goToRewardsPledge` signal observer to

```
// L:98
self.viewModel.outputs.goToRewardPledge
  .observeForControllerAction()
  .observeValues { [weak self] project, reward in
    vc.configure(with: reward)
    vc.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(ProjectPamphletContentViewController.dismissRewardPledge(_:))) // Note: should not be handled by the controller itself but by its presenting controller (this controller)

    self?.present(nc, animated: true)
    }
  }
```

and also add the dismiss function to `ProjectPamphletContentViewController` like so
```
@objc private func dismissRewardPledge(_ button: UIBarButtonItem) {
  self.dismiss(animated: true, completion: nil)
}
```

# 👀 See

| iPhone SE | iPhone XS | iPad Pro 11 |
| --- | --- | --- |
| Screen Shot 2019-03-29 at 2 53 49 PM | Screen Shot 2019-03-29 at 2 53 50 PM | Screen Shot 2019-03-29 at 2 53 51 PM |

# ✅ Acceptance criteria

- [ ] Presenting `PledgeVC` shows table with 5 rows and prints the name of selected project in the console
- [ ] Checkout using web still works

### DIFF
--- a/Kickstarter-iOS/DataSources/PledgeDataSource.swift
+++ b/Kickstarter-iOS/DataSources/PledgeDataSource.swift
@@ -1,0 +1,46 @@
+import Foundation
+import KsApi
+import Library
+
+final class PledgeDataSource: ValueCellDataSource {
+  enum Section: Int {
+    case project
+    case inputs
+    case summary
+  }
+
+  func load(reward: Reward) {
+    self.appendRow(
+      value: "Description",
+      cellClass: PledgeRowCell.self,
+      toSection: Section.project.rawValue
+    )
+
+    self.appendRow(
+      value: "Your pledge amount",
+      cellClass: PledgeRowCell.self,
+      toSection: Section.inputs.rawValue
+    )
+
+    self.appendRow(
+      value: "Your shipping location",
+      cellClass: PledgeRowCell.self,
+      toSection: Section.inputs.rawValue
+    )
+
+    self.appendRow(
+      value: "Total",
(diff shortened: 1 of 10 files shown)


dusi:
This plumbing work is split in 2 (this one + #653) in order for an easy DIFF comparison. Otherwise the DIFF would be in thousands because Github couldn't resolve renaming files (and using previous files as new files) properly

PR #655, opened 2019-04-02
Present SFSafariViewController only for http and https schemes

dusi:
# 📲 What

Fixes an issue that we've been seeing recently that caused `SFSafariViewController` to crash when passed a URL that doesn't have `http` or `https` scheme (i.e. FB live stream). The fallback solution should now be testing whether the url can be opened and opening it in that case.

# 🤔 Why

This crash has recently escalated quite a bit during an update of Throw Throw burrito.

# 🛠 How

As much as this is simple `if`/`else` condition...in order to test this we have to conform to a new protocol `UIApplicationType` and use a mock object.

# ✅ Acceptance criteria

- [ ] Go to `Update` page and tap on any link
- [ ] Go to `Update` page and tap on link to a live stream


Kickstarter-iOS/Views/Controllers/CheckoutViewController.swift
```diff
@@ -132,12 +132,6 @@ internal final class CheckoutViewController: DeprecatedWebViewController {
     self.present(vc, animated: true, completion: nil)
   }
 
-  fileprivate func goToSafariBrowser(url: URL) {
```
dusi: There were 6 duplicated functions thorough our codebase...it should now be much DRYer

Library/UIViewController+URL.swift
```diff
@@ -0,0 +1,16 @@
+import SafariServices
+import UIKit
+
+public extension UIViewController {
+  static var supportedURLSchemes = ["http", "https"]
+
+  func goTo(url: URL, application: UIApplicationType = UIApplication.shared) {
```
justinswart: I think this dependency injection is totally fine - just wondering if we want to deviate from using `Environment` at this point in time?

dusi: Can you elaborate a bit? Not sure where you going with `Environment`? This injection is mostly for testing purposes

justinswart: Oh, yes I mean we currently do all dependency injection through `AppEnvironment`/`Environment`, so this kind of injection (passing into a function) is fine, it's just that the rest of our injection hasn't been done that way. I'm more opening this up for discussion than anything else, I am not particularly opinionated about it.

dusi: Oh right...I'll see what I can do and try to inject this through `AppEnvironment` instead

PR #658, opened 2019-04-09
[Investment day] Convert submodules to Carthage

justinswart:
# 📲 What

This changes our dependency management to make use of [Carthage](https://github.com/Carthage/Carthage) rather than git submodules.

# 🤔 Why

While git submodules are fine in and of themselves, I believe Carthage will make it easier for us to manage our dependencies and to also have a clearer understanding of how our dependencies are linked. During this process I noticed a number of areas where things were previously working simply because they were _implicitly_ lifted into the same namespace or because framework test targets were running using application hosts which isn't necessary. Now those frameworks are clearly separated and must be explicitly linked against each other to share code.

### What's the upside?

- It's simpler than git submodules.
- It _should_ speed up build times and cause Xcode to do less indexing (but I have to still prove that).
- The reason for that is because dependencies are not compiled at build time along with the app but instead are pre-built by Carthage as dynamic Swift libraries and linked against in the Xcode project.
- It should help us to manage all of our dependencies more neatly in one place.
- We can cache the pre-built dependencies on CircleCI which may give us better build times (to be proven).

### What's the downside?

- It's a sweeping change to how our dependencies are linked so we need to test it thoroughly.
- `make bootstrap` will now take a lot longer to run as Carthage will pre-build all of the dependencies instead of just checking out their source code.
  - There are ways we can manage this, for example on CircleCI we cache Carthage's build directory and we could do something similar locally. Although generally we should only run `make bootstrap` once, or if we encounter any problems.

# 🛠 How

1. Remove `.gitmodules` file (all submodules).
2. Add `Cartfile` along with all the same dependencies.
3. Pin those dependencies at the same versions and commit hashes as the submodules to ensure the same versions are being used as before.
4. Build, fix, rebuild, rinse and repeat until the project is running and all of its tests are passing.

# 👀 See

Below are a bunch of before and after screenshots of the Xcode configuration to make it easier to see what's changed.

### Project navigator
Note: Most of these will be removed when we deprecate the LiveStream framework.

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-04-09 at 7 35 09 PM| Screen Shot 2019-04-09 at 7 35 26 PM |

### App target dependencies
| Before 🐛 | After 🦋 |
| --- | --- |
|Screen Shot 2019-04-09 at 7 48 29 PM|Screen Shot 2019-04-09 at 7 48 32 PM|

### App build phases
| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-04-09 at 7 50 16 PM | Screen Shot 2019-04-09 at 7 50 12 PM |

Similarly, throughout the other targets these things were simplified. And in order for tests to run, we now need to copy the dynamic frameworks into the build directory's `Frameworks` folder. Which is similar to what we had to do to get Stripe and Playgrounds working.

### Library tests
This makes use of Carthage's [documented](https://github.com/Carthage/Carthage#adding-frameworks-to-unit-tests-or-a-framework) method for adding dynamic Swift frameworks to test targets.

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-04-09 at 7 59 47 PM | Screen Shot 2019-04-09 at 7 59 53 PM |

# ✅ Acceptance criteria

- [ ] The team prefers this approach.
- [ ] Thorough regression test of the app. Unfortunately this is wide-reaching and requires us to do some solid testing of the entire app.

# ⏰ TODO

- [x] Fix ReactiveExtensions tvOS target.
- [x] Move the two deleted tests to a UITest target.


.circleci/config.yml
```diff
@@ -62,10 +62,25 @@ jobs:
...
             - stripe-sdk-cache-{{ checksum "stripe_version.txt" }}
 
+      - run:
+          name: Store Xcode Version
+          command: xcodebuild -version > xcode_version.txt
+
+      - restore_cache:
+          name: Restore Carthage Cache
+          keys:
+            - carthage-cache-{{ checksum "Cartfile.resolved" }}-{{ checksum "xcode_version.txt" }}
```
justinswart: We combine the checksum of `Cartfile.resolved` with the current Xcode version so that we know it will rebuild dependencies if we use a new version of Xcode on CI.

Cartfile
```diff
@@ -0,0 +1,18 @@
+###Internal
+
+github "kickstarter/Kickstarter-Prelude" "7a56163ad02d817aeb022f6dcb3820a733d5ec4c"
```
justinswart: Some of these are pinned to commit hashes, that's just to ensure that they're unchanged from before for those dependencies that had submodules referencing non-release/tagged commits.

Kickstarter-iOS/DataSources/DashboardDataSource.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import Library
+import UIKit
```
justinswart: After moving things around these sorts of `UIKit` and `Foundation` imports became necessary as they are no longer 'implicitly' imported.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -10,7 +10,7 @@ import UserNotifications
 @testable import LiveStream
 @testable import Kickstarter_Framework
 @testable import KsApi
-@testable import ReactiveExtensions_TestHelpers
+import ReactiveExtensions_TestHelpers
```
justinswart: We can't use `@testable` to import `ReactiveExtensions_TestHelpers` because it's external to our project now. This appears to not be a problem.

The one thing I had to do on the framework side was to make `TestObserver` public to workaround this, I don't believe that's an issue.

Library/UIViewController+URLTests.swift
```diff
@@ -8,40 +8,6 @@ final class UIViewControllerURLTests: TestCase {
     XCTAssertEqual(UIViewController.supportedURLSchemes, ["http", "https"])
   }
 
-  func testGoToHttpScheme() {
```
justinswart: @dusi these tests now need to move to a UITest because we no longer run Library tests with a host application. I believe this is a more correct approach. Using `UIWindow` in a unit test is not allowed as it throws an exception:

```
NSInternalInconsistencyException: props must have a valid clientID
```

Makefile
```diff
@@ -12,8 +12,6 @@ DIST_BRANCH = $(RELEASE)-dist
 OPENTOK_VERSION ?= 2.10.2
 FABRIC_SDK_VERSION ?= 3.10.5
 FABRIC_SDK_URL ?= https://s3.amazonaws.com/kits-crashlytics-com/ios/com.twitter.crashlytics.ios/INSERT_SDK_VERSION/com.crashlytics.ios-manual.zip
-STRIPE_SDK_VERSION ?= 13.2.0
```
justinswart: Strip is now downloaded via Carthage.

bin/carthage.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
justinswart: This script pins Carthage to a particular version and handled caching of the `/Carthage` directory on CircleCI.

carthage-input-files.xcfilelist
```diff
@@ -0,0 +1,16 @@
+# Carthage input files
```
dusi: Is `.xcfilelist` a suggested extension for these scripts?

justinswart: Yip - https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes/build_system_release_notes_for_xcode_10

> Run Script build phases now support declaring input and output files in a .xcfilelist file.

dusi:
Just a quick note that we should probably give everyone on the team heads up when this PR is merged to delete `Frameworks` folder which will no longer be needed.

justinswart:
[Git submodules -> Carthage](https://trello.com/c/fseDWUmo/1359-git-submodules-carthage)

PR #659, opened 2019-04-11
[Native Checkout] "Back this project" button

ifbarrera:
# 📲 What

Adds the "Back this project" button to the project page when the `ios_native_checkout` flag is enabled.

This work *does not* include state-handling based on project states.

# 🤔 Why

So users can navigate to the reward selection screen.

# 🛠 How

Adds the "Back this project" button to the `ProjectPamphletViewController` if the `ios_native_checkout` flag is enabled. Otherwise, the "Back this project" button is not added.

# 👀 See

Tested on:
iPhone X (12.1)
iPhone 7 (11.4)
iPhone 5s (10.3.1)

| Before 🐛 | After 🦋 |
| --- | --- |
|3aJCj59iTC|OdCLej5Miq|

| iPhone 5 (10.3.1) | iPhone 7 (11.4) | iPhone X (12.1) |
| --- | --- | --- |
|ggVDhPRX3n|OdCLej5Miq|CQNtfeJPHX|

**Landscape Mode**

Simulator Screen Shot - iPhone 5s - 2019-04-11 at 17 10 25

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] **Feature flag on** Login as an admin on staging. Close the app. Relaunch app. Navigate to Project page. You should see the "Back this project" button
- [x] **Feature flag off** Log out from staging. Close the app. Relaunch the app. Navigate to Project page. You should *not* see the "Back this project" button.

# ⏰ TODO

- [x] Confirm default flagging behaviour


ifbarrera:
[Add "Back this Project" button to project page](https://trello.com/c/sIHggcy3/1321-add-back-this-project-button-to-project-page)

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -69,9 +104,52 @@ public final class ProjectPamphletViewController: UIViewController {
...
+      |> \.layoutMargins .~ .init(all: backThisProjectContainerViewMargins)
+
+    _ = self.backThisProjectButton
+      |> greenCTAButtonStyle
+      |> UIButton.lens.title(for: .normal) %~ { _ in
+        return Strings.project_back_button()
+    }
+
+    _ = self.backThisProjectButton.titleLabel
+      ?|> \.font .~ UIFont.boldSystemFont(ofSize: 15)
```
dusi: Is there a reason we're not using ksr font? Do we not want to support Dynamic Type?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -17,6 +17,31 @@ public final class ProjectPamphletViewController: UIViewController {
 
   @IBOutlet weak private var navBarTopConstraint: NSLayoutConstraint!
 
+  private let backThisProjectContainerViewMargins = Styles.grid(3)
+  private let backThisProjectContainerView: UIView = {
+    return UIView() |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private let backThisProjectButton: UIButton = {
+     return UIButton(type: .custom) |> \.translatesAutoresizingMaskIntoConstraints .~ false
```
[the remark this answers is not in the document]
dusi: Wonder if by using `.custom` the button does need explicitly setting selected/highlighted state color. Right now tapping and holding on the button doesn't give user a hint that anything is happening.

Library/Styles/BaseStyles.swift
```diff
@@ -4,6 +4,7 @@ import UIKit
 
 public enum Styles {
   public static let cornerRadius: CGFloat = 4.0
+  public static let buttonMinHeight: CGFloat = 44.0
```
dusi: This value seems like such a generic constant (based on HIG) not really related to just buttons..how about calling it something like `minTouchDimensions` or `minTappableDimensions`..etc?

Library/Styles/BaseStyles.swift
```diff
@@ -141,6 +142,14 @@ public let separatorStyle: ViewStyle = { (view: UIView) in
     |> \.accessibilityElementsHidden .~ true
 }
 
+public let greenCTAButtonStyle: ButtonStyle = { (button: UIButton) in
+  button
+    |> roundedStyle(cornerRadius: 12)
+    |> UIButton.lens.backgroundColor(for: .normal) .~ .ksr_green_500
+    |> UIButton.lens.backgroundColor(for: .highlighted) .~ .ksr_green_500
```
dusi: Ah, I think this is why the button doesn't seem to highlight when selected/highlighted..we're using the same color 🤔

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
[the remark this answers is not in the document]
ifbarrera: Yeah with `.custom` you have to set the highlighted color explicitly. I just set it to the same color because I'm not sure what color we want :\ the system behavior seemed too dramatic - I was planning on getting some design feedback since I want this to be the same styling we use for the buttons on the reward cells.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -32,6 +57,11 @@ public final class ProjectPamphletViewController: UIViewController {
   public override func viewDidLoad() {
     super.viewDidLoad()
 
+    if shouldShowNativeCheckout() {
+      self.setupViews()
```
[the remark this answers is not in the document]
ifbarrera: Yep I can combine them! I split them before when I was figuring out a bug with the constraints being set.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: Yeah I figured we don't want to support dynamic type on this button, similar to how we don't support dynamic type on navigation items. What do you think?

ifbarrera:
@dusi the padding in the screenshots is because we manually set the view's height: https://github.com/kickstarter/ios-oss/blob/back-this-project-button/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift#L63

This is so that the screenshots actually capture the entirety of rendered content to catch discrepancies. But because it's just a hard-coded, best-estimate value, most of them are off a little.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
dusi: Currently most of our buttons support Dynamic Type (even though not great) - mostly because we use UIButton subclasses. Would it be too hard to implement this in the same manner as we do now? (please see the screenshot)

Screen Shot 2019-04-12 at 3 48 12 PM

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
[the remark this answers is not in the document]
ifbarrera: @dusi I'm updating the `.highlighted` and `.selected` background color to just be the same color with an alpha of 0.8 for now - we can change it later during a design sweep if we want to

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: Yep can do! 👍

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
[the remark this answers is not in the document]
justinswart: For what it's worth, on https://github.com/kickstarter/ios-private/pull/438, I used a `.custom` close button and seem to get the highligh state for free 🤔

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -69,9 +103,58 @@ public final class ProjectPamphletViewController: UIViewController {
     }
   }
 
+  private func configureViews() {
+    // Configure subviews
+    self.backThisProjectContainerView.addSubview(self.backThisProjectButton)
+
+    self.view.addSubview(self.backThisProjectContainerView)
+    self.view.bringSubviewToFront(backThisProjectContainerView)
```
justinswart: Why necessary to bring to front? Also missing a `self` in this second one.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -69,9 +103,58 @@ public final class ProjectPamphletViewController: UIViewController {
...
+
+    self.backThisProjectButton.addTarget(self, action: #selector(backThisProjectTapped), for: .touchUpInside)
+
+    // Configure constraints
+    // swiftlint:disable line_length
```
justinswart: I would just indent these like this then we don't have to disable `line_length` rule. Also disabling it in this way disables it for the rest of the file.

```swift
    let backThisProjectContainerViewConstraints = [
      self.backThisProjectContainerView.leftAnchor.constraint(equalTo: self.view.leftAnchor),
      self.backThisProjectContainerView.rightAnchor.constraint(equalTo: self.view.rightAnchor),
      self.backThisProjectContainerView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor)
    ]
    
    let containerMargins = self.backThisProjectContainerView.layoutMarginsGuide

    let backThisProjectButtonConstraints = [
      self.backThisProjectButton.leftAnchor.constraint(equalTo: containerMargins.leftAnchor),
      self.backThisProjectButton.rightAnchor.constraint(equalTo: containerMargins.rightAnchor),
      self.backThisProjectButton.bottomAnchor.constraint(equalTo: containerMargins.bottomAnchor),
      self.backThisProjectButton.topAnchor.constraint(equalTo: containerMargins.topAnchor),
      self.backThisProjectButton.heightAnchor.constraint(equalToConstant: Styles.buttonMinHeight)
    ]
```

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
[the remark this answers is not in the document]
dusi: Yeah, I think we figured out in one of the comments bellow that it's due to the fact that both `.normal` & `.highlighted` colors are the same

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -101,12 +184,57 @@ public final class ProjectPamphletViewController: UIViewController {
     self.viewModel.inputs.willTransition(toNewCollection: newCollection)
   }
 
-  private func setInitial(constraints: [NSLayoutConstraint?], constant: CGFloat) {
+  // MARK: - Private View Setup Functions
+  private func configureSublayers() {
+    if let shapeLayer = self.backThisProjectContainerView.layer.sublayers?.first as? CAShapeLayer {
```
justinswart: Little bit confused by what's going on here? If we're reaching in for this layer shouldn't we just keen a reference to it? Also, if it's the first layer then we could probably just check if the sublayer count is greater than 0 🤔

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
dusi: I've noticed that the button now supports Dynamic Type (resizes the font)...but still cuts of the text in the largest setting. What do you think about making the button support multiple lines and allowing it to display the whole text?

To clarify, this is what's happening currently:
IMG_0123

Adding a simple implementation of a custom button (that could potentially fix our other buttons (in the above screenshot) like so can fix some of those problems:

```
// Could probably name it better but I was just quickly trying to get it working
class CustomButton: UIButton {
  override var intrinsicContentSize: CGSize {
    let size = self.titleLabel?.intrinsicContentSize ?? .zero
    let titleInsets = self.titleEdgeInsets

    return CGSize(
      width: titleInsets.left + size.width  + titleInsets.right,
      height: titleInsets.top + size.height + titleInsets.bottom
    )
  }

  override func layoutSubviews() {
    super.layoutSubviews()

    self.titleLabel?.preferredMaxLayoutWidth = self.titleLabel?.frame.width ?? 0
  }
}
```

Which fixes the button and allows multiline text if configured like so
```
self.backThisProjectButton.titleLabel?.numberOfLines = 0 // could be in the styling function
```

we would of course have to replace button's height constraint to allow for `greaterThan` as oppose to equal

```
self.backThisProjectButton.heightAnchor.constraint(greaterThanOrEqualToConstant: Styles.minTouchSize.height)
```

IMG_0125

We can also go one step further and make the spacing slightly better:
```
self.backThisProjectButton.titleEdgeInsets = UIEdgeInsets(topBottom: Styles.grid(2)) // could be in the styling function
```

IMG_0126

And finally we would have to fix the table view's content insets to accommodate for the button properly (this screenshot demonstrates scroll view scrolled all the way to the bottom - you can see how it's cutting off the last row)

IMG_0124

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: The `backthisProjectContainerView`'s layer starts off with one sub-layer: the button that's added as a subview. Once we add the shape layer, the count goes up to 2. Because of this, our options are either check if `backThisProjectContainerView.layer.sublayers.count == 1` or check that the first sublayer is the shape layer we added. Imo checking that the first sublayer is a `CAShapeLayer` is more specific, which is why I decided to go with that.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -17,6 +17,31 @@ public final class ProjectPamphletViewController: UIViewController {
 
   @IBOutlet weak private var navBarTopConstraint: NSLayoutConstraint!
 
+  private let backThisProjectContainerViewMargins = Styles.grid(3)
+  private let backThisProjectContainerView: UIView = {
+    return UIView(frame: .zero) |> \.translatesAutoresizingMaskIntoConstraints .~ false
```
dusi: I wonder if we should sooner or later start using a convenience functions / initializers to create our views that we'll be building programmatically. One of the pesky things is to keep remembering not to forget enabling auto layout which is by default off. This means each of our instance will have to have `translatesAutoresizingMaskIntoConstraints` set to `false`.

How would you folks feel about introducing something like this, that would allow us constructing views in more consistent way and we would not have to think about forgetting enabling auto layout?

Screen Shot 2019-04-15 at 12 41 42 PM

Cc: @justinswart @Scollaco @cdolm92 

Note: By all means this should probably not be part of this PR

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: Not necessary, will remove 👍

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -17,6 +17,31 @@ public final class ProjectPamphletViewController: UIViewController {
...
+  }()
+
+  private let backThisProjectButton: UIButton = {
+     return UIButton(type: .custom) |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private var backThisProjectContainerSublayer: CAShapeLayer {
```
dusi: I'm not sure if you were going to do some changes to this logic that we discussed on the call with Justin, but I'm starting to think this should be a `function` that takes a `frame` (`backThisProjectContainerView.bounds`), maybe even `radii` with default value of 16. This way it will be more reusable ¯\_(ツ)_/¯

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -69,9 +103,58 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    }
+  }
+
   public override func bindViewModel() {
     super.bindViewModel()
 
+    self.viewModel.outputs.goToRewards
+      .observeForControllerAction()
+      .observeValues { [weak self] params in
+        let (project, refTag) = params
```
dusi: Is there a reason we won't be needing `reward` in this case?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -101,12 +184,57 @@ public final class ProjectPamphletViewController: UIViewController {
...
+  }
 
+  private func setInitial(constraints: [NSLayoutConstraint?], constant: CGFloat) {
     constraints.forEach {
       $0?.constant = constant
     }
   }
+
+  private func goToRewards(project: Project, refTag: RefTag?) {
+
```
dusi: This could actually be already hooked up to the real VC. Again depending whether we'll be passing a `reward` or not this could look like:

```
private func goToRewards(project: Project, refTag: RefTag?) {
    let leftBarButtonItem = UIBarButtonItem(
      barButtonSystemItem: .cancel,
      target: self,
      action: #selector(dismissPledge)
    )

    vc.configure(with: project.rewards.first!)
    vc.navigationItem.leftBarButtonItem = leftBarButtonItem

  }
```

and we'd probably want to declare method responsible for dismissal (I'd suggest we do it here since the controller itself should not know how to present/dismiss itself to separate concerns - our app doesn't follow this pattern so it's just a suggestion).

```
@objc func dismissPledge() {
    self.dismiss(animated: true)
  }
```

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: I like it 😎

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: All the rewards are in `project.rewards`!

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: Done

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -69,9 +103,58 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    let backThisProjectButtonConstraints = [self.backThisProjectButton.leftAnchor.constraint(equalTo: containerMargins.leftAnchor),
+                                            self.backThisProjectButton.rightAnchor.constraint(equalTo: containerMargins.rightAnchor),
+                                            self.backThisProjectButton.bottomAnchor.constraint(equalTo: containerMargins.bottomAnchor),
+                                            self.backThisProjectButton.topAnchor.constraint(equalTo: containerMargins.topAnchor),
+                                            self.backThisProjectButton.heightAnchor.constraint(equalToConstant: Styles.minTouchSize.height)
+    ]
+
+    NSLayoutConstraint.activate([backThisProjectContainerViewConstraints,
+                                 backThisProjectButtonConstraints].flatMap { $0 })
```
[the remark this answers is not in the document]
ifbarrera: Just changed to a simple array concat 😄

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: @justinswart take a look at the updated code, now we're just storing a reference to the layer as you suggested and updating the layer's `path`

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: @dusi sounds like a great idea for a future refactor if we end up wanting to reuse this layer!

Library/Styles/BaseStyles.swift
ifbarrera: Now using `greenButtonStyle` from `ButtonStyles.swift` which has highlighted state handling 👍

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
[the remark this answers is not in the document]
ifbarrera: Updated to use `greenButtonStyle` from `ButtonStyles.swift` which has highlighted state handling!

Library/Styles/BaseStyles.swift
ifbarrera: Updated to `minTouchSize: CGSize`

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -143,3 +268,29 @@ extension ProjectPamphletViewController: ProjectNavBarViewControllerDelegate {
...
+    |> UIButton.lens.titleEdgeInsets .~ .init(topBottom: Styles.grid(1), leftRight: Styles.grid(2))
+    |> UIButton.lens.title(for: .normal) %~ { _ in
+      return Strings.project_back_button()
+  }
+}
+
+private var backThisProjectButtonTitleLabelStyle = { (titleLabel: UILabel?) -> UILabel? in
+  _ = titleLabel
+    ?|> \.font .~ UIFont.ksr_headline()
+    ?|> \.numberOfLines .~ Int(0)
```
dusi: Isn't `0` inferred to `Int` by default?

Kickstarter-iOS/Views/MultiLineButton.swift
```diff
@@ -0,0 +1,24 @@
...
+    return CGSize(
+      width: titleInsets.left + size.width  + titleInsets.right,
+      height: titleInsets.top + size.height + titleInsets.bottom
+    )
+  }
+
+  override func layoutSubviews() {
+    super.layoutSubviews()
+
+    let width = self.titleLabel?.frame.width ?? CGFloat(0)
```
dusi: Was this `CGFloat(0)` also intentional?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
ifbarrera: Oh I was trying to help the compiler here as much as possible 😭

dusi:
🚢 🚢 🚢 
🚢 🚢 🚢 
🚢 🚢 🚢

dannyalright:
I'm noticing on iPad that the button spans the full width of the screen instead of the default 'Readable Content' width [from Xcode](https://medium.com/@hacknicity/how-ipad-apps-adapt-to-the-new-11-and-12-9-ipads-pro-cabd1c0e5f20). [Original Abstract spec](https://share.goabstract.com/452a8d52-b8cc-4cb0-8ba9-f92ac55c4819?mode=design&sha=11002fd4791d653ec2fb97c0e7377769d811a53c).

| KickAlpha | Xcode Readable Content |
| --- | --- |
| kickalpha | xcode |

How should we go about fixing this?

Also, the transition on iPad seems a little disjointed compared to iPhone. The overlay seems to be in progress so I'm assuming it's not meant to be 1:1 yet.

dusi:
## Button

Yeah I agree..the button should not stretch more than the readable size (or I could even see us being more radical and making it way smaller)

For example this is from one of the WWDC talks this year which resembles our UI.

Screen Shot 2019-07-08 at 1 27 12 PM

## Modal presentation

If we're going to present the project screen full screen I'd consider tweaking the transition to be more like a fade-in / fade-out (as oppose to modal presentation)

or alternatively present this as a `.pageSheet` ¯\_(ツ)_/¯

dannyalright:
@dusi nice. How should we follow up on the button & modal presentation UI tweaks? Are they small enough to pair on in an hour one investment day?

dusi:
Yeah, let's do it this Friday..I'll put this convo on my local todo list

PR #660, opened 2019-04-11
💲[Native Checkout] Pledge Amount Cell

dusi:
# 📲 What

Implements pledge amount cell UI.

# 🤔 Why

New design = new cell 😄 

# 🛠 How

We've added custom cell `PledgeAmountCell` as well a custom input view for the amount `AmountInputView` which properly aligns the amount value and the currency.

Please note that user interaction has not been enabled (this is simply a visual layout work).

This is me fighting `UIKit` this week...thanks Apple
(I'm the hamster)
parrot

# 📍 Note

```

```

# 👀 See

| iPhone SE | iPhone X |
| --- | --- |
| Screen Shot 2019-04-11 at 4 25 21 PM | Screen Shot 2019-04-11 at 4 25 23 PM |

# ♿️ Accessibility 

For the `Dynamic Type` support we're switching the axis of the container that holds the stepper & the input field. If the font gets too big the axis are vertical...for anything normal-ish the axis are horizontal.

| iPhone SE | iPhone X |
| --- | --- |
| Screen Shot 2019-04-11 at 4 25 49 PM | Screen Shot 2019-04-11 at 4 25 40 PM |

`Dynamic Type`
- [ ] Stepper & input field are aligned horizontally for smaller/normal-ish font sizes
- [ ] Stepper & input field are aligned vertically for accessibility (larger) font sizes

`VoiceOver`
- [ ] The title is read as `Your pledge amount` (and is localized)
- [ ] The minus button is read as `Decrement, button` (alternatively dimmed if it's in default state)
- [ ] The plus button is read as `Increment, button`
- [ ] The currency is read as `US dollars` (or any other currency)
- [ ] The amount is read as `X, textfield` (plus instructions on editing the textfield)

# ✅ Acceptance criteria

- [ ] The cell matches designs
- [ ] Stepper's default state for the `-` button is disabled
- [ ] Stepper's `-` & `+` buttons have highlighted state (light shade of gray)
- [ ] Currency's ascender is aligned to amount's [ascender](https://en.wikipedia.org/wiki/Ascender_(typography))


Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
```diff
@@ -1455,4 +1456,4 @@
 "updates_count_updates.zero" = "%{updates_count} updates";
 "via_custom" = "via custom";
 "via_external" = "via external";
-"via_kickstarter" = "via Kickstarter";
\ No newline at end of file
+"via_kickstarter" = "via Kickstarter";
```
dusi: 🤦‍♂️

dusi:
Please note that there are some deprecated snapshot tests included, I'm not entirely sure but they must have been missed on the feature branch.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    let alignment: UIStackView.Alignment = (isAccessibilityCategory ? .leading : .center)
+    let axis: NSLayoutConstraint.Axis = (isAccessibilityCategory ? .vertical : .horizontal)
+    let distribution: UIStackView.Distribution = (isAccessibilityCategory ? .equalSpacing : .fill)
+
+    _ = self.inputStackView
+      |> \.alignment .~ alignment
+      |> \.axis .~ axis
+      |> \.distribution .~ distribution
+      |> \.spacing .~ (isAccessibilityCategory ? Styles.grid(1) : 0)
```
ifbarrera: I noticed that if you type a very long number in the `AmountInputView` that it goes right up to the right of the stepper. Should maybe keep the spacing at `Styles.grid(1)` even when the axis is horizontal?

image

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -0,0 +1,124 @@
...
+      currency: value.currency
+    )
+  }
+}
+
+// MARK: - Styles
+
+private let inputStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
+  stackView
+    |> \.spacing .~ Styles.grid(1)
```
ifbarrera: Yeah actually it seems that this isn't consistent with the spacing in `func configureView(for traitCollection: UITraitCollection)`

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -15,32 +15,41 @@ class PledgeTableViewController: UITableViewController {
...
 
   override func viewDidLoad() {
     super.viewDidLoad()
 
     _ = self.tableView
-      |> tableViewStyle
       |> \.dataSource .~ self.dataSource
 
     self.tableView.register(PledgeRowCell.self, forCellReuseIdentifier: "PledgeRowCell")
+    self.tableView.register(PledgeAmountCell.self, forCellReuseIdentifier: "PledgeAmountCell")
```
ifbarrera: Should we maybe convert these reuse identifiers to an enum?

Library/UIView+AutoLayout.swift
```diff
@@ -16,6 +16,11 @@ extension UIView {
     NSLayoutConstraint.activate(constraints)
   }
 
+  public func addSubviewConstrainedToEdges(_ view: UIView) {
```
ifbarrera: Combining adding the subview with constraining to edges doesn't feel like it really adds value, and it also hides how the subview is being added from the "owner" of the subview (the VC). I would argue we don't need this "helper" and we should keep adding subviews and constraining them separate.

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -57,10 +66,24 @@ class PledgeTableViewController: UITableViewController {
...
     |> \.contentInset .~ UIEdgeInsets(top: -35)
     |> \.sectionFooterHeight .~ 10
     |> \.sectionHeaderHeight .~ 0
-    |> \.separatorStyle .~ .none
+
+  if #available(iOS 11, *) { } else {
+    let estimatedHeight: CGFloat = 44
+
+    return style
+      |> \.contentInset .~ UIEdgeInsets(top: 30)
```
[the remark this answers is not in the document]
dusi: That's a really good question.

I've decided to use `UITableView` `.grouped` style in order to not have floating headers/footers (which we thought might come handy for the section separators).

By default the `.grouped` style adds certain padding above the first row so this removes that padding. In addition to that the value of insets is different on iOS 11+ and iOS 10 (so this line is a special case - which with our deprecation plans might disappear soon). There's also a know issue on iOS 10 (when rotating to landscape) that still shows the padding - I tried couple workarounds there but without luck...so hopefully we can remove iOS 10 support before this thing gets released. Otherwise we'd have to use `.plain` style and use cells for separators.

Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -9,16 +9,16 @@ final class PledgeDataSource: ValueCellDataSource {
     case summary
   }
 
-  func load(reward: Reward) {
+  func load(amount: Double, currency: String) {
```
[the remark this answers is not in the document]
dusi: Good point.

I've actually previously passed `Reward` object and accessed its properties...but since some of the values now come from `Project` we would also have to pass `Project` and `Reward`. As soon as this happened my thinking was: `Let's reduce the complexity for now" and when we need to pass more data either add more attributes or pass entire objects.`. I'd prefer for now going this way and if we find out that the amount of information passed in is making the function attributes too big we should think about your approach from above. For now this will stay more isolated and with less dependencies on model layer.

Hope that makes sense?

Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -0,0 +1,118 @@
...
+      |> textFieldStyle
+
+    _ = self.stackView
+      |> stackViewStyle
+
+    // Align label's ascender to text field's ascender
+    if let textFieldFont = self.textField.font, let labelFont = self.label.font {
+      let constant = (textFieldFont.capHeight - labelFont.capHeight) / 2
+
+      self.labelCenterYAnchor = NSLayoutConstraint(
```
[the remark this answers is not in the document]
dusi: LOL, I don't know where was I going with this 😆

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -0,0 +1,124 @@
...
+    _ = self
+      |> \.accessibilityElements .~ [self.label, self.stepper, self.amountInputView]
+      |> \.backgroundColor .~ UIColor.hex(0xf0f0f0) // Extract to Colors.json
+
+    self.contentView.addSubviewConstrainedToEdges(self.rootStackView)
+
+    self.rootStackView.addArrangedSubview(self.label)
+    self.rootStackView.addArrangedSubview(self.inputStackView)
+    self.inputStackView.addArrangedSubview(self.stepper)
+    self.inputStackView.addArrangedSubview(UIView(frame: .zero))
```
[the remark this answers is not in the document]
dusi: It's a spacer...it fills in the space between the stepper and the amount input.

Please see attached specs.

Screen Shot 2019-04-12 at 12 44 48 PM

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
dusi: I completely forgot about this..thanks for bringing it up...I'll see if I can use the separator and it's min width or any other option. It actually seems to behave slightly different on device vs SIM.

Library/UIView+AutoLayout.swift
dusi: I'm kinda conflicted about the fact that this doesn't add value Isabel

A) it merges two common operations into one (adding subview + settings constraints) - technically one wouldn't work without the other
B) the name explicitly says what it does when establishing constraints so I'm not sure how does it hide this from the owner

I'm fine with changing this if you feel strongly about it...I feel like I've seen this in some UIKit library but can't remember exactly which one ¯\_(ツ)_/¯ 

Let me know, I'll make the appropriate changes.

Kickstarter-iOS/DataSources/PledgeDataSource.swift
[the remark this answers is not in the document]
ifbarrera: Fair enough! Thanks for elaborating 🙏

Library/UIView+AutoLayout.swift
ifbarrera: These are fair points. My concerns are around adding helpers that prescribe "how" a subview is added to the hierarchy. Sometimes it's important to specify where the subview is added (`insertSubview: at:` etc) although admittedly the majority of the time we do end up using `addSubview`. I also have a general feeling that adding subviews and settings constraints are operations that should be grouped by together by type: all subviews added in the right order first, then all constraints added. Imo keeping them separate makes it easier to reason about how a view is being configured by splitting up *how views are added* from *how they're positioned*. 

I actually just noticed that we already have `addSubviewConstrainedToMargins` below, so we already have a precedent for introducing this function 🤷‍♀️ 

@justinswart thoughts on this?

Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -0,0 +1,119 @@
...
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+    super.traitCollectionDidChange(previousTraitCollection)
+
+    let traitCollection = self.traitCollection
+
+    if previousTraitCollection?.preferredContentSizeCategory
```
justinswart: Does this not work if you just put it in `bindStyles`? I believe that is called when the traitcollection changes.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -0,0 +1,128 @@
...
+      amount: String(format: "%i", value.amount),
+      placeholder: "\(0)",
+      currency: value.currency
+    )
+  }
+}
+
+// MARK: - Styles
+
+func inputStackViewStyle(_ isAccessibilityCategory: Bool) -> ((UIStackView) -> UIStackView) {
```
justinswart: Perhaps we should test these to see that the `UIStackView` that's passed in is configured correctly.

Library/UIView+AutoLayout.swift
justinswart: I guess as we continue building out view programmatically we risk starting to create our own AutoLayout DSL and then sticking to conventions across the project 🤔  While I don't have strong feelings about this in particular, perhaps a way to achieve both things is through function composition?

Here's a quick example I cobbled together, of course better function and argument naming would make this more readable.

```swift
// Functions

public func ksr_addSubviewToParent() -> ((UIView, UIView) -> (UIView, UIView)) {
  return { subview, parent in
    parent.addSubview(subview)
    return (subview, parent)
  }
}

public func ksr_constrainToEdgesInParent() -> ((UIView, UIView) -> (UIView, UIView)) {
  return { subview, parent in
    subview.translatesAutoresizingMaskIntoConstraints = false

    let constraints = [
      subview.leadingAnchor.constraint(equalTo: parent.leadingAnchor),
      subview.trailingAnchor.constraint(equalTo: parent.trailingAnchor),
      subview.topAnchor.constraint(equalTo: parent.topAnchor),
      subview.bottomAnchor.constraint(equalTo: parent.bottomAnchor)
    ]

    NSLayoutConstraint.activate(constraints)

    return (subview, parent)
  }
}

// Usage

override public func viewDidLoad() {
  super.viewDidLoad()

  let view = UIView()

  _ = (view, self.view)
    |> ksr_addSubviewToParent()
    |> ksr_constrainToEdgesInParent()
}

Decisions could be made as to whether we want to pass both the child and parent as arguments and receive both as return values to pipe forward or if we just want to receive the subview to do more things to it.

Anyway, just an idea!
```

ifbarrera: I like this approach a lot @justinswart, it's more consistent with the overall patterns in our codebase 👍

dusi: dumbdumber

Kickstarter-iOS/Views/AmountInputView.swift
dusi: Yeah, it works without it thanks to our swizzling....why did I do this Justin? 🤦‍♂️

PR #664, opened 2019-04-22
💲[Native Checkout] Rewards Collection View Plumbing

ifbarrera:
# 📲 What

Adds plumbing/simple implementation for the rewards screen.

This PR includes the following:

- adds `RewardsCollectionViewController` & `RewardsCollectionViewControllerTests`
- adds `RewardsCollectionViewDataSource` and `RewardsCollectionViewDataSourceTests`
- adds `RewardsCollectionViewModel` and `RewardsCollectionViewModelTests`
- renamed old `RewardCell` -> `DeprecatedRewardCell` & `RewardCellViewModel` -> `DeprecatedRewardCellViewModel`
- creates a new `RewardCell` file
- hooks up `RewardCollectionViewController` to the "Back this project" button

This PR **does not** include:
- custom scrolling behavior for "peeking" functionality

# 🤔 Why

To reduce future PR sizes when working on rewards.

# 👀 See

w0nmTRHfHE

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Sign in to an admin account on staging. Navigate to project page. Tap "Back this project". You should see a bare implementation of the rewards screen.
- [ ] Sign out. Navigate to a project page. You should *not* see the "Back this project" button, and you should be able to navigate to a reward as normal.


Kickstarter-iOS/Views/Cells/DeprecatedRewardCell.swift
```diff
@@ -0,0 +1,268 @@
+import KsApi
```
ifbarrera: Nothing in this file is actually new. This file was renamed from `RewardCell.swift` -> `DeprecatedRewardCell.swift`

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -1,268 +1,33 @@
+import Foundation
```
ifbarrera: All the contents in this file are new. Probably best/easiest to review the whole thing by clicking `View file`.

Library/DataSource/UICollectionView-Extensions.swift
```diff
@@ -5,6 +5,10 @@ public extension UICollectionView {
     register(cellClass, forCellWithReuseIdentifier: cellClass.description())
   }
 
+  public func register<T: ValueCell>(_ cellClass: T.Type) {
```
ifbarrera: This is a helper that is compatible with `ValueCellDataSource`, which has a default implementation for `T.defaultReusableId` that's slightly different from the above `cellClass.description()`.

ifbarrera: I think ideally all of our cells should conform to `ValueCell`, so there can be consistency in how we register and configure all our cells.

ifbarrera:
[Rewards Collection View Plumbing](https://trello.com/c/0p5CGxwm/1334-rewards-collection-view-plumbing)

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+import Prelude
+
+final class RewardsCollectionViewController: UICollectionViewController {
+  private let dataSource = RewardsCollectionViewDataSource()
+  private let viewModel = RewardsCollectionViewModel()
+
+  // Hidden scroll view used for paging
+  private let hiddenPagingScrollView: UIScrollView = {
+    UIScrollView()
+      |> \.backgroundColor .~ UIColor.red
```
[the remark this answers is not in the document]
ifbarrera: Whoops will remove!

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+  }()
+
+  private let layout: UICollectionViewFlowLayout = {
+    UICollectionViewFlowLayout()
+      |> \.minimumLineSpacing .~ Styles.grid(3)
+      |> \.sectionInset .~ .init(all: Styles.grid(6))
+      |> \.scrollDirection .~ .horizontal
+  }()
+
+  private let peekAmountInset = Styles.grid(3)
```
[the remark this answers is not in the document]
ifbarrera: This is left over from testing, will remove!

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+import Library
+import KsApi
+import Prelude
+
+final class RewardsCollectionViewController: UICollectionViewController {
+  private let dataSource = RewardsCollectionViewDataSource()
+  private let viewModel = RewardsCollectionViewModel()
+
+  // Hidden scroll view used for paging
+  private let hiddenPagingScrollView: UIScrollView = {
```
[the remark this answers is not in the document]
ifbarrera: They wont be re-instantiated because they're `let`s 🤔

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,124 @@
...
+    UIScrollView()
+      |> \.backgroundColor .~ UIColor.red
+      |> \.isPagingEnabled .~ true
+      |> \.isHidden .~ true
+  }()
+
+  private let layout: UICollectionViewFlowLayout = {
```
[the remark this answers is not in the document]
ifbarrera: The `flowLayout` var is a helper for grabbing the current layout of the `UICollectionView` and checking whether it's of type `UICollectionViewFlowLayout`. I see your point that we could just refer to it via the `layout` var that we use to configure the collection view layout, but I'd be wary of relying on the fact that the layout hasn't changed. I think it's safer to always access the layout via `self.collectionView.layout` to guarantee that you're actually grabbing the current layout directly from the collection view and not some layout that's no longer active.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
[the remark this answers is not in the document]
ifbarrera: Oh actually it's used in this PR, but removed once the hidden scroll view is implemented: https://github.com/kickstarter/ios-oss/pull/665/files#diff-16fe2eb0466c404db2f7410f45e74d64L25

Cool if I leave it for now given I already have a PR open that removes it?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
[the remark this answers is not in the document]
ifbarrera: And to my own point, `flowLayout` should be `self.collectionView.layout` not `self.collectionViewLayout` 😅  will update

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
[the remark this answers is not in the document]
justinswart: Ah, I missed that, thought they were computed `var`s 😬

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -1,268 +1,33 @@
...
-    super.awakeFromNib()
+  override init(frame: CGRect) {
+    super.init(frame: frame)
 
-    let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(tapped))
-    tapRecognizer.cancelsTouchesInView = false
-    tapRecognizer.delaysTouchesBegan = false
-    tapRecognizer.delaysTouchesEnded = false
-    self.addGestureRecognizer(tapRecognizer)
+    configureViews()
```
justinswart: How about a little `self` here 😬

Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
```diff
@@ -166,14 +167,28 @@ extension ProjectNavigatorViewController: ProjectPamphletViewControllerDelegate
...
+
+    self.present(navigationController, animated: true)
+  }
 }
 
+// MARK: - UIGestureRecognizerDelegate
+
 extension ProjectNavigatorViewController: UIGestureRecognizerDelegate {
   func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
     return true
```
justinswart: Is this for the hidden scrollview? Do we want to return true only for that scrollview's gesture?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+      |> \.minimumLineSpacing .~ Styles.grid(3)
+      |> \.sectionInset .~ .init(all: Styles.grid(6))
+      |> \.scrollDirection .~ .horizontal
+  }()
+
+  private var flowLayout: UICollectionViewFlowLayout? {
+    return self.collectionView.collectionViewLayout as? UICollectionViewFlowLayout
+  }
+
+  private let peekAmountInset = Styles.grid(3)
```
dusi: Do we need a property for this? We're only accessing it twice from the same scope of `viewDidLayoutSubviews`?

Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
ifbarrera: This is not new, I think it's for the pull-down effect in the project screen.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -1,268 +1,33 @@
...
+    _ = self.contentView
+      |> \.backgroundColor .~ .ksr_grey_200
 
-    self.viewModel.outputs.notifyDelegateRewardCellWantsExpansion
-      .observeForUI()
-      .observeValues { [weak self] in
-        self.doIfSome { $0.delegate?.rewardCellWantsExpansion($0) }
-    }
+    _ = containerView
+      |> \.backgroundColor .~ UIColor.white
```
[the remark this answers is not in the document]
ifbarrera: Seems a little unnecessary for one line of styling code but will do

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -1,268 +1,33 @@
...
-  @IBOutlet fileprivate weak var shippingLocationsSummaryLabel: UILabel!
-  @IBOutlet fileprivate var separatorViews: [UIView]!
-  @IBOutlet fileprivate weak var titleDescriptionStackView: UIStackView!
-  @IBOutlet fileprivate weak var viewYourPledgeButton: UIButton!
-  @IBOutlet fileprivate weak var youreABackerCheckmarkImageView: UIImageView!
-  @IBOutlet fileprivate weak var youreABackerContainerView: UIView!
-  @IBOutlet fileprivate weak var youreABackerLabel: UILabel!
-  @IBOutlet fileprivate weak var youreABackerStackView: UIStackView!
+final class RewardCell: UICollectionViewCell, ValueCell {
+  private let containerView = UIView(frame: .zero)
```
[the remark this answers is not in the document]
ifbarrera: Is there a compelling reason to use `lazy vars` over regular `let`s? My thinking is that the intention is that these views are *always* added as subviews (and thus, initialized) and so it doesn't make sense to make them lazy 🤔

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -0,0 +1,53 @@
...
+    self.reloadDataWithRewards = Signal.combineLatest(
+      self.configureWithProjectProperty.signal.skipNil(),
+      self.viewDidLoadProperty.signal
+    )
+      .map(first)
+      .map { $0.rewards }
+  }
+
+  private let configureWithProjectProperty = MutableProperty<Project?>(nil)
+  private let configureWithRefTagProperty = MutableProperty<RefTag?>(nil)
```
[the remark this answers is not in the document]
ifbarrera: It's something we'll definitely use so I thought might as well add it now.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
ifbarrera: This is removed in the[ "peek" functionality PR](https://github.com/kickstarter/ios-oss/pull/665) so to prevent merge conflicts I'd rather just leave it here for now.

Kickstarter-iOS/Views/Cells/RewardCell.swift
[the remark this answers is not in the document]
dusi: At this point (and for this screen it doesn't really matter)...I was talking more in terms of consistency with the other work we've done so far...It's fine either way! 👍

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+    let rewardsCollectionVC = RewardsCollectionViewController()
+    rewardsCollectionVC.viewModel.inputs.configure(with: project, refTag: refTag)
+
+    return rewardsCollectionVC
+  }
+
+  init() {
+    super.init(collectionViewLayout: self.layout)
+
+    let closeButton = UIBarButtonItem(image: UIImage(named: "icon--cross"),
```
dusi: I've seen this pattern in our code base (`VC` performing presentation/dismissal and other workflow actions) and was wondering if this is something we could start chatting around? Cc: @justinswart @cdolm92 @Scollaco 

This approach has been mostly used by Apple in their demo projects (mostly for simplicity of the demo) but it gives the controller powers (knowledge) it shouldn't really know about. In other words the controller should not know how to dismiss itself (it should be the role of it's presenting/parent view controller or ideally something like flow controller). It's a long way to an ideal state (flow controllers), or other patterns, but we could at least start talking whether we'd want to inject these actions from outside of the controller so that it's implementation agnostic (this is a common practice when defining React Native components for example and passing their actions from the parent component).

Example would be

```swift
// in parent view controller
let vc = RewardsCollectionViewController.instantiate()
vc.dismissHandler = { [weak self] in
  self?.dismiss(animated: true)
}

```

or

```swift
let vc = RewardsCollectionViewController.instantiate()
vc.leftBarButtonItem = UIBarButtonItem(image: UIImage(named: "icon--cross", style: .plain, target: self, action: #selector(ParentViewController.closeButtonTapped))

```

This way this controller can either be presented modally or pushed on the navigation stack and by providing dismissal action it could either be dismissed or popped. If we didn't inject the action based on the workflow / presentation context imaging we pushed this on the stack...the internally declared dismissal would not work because it only assumes modal presentation (this removes reusability and flexibility).

ifbarrera: I think this is something we should discuss as a team & approach more holistically in our codebase if it's something we want to move towards. I think it's a good topic to bring up for sure though, and will become more relevant as our app gets larger.

justinswart: Yeah, worth exploring further but outside the scope of this PR. While it's true that we may want to have dismissal stuff be declared at the presentation callsite, we should just be smart about how we do that. For example, I would want to avoid a situation where a child is calling to a parent via delegate methods to have something presented. UIKit happily presents VC's from any VC in the stack and I don't necessarily think that's a problem. What I'm saying is that the delegate pattern is bad for this but potentially exploring a better abstraction like a Presenter or Flow controller may be worthwhile.

PR #665, opened 2019-04-23
💲[Native Checkout] "Peek" functionality using hidden scroll view

ifbarrera:
# 📲 What

Implements "peeking" functionality using the technique described [here](http://khanlou.com/2013/04/paging-a-overflowing-collection-view/) by which a hidden scroll view is used to to override `UICollectionView`s default scrolling behaviour.

# 🤔 Why

For consistency with designs.

# 🛠 How

A scroll view is added to the view and set to hidden. The scroll view's content size is set to be the same as the collection view's content size, but the scroll view's bounds are set to the desired "page" size (the width of the `RewardCell`). Then, the scroll view's panGestureRecognizer is added to the collection view's gesture recognizers, and the collection view's own `panGestureRecognizer` is disabled. By overriding `scrollViewDidScroll`, we can intercept the scroll view's paging scroll view offsets, and manually set them on the collection view to achieve the desired paging behaviour.

# 👀 See

** iPhone X (12.1)**
pHVI7G1Lwb

*Landscape*

oyAjiTCie3

** iPhone 7 (11.4) **

NGcHzTrlvg

** iPhone 5s (10.3.1) **

z3dW9Z1Cg4

# ✅ Acceptance criteria

- [x] Log in as admin on staging. Background the app. Navigate to the project page. Tap "Back this project". Scrolling through rewards, you should see the current reward centered on the screen.


ifbarrera:
[Implement "peek" effect in rewards collection view](https://trello.com/c/ZRciGmOn/1343-implement-peek-effect-in-rewards-collection-view)

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -104,21 +98,105 @@ final class RewardsCollectionViewController: UICollectionViewController {
...
+  private func updateHiddenScrollViewBoundsIfNeeded() {
+    guard let layout = flowLayout else { return }
+
+    let (contentSize, pageSize, contentInsetLeftRight) = self.hiddenScrollViewData(from: layout,
+                                                                                   using: self.collectionView)
+    let needsUpdate = self.collectionView.contentInset.left != contentInsetLeftRight
+      || self.hiddenPagingScrollView.contentSize != contentSize
+
+    // Check if orientation or frame has changed
+    if needsUpdate {
```
justinswart: Can we use a `guard needsUpdate`?

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -6,6 +6,8 @@ import Prelude
 final class RewardCell: UICollectionViewCell, ValueCell {
   private let containerView = UIView(frame: .zero)
 
+  static let fixedCardWidth: CGFloat = 249
```
justinswart: Not based on device width?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -73,18 +73,9 @@ final class RewardsCollectionViewController: UICollectionViewController {
 
     guard let layout = self.flowLayout else { return }
```
dusi: In an ideal scenario...this `guard` is repeated twice in the flow...first time here...second time in `updateHiddenScrollViewBoundsIfNeeded`. In order to make each function independent of the state of the view controller could we do this

```swift
// Callsite
guard let layout = self.flowLayout else { return }

// or maybe even also injecting layout to a new function and extracting this to `updateItemSize(for: layout)`
layout.itemSize = self.calculateItemSize(from: layout, using: self.collectionView)

self.updateHiddenScrollViewBoundsIfNeeded(for: layout)

// Updated function input
private func updateHiddenScrollViewBoundsIfNeeded(for layout: UICollectionViewFlowLayout) {
  // Then we don't need to guard here
}

// Possibly new function
private func updateItemSize(for layout: UICollectionViewFlowLayout) {
  // Then we don't need to guard here
}
```

or even something else?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -104,21 +98,105 @@ final class RewardsCollectionViewController: UICollectionViewController {
...
+  // MARK: - Private Helpers
+
+  private func configureHiddenScrollView() {
+    _ = self.hiddenPagingScrollView
+      |> \.delegate .~ self
+
+    _ = (self.hiddenPagingScrollView, self.view)
+      |> ksr_addSubviewToParent()
+
+    self.view.sendSubviewToBack(self.hiddenPagingScrollView)
```
dusi: Would it make sense to create another function helper for inserting subviews? This way we wouldn't need to do 2 operations (add & sending to back)? But can do instead...

```swift
func ksr_insertSubviewToParentAt(index: Int) -> (UIView, UIView) -> (UIView, UIView) {
  return { (subview, parent) in
    parent.insertSubview(subview, at: index)

    return (subview, parent)
  }
}
```

and call it like so

```swift
_ = (self.hiddenPagingScrollView, self.view)
  |> ksr_insertSubviewToParentAt(index: 0)
```

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -104,21 +98,105 @@ final class RewardsCollectionViewController: UICollectionViewController {
...
+        |> \.contentInset .~ .init(top: top,
+                                   left: contentInsetLeftRight,
+                                   bottom: bottom,
+                                   right: contentInsetLeftRight)
+
+      self.collectionView.contentOffset.x = -contentInsetLeftRight
+    }
+  }
+
+  typealias HiddenScrollViewData = (contentSize: CGSize, pageSize: CGSize, contentInsetLeftRight: CGFloat)
```
dusi: `private`?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -104,21 +98,105 @@ final class RewardsCollectionViewController: UICollectionViewController {
...
+                             height: collectionView.contentSize.height)
+
+    let contentInsetLeftRight = (collectionView.frame.width - itemSize.width) / 2
+
+    return (contentSize, pageSize, contentInsetLeftRight)
+  }
+
+  private func calculateItemSize(from layout: UICollectionViewFlowLayout,
+                                 using collectionView: UICollectionView) -> CGSize {
+    let cardWidth: CGFloat = RewardCell.fixedCardWidth
```
dusi: Wouldn't the `cardWidth` infer its type as `CGFloat`?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
ifbarrera: I like it, will update the `updateHiddenScrollViewBoundsIfNeeded` to accept `layout` as an argument 👍

Kickstarter-iOS/Views/Cells/RewardCell.swift
ifbarrera: @justinswart no, chatted with Izzy about it and it seems the designs do imply that the card width should stay constant.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -4,6 +4,11 @@ import KsApi
 import Prelude
 
 final class RewardsCollectionViewController: UICollectionViewController {
+  enum Layout {
```
dusi: In order to decouple this and make it private for the scope of `RewardsCollectionViewController` could we move this outside of the VC body and make it private just like we do in `PledgeDescriptionCell`. Otherwise since this is internal it can be accessed from anywhere in this module like so: `RewardsCollectionViewController.Layout.Card.width`.

So the final implementation would be more consistent with what we already have.

```swift
private enum Layout {
  enum Card {
    static let width: CGFloat = 249
  }
}

final class RewardsCollectionViewController: UICollectionViewController {
  // MARK: - Properties
```

Library/UIView+AutoLayout.swift
```diff
@@ -7,6 +7,13 @@ public func ksr_addSubviewToParent() -> ((UIView, UIView) -> (UIView, UIView)) {
   }
 }
 
+public func ksr_insertSubviewInParent(at index: Int) -> ((UIView, UIView) -> (UIView, UIView)) {
```
dusi: I just realized we don't have tests for these? Shouldn't be so hard to add?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -74,18 +80,13 @@ final class RewardsCollectionViewController: UICollectionViewController {
...
-    var itemWidth = collectionViewSize.width - sectionInsets.left - 2 * peekAmountInset
+  override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+    super.viewWillTransition(to: size, with: coordinator)
 
-    if [.landscapeLeft, .landscapeRight].contains(UIDevice.current.orientation) {
-      itemWidth = collectionViewSize.width / 3 - sectionInsets.left - 2 * peekAmountInset
-    }
-
-    layout.itemSize = CGSize(width: itemWidth, height: itemHeight)
+    self.flowLayout?.invalidateLayout()
```
dusi: I've noticed that no matter what the current horizontal offset rotating the device always resets it to 0 therefore scrolls back to the very first reward. There should be a way to either cache current page and upon rotation restore the scroll offset (maybe inside the coordinator completion handler)?

Library/UIView+AutoLayout.swift
ifbarrera: We actually don't have tests for any of these functions, but sure I can add one.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
ifbarrera: I think I'd prefer to implement this separately as an enhancement. I'll create a card 👍

PR #666, opened 2019-04-24
💲[Native Checkout] Shipping Location Cell 1/3 (UI)

dusi:
# 📲 What

Implements UI of the shipping location cell

# 🛠 How

In the same manner as we've added pledge amount cell we're continuing with shipping location cell.

Please note that this PR only considers visual appearance of the cell and that some values are hard coded + this work still misses user interaction.

This PR is the 1st out of 3 which will also address (passing correct data, formatting shipping amount correctly).

# 📍Note

```

```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-04-11 at 4 25 23 PM  | Screen Shot 2019-04-23 at 5 20 01 PM |

# ♿️ Accessibility 

**Dynamic Type**

Note how the country title truncates tail (there's an issue with multi-line labels/buttons in stack views) so currently we're simply using 1 line and might fix this in the future.

| Default | Large | Accessibility Large |
| --- | --- | --- |
| Screen Shot 2019-04-23 at 4 37 25 PM | Screen Shot 2019-04-23 at 4 37 29 PM | Screen Shot 2019-04-23 at 4 37 32 PM |
| Screen Shot 2019-04-23 at 4 38 02 PM | Screen Shot 2019-04-23 at 4 38 05 PM | Screen Shot 2019-04-23 at 4 38 07 PM |

**VoiceOver**

Shipping location cell exposes 3 accessible elements: title, button and label and reads them as follows

- [ ] Title is read by VO like: `Your shipping location, heading` and is localized
- [ ] Button is read by VO like: `{Country name}, button`
- [ ] Label is read by VO like: `+$7.5` (for now)

# ✅ Acceptance criteria

- [ ] Visual style matches design specs

(please note that the button doesn't use the small disclosure indicator, I'd suggest against this design approach due to it being mostly web pattern, not a mobile one)

### DIFF
--- a/Kickstarter-iOS/DataSources/PledgeDataSource.swift
+++ b/Kickstarter-iOS/DataSources/PledgeDataSource.swift
@@ -23,8 +23,8 @@ final class PledgeDataSource: ValueCellDataSource {
     )
 
     self.appendRow(
-      value: "Your shipping location",
-      cellClass: PledgeRowCell.self,
+      value: (location: "British Indian Ocean Territory", currency: "$", rate: 7.50),
+      cellClass: PledgeShippingLocationCell.self,
       toSection: Section.inputs.rawValue
     )
 
@@ -37,9 +37,11 @@ final class PledgeDataSource: ValueCellDataSource {
 
   override func configureCell(tableCell cell: UITableViewCell, withValue value: Any) {
     switch (cell, value) {
+    case let (cell as PledgeAmountCell, value as (Double, String)):
+      cell.configureWith(value: value)
     case let (cell as PledgeRowCell, value as String):
       cell.configureWith(value: value)
-    case let (cell as PledgeAmountCell, value as (Double, String)):
+    case let (cell as PledgeShippingLocationCell, value as (String, String, Double)):
       cell.configureWith(value: value)
     default:
       assertionFailure("Unrecognized (cell, viewModel) combo.")
--- a/Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
+++ b/Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
@@ -4,15 +4,15 @@ import XCTest
 @testable import KsApi
 
 final class PledgeDataSourceTests: XCTestCase {
-  let collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewLayout())
   let dataSource = PledgeDataSource()
+  let tableView = UITableView(frame: .zero, style: .plain)
 
   func testLoad() {
     dataSource.load(amount: 100, currency: "USD")
 
-    XCTAssertEqual(3, self.dataSource.numberOfSections(in: collectionView))
-    XCTAssertEqual(1, self.dataSource.collectionView(collectionView, numberOfItemsInSection: 0))
-    XCTAssertEqual(2, self.dataSource.collectionView(collectionView, numberOfItemsInSection: 1))
-    XCTAssertEqual(1, self.dataSource.collectionView(collectionView, numberOfItemsInSection: 2))
+    XCTAssertEqual(3, self.dataSource.numberOfSections(in: self.tableView))
+    XCTAssertEqual(1, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
+    XCTAssertEqual(2, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 1))
+    XCTAssertEqual(1, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 2))
   }
 }
--- a/Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
+++ b/Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
@@ -7,12 +7,18 @@ final class PledgeDataSourceTests: XCTestCase {
   let dataSource = PledgeDataSource()
   let tableView = UITableView(frame: .zero, style: .plain)
 
+  // swiftlint:disable line_length
   func testLoad() {
-    dataSource.load(amount: 100, currency: "USD")
+    self.dataSource.load(amount: 100, currency: "USD")
 
     XCTAssertEqual(3, self.dataSource.numberOfSections(in: self.tableView))
     XCTAssertEqual(1, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 0))
     XCTAssertEqual(2, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 1))
     XCTAssertEqual(1, self.dataSource.tableView(self.tableView, numberOfRowsInSection: 2))
+    XCTAssertEqual(PledgeRowCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: 0))
+    XCTAssertEqual(PledgeAmountCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: 1))
+    XCTAssertEqual(PledgeShippingLocationCell.defaultReusableId, self.dataSource.reusableId(item: 1, section: 1))
+    XCTAssertEqual(PledgeRowCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: 2))
   }
+  // swiftlint:enable line_length
 }
--- a/Kickstarter-iOS/Views/AmountInputView.swift
+++ b/Kickstarter-iOS/Views/AmountInputView.swift
@@ -15,12 +15,12 @@ class AmountInputView: UIView {
   override init(frame: CGRect) {
     super.init(frame: frame)
 
-    self.stackView.addArrangedSubview(self.label)
-    self.stackView.addArrangedSubview(self.textField)
-
     _ = (self.stackView, self)
       |> ksr_addSubviewToParent()
       |> ksr_constrainViewToEdgesInParent()
+
+    self.stackView.addArrangedSubview(self.label)
+    self.stackView.addArrangedSubview(self.textField)
   }
 
   required init?(coder aDecoder: NSCoder) {
@@ -33,7 +33,8 @@ class AmountInputView: UIView {
     super.bindStyles()
 
     _ = self
-      |> viewStyle
+      |> checkoutWhiteBackgroundStyle
+      |> checkoutRoundedCornersStyle
 
     _ = self.label
       |> labelStyle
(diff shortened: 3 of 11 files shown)


Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -23,8 +23,8 @@ final class PledgeDataSource: ValueCellDataSource {
     )
 
     self.appendRow(
-      value: "Your shipping location",
-      cellClass: PledgeRowCell.self,
+      value: (location: "British Indian Ocean Territory", currency: "$", rate: 7.50),
```
dusi: These values are currently hard coded in order for this PR to be smaller. I'll be passing them from the `VM` in the next PR.

Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
```diff
@@ -4,15 +4,15 @@ import XCTest
 @testable import KsApi
 
 final class PledgeDataSourceTests: XCTestCase {
-  let collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewLayout())
```
dusi: I must have copy-pasted this from a different data source test forgetting about the fact that we're actually dealing with `UITableView`.

Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -15,12 +15,12 @@ class AmountInputView: UIView {
   override init(frame: CGRect) {
     super.init(frame: frame)
 
-    self.stackView.addArrangedSubview(self.label)
-    self.stackView.addArrangedSubview(self.textField)
-
     _ = (self.stackView, self)
       |> ksr_addSubviewToParent()
       |> ksr_constrainViewToEdgesInParent()
+
```
dusi: For consistency with other views

Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -86,12 +87,6 @@ private let stackViewStyle: StackViewStyle = { (stackView: UIStackView) in
     |> \.isLayoutMarginsRelativeArrangement .~ true
 }
 
-private let viewStyle: ViewStyle = { (view: UIView) in
```
dusi: These styles are now shared so I extracted these to common styles file

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -0,0 +1,106 @@
...
+    _ = (self.rootStackView, self.contentView)
+      |> ksr_addSubviewToParent()
+      |> ksr_constrainViewToEdgesInParent()
+
+    self.rootStackView.addArrangedSubview(self.titleLabel)
+    self.rootStackView.addArrangedSubview(self.adaptableStackView)
+
+    self.adaptableStackView.addArrangedSubview(self.countryButton)
+    self.adaptableStackView.addArrangedSubview(self.spacer)
+    self.adaptableStackView.addArrangedSubview(self.amountLabel)
```
ifbarrera: Wondering if we can do something like this with stack views:

```
    _ = ([self.countryButton, self.spacer, self.amountLabel], self.adaptableStackView)
      |> ksr_addArrangedSubviews()
```

By defining:

```
public func ksr_addArrangedSubviews() -> (([UIView], UIStackView) -> UIStackView) {
  return { (subviews, stackView) in
    subviews.forEach { subview in
      stackView.addArrangedSubview(subview)
    }

    return stackView
  }
}
```

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -0,0 +1,106 @@
...
+    _ = self
+      |> checkoutBackgroundStyle
+
+    _ = self.adaptableStackView
+      |> checkoutAdaptableStackViewStyle(self.traitCollection.ksr_isAccessibilityCategory())
+
+    _ = self.amountLabel
+      |> amountLabelStyle
+
+    _ = self.countryButton
```
ifbarrera: Should we consider adding a hint to this button? `"Tap to change your shipping location"`?

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -0,0 +1,106 @@
...
+      |> \.text %~ { _ in Strings.Your_shipping_location() }
+
+    _ = self.rootStackView
+      |> checkoutStackViewStyle
+  }
+
+  // MARK: - Configuration
+
+  func configureWith(value: (location: String, currency: String, rate: Double)) {
+    self.countryButton.setTitle(value.location, for: .normal)
```
ifbarrera: Should these values go into a view model? Not sure if a VM is out of the scope of this work.

Library/Styles/CheckoutStyles.swift
```diff
@@ -0,0 +1,50 @@
...
+      top: Styles.grid(2), left: Styles.grid(4), bottom: Styles.grid(3), right: Styles.grid(4)
+    )
+    |> \.spacing .~ (Styles.grid(1) + Styles.gridHalf(1))
+}
+
+public let checkoutTitleLabelStyle: LabelStyle = { (label: UILabel) in
+  label
+    |> \.accessibilityTraits .~ UIAccessibilityTraits.header
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.font .~ UIFont.ksr_headline()
```
ifbarrera: Looking at the abstract files, I think the title labels are meant to have a font size of 15pts.

ifbarrera: (the default for `ksr_headline()` is 17pts)

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -73,39 +77,6 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
...
-    |> \.numberOfLines .~ 0
-    |> \.text %~ { _ in Strings.Your_pledge_amount() }
-}
-
-private let rootStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
-  stackView
-    |> \.axis .~ NSLayoutConstraint.Axis.vertical
-    |> \.isLayoutMarginsRelativeArrangement .~ true
-    |> \.layoutMargins .~ UIEdgeInsets(
-      top: Styles.grid(2), left: Styles.grid(4), bottom: Styles.grid(3), right: Styles.grid(4)
```
ifbarrera: If these margins are standard for all of the cells in the pledge VC, I wonder if it might make sense to put them on the cell's `contentView`s shared checkout style instead, so other cells can also have the same layout margins?

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
dusi: I like this! ❤️

dusi: I've slightly tweaked the naming so that it more resembles our other functions on `UIView`

i.e. just like when you're reading this from left to right

```
_ = (subview, parent)
  |> ksr_addSubviewToParent() // subview to parent (same order)
  |> ksr_constrainViewToEdgesInParent() // view to parent (same order)
```

we now do

```
_ = ([firstView, secondView, thirdView], stackView)
```

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
dusi: I'm undecided on this one. On one hand, that button looks like a button (so it should be treated like a button - `Title, button` (a11Y-title + a11y-trait)...on the other hand we currently use a different pattern for this with more descriptive action steps `Double tap to` (which in my opinion goes slightly against HIG).

Maybe we can talk about this offline?

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
dusi: Do you mean into cell's VM? That might be out of scope and again, I'd prefer to use these cells as a thin UI layer. I'm not really sure which direction do we want to go :/

Library/Styles/CheckoutStyles.swift
dusi: Good eye! Thanks for pointing this out.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
dusi: Do you mean simply moving `layoutMargins` to its own style (`checkoutCellContentViewStyle`) and removing it from here?

I'm not sure if I'm doing it wrong but it doesn't work for me :/

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -0,0 +1,105 @@
+import Library
+import Prelude
+import UIKit
+
+final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
+  // MARK: - Properties
+
+  private lazy var adaptableStackView: UIStackView = { UIStackView(frame: .zero) }()
```
justinswart: What does adaptable mean here? Might help to name it based on what it contains for quicker reference.

Library/UIStackView.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit.UIStackView
+
+public func ksr_addArrangedSubviewsToStackView() -> (([UIView], UIStackView) -> UIStackView) {
+  return { (subviews, stackView) in
+    subviews.forEach { subview in
```
justinswart: 😬 
```swift
subviews.forEach(stackView.addArrangedSubview)
```

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
dusi: It adapts its axis based on the accessibility font size...for regular sizes it uses horizontal axes..for a11y sizes it uses vertical

dusi: I've used to call it `inputStackView` but renamed it to this because it spans between multiple cells and always acts as the a11y adaptable container.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
ifbarrera: Hmm I wonder if you need to set `preservesSuperviewLayoutMargins` on the stack view? 🤔or alternatively you might need to constraint the stack view to the `contentView`s margins. This isn't super important but I just wanted to see if there was a way to share layout margins among all the pledge cells.

PR #667, opened 2019-04-26
💲[Native Checkout] Shipping Location Cell 2/3 (Currency formatter 🤯)

dusi:
# 📲 What

Adds the ability to create formatted attributed string based on designs.

# 🤔 Why

This should be a flexible solution for us to display amount values with specific currency symbol and fraction digits based on the designs.

# 🛠 How

Using an attributed string (don't worry, everything is tested :))))))

# 📍Note

```

```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-04-23 at 5 20 01 PM | Screen Shot 2019-04-25 at 6 01 13 PM |

But wait, there's more..

There's now a new playground that allows you to see various formats and allows you to test new ones as well.

Screen Shot 2019-04-25 at 5 59 44 PM

# ✅ Acceptance criteria

- [x] Currency is formatted as a superscript
- [x] Amount's (decimal numbers before the decimal point) are formatted normally
- [x] Amount's (decimal numbers after the decimal point) and the decimal point itself are formatted as a superscript

Please note that loading this data properly will be done in separate PR as well as prefixing the value with + symbol (based on designs)


dusi:
I don't think I've ever written these many tests...hopefully they sanity check this formatter a bit

Kickstarter-iOS.playground/Pages/Strings+Currency.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,129 @@
+import UIKit
+@testable import Library
+
+// Currency symbol
+
+attributedCurrencyString(
```
dusi: If you run the playground you'll get a nice inline preview of the formatted string. You can also play with various inputs.

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -86,8 +94,6 @@ final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
 private let amountLabelStyle: LabelStyle = { (label: UILabel) in
   label
     |> \.adjustsFontForContentSizeCategory .~ true
-    |> \.font .~ UIFont.ksr_title1()
```
dusi: This is no longer needed...these are instead passed to the formatter/attributter

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
...
+public func attributedCurrencyString(
+  currencySymbol: String,
+  amount: Double,
+  fractionDigits: UInt,
+  font: UIFont,
+  superscriptFont: UIFont,
+  foregroundColor: UIColor
+  ) -> NSAttributedString {
+  // Drop decimal places
+  let formattedString = String(format: "\(currencySymbol)%.\(fractionDigits)f", amount)
```
dusi: We strip down unnecessary decimal points..

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
...
+
+  let attributedString = NSMutableAttributedString(string: formattedString)
+  let franctionDigitsAndSeparator = Int(fractionDigits == 0 ? fractionDigits : fractionDigits + 1)
```
dusi: We don't want to display the separator if there are no fraction digits..

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
...
+  let range = NSRange(location: 0, length: attributedString.length)
+  let suffix = NSRange(
+    location: attributedString.length - franctionDigitsAndSeparator,
+    length: franctionDigitsAndSeparator
+  )
+
+  // Calculate vertical offset based on the height of a capital character of the two fonts
+  let maxCapHeight: CGFloat = max(font.capHeight, superscriptFont.capHeight)
+  let minCapHeight: CGFloat = min(font.capHeight, superscriptFont.capHeight)
+  let multiplier: CGFloat = font.capHeight > superscriptFont.capHeight ? 1 : 0
```
dusi: We don't want to offset fonts if the default font is smaller than the superscript font..

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
...
+  let baselineOffset = NSNumber(value: Float(multiplier * (maxCapHeight - minCapHeight)))
+
+  // Set font for the whole string
+  attributedString.addAttribute(NSAttributedString.Key.font, value: font, range: range)
```
dusi: And here we finally

1) Set text color and font for the whole attributed string
2) Override font for attributed currency substring
3) Override font for attributed fraction + separator substring

😎

Library/String+CurrencyTests.swift
```diff
@@ -0,0 +1,409 @@
+import XCTest
+@testable import Library
+
+// swiftlint:disable line_length
+final class StringCurrencyTests: XCTestCase {
+  func testCurrencySymbol_NoSymbol() {
```
dusi: The first batch of tests is testing only limited inputs (i.e. currency, amount, fractions, etc.).

The second batch of tests (prefixed `testCombined..`) is testing some common combinations.

Library/String+CurrencyTests.swift
```diff
@@ -0,0 +1,409 @@
...
+      font: font,
+      superscriptFont: superscriptFont,
+      foregroundColor: foregroundColor
+    )
+    let range = NSRange(location: 0, length: currencySymbol.count)
+    let attributedSubstring = attributedString.attributedSubstring(from: range)
+    let attributes = attributedSubstring.attributes(at: 0, longestEffectiveRange: nil, in: range)
+    let offset = font.capHeight - superscriptFont.capHeight
+
+    XCTAssertEqual("$99.97", attributedString.string)
```
dusi: An easy way to read the test is to look at the first assert in each test that shows the resulting formatted string...

Then based on designs read how are we styling currency prefix and fraction suffix

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -3369,7 +3373,6 @@
 				77FD8B45216D6245000A95AC /* LoadingBarButtonItemView.swift */,
 				D63BBD382180BE5D007E01F0 /* PaymentMethodsFooterView.swift */,
 				374F507822614A1000DE6746 /* PledgeFooterView.swift */,
-				5993DEBD1CE296F000925494 /* ProfileHeaderView.swift */,
```
dusi: These file removals must have been dangling in the project for a while...we've removed them some time back.

Kickstarter-iOS.playground/Pages/Strings+Currency.xcplaygroundpage/Contents.swift
dusi: On the second though the inline preview might not persist between computers (I'm not 100% sure about it)

Library/String+CurrencyTests.swift
dusi: You can simply copy paste each test into the `Strings+Currency.playground` and see for yourself

Library/String+Currency.swift
ifbarrera: Any reason not to use the formatter we already have? 
```
Format.formattedCurrency(_ amount: Any, country: Project.Country, omitCurrencyCode: Bool = true, env: Environment = AppEnvironment.current)
``` 
... returns a formatted currency amount which is localized and uses a cached `NumberFormatter`.

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
...
+    length: franctionDigitsAndSeparator
+  )
+
+  // Calculate vertical offset based on the height of a capital character of the two fonts
+  let maxCapHeight: CGFloat = max(font.capHeight, superscriptFont.capHeight)
+  let minCapHeight: CGFloat = min(font.capHeight, superscriptFont.capHeight)
+  let multiplier: CGFloat = font.capHeight > superscriptFont.capHeight ? 1 : 0
+  let baselineOffset = NSNumber(value: Float(multiplier * (maxCapHeight - minCapHeight)))
+
+  // Set font for the whole string
```
justinswart: Any benefit to grouping things like this by range?

```swift
// Set font and foreground color for the whole string
attributedString.setAttributes([
  NSAttributedString.Key.font: font,
  NSAttributedString.Key.foregroundColor: foregroundColor
], range: range)

// Replace prefix font and offset vertically from the baseline
attributedString.setAttributes([
  NSAttributedString.Key.font: superscriptFont,
  NSAttributedString.Key.baselineOffset: baselineOffset
], range: prefix)

// Replace suffix font and offset vertically from the baseline
attributedString.setAttributes([
  NSAttributedString.Key.font: superscriptFont,
  NSAttributedString.Key.baselineOffset: baselineOffset
], range: suffix)
```

Library/String+CurrencyTests.swift
```diff
@@ -0,0 +1,409 @@
...
+    XCTAssertEqual(currencySymbol, attributedSubstring.string)
+    XCTAssertEqual(superscriptFont, attributes[NSAttributedString.Key.font] as? UIFont)
+    XCTAssertEqual(foregroundColor, attributes[NSAttributedString.Key.foregroundColor] as? UIColor)
+    XCTAssertEqual(offset, (attributes[NSAttributedString.Key.baselineOffset] as? NSNumber) as? CGFloat)
```
justinswart: Interesting that it needs the cast to `NSNumber` first 🤔

dusi: What was I even thinking? 😄

Library/String+Currency.swift
dusi: Ah, this is great...I'll try using the formatter...had no idea we already have it!

Library/String+Currency.swift
dusi: Could do! Thought it was more readable to do 1 command per line but yeah, I'm not opposed to this.

dusi: The only downside to that is that `setAttributes` replaces all the previously set attributes so we have to additionally also set the foreground color like so:

```
// Set font and foreground color for the whole string
attributedString.setAttributes([
  NSAttributedString.Key.font: font,
  NSAttributedString.Key.foregroundColor: foregroundColor
], range: range)

// Replace prefix font and offset vertically from the baseline
attributedString.setAttributes([
  NSAttributedString.Key.font: superscriptFont,
  NSAttributedString.Key.baselineOffset: baselineOffset,
  NSAttributedString.Key.foregroundColor: foregroundColor
], range: prefix)

// Replace suffix font and offset vertically from the baseline
attributedString.setAttributes([
  NSAttributedString.Key.font: superscriptFont,
  NSAttributedString.Key.baselineOffset: baselineOffset,
  NSAttributedString.Key.foregroundColor: foregroundColor
], range: suffix)
```

justinswart: Ah, no that's not worth it, I'd leave as is 👍

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -53,6 +53,8 @@ final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
     _ = self.adaptableStackView
       |> checkoutAdaptableStackViewStyle(self.traitCollection.ksr_isAccessibilityCategory())
 
+    _ = self.amountLabel
+      |> checkoutBackgroundStyle
```
dusi: Setting background color on some of the labels to get rid of blended layers (in spirit of performance improvement and best practices)

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-05-01 at 12 09 16 PM | Screen Shot 2019-05-01 at 12 07 28 PM |

Library/String+Currency.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+import UIKit.UIColor
+import UIKit.UIFont
+
+public func attributedCurrencyString(
```
justinswart: Oh, something else I wanted to suggest, seeing as this is not actually an extention on `String` should the filename be `String+Currency.swift`? This may be fine to place in `SharedFunctions`? 🤷‍♂

dusi:
A different approach to a formatter...let me know if you prefer to pair on this @justinswart

Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -24,39 +28,37 @@ final class PledgeDataSource: ValueCellDataSource {
...
       toSection: Section.inputs.rawValue
     )
 
-    self.loadSummarySection(isLoggedIn: isLoggedIn)
-  }
-
-  private func loadSummarySection(isLoggedIn: Bool) {
     self.appendRow(
-      value: "Total",
+      value: Strings.Total(),
```
dusi: Noticed that this string wasn't localized

Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -24,39 +28,37 @@ final class PledgeDataSource: ValueCellDataSource {
...
+        toSection: Section.summary.rawValue
+      )
     }
   }
 
   override func configureCell(tableCell cell: UITableViewCell, withValue value: Any) {
     switch (cell, value) {
     case let (cell as PledgeAmountCell, value as (Double, String)):
       cell.configureWith(value: value)
+    case let (cell as PledgeContinueCell, value as ()):
```
dusi: Ordered alphabetically

Library/Format.swift
```diff
@@ -384,7 +408,7 @@ private struct NumberFormatterConfig {
   fileprivate let currencySymbol: String
 
   fileprivate func formatter() -> NumberFormatter {
-    let formatter = NumberFormatter()
+    let formatter = AttributedNumberFormatter()
```
dusi: My thinking behind this was that I didn't want to introduce a new formatter (cached formatter) since it would be identical to the `defaultCurrencyFormatter` which is already heavily covered by tests (these tests would be duplicate while there's no change to the functionality of the formatter other than extending it with attributed string support).

In order to set attributes call site can simply test for class type and cast this object properly.

I've done many iterations on this and at the end this seemed like a good tradeoff between keeping the functionality same (not duplicating tests) and adding attributed output support.

More than happy to discuss this

Library/NumberFormatter.swift
```diff
@@ -0,0 +1,67 @@
...
+  // MARK: - Properties
+
+  var defaultAttributes: String.Attributes = [:]
+  var currencySymbolAttributes: String.Attributes = [:]
+  var decimalSeparatorAttributes: String.Attributes = [:]
+  var fractionDigitsAttributes: String.Attributes = [:]
+
+  // MARK: - Attributed string
+
+  override func attributedString(for obj: Any, withDefaultAttributes attrs: [NSAttributedString.Key: Any]? = nil) -> NSAttributedString? {
```
dusi: Documentation indicates that this has to be overridden.

Doing so makes it pretty trivial for us to set various attributes to various portions of the output string (including currency symbol, fraction digits and decimal separator)

Library/NumberFormatterTests.swift
```diff
@@ -0,0 +1,150 @@
+@testable import Library
+import XCTest
+
+final class NumberFormatterTests: TestCase {
```
dusi: These tests should only test the attributed string (the format is tested in FormatTests)

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -32,18 +38,31 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
-         currencySymbol(forCountry: project.country).trimmed(),
-         reward.estimatedDeliveryOn
-          .map { Format.date(secondsInUTC: $0, template: "MMMMyyyy", timeZone: UTCTimeZone) } ?? "") }
+    let amountCurrencyDeliveryShipping = projectAndReward.signal
+      .map { (project, reward) -> (Double, String, String, (String, NSAttributedString?)) in
+        let amount = reward.minimum
+        let currency = currencySymbol(forCountry: project.country).trimmed()
+        let delivery = reward.estimatedDeliveryOn
+          .map { Format.date(secondsInUTC: $0, template: "MMMMyyyy", timeZone: UTCTimeZone) } ?? ""
+        let shipping = (
```
dusi: Currently static data that will be loaded properly in the next PR

Library/Format.swift
```diff
@@ -105,6 +105,30 @@ public enum Format {
...
+    let symbol = currencySymbol(forCountry: country, omitCurrencyCode: omitCurrencyCode, env: env)
+    let config = NumberFormatterConfig.defaultCurrencyConfig
+      |> NumberFormatterConfig.lens.locale .~ env.locale
+      |> NumberFormatterConfig.lens.currencySymbol .~ symbol
+      |> NumberFormatterConfig.lens.maximumFractionDigits .~ 2
+
+    guard let formatter = NumberFormatterConfig.cachedFormatter(forConfig: config)
+      as? AttributedNumberFormatter else { return nil }
+
+    _ = formatter
```
justinswart: Pity about having to do this discarded value 🤔  Only real alternative I could come up with was:

```swift
    return (
      formatter
        |> \.defaultAttributes .~ defaultAttributes
        |> \.currencySymbolAttributes .~ superscriptAttributes
        |> \.decimalSeparatorAttributes .~ superscriptAttributes
        |> \.fractionDigitsAttributes .~ superscriptAttributes
      )
      .attributedString(for: amount)
```
😅

dannyalright:
> There's now a new playground that allows you to see various formats and allows you to test new ones as well.
> 
> Screen Shot 2019-04-25 at 5 59 44 PM

Sick! Keen to learn more about this stuff.

PR #663, opened 2019-04-22
💲[Native Checkout] Description Cell

cdolm92:
# 📲 What

Implements UI of Description Cell.

# 🤔 Why

Part of Native Checkout.

# 🛠 How

Custom `PledgeDescriptionCell`

# 👀 See

Trello, screenshots, external resources?

| iPhone SE  | iPhone X |
| --- | --- |
|Screen Shot 2019-04-22 at 10 37 28 AM|Screen Shot 2019-04-22 at 10 38 37 AM|

# ♿️ Accessibility 
| iPhone SE  | iPhone X |
| --- | --- |
|Screen Shot 2019-04-22 at 10 44 24 AM|Screen Shot 2019-04-22 at 10 46 04 AM|

- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Cell matches the design

PR #668, opened 2019-04-26
💲[Native Checkout] Description Cell User Interaction

cdolm92:
# 📲 What

Adds user interaction to this cell which takes user to the Trust and Safety screen.

# 🛠 How

The `PledgeDescriptionCell` has a delegate method that the `PledgeTableViewController` conforms to to present the Trust and Safety screen. This is done using a `UITapGestureRecognizer` on the `learnMoreLabel`.
# 👀 See

description-cell-user-interaction

# ✅ Acceptance criteria

- [ ] Navigate to the Pledge Screen and tap on the green text, `Learn more about accountability`
- [ ] Trust and Safety screen appears

# ⏰ TODO

- [x] Refactor
- [x] Design feedback
- [x] Need to correct `Trust and Safety` screen presentation
- [x] Remove gesture change to button 
- [x] Merge https://github.com/kickstarter/ios-oss/pull/663


cdolm92:
I first implemented this using a UITapGesture w/ Label, @dusi  then suggested that I use a button instead so I did. And I think it's best to leave it as is.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -90,17 +101,10 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
...
-    _ = self.learnMoreLabel
-      |> learnMoreLabelStyle
-  }
-
-  // MARK: - Configuration
-
-  func configureWith(value: String) {
-    _ = self.dateLabel
-      |> \.text .~ value
+    _ = self.learnMoreButton
```
ifbarrera: Are these styles split up for compiler reasons? Or can we do:

```
_ = self.learnMoreButton
  |> checkoutBackgroundStyle
  |> learnMoreButtonStyle
```

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -46,6 +55,8 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
       |> ksr_constrainViewToEdgesInParent()
 
     self.configureStackView()
+    self.learnMoreButton.addTarget(self, action: #selector(learnMoreButtonTapped), for: .touchUpInside)
```
ifbarrera: Can we move all this out of the `init` and into a private `configureSubviews` function?

```
  override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
    super.init(style: style, reuseIdentifier: reuseIdentifier)

    self.configureSubviews()
    self.bindViewModel()
  }
```

Library/ViewModels/PledgeDescriptionCellViewModel.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+import KsApi
+import Prelude
+import ReactiveSwift
+import ReactiveExtensions
+import Result
+
+public protocol PledgeDescriptionCellViewModelInputs {
+  func configureWith(estimatedDeliveryDate: String)
+  func tapped()
```
ifbarrera: Can we use a more specific name for this input? Maybe `learnMoreButtonTapped`?

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -25,9 +32,11 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
   private lazy var estimatedDeliveryLabel: UILabel = { UILabel(frame: .zero) }()
   private lazy var dateLabel: UILabel = { UILabel(frame: .zero) }()
   private lazy var descriptionLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var learnMoreLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var spacerView: UIView = {
-    return UIView(frame: .zero) |> \.translatesAutoresizingMaskIntoConstraints .~ false }()
+  private lazy var spacerView: UIView = { UIView(frame: .zero) }()
```
ifbarrera: Do we need the closure here? Or can't we just do `private lazy var spacerView = UIView(frame: .zero)`?

Kickstarter-iOS/ViewModels/HelpWebViewModel.swift
```diff
@@ -45,7 +45,7 @@ internal final class HelpWebViewModel: HelpWebViewModelType, HelpWebViewModelInp
   }
 }
 
-private func urlForHelpType(_ helpType: HelpType, baseUrl: URL) -> URL? {
+public func urlForHelpType(_ helpType: HelpType, baseUrl: URL) -> URL? {
```
justinswart: We've made this public so that we can share the functionality, may be worth moving to `SharedFunctions.swift` and adding tests. But I'd suggest doing that in a separate PR.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
justinswart: Good point! We actually don't need the closure. So there's going to be a number of places that we can change this. I think the function stuff is a hangover from having set local vars in that scope before to set properties on it, but seeing as we pipe forward to set these things it can all be done in-line.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -46,6 +51,7 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
       |> ksr_constrainViewToEdgesInParent()
 
     self.configureStackView()
+    self.bindViewModel()
```
justinswart: Note-to-self that we're going to find a way to override `UITableViewCell` initializers to not have to do this.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -85,34 +91,26 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
     _ = self.dateLabel
       |> dateLabelStyle
 
-    _ = self.descriptionLabel
+    _ = self.learnMoreTextView
```
justinswart: Button and label now replaced with `UITextView`.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -163,20 +200,42 @@ private let dateLabelStyle: LabelStyle = { (label: UILabel) in
     |> \.numberOfLines .~ 0
 }
 
-private let descriptionLabelStyle: LabelStyle = { (label: UILabel) in
-  label
-    |> \.text  %~ { _ in Strings.Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life() }
-    |> \.textColor .~ UIColor.ksr_text_dark_grey_500
-    |> \.font .~ UIFont.ksr_caption1()
+private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) -> UITextView in
```
justinswart: Had to split up the styling here for the compiler to get with... the program 😅

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -163,20 +200,42 @@ private let dateLabelStyle: LabelStyle = { (label: UILabel) in
...
 }
 
-private let learnMoreLabelStyle: LabelStyle = { (label: UILabel) in
-  label
-    |> \.text  %~ { _ in "\(Strings.Learn_more_about_accountability())." }
-    |> \.textColor .~ UIColor.ksr_green_500
-    |> \.font .~ UIFont.ksr_caption1()
-    |> \.adjustsFontForContentSizeCategory .~ true
-    |> \.numberOfLines .~ 0
+private func attributedLearnMoreText() -> NSAttributedString {
```
justinswart: I will do some work in a separate PR to create a reusable function for constructing these seeing as we'll need it more and more going forward.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
cdolm92: These are split because the compiler gives us an error. This is because `learnMoreButtonStyle` is expecting a `UITextView` but `checkoutBackgroundStyle` is returning a `UIView`. 
Written like this it would work: 
`_ = self.learnMoreButton
  |> learnMoreButtonStyle
  |> checkoutBackgroundStyle
` 
 Let me know if I should make that change.

cdolm92:
Hi folks @ifbarrera @justinswart, let me know if this is g2g or if you have more feedback. thanks!

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
cdolm92: We also split it because the first part sets the background and the next the style. That way we don't have to worry about the order of passing the `UITextView`.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -163,20 +198,42 @@ private let dateLabelStyle: LabelStyle = { (label: UILabel) in
...
+  let linkRange = string.range(of: Strings.Learn_more_about_accountability())
+  let stringRange = string.range(of: string as String)
+
+  let attributedString = NSMutableAttributedString(string: string as String)
+
+  let url = urlForHelpType(
+    HelpType.trust, baseUrl: AppEnvironment.current.apiService.serverConfig.webBaseUrl
+  )
+
+  attributedString.addAttribute(.font, value: UIFont.ksr_caption1(size: 12), range: stringRange)
```
ifbarrera: I think the default font size for `.ksr_caption1` is 12pts so you don't need to explicitly pass that in here.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -124,10 +126,43 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
...
+  internal override func bindViewModel() {
+    super.bindViewModel()
+
+    self.dateLabel.rac.text = self.viewModel.outputs.estimatedDeliveryText
+
+    self.viewModel.outputs.presentTrustAndSafety
+      .observeForUI()
+      .observeValues { [weak self] in
+        guard let _self = self else { return }
+        self?.delegate?.pledgeDescriptionCellDidPresentTrustAndSafety(_self)
```
dusi: Man, this seems really confusing (using two `self` references)...wouldn't this suffice?
```
self.delegate.pledgeDescriptionCellDidPresentTrustAndSafety(self)
```

Kickstarter-iOS/Views/Controllers/DeprecatedRewardPledgeViewController.swift
```diff
@@ -230,7 +230,8 @@ internal final class DeprecatedRewardPledgeViewController: UIViewController {
 
     _ = self.disclaimerButton
       |> UIButton.lens.accessibilityLabel %~ { _ in
-        Strings.Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life()
+        Strings.Kickstarter_is_not_a_store()
```
dusi: Thanks for reverting this back..missed it in my prior PR review

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -64,6 +64,22 @@ class PledgeTableViewController: UITableViewController {
     let footerView = tableView.dequeueReusableHeaderFooterView(withClass: PledgeFooterView.self)
     return footerView
   }
+
+  internal override func tableView(_ tableView: UITableView,
+                                   willDisplay cell: UITableViewCell,
+                                   forRowAt indexPath: IndexPath) {
+    if let descriptionCell = cell as? PledgeDescriptionCell {
+      descriptionCell.delegate = self
```
dusi: Could have used prelude for this `_ = descriptionCell |> \.delegate .~ self`

PR #671, opened 2019-05-01
💲[Native Checkout] Reverts strings & snapshots

dusi:
# 📲 What

Reverts strings & snapshots

# 🤔 Why

Reverting strings back should revert `DeprecatedPledgeViewController` snapshots back to their original state and hopefully after being merged to `description cell` branch the only snapshot that should be updated would be the one for `Base` (English) localization due to a change of `It’s a way to bring creative projects to life.` to `It's a way to bring creative projects to life.`


Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -51,9 +51,8 @@ final class PledgeDescriptionCell: UITableViewCell, ValueCell {
     self.configureStackView()
 
     NSLayoutConstraint.activate([
-      self.containerImageView.widthAnchor.constraint(equalToConstant: Layout.Container.width),
```
dusi: Setting width constraint on the container was overriding image constraints....we actually don't need to set container's constraints since it will stretch based on the image view.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -166,7 +162,7 @@ private let descriptionLabelStyle: LabelStyle = { (label: UILabel) in
   label
     |> \.text  %~ { _ in Strings.Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life() }
     |> \.textColor .~ UIColor.ksr_text_dark_grey_500
-    |> \.font .~ UIFont.ksr_subhead(size: 12)
```
dusi: This font seemed closer to the designs and allows text to fit for the default size.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -3443,8 +3443,9 @@
 				A773531E1D5E8AEF0017E239 /* MostPopularSearchProjectCell.swift */,
 				A7B1EBB21D90496A00BEE8B3 /* NoRewardCell.swift */,
 				A74382041D3458C900040A95 /* PaddingCell.swift */,
-				37DEC22D2257CDD50051EF9B /* PledgeRowCell.swift */,
 				370ACAFF225D337900C8745F /* PledgeAmountCell.swift */,
+				D79A054A225E9B6B004BC6A8 /* PledgeDescriptionCell.swift */,
```
dusi: These files were not sorted alphabetically

PR #672, opened 2019-05-08
💲[Native Checkout] Attributed String Links Function

justinswart:
# 📲 What

Adds a function to assist in creating `NSAttributedString`s having parts of the string as tappable links when added to a `UITextView`.

# 🤔 Why

While working on #668 we noticed that on some of our designs we're going to need the functionality of having parts of the text in a label tappable. `UITextView` makes this relatively easy to handle but the `NSAttributedString` API doesn't. This function should make it easier to construct these strings.

# 🛠 How

Added the `ksr_attributedString(_:with:)` function which takes an `NSAttributedString` as well as an array of `AttributedLinkData` which is the `String` to link, the `URL` to link to and a dictionary of any additional properties to apply to the same string (bold/italic, etc).

**Note:** The one thing that this function doesn't do is handle linking the same words multiple times in a string. It uses `NSString`'s `range(of:)` function and this does a linear search from the beginning of the string. I think that this should almost always suffice but we may want to consider expanding on it in the future, I just think it's going to add a lot of complexity for what we need right now. Happy to discuss 😊 
 
# 👀 See

| Regular size | Large size |
| --- | --- |
| Screen Shot 2019-05-07 at 8 57 59 PM | Screen Shot 2019-05-07 at 9 11 16 PM |


justinswart:
[Attributed String Links Helper Function](https://trello.com/c/9ZC4Hqte/1356-attributed-string-links-helper-function)

Library/SharedFunctionsTests.swift
```diff
@@ -182,4 +182,62 @@ final class SharedFunctionsTests: TestCase {
...
+      .attribute(.foregroundColor, at: anotherLinkRange.location, longestEffectiveRange: nil, in: fullRange)
+
+    XCTAssertNotNil(aLinkAttribute, "a link is a link")
+    XCTAssertNotNil(aLinkColorAttribute, "a link has a foreground color")
+    XCTAssertNotNil(anotherLinkAttribute, "another link is a link")
+    XCTAssertNil(anotherLinkColorAttribute, "another link has no foreground color")
+  }
+
+  func testByPledgingYouAgree() {
+    let string = Strings.By_pledging_you_agree() as NSString
```
justinswart: Something worth noting is that this string, _By_pledging_you_agree_ is actually the following text:

> By pledging you agree to Kickstarter’s Terms of Use, Privacy Policy and Cookie Policy.

In our designs these are links:

Screen Shot 2019-05-07 at 9 24 00 PM

This _may_ be problematic when it comes to translations but let's chat more offline.

Library/SharedFunctions.swift
```diff
@@ -174,3 +174,19 @@ public func ksr_isOSVersionAvailable(_ version: Double) -> Bool {
...
+
+public typealias AttributedLinkData = (text: String, url: URL?, attributes: [NSAttributedString.Key: Any])
+
+public func ksr_attributedString(_ attributedString: NSAttributedString,
+                                 with links: [AttributedLinkData]) -> NSAttributedString {
+  let string = attributedString.string as NSString
+  let mutableAttributedString = NSMutableAttributedString(attributedString: attributedString)
+
+  links.map { text, url, attributes -> (attributes: [NSAttributedString.Key: Any], range: NSRange) in
+    (attributes.merging([.link: url as Any]) { $1 }, string.range(of: text))
```
justinswart: Merges the link attribute, it's respective URL and any other attributes that are passed in for that particular text.

ifbarrera: I would if using `localizedStandardRange(of str: String)` might be safer here?

justinswart:
Let's decide if this is still useful based on #673

Library/SharedFunctions.swift
```diff
@@ -174,3 +174,19 @@ public func ksr_isOSVersionAvailable(_ version: Double) -> Bool {
 
   return false
 }
+
+public typealias AttributedLinkData = (text: String, url: URL?, attributes: [NSAttributedString.Key: Any])
+
+public func ksr_attributedString(_ attributedString: NSAttributedString,
```
dusi: Looking at this, I'm seeing some duplication so I'm not sure what was the original decision to make this a function as oppose to an extension function on `NSAttributedString`?

I wonder if we can:

1) Make it an extension on `NSMutableAttributedString` (since we're already passing one in as a param)

```swift
extension NSMutableAttributedString {
  public var /* or func */ copyWith(links: [AttributedLinkData]) -> NSAttributedString {
    // I chose copy because if I remember correctly it usually represented conversion from mutable to an immutable object in ObjC 🤔 
  }
}
```

so then we can call it like so:

```swift
mutableAttributedString.with(links: [(Strings.Learn_more_about_accountability(), url, [:])])
```

2) Maybe make it more readable like this

```swift
let linkData = AttributedLinkData(
  text: Strings.Learn_more_about_accountability(),
  url: url,
  attributes: [:]
)

mutableAttributedString.with(links: [linkData])
```

3) Possibly make attributes optional with default value like so:

```swift
public typealias AttributedLinkData = (
  text: String, 
  url: URL?, 
  attributes: [NSAttributedString.Key: Any]? = [:])
```

so that we can shorten the above to

```swift
let linkData = AttributedLinkData(
  text: Strings.Learn_more_about_accountability(),
  url: url
)
```

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -222,18 +222,19 @@ private func attributedLearnMoreText() -> NSAttributedString {
...
   )
 
-  attributedString.addAttribute(.font, value: UIFont.ksr_caption1(), range: stringRange)
-  attributedString.addAttribute(.foregroundColor, value: UIColor.ksr_text_dark_grey_500, range: stringRange)
-  attributedString.addAttribute(.link, value: url as Any, range: linkRange)
-
-  return attributedString
+  return ksr_attributedString(
+    attributedString,
+    with: [(Strings.Learn_more_about_accountability(), url, [:])]
```
dusi: How about we use `AttributedLinkData` for better readability here? It's kinda hard to scan through that line

Library/SharedFunctionsTests.swift
```diff
@@ -182,4 +182,62 @@ final class SharedFunctionsTests: TestCase {
...
+      .attribute(.link, at: aLinkRange.location, longestEffectiveRange: nil, in: fullRange)
+    let aLinkColorAttribute = stringWithLinks
+      .attribute(.foregroundColor, at: aLinkRange.location, longestEffectiveRange: nil, in: fullRange)
+    let anotherLinkAttribute = stringWithLinks
+      .attribute(.link, at: anotherLinkRange.location, longestEffectiveRange: nil, in: fullRange)
+    let anotherLinkColorAttribute = stringWithLinks
+      .attribute(.foregroundColor, at: anotherLinkRange.location, longestEffectiveRange: nil, in: fullRange)
+
+    XCTAssertNotNil(aLinkAttribute, "a link is a link")
+    XCTAssertNotNil(aLinkColorAttribute, "a link has a foreground color")
```
dusi: If I'm looking at this right, it looks like this would succeed even if there was not a link but the attributed string had a color. Should we maybe explicitly test the color (values)?

Library/SharedFunctionsTests.swift
```diff
@@ -182,4 +182,62 @@ final class SharedFunctionsTests: TestCase {
     XCTAssertTrue(ksr_isOSVersionAvailable(12.123))
     XCTAssertTrue(ksr_isOSVersionAvailable(12.9))
   }
+
+  func testAttributedStringWithLinks() {
```
dusi: Should we also add a test that demonstrates inability to identify multiple occurrences of the same link / text?

Library/SharedFunctions.swift
dusi: On second thought this could even be extension on `String` in case we want to assure not mixing styles?

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -222,18 +222,19 @@ private func attributedLearnMoreText() -> NSAttributedString {
   \(Strings.Learn_more_about_accountability())
   """ as NSString
```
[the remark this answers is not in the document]
justinswart: Yeah, `NSRange` 🤷‍♂

Library/SharedFunctions.swift
justinswart: You can have defaults in a `typealias`?

justinswart:
Thanks for the feedback - I think a more important question before spending more time on this though is - do we need this function if we're going to construct these things from HTML? 🤔

justinswart:
Closing this in favour of #673.

PR #673, opened 2019-05-08
💲[Native Checkout] Use HTML string with anchor tags

justinswart:
# 📲 What

Updates our way of inserting tappable links in `UITextView` by generating the `NSAttributedString` directly from HTML.

# 🤔 Why

In #672 and #668 we explored a way to add links to `NSAttributedString` using `UITextView` by searching for specific phrases in that string and applying link attributes to those ranges.

While this is fine in and of itself, it may be problematic when we start to work with translations and try to search for phrases in translated strings. A safer route would be for the entire string to be translated at once and wrap the phrases we care about in anchor tags in good old fashioned HTML.

# 🛠 How

- Added a hard-coded `String` for now which I'll submit for translation if we're happy with this approach.
- Generated our `NSAttributedString` from that.

# 👀 See

Still seems legit:

Screen Shot 2019-05-08 at 3 14 57 PM

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

😕 This appears to be quite [challenging](https://stackoverflow.com/questions/18281525/uitextview-with-white-background-showing-as-color-blended-layer-in-instruments-a) with `UITextView`
- ❌ Optimized Blended Layers (screenshots)

# ✅ Acceptance criteria

- [ ] Comfortable with this approach

# ⏰ TODO

- [ ] Add string for translation

### DIFF
--- a/Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
@@ -216,24 +216,29 @@ private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) -> 
   return textView
 }
 
-private func attributedLearnMoreText() -> NSAttributedString {
-  let string = """
-  \(Strings.Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life())
-  \(Strings.Learn_more_about_accountability())
-  """ as NSString
-
-  let linkRange = string.range(of: Strings.Learn_more_about_accountability())
-  let stringRange = string.range(of: string as String)
-
-  let attributedString = NSMutableAttributedString(string: string as String)
-
-  let url = urlForHelpType(
-    HelpType.trust, baseUrl: AppEnvironment.current.apiService.serverConfig.webBaseUrl
+private func attributedLearnMoreText() -> NSAttributedString? {
+  // swiftlint:disable line_length
+  let string = localizedString(
+    key: "Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life_Learn_more_about_accountability",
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -216,24 +216,29 @@ private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) ->
...
+  guard let attributedString = try? NSMutableAttributedString(
+    data: Data(bytes: string.utf8),
+    options: [.documentType: NSAttributedString.DocumentType.html],
+    documentAttributes: nil
+  ) else { return nil }
+
+  let attributes: [NSAttributedString.Key: Any] = [
+    .font: UIFont.ksr_caption1(),
+    .foregroundColor: UIColor.ksr_text_dark_grey_500,
+    .underlineStyle: 0
```
justinswart: `NSUnderlineStyle.none` is buggy in Swift 4.2 🤷‍♂ 
https://bugs.swift.org/browse/SR-8184?page=com.atlassian.jira.plugin.system.issuetabpanels%3Achangehistory-tabpanel

justinswart: Also, I had to apply this here to the whole string, it was having no affect in the `UITextView`'s `linkTextAttributes` property 🤦‍♂ . I think this means that if we were wanting to be any more advanced with our links, like say let them have a different font or something, we may have to take a more manual approach.

justinswart:
[Attributed String Links Helper Function](https://trello.com/c/9ZC4Hqte/1356-attributed-string-links-helper-function)

justinswart:
> Just want to make sure the foreground color on the link is coming from the linkTextAttributes and not the window's tintColor.

Doing this _does_ make the link blue, so it seems to be using this.

```swift
    |> \.linkTextAttributes .~ [
      .foregroundColor: UIColor.blue
    ]
```

Screen Shot 2019-05-10 at 10 48 10 AM

PR #674, opened 2019-05-09
💲[Native Checkout] Sheet Overlay Container

ifbarrera:
# 📲 What

This PR adds `SheetOverlayContainerViewController`, a view controller that's intended to act as a custom container to **any** other view controller, rendering the child VC in a "card-like" way. It also comes with it's own custom transition animator, `SheetOverlayTransitionAnimator` which performs a custom modal presentation transition.

What the `SheetOverlay` does:
- contains any view controller within a "card" view that displays over the presenting view controller
- animates the presentation in a custom way

What the `SheetOverlay` **does not** do:
- manage navigation for you (dismissing, pushing, popping VCs)
- pass messages to/from the contained VC

Also in this PR:
- refactored the `Back this project` button's container view's layer to use `maskedCorners` to mask only the top right and top left corners, instead of creating a custom shape layer to achieve this. `maskedCorners` is only available on iOS 11 and above, which is fine given we're in the process of deprecating iOS 10

# 🤔 Why

This is required to present existing view controllers with the "card" design, as required in several places during the new checkout flow.

# 🛠 How

**New**
- `SheetOverlayContainerViewController`
- `SheetOverlayTransitionAnimator`

# 👀 See

**SheetOverlayContainerViewController configured with LoginToutViewController**
Simulator Screen Shot - iPhone 8 - 2019-05-09 at 11 54 10
Simulator Screen Shot - iPhone 8 - 2019-05-09 at 11 54 14

vKzZoX1mTv

**SheetOverlayContainerViewController configured with AddNewCardViewController**
Simulator Screen Shot - iPhone X - 2019-05-09 at 11 57 26

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

**How to test**

In `ProjectNavigatorViewController`, paste the following code into the function on line 171:

```
    let loginOutVC = LoginToutViewController.configuredWith(loginIntent: .backProject)
    let navigationController = UINavigationController(rootViewController: loginOutVC)
    let containerVC = SheetOverlayContainerViewController(childViewController: navigationController,
                                                          childViewOffset: 50)

    self.present(containerVC, animated: true)
```

Note that you can initialize `SheetOverlayContainerViewController` with *any* view controller.

- [ ] Log in on staging with an admin user. Navigate to any project. Tap "Back this project". The login/signup screen should animate from the bottom with a "sheet" animation, and partially cover the underlying screen.


Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -30,17 +30,6 @@ public final class ProjectPamphletViewController: UIViewController {
       |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  private let backThisProjectContainerSublayer: CAShapeLayer = {
```
ifbarrera: We don't need this `CAShapeLayer` anymore because we can set all these properties on the `backThisProjectContainerView`'s layer and use `maskedCorners` instead.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -102,41 +90,25 @@ public final class ProjectPamphletViewController: UIViewController {
     }
   }
 
-  private func configureViews() {
```
ifbarrera: Just moved this down under `Private Setup Functions`

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -102,41 +90,25 @@ public final class ProjectPamphletViewController: UIViewController {
...
+      |> \.shadowColor .~ UIColor.black.cgColor
+      |> \.shadowOpacity .~ 0.12
+      |> \.shadowOffset .~ CGSize(width: 0, height: -1.0)
+      |> \.shadowRadius .~ 1.0
+      |> \.cornerRadius .~ CheckoutStyleConstants.cardStyleRoundedCornerRadius
+
+    if #available(iOS 11.0, *) {
+      _ = self.backThisProjectContainerView.layer
+        |> \.maskedCorners .~ [.layerMaxXMinYCorner, .layerMinXMinYCorner]
+    }
```
ifbarrera: No `else` because we're removing iOS 10 support imminently.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift
```diff
@@ -62,7 +62,7 @@ internal final class ProjectPamphletViewControllerTests: TestCase {
         let (parent, _) = traitControllers(device: device, orientation: .portrait, child: vc)
         parent.view.frame.size.height = device == .pad ? 1_200 : 800
 
-        FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)")
+        FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)", tolerance: 0.01)
```
ifbarrera: These snapshots broke when I updated the "Back this project"'s layer code. Looking at the snapshot diff, there was nothing visibly different. I thought it might be better to bump the tolerance of these snapshots instead of regenerating all new ones for something that's not actually visually changing 🤔

ifbarrera:
[Create reusable animation for "slide up" card view functionality](https://trello.com/c/VNllysZN/1357-create-reusable-animation-for-slide-up-card-view-functionality)

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
justinswart: 🔜

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift
justinswart: Weird 🤔 sometimes in the past i recall the scroll indicators changing for some reason.

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,58 @@
+import UIKit
+import Library
+import Prelude
+
+/* SheetOverlayContainerViewController intended to be used as a container for another view controller
```
justinswart: Thanks for adding the documentation, I think the format we've used elsewhere like in `SharedFunctions.swift` is:

```swift
/**
 SheetOverlayContainerViewController intended to be used as a container for another view controller
 that you would like to render as a "sheet" or "card" that partially covers the content beneath it.
*/
```

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -102,41 +90,25 @@ public final class ProjectPamphletViewController: UIViewController {
...
   public override func bindStyles() {
     super.bindStyles()
 
     _ = self.backThisProjectContainerView
       |> \.layoutMargins .~ .init(all: backThisProjectContainerViewMargins)
 
+    _ = self.backThisProjectContainerView.layer
+      |> \.backgroundColor .~ UIColor.white.cgColor
+      |> \.shadowColor .~ UIColor.black.cgColor
+      |> \.shadowOpacity .~ 0.12
```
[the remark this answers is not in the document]
ifbarrera: Yeah that's kind of my rule of thumb, which is why I made the corner radius a constant. 🤷‍♀

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
ifbarrera: Oh ok. Will update 👍

dusi:
I was wondering about the way we handle presentation on iPad...in terms of the `width`. Currently we present project screen in full width (which I think is an oversight from when we implemented this - we didn't really kept iPad in mind)...so I would like us to consider limiting the width of the modal.

This could be done couple ways:

1) Setting the modal presentation style to `UIModalPresentationStyle.pageSheet` which uses full width on iPad in portrait mode and iPad's portrait width in landscape mode
2) Setting this manually (again to the lower value of iPad screen size or based on `UIInterfaceOrientationIdiom`

Example
sheet-width

dusi:
Also should we be able to control the expanded height of the modal? (I don't think we've discussed this)

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,59 @@
+import UIKit
+import Library
+import Prelude
+
+/**
+ SheetOverlayContainerViewController intended to be used as a container for another view controller
+ that you would like to render as a "sheet" or "card" that partially covers the content beneath it.
+ */
+
+final class SheetOverlayContainerViewController: UIViewController {
```
dusi: I've given this class some more thought and I have to ask if we even need it? Most of the work is done by the animator anyways...so with the same animator we should be able to present `UINavigationController` or just any other `UIVC` class...the animator would just have to ensure rounding corners and style (but I think that's fine coz it's already doing aplenty.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -188,19 +160,38 @@ public final class ProjectPamphletViewController: UIViewController {
...
-  private func configureSublayers() {
-    let updatedPath = UIBezierPath(roundedRect: self.backThisProjectContainerView.bounds,
-                                   byRoundingCorners: [.topLeft, .topRight],
-                                   cornerRadii: CGSize(width: 16, height: 16))
 
-    _ = self.backThisProjectContainerSublayer
-      |> \.path .~ updatedPath.cgPath
+  private func configureViews() {
+    // Configure subviews
+    self.backThisProjectContainerView.addSubview(self.backThisProjectButton)
```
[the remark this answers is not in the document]
ifbarrera: Ah yes I don't think this helper existed when this was written. Will update.

Kickstarter-iOS/Views/Transitions/SheetOverlayTransitionAnimator.swift
```diff
@@ -0,0 +1,99 @@
...
+                               transitionContext: transitionContext)
+    } else {
+      self.animateDismissal(fromViewController: fromVC,
+                            toViewController: toVC,
+                            containerView: containerView,
+                            transitionContext: transitionContext)
+    }
+  }
+
+  private func animatePresentation(
```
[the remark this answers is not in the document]
ifbarrera: I think it describes what's happening in these functions more accurately to use `animatePresentation` and `animateDismissal`

Kickstarter-iOS/Views/Transitions/SheetOverlayTransitionAnimator.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    let toFrame = toVC.view.frame.offsetBy(dx: 0, dy: fromVC.view.frame.height)
+
+    UIView.animate(
+      withDuration: self.transitionDuration(using: transitionContext),
+      delay: 0,
+      options: .curveEaseInOut,
+      animations: { [weak self] in
+        fromVC.view.frame = toFrame
+        self?.darkOverlayView.alpha = 0.0
```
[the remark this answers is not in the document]
ifbarrera: I noticed that the `ProjectNavigatorTransitionAnimator` didn't use lenses to set animatable properties - not sure if this was intentional, it could be a performance thing?

Kickstarter-iOS/Views/Transitions/SheetOverlayTransitionAnimator.swift
```diff
@@ -0,0 +1,99 @@
...
+        self?.darkOverlayView.alpha = 1.0
+    }, completion: { [weak self] _ in
+      toVC.view.backgroundColor = self?.darkOverlayView.backgroundColor
+      self?.darkOverlayView.removeFromSuperview()
+
+      transitionContext.completeTransition(!transitionContext.transitionWasCancelled)
+    })
+  }
+
+  private func animateDismissal(fromViewController fromVC: UIViewController,
```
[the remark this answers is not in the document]
ifbarrera: I'd say no, because it's super close to `UIKit` and there's no real business logic here.

Library/Styles/CheckoutStyles.swift
```diff
@@ -1,6 +1,10 @@
 import Prelude
 import UIKit
 
+public enum CheckoutStyleConstants {
```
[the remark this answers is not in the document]
ifbarrera: Hmmm good point, I'll change it to a layout style 👍

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+  private let transitionAnimator = SheetOverlayTransitionAnimator()
+
+  init(childViewController: UIViewController, childViewOffset: CGFloat) {
```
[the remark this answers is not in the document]
ifbarrera: I think having the default offset be zero kind of defeats the purpose of this behavior, but yeah can make it optional with a default of 45 or something else reasonable.

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+    super.init(nibName: nil, bundle: nil)
+
+    _ = self
+      |> \.modalPresentationStyle .~ .custom
+      |> \.transitioningDelegate .~ self
+
+    self.addChild(childViewController)
+    self.configureChildView(view: childViewController.view, offset: childViewOffset)
+
+    childViewController.willMove(toParent: self)
```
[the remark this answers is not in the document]
ifbarrera: Whoops yep you're right! Always forget the exact steps of adding a child VC 😅

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  private func configureChildView(view: UIView, offset: CGFloat) {
+    self.view.addSubview(view)
```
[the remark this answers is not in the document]
ifbarrera: I can update to use the convenience functions. We could use autolayout here but it felt a bit heavy-handed for something so simple. 🤔

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
ifbarrera: I think the benefit of having this class is that it abstracts the configuration of this behavior, which makes it very simple to use with one line of code:

`let sheetOverlayViewController = SheetOverlayContainerViewController(childViewController: navigationController, offset: 50)`

The `SheetOverlay` view controller takes care of holding a reference to the transition animator and conforming to `UIViewControllerTransitioningDelegate`, something that would have to be done by the presenting view controller if we didn't have the `SheetOverlay` class. 

I'd also be wary of putting too much view controller configuration code in the transition animator - I think we should try to limit the behaviour in the animator objects to purely animation-specific functionality for better separation of concerns.

ifbarrera:
> I was wondering about the way we handle presentation on iPad...in terms of the `width`. Currently we present project screen in full width (which I think is an oversight from when we implemented this - we didn't really kept iPad in mind)...so I would like us to consider limiting the width of the modal.
> 
> This could be done couple ways:
> 
> 1. Setting the modal presentation style to `UIModalPresentationStyle.pageSheet` which uses full width on iPad in portrait mode and iPad's portrait width in landscape mode
> 2. Setting this manually (again to the lower value of iPad screen size or based on `UIInterfaceOrientationIdiom`
> 
> Example
> sheet-width

Yep good point! Let me play around with it and see what makes the most sense.

ifbarrera:
> Also should we be able to control the expanded height of the modal? (I don't think we've discussed this)

What do you mean by this?

Kickstarter-iOS/Views/Controllers/SheetOverlayContainerViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+  private let transitionAnimator = SheetOverlayTransitionAnimator()
+
+  init(childViewController: UIViewController, childViewOffset: CGFloat) {
+    super.init(nibName: nil, bundle: nil)
+
+    _ = self
+      |> \.modalPresentationStyle .~ .custom
+      |> \.transitioningDelegate .~ self
+
+    self.addChild(childViewController)
```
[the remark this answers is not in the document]
ifbarrera: Interestingly, moving the child VC setup to the `viewDidLoad` actually breaks the animation 🤔The Apple docs don't specify when the right time to configure a child-parent relationship programmatically is, but they do say this about about configuration using interface builder: `"The children must be instantiated at the same time as the parent so that the appropriate parent-child relationships can be created."` Is it possible that this is also true for programmatic setup?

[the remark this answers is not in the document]
ifbarrera: ^ scratch that, the `modalPresentationStyle` is what needs to be set in the `init`.

Library/Styles/CheckoutStyles.swift
```diff
@@ -72,3 +68,8 @@ public let checkoutWhiteBackgroundStyle: ViewStyle = { (view: UIView) in
   view
     |> \.backgroundColor .~ UIColor.white
 }
+
+public let checkoutLayerCardStyleRounded: LayerStyle = { layer in
```
dusi: For consistency and maybe easier refactoring, could the `Style` be at the end of the constant? (i.e. `checkoutLayerCardRoundedStyle` or `checkoutRoundedCornerStyle`)?

Library/UIView+AutolayoutTests.swift
```diff
@@ -6,11 +6,49 @@ final class UIViewAutoLayoutExtensionTests: TestCase {
     let view1 = UIView(frame: .zero)
     let view2 = UIView(frame: .zero)
 
-    _ = (view2, view1)
+    let targetView = UIView(frame: .zero)
 
-    _ = ksr_insertSubviewInParent(at: 0)(view2, view1)
+    targetView.addSubview(view1)
```
dusi: Could this use `ksr_addSubviewToParent`?

ifbarrera: It could, but because we're testing one very specific `ksr` helper I don't want to couple another `ksr` function to these tests.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
dusi: We can potentially remove this now since the iOS 10 deprecation has been merged to `master` and should have been merged to `feature-native-checkout`

ifbarrera: They're removed here: https://github.com/kickstarter/ios-oss/pull/679

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+final class SheetOverlayViewController: UIViewController {
+  // MARK: - Properties
+
+  private let childViewController: UIViewController
+  private let offset: CGFloat
+  private let transitionAnimator = SheetOverlayTransitionAnimator()
+
+  init(childViewController: UIViewController, offset: CGFloat = 45.0) {
```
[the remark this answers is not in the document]
dusi: Unless your intension was to mirror the property (then `init(childViewController:)...` might probably be fine 👍

PR #675, opened 2019-05-09
Deprecate iOS 10

justinswart:
# 📲 What

- Removes support for iOS 10.
- Sets our minimum deployment target to iOS 11.4.
- Removes old support for tvOS.
- Removes deprecated social sharing types.

# 🤔 Why

Our iOS 10 usage has dropped below our threshold for maintaining support for it.

# 🛠 How

- Set minimum deployment target to iOS 11.4.
- Removed `#available(iOS 11.0, *)` checks.
- Removed old tvOS deployment target setting.
- Removed `#if os(iOS)` checks as we only support iOS in the app.

# ✅ Acceptance criteria

- [ ] No longer runs on iOS 10 devices.
- [x] I didn't miss any code/settings that would now be redundant with iOS 11.4+.
- [x] Confirm that all of our sharing works correctly through `UIActivityViewController`.


Library/UITraitCollection.swift
```diff
@@ -2,19 +2,6 @@ import UIKit
 
 public extension UITraitCollection {
   public func ksr_isAccessibilityCategory() -> Bool {
```
justinswart: We could remove the helper function 🤷‍♂

Library/ViewModels/ShareViewModel.swift
```diff
@@ -42,17 +41,11 @@ public protocol ShareViewModelInputs {
...
-  /// Call from the SLComposeViewController's completion handler
-  func shareComposeCompletion(result: SLComposeViewControllerResult)
-
   /// Call when the direct-share twitter button is pressed.
   func twitterButtonTapped()
 }
 
 public protocol ShareViewModelOutputs {
-  /// Emits when the share compose controller show be presented.
-  var showShareCompose: Signal<SLComposeViewController, NoError> { get }
```
justinswart: These outputs were not being used so I think all of the code referring to the old `Social` framework was unused.

Library/ViewModels/ShareViewModel.swift
```diff
@@ -1,10 +1,8 @@
-#if os(iOS)
 import KsApi
 import LiveStream
 import Prelude
 import ReactiveSwift
 import Result
-import Social
```
justinswart: We use `UIActivityViewController` for all of this now, no need to import `Social`.

Library/Koala/Koala.swift
```diff
@@ -2157,29 +2157,25 @@ private func properties(liveStreamEvent: LiveStreamEvent,
...
+    return "facebook"
+  } else if shareType == .message {
+    return "message"
+  } else if shareType == .mail {
+    return "email"
+  } else if shareType == .copyToPasteboard {
+    return "copy link"
+  } else if shareType == .postToTwitter {
+    return "twitter"
+  } else if shareType == UIActivity.ActivityType("com.apple.mobilenotes.SharingExtension") {
```
justinswart: Note-to-self to check if this is legit.

justinswart:
[iOS: Drop support for OS 10 and lower](https://trello.com/c/5aqWEwon/1352-ios-drop-support-for-os-10-and-lower)

Library/UITraitCollection.swift
dusi: Yep the sole purpose for this was iOS 10 support...so kill it with fire! :)

dusi:
It still looks like that some of our dependencies reference tvOS...any way to get rid of this?
Screen Shot 2019-05-10 at 10 47 07 AM

justinswart:
> It still looks like that some of our dependencies reference tvOS...any way to get rid of this?

All gone in #658 😂

Library/ViewModels/ShareViewModel.swift
```diff
@@ -42,17 +41,11 @@ public protocol ShareViewModelInputs {
   /// Call from the UIActivityViewController's completion handler.
   func shareActivityCompletion(with data: ShareActivityCompletionData)
 
-  /// Call from the SLComposeViewController's completion handler
-  func shareComposeCompletion(result: SLComposeViewControllerResult)
```
dusi: It also seems that `func facebookButtonTapped` and `func twitterButtonTapped` are unused

justinswart: Ah thanks!

Library/Koala/Koala.swift
justinswart: Appears legit - https://stackoverflow.com/questions/31792506/how-to-exclude-notes-and-reminders-apps-from-the-uiactivityviewcontroller

Library/Styles/SettingsStyles.swift
```diff
@@ -139,44 +139,27 @@ public func settingsLabelStyle(_ label: UILabel) -> UILabel {
...
 }
 
 public func settingsGroupedTableViewStyle(_ tableView: UITableView) -> UITableView {
-  let style = tableView
+  return tableView
     |> \.allowsSelection .~ false
     |> \.backgroundColor .~ .ksr_grey_200
     |> \.separatorInset .~ .zero
-
-  if #available(iOS 11, *) { } else {
```
ifbarrera: 🙌

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -62,11 +62,7 @@ public final class ProjectPamphletViewController: UIViewController {
   }
 
   private var initialTopConstraint: CGFloat {
-    if #available(iOS 11.0, *) {
-      return parent?.view.safeAreaInsets.top ?? 0.0
-    } else {
-      return UIApplication.shared.statusBarFrame.size.height
-    }
+    return self.parent?.view.safeAreaInsets.top ?? 0.0
```
ifbarrera: I think here we should just be able to use `self.view.safeAreaInsets.top` as well?

PR #677, opened 2019-05-13
Fix playgrounds using Carthage frameworks

justinswart:
# 📲 What

Fixes an annoying problem with Playgrounds in that they refuse to compile but without any meaningful error message:

```
error: Couldn't lookup symbols:
 protocol witness table for __C.UIViewController : PlaygroundSupport.PlaygroundLiveViewable in PlaygroundSupport
 PlaygroundSupport.PlaygroundPage.current.unsafeMutableAddressor : PlaygroundSupport.PlaygroundPage
 PlaygroundSupport.PlaygroundPage.liveView.setter : Swift.Optional<PlaygroundSupport.PlaygroundLiveViewable>
 type metadata accessor for PlaygroundSupport.PlaygroundPage
```

**Note:** This does not fix the following playgrounds as they are not compiling due to code-changes in the models and lenses that they touch. This should be fixed separately:

- `Comments`
- `CreatorDashboard`
- `FindFriends`
- `GraphQL`
- `LoginTout`
- `ProjectActivities`
- `Search`
- `Settings`
- `Signup`

# 🤔 Why

We'd like to be able to use Playgrounds again as we've done in the past.

# 🛠 How

What appears to be key in the above error message is `Couldn't lookup symbols`. What is not helpful is that the error does not tell us which symbols couldn't be looked up. After some extensive Googling and trying many different things, @dusi and I were able to determine that:

All Carthage-built frameworks that are `import`ed in a target _must_ be:
1. Added to added to that target's `Linked Frameworks and Libraries` (even if the target builds without them being there).
2. Be added to a `Copy Files` build phase which copies them to the `Frameworks` directory.

One then needs to clean the build folder, rebuild and sometimes exit and restart Xcode (🤦‍♂ ).

This appears to, with relative reliability, get playgrounds to compile again.

In order to find all of the necessary imports we ran the command `grep -r -h 'import' . | sort -u | uniq` in terminal and that target's root folder respectively which produced a list like:

Screen Shot 2019-05-13 at 12 05 02 PM

We then, one-by-one, made sure that each of these was added in the correct places as explained above.

# 👀 See

http://justthesam.com/2015/12/using-a-carthage-framework-in-a-playground/
https://m.pardel.net/using-3rd-party-frameworks-in-swift-playgrounds-fd415f6dba10
https://blog.mazur.me/playground-driven-development-and-carthage/
https://www.google.com/search?q=carthage+playgrounds

Screen Shot 2019-05-10 at 4 01 37 PM

# ✅ Acceptance criteria

- [x] Playgrounds do not crash with the obscure symbols error.

# ⏰ TODO

- [ ] Update Playgrounds that are not compiling due to code changes in separate PR.


justinswart:
[Fix playgrounds](https://trello.com/c/zgURZKXS/1360-fix-playgrounds)

dusi:
Would it be possible to sort `Embedded Binaries` & `Linked Frameworks and Libraries` alphabetically for easier readability / scanning through?

justinswart:
> Would it be possible to sort Embedded Binaries & Linked Frameworks and Libraries alphabetically for easier readability / scanning through?

😑

justinswart:
One can't seem to reorder Embedded Binaries, but did it for the linked libraries! 😬 😅

dusi:
I'm sorry about the `Embedded Binaries` @justinswart , I've realized those can't be sorted couple hours after suggesting it...at least it works for `Linked Frameworks..`

PR #680, opened 2019-05-17
Default recommended projects

Scollaco:
# 📲 What
Changes Discover behavior by setting `Recommended For You` as default filter if user is logged in. If the user is logged out, the behavior shouldn't change. 

# 🤔 Why
This feature existed in the past, but due GDPR we had decided to remove it temporarily until have a better understanding of its implications.

# 🛠 How
But changing the `initialParams` on `DiscoveryViewModel` to include recommended projects if user is logged in.

# 👀 See
recs2

# ✅ Acceptance criteria
- [x] Being logged out, I open the app and the default filter should be `All Projects`
- [x] I login on my account
- [x] If my Recommendations setting on `Settings > Account > Privacy` is `OFF`,  Discovery should still show `All Projects` as default filter.
- [x] If my Recommendations setting on `Settings > Account > Privacy` is `ON`,  Discovery should show `Recommended for you` as default filter.
- [x] Opting out of recommendations in `Settings > Account > Privacy` and going back to Explore tab should show `All Projects` even if `Recommended for you` was previously selected
- [x] I logout and Discovery should show `All Projects` as default filter.


Scollaco:
[Default discovery feed to recommended projects](https://trello.com/c/08dG7fUr/1372-default-discovery-feed-to-recommended-projects)

Scollaco:
[Default discovery feed to recommended projects](https://trello.com/c/08dG7fUr/1372-default-discovery-feed-to-recommended-projects)

Library/ViewModels/DiscoveryViewModelTests.swift
```diff
@@ -85,6 +83,37 @@ internal final class DiscoveryViewModelTests: TestCase {
...
+
+    withEnvironment(config: Config.template, currentUser: user) {
+      self.vm.inputs.viewDidLoad()
+      self.vm.inputs.viewWillAppear(animated: false)
+
+      self.configureNavigationHeader.assertValues([initialParams])
+    }
+  }
+
+  func testLoadRecommendedProjectsIntoDataSource_UserOptedInOfRecommendations() {
```
dusi: Somehow `OptedInOf` sounds really weird to me ... would it make more sense if we used the following?

```
_UserRecommendationsOptedOut
```
and
```
_UserRecommendationsOptedIn
```

Library/Notifications.swift
```diff
@@ -14,6 +15,7 @@ public enum CurrentUserNotifications {
 
 public enum UserInfoKeys {
   public static let context = "UserInfoKeys.context"
+  public static let userOptedOutRecommendations = "UserInfoKeys.userOptedOutRecommendations"
```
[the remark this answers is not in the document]
dusi: BTW is this being used anywhere? Looks like it's not referenced from anywhere else in the codebase? 🤔

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -7,6 +7,10 @@ import ReactiveExtensions
 import Result
 
 public protocol DiscoveryViewModelInputs {
+
+  /// Call when Recommendations setting changes on Settings > Account > Privacy > Recommendations.
+  func didChangeRecommendationsSetting()
```
dusi: The notification uses singular `recommendationSettingChanged` (for `recommendation`) yet here we're using plural (`recommendations`)...would you mind making these consistent?

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -63,8 +67,18 @@ public protocol DiscoveryViewModelType {
 
 private func initialParam() -> DiscoveryParams {
```
justinswart: This doesn't need to be an instance function as far as I can tell?

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -77,10 +91,11 @@ DiscoveryViewModelOutputs {
     self.configureSortPager = self.configurePagerDataSource
 
     let currentParams = Signal.merge(
-      self.viewWillAppearProperty.signal.take(first: 1).map { _ in initialParam() },
+      self.didChangeRecommendationsSettingProperty.signal.map { _ in initialParam() },
```
justinswart: Could we rather merge the two for the `initialParam`:

```swift
let initialParam = Signal.merge(
  self.didChangeRecommendationsSettingProperty.signal,
  self.viewWillAppearProperty.signal
)
.ignoreValues()
.map { initialParam() } // if we move initialParam() to file scope this can be point-free.

let currentParams = Signal.merge(
  initialParam,
  self.filterWithParamsProperty.signal.skipNil()
)
```

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -63,8 +67,18 @@ public protocol DiscoveryViewModelType {
...
 
+  let user = AppEnvironment.current.currentUser
+  let optedOutOfRecommendations = user?.optedOutOfRecommendations ?? true
+
+  if user == nil || optedOutOfRecommendations {
     return DiscoveryParams.defaults
       |> DiscoveryParams.lens.includePOTD .~ true
+  }
+
+  return DiscoveryParams.defaults
```
justinswart: If the user is not nil we return these defaults?

Library/Notifications.swift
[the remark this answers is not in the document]
Scollaco: It was being used before, but after making some changes, it became unnecessary. Will clean it up.

Library/ViewModels/DiscoveryViewModel.swift
Scollaco: Good idea. I refactored the code following your suggestion, but had to change the implementation slightly in order to cover all edge cases.

Library/ViewModels/DiscoveryViewModel.swift
Scollaco: We return these defaults if:
1- User is nil
2- User is not nil and ` user.optedOutOfRecommendations == true`

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -61,26 +65,42 @@ public protocol DiscoveryViewModelType {
...
 
-private func initialParam() -> DiscoveryParams {
+public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewModelInputs,
+DiscoveryViewModelOutputs {
+
+  private static func initialParams() -> DiscoveryParams {
+    let user = AppEnvironment.current.currentUser
+    let optedOutOfRecommendations = user?.optedOutOfRecommendations ?? true
 
+    if user == nil || optedOutOfRecommendations {
```
justinswart: I think the logic still confuses me a little with the `||` and the coalescing to `true`, how about this?

```swift
  private static func initialParams() -> DiscoveryParams {
    guard AppEnvironment.current.currentUser?.optedOutOfRecommendations == .some(false) else {
      return DiscoveryParams.defaults
        |> DiscoveryParams.lens.includePOTD .~ true
    }

    return DiscoveryParams.defaults
      |> DiscoveryParams.lens.includePOTD .~ true
      |> DiscoveryParams.lens.backed .~ false
      |> DiscoveryParams.lens.recommended .~ true
  }
```
The way I read this is - "By default we are opted out of recommendations (GDPR), so unless this is explicitly `false` (user intervention), we only display POTD." Does that sound right?

Scollaco: I like this solution. It's much simpler!

justinswart:
Hey I think the only thing left to figure out is why this generated so many new snapshots?

Scollaco:
> Hey I think the only thing left to figure out is why this generated so many new snapshots?

Ok, I think the problem was that I'd generated snapshots using the iPhoneX simulator 🤦‍♂ . I reverted the commits and once I re-ran the tests using the iphone 8 simulator, all the viewController tests passed, so snapshots were no longer necessary.

PR #681, opened 2019-05-17
🚫 Danger CI Integration

dusi:
# 📲 What

Adds `Danger` as one of our CI build steps. That runs:
- Swiftlint
- SwiftFormat

# 🤔 Why

Danger will allow us to enforce code format (`SwiftFormat`) as well as code linter (`SwiftLint`) and hopefully take over the role of being opinionated about our code so that we don't have to.

# 🛠 How

Integrated as Ruby gem...I've initially tried setting things up using SPM but it showed more tricky than I was able to grasp on Friday 🍺 Since we're currently in a desperate need of using these tools the Ruby version out values having a Swift integration. We can always migrate in the future.

# ✅ Acceptance criteria

- [x] Danger should complain about the size of this PR (as that's currently the only rule)

# ⏰ TODO

- [x] Discuss pros of using `Only build pull requests` on our team as a CircleCI setting


dusi:
New phone .. who dis?

.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.2
```
dusi: Remove

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ base_job: &base_job
     XCODE_VERSION: *xcode_version
 
 jobs:
-  swiftlint-and-cache:
+  build-and-cache:
```
dusi: This renaming will require tweaking our `master` branch protection rule but since this step is no longer going to be using `Swiftlint` we should use a better naming

.circleci/config.yml
```diff
@@ -44,53 +44,73 @@ jobs:
...
             - fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
 
       - restore_cache:
+          name: Restore Stripe SDK
           keys:
             - stripe-sdk-cache-{{ checksum "stripe_version.txt" }}
 
       - run:
-          name: Store Xcode Version
+          name: Cache Xcode Version
```
justinswart: I think this should still be `Store Xcode Version` it does the same thing as `Store Stripe Version` and `Store Fabric SDK Version` above. It's basically writing the Xcode version to a file so that it can later be `checksum`'d as a part of the cache key for restoring the Carthage cache.

This just works around a shortcoming in CircleCI's caching mechanism in that you can't access custom environment vars in the `yml` file 😕

.circleci/config.yml
```diff
@@ -44,53 +44,73 @@ jobs:
...
 
       - restore_cache:
-          name: Restore Carthage Cache
+          name: Restore Carthage
           keys:
             - carthage-cache-v1-{{ checksum "Cartfile.resolved" }}-{{ checksum "xcode_version.txt" }}
 
       - run:
-          name: Make bootstrap
+          name: Bootstrap
```
justinswart: A lot of these names repeat throughout this file, I'd either not make the change or update it everywhere 😅

Makefile
```diff
@@ -41,11 +41,6 @@ clean:
 dependencies: carthage-bootstrap configs secrets fabric
 
 bootstrap: hooks dependencies
-	brew update || brew update
-	brew unlink swiftlint || true
-	brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/686375d8bc672a439ca9fcf27794a394239b3ee6/Formula/swiftlint.rb
```
justinswart: If we're removing the pinning of `swiftlint` here do we have the binary checked into the repo?

dusi: Good question...I'll make it so

.circleci/config.yml
```diff
@@ -44,10 +44,12 @@ jobs:
           command: *store_stripe_version
 
       - restore_cache:
+          name: Restore Fabric SDK
           keys:
             - fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
 
       - restore_cache:
+          name: Restore Stripe SDK
```
justinswart: Note to self: Stripe caching should have been removed in #658. Will open a separate PR for that.

justinswart: #685

PR #683, opened 2019-05-22
💲[Native Checkout] Reward Selection (no animation)

ifbarrera:
# 📲 What

Adds functionality for navigating to the `Pledge` screen when a reward is tapped from the `RewardsCollectionViewController`. Note that right now it's a simple `push` navigation, since we plan on implementing the custom transition in a separate item of work.

# 🤔 Why

So we can finally have an end-to-end flow for the new native checkout experience ❇️ 

# 🛠 How

Overrides the reward collection view's `collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath)` function, and passes the selected index to the view model as an input. The view model then grabs the reward selected at the given index, and combines it with the project object and the optional `refTag` to pass as an output. The `goToPledge` output is observed and subsequently creates and pushes the `PledgeViewController`.

# 👀 See

BMklK0dMkq

# ♿️ Accessibility 

☝️ Note that right now the reward cells don't have an accessibility value, because it will be set as the title of the reward once we actually populate the reward cells with content.

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Log in on an *admin* account to staging. Background the app, then open it again. Navigate to any project. Tap "Back this Project". Tap on any reward "card" and you should then be navigated to the pledge screen.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
@@ -13,7 +13,7 @@ final class RewardsCollectionViewController: UICollectionViewController {
   // MARK: - Properties
 
   private let dataSource = RewardsCollectionViewDataSource()
-  private let viewModel = RewardsCollectionViewModel()
+  fileprivate let viewModel = RewardsCollectionViewModel()
 
   private let hiddenPagingScrollView: UIScrollView = {
     UIScrollView()
@@ -111,6 +111,12 @@ final class RewardsCollectionViewController: UICollectionViewController {
         self?.dataSource.load(rewards: rewards)
         self?.collectionView.reloadData()
     }
+
+    self.viewModel.outputs.goToPledge
+      .observeForControllerAction()
+      .observeValues { [weak self] project, reward, refTag in
+        self?.goToPledge(project: project, reward: reward, refTag: refTag)
+    }
   }
 
   // MARK: - Private Helpers
@@ -190,10 +196,25 @@ final class RewardsCollectionViewController: UICollectionViewController {
     return CGSize(width: itemWidth, height: itemHeight)
   }
 
+  private func goToPledge(project: Project, reward: Reward, refTag: RefTag?) {
+    let pledgeViewController = PledgeViewController.instantiate()
+    pledgeViewController.configureWith(project: project, reward: reward)
+
+    self.navigationController?.pushViewController(pledgeViewController, animated: true)
+  }
+
   // MARK: - Public Functions
 
   @objc func closeButtonTapped() {
     self.navigationController?.dismiss(animated: true)
+  }
+}
+
+// MARK: - UICollectionViewDelegate
+
+extension RewardsCollectionViewController {
+  override func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+    self.viewModel.inputs.rewardSelected(at: indexPath.row)
   }
 }
 
--- a/Library/ViewModels/RewardsCollectionViewModel.swift
+++ b/Library/ViewModels/RewardsCollectionViewModel.swift
@@ -4,10 +4,10 @@ import Result
 import KsApi
 import Prelude
 
-public typealias GoToPledgeData = (project: Project, reward: Reward, refTag: RefTag?)
+public typealias PledgeData = (project: Project, reward: Reward, refTag: RefTag?)
 
 public protocol RewardsCollectionViewModelOutputs {
-  var goToPledge: Signal<GoToPledgeData, NoError> { get }
+  var goToPledge: Signal<PledgeData, NoError> { get }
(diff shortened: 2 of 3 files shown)


ifbarrera:
[Reward Selection - No Animation](https://trello.com/c/FByCNlid/1368-reward-selection-no-animation)

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -190,13 +196,28 @@ final class RewardsCollectionViewController: UICollectionViewController {
     return CGSize(width: itemWidth, height: itemHeight)
   }
 
+  private func goToPledge(project: Project, reward: Reward, refTag: RefTag?) {
```
Scollaco: The `refTag` argument is not being used here. Could you remove it from the function signature?

Scollaco: Is there a reason why we are passing the `refTag` on this function? It's not being used here.

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -27,6 +31,17 @@ RewardsCollectionViewModelInputs, RewardsCollectionViewModelOutputs {
...
+
+    let selectedReward = reloadDataWithRewards
+      .takePairWhen(self.rewardSelectedIndexProperty.signal.skipNil())
+      .map { rewards, index in rewards[index] }
+
+    self.goToPledge = Signal.combineLatest(self.configureWithProjectProperty.signal.skipNil(),
+      selectedReward,
+      self.configureWithRefTagProperty.signal)
+      .map { project, reward, refTag in
+        return GoToPledgeData(project: project, reward: reward, refTag: refTag)
```
Scollaco: A suggestion here: we usually use names starting with `goTo` to define an action that brings the user to a different screen in the app. Maybe `PledgeData(project:reward:refTag)` would be better?

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
ifbarrera: Yeah we're definitely going to need to pass it to the `PledgeViewController`, so I added it in anticipation for that.

Library/ViewModels/RewardsCollectionViewModel.swift
ifbarrera: Yep that's better, thanks for the suggestion!

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -13,7 +13,7 @@ final class RewardsCollectionViewController: UICollectionViewController {
   // MARK: - Properties
 
   private let dataSource = RewardsCollectionViewDataSource()
-  private let viewModel = RewardsCollectionViewModel()
+  fileprivate let viewModel = RewardsCollectionViewModel()
```
justinswart: Why did the visibility change? 🤔

ifbarrera: Ah good catch! I'd forgotten that Swift 4 extended `private` access to include extensions in the same file 👍

Library/ViewModels/RewardsCollectionViewModelTests.swift
```diff
@@ -33,4 +39,29 @@ final class RewardsCollectionViewModelTests: TestCase {
...
+
+    self.goToPledgeProject.assertValues([project])
+    //swiftlint:disable:next force_unwrapping
+    self.goToPledgeReward.assertValues([project.rewards.first!])
+    self.goToPledgeRefTag.assertValues([.activity])
+
+    self.vm.inputs.rewardSelected(at: project.rewards.endIndex - 1)
+
+    self.goToPledgeProject.assertValues([project, project])
+    //swiftlint:disable:next force_unwrapping
```
dusi:
```suggestion
    // swiftlint:disable:next force_unwrapping
```

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -27,6 +31,17 @@ RewardsCollectionViewModelInputs, RewardsCollectionViewModelOutputs {
     )
       .map(first)
       .map { $0.rewards }
+
+    let selectedReward = reloadDataWithRewards
+      .takePairWhen(self.rewardSelectedIndexProperty.signal.skipNil())
+      .map { rewards, index in rewards[index] }
```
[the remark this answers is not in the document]
ifbarrera: Since the `dataSource` is populated with rewards, I'd say it's highly unlikely that we would ever have an out of index exception. I'd err on the side of "let's keep an eye on it" and address the underlying issue if it comes up, since that would be a symptom of a more serious bug somewhere else in this view model if that were to happen.

PR #684, opened 2019-05-23
Fix Carthage App Store submission

justinswart:
# 📲 What

Fixes an error that we were receiving when submitting to the App Store, builds being rejected on upload with the error `ERROR ITMS-90206: "Invalid Bundle. The bundle at '<nested framework>' contains disallowed file 'Frameworks'."`

Where `<nested framework>` was `Library`, `KsApi` or `Kickstarter_Framework`.

# 🤔 Why

This began happening since the work we did in #658, #677 and #678. Most notably the work in #677 which adds a workaround for Playgrounds by copying 3rd party frameworks to the `/Frameworks` directory of each respective nested framework in order for Playgrounds to access them.

While this does appear to fix Playgrounds, Apple does not allow frameworks to be nested in this way as it is the app bundle's responsibility to make these 3rd party frameworks available to its own nested frameworks by embedding them. See [TN2435](https://developer.apple.com/library/archive/technotes/tn2435/_index.html).

# 🛠 How

In order to have playgrounds work in Xcode but not upload the 3rd party frameworks to the App Store we wrapped our `carthage copy-frameworks` script in an `if` statement in a `bash` script to only run in `Debug` configuration. It took some time to realize that this script _was_ running correctly but, what appeared to be an Xcode bug, was resulting in an empty `Frameworks` directory to be created in each of our nested frameworks at the time of archiving. The only reason that I can surmise for this is that we have input and output `.xcfilelist` files which Xcode may be parsing and preparing a `Frameworks` directory for.

The final workaround for this then is to add an `else` condition which removes that directory if it exists 🤷‍♂️. This appears to also be an occasional issue with Cocoapods: https://stackoverflow.com/a/33092433.

We are now successfully uploading to the app store again.


Carthage-xcfilelist/app-input-files.xcfilelist
```diff
@@ -9,7 +9,6 @@ $(SRCROOT)/Carthage/Build/iOS/OnePasswordExtension.framework
 $(SRCROOT)/Carthage/Build/iOS/Prelude.framework
 $(SRCROOT)/Carthage/Build/iOS/Prelude_UIKit.framework
 $(SRCROOT)/Carthage/Build/iOS/ReactiveExtensions.framework
-$(SRCROOT)/Carthage/Build/iOS/ReactiveExtensions_TestHelpers.framework
```
justinswart: Although this didn't appear to be problematic for the upload, we should not be bundling our test helpers with the app.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -4020,7 +4020,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "bin/carthage-debug-copy.sh\n";
+			shellScript = "${SRCROOT}/bin/carthage-debug-copy.sh\n";
```
justinswart: Added `SRCROOT` for safety 😅

PR #688, opened 2019-05-24
Update for Xcode 10.2

justinswart:
# 📲 What

Updates our codebase to compile in Xcode 10.2.

# 🤔 Why

We've been held back a little on being able to use Xcode 10.2 and Swift 5. This PR aims to make the necessary changes for us to use Xcode 10.2 but, unfortunately, not Swift 5 just yet:

While attempting to compile the app using Swift 5 I ran into the following error:

Screen Shot 2019-05-24 at 1 00 56 PM

Although I'm not entirely certain just yet, it appears that our use of keypath's for lensing _might_ have been exploiting a bug in Swift 4.2 in that `WritableKeyPath` was allowing us to bypass private setters 🤔 . This may be related: https://forums.swift.org/t/new-ambiguity-with-keypaths-in-swift-5/20668. I'll continue investigating but this is problematic if true.

# 🛠 How

- Updated ReactiveExtensions to compile with Carthage.
- Moved HockeySDK to be downloaded by our download script (https://github.com/Carthage/Carthage/issues/2738).
- Updated CircleCI config to use the Xcode 10.2.1 image.
- Updated ReactiveSwift to [5.0.1](https://github.com/ReactiveCocoa/ReactiveSwift/releases/tag/5.0.1).
  - I've stopped short of updating to [6.0.0](https://github.com/ReactiveCocoa/ReactiveSwift/releases/tag/6.0.0) because that deprecates ReactiveSwift's `Result` type in favour of the one that is now included in Swift 5.

PR #689, opened 2019-05-28
Update to Swift 5

justinswart:
# 📲 What

As a follow-up to #688 this updates our project settings to use Swift 5. It also changes all of our `fileprivate(set) var` and `private(set) var` modifiers in our `KsApi` models to be `public var`.

**Note:** This does _not_ silence modifier warnings throughout the codebase (redundant `public`, etc.). It's my expectation that we will use SwiftFormat to update these in one pass.

# 🤔 Why

We like Swift and we particularly like to use the latest Swift.

We learnt in #688 that the way in which we were using `WritableKeyPath<Root, Value>` was in fact exploiting a bug that existed in Swift prior to 5.0, that is - inaccessible setters should not be accessible via `WritableKeyPath`s. This is corroborated by [this PR](https://github.com/apple/swift/pull/16452) in the Swift repo. The only reasonable way to solve this, as far as I can see, is to simply make these setters `public var`.

# 🛠 How

- `index(of:)` -> `firstIndex(of:)`.
- `index(where:)` -> `firstIndex(where:)`.
- Explicitly cast to `Any` wherever we had warnings about a type being coerced to `Any`.
- `fileprivate(set) var`, `private(set) var` -> `public var`.

# ✅ Acceptance criteria

- [ ] We're happy to make our setters `public`.
- [x] Everything builds and runs as before.


justinswart:
[Update KeyPath lenses to work with Swift 5](https://trello.com/c/f7OIXHWA/1381-update-keypath-lenses-to-work-with-swift-5)

justinswart:
[iOS: Upgrade to Swift 5 and Xcode 10.2 compatibility](https://trello.com/c/ycihRBvl/1366-ios-upgrade-to-swift-5-and-xcode-102-compatibility)

dusi:
Just remembered we should also update the `SwiftFormat` version in the `bin/format` helper - which is currently set to 4.2 `--swiftversion 4.2`

justinswart:
Note, I won't be addressing the formatting issues in this PR to keep its size relatively small. The plan is for @dusi to run that on the whole codebase.

Kickstarter-iOS/Views/Cells/DashboardActionCell.swift
```diff
@@ -45,6 +45,7 @@ internal final class DashboardActionCell: UITableViewCell, ValueCell {
     _ = self |> baseTableViewCellStyle()
     self.isAccessibilityElement = false
     self.accessibilityElements = [self.activityButton, self.messagesButton, self.postUpdateButton]
+      .map { $0 as Any }
```
dusi: Should we use `.compactMap { $0 }` to explicitly state that we're dealing with optionals?

BTW: we're not calling `super.bindStyles()` (not sure if it's intentional here)

justinswart: Ah, sure I can change that to `compactMap`. I can add the super call... or perhaps a separate PR, don't really want to make any functional changes in this kind of PR.

PR #691, opened 2019-05-28
💲[Native Checkout] Shipping location cell plus sign

dusi:
# 📲 What

Adds plus sign prefix to the shipping location cell

# 🤔 Why

Designs.

# 🛠 How

We're still using localized string (except we don't pass any value to it - since we only want to get back plus sign and potentially a space in some languages - i.e. French)

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-05-28 at 4 21 53 PM | Screen Shot 2019-05-28 at 4 22 19 PM |

# ✅ Acceptance criteria

- [x] Shipping location amount is prefixed with a plus sign `+` and is localized (i.e. instead of `+` it is `+ ` in French)

Please note that currently Dynamic Type is broken for the shipping location cell - this will be addressed in #693


Library/String+Attributed.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+func + (left: NSAttributedString, right: NSAttributedString) -> NSAttributedString {
```
dusi: This allows us concatenating two attributed strings like so: `first + second`

dusi:
[[Checkout] Pledge Screen - Shipping Location Cell](https://trello.com/c/01sl7Var/1287-checkout-pledge-screen-shipping-location-cell)

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -46,7 +46,7 @@ final class PledgeViewModelTests: TestCase {
         [Format.date(secondsInUTC: estimatedDelivery, template: "MMMMyyyy", timeZone: UTCTimeZone)]
       )
       self.shippingLocation.assertValues(["Brooklyn"])
-      self.shippingAmount.assertValues(["$7.50"])
+      self.shippingAmount.assertValues(["+$7.50"])
```
justinswart: Did we want to add a test here for French to show the space after the `+`?

Library/String+Attributed.swift
justinswart: Fancy!

dusi:
@ifbarrera do you want me to put blocker on this PR in order for your branch to catch up? This doesn't have that many architectural changes to the `feature-native-checkout` branch so should not cause you much harm (but feel free to review first)

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -44,13 +44,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
           .map { Format.date(secondsInUTC: $0, template: "MMMMyyyy", timeZone: UTCTimeZone) } ?? ""
         let shipping = (
           "Brooklyn",
-          Format.attributedCurrency(
```
ifbarrera: If you take a look at https://github.com/kickstarter/ios-oss/pull/682/files, I've already removed this call to `Format.attributedCurrency` in favor of sending the `Project` directly through to the cell, where we can pipe it into a `ViewModel` that does this formatting. How should we resolve these two approaches?

One thing we could do is leave the `shippingValue(for: project)` function at the bottom of this file, but remove it from anywhere that it's called - that way we can simply move it over to the `PledgeShippingLocationCellViewModel` for usage once that refactor happens, and we shouldn't have any merge conflicts. What do you think?

PR #692, opened 2019-05-29
Ensure tests are only run on iPhone 8

justinswart:
# 📲 What

Ensures that we only run tests and record screenshots on the iPhone 8 simulator.

If for whatever reason we need to run unit tests on a different device, we can just comment out this line.

# 🤔 Why

It's happened once or twice in the past that we accidentally record a bunch of screenshots on the wrong device which results in thousands of new screenshots being committed to the repo. This is just a safety catch to help us prevent that.

# 🛠 How

Put a check at the top of `TestCase` `setUp()` function which is called before all tests run.

# ✅ Acceptance criteria

- [ ] All tests pass as before.
- [ ] Running tests on any other device causes Xcode to fail and to not continue running tests.

### DIFF
--- a/Library/TestHelpers/TestCase.swift
+++ b/Library/TestHelpers/TestCase.swift
@@ -26,10 +26,7 @@ internal class TestCase: FBSnapshotTestCase {
   override func setUp() {
     super.setUp()
 
-    let preferredDeviceName = "iPhone 8"
-    if !UIDevice.current.name.contains(preferredDeviceName) {
-      fatalError("Please only test and record screenshots on \(preferredDeviceName)")
-    }
+    preferredSimulatorCheck()
 
     UIView.doBadSwizzleStuff()
     UIViewController.doBadSwizzleStuff()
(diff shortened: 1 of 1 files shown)


Library/TestHelpers/TestCase.swift
```diff
@@ -25,6 +25,12 @@ internal class TestCase: FBSnapshotTestCase {
 
   override func setUp() {
     super.setUp()
+
+    let preferredDeviceName = "iPhone 8"
+    if !UIDevice.current.name.contains(preferredDeviceName) {
```
dusi: Should we be getting this from `AppEnvironment.current` so that we can potentially inject our own value down the road (and maybe for consistency)?

justinswart: I don't think this is really for `AppEnvironment`, that's really only for dependency injection and I don't know that we'd want to inject stuff in unit tests _for_ unit tests 🤔 I think this is like a safety thing much like we have in our swiftlint.yml to fail if we leave `self.recordMode = true` in the test.

PR #693, opened 2019-05-30
💲[Native Checkout] Refactor cells to be backed by view model

dusi:
# 📲 What

Refactors pledge screen architecture to back each cell with a view model (this should be consistent with the rest of the codebase and make updating UI slightly less frustrating)

# 🤔 Why

By passing values from VC's VM it was challenging to pass the right font size which broke Dynamic Support in the process.

# 🛠 How

Each cell now has a VM that's populated only with information that it needs for the cell to display values. In addition to that we now can bind `text` / `attributedText` props on specific UI components making it more consistent with the rest of the codebase.

# 👀 See

Notice how the shipping location amount didn't scale previously.

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-05-30 at 11 01 58 AM | Screen Shot 2019-05-30 at 10 57 01 AM |

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Shipping location supports Dynamic Type scaling properly
- [ ] Shipping amount supports Dynamic Type scaling properly

### DIFF
--- a/Kickstarter-iOS/Views/AmountInputView.swift
+++ b/Kickstarter-iOS/Views/AmountInputView.swift
@@ -5,8 +5,8 @@ import UIKit
 class AmountInputView: UIView {
   // MARK: - Properties
 
-  private lazy var label: UILabel = { UILabel(frame: .zero) }()
-  private lazy var textField: UITextField = { UITextField(frame: .zero) }()
+  private(set) lazy var label: UILabel = { UILabel(frame: .zero) }()
+  private(set) lazy var textField: UITextField = { UITextField(frame: .zero) }()
   private lazy var stackView: UIStackView = { UIStackView(frame: .zero) }()
   private var labelCenterYAnchor: NSLayoutConstraint?
 
--- a/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
@@ -1,3 +1,4 @@
+import KsApi
 import Library
 import Prelude
 import Prelude_UIKit
@@ -5,6 +6,8 @@ import UIKit
 
 final class PledgeAmountCell: UITableViewCell, ValueCell {
   // MARK: - Properties
+
+  private let viewModel = PledgeAmountCellViewModel()
 
   private lazy var adaptableStackView: UIStackView = { UIStackView(frame: .zero) }()
   private lazy var amountInputView: AmountInputView = { AmountInputView(frame: .zero) }()
@@ -36,6 +39,8 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
       |> ksr_addArrangedSubviewsToStackView()
 
     self.spacer.widthAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(3)).isActive = true
+
+    self.bindViewModel()
   }
 
   required init?(coder _: NSCoder) {
@@ -68,18 +73,19 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
       |> stepperStyle
   }
 
+  // MARK: - Binding
+
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.amountInputView.label.rac.text = self.viewModel.outputs.currency
+    self.amountInputView.textField.rac.text = self.viewModel.outputs.amount
+  }
+
   // MARK: - Configuration
 
-  func configureWith(value: PledgeDataSource.PledgeInputRow) {
-    guard case let .pledgeAmount(amount, currencySymbol) = value else {
-      return
-    }
-
-    self.amountInputView.configureWith(
-      amount: String(format: "%i", amount),
-      placeholder: "\(0)",
-      currency: currencySymbol
-    )
+  func configureWith(value: (project: Project, reward: Reward)) {
+    self.viewModel.inputs.configureWith(project: value.project, reward: value.reward)
   }
 }
 
--- a/Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
@@ -104,11 +104,7 @@ public final class ProjectPamphletContentViewController: UITableViewController {
     }
   }
 
-  public override func tableView(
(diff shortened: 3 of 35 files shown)


Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -5,8 +5,8 @@ import UIKit
 class AmountInputView: UIView {
   // MARK: - Properties
 
-  private lazy var label: UILabel = { UILabel(frame: .zero) }()
-  private lazy var textField: UITextField = { UITextField(frame: .zero) }()
+  private(set) lazy var label: UILabel = { UILabel(frame: .zero) }()
```
dusi: Exposing these in order for us to being able to bind these with `rac` extensions

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -36,6 +39,8 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
       |> ksr_addArrangedSubviewsToStackView()
 
     self.spacer.widthAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(3)).isActive = true
+
+    self.bindViewModel()
```
dusi: We should consider swizzling `UITableViewCell` and do this automatically in `initWithFrame` for cells created programmatically

Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
```diff
@@ -104,11 +104,7 @@ public final class ProjectPamphletContentViewController: UITableViewController {
     }
   }
 
-  public override func tableView(
-    _: UITableView,
-    willDisplay cell: UITableViewCell,
-    forRowAt _: IndexPath
-  ) {
+  public override func tableView(_: UITableView, willDisplay cell: UITableViewCell, forRowAt _: IndexPath) {
```
dusi: Since `SwiftFormat` stripped unused arguments it now fits on 1 line 😎

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: Yeah, @ifbarrera and I worked on a solution for this that should work for all our programmatic views, will open a PR.

Library/ViewModels/PledgeAmountCellViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    let reward = self.projectAndRewardProperty.signal
+      .skipNil()
+      .map(second)
+
+    self.amount = reward
+      .map { String(format: "%.0f", $0.minimum) }
+
+    self.currency = project
+      .map { currencySymbol(forCountry: $0.country).trimmed() }
```
justinswart: Would we ever want `currencySymbol` to not be trimmed? we could move that to inside the function.

Library/ViewModels/PledgeDescriptionCellViewModel.swift
```diff
@@ -21,14 +21,18 @@ public protocol PledgeDescriptionCellViewModelType {
 public final class PledgeDescriptionCellViewModel: PledgeDescriptionCellViewModelType,
   PledgeDescriptionCellViewModelInputs, PledgeDescriptionCellViewModelOutputs {
   public init() {
-    self.estimatedDeliveryText = self.estimatedDeliveryDateProperty.signal
+    self.estimatedDeliveryText = self.rewardProperty.signal
+      .skipNil()
+      .map { $0.estimatedDeliveryOn }
+      .skipNil()
+      .map { Format.date(secondsInUTC: $0, template: "MMMMyyyy", timeZone: UTCTimeZone) }
```
justinswart: `MMMMyyyy` repeats a couple times in the codebase. Candidate for a constant?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -68,7 +45,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     self.viewDidLoadProperty.value = ()
   }
 
-  public let reloadWithData: Signal<PledgeTableViewData, Never>
+  public let reloadWithData: Signal<(Project, Reward, Bool), Never>
```
justinswart: Did we not just want to keep `PledgeTableViewData` and update the typealias?

Library/ViewModels/PledgeAmountCellViewModel.swift
dusi: Sure, was simply using what we used before

Library/ViewModels/PledgeViewModel.swift
dusi: I dunno, this seemed more readable to me...I think up to 3 elements I'd prefer this, 3+ I'd consolidate it to a typealias ¯\_(ツ)_/¯

Library/ViewModels/PledgeAmountCellViewModel.swift
dusi: I looked at the internal logic a bit more and with the internal logic it seems more appropriate to have it here (it's either using trimmed 2x at call site like we do now or adding it to possible 5 `return` statements). Of course we can add a convenience wrapper but I'm not really sure if that saves us that much since we'd have to also test it I guess.

Leaving as is unless strong opinions on this.

Library/SharedFunctionsTests.swift
```diff
@@ -181,4 +183,59 @@ final class SharedFunctionsTests: TestCase {
...
+    withEnvironment(config: config) {
+      let locations = [
+        Location.template |> Location.lens.country .~ "US",
+        Location.template |> Location.lens.country .~ "CZ",
+        Location.template |> Location.lens.country .~ "CA"
+      ]
+      let shippingRules = locations.map { ShippingRule.template |> ShippingRule.lens.location .~ $0
+      }
+      let shippingRule = defaultShippingRule(fromShippingRules: shippingRules)
+      XCTAssertEqual(shippingRules[1], shippingRule)
```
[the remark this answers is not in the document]
dusi: I've updated the tests (due to the not so ideal `Equatable` conformance) so they're now more specific

PR #694, opened 2019-05-30
🦡 [App Badges] Set badge count on activities tab bar item

justinswart:
# 📲 What

Sets the app's current badge count value on the activities tab bar item.

This is part of a larger body of work to bring badging to the app. **Future** PR's will include:

- resetting badge count upon navigating to the activities tab.
- making a GraphQL request to reset the count on the back-end.
- updating the badge count when a push notification is received while the app is open.

# 🤔 Why

We've allowed for badges in the app for some time now but work has recently been done to support this on the back-end. This brings additional functionality to allow us to display the amount of unread activities in the app and to mark them as read via GraphQL.

# 🛠 How

- On app launch set the value of `AppEnvironment.current.application.applicationIconBadgeNumber` which is `UIApplication.shared.applicationIconBadgeNumber` via our `ApplicationType` protocol on `AppEnvironment`.
- Clamp this value at `99` and display a `+` if this value exceeds that.
- Do not show a badge for `0` or less.

# 👀 See

| 0 unread | 5 unread | 99 unread | > 99 unread |
| --- | --- | --- | --- |
| Screen Shot 2019-05-30 at 2 38 12 PM | Screen Shot 2019-05-30 at 2 36 59 PM | Screen Shot 2019-05-30 at 2 37 28 PM | Screen Shot 2019-05-30 at 2 37 50 PM |

# ✅ Acceptance criteria

- [ ] Badge count reflects current value of `UIApplication.shared.applicationIconBadgeNumber` on app launch.
- [ ] No badge is shown for zero or less.
- [ ] `99+` is shown for an amount greater than `99`.


justinswart:
[New activity indicator inside the app](https://trello.com/c/YQMVZ6mv/1383-new-activity-indicator-inside-the-app)

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -375,3 +391,13 @@ extension TabBarItem: Equatable {
     }
   }
 }
+
+func activitiesBadgeValue() -> String? {
```
dusi: `private`?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -4,7 +4,10 @@ import Prelude
 import ReactiveSwift
 import UIKit
 
+public let maxBadgeValue = 99
```
dusi: Since we're only using this constant from the function bellow does it still make sense to have it here (as oppose to inside the function)?

If you think it's better to have it as a global constant how about we extract this to `Constants.swift` or something and maybe make it an enum?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -375,3 +391,13 @@ extension TabBarItem: Equatable {
     }
   }
 }
+
+func activitiesBadgeValue() -> String? {
+  let count = min(AppEnvironment.current.application.applicationIconBadgeNumber, maxBadgeValue)
+
+  guard count > 0 else { return nil }
+
+  let plusSign = AppEnvironment.current.application.applicationIconBadgeNumber > maxBadgeValue ? "+" : ""
```
dusi: Also looks like we're repeating some code here..how about
```swift
private func activitiesBadgeValue() -> String? {
  let applicationIconBadgeNumber = AppEnvironment.current.application.applicationIconBadgeNumber
  let count = min(applicationIconBadgeNumber, maxBadgeValue)

  guard count > 0 else { return nil }

  let plusSign = applicationIconBadgeNumber > maxBadgeValue ? "+" : ""

  return "\(count)\(plusSign)"
}
```

Library/UIViewController+URLTests.swift
```diff
@@ -43,19 +43,3 @@ final class UIViewControllerURLTests: TestCase {
     }
   }
 }
-
```
dusi: Thanks for cleaning this up!

justinswart:
> Wondering about accessibility...should we have a custom a11y value with more human description than 99+ ... maybe more than 99?

@dusi good point, will address a11y in a separate PR 👍

PR #696, opened 2019-05-31
🃏 [Investment day] Remove all manual Equatable conformance

justinswart:
# 📲 What

Removes all manual conformance to `Equatable` via `func == (lhs:rhs:)`.

# 🤔 Why

Since Swift 4.1 conformance to `Equatable` has been [automatically synthesized](https://github.com/apple/swift-evolution/blob/master/proposals/0185-synthesize-equatable-hashable.md) by the compiler. There is no need for us to manually implement these and, in fact, we get more accurate comparisons because many of our models only compared using `id`.

A future PR will do the same for our manual `Hashable` conformances.

# 🛠 How

- Went through our models one-by-one and removed their implementation of `func == (lhs:rhs:)`.
- Where necessary, applied `Equatable` to each nested member until the parent synthesized successfully.
- Fixed tests that were comparing on `id` alone.

# 📈 Risks

Previously many of our models were compared on `id` alone. By automatically synthesizing conformance to `Equatable` this comparison is now made on _all_ of the model's properties. In most cases this _should_ be fine, however in a situation where comparison on `id` alone was assumed/expected we now run the risk of introducing a bug. I'm not aware of anywhere in the codebase that this is happening but I would want to spend some time looking out and testing for it.

# ✅ Acceptance criteria

- [ ] Dashboard works as before (a number of these tests failed, just want to make sure, it's gtg).


Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -650,7 +650,7 @@ final class AppDelegateViewModelTests: TestCase {
 
     let params = .defaults
       |> DiscoveryParams.lens.sort .~ .newest
-    self.goToDiscovery.assertValues([params])
+    XCTAssertEqual(self.goToDiscovery.values.first??.queryParams, params.queryParams)
```
justinswart: These tests were only comparing on `queryParams` before so I've kept them that way.

KsApi/models/CategoryTests.swift
```diff
@@ -58,7 +58,7 @@ class CategoryTests: XCTestCase {
   }
 
   func testParent() {
-    XCTAssertEqual(Category.illustration.parent, Category.art)
+    XCTAssertEqual(Category.illustration.parent?.id, Category.art.id)
```
justinswart: These tests only checked `id` before, kept them the same.

KsApi/models/ShippingRule.swift
```diff
@@ -17,12 +17,6 @@ extension ShippingRule: Argo.Decodable {
   }
 }
 
-extension ShippingRule: Equatable {}
-public func == (lhs: ShippingRule, rhs: ShippingRule) -> Bool {
-  // todo: change to compare id once that api is deployed
```
justinswart: lol

Library/TestHelpers/AlertError+Equatable.swift
```diff
@@ -1,16 +0,0 @@
-@testable import Library
```
justinswart: This whole file is no longer needed.

Library/ViewModels/DashboardReferrersCellViewModelTests.swift
```diff
@@ -192,25 +192,40 @@ internal final class DashboardReferrersCellViewModelTests: TestCase {
 
     self.vm.inputs.backersButtonTapped()
```
justinswart: This test confuses me, it seems like it wasn't working properly before at all 🤔

Library/ViewModels/DashboardRewardsCellViewModelTests.swift
```diff
@@ -99,25 +101,12 @@ internal final class DashboardRewardsCellViewModelTests: TestCase {
     self.rewardsRowCountry.assertValues([.us])
     self.rewardsRowTotalPledged.assertValues([1_500])
     self.hideSeeAllTiersButton.assertValues([true])
-
-    // switched project
```
justinswart: Some old commented out code.

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -557,10 +557,16 @@ internal final class DashboardViewModelTests: TestCase {
...
+
+      let presentProjectDrawerProjectIds = self.presentProjectsDrawer.values
+        .flatMap { $0.map { $0.project.id } }
+      let projectDataIds = [
         [projectData1, projectData2], [projectData1, projectData2],
         [projectData1, projectData2]
-      ])
+      ]
+      .flatMap { $0.map { $0.project.id } }
+      XCTAssertEqual(presentProjectDrawerProjectIds, projectDataIds)
```
justinswart: Perhaps not the neatest but this keeps the same logic that the tests had before by only comparing on `project.id`.

Library/ViewModels/UpdateDraftViewModelTests.swift
```diff
@@ -556,6 +556,8 @@ final class UpdateDraftViewModelTests: TestCase {
 
   func testPreviewTappedWithEdits() {
     let draft = UpdateDraft.template
+    let updatedDraft = draft
+      |> UpdateDraft.lens.update.title .~ "Hello, world!"
```
justinswart: This updates the model for comparison since its title is changed.

justinswart:
[Remove manual `Equatable` conformance](https://trello.com/c/cpl4OPzr/303-remove-manual-equatable-conformance)

justinswart:
@dusi yip, messages cell wasn't appearing because comparison wasn't being explicitly done on `id` 🤦‍♂ 

https://github.com/kickstarter/ios-oss/pull/696/commits/13c9027e2affc175bb0043c65559f73fef96cfed

I think we might actually have to spent more time and audit every occurrence of `==` and `!=` to be sure it's not doing the wrong comparison 🤔

justinswart:
Unfortunately with the fairly large unknowns imposed by this PR and limited time for us to audit every equality comparison with confidence (plus an interest in not having these forever-blocked PR's) I'm going to close this for now. Thanks for your time!

justinswart:
We're back! As discussed, this was closed for the wrong reasons and is something I'd still like to one day get to.

justinswart:
lol

PR #698, opened 2019-06-05
🦡 [App Badges] Update badge value on push and lifecycle events

justinswart:
# 📲 What

Adds the remaining functionality for badging in the app, in particular setting the badge value on the activities tab bar item when various events occur (see ACs below).

# 🤔 Why

We're now receiving badges via push notification payloads and via the `User` model, this adds the necessary functionality to update the UI to accurately reflect those.

# 🛠 How

- Added inputs to `RootViewModel` so that we can update it when `applicationWillEnterForeground` and when the app is launched from background after having been updated with a badge value and a push notification is tapped.
- Added `ClearUserUnseenActivityMutation` to clear the count on the back-end when the activities tab bar item is selected.
- Added a localized string for the clamped `99+` plus string so that translators can move the `+` around if necessary.

# 👀 See

| Push in app, clear on tap | Count from user refresh | Clear on activities refresh | Clear on logout |
| --- | --- | --- | --- |
|  | no-push-count-from-user-clear-on-tap | clear-on-refresh | clear-on-logout |

# ✅ Acceptance criteria

**How to test:** Download [NWPusher](https://github.com/noodlewerk/NWPusher), run the build on device and grab the device push token from the console. Configure NWPusher with the `com.kickstarter.kickstarter.debug` APNS push certificate. Send yourself a payload like:

```json
{
  "aps": {
    "alert": "Push notification with badge value 80",
    "badge": 80,
    "sound": "default"
  }
}
```

**Note:** Messages do not affect the badge value, only activities.

- [ ] Push notification with badge in payload sets badge value on activities tab bar item clamped at `99`.
- [ ] `99+` is displayed for values greater than `99` and no value is shown for `0` or below.
- [ ] If push is received when app is open the tab bar item's badge value is updated.
- [ ] Select activities tab, background app, receive push with badge count, foreground app. Activities tab should now have a value and you should still be on that tab. Pull to refresh to clear it and refresh activities.
- [ ] If the current user is updated, the badge should be updated with that user's `unseenActivityCount` value (generate an activity to create this value).
- [ ] Navigating to the activities tab and pulling to refresh on activities should both clear the `unseenActivityCount` on the server (generate an activity to create this value).
- [ ] If the user logs out, the badge value is cleared.
- [ ] Tapping the activities tab clears the badge value immediately and fires off an asynchronous 'fire and forget' GraphQL call to reset the activity count on the back-end. Nothing happens if this fails but coming back to the app should show the original count from the refreshed `User` model.
- [ ] BONUS AC: Install an older version of the app and log in. Overwrite that version with this build. You should _still_ be logged in. This is worth testing since we've modified the `User` model.

### DIFF
--- a/Kickstarter-iOS/ViewModels/RootViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModel.swift
@@ -275,17 +275,17 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
       )
       .map { _, index -> RootTabBarItemBadgeValueData in (nil, index) }
 
-    let badgeValueOnUserSessionStarted = self.userSessionStartedProperty.signal
+    let badgeValueOnUserUpdated = self.currentUserUpdatedProperty.signal
       .map { _ in AppEnvironment.current.currentUser?.unseenActivityCount }
 
     let updateBadgeValueFromNotification = selectedIndexAndActivityViewControllerIndex
       .takePairWhen(self.didReceiveBadgeValueSignal)
 
-    let updateBadgeValueOnUserSessionStarted = selectedIndexAndActivityViewControllerIndex
-      .takePairWhen(badgeValueOnUserSessionStarted)
+    let updateBadgeValueOnUserUpdated = selectedIndexAndActivityViewControllerIndex
+      .takePairWhen(badgeValueOnUserUpdated)
 
-    let updateBadgeValueOnUserSessionStartedOrFromNotification = Signal.merge(
-      updateBadgeValueOnUserSessionStarted,
+    let updateBadgeValueOnUserUpdatedOrFromNotification = Signal.merge(
+      updateBadgeValueOnUserUpdated,
       updateBadgeValueFromNotification
     )
     .map(unpack)
@@ -302,7 +302,7 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
 
     self.setBadgeValueAtIndex = Signal.merge(
       updateBadgeValueOnLifecycleEvents,
-      updateBadgeValueOnUserSessionStartedOrFromNotification,
+      updateBadgeValueOnUserUpdatedOrFromNotification,
       clearBadgeValueOnUserSessionEnded,
       clearBadgeValueOnActivitiesTabSelected
     )
--- a/Kickstarter-iOS/ViewModels/RootViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModel.swift
@@ -156,6 +156,9 @@ internal protocol RootViewModelOutputs {
 
   /// Emits data for setting tab bar item styles.
   var tabBarItemsData: Signal<TabBarItemsData, Never> { get }
+
+  /// Emits a User that can be used to replace the current user in the environment.
+  var updateUserInEnvironment: Signal<User, Never> { get }
 }
 
 internal protocol RootViewModelType {
@@ -266,15 +269,6 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
       activityViewControllerIndex
     )
 
-    let clearBadgeValueOnActivitiesTabSelected = selectedIndexAndActivityViewControllerIndex
-      .filter(==)
-      .on(
-        value: { _ in
-          _ = AppEnvironment.current.apiService.clearUserUnseenActivity(input: EmptyInput())
-        }
-      )
-      .map { _, index -> RootTabBarItemBadgeValueData in (nil, index) }
-
     let badgeValueOnUserUpdated = self.currentUserUpdatedProperty.signal
       .map { _ in AppEnvironment.current.currentUser?.unseenActivityCount }
 
@@ -289,9 +283,6 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
       updateBadgeValueFromNotification
     )
     .map(unpack)
-    .filter { selectedIndex, activitiesViewControllerIndex, _ in
-      selectedIndex != activitiesViewControllerIndex
-    }
     .map { _, index, value in
       (activitiesBadgeValue(with: value), index)
     }
@@ -300,12 +291,27 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
       .takePairWhen(self.userSessionEndedProperty.signal)
       .map { index, _ in (activitiesBadgeValue(with: nil), index) }
 
+    let currentBadgeValue = MutableProperty<String?>(nil)
+
+    let clearBadgeValueOnActivitiesTabSelected = selectedIndexAndActivityViewControllerIndex.filter(==)
+      .flatMap { _, index in currentBadgeValue.producer.map { ($0, index) } }
+      .filter { value, _ in value != nil }
+      .map { _, index -> RootTabBarItemBadgeValueData in (nil, index) }
+
     self.setBadgeValueAtIndex = Signal.merge(
       updateBadgeValueOnLifecycleEvents,
       updateBadgeValueOnUserUpdatedOrFromNotification,
       clearBadgeValueOnUserSessionEnded,
       clearBadgeValueOnActivitiesTabSelected
     )
+
+    currentBadgeValue <~ self.setBadgeValueAtIndex.map { $0.0 }
+
+    self.updateUserInEnvironment = clearBadgeValueOnActivitiesTabSelected
+      .switchMap { _ in
+        updatedUserWithClearedActivityCountProducer()
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+      }
 
     let shouldSelectIndex = self.shouldSelectIndexProperty.signal
       .skipNil()
@@ -408,6 +414,7 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
   internal let setViewControllers: Signal<[RootViewControllerData], Never>
   internal let switchDashboardProject: Signal<(Int, Param), Never>
   internal let tabBarItemsData: Signal<TabBarItemsData, Never>
+  internal let updateUserInEnvironment: Signal<User, Never>
 
   internal var inputs: RootViewModelInputs { return self }
   internal var outputs: RootViewModelOutputs { return self }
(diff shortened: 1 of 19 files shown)


Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -244,9 +250,62 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
...
+    let selectedIndexAndActivityViewControllerIndex = Signal.combineLatest(
+      self.selectedIndex,
+      activityViewControllerIndex
+    )
+
+    let clearBadgeValueOnActivitiesTabSelected = selectedIndexAndActivityViewControllerIndex
+      .filter(==)
+      .on(
+        value: { _ in
+          _ = AppEnvironment.current.apiService.clearUserUnseenActivity(input: EmptyInput())
```
justinswart: Open to suggestions on how to test this. Currently it's just injected as a side-effect when the tab bar item is selected. I don't really want to create a void output simply to say 'this emitted' as it causes no update to the UI if it succeeds or fails. One option I thought of is to track it through Koala and assert that the event was tracked?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -244,9 +250,62 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
...
+    }
+    .map { _, index, value in
+      (activitiesBadgeValue(with: value), index)
+    }
+
+    let clearBadgeValueOnUserSessionEnded = activityViewControllerIndex
+      .takePairWhen(self.userSessionEndedProperty.signal)
+      .map { index, _ in (activitiesBadgeValue(with: nil), index) }
+
+    self.setBadgeValueAtIndex = Signal.merge(
```
justinswart: A surprising number of events on which one needs to set this badge value!

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -398,15 +468,20 @@ extension TabBarItem: Equatable {
...
-  guard count > 0 else { return nil }
+  guard clampedBadgeValue > 0 else { return nil }
 
-  let plusSign = applicationIconBadgeNumber > maxBadgeValue ? "+" : ""
+  guard badgeValue > maxBadgeValue else {
+    return "\(clampedBadgeValue)"
+  }
 
-  return "\(count)\(plusSign)"
+  return localizedString(
```
justinswart: Will create a related PR for this string to be translated.

Kickstarter-iOS/Views/Controllers/ActivitiesViewController.swift
```diff
@@ -169,6 +169,12 @@ internal final class ActivitiesViewController: UITableViewController {
       .observeValues { [weak self] project, update in
         self?.goToUpdate(project: project, update: update)
       }
+
+    self.viewModel.outputs.clearBadgeValue
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.parent?.tabBarItem.badgeValue = nil
```
justinswart: This VC is nested so we need to reach out to its parent for the `tabBarItem` 🤷‍♂

Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -28,6 +29,15 @@ public final class RootTabBarViewController: UITabBarController {
     super.viewDidLoad()
     self.delegate = self
 
+    self.applicationWillEnterForegroundObserver = NotificationCenter
+      .default
+      .addObserver(
+        forName: UIApplication.willEnterForegroundNotification,
```
justinswart: I figured observing this notification was better than having to pass this through from `AppDelegate`. It's the same event.

KsApi/models/User.swift
```diff
@@ -16,6 +16,7 @@ public struct User {
   public var showPublicProfile: Bool?
   public var social: Bool?
   public var stats: Stats
+  public var unseenActivityCount: Int?
```
justinswart: Optional so that we don't break the deserialisation of users stored in `UserDefaults`.

justinswart:
[New activity indicator inside the app](https://trello.com/c/YQMVZ6mv/1383-new-activity-indicator-inside-the-app)

Library/ViewModels/ActivitiesViewModel.swift
```diff
@@ -156,6 +159,16 @@ public final class ActivitiesViewModel: ActivitiesViewModelType, ActitiviesViewM
       }
 
     self.activities = Signal.merge(clearedActivitiesOnSessionEnd, updatedActivities)
+    self.clearBadgeValue = Signal.zip(
+      self.refreshProperty.signal,
+      updatedActivities.skip(first: 1)
+    )
+    .on(
+      value: { _ in
+        _ = AppEnvironment.current.apiService.clearUserUnseenActivity(input: EmptyInput())
```
justinswart: Also a candidate for testing 🤷‍♂

Kickstarter-iOS/ViewModels/RootViewModel.swift
eoji: After I clear the count, I update the user and test that it emits again. Would that work here?

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -244,9 +250,62 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
...
+        (activitiesBadgeValue(with: AppEnvironment.current.application.applicationIconBadgeNumber), index)
+      }
+
+    let selectedIndexAndActivityViewControllerIndex = Signal.combineLatest(
+      self.selectedIndex,
+      activityViewControllerIndex
+    )
+
+    let clearBadgeValueOnActivitiesTabSelected = selectedIndexAndActivityViewControllerIndex
+      .filter(==)
```
ifbarrera: I wonder if it might be a good idea to add an additional check here that first determines whether or not there actually *is* a badge on the activities tab before sending off the request to clear it? I think we should try to reduce the number of network requests as much as possible.

Kickstarter-iOS/ViewModels/RootViewModel.swift
justinswart: Ah, yes I think that makes sense! Thanks for the suggestion. For the most part I was just relying on mostly having the badge value from push but actually I should refresh the user to avoid an inconsistent state.

Kickstarter-iOS/ViewModels/RootViewModelTests.swift
```diff
@@ -87,6 +87,108 @@ final class RootViewModelTests: TestCase {
...
+    withEnvironment(application: mockApplication) {
+      self.vm.inputs.viewDidLoad()
+
+      self.setBadgeValueAtIndexValue.assertValues(["99+"])
+      self.setBadgeValueAtIndexIndex.assertValues([1])
+    }
+
+    mockApplication.applicationIconBadgeNumber = 50
+
+    withEnvironment(application: mockApplication) {
```
ifbarrera: Is it necessary to re-set the environment here? 🤔

Kickstarter-iOS/ViewModels/RootViewModel.swift
ifbarrera: Can the mutation return the new `unseenActivityCount`? If so can we just set it on the current user? 🤷‍♀

justinswart: Yeah I'm down with using the response from the mutation 👍

justinswart: Ok, this now uses the value returned from the mutation and updates the current user on `AppEnvironment`. This is a "fire and forget" call so we don't flip the value back if the call fails.

Library/ViewModels/ActivitiesViewModel.swift
justinswart: This is now tested via our `updateUserInEnvironment` outputs.

Kickstarter-iOS/ViewModels/RootViewModel.swift
justinswart: We now only clear this if we know we have a badge value.

Kickstarter-iOS/ViewModels/RootViewModelTests.swift
justinswart: Nope! 👍

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -244,9 +253,65 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
...
+    self.setBadgeValueAtIndex = Signal.merge(
+      updateBadgeValueOnLifecycleEvents,
+      updateBadgeValueOnUserUpdatedOrFromNotification,
+      clearBadgeValueOnUserSessionEnded,
+      clearBadgeValueOnActivitiesTabSelected
+    )
+
+    currentBadgeValue <~ self.setBadgeValueAtIndex.map { $0.0 }
+
+    self.updateUserInEnvironment = clearBadgeValueOnActivitiesTabSelected
```
ifbarrera: Should we add a check here for `AppEnvironment.current.currentUser != nil` before firing off the request?

Library/SharedFunctionsTests.swift
```diff
@@ -179,4 +181,58 @@ final class SharedFunctionsTests: TestCase {
     XCTAssertTrue(ksr_isOSVersionAvailable(12.123))
     XCTAssertTrue(ksr_isOSVersionAvailable(12.9))
   }
+
+  func testupdatedUserWithClearedActivityCountProducer_Success() {
```
ifbarrera: Minor typo here should be `testUpdated` (camel cased)

Kickstarter-iOS/ViewModels/RootViewModel.swift
justinswart: Hmm maybe but this does happen in the function that's called in the `switchMap` - `updatedUserWithClearedActivityCountProducer()` 🤷‍♂

Library/ViewModels/ActivitiesViewModel.swift
```diff
@@ -169,6 +169,7 @@ public final class ActivitiesViewModel: ActivitiesViewModelType, ActitiviesViewM
     .ignoreValues()
 
     self.updateUserInEnvironment = self.clearBadgeValue
+      .filter { _ in AppEnvironment.current.currentUser != nil }
```
ifbarrera: Do we need to add a test for this in `ActivitiesViewModelTests`

justinswart: Woops, yes it should thanks.

PR #699, opened 2019-06-07
Xcode 11 Support

dusi:
# 📲 What

Makes the project compatible with `Xcode 11`.

# 🤔 Why

Staying on the bleeding edge of 🍎 technologies.
- Xcode 11 offers some amazing prototyping tools that we would love to start using soon
- Xcode 11 makes it possible to build for `iOS 13` (and check some of the upcoming features i.e. `Dark Mode`)

# 🛠 How

By replacing Carthage dependencies (where possible) with Swift Package Dependency we're not only able to remove a lot of boiler plate in our project file but as a side effect playgrounds seem more stable now

# 📌 Note

Please note that in order to being able to `build` you'll have to temporarily also build `Prelude` locally with the following addition to `Lens.swift` (this is in order for being able to lens over an optional, which `Xcode 11` was complaining about (mostly when lensing over an optional property on an UI element).

```swift
// Line:173
public func lens<Whole, Part>(_ keyPath: WritableKeyPath<Whole, Part?>) -> Lens<Whole, Part> {
  return Lens<Whole, Part>(
    view: {
      let keyPath = keyPath as! WritableKeyPath<Whole, Part>  // we need to make sure to not use force cast here

      return $0[keyPath: keyPath]
  },
    set: { part, whole in
      var copy = whole
      copy[keyPath: keyPath] = part
      return copy
  }
  )
}

// Line:200
public func .~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, part: Part?) -> ((Whole) -> Whole) {
  return lens(keyPath) .~ part
}

// Line:218
public func %~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, f: @escaping (Part) -> Part)
  -> ((Whole) -> Whole) {

    return lens(keyPath) %~ f
}
```

# ⚠️ Warning

Even though the project builds it doesn't run on `Simulator` just yet, this is due to a problem with embedding `ReactiveSwift` ... see the following console warning:

```
dyld: Library not loaded: @rpath/ReactiveSwift.framework/ReactiveSwift
  Referenced from: /Users/paveldusatko/Library/Developer/CoreSimulator/Devices/82163363-2EDA-4D43-809C-A5195FBC38D4/data/Containers/Bundle/Application/59FFD665-68C5-479B-87AF-C544657B2726/KickDebug.app/Frameworks/ReactiveExtensions.framework/ReactiveExtensions
  Reason: image not found
```


justinswart:
This should be a draft PR 😄 also when I put the force cast in it was just to get it to compile, that likely breaks the actual Lens so need to still better understand how to deal with the optional keypath there 🤔

Does that @rpath error not go away if you embed ReactiveSwift?

justinswart:
Ok, had another look, we don't need the force cast, really we just want to do nothing if the `Part` is `nil`, so we can just delete our modified version of `func lens<Whole, Part>(_ keyPath: WritableKeyPath<Whole, Part?>) -> Lens<Whole, Part>`

and update the `.~` and `%~` operators to be overloaded with these:

```swift
public func .~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, part: Part?) -> ((Whole) -> Whole) {
  guard let part = part else {
    return { $0 }
  }

  return lens(keyPath) .~ part
}

// haven't been able to test this one because I can't get the project to build right now 🤷‍♂ 
public func %~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, f: @escaping (Part?) -> Part?)
  -> ((Whole) -> Whole) {
    return lens(keyPath) %~ f
}
```

PR #700, opened 2019-06-07
💲[Native Checkout] Rewards - Basic

ifbarrera:
# 📲 What

New reward cell implementation for the most basic state: non-backer viewing a live project.

This PR includes:
- displaying included items list
- displaying reward minimum in the project's currency
- displaying a conversion label based on the user's currency preferences
- tapping the card or the "Pledge" button should take you to the pledge screen
- scrolling on long rewards should cause the `Pledge` button to float
- works with dynamic type

This PR *does not* include:
- handling for backed rewards, non-live projects
- proper visual handling for limited rewards
- proper visual handling for unavailable rewards
- proper visual handling for time-based rewards
- reward ordering/sorting
- **not** optimized for voice over yet

# 🤔 Why

So a user can examine rewards to select their reward and pledge to a project.

# 🛠 How

- pre-pended `Deprecated` to several classes
- took most of the logic directly out of the existing `DeprecatedRewardCellViewModel`
- cleaned up some of the tests so it's more obvious what they're doing
- generated new screenshots for `RewardsCollectionViewControllerTests`
- updated strings

# 👀 See

| iPad | iPhone 8 | iPhone X | iPhone 5 |
| --- | --- | --- | --- |
| Simulator Screen Shot - iPad Pro (12 9-inch) - 2019-06-06 at 18 08 01|Simulator Screen Shot - iPhone 8 - 2019-06-07 at 15 43 43|Simulator Screen Shot - iPhone X - 2019-06-06 at 18 13 43|Simulator Screen Shot - iPhone 5s - 2019-06-06 at 17 52 05|

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Login on staging with an admin account. Background the app. Open the app again, and navigate to any project. Tap "Back this project" - You should see all rewards.


Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+  private let pledgeButtonLayoutGuide = UILayoutGuide()
+  private let priceStackView = UIStackView(frame: .zero)
+  private let rewardTitleLabel = UILabel(frame: .zero)
+  private let scrollView = UIScrollView(frame: .zero)
 
   override init(frame: CGRect) {
     super.init(frame: frame)
 
     self.configureViews()
+    self.bindViewModel()
```
justinswart: Note to self to do a PR adding this to our swizzling.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+
+    [
+      self.baseStackView,
+      self.priceStackView,
+      self.includedItemsStackView,
+      self.descriptionStackView
+    ]
+    .forEach { stackView in
+      _ = stackView
+        |> sectionStackViewStyle
```
justinswart: You can replace the `forEach` with `||>`

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+    }
+
+    _ = self.baseStackView
+      |> baseStackViewStyle
+
+    _ = self.priceStackView
+      |> priceStackViewStyle
+
+    [self.includedItemsTitleLabel, self.descriptionTitleLabel].forEach { label in
```
justinswart: This can also be `||>`

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+      _ = label
+        |> baseRewardLabelStyle
+        |> sectionTitleLabelStyle
+    }
+
+    _ = self.includedItemsTitleLabel
+      |> \.text %~ { _ in Strings.project_view_pledge_includes() }
+
+    self.includedItemsStackView.subviews.enumerated().forEach { index, view in
```
justinswart: Can I suggest:

```swift
self.includedItemsStackView.subviews
  .dropFirst()
  .map { $0 as? UILabel }
  .compact()
  ||> baseRewardLabelStyle
  ||> sectionBodyLabelStyle
```

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+    _ = topConstraint
+      |> \.priority .~ .defaultLow
+      |> \.isActive .~ true
+
+    let contentMargins = self.contentView.layoutMarginsGuide
+
+    NSLayoutConstraint.activate([
+      pledgeButton.leftAnchor.constraint(equalTo: contentMargins.leftAnchor),
+      pledgeButton.rightAnchor.constraint(equalTo: contentMargins.rightAnchor),
+      // swiftlint:disable:next line_length
```
justinswart: Don't think we still need the `line_length` disables in some of these cases.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+      separator.heightAnchor.constraint(equalToConstant: 1).isActive = true
+
+      var itemViews: [UIView] = [label]
+
+      if index != items.endIndex - 1 {
+        itemViews.append(separator)
+      }
+
+      return itemViews
+    }.flatMap { $0 }
```
justinswart: Can use `compact()`

justinswart: Actually I think `flatMap` is correct here, didn't realize you're flattening the inner array.

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
```diff
@@ -118,6 +118,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      region = "US"
```
justinswart: What dis 🤔

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+        |> sectionBodyLabelStyle
+        |> \.text .~ item
+
+      let separator = UIView()
+        |> separatorStyle
+      separator.heightAnchor.constraint(equalToConstant: 1).isActive = true
+
+      var itemViews: [UIView] = [label]
+
+      if index != items.endIndex - 1 {
```
justinswart: I tend to prefer `dropLast()` and `dropFirst()` over index comparisons 🤷‍♂ Here's a way one could do that but feel free to ignore if you feel it adds no material benefit or readability gain:

```swift
// we want all of the labels formatted the same
let includedItemViews = items.map { item -> UIView in
  let label = UILabel()
    |> baseRewardLabelStyle
    |> sectionBodyLabelStyle
    |> \.text .~ item

  return label
}

// we want all of them separated except the last one - dropLast()
let separatedItemViews = includedItemViews.dropLast().map { view -> [UIView] in
  let separator = UIView()
    |> separatorStyle
  separator.heightAnchor.constraint(equalToConstant: 1).isActive = true

  return [view, separator]
}
.flatMap { $0 }

let allItemViews = [self.includedItemsTitleLabel]
  + separatedItemViews
  + [includedItemViews.last].compact() // we append the last unseparated label here
```

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+
+    self.pledgeButton.addTarget(self, action: #selector(self.pledgeButtonTapped), for: .touchUpInside)
+
+    let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(self.rewardCardTapped))
+    self.containerView.addGestureRecognizer(tapGestureRecognizer)
   }
 
-  func configureWith(value _: Reward) {}
+  private func setupConstraints() {
+    let baseStackView = self.baseStackView
```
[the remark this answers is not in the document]
ifbarrera: Just to shorten calls further down.

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
ifbarrera: Whoops will revert

ifbarrera:
@justinswart 

1. Can you make a note of which text colors/fonts look off? I can do another pass or we can create a card to address this during a design pass.
2. I'll make a card to add the gradient while scrolling.
3. We already have a [card](https://trello.com/c/ZOb4MD3Y/953-rewards-carousel-should-maintain-its-scroll-position-when-rotating-the-device) to address caching the offset for landscape/portrait rotations.

justinswart:
> Can you make a note of which text colors/fonts look off? I can do another pass or we can create a card to address this during a design pass.

Sure, this is looking at the designs on [InVision](https://kickstarter.invisionapp.com/d/main#/console/16837064/349926395/preview) and comparing:

Screen Shot 2019-06-14 at 10 05 09 AM

Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
```diff
@@ -188,6 +188,10 @@ extension ProjectNavigatorViewController: ProjectPamphletViewControllerDelegate
 
     let navigationController = UINavigationController(rootViewController: rewardsViewController)
 
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      navigationController.modalPresentationStyle = .formSheet
```
dusi: Looks like the rest of our codebase doesn't really bother checking for this...I wonder if we should just omit this check and always set it to `formSheet`...it will only apply on iPads based on the docs:
```
The presentation style determines how a modally presented view controller is displayed on the screen. On iPhone and iPod touch devices, modal view controllers are always presented full-screen, but on iPad devices there are several different presentation options. For a list of possible presentation styles, and their compatibility with the available transition styles, see the “UIModalPresentationStyle” constant descriptions.
```

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -25,7 +25,7 @@ final class RewardsCollectionViewController: UICollectionViewController {
     UICollectionViewFlowLayout()
       |> \.minimumLineSpacing .~ Styles.grid(3)
       |> \.minimumInteritemSpacing .~ 0
-      |> \.sectionInset .~ .init(topBottom: Styles.grid(6))
+      |> \.sectionInset .~ .zero
```
dusi: Do we need to set this explicitly? This is what docs say about `sectionInset`: `The default edge insets are all set to 0.`

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -0,0 +1,215 @@
...
+  public init() {
+    let projectAndRewardOrBacking: Signal<(Project, Either<Reward, Backing>), Never> =
+      self.projectAndRewardOrBackingProperty.signal.skipNil()
+
+    let project: Signal<Project, Never> = projectAndRewardOrBacking.map(first)
+
+    let reward: Signal<Reward, Never> = projectAndRewardOrBacking
+      .map { project, rewardOrBacking -> Reward in
+        rewardOrBacking.left
+          ?? rewardOrBacking.right?.reward
```
dusi: Am I the only one really confused about what's going on in here?

Library/ViewModels/RewardCellViewModel.swift
```diff
@@ -0,0 +1,215 @@
...
+        }
+      }
+      .map(Strings.About_reward_amount(reward_amount:))
+
+    self.rewardMinimumLabelText = projectAndRewardOrBacking
+      .map(formattedAmountForRewardOrBacking(project:rewardOrBacking:))
+
+    self.descriptionLabelText = reward
+      .map { $0.isNoReward ?
+        Strings.Pledge_any_amount_to_help_bring_this_project_to_life() : $0.description
```
dusi: Looks like it would fit on 1 line

dusi:
**Top clipping**
Was this really the final consensus (as oppose to clipping the inside of the rounded container)

Screen Shot 2019-06-14 at 3 21 22 PM

**Bottom corners**
Also should these corners be rounded?

Screen Shot 2019-06-14 at 3 21 49 PM


**Bottom corners****Bottom corners**
Also should these corners be rounded?lso should these corners be rounded?

Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
```diff
@@ -429,8 +441,10 @@
 "Tap_the_heart_on_each_project_page_to_save_it" = "Tap the heart on each project page to save it. We’ll remind you 48 hours before the project ends.";
 "Tap_the_star_on_a_project_to_get_notified" = "Tap the star icon on a project to get notified 48 hours before it ends.";
 "Tap_the_star_on_each_project_page_to_save_it" = "Tap the star on each project page to save it. We’ll remind you 48 hours before the project ends.";
+"Ten_is_just_the_beginning" = "Ten is just the beginning.";
```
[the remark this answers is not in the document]
justinswart: I didn't merge the KSR10 stuff to master so that I didn't have to remove it again 😄  I'm working on improving our strings script so that we only generate strings for they keys that we want to.

Kickstarter-iOS/Views/Cells/RewardCell.swift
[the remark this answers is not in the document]
ifbarrera: I'll do another pass of this to try to make it more readable 👍

Kickstarter-iOS/DataSources/RewardsCollectionViewDataSourceTests.swift
```diff
@@ -10,11 +11,15 @@ final class RewardsCollectionViewDataSourceTests: XCTestCase {
   )
 
   func testLoadRewards() {
-    let rewards = [Reward.template, Reward.template]
+    let project = Project.cosmicSurgery
+    let rewardsData = project.rewards.map { (project, Either<Reward, Backing>.left($0)) }
 
-    self.dataSource.load(rewards: rewards)
+    self.dataSource.load(rewardsData)
```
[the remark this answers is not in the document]
ifbarrera: This PR doesn't include tests and updates to support displaying `Backing`s

Kickstarter-iOS/Locales/de.lproj/Localizable.strings
```diff
@@ -70,7 +71,7 @@
 "Closes_project" = "Schließt das Projekt.";
 "Collapses_subcategories" = "Reduziert die Anzeige der Unterkategorien.";
 "Collections" = "Sammlungen";
-"Comment_reply_digest" = "Comment reply digest";
+"Comment_reply_digest" = "Tägliche Zusammenfassung Kommentare";
```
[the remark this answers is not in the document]
ifbarrera: @dusi right now we don't have a way to specify which strings to download with the `make strings` script. When you run the script, it pulls down all changes to localized strings.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+      return itemViews
+    }.flatMap { $0 }
+
+    let allItemViews = [self.includedItemsTitleLabel] + includedItemViews
+
+    _ = (allItemViews, self.includedItemsStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+  }
+
+  internal func configureWith(value: (Project, Either<Reward, Backing>)) {
```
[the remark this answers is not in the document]
ifbarrera: I thought we'd talked about naming the elements if there are more than 2 or 3 🤔 

It doesn't seem super useful here since there are only two.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
+  @objc func rewardCardTapped() {
+    self.viewModel.inputs.rewardCardTapped()
+  }
+}
+
+// MARK: - Styles
+
+private let baseRewardLabelStyle: LabelStyle = { label in
+  label
+    |> \.numberOfLines .~ 0
```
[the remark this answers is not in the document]
ifbarrera: Are you suggesting doing: `|> \.numberOfLines .~ Int(0)` here? The compiler isn't currently showing a type-check warning here so it doesn't seem necessary. 🤔

Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
ifbarrera: Sure 👍

ifbarrera:
> **Top clipping**
> Was this really the final consensus (as oppose to clipping the inside of the rounded container)
> 
> Screen Shot 2019-06-14 at 3 21 22 PM
> 
> **Bottom corners**
> Also should these corners be rounded?
> 
> Screen Shot 2019-06-14 at 3 21 49 PM
> 
> **Bottom corners****Bottom corners**
> Also should these corners be rounded?lso should these corners be rounded?

@dusi yeah this is how it's intended to work. Please see the design prototype:

https://drive.google.com/file/d/19XSbiKgSSvkehMEXweJ3mVKIPI2PDbRb/view

The intention is that the page control will sit at the bottom and the cards should scroll below the page control and the home indicator safe area.

ifbarrera:
> Couple comments.
> 
> I've also noticed that some of the localized strings cause changes to an existing functionality in the deprecated controllers. We've previously tried to keep the feature branch free of such changes (imagine we want to A/B test the new checkout - we'd still need to have access to the deprecated checkout). But now that it's been modified, we haven't really technically QA'd the updated deprecated one.

There was one Japanese string that was updated in the deprecated checkout experience. I regenerated the snapshots and visually QA'd the diff between the old and the new snapshots - given I can't read Japanese, all I could verify is that the text was wrapping correctly and that there wasn't any overlap of copy. I think given that was the only thing that changed it's relatively low risk.

Regardless, I think we should still plan on regression-testing the deprecated checkout before shipping the new checkout. Some of the feature flag tools I've worked on will facilitate that, and it'll give us confidence that we haven't introduced any regressions to the legacy experience.

Kickstarter-iOS/Views/Cells/RewardCell.swift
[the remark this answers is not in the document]
dusi: I'm sorry, I was referring to the other rows (basically anything that's not a primitive like `lineBreakMode`, `backgroundColor`, etc.)

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -2,32 +2,352 @@ import Foundation
...
 import Library
 import Prelude
+import ReactiveSwift
+
+protocol RewardCellDelegate: AnyObject {
+  func rewardCellDidTapPledgeButton(_ rewardCell: RewardCell, rewardId: Int)
+}
 
 final class RewardCell: UICollectionViewCell, ValueCell {
+  weak var delegate: RewardCellDelegate?
```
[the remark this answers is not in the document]
ifbarrera: Moved the `//MARK` to the top 👍

Kickstarter-iOS/Locales/de.lproj/Localizable.strings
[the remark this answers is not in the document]
dusi: I know 😞 

It just feels weird that our limited tooling allows for unrelated changes to make it to this PR. Would it be hard to remove using Xcode's DIFF comparison editor or any other GUI client for git? Unfortunately it's a manual process but it could be simplified if we only chose to commit strings we're interested in (given there is far less of those than our script has changed)

Screen Shot 2019-06-17 at 9 28 35 AM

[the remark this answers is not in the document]
ifbarrera: I imagine the intention with this workflow was that all user interfaces should be flexible enough to support string updates as they come. If we're going to stop updating strings we need to come up with a new process for translators to request that we update specific strings in specific parts of the app as needed. My concern with "cherry-picking" strings (besides the fact that it's super manual) is that we'd be ignoring string changes that have intentionally been added by the international team. I think this is a much larger process topic given that the way it's worked in the past is we always update all strings so there isn't any communication between the translation team and us about what needs to be updated because they just update the string on their end and it gets updated on our end the next time someone runs `make strings`.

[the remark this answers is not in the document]
ifbarrera: I reverted all the other strings but I do think we should have a wider discussion about this before committing to changing our workflow.

Library/ViewModels/RewardCellViewModel.swift
justinswart: I think this is how it was written before in what is now `DeprecatedRewardCellViewModel.swift` - https://github.com/kickstarter/ios-oss/pull/700/files#diff-be538dae7b71e33a0e267c88f43d2024R52. Agreed the coalescing makes is a little hard to read but I think @ifbarrera didn't want to make any changes to how the code works as far as possible.

Kickstarter-iOS/Views/Cells/RewardCell.swift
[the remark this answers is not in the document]
dusi: Cool beans..this is unfortunately CPU dependent so it might be fine on some machines but not as performant on other machines...I thought we have agreed to this earlier when working on the checkout but maybe it got missed in conversation ¯\_(ツ)_/¯

ifbarrera:
[User Can See All Rewards in the Basic State](https://trello.com/c/rA9MQuIz/1369-user-can-see-all-rewards-in-the-basic-state)

PR #701, opened 2019-06-07
🛠 Xcode 11 Support

dusi:
# 📲 What

Makes the project build with `Xcode 11`.

# 🤔 Why

Staying on the bleeding edge of 🍎 technologies.
- `Xcode 11` offers some amazing prototyping tools that we would love to start using soon (unfortunately Xcode Preview requires macOS Catalina)
- Xcode 11 makes it possible to build for `iOS 13` (and check some of the upcoming features i.e. `Dark Mode`)

# 🛠 How

By replacing Carthage dependencies (where possible) with Swift Package Dependency we're not only able to remove a lot of boiler plate in our project file but as a side effect playgrounds seem more stable now

# 📌 Note

Please note that in order to being able to `build` you'll have to temporarily also build `Prelude` locally with the following addition to `Lens.swift` (this is in order for being able to lens over an optional, which `Xcode 11` was complaining about (mostly when lensing over an optional property on an UI element).

```swift
// Line:173
public func lens<Whole, Part>(_ keyPath: WritableKeyPath<Whole, Part?>) -> Lens<Whole, Part> {
  return Lens<Whole, Part>(
    view: {
      let keyPath = keyPath as! WritableKeyPath<Whole, Part>  // we need to make sure to not use force cast here

      return $0[keyPath: keyPath]
  },
    set: { part, whole in
      var copy = whole
      copy[keyPath: keyPath] = part
      return copy
  }
  )
}

// Line:200
public func .~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, part: Part?) -> ((Whole) -> Whole) {
  return lens(keyPath) .~ part
}

// Line:218
public func %~ <Whole, Part> (keyPath: WritableKeyPath<Whole, Part?>, f: @escaping (Part) -> Part)
  -> ((Whole) -> Whole) {

    return lens(keyPath) %~ f
}
```

# ⚠️ Warning

Even though the project builds it doesn't run on `Simulator` just yet, this is due to a problem with embedding `ReactiveSwift` ... see the following console warning:

```
dyld: Library not loaded: @rpath/ReactiveSwift.framework/ReactiveSwift
  Referenced from: /Users/paveldusatko/Library/Developer/CoreSimulator/Devices/82163363-2EDA-4D43-809C-A5195FBC38D4/data/Containers/Bundle/Application/59FFD665-68C5-479B-87AF-C544657B2726/KickDebug.app/Frameworks/ReactiveExtensions.framework/ReactiveExtensions
  Reason: image not found
```


dusi:
Will move this to private for now..

PR #702, opened 2019-06-10
💲[Native Checkout]  Project Pledge States UI

cdolm92:
# 📲 What

The four of the five core states a project can be in:
- `Back this project`
- `Manage`
- `View your pledge`
- `View rewards`

# 🤔 Why

Depending on the project pledge state users can back, manage, view pledge or reward with this CTA.

# 🛠 How

The `PledgeCTAContainerView` displays the CTA depending on state. The state is determined in the `PledgeCTAContainerViewViewModel` where `func projectStateButton(backer: User, project: Project)` 
returns a `PledgeStateCTAType`.

# 👀 See

| Live Project 🐛 | Non-Live Project 🦋 |
| --- | --- |
|Simulator Screen Shot - iPhone Xs - 2019-06-17 at 14 07 59| Simulator Screen Shot - iPhone Xs - 2019-06-17 at 14 08 05| 
| --- | --- |
|Simulator Screen Shot - iPhone Xs - 2019-06-17 at 14 08 16|Simulator Screen Shot - iPhone Xs - 2019-06-17 at 14 08 36|

# ✅ Acceptance criteria

**Test Non-Live/Backed Project**
- [x] Navigate to `Profile` tab
- [x] Scroll and tap on a Non-Live project under the `backed` screen 
- [x] View Your Pledge CTA should show in the project page

**Test Non-Live/Not Backed Project**
- [x] Navigate to `Search` tab, search for a non-live/unbacked project (ex. Cat cafe)
- [x] Scroll and tap on a Non-Live/Unbacked project
- [x] View Reward CTA should show in project page
 * Repeat steps w/ `Unsuccessful` projects too.

**Test Live/Not Backed Project**
- [x] Tap on project in `Explore` screen
- [x] Scroll and tap on a Live/Unbacked project
- [x] Back this project CTA should show in project page

**Test Live/Backed Project**
- [x] Navigate to `Profile` tab
- [x] Scroll and tap on a Live project under the `backed` screen 
- [x] Manage CTA should show in the project page

# ⏰ TODO

- [x] Fix snapshots


Kickstarter-iOS/Views/Cells/ProjectPamphletMainCell.swift
```diff
@@ -49,8 +49,6 @@ internal final class ProjectPamphletMainCell: UITableViewCell, ValueCell {
   @IBOutlet fileprivate weak var readMoreButton: UIButton!
   @IBOutlet fileprivate weak var stateLabel: UILabel!
   @IBOutlet fileprivate weak var statsStackView: UIStackView!
-  @IBOutlet fileprivate weak var youreABackerContainerView: UIView!
```
ifbarrera: How come we removed this? 🤔

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
```diff
@@ -0,0 +1,88 @@
+import Library
+import KsApi
+import Prelude
+import UIKit
+
+class ProjectStatesContainerView: UIView {
+  fileprivate let vm: ProjectStatesContainerViewViewModelType = ProjectStatesContainerViewViewModel()
+  // MARK: - Properties
+
+  private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
```
ifbarrera: If we're not doing any additional configuration, I'd say we should prefer using straight `let`s:

`private let rootStackView = UIStackView(frame: .zero)`

What do others think about this? cc @justinswart @dusi

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
```diff
@@ -0,0 +1,88 @@
...
+      |> ksr_addArrangedSubviewsToStackView()
+
+    self.button.rac.title = self.vm.outputs.buttonTitleText
+    self.button.rac.backgroundColor = self.vm.outputs.buttonBackgroundColor
+    self.labelStackView.rac.hidden = self.vm.outputs.stackViewIsHidden
+    self.label.rac.text = self.vm.outputs.rewardTitle
+  }
+
+  func configureWith(project: Project, user: User) {
+    _ = self.button
```
ifbarrera: Should all of these styling things go under the `bindStyles` function instead of `configureWith`?

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
```diff
@@ -0,0 +1,88 @@
+import Library
+import KsApi
+import Prelude
+import UIKit
+
+class ProjectStatesContainerView: UIView {
```
ifbarrera: `final`?

Kickstarter-iOS/Views/Storyboards/ProjectPamphlet.storyboard
```diff
@@ -428,24 +428,6 @@
                                                 </view>
                                             </subviews>
                                         </stackView>
-                                        <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="wDa-3z-Se6">
```
ifbarrera: Just wondering why we removed this view here too

justinswart:
Hey @cdolm92 I know this is WIP but looking at these changes I think things may have gotten lost in your merge. I think it might be easier to roll back to before the merge and instead reapply your changes on top of `master`. Happy to help with that!

Kickstarter-iOS/Views/Cells/ProjectPamphletMainCell.swift
cdolm92: Product decision - It's not in the design and it would be redundant to have `You're a backer` label on the project cell and manage CTA.

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
cdolm92: I did it this way because it was previously suggested I should. This is a convention that the team should figure out because it's getting confusing. cc @justinswart @dusi

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,93 @@
...
+  default:
+    return PledgeStateCTAType.viewRewards
+  }
+}
+
+private func formattedAmount(for backing: Backing) -> String {
+  let amount = backing.amount - Double(backing.shippingAmount ?? 0)
+  let backingAmount = floor(amount) == amount
+    ? String(Int(amount))
+    : String(format: "%.2f", backing.amount)
```
Scollaco: Web and Android rounds the amount (`Int(round(backing.amount))`) instead of showing decimals, so to keep consistent with the other platforms, I would do the same.

cdolm92: Thanks for bringing this up! Checked in w/ @dannyalright and we're going to show the exact amount that the user pledged

dannyalright: yeah. I feel like this is a case where the other platforms do it 'wrong'. Breaking platform consistency is worth it.

I'll follow up with the other platforms to see if we can get them to show the true amount.

dannyalright: Just spoke to Izzy for context. Could we please make that amount reflect the **reward minimum** amount? So the number should match what the user sees on the reward tile in the next screen. 

Reward minimums never have decimals, if it helps.

Library/Strings.swift
```diff
@@ -727,6 +727,28 @@ contributeurs"
...
+backers"
+   - **de**: "%{backers_count}
+Unterstützer"
+   - **es**: "%{backers_count}
+patrocinadores"
+   - **fr**: "%{backers_count}
+contributeurs"
+   - **ja**: "%{backers_count} バッカー"
+  */
+  public static func Backers_count_separator_backers_other(backers_count: String) -> String {
```
dusi: Looks like the size of this PR is mostly caused by the generated strings...was it intentional to include these as well?
Let's say we now use one of these we won't even show their default (English) representation but the key instead.

Should we pair on this?

Kickstarter-iOS/Views/Cells/ProjectPamphletMainCell.swift
```diff
@@ -231,16 +229,6 @@ internal final class ProjectPamphletMainCell: UITableViewCell, ValueCell {
     _ = self.statsStackView
       |> UIStackView.lens.isAccessibilityElement .~ true
       |> UIStackView.lens.backgroundColor .~ .white
-
-    _ = self.youreABackerContainerView
```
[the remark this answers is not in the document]
dusi: The reason I'm asking is because I've noticed @ifbarrera 's earlier comment and it would all be cool in case we're using the `CTA` button but since that is feature flagged it is potentially possible we will still want to show this screen with the old vertical rewards on it (maybe as an A/B test).

So in case of CTA removing this button makes sense because the CTA button itself is very descriptive and replaces this functionality, but

in case of the old design, users won't have any way of telling the state from the UI

Just wanted to double check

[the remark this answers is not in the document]
dannyalright: Removing the `You're a backer` from the screen was not intentional. It will be removed from the once we redesign the rest of the Project screen. The confusion is my fault; the mockups were just using a repeated project page to give context. They did not represent state but must have been bundled in when engineering tasks were made.

I was fine with letting it drop-off but you raise a good point RE still keeping vertical rewards around in some circumstances. Let's re-add the `You're a backer` for now then, please @cdolm92.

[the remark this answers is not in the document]
cdolm92: cool 👍🏾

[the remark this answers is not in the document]
ifbarrera: @dannyalright @cdolm92 I think we could actually add this back in a way that would allow us to conditionally show/hide it depending on whether the `featureNativeCheckout` feature flag is enabled 🤔 maybe that's the happy medium?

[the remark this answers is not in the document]
dannyalright: fine with me! Good idea

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
dusi: I also thought we agreed on `lazy var`s for the time being...most importantly because as `lazy vars` support both single and multi line assignments compared to `let`s that don't.

For example

```swift
lazy var tableView: UITableView = { UITableView(frame: .zero, style: .plain) }
```

for single line property

```swift
lazy var tableView: UITableView = { 
  UITableView(frame: .zero, style: .plain)
    |> \.dataSource .~ self
}
```

for multilane property (the above is not supported by Swift - as far as I know)

in addition to that `lazy var`s are loaded lazily so if there's a view component we need to add later in the lifecycle (as oppose to init) this will be optimized solution.

Library/PledgeStateCTAType.swift
```diff
@@ -0,0 +1,41 @@
+import UIKit
+
+public enum PledgeStateCTAType {
+  case pledge
+  case manage
+  case viewBacking
+  case viewRewards
+
+  public var buttonTitle: String {
```
dusi: Could we order alphabetically (so put it after `buttonBackgroundColor`)?

Library/Styles/Colors.swift
```diff
@@ -179,6 +179,11 @@ extension UIColor {
     return .hex(0x1A8CFF)
   }
 
+  // 0x2b60ff
+  public static var ksr_blue: UIColor {
+    return .hex(0x2b60ff)
```
dusi: Just a question...was this generated by our `Colors.script`? Looks like the casing is different so just wanted to double check..we would lose these changes next time someone runs the script.

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+public protocol PledgeCTAContainerViewViewModelType {
+  var inputs: PledgeCTAContainerViewViewModelInputs { get }
+  var outputs: PledgeCTAContainerViewViewModelOutputs { get }
+}
+
+public final class PledgeCTAContainerViewViewModel: PledgeCTAContainerViewViewModelType,
+  PledgeCTAContainerViewViewModelInputs, PledgeCTAContainerViewViewModelOutputs {
+  public init() {
+    let projectAndUser = self.projectAndUserProperty.signal.skipNil()
+    let project = projectAndUser.map { $0.0 }
```
dusi: We can be point free here: 
```suggestion
    let project = projectAndUser.map(first)
```

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    let pledgeState = project
+      .map { pledgeStateButton(project: $0) }
```
dusi: and here as well: 
```suggestion
      .map(pledgeStateButton(project:))
```

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+    }
+    let backing = backingEvent.values()
+    let projectAndBacking = Signal.combineLatest(project, backing)
+
+    self.buttonTitleText = pledgeState.map { $0.buttonTitle }
+    self.buttonBackgroundColor = pledgeState.map { $0.buttonBackgroundColor }
+    self.stackViewIsHidden = pledgeState.map { $0.stackViewIsHidden }
+
+    self.rewardTitle = projectAndBacking
+      .map { (arg) -> String in
```
dusi: Do we need the ambiguous `arg` here? Looks like the following works on my machine..

```swift
  .map { (project, backing) -> String in
```

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+  public var inputs: PledgeCTAContainerViewViewModelInputs { return self }
+  public var outputs: PledgeCTAContainerViewViewModelOutputs { return self }
+
+  public let buttonBackgroundColor: Signal<UIColor, Never>
+  public let buttonTitleText: Signal<String, Never>
+  public let rewardTitle: Signal<String, Never>
+  public let stackViewIsHidden: Signal<Bool, Never>
+}
+
+private func pledgeStateButton(project: Project) -> PledgeStateCTAType {
```
dusi: Can we separate this section?

```suggestion

// MARK: - Functions

private func pledgeStateButton(project: Project) -> PledgeStateCTAType {
```

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+  guard let projectIsBacked = project.personalization.isBacking
+  else { return PledgeStateCTAType.viewRewards }
```
dusi: Looks like this would fit on 1 line

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+
+  switch project.state {
+  case .live:
+    return projectIsBacked ? PledgeStateCTAType.manage : PledgeStateCTAType.pledge
```
dusi: Here we can probably loose `PledgeStateCTAType` type for readability. This will most likely not cause any compiler slowdown since we're not inferring this using our pipes so we should be good.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -64,12 +67,21 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
+      .skipNil()
+
     self.goToRewards = freshProjectAndRefTag
       .takeWhen(self.backThisProjectTappedProperty.signal)
       .map { project, refTag in
         (project, refTag)
       }
 
+    let project = freshProjectAndRefTag
+      .map { project, _ in project }
```
dusi: We could again be point free: 
```suggestion
      .map(first)
```

Kickstarter-iOS/Views/ProjectStatesContainerView.swift
ifbarrera: Not all properties require access to `self` to be configured. Furthermore, I'd argue that we should always be explicitly setting things like `delegate`s and `dataSource`s in `viewDidLoad` so that we have certainty on when things those are getting assigned - otherwise, it becomes harder to reason about exactly when (for example) the table view's `dataSource` is getting set, since it's lazily evaluated 🤷‍♀ 

I guess my feeling is that we should only be using `lazy vars` when we actually need to defer a property's initialization for some reason, for example: it's only needed in certain cases, or it's very expensive to initialize (date formatters etc). Most properties we're looking to create are needed almost right away (when they're added to the view hierarchy, usually on `viewDidLoad`) which I *think* makes their deferred initialization mostly pointless. My feeling is that *always* using `lazy vars` to declare properties even when their "lazy" functionality isn't really needed is mostly a stylistic preference. If we want to use that approach just so everything's consistent, then I guess that's fine.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift
```diff
@@ -84,6 +86,56 @@ internal final class ProjectPamphletViewControllerTests: TestCase {
...
+      |> Project.lens.state .~ .successful
+
+    [Device.phone4inch, Device.phone5_5inch, Device.phone5_8inch].forEach { device in
+      let language = Language.en
+      withEnvironment(config: config, currentUser: currentUser, language: language) {
+        let vc = ProjectPamphletViewController.configuredWith(
+          projectOrParam: .left(backedProject), refTag: nil
+        )
+
+        let (parent, _) = traitControllers(device: device, orientation: .landscape, child: vc)
```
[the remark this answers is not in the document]
justinswart: Look if you ask me I would just lock the app in portrait so I'm not sure I'm the right person to give an opinion 🤷‍♂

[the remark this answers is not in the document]
cdolm92: I decided to do landscape testing instead of portrait here for all cases because it generated the least amount of snapshots.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -160,40 +171,6 @@ public final class ProjectPamphletViewController: UIViewController {
...
-
-  private func configureViews() {
-    _ = (self.backThisProjectButton, self.backThisProjectContainerView)
-      |> ksr_addSubviewToParent()
-
-    _ = (self.backThisProjectContainerView, self.view)
-      |> ksr_addSubviewToParent()
-
-    self.backThisProjectButton.addTarget(
-      self, action: #selector(ProjectPamphletViewController.backThisProjectTapped), for: .touchUpInside
```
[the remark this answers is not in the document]
cdolm92: I renamed it on this file but I might remove it since the button is in the `PledgeCTAContainerView`

dannyalright:
> ### `MultiLineButton`
> I wanted to verify with @ifbarrera & @dannyalright that the button is supposed to not change its height and use multiple lines for longer texts (i.e. with Dynamic Type accessibility font)? I thought this used to work previously but now I couldn't make the button expand vertically). Instead it's been trimming the text in the middle

> Screen Shot 2019-06-20 at 9 19 45 PM

I'm concerned about the height of this button taking up too much space if we expand it vertically as its container is fixed to the button. Thoughts? Otherwise I'm in favour of expansion.

> ### Spacing
> @dannyalright is this spacing enough? Seems like the text is really close to the button

> Screen Shot 2019-06-20 at 9 30 26 PM

@cdolm92 could you please add another unit (6pt) of spacing? Sorry for not specifying this in the design work.

> ### Dynamic Type
> I've noticed that scaling the font up collapses the information we're showing to the user into a very limited and unusable state (see bellow).

> Screen Shot 2019-06-20 at 9 30 34 PM

> Wonder if we should design this button with more accessibility in mind @dannyalright maybe similar to what we do with some of the cells on the rewards pledge (see bellow).

> Screen Shot 2019-06-20 at 10 31 17 PM

> This is an example of the same cell under normal size (items are aligned horizontally)

> Screen Shot 2019-06-20 at 10 31 25 PM

while for accessibility sizes we align items vertically. This gives the user enough information at all times.

Agreed. @cdolm92 could you please make the button wrap?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+// MARK: - Styles
+
+private let projectStateButtonStyle: ButtonStyle = { (button: UIButton) in
+  button
+    |> roundedStyle(cornerRadius: 12)
+    |> UIButton.lens.titleColor(for: .normal) .~ .white
+    |> UIButton.lens.titleLabel.font .~ .ksr_headline(size: 15)
```
ifbarrera: You might also need to set `\.lineBreakMode .~ NSLineBreakMode.byWordWrapping` to get the text to wrap.

ifbarrera: I wonder if it makes sense to refactor `checkoutGreenButtonStyle` in `CheckoutStyles.swift` into `checkoutButtonStyle` by removing the `greenButtonStyle` to make it a more generic button style, then add the color separately? Just thinking since the style of this button is identical to `checkoutGreenButtonStyle` with the exception of the color 🤔

dusi:
@dannyalright just for a reference we've used this `MultiLineButton` to also expand some other buttons which are part of the checkout workflow in order for the user to see the whole thing. We've made this decision while you were away.

Here's an example of how it looks (hopefully this wasn't a huge misstep)
Screen Shot 2019-06-21 at 2 36 16 PM

cdolm92:
> > ### `MultiLineButton`
> > I wanted to verify with @ifbarrera & @dannyalright that the button is supposed to not change its height and use multiple lines for longer texts (i.e. with Dynamic Type accessibility font)? I thought this used to work previously but now I couldn't make the button expand vertically). Instead it's been trimming the text in the middle
> 
> > Screen Shot 2019-06-20 at 9 19 45 PM
> 
> I'm concerned about the height of this button taking up too much space if we expand it vertically as its container is fixed to the button. Thoughts? Otherwise I'm in favour of expansion.
> 
> > ### Spacing
> > @dannyalright is this spacing enough? Seems like the text is really close to the button
> 
> > Screen Shot 2019-06-20 at 9 30 26 PM
> 
> @cdolm92 could you please add another unit (6pt) of spacing? Sorry for not specifying this in the design work.
> 
> > ### Dynamic Type
> > I've noticed that scaling the font up collapses the information we're showing to the user into a very limited and unusable state (see bellow).
> 
> > Screen Shot 2019-06-20 at 9 30 34 PM
> 
> > Wonder if we should design this button with more accessibility in mind @dannyalright maybe similar to what we do with some of the cells on the rewards pledge (see bellow).
> 
> > Screen Shot 2019-06-20 at 10 31 17 PM
> 
> > This is an example of the same cell under normal size (items are aligned horizontally)
> 
> > Screen Shot 2019-06-20 at 10 31 25 PM
> 
> while for accessibility sizes we align items vertically. This gives the user enough information at all times.
> 
> Agreed. @cdolm92 could you please make the button wrap?

The spacing is on iPhone SE and so I think it's fine as it is.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -24,12 +24,12 @@ public final class ProjectPamphletViewController: UIViewController {
...
-  private let backThisProjectContainerViewMargins = Styles.grid(3)
-  private let backThisProjectContainerView: UIView = {
-    UIView(frame: .zero) |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  private let pledgeCTAContainerViewMargins = Styles.grid(3)
+  private let pledgeCTAContainerView: PledgeCTAContainerView = {
+    PledgeCTAContainerView(frame: .zero) |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  private let backThisProjectButton: UIButton = {
+  private let pledgeCTAButton: UIButton = {
```
ifbarrera: Hey @cdolm92 I know this is already merged but did we forget to remove this button? I don't see it being used in this file. Same goes for the `backThisProjectTapped` function in this file.

cdolm92: Thanks for pointing this out! I will make a card to remove this.

PR #703, opened 2019-06-11
Update Facebook SDK

justinswart:
# 📲 What

Updates our Facebook SDK to [5.0.2](https://github.com/facebook/facebook-objc-sdk/releases/tag/v5.0.2). At the same time this PR moves the management of this dependency to Carthage and thus removes the SDK files from our repo unfortunately resulting in a large diff.

**Note:** Ignore/filter the deleted .h files in the diff to make reviewing easier.

# 🤔 Why

Our Facebook SDK was quite outdated and we've had reports of failed logins as well as friends lists not appearing in the app. Although it was also since discovered that our app was pending app review by Facebook it was also time to update this SDK and remove it from our repo.

# 🛠 How

- Moved dependency to `Cartfile`.
- Renamed deprecated types (Facebook dropped the `FBSDK` prefix in front of their types for Swift codebases).
- Updated our `FacebookAppDelegateProtocol` for tests and conformed the new `ApplicationDelegate` (previously `FBSDKApplicationDelegate`) to this.

# ✅ Acceptance criteria

**Note:** Both of these are dependent on Facebook first completing their app review.

- [ ] Facebook login works
- [ ] Friends list loads correctly in the app

# ⏰ TODO

- [ ] Confirm what to do about `FacebookAutoLogAppEventsEnabled` and `FacebookAdvertiserIDCollectionEnabled` console warnings.


Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -443,7 +443,7 @@ final class AppDelegateViewModelTests: TestCase {
     withEnvironment(apiService: MockService(fetchUserSelfError: error), currentUser: .template) {
       self.forceLogout.assertValueCount(0)
 
-      vm.inputs.applicationDidFinishLaunching(
+      self.vm.inputs.applicationDidFinishLaunching(
```
justinswart: SwiftFormat u drunk?

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -214,7 +216,7 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
       .take(first: 1)
       .observeValues { appOptions in
         _ = AppEnvironment.current.facebookAppDelegate.application(
-          appOptions.application,
+          appOptions.application ?? UIApplication.shared,
```
justinswart: The Facebook ApplicationDelegate interface updated to receive this as non-optional. Coalescing here is the quickest fix, an alternative would be to the mock this as a whole different type throughout the codebase which seemed like more work than its worth. It turns out this value is `nil` when tests run 🤷‍♂

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -223,7 +225,10 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
 
     self.facebookOpenURLReturnValue <~ openUrl.map {
       AppEnvironment.current.facebookAppDelegate.application(
-        $0.application, open: $0.url, sourceApplication: $0.sourceApplication, annotation: $0.annotation
+        $0.application ?? UIApplication.shared,
```
justinswart: Here too.

justinswart:
[Update Facebook SDK](https://trello.com/c/zXVrlelb/1405-update-facebook-sdk)

PR #705, opened 2019-06-13
Feature Flag Tools 🔨

ifbarrera:
# 📲 What

Adds some debug tooling for toggling active feature flags.

Also in this PR:
- refactored `BetaToolsViewController` to use programatic views (deleted the storyboard) and to use a table view instead of a stack view.

# 🤔 Why

To make it easier to test features that are hidden or turned off via feature flags.

# 🛠 How

To use the new tool, you navigate to the debug menu (the little bug icon on the top right of the discovery screen). The first row in the menu is `Feature Flags` which when tapped, displays a list of active feature flags with a switch corresponding to their current "enabled" status. Active feature flags are defined by cross-referencing a dictionary of `[String: Bool]` features returned from the `/config` endpoint with our local `Feature` enum. If the feature exists in the enum, it's displayed in the list of active features - otherwise it's ignored.

Toggling the switch on a given feature updates the `AppEnvironment`s config with the updated features dictionary, which will remain active **only until the /config endpoint is called again**. The endpoint is usually called on `ApplicationDidEnterForeground`, which means that if you background and foreground the app, you will lose your "configuration" of feature flags from the debug menu.

# 👀 See

T87iUAUaZ2

# ✅ Acceptance criteria

Feature flagging tool:
- [ ] **Logged in**  on a debug build, tap the bug icon on the top right of the discovery screen. You should see "Feature Flags" as the first option in the table. Tap "Feature Flags". You should see "Native Checkout" as the only active feature flag, and it should be **enabled**.
- [ ] **Logged out** on a debug build, tap the bug icon on the top right of the discovery screen. You should see "Feature Flags" as the first option in the table. Tap "Feature Flags". You should see "Native Checkout" as the only active feature flag, and it should be disabled.

Beta tools:
- [ ] Changing environment should work as expected.
- [ ] Changing language should work as expected.
- [ ] Submitting beta feedback should work as expected.
- [ ] Navigating to debugging push notifications should work as expected.


ifbarrera:
[Show Feature Flags on internal builds](https://trello.com/c/jRcF33VI/1408-show-feature-flags-on-internal-builds)

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -5,119 +5,169 @@ import Prelude
...
+      .observeForControllerAction()
+      .observeValues { [weak self] in
+        self?.showMailDisabledAlert()
+      }
   }
 
-  @IBAction fileprivate func betaFeedbackButtonTapped(_: Any) {
+  // MARK: - Selectors
+
+  @objc func betaFeedbackButtonTapped() {
```
dusi: `private`?

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -5,119 +5,169 @@ import Prelude
...
     self.viewModel.inputs.betaFeedbackButtonTapped(canSendMail: MFMailComposeViewController.canSendMail())
   }
 
-  @IBAction fileprivate func betaDebugPushNotificationsButtonTapped(_: Any) {
+  @objc func doneTapped() {
```
dusi: `private`

and also are we not routing this through VM for simplicity?

dusi: Also if we pass this action from outside this controller would not need to know about how to dismiss itself (so no matter if we present it modally or on a navigation stack the presenting view controller would decide the dismissal logic.

(just a thought)

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -135,16 +185,21 @@ internal final class BetaToolsViewController: UIViewController {
     self.present(alert, animated: true, completion: nil)
   }
 
-  private func showEnvironmentActionSheet() {
+  private func showEnvironmentActionSheet(sourceViewIndex: Int) {
+    guard let sourceView = self.tableView
```
dusi: It looks like this is still a bit off on iPad...I wonder if we could present this from the centre of the cell (since the cell is the element that you're interacting with (not the label) and therefore we can tap anywhere on it).

I tried quickly fixing this but wasn't really working just right...wonder if it's worth spending a bit time on this...it should be as simple as adjusting the coordinates of the view or maybe even better passing `sourceRect` to our convenience function to present popover.

This shows how to popover overlays the label and is technically aligning to `0, 0` coordinate of the label (ideally tis would be `midX` & `midY` of the cell maybe with some optional offset as well)
Screen Shot 2019-06-17 at 10 10 25 AM

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
+import Foundation
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+final class FeatureFlagToolsViewController: UITableViewController {
+  private let viewModel = FeatureFlagToolsViewModel()
```
dusi: Is there a reason why are we not placing `VM` under properties? Maybe if this is something we should do we need to document it so that everyone embraces it.

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+  // MARK: - Properties
+
+  private var features = [(Feature, Bool)]()
+  private let reuseId = "FeatureFlagTools.TableViewCell"
```
dusi: You can register a cell in more type safe manner using its class name like we do elsewhere `self.tableView.registerCellClass(SettingsTextInputCell.self)` and then reuse it like so `tableView.dequeueReusableCell(withClass: SettingsTextInputCell.self, for: indexPath)` ... unless there was an issue with the base `UITableViewCell` class?

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+  static func insantiate() -> FeatureFlagToolsViewController {
+    return FeatureFlagToolsViewController.init(nibName: nil, bundle: nil)
```
dusi: Again, could use designated initializer `init(style:)`

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+    self.viewModel.outputs.updateConfigWithFeatures
+      .observeForUI()
+      .observeValues { [weak self] features in
+        self?.updateConfig(with: features)
+      }
+  }
+
+  @objc func switchTogged(_ switchControl: UISwitch) {
```
dusi: `private`?

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    self.viewModel.inputs.didUpdateConfig()
+  }
+}
+
+// MARK: - UITableViewDataSource
+
+extension FeatureFlagToolsViewController {
+  override func numberOfSections(in _: UITableView) -> Int {
+    return 1
```
dusi: This should be a default number of sections so this data source function can be omitted.

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+  override func tableView(_: UITableView, numberOfRowsInSection _: Int) -> Int {
+    return self.features.count
+  }
+
+  override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+    let cell = tableView.dequeueReusableCell(withIdentifier: self.reuseId, for: indexPath)
+    let feature = self.features[indexPath.row]
+
+    let switchControl = UISwitch(frame: .zero)
```
dusi: Was the reason for not creating a custom cell with a switch control simplicity here? Looks like we're re-creating the cell every time.

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    _ = cell.textLabel
+      ?|> titleLabelStyle
+      ?|> \.text .~ feature.0.description
+
+    return cell
+  }
+}
+
+private let titleLabelStyle: LabelStyle = { label in
```
dusi: Can we use pragma mark to separate these `Styles`?

KsApi/models/Feature.swift
```diff
@@ -3,3 +3,11 @@ import Foundation
 public enum Feature: String {
   case checkout = "ios_native_checkout"
 }
+
+extension Feature: CustomStringConvertible {
```
dusi: 🙌

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+  @objc func switchTogged(_ switchControl: UISwitch) {
+    self.viewModel.inputs.setFeatureAtIndexEnabled(index: switchControl.tag, isEnabled: switchControl.isOn)
+  }
+
+  // MARK: - Private Helpers
+
+  private func updateConfig(with features: Features) {
+    guard let config = AppEnvironment.current.config else { return }
+
+    let updatedConfig = config |> \.features .~ features
```
ifbarrera: @justinswart here's the offending code 😄

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
ifbarrera: I think historically we've always but the view model at the top 🤷‍♀ I can move it to be part of the properties.

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
ifbarrera: Yeah just registering a basic `UITableViewCell` breaks `classNameWithoutModule`

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
ifbarrera: Yeah all of this was just to keep it as simple as possible. We're not really optimizing for performance here.

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
dusi: `FeatureFlagToolsViewCon` :trollface: `er.insane()`

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -155,13 +210,24 @@ internal final class BetaToolsViewController: UIViewController {
...
       name: Notification.Name.ksr_userLocalePreferencesChanged,
       object: nil,
       userInfo: nil
     )
+
+    self.navigationController?.popViewController(animated: true)
+  }
+
+  private func updateEnvironment(environment: EnvironmentType) {
+    let serverConfig = ServerConfig.config(for: environment)
```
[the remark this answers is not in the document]
justinswart: Just one thing to bear in mind when doing that sort of thing inside the view model is - if the environment is involved in the chain of signals that perform that replacement you may run the risk of the environment being switched out from underneath and the subscriptions being lost. Little tricky to explain but I had this happen once before something like:

```swift
AppEnvironment.current.signal
  .map { // transform something... }
  .observeValues { // replace environment }
```
Because this depends on the current environment doing the replacement here can be problematic. This might not exactly apply in what @Scollaco is suggesting 😄 My feeling is to do the replacement outside of the view model as it is now.

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
justinswart: How about we just change the type of alert to `.alert` instead of `.actionSheet` for iPad, this looks better to me and we also wouldn't need to care about the `sourceView`:

Screen Shot 2019-06-19 at 10 33 46 AM

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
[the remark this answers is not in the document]
ifbarrera: I can do that! 👍

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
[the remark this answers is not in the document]
ifbarrera: Yeah I thought about this and my feeling is that actually updating the environment is more of a side effect. This way, we can be certain that the right `EnvironmentType` is being sent, and we've separately tested that side-effect behavior works as expected.

We also do it this way in [`AppDelegate`](https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/AppDelegate.swift#L68)

Kickstarter-iOS/Views/Controllers/DiscoveryNavigationHeaderViewController.swift
```diff
@@ -138,12 +138,14 @@ internal final class DiscoveryNavigationHeaderViewController: UIViewController {
...
       .observeForControllerAction()
       .observeValues { [weak self] in
+        self?.betaToolsButton.tintAdjustmentMode = .dimmed
         self?.showDiscoveryFilters(selectedRow: $0)
       }
 
     self.viewModel.outputs.dismissDiscoveryFilters
       .observeForControllerAction()
       .observeValues { [weak self] in
+        self?.betaToolsButton.tintAdjustmentMode = .normal
```
ifbarrera: Whoops I was testing something here, ignore this will fix in a sec!

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
ifbarrera: I changed this to a modal presentation style, so now it's being used!

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
ifbarrera: I already updated it to pass the `sourceRect`, but I'm cool with either solution.

justinswart: My preference is `.alert` but it wouldn't say it's a strong feeling.

ifbarrera: Updated!

dannyalright:
Late to this party. Cool PR.

Agree with @dusi's comment:

> From the UX perspective I was wondering if..

> 1. We should present this modally (to emphasize interruption in the user workflow)
> 2. Whether we should use .formSheet modal presentation on the iPad (similar to our Settings - for consistency and since there's not that much content so it's just extravagant when presented full screen on iPad)
> 3. Use Beta Tools title in the nav bar? (we've previously had this I believe as the first label on the screen but removed it. Maybe also add titles when we push on to the Feature toggles or Notifications screens?

@dusi is this a good investment day task?

dusi:
Yep, I think it would be a simple investment day task.

Also, I'd consider rewriting this in `SwiftUI` since this is only an internal tool and it doesn't really matter it will be iOS 13 only...before doing any further logical changes to this.

dannyalright:
@dusi dang. Maybe we should schedule it for post-September then, when Catalina is public (and therefore SwiftUI is more accessible)?

PR #706, opened 2019-06-14
🦡 [App Badges] Improve voice-over support for badge value

justinswart:
# 📲 What

- Adds the localized string for `99+` so that the `+` can be moved around/modified by translators.
- Displays the uncapped badge value when voice over is enabled.

# 🤔 Why

iOS voice over natively reads badge value as "1 item", "2 items", "100 items", etc. This functionality was lost when we displayed the badge value as `99+` when it exceeded `99`, the screen reader would just read "ninety-nine plus". `UITabBarItem` does not provide access to its badge value view and digging through the view hierarchy for `_UIBadgeView` to add an accessibility hint seemed like the wrong approach.

# 🛠 How

Added an input to `RootViewModel` so that when voice over is enabled or disabled we can display the uncapped or capped badge value respectively.

# 👀 See

Enabling and disabling voice-over - note the screen reader focus appearing at the top and the badge value updating accordingly:

# ♿️ Accessibility

- [ ] Works with VoiceOver

# ✅ Acceptance criteria

- [ ] Run on device, send yourself a badge value greater than 99, enable and disable voice over.

### DIFF
--- a/Kickstarter-iOS/ViewModels/RootViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModel.swift
@@ -483,13 +483,7 @@ private func activitiesBadgeValue(with value: Int?) -> String? {
 
   guard clampedBadgeValue > 0 else { return nil }
 
-  guard badgeValue > maxBadgeValue else {
-    return "\(clampedBadgeValue)"
-  }
-
-  return localizedString(
-    key: "activities_badge_value_plus",
-    defaultValue: "%{activities_badge_value}+",
-    substitutions: ["activities_badge_value": "\(clampedBadgeValue)"]
-  )
+  return badgeValue > maxBadgeValue
+    ? Strings.activities_badge_value_plus(activities_badge_value: "\(clampedBadgeValue)")
+    : "\(clampedBadgeValue)"
 }
--- a/Kickstarter-iOS/ViewModels/RootViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/RootViewModel.swift
@@ -132,6 +132,9 @@ internal protocol RootViewModelInputs {
 
   /// Call from the controller's `viewDidLoad` method.
   func viewDidLoad()
+
+  /// Call when voice over is enabled or disabled
+  func voiceOverStatusDidChange()
 }
 
 internal protocol RootViewModelOutputs {
@@ -260,8 +263,8 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
 
     let updateBadgeValueOnLifecycleEvents = activityViewControllerIndex
       .takeWhen(lifecycleEvents)
-      .map { index in
-        (activitiesBadgeValue(with: AppEnvironment.current.application.applicationIconBadgeNumber), index)
+      .map { index -> (Int?, RootViewControllerIndex) in
+        (AppEnvironment.current.application.applicationIconBadgeNumber, index)
       }
 
     let selectedIndexAndActivityViewControllerIndex = Signal.combineLatest(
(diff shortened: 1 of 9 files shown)


Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -56,13 +57,21 @@ public final class RootTabBarViewController: UITabBarController {
...
+
     self.viewModel.outputs.updateUserInEnvironment
-      .observeValues { [weak self] user in
+      .observeValues { user in
         AppEnvironment.updateCurrentUser(user)
         NotificationCenter.default.post(.init(name: .ksr_userUpdated))
       }
 
-    self.userLocalePreferencesChanged = NotificationCenter
+    self.userLocalePreferencesChangedObserver = NotificationCenter
```
justinswart: Updated this name for consistency.

Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -77,10 +86,16 @@ public final class RootTabBarViewController: UITabBarController {
...
-    self.sessionEndedObserver.doIfSome(NotificationCenter.default.removeObserver)
-    self.sessionStartedObserver.doIfSome(NotificationCenter.default.removeObserver)
-    self.userUpdatedObserver.doIfSome(NotificationCenter.default.removeObserver)
-    self.applicationWillEnterForegroundObserver.doIfSome(NotificationCenter.default.removeObserver)
+    [
+      self.applicationWillEnterForegroundObserver,
+      self.sessionStartedObserver,
+      self.sessionEndedObserver,
+      self.userUpdatedObserver,
+      self.userLocalePreferencesChangedObserver,
```
justinswart: It looks like `userLocalePreferencesChangedObserver` was leaking before 😱

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -260,8 +263,8 @@ internal final class RootViewModel: RootViewModelType, RootViewModelInputs, Root
 
     let updateBadgeValueOnLifecycleEvents = activityViewControllerIndex
       .takeWhen(lifecycleEvents)
-      .map { index in
-        (activitiesBadgeValue(with: AppEnvironment.current.application.applicationIconBadgeNumber), index)
+      .map { index -> (Int?, RootViewControllerIndex) in
```
justinswart: Had to help the compiler along a few times here because of the merged Signal being a mix of `Int` and `Int?`

Kickstarter-iOS/ViewModels/RootViewModel.swift
```diff
@@ -477,19 +489,14 @@ extension TabBarItem: Equatable {
...
-  guard badgeValue > maxBadgeValue else {
-    return "\(clampedBadgeValue)"
-  }
-
-  return localizedString(
-    key: "activities_badge_value_plus",
-    defaultValue: "%{activities_badge_value}+",
-    substitutions: ["activities_badge_value": "\(clampedBadgeValue)"]
-  )
+  return (badgeValue > maxBadgeValue) && !isVoiceOverRunning
```
dusi: Is `badgeValue > maxBadgeValue` ever going to happen?

justinswart: Yes, the push notification can be any amount, this is the check to see if it's say `100` or more in the push payload, then display `99+`. When voiceover is enabled, we show the actual value from the payload without any `+`.

dusi:
Looks like we're going to have to also clamp it (or at least visually) for when Voice Over is on...maybe we should just not complicate things and use single logic for both cases (VoiceOver ON/OFF)?

IMG_0130

justinswart:
> Looks like we're going to have to also clamp it (or at least visually) for when Voice Over is on...maybe we should just not complicate things and use single logic for both cases (VoiceOver ON/OFF)?

We chatted on Slack, but just so that it's also recorded here - we show the uncapped value to retain the built-in behaviour of voice-over, so that the screen reader says "one hundred items" and not "99 plus".

Given that the user would have to have more than `9,999,999` unseen activity notifications before this becomes a UI issue (and only for visually impaired users with voice-over enabled), I don't think we should change this for that edge case.

PR #709, opened 2019-06-18
Move Facebook delegate methods to app target

justinswart:
# 📲 What

Moves calls that we make to `FBSDKApplicationDelegate` so that they are made from within our main app target.

# 🤔 Why

We've recently found, in a very specific Facebook login scenario, that users upon returning to our app - after completing a login flow via the Facebook app - are caught in an infinite loop. I believe that this change addresses the issue.

**Note:** This does _not_ affect users logging in via a browser flow and authenticating Facebook in Safari, _only_ if they select 'login via the app'.

In trying to track down the cause I found that placing the call to the `FBSDKApplicationDelegate` directly in our own `AppDelegate` in `application(_:open:options:)` worked without getting caught in any loop. This is problematic for us because we use our view model to mock the response from `FBSDKApplicationDelegate` in our unit tests. I then found that simply by moving the call to `FBSDKApplicationDelegate` just one level, into our view model's input function, I could reproduce the infinite loop again. This suggested to me that something to do with the bundle identifier of the framework from which the call is made might be problematic for `FBSDK`. It's possible that they perform some check internally against the bundle and if the two differ then the call is considered to have not been handled and it is called again (I surmise).

# 🛠 How

- Moved all Facebook-related calls and tests from `AppDelegateViewModel` into `FacebookLoginViewModel`.
- Made `FacebookLoginViewModel` a direct member of the application target.
- Works 💯 .

# ✅ Acceptance criteria

- [x] Test that deep-linking is unaffected.
- [x] Test login through Facebook via the app, not the browser.
- [x] Test logging in via the browser too.

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -208,6 +208,11 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     self.window?.tintColor = .ksr_green_700
 
+    self.viewModel.inputs.applicationDidFinishLaunching(
+      application: application,
+      launchOptions: launchOptions
+    )
+
     self.fbLoginViewModel.inputs.applicationDidFinishLaunching(
       application: application,
       launchOptions: launchOptions
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -220,7 +220,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     UNUserNotificationCenter.current().delegate = self
 
-    return self.fbLoginViewModel.outputs.applicationDidFinishLaunchingReturnValue
+    return self.viewModel.outputs.applicationDidFinishLaunchingReturnValue
   }
 
   func applicationWillEnterForeground(_: UIApplication) {
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -21,6 +21,7 @@ import UserNotifications
 internal final class AppDelegate: UIResponder, UIApplicationDelegate {
   var window: UIWindow?
   fileprivate let viewModel: AppDelegateViewModelType = AppDelegateViewModel()
+  fileprivate let fbLoginViewModel: FacebookLoginViewModelType = FacebookLoginViewModel()
 
   internal var rootTabBarController: RootTabBarViewController? {
     return self.window?.rootViewController as? RootTabBarViewController
@@ -207,14 +208,14 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     self.window?.tintColor = .ksr_green_700
 
-    self.viewModel.inputs.applicationDidFinishLaunching(
+    self.fbLoginViewModel.inputs.applicationDidFinishLaunching(
       application: application,
       launchOptions: launchOptions
     )
 
     UNUserNotificationCenter.current().delegate = self
 
-    return self.viewModel.outputs.applicationDidFinishLaunchingReturnValue
+    return self.fbLoginViewModel.outputs.applicationDidFinishLaunchingReturnValue
   }
 
   func applicationWillEnterForeground(_: UIApplication) {
@@ -238,7 +239,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     open url: URL,
     options: [UIApplication.OpenURLOptionsKey: Any] = [:]
   ) -> Bool {
-    return self.viewModel.inputs.applicationOpenUrl(
+    return self.fbLoginViewModel.inputs.applicationOpenUrl(
       application: app,
       url: url,
       options: options
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -239,11 +239,9 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     open url: URL,
     options: [UIApplication.OpenURLOptionsKey: Any] = [:]
   ) -> Bool {
-    return self.fbLoginViewModel.inputs.applicationOpenUrl(
-      application: app,
-      url: url,
-      options: options
-    )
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -238,11 +239,9 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
     options: [UIApplication.OpenURLOptionsKey: Any] = [:]
   ) -> Bool {
-    return self.viewModel.inputs.applicationOpenUrl(
-      application: app,
-      url: url,
-      options: options
-    )
+    self.viewModel.inputs.applicationOpenUrl(application: app, url: url, options: options)
+
+    return self.fbLoginViewModel.inputs.applicationOpenUrl(application: app, url: url, options: options)
```
justinswart: This pings our `AppDelegateViewModel` first so that deep-links are handled and then returns the value from `FBSDKApplicationDelegate`. I considered doing something like:

```swift
let deepLinkHandled = self.viewModel.inputs.applicationOpenUrl(application: app, url: url, options: options)

guard deepLinkHandled else {
  return self.fbLoginViewModel.inputs.applicationOpenUrl(application: app, url: url, options: options)
}

return deepLinkHandled
```
But in fact leaving it the way it is now more closely resembles how it's been all along and I'm just not sure if this is something I want to change at this stage 🤔. Open to thoughts on that.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -211,26 +208,8 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     self.postNotification = self.currentUserUpdatedInEnvironmentProperty.signal
       .mapConst(Notification(name: .ksr_userUpdated, object: nil))
 
-    self.applicationLaunchOptionsProperty.signal.skipNil()
```
justinswart: These have been moved to `FacebookLoginViewModel`.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -484,12 +462,11 @@ final class AppDelegateViewModelTests: TestCase {
 
     XCTAssertEqual(["App Open", "Opened App"], self.trackingClient.events)
 
-    let result = self.vm.inputs.applicationOpenUrl(
```
justinswart: This no longer has a return value because we never used it before, it was always `false`.

Kickstarter-iOS/ViewModels/FacebookLoginViewModel.swift
```diff
@@ -0,0 +1,91 @@
+import Library
+import UIKit
+import ReactiveSwift
+
+public protocol FacebookLoginViewModelInputs {
```
justinswart: All of the code in this file was just copy-pasted from `AppDelegateViewModel`.

Kickstarter-iOS/ViewModels/FacebookLoginViewModelTests.swift
```diff
@@ -0,0 +1,31 @@
+@testable import Kickstarter_Framework
+@testable import Library
+import ReactiveExtensions_TestHelpers
+import ReactiveSwift
+import XCTest
+
+final class FacebookLoginViewModelTests: TestCase {
```
justinswart: The only test that was testing this stuff copied from `AppDelegateViewModelTests`.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -207,14 +208,14 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     self.window?.tintColor = .ksr_green_700
 
-    self.viewModel.inputs.applicationDidFinishLaunching(
```
ifbarrera: I *think* we still need to call `self.viewModel.inputs.applicationDidFinishLaunching` somewhere in here - it's used for a bunch of stuff in `AppDelegateViewModel` 🤔

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -207,14 +208,14 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    self.fbLoginViewModel.inputs.applicationDidFinishLaunching(
       application: application,
       launchOptions: launchOptions
     )
 
     UNUserNotificationCenter.current().delegate = self
 
-    return self.viewModel.outputs.applicationDidFinishLaunchingReturnValue
+    return self.fbLoginViewModel.outputs.applicationDidFinishLaunchingReturnValue
```
ifbarrera: I wonder if we should still return our app delegate's `applicationDidFinishLaunchingReturnValue` here? It seems that `self.fbLoginViewModel.outputs.applicationDidFinishLaunchingReturnValue` is just hardcoded to `true` 🤔

ifbarrera: Actually do we need this property at all? https://github.com/kickstarter/ios-oss/pull/709/files#diff-66616e30e2b9d34ca192ff1ace432180R86

It would seem that it's value should be set from the return value facebook's `didFinishLaunchingWithOptions` [here](https://github.com/kickstarter/ios-oss/pull/709/files#diff-66616e30e2b9d34ca192ff1ace432180R38), but it looks like we've always just discarded that return value.

Kickstarter-iOS/AppDelegate.swift
justinswart: lol yes good catch - that would've broke things

Kickstarter-iOS/AppDelegate.swift
justinswart: The only time we set this value is from the shortcut items (long-press in springboard):

```swift
self.applicationDidFinishLaunchingReturnValueProperty <~ self.applicationLaunchOptionsProperty.signal
  .skipNil()
  .map { _, options in options?[UIApplication.LaunchOptionsKey.shortcutItem] == nil }
```

justinswart:
Closing as this fix is too brittle.

PR #711, opened 2019-06-18
💲[Native Checkout] Stepper haptics

dusi:
# 📲 What

Adds haptic feedback to the amount stepper.

Based on design specs:
* We're using `selection` feedback for incrementing/decrementing
* We're using `notification` feedback for reaching minimum/maximum value

# 🤔 Why

Providing great UX to our users

# 🛠 How

The logic is pretty simple, when the stepper reaches its minimum or maximum value we trigger `UIFeedbackGenerator.ksr_generateWarningFeedback()`, when the stepper increments/decrements within its limited we trigger `UIFeedbackGenerator.ksr_generateSelectionFeedback()` (iOS does the rest for us.

Please note that big chunk of work was actually refactoring our haptic feedback mechanism in order for it to being more testable and remove some duplicates.

Previously we used `UIButton` extension proxy to trigger feedback on `UIFeedbackGenerator`...some of those proxies were not even used so I've removed them and hopefully this will make things simpler.

So where we would previously use:
```swift
self.button.ksr_success()
```
we now don't need to reference self (mostly in closures) and can directly call
```swift
UIFeedbackGenerator.ksr_generateSuccess()
```
in addition to that this is going through `AppEnvironment.current` which allows us to actually observe whether the correct feedback was triggered or not and allows us to mock it.

# ✅ Acceptance criteria

You'll notice that currently we hardcode `min`, `max` and `initial value` (this is for the testing purposes and will be fixed in future work...right now we're using `10` for minimum stepper value, `20` for maximum stepper value and `15` for default stepper value).

- [ ] Incrementing or decrementing by either holding `-` or `+` stepper button triggers selection haptic feedback (very subtle)
- [ ] Reaching minimum or maximum triggers notification haptic feedback (more prominent)

# ⏰ TODO

This PR does not include the following (which will be separate pieces of work)
- [ ] Dynamically set `min`, `max` and `initial` values based on project setting
- [ ] Dynamically set `incrementation step value` (currently set to `1`)
- [ ] Allow stepper to interoperate with textfield input
- [ ] Limit textfield input to digits and decimal separators only (based on locale)


Kickstarter-iOS/Views/Cells/DiscoveryPostcardCell.swift
```diff
@@ -257,21 +257,15 @@ internal final class DiscoveryPostcardCell: UITableViewCell, ValueCell {
 
     self.watchProjectViewModel.outputs.generateImpactFeedback
       .observeForUI()
-      .observeValues { [weak self] in
```
dusi: We now no longer need `self` reference...in the future this could also be refactored into a side effect so there would not be a need for `generateImpactFeedback` signal from the VM. I identified that this would make the changes much bigger and decided against doing that.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -80,13 +86,45 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
 
     self.amountInputView.label.rac.text = self.viewModel.outputs.currency
     self.amountInputView.textField.rac.text = self.viewModel.outputs.amount
+
+    self.viewModel.outputs.stepperInitialValue
```
dusi: I will update `ReactiveExtensions` project to have the ability to use `rac` assignments. Maybe in a separate PR?

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
```diff
@@ -122,21 +122,15 @@ public final class ProjectNavBarViewController: UIViewController {
 
     self.watchProjectViewModel.outputs.generateImpactFeedback
       .observeForUI()
-      .observeValues { [weak self] in
-        self?.saveButton.generateImpactFeedback(style: .light)
```
dusi: Team notice that this is a thing of past (these proxies were just duplicating things and made a lot of reasoning about our code harder than need be).

So previously the flow would be
```swift
button.generateImpactFeedback
  -> UIFeedbackGenerator.ksr_impact()
    -> UIImpactFeedbackGenerator(style: style).impactOccurred()`
```

now we're removing one layer and doing
```swift
UIFeedbackGenerator.ksr_generateImpactFeedback()
  -> AppEnvironment.current.impactFeedbackGenerator.impactOccured
```
(please not that the naming is now more consistent) & (this also allows us to test things)

Cc: @justinswart @cdolm92 @ifbarrera @Scollaco

Library/Environment.swift
```diff
@@ -73,13 +73,19 @@ public struct Environment {
   /// The current set of launched countries for Kickstarter.
   public let launchedCountries: LaunchedCountries
 
+  /// Impact haptic feedback with light style
+  public let lightImpactFeedbackGenerator: UIImpactFeedbackGeneratorType
```
dusi: By making these injectable we can test haptics 🙌

Library/UIButton+HapticFeedback.swift
```diff
@@ -1,15 +0,0 @@
-import UIKit
-
-extension UIButton {
-  public func generateSelectionFeedback() {
-    UIFeedbackGenerator.ksr_selection()
-  }
-
-  public func generateSuccessFeedback() {
```
dusi: I believe previously only `generateSuccessFeedback` was used through proxy so the other 2 functions were unused

Library/UIFeedbackGenerator+Extensions.swift
```diff
@@ -1,15 +1,19 @@
 import UIKit
 
 extension UIFeedbackGenerator {
-  public static func ksr_success() {
-    UINotificationFeedbackGenerator().notificationOccurred(.success)
+  public static func ksr_generateImpactFeedback(style _: UIImpactFeedbackGenerator.FeedbackStyle = .light) {
```
dusi: Sorting alphabetically + more consistent naming

Library/ViewModels/PledgeAmountCellViewModel.swift
```diff
@@ -29,20 +35,61 @@ public final class PledgeAmountCellViewModel: PledgeAmountCellViewModelType,
...
+    self.amount = Signal.merge(
+      initialValue,
+      self.stepperValueProperty.signal
+    )
+    .map { String(format: "%.0f", $0) }
 
     self.currency = project
       .map { currencySymbol(forCountry: $0.country).trimmed() }
+
+    let minAndMax = Signal.combineLatest(project, reward)
```
dusi: Hardcoding `min`, `max` and `initialValue` for now..will be fixed in future PRs

Library/Environment.swift
```diff
@@ -117,11 +126,14 @@ public struct Environment {
     koala: Koala = Koala(client: KoalaTrackingClient(endpoint: .production)),
     language: Language = Language(languageStrings: Locale.preferredLanguages) ?? Language.en,
     launchedCountries: LaunchedCountries = .init(),
+    lightImpactFeedbackGenerator: UIImpactFeedbackGeneratorType = UIImpactFeedbackGenerator(style: .light),
```
justinswart: Given that at the callsite we access all of these through an extension on `UIFeedbackGenerator`, would it not be less to add to the `Environment` (and less to mock) if we just added a type like `UIFeedbackGeneratorType`?

I guess one other question I have is - does it matter if these generator instances exist for the lifetime of the app? I'm guessing they're pretty lightweight like tap gestures and it doesn't matter too much, but perhaps if we moved the instantiation of them into the type that we create then they'd only exist when we need them.

Lastly - I've seen in the documentation that we're supposed to call [prepare()](https://developer.apple.com/documentation/uikit/uifeedbackgenerator/2369818-prepare), I don't think we were doing that before, but perhaps it's a good idea to add it.

Library/TestHelpers/XCTestCase+AppEnvironment.swift
```diff
@@ -4,6 +4,7 @@ import KsApi
 import ReactiveSwift
 import XCTest
 
+// swiftlint:disable line_length
 extension XCTestCase {
   // Pushes an environment onto the stack, executes a closure, and then pops the environment from the stack.
   func withEnvironment(_ env: Environment, body: () -> Void) {
```
justinswart: When adding stuff to `Environment` be sure to also carry the arguments through to the `pushEnvironment` and `replaceCurrentEnvironment` functions in `Library/AppEnvironment.swift` too 👍

Cartfile
```diff
@@ -1,7 +1,7 @@
 ### Internal
 
 github "kickstarter/Kickstarter-Prelude" "c96c90b026d45839724e827f881dcba3ec812725"
-github "kickstarter/Kickstarter-ReactiveExtensions" "665b5cd4941e54e5d120b80218f62d6ae0e66dac"
+github "kickstarter/Kickstarter-ReactiveExtensions" "1b12cd6236aaace5e90a75332309c15c3585e162"
```
dusi: Please also review `ReactiveExtensions` updates

Kickstarter-iOS/Library/SharedFunctions.swift
```diff
@@ -1,6 +1,38 @@
 import Library
 import UIKit
 
+// MARK: - Haptic feedback
+
+func generateImpactFeedback(
```
dusi: Convenience functions should now be unified place to trigger with the following improvements:
- we call `prepare` to get the HW ready for haptic feedback
- we do not store reference to the haptic feedback generator (so the system releases it right after)

Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -3,6 +3,42 @@
 import XCTest
 
 internal final class SharedFunctionsTests: XCTestCase {
+  func testGenerateImpactFeedback() {
```
dusi: Tests to make sure that we call both `prepare` and the right haptic function

Kickstarter-iOS/Library/UIFeedbackGeneratorType.swift
```diff
@@ -0,0 +1,31 @@
+import UIKit
```
dusi: Protocol that allows us to mock and test our haptic feedback

Kickstarter-iOS/TestHelpers/MockAppEnvironment.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
```
dusi: Previously this was part of the test file that I touched, so I've extracted this since I was doing the same work for `MockFeedbackGenerator` (keeps the test file small and readable)

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -21,7 +21,7 @@ public protocol WatchProjectViewModelOutputs {
   var generateSelectionFeedback: Signal<(), Never> { get }
 
   /// Emits when haptic feedback should be generated
-  var generateSuccessFeedback: Signal<(), Never> { get }
+  var generateNotificationSuccessFeedback: Signal<(), Never> { get }
```
dusi: Renamed in order to be consistent with our haptic feedback convenience functions

Kickstarter-iOS/Library/SharedFunctions.swift
```diff
@@ -1,6 +1,38 @@
...
+  feedbackGenerator: UIImpactFeedbackGeneratorType = UIImpactFeedbackGenerator(style: .light)
+) {
+  feedbackGenerator.prepare()
+  feedbackGenerator.impactOccurred()
+}
+
+func generateNotificationSuccessFeedback(
```
justinswart: Any concern with these not being prefixed `ksr_`?

Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -3,6 +3,42 @@
...
+    let mockFeedbackGenerator = MockImpactFeedbackGenerator()
+
+    generateImpactFeedback(feedbackGenerator: mockFeedbackGenerator)
+
+    XCTAssertTrue(mockFeedbackGenerator.prepareWasCalled)
+    XCTAssertTrue(mockFeedbackGenerator.impactOccurredWasCalled)
+  }
+
+  func testGEnerateNotificationSuccessFeedback() {
```
justinswart: `GE` -> `Ge`

Kickstarter-iOS/Library/SharedFunctions.swift
dusi: Since those are not extension functions or anything like that I went with simplicity

Kickstarter-iOS/Library/SharedFunctionsTests.swift
dusi: 😅

dusi: Next time if you use Github's `suggestion` I can update it right from this PR

Kickstarter-iOS/Views/Controllers/ProjectNavBarViewController.swift
dusi: After some feedback from Justin it's now even simpler...we only have one publicly available function for each haptic type.

So what was previously this

```swift
button.generateImpactFeedback
  -> UIFeedbackGenerator.ksr_impact()
    -> UIImpactFeedbackGenerator(style: style).impactOccurred()
```

is now
```swift
generateImpactFeedback()
```

PR #712, opened 2019-06-19
Fix Facebook Login Infinite Loop

justinswart:
# 📲 What

This is my second attempt at fixing the strange infinite loop that we are experiencing with Facebook logins authorized via the Facebook app. See more info in the closed PR #709.

# 🤔 Why

The fix in #709 was too brittle and I believe this to be a simpler and more correct approach.

# 🛠 How

- Remove `FacebookAppDelegateProtocol` which we used to mock responses from `FBSDKApplicationDelegate`.
  - I think that it's largely unnecessary to test a response from FBSDK delegate. In the application we simply return this value via our `UIApplicationDelegate`.
- Remove the pattern we had of piping this through a `MutableProperty` in our `AppDelegateViewModel`. This was problematic and caused the application to freeze upon returning from a login flow via the FB app.

# ✅ Acceptance criteria

- [ ] Deep-linking works as before.
- [ ] FB login [scenarios](https://developers.facebook.com/docs/facebook-login/testing-your-login-flow/) work.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -43,10 +45,6 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       )
     )
 
-    // NB: We have to push this shared instance directly because somehow we get two different shared
-    //     instances if we use the one from `Environment.init`.
-    AppEnvironment.replaceCurrentEnvironment(facebookAppDelegate: ApplicationDelegate.shared)
```
justinswart: No longer needed as we don't mock this anymore.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -238,11 +236,17 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
     options: [UIApplication.OpenURLOptionsKey: Any] = [:]
   ) -> Bool {
-    return self.viewModel.inputs.applicationOpenUrl(
-      application: app,
-      url: url,
-      options: options
-    )
+    let fbReturnValue = ApplicationDelegate.shared.application(app, open: url, options: options)
+
+    if !fbReturnValue {
```
justinswart: If FB won't handle the URL, we just return whether we handle it which is always `true`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -697,7 +676,7 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     options: [UIApplication.OpenURLOptionsKey: Any]
   ) -> Bool {
     self.applicationOpenUrlProperty.value = (application, url, options)
-    return self.facebookOpenURLReturnValue.value
+    return true
```
justinswart: In my opinion this should always be `true`. If something has caused our app to open via a URL, we're going to handle that. We would only return `false` here if we specifically know our app does not handle a particular URL. And we only know that once we've tried to find a redirect URL which is asynchronous and wouldn't work for this return value.

PR #715, opened 2019-06-21
💲[Native Checkout] Pledge Summary Cell Layout

justinswart:
# 📲 What

Layout for `PledgeSummaryCell`.

**Note:** Layout only, I will likely be adding a view model in an upcoming PR and handling the updating of state between the cells when the amount changes.

# 🛠 How

This makes use of the same approach that we used in `PledgeDescriptionCell` using `UITextView` to set an `NSAttributedString` which is formatted with tappable links from HTML.

**Note:** Currently there is some arbitrary padding being added below the text which I am looking into and will fix in an upcoming PR:

Screen Shot 2019-06-21 at 10 32 43 AM

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-21 at 12 15 19 PM | Screen Shot 2019-06-21 at 12 12 23 PM |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type

# ✅ Acceptance criteria

- [x] Layout matches https://kickstarter.invisionapp.com/d/main#/console/16837064/349926403/preview
- [x] Tapping links currently prints them out in the console (opening Safari to come).

# ⏰ TODO

- [ ] Handle state between cells when amount changes.
- [ ] Fix padding below `UITextView`s.


Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -34,8 +34,8 @@ final class PledgeDataSource: ValueCellDataSource {
     }
 
     self.appendRow(
-      value: Strings.Total(),
-      cellClass: PledgeRowCell.self,
+      value: (project, 240, 25),
```
justinswart: Hard-coded for now, will handle this as part of handling the state changing between cells.

Kickstarter-iOS/DataSources/PledgeDataSourceTests.swift
```diff
@@ -61,7 +61,7 @@ final class PledgeDataSourceTests: XCTestCase {
     XCTAssertEqual(PledgeDescriptionCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: PledgeDataSource.Section.project.rawValue))
     XCTAssertEqual(PledgeAmountCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: PledgeDataSource.Section.inputs.rawValue))
     XCTAssertEqual(PledgeShippingLocationCell.defaultReusableId, self.dataSource.reusableId(item: 1, section: PledgeDataSource.Section.inputs.rawValue))
-    XCTAssertEqual(PledgeRowCell.defaultReusableId, self.dataSource.reusableId(item: 0, section: PledgeDataSource.Section.summary.rawValue))
```
justinswart: Do we want to delete `PledgeRowCell.swift` now?

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -203,18 +203,8 @@ private let dateLabelStyle: LabelStyle = { (label: UILabel) in
 
 private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) -> UITextView in
   _ = textView
+    |> tappableLinksViewStyle
     |> \.attributedText .~ attributedLearnMoreText()
-    |> \.isScrollEnabled .~ false
```
justinswart: Moved to shared function.

Kickstarter-iOS/Views/Cells/PledgeDescriptionCell.swift
```diff
@@ -223,25 +213,13 @@ private func attributedLearnMoreText() -> NSAttributedString? {
   // swiftlint:disable line_length
   let string = localizedString(
     key: "Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life_Learn_more_about_accountability",
-    defaultValue: "<p>Kickstarter is not a store. It's a way to bring creative projects to life.</br><a href=\"https://www.kickstarter.com/trust\">Learn more about accountability</a><p>"
+    defaultValue: "<p>Kickstarter is not a store. It's a way to bring creative projects to life.</br><a href=\"%{trust_link}\">Learn more about accountability</a><p>",
```
justinswart: Will add these for translation.

Library/Styles/CheckoutStyles.swift
```diff
@@ -91,3 +91,40 @@ public let checkoutLayerCardRoundedStyle: LayerStyle = { layer in
...
+    |> \.textContainerInset .~ UIEdgeInsets.zero
+    |> \.textContainer.lineFragmentPadding .~ 0
+    |> \.linkTextAttributes .~ [
+      .foregroundColor: UIColor.ksr_green_500
+    ]
+
+  return textView
+}
+
+public func checkoutAttributedLink(with string: String) -> NSAttributedString? {
```
justinswart: Not _really_ a style but kind've is. Figured it's fine to share here for now.

justinswart:

justinswart:
Note, I have merged #717 since it was approved, you can view that diff to make it easier to see what is specific to this PR.

Library/ViewModels/PledgeSummaryCellViewModel.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import Prelude
+import ReactiveSwift
+
+public protocol PledgeSummaryCellViewModelInputs {
+  func tapped(_ url: URL)
```
Scollaco: In other places where `tapped`  is used, we usually make the argument name explicit. It's easier to read and understand what was tapped, so by using `tapped(url: URL)`, we call it with `tapped(url: $0)`, instead of `tapped($0)`, for example.

Scollaco: I just noticed that this code is from the #717 😛 . But feel free to keep how it is if you think it's not necessary.

justinswart: As long as the type and be inferred from the function signature, in this case `URL` we don't need to reiterate that in the argument name. If this was `tapped(index: Int)` we would keep `index` as an argument because it's not obvious from the type what `index` is. Makes sense?

PR #717, opened 2019-06-21
💲[Native Checkout] Pledge Summary Cell Handle Links

justinswart:
# 📲 What

Adds functionality to open links that are tapped in the `PledgeSummaryCell`. This is one of the TODO items in #715.

# 🛠 How

When we construct the HTML to build the attributed string for the `UITextView` to display as tappable links, we derive these from the `HelpType`. The `UITextViewDelegate` passes us the `URL` that was tapped and we pass this to our view model for translation back to an `HelpType` which is the expected type for our `HelpWebViewController`.

# 👀 See

# ✅ Acceptance criteria

- [x] All links in the summary cell are tappable and open a web view modally.

### DIFF
--- a/Kickstarter-iOS/ViewModels/HelpWebViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/HelpWebViewModel.swift
@@ -24,7 +24,7 @@ internal final class HelpWebViewModel: HelpWebViewModelType, HelpWebViewModelInp
   internal init() {
     self.webViewLoadRequest = self.helpTypeProperty.signal.skipNil()
       .takeWhen(self.viewDidLoadProperty.signal)
-      .map { urlForHelpType($0, baseUrl: AppEnvironment.current.apiService.serverConfig.webBaseUrl) }
+      .map { $0.url(withBaseUrl: AppEnvironment.current.apiService.serverConfig.webBaseUrl) }
       .skipNil()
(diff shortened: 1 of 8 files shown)


Kickstarter-iOS/ViewModels/HelpWebViewModel.swift
```diff
@@ -44,22 +44,3 @@ internal final class HelpWebViewModel: HelpWebViewModelType, HelpWebViewModelInp
     self.viewDidLoadProperty.value = ()
   }
 }
-
-public func urlForHelpType(_ helpType: HelpType, baseUrl: URL) -> URL? {
```
justinswart: Moved this to be a function on the `enum`.

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -68,10 +68,15 @@ class PledgeTableViewController: UITableViewController {
   }
 
   internal override func tableView(_: UITableView, willDisplay cell: UITableViewCell, forRowAt _: IndexPath) {
-    if let descriptionCell = cell as? PledgeDescriptionCell {
```
justinswart: Felt like a `switch` was more readable here.

justinswart:

PR #718, opened 2019-06-22
💲[Native Checkout] Pledge Summary Cell Handle State Change

justinswart:
# 📲 What

Handles state changing between the cells in `PledgeViewController`. A continuation of the work in #715 and #717.

**Note:** Currently the software keyboard does not move the `UITableView` up when selecting the amount field, this should be addressed separately.

# 🤔 Why

Given that we're using a `UITableView`, the various components of this view (cells) do not have any context of the other data on the screen. When a user updates values we need that to reflect on the summary cell accurately.

# 🛠 How

As we update the amount that we're backing or change the shipping selection, delegates will pass this back to the `PledgeViewModel`, update the data source and re-configure the `PledgeSummaryCell`. Initially we thought we could simple reload the `IndexPath` of the `PledgeSummaryCell` but that causes a jarring bounce as the `UITableView` reloads that row. Instead I've retained a reference to the cell and simply reconfigured it with the new value.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Updating the amount value updates the summary value.
- [ ] If there is a shipping value that should be included in the total.


justinswart:

Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -12,34 +12,36 @@ final class PledgeDataSource: ValueCellDataSource {
 
   // MARK: - Load
 
-  func load(project: Project, reward: Reward, isLoggedIn: Bool) {
+  func load(data: PledgeViewData) {
+    self.clearValues()
```
justinswart: We need to clear values so that this function is idempotent when updating.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -3,13 +3,23 @@ import KsApi
 import Prelude
 import ReactiveSwift
 
+public typealias PledgeViewData = (
```
justinswart: I think we had `PledgeViewData` before but it disappeared 🤷‍♂.  I think that `typealias`es help us in here in that if the data structure changes we don't have to update all the function signatures that use that type.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -3,13 +3,23 @@ import KsApi
...
   func configureWith(project: Project, reward: Reward)
+  func pledgeAmountDidUpdate(to amount: Double)
+  func shippingAmountDidUpdate(to amount: Double)
   func viewDidLoad()
 }
 
 public protocol PledgeViewModelOutputs {
-  var reloadWithData: Signal<(Project, Reward, Bool), Never> { get }
+  var reloadWithData: Signal<PledgeViewData, Never> { get }
+  var updateWithData: Signal<PledgeViewData, Never> { get }
```
justinswart: `reload` and `update` are fairly similar in naming... suggestions?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -14,13 +14,25 @@ final class PledgeViewModelTests: TestCase {
...
     super.setUp()
 
+    self.vm.outputs.reloadWithData.observe(self.reloadWithData.observer)
     self.vm.outputs.reloadWithData.map { $0.0 }.observe(self.project.observer)
     self.vm.outputs.reloadWithData.map { $0.1 }.observe(self.reward.observer)
     self.vm.outputs.reloadWithData.map { $0.2 }.observe(self.isLoggedIn.observer)
+    self.vm.outputs.reloadWithData.map { $0.3 }.observe(self.total.observer)
+
+    self.vm.outputs.updateWithData.observe(self.updateWithData.observer)
+    self.vm.outputs.updateWithData.map { $0.0 }.observe(self.project.observer)
```
justinswart: Given that `reloadWithData` and `updateWithData` emit the same values and `reloadWithData` only emits once (and we test this), I figured it was fine to bind them to the same `TestObserver`s.

Kickstarter-iOS/DataSources/PledgeDataSource.swift
```diff
@@ -12,34 +12,36 @@ final class PledgeDataSource: ValueCellDataSource {
...
 
     self.appendRow(
-      value: (project, reward),
+      value: (data.project, data.reward),
       cellClass: PledgeAmountCell.self,
       toSection: Section.inputs.rawValue
     )
 
-    if reward.shipping.enabled {
+    if data.reward.shipping.enabled {
```
ifbarrera: I'm wondering if we should move this logic to the view model? 🤔 I know we have a test for this in the data source tests but it seems to me like something that we should keep in the view model in case we decide to swap out this table view for a stack view, let's say.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -80,6 +88,19 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
...
+      .observeValues { [weak self] amount in
+        guard let self = self else { return }
+        self.delegate?.pledgeAmountCell(self, didUpdateAmount: amount)
+      }
+  }
+
+  // MARK: - Actions
+
+  @objc func amountUpdated(_ textField: UITextField) {
+    self.viewModel.inputs.amountUpdated(to: textField.text ?? "")
```
ifbarrera: Should we just send through the `textField.text` regardless of whether it's `nil` or not and keep that logic all in the view model?

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -3,9 +3,14 @@ import Library
 import Prelude
 import UIKit
 
+protocol PledgeShippingLocationCellDelegate: AnyObject {
+  func pledgeShippingCell(_ cell: PledgeShippingLocationCell, didUpdateShippingAmount amount: Double)
```
ifbarrera: I wonder if we should just pass back the whole `ShippingRule` which includes `shippingRule.cost` (`Double`) but also the shipping location. I know we'll definitely need to send the shipping location to the `createBacking` mutation, so we might as well pass back the whole thing now? 🤷‍♀

Library/ViewModels/PledgeViewModel.swift
ifbarrera: Hehe pretty sure `PledgeViewData` has been added and removed like 10 times 😄

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -3,13 +3,23 @@ import KsApi
 import Prelude
 import ReactiveSwift
 
+public typealias PledgeViewData = (
+  project: Project,
+  reward: Reward,
+  isLoggedIn: Bool,
+  total: Double
```
ifbarrera: Can we use a slightly more descriptive name here? Maybe `pledgeTotal`?

Library/ViewModels/PledgeViewModel.swift
ifbarrera: Looks like `updateWithData` is basically just 1. reloading the data source and 2. re-configuring the summary cell. Is it possible to split up those two things and have `updateWithData` just do 2.? Then we can rename it to `configureSummaryCell` or something similar.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -31,20 +41,50 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       .map { _ in AppEnvironment.current.currentUser }
       .map(isNotNil)
 
-    self.reloadWithData = Signal.combineLatest(project, reward, isLoggedIn)
+    let shippingAmount = Signal.merge(
+      self.shippingAmountSignal,
+      projectAndReward.mapConst(0)
```
ifbarrera: I'm assuming that the default shipping cost is coming from the shipping cell's view model once it's configured?

Kickstarter-iOS/DataSources/PledgeDataSource.swift
dusi: Do you mean appending the row logic Isabel? Can you elaborate a bit more on this?

I think the way this is currently defined here should be correct since this is the whole purpose of the data source (to define how many rows a table view or collection view or any list UI will show). 

If we decide to switch architectures then this logic would probably move on to the controller's `viewDidLoad` where we would construct the stack view (unless we do some conditional UI loading there).

But I might not follow your question just right 🤔

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -40,6 +45,9 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
 
     self.spacer.widthAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(3)).isActive = true
 
+    self.amountInputView.textField
+      .addTarget(self, action: #selector(self.amountUpdated(_:)), for: .editingChanged)
```
dusi: Can we replace `self.amountUpdated` with class name instead `PledgeAmountCell.amountUpdated`? 

This is SwiftFormat's doing unfortunately 😄 If the selector is left without `self` SwiftFormat will add it there, so therefore I think we could be explicit and use the class name instead

Kickstarter-iOS/DataSources/PledgeDataSource.swift
ifbarrera: Oh I just mean instead of doing `data.reward.shipping.enabled` here, keeping that in the view model and just passing the actual value `isShippingEnabled` via the `PledgeViewData`:

```
if data.isShippingEnabled {
// append shipping row
}
```

That way we're separating out **how** we determine whether the cell is added or not (ie. the business logic) from the actual hiding/showing of the cell.

Does that make sense?

dusi:
I've noticed 2 things with the UX

1. Adjusting the font size
- I realize our inputs will never be this big but since there's a text field input it's pretty easy to enter arbitrary large value (see screenshot)
- should we not have a min scale value for the total amount label?

Screen Shot 2019-06-24 at 10 17 53 AM

2. Min value when no value entered
- I suppose keeping the last total when the text field value is empty is legit ... just double checking this was intentional (I'm not really sure about an ideal behaviour here)

Screen Shot 2019-06-24 at 10 18 37 AM

Library/ViewModels/PledgeViewModel.swift
justinswart: Well, currently, `reloadWithData` also _reloads_ the `UITableView` so we can't update the data source with that because we only want to silently update it and configure the cell. I also considered `pledgeViewDataAndReload` which would be like `(PledgeViewData, Bool)` where the `Bool` is whether or not to reload the table view. If we did this then we could have just one other signal to reconfigure the cell after the table view is reloaded. However - I am weary of running into that race condition we had before in another PR 🤔

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: Yip, I think I might've done this to test 👍

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
justinswart: Sure, I can update that.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: This change is already made in #719... it may be better to try and merge that first, that way the stepper will also update the total.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: Will also wait for #719 to merge for this.

Library/ViewModels/PledgeViewModel.swift
ifbarrera: Ah okay I see! Yeah this makes sense then.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: This is merged 👍

Library/ViewModels/PledgeViewModel.swift
justinswart: Yes but not all rewards will have shipping so we map this to zero to ensure that this signal emits at least once.

Kickstarter-iOS/DataSources/PledgeDataSource.swift
justinswart: Oh right, so you're suggesting we pass only the information that's needed right? I think the only downside to that would be duplication of data seeing as we have to pass `reward` to the datasource anyway?

PR #719, opened 2019-06-23
💲[Native Checkout] Pledge amount Stepper and Textfield input + Done button

dusi:
# 📲 What

* Makes both inputs (`stepper` and `text field`) work together
 * `Stepper` input updates `text field` accordingly
 * `Text field` input updates `stepper` accordingly
* Adds `Done` button to the keyboard that allows dismissing the keyboard if enabled
* Adds the ability to dismiss the keyboard by tapping anywhere on the screen

# 🤔 Why

The two inputs should be able to work together as well as the user should have a proper way of dismissing the keyboard. Also by enabling/disabling the `Done` button the user gets a visual cue that the entered value is valid or not.

# 🛠 How

Each input affects the other input component accordingly. So if the `stepper` value changes (increments or decrements) it updates the `text field` value. 

On the other hand if the `text field` input changes it updates the `stepper` accordingly. 

If the user enters value using the `text field` outside of the `min`/`max` range of the `stepper` it properly disables `steppers`'s `+` or `-` button based on the value. It also properly updates `stepper`'s `min`/`max` value accordingly so that next time `stepper` is used as input it starts incrementing from `min` (if the value entered using `text field` was bellow `min`) or decrementing from `max` (if the value entered using `text field` was above `max`).

# 👀 See

stepper-v-textfield

# ♿️ Accessibility 

This PR shouldn't have UI implications so no ACs here

# 🏎 Performance

This PR shouldn't have UI implications so no ACs here

# ✅ Acceptance criteria

**Stepper buttons are enabled/disabled properly**

- [x] Incrementing stepper to its `max` value (currently hardcoded to 20) disables the `+` button when the max value is reached
- [x] Decrementing stepper to its `min` value (currently hardcoded to 10) disables the `-` button when the max value is reached

**Done button is enabled/disabled properly**

- [x] The keyboard input has a toolbar with a `Done` button
- [x] Entering value lower than `10` disables the `Done` button (by `text field` input)
- [x] Entering value greater than `20` disables the `Done` button (by `text field` input)
- [x] Entering value within the `10` to `20` range enables the `Done` button

**Entering value through text field properly updates the stepper**

- [x] Entering value using the `text field` lower than `10` disables `stepper`'s `-` button
- [x] Entering value using the `text field` lower than `10` sets `stepper`'s `min` value to `10` so that next time the `+` button is tapped the amount value gets set to `11` (this is UIKit's behaviour so not much tweaking available)

- [x] Entering value using the `text field` greater than `20` disable's `stepper`'s `+` button
- [x] Entering value using the `text field` greater than `20` sets `stepper`'s `max` value to `20` so that next time the `-` button is tapped the amount value gets set to `19` (this is UIKit's behaviour so not much tweaking available)

**Done button**

- [x] Tapping the `Done` button dismisses the keyboard
- [x] Tapping anywhere on the screen dismisses the keyboard (even if the `Done` button is disabled)
- [x] Tapping `stepper`'s `-`/`+` buttons while the keyboard is on does not dismiss the keyboard

# ⏰ TODO

- [ ] Disable non-numeric input


Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -21,6 +49,10 @@ class AmountInputView: UIView {
 
     _ = ([self.label, self.textField], self.stackView)
       |> ksr_addArrangedSubviewsToStackView()
+
+    self.doneButton.heightAnchor.constraint(equalToConstant: Layout.Button.height).isActive = true
```
justinswart: Should we use `NSLayoutConstraint.activate()` when more than one is being activated?

Library/ViewModels/PledgeAmountCellViewModel.swift
```diff
@@ -71,25 +75,67 @@ public final class PledgeAmountCellViewModel: PledgeAmountCellViewModelType,
     self.generateNotificationWarningFeedback = stepperValueChanged
       .filter { min, max, value in value <= min || max <= value }
       .ignoreValues()
+
+    let textFieldValue = self.textFieldValueProperty.signal
```
justinswart: You can consider this here:

```swift
let textFieldValue = self.textFieldValueProperty.signal
  .skipNil()
  .map(Double.init)
  .skipNil()
```

PR #720, opened 2019-06-23
🧘‍♀️[Xcode] Zen out build warnings

dusi:
# 📲 What

Cleans up all outstanding Xcode warnings

# 🤔 Why

Warnings should be an indication that something is wrong. We haven't really pay much attention to this because our project has been producing at least couple warnings in the past and therefore it's easy to introduce new ones (because we simply don't take them seriously).

# 🛠 How

Updated Xcode project with Xcode's update tool

# 👀 See

| Before | After |
| --- | --- |
| Screen Shot 2019-06-23 at 8 47 22 PM | Screen Shot 2019-06-23 at 8 46 48 PM |

## 🐛 Architectures - Before
Screen Shot 2019-06-22 at 8 23 30 PM

## 🦋 Architectures - After
Screen Shot 2019-06-22 at 8 23 56 PM

## 🐛 Code signing - Before
Screen Shot 2019-06-22 at 8 20 47 PM

## 🦋 - Code signing - After
Screen Shot 2019-06-22 at 8 20 49 PM 

# ✅ Acceptance criteria

- [x] 32-bit devices are still not allowed
- [x] Project builds and produces no warnings (other than type-check limit in some expensive functions)


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -5544,7 +5548,6 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = NO;
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				ARCHS = "$(ARCHS_STANDARD)";
```
dusi: It looks like that after removing this it still resolves into `Standard Architecture` anyways ¯\_(ツ)_/¯

Kickstarter-iOS/Views/Controllers/ActivitiesViewController.swift
```diff
@@ -177,7 +177,7 @@ internal final class ActivitiesViewController: UITableViewController {
       }
 
     self.viewModel.outputs.updateUserInEnvironment
-      .observeValues { [weak self] user in
+      .observeValues { user in
```
dusi: And it all started because of this 😄

PR #721, opened 2019-06-24
Persist Debug Feature Flags

ifbarrera:
# 📲 What

[This PR](https://github.com/kickstarter/ios-oss/pull/705) added support for toggling local feature flags to make it easier to test feature flagged... features 😄. One issue with that PR was that because feature flags are stored in the `Config` object, and the object is refreshed on `ApplicationWillEnterForeground`, our feature flag settings would get reset if you background the app. This could lead to a confusing debugging experience - this PR fixes that issue by "caching" the debug feature flag settings.

# 🤔 Why

See "What".

# 🛠 How

Store debug settings in `AppEnvironment` and always prefer to take the debug settings, when updating the config.

# 👀 See

Notice how backgrounding the app doesn't reset our feature flag setting:

hMkWBSgR43

# ✅ Acceptance criteria

- [ ] Navigate to feature flag tools, toggle the native checkout feature flag. Navigate to a project and check that the feature is actually togged. Background the app. Navigate again to the feature flags tool and check that the flag is still toggled to whatever the previous value was. Navigate back to the project page and check that the feature is actually toggled still.

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -189,6 +189,7 @@
 		77556F5720A0B41E008CEA57 /* ThanksViewControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77556F5520A0B408008CEA57 /* ThanksViewControllerTests.swift */; };
 		7758A8372097BADF0018B96D /* DiscoveryProjectCategoryView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7758A83120979BE50018B96D /* DiscoveryProjectCategoryView.swift */; };
 		7758A8382097BB090018B96D /* DiscoveryProjectCategoryView.xib in Resources */ = {isa = PBXBuildFile; fileRef = 7758A8352097A8180018B96D /* DiscoveryProjectCategoryView.xib */; };
+		775D922922BD7CFD00442301 /* DebugData.swift in Sources */ = {isa = PBXBuildFile; fileRef = 775D922822BD7CFD00442301 /* DebugData.swift */; };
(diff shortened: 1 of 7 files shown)


KsApi/models/DebugData.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public struct DebugData {
+  public let config: Config?
```
ifbarrera: Kept this as optional in case we want to add to the struct in the future.

ifbarrera:
[Show Feature Flags on internal builds](https://trello.com/c/jRcF33VI/1408-show-feature-flags-on-internal-builds)

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -2463,6 +2465,14 @@
 			path = mutations;
 			sourceTree = "<group>";
 		};
+		777B88E322BD8E620027640F /* Recovered References */ = {
```
justinswart: Xcode sometimes adds this `Recovered References` folder, we can just delete it I think:

Screen Shot 2019-06-24 at 10 27 37 AM

Library/AppEnvironment.swift
```diff
@@ -54,10 +60,12 @@ public struct AppEnvironment: AppEnvironmentType {
   }
 
   public static func updateConfig(_ config: Config) {
+    let debugConfig = self.current.debugData?.config
```
justinswart: Can we coalesce here, then we don't have to do it 3 times below:

```swift
let config = self.current.debugData?.config ?? config
```

KsApi/models/DebugData.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public struct DebugData {
```
justinswart: Hmm just wondering if this should be a `KsApi` model... it's really just a wrapper type that we're using in the app, should perhaps just live in `Library?`

PR #722, opened 2019-06-25
💲[Native Checkout] Disable non-numeric, non-decimal separator input for the amount

dusi:
# 📲 What

Exactly what the title says

# 🤔 Why

We don't want users on iPads or with hardware keyboards to being able to input anything other than digits and decimal separator (localized)

# 🛠 How

Simply comparing input string to whether it's a numeric or decimal separator (based on `Locale`) and only allowing text field to proceed or not if it validates properly.

# ✅ Acceptance criteria

On iPad (with SW or HW keyboard connected) or on iPhone with HW keyboard connected (i.e. SIM) try the following input

- [ ] Entering any digit (`0123456789`) works
- [ ] Entering decimal number separator (i.e. `.` for `us` local or `,` for `jp` locale) works
- [ ] Entering multiple decimal number separators does not work
- [ ] Entering anything else does not work
- [ ] Dismissing the keyboard (by tapping anywhere on the screen) will keep the value untouched if it's within the `min`/`max` range (so currently `10-20`)
- [ ] Dismissing the keyboard (by tapping anywhere on the screen) while `text field` input is empty will set the value to `min` and dismiss keyboard
- [ ] Dismissing the keyboard (by tapping anywhere on the screen) while `text field` input is value 
smaller than 10 will set the text field value to `10` and dismiss keyboard
- [ ] Dismissing the keyboard (by tapping anywhere on the screen) while `text field` input is value greater than 20 will set the text field value to `20` and dismiss keyboard

### DIFF
--- a/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
@@ -141,7 +141,7 @@ extension PledgeAmountCell: UITextFieldDelegate {
     _ textField: UITextField, shouldChangeCharactersIn _: NSRange, replacementString string: String
   ) -> Bool {
     let decimalSeparatorCharacters = CharacterSet.ksr_decimalSeparators()
-    let existingCharacters = CharacterSet(charactersIn: textField.text ?? String.init())
+    let existingCharacters = CharacterSet(charactersIn: textField.text ?? "")
     let inputCharacters = CharacterSet(charactersIn: string)
     let numericCharacters = CharacterSet.ksr_numericCharacters()
 
--- a/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
@@ -38,6 +38,8 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
     _ = ([self.stepper, self.spacer, self.amountInputView], self.adaptableStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
+    self.amountInputView.textField.delegate = self
+
     self.spacer.widthAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(3)).isActive = true
 
     self.stepper.addTarget(
@@ -109,6 +111,25 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
 
   @objc func stepperValueChanged(_ stepper: UIStepper) {
     self.viewModel.inputs.stepperValueChanged(stepper.value)
+  }
+}
+
+extension PledgeAmountCell: UITextFieldDelegate {
+  func textField(
+    _ textField: UITextField, shouldChangeCharactersIn _: NSRange, replacementString string: String
+  ) -> Bool {
+    let decimalSeparatorCharacters = CharacterSet.ksr_decimalSeparators()
+    let existingCharacters = CharacterSet(charactersIn: textField.text ?? String.init())
+    let inputCharacters = CharacterSet(charactersIn: string)
+    let numericCharacters = CharacterSet.ksr_numericCharacters()
+
+    if numericCharacters.isSuperset(of: inputCharacters) {
+      return true
+    } else if decimalSeparatorCharacters.isSuperset(of: inputCharacters) {
+      return !decimalSeparatorCharacters.isSubset(of: existingCharacters)
+    } else {
+      return false
+    }
   }
 }
 
--- a/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
@@ -141,7 +141,7 @@ extension PledgeAmountCell: UITextFieldDelegate {
     _ textField: UITextField, shouldChangeCharactersIn _: NSRange, replacementString string: String
   ) -> Bool {
     let decimalSeparatorCharacters = CharacterSet.ksr_decimalSeparators()
-    let existingCharacters = CharacterSet(charactersIn: textField.text ?? "")
+    let existingCharacters = CharacterSet(charactersIn: textField.text.coalesceWith(""))
     let inputCharacters = CharacterSet(charactersIn: string)
     let numericCharacters = CharacterSet.ksr_numericCharacters()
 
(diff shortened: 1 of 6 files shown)


Library/CharacterSet.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public extension CharacterSet {
```
dusi: I've considered using `private extension` but in order for us to test this I think this is a better approach..

Library/CharacterSetTests.swift
```diff
@@ -0,0 +1,39 @@
...
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("0"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("1"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("2"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("3"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("4"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("5"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("6"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("7"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("8"))
+    XCTAssertTrue(CharacterSet.ksr_numericCharacters().contains("9"))
```
dusi: We can potentially also test alphabet symbols, but I thought it might be overkill ¯\_(ツ)_/¯

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -134,6 +136,25 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
   }
 }
 
+extension PledgeAmountCell: UITextFieldDelegate {
+  func textField(
+    _ textField: UITextField, shouldChangeCharactersIn _: NSRange, replacementString string: String
+  ) -> Bool {
+    let decimalSeparatorCharacters = CharacterSet.ksr_decimalSeparators()
+    let existingCharacters = CharacterSet(charactersIn: textField.text ?? String.init())
```
justinswart: `String.init()` -> `""`?

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -134,6 +136,25 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
...
+    let inputCharacters = CharacterSet(charactersIn: string)
+    let numericCharacters = CharacterSet.ksr_numericCharacters()
+
+    if numericCharacters.isSuperset(of: inputCharacters) {
```
justinswart: Potentially one less return if I have the logic correct:
```swift
guard decimalSeparatorCharacters.isSuperset(of: inputCharacters) else {
  return numericCharacters.isSuperset(of: inputCharacters)
}

return !decimalSeparatorCharacters.isSubset(of: existingCharacters)
```

Library/CharacterSet.swift
```diff
@@ -0,0 +1,14 @@
...
+  static func ksr_decimalSeparators() -> CharacterSet {
+    if let decimalSeparator = AppEnvironment.current.locale.decimalSeparator {
```
justinswart: One return? 🤷‍♂ 
```swift
return CharacterSet(charactersIn: AppEnvironment.current.locale.decimalSeparator ?? "")
```

Scollaco:
I had the opportunity to run on my device and have a couple suggestions that can improve (IMO) the UX:
- How it currently works, we can add as much decimals as we want. What do you think if we allow the user to type a max of 2 decimals (or whatever is acceptable for the current localization)?
- If we tap any amount, clear the textField, and tap anywhere on the screen to dismiss the keyboard, the textField will remain without text. I think it would be a good idea to reset the field to the `minimum` if this happens.

I can also create a card/PR to fix this, since it seems to be beyond the scope of this PR.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
dusi: LOL..exactly what I had...then I was like...no Justin will prefer `String.init()` 🤦‍♂

dusi:
Thanks for the suggestions @Scollaco 

I think for the amount of digits, we don't really have localized variants and I can imagine some currencies might support 3 some might prefer 2 so for now I would keep it this way and additionally fix this later on.

As for dismissing the keyboard when the field is empty, I believe I've discussed this with Danny and you're right..I'll see what I can do about it (since the logic is triggered on the `VC` not on the `VM`).

CC: @dannyalright in case you're strongly opinionated about any of these.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
dusi: Unfortunately your condition doesn't support deleting characters so I think I'll keep it the way it is for now.

dannyalright:
good eye, @Scollaco.

> If we tap any amount, clear the textField, and tap anywhere on the screen to dismiss the keyboard, the textField will remain without text. I think it would be a good idea to reset the field to the minimum if this happens.

This is what I imagined would happen—the field would reset to the minimum.

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
justinswart: 😂 we actually have a helper that I like called `coalesceWith` which is just `??` under the hood. but it would read `textField.text.coalesceWith("")`, I'm cool with either.

dusi: That's good to know...I can use that one...but it's not something we do consistently

Library/ViewModels/PledgeAmountCellViewModel.swift
```diff
@@ -8,11 +8,11 @@ public protocol PledgeAmountCellViewModelInputs {
   func configureWith(project: Project, reward: Reward)
   func doneButtonTapped()
   func stepperValueChanged(_ value: Double)
+  func textFieldDidEndEditing(_ value: String?)
   func textFieldValueChanged(_ value: String?)
 }
 
 public protocol PledgeAmountCellViewModelOutputs {
-  var amount: Signal<String, Never> { get }
```
dusi: Renamed to `textFieldValue` for better readability

PR #723, opened 2019-06-25
💲[Native Checkout] UITextView bottom padding improvements

justinswart:
# 📲 What

Improves arbitrary padding that was appearing at the bottom of the `UITextView` controls that we use for our in-line tappable links.

# 🤔 Why

There appears to be a bug in `UITextView` in that it sometimes adds spacing below the text and one has to implement this workaround to correct it 🤷‍♂ Reference: https://stackoverflow.com/questions/746670/how-to-lose-margin-padding-in-uitextview

# 🛠 How

Added workaround that recalculates the textview's height based on its width and applies it using a height constraint.

**Note:** This breaks if we change the font size using the accessibility inspector, as well as if the font size changes on device while the app is running. I wasn't able to prevent this and I feel we just have to accept that it's unlikely that users will frequently change the text size _while_ using the app.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 4 26 17 PM | Screen Shot 2019-06-25 at 4 22 41 PM |

# ✅ Acceptance criteria

- [x] `UITextView`s on the `PledgeViewController` don't add tons of arbitrary space at the bottom.
- [x] Expected to not work while changing font size during active app session.


dusi:
@justinswart don't forget to change the `base` branch so that we don't merge it to black hole 😄

Library/UITextView+SizeThatFits.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+import UIKit
+
+public extension UITextView {
+  func sizeThatFitsCurrentWidth() -> CGSize {
```
dusi: Since this is an extension should we do `ksr_sizeThat...`?

justinswart:
> Looks good...I've only noticed one issue which must be an edge case (unless there's some kind of a height constraint) for the smallest font selection (see screenshot).

I suspect this is because the thumbnail is causing it to grow to fill the `UIStackView` 🤔

PR #724, opened 2019-06-26
💲[Native Checkout] Project Pledge States UI - UI Fixes

dusi:
# 📲 What

Fixes `Dynamic Type` issues

# 🛠 How

Please see inline comments

# 👀 See

Please see inline comments

# ♿️ Accessibility 

- [ ] Dynamic type works for all the inline screenshots that were not working properly before

### DIFF
--- a/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
+++ b/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
@@ -8,14 +8,27 @@ final class PledgeCTAContainerView: UIView {
 
   private let vm: PledgeCTAContainerViewViewModelType = PledgeCTAContainerViewViewModel()
 
-  private lazy var amountAndRewardTitleStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var amountAndRewardTitleStackView: UIStackView = {
+    UIStackView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
   private lazy var amountOrRewardLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var pledgeCTAbutton: UIButton = {
+  private lazy var pledgeButton: UIButton = {
     MultiLineButton(type: .custom)
       |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var rootStackView: UIStackView = {
+    UIStackView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private lazy var spacer: UIView = {
+    UIView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
   private lazy var youreABackerLabel: UILabel = { UILabel(frame: .zero) }()
 
   // MARK: - Lifecycle
@@ -30,17 +43,12 @@ final class PledgeCTAContainerView: UIView {
     _ = ([self.youreABackerLabel, self.amountOrRewardLabel], self.amountAndRewardTitleStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    _ = ([self.amountAndRewardTitleStackView, self.pledgeCTAbutton], self.rootStackView)
+    _ = ([self.amountAndRewardTitleStackView, self.spacer, self.pledgeButton], self.rootStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    NSLayoutConstraint.activate(
-      [
-        self.pledgeCTAbutton.heightAnchor.constraint(
-          greaterThanOrEqualToConstant: Styles.minTouchSize.height
-        ),
-        self.pledgeCTAbutton.trailingAnchor.constraint(equalTo: self.rootStackView.trailingAnchor)
-      ]
-    )
+    NSLayoutConstraint.activate([
+      self.pledgeButton.heightAnchor.constraint(greaterThanOrEqualToConstant: Styles.minTouchSize.height)
+    ])
 
     self.bindViewModel()
   }
@@ -54,30 +62,32 @@ final class PledgeCTAContainerView: UIView {
   override func bindStyles() {
     super.bindStyles()
 
-    _ = self.pledgeCTAbutton
-      |> projectStateButtonStyle
+    let isAccessibilityCategory = self.traitCollection.preferredContentSizeCategory.isAccessibilityCategory
+
+    _ = self.amountAndRewardTitleStackView
+      |> \.axis .~ NSLayoutConstraint.Axis.vertical
+      |> \.isLayoutMarginsRelativeArrangement .~ true
+
+    _ = self.amountOrRewardLabel
+      |> \.font .~ .ksr_headline(size: 14)
+      |> \.textColor .~ .ksr_dark_grey_500
+      |> \.numberOfLines .~ 0
+
+    _ = self.pledgeButton
+      |> projectStateButtonStyle(
+        isAccessibilityCategory,
+        amountAndRewardTitleStackViewIsHidden: self.amountAndRewardTitleStackView.isHidden
+      )
+
+    _ = self.rootStackView
+      |> adaptableStackViewStyle(isAccessibilityCategory)
+      |> \.isLayoutMarginsRelativeArrangement .~ true
+      |> \.layoutMargins .~ UIEdgeInsets.init(topBottom: Styles.grid(3), leftRight: Styles.grid(3))
 
     _ = self.youreABackerLabel
       |> \.font .~ .ksr_headline(size: 14)
       |> \.text %~ { _ in Strings.Youre_a_backer() }
-
-    _ = self.amountOrRewardLabel
-      |> \.font .~ .ksr_caption1(size: 14)
-      |> \.textColor .~ .ksr_dark_grey_500
-
-    _ = self.rootStackView
-      |> checkoutAdaptableStackViewStyle(
-        self.traitCollection.preferredContentSizeCategory.isAccessibilityCategory
-      )
-      |> \.distribution .~ UIStackView.Distribution.equalCentering
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
-      |> \.isLayoutMarginsRelativeArrangement .~ true
-      |> \.layoutMargins .~ UIEdgeInsets.init(topBottom: Styles.grid(3), leftRight: Styles.grid(3))
-
-    _ = self.amountAndRewardTitleStackView
-      |> \.axis .~ NSLayoutConstraint.Axis.vertical
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
-      |> \.isLayoutMarginsRelativeArrangement .~ true
+      |> \.numberOfLines .~ 0
   }
 
   // MARK: - Binding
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -8,14 +8,27 @@ final class PledgeCTAContainerView: UIView {
...
 
-  private lazy var amountAndRewardTitleStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var amountAndRewardTitleStackView: UIStackView = {
+    UIStackView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
   private lazy var amountOrRewardLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var pledgeCTAbutton: UIButton = {
+  private lazy var pledgeButton: UIButton = {
```
dusi: Renamed for simplicity

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -8,14 +8,27 @@ final class PledgeCTAContainerView: UIView {
 
   private let vm: PledgeCTAContainerViewViewModelType = PledgeCTAContainerViewViewModel()
 
-  private lazy var amountAndRewardTitleStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var amountAndRewardTitleStackView: UIStackView = {
```
dusi: Enabling auto layout in initializers thorough this file so that we don't have to combine it with style

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -30,17 +43,12 @@ final class PledgeCTAContainerView: UIView {
     _ = ([self.youreABackerLabel, self.amountOrRewardLabel], self.amountAndRewardTitleStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    _ = ([self.amountAndRewardTitleStackView, self.pledgeCTAbutton], self.rootStackView)
+    _ = ([self.amountAndRewardTitleStackView, self.spacer, self.pledgeButton], self.rootStackView)
```
dusi: We had to add a spacer that would properly align the button to the right and fill in the space between the labels and the button.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -54,41 +62,44 @@ final class PledgeCTAContainerView: UIView {
...
-      |> \.text %~ { _ in Strings.Youre_a_backer() }
+    _ = self.amountAndRewardTitleStackView
+      |> \.axis .~ NSLayoutConstraint.Axis.vertical
+      |> \.isLayoutMarginsRelativeArrangement .~ true
 
     _ = self.amountOrRewardLabel
-      |> \.font .~ .ksr_caption1(size: 14)
+      |> \.font .~ .ksr_headline(size: 14)
       |> \.textColor .~ .ksr_dark_grey_500
+      |> \.numberOfLines .~ 0
```
dusi: | Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 5 49 33 PM copy | Screen Shot 2019-06-25 at 5 50 21 PM copy |

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -30,17 +43,12 @@ final class PledgeCTAContainerView: UIView {
...
-    _ = ([self.amountAndRewardTitleStackView, self.pledgeCTAbutton], self.rootStackView)
+    _ = ([self.amountAndRewardTitleStackView, self.spacer, self.pledgeButton], self.rootStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    NSLayoutConstraint.activate(
-      [
-        self.pledgeCTAbutton.heightAnchor.constraint(
-          greaterThanOrEqualToConstant: Styles.minTouchSize.height
-        ),
-        self.pledgeCTAbutton.trailingAnchor.constraint(equalTo: self.rootStackView.trailingAnchor)
```
dusi: | Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 5 50 46 PM | Screen Shot 2019-06-25 at 5 49 33 PM |

This constraint caused the button to align to the edge of the stack view. This was easily reproducible if you made dynamic type font large and open any pledge detail screen.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -100,12 +111,31 @@ final class PledgeCTAContainerView: UIView {
...
+      |> \.axis .~ axis
+      |> \.spacing .~ spacing
+  }
+}
+
+private func projectStateButtonStyle(
+  _ isAccessibilityCategory: Bool, amountAndRewardTitleStackViewIsHidden: Bool
+) -> (ButtonStyle) {
+  return { (button: UIButton) in
+    let lineBreakMode: NSLineBreakMode = isAccessibilityCategory || amountAndRewardTitleStackViewIsHidden
```
dusi: | Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 5 50 09 PM | Screen Shot 2019-06-25 at 5 49 02 PM |

This hopefully fixes the issue of not using `wordWrap` when the button can fit properly. Maybe there's a way around with `compression resistance` or `content hugging` priorities but I didn't have much luck..Cc: @justinswart 🤔

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -47,7 +50,7 @@ public final class PledgeCTAContainerViewViewModel: PledgeCTAContainerViewViewMo
           basicPledge,
           country: project.country,
           omitCurrencyCode: project.stats.omitUSCurrencyCode
-        )
+        ).trimmed()
```
dusi: | Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 5 49 02 PM | Screen Shot 2019-06-25 at 5 50 09 PM |

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -100,12 +111,31 @@ final class PledgeCTAContainerView: UIView {
...
+      ? NSLineBreakMode.byWordWrapping : NSLineBreakMode.byTruncatingTail
+
+    return button
+      |> roundedStyle(cornerRadius: 12)
+      |> UIButton.lens.titleColor(for: .normal) .~ UIColor.white
+      |> UIButton.lens.titleLabel.font .~ UIFont.ksr_headline(size: 15)
+      |> UIButton.lens.layer.borderWidth .~ 0
+      |> UIButton.lens.titleEdgeInsets .~ .init(topBottom: Styles.grid(1), leftRight: Styles.grid(2))
+      |> (UIButton.lens.titleLabel .. UILabel.lens.textAlignment) .~ NSTextAlignment.center
```
dusi: | Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-25 at 5 48 36 PM | Screen Shot 2019-06-25 at 5 48 36 PM |

Fixes the centering

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -54,41 +62,44 @@ final class PledgeCTAContainerView: UIView {
...
       |> \.layoutMargins .~ UIEdgeInsets.init(topBottom: Styles.grid(3), leftRight: Styles.grid(3))
 
-    _ = self.amountAndRewardTitleStackView
-      |> \.axis .~ NSLayoutConstraint.Axis.vertical
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
-      |> \.isLayoutMarginsRelativeArrangement .~ true
+    _ = self.youreABackerLabel
+      |> \.font .~ .ksr_headline(size: 14)
+      |> \.text %~ { _ in Strings.Youre_a_backer() }
+      |> \.numberOfLines .~ 0
```
dusi: Same as above

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -54,41 +62,44 @@ final class PledgeCTAContainerView: UIView {
...
   override func bindViewModel() {
     super.bindViewModel()
 
-    self.pledgeCTAbutton.rac.title = self.vm.outputs.buttonTitleText
-    self.pledgeCTAbutton.rac.backgroundColor = self.vm.outputs.buttonBackgroundColor
     self.amountAndRewardTitleStackView.rac.hidden = self.vm.outputs.stackViewIsHidden
     self.amountOrRewardLabel.rac.text = self.vm.outputs.rewardTitle
+    self.pledgeButton.rac.backgroundColor = self.vm.outputs.buttonBackgroundColor
+    self.pledgeButton.rac.title = self.vm.outputs.buttonTitleText
+    self.spacer.rac.hidden = self.vm.outputs.spacerIsHidden
```
dusi: We now also need to hide the spacer so we need 1 additional signal that should be tested as well

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -100,12 +111,31 @@ final class PledgeCTAContainerView: UIView {
...
 
-private let projectStateButtonStyle: ButtonStyle = { (button: UIButton) in
-  button
-    |> roundedStyle(cornerRadius: 12)
-    |> UIButton.lens.titleColor(for: .normal) .~ .white
-    |> UIButton.lens.titleLabel.font .~ .ksr_headline(size: 15)
-    |> UIButton.lens.layer.borderWidth .~ 0
-    |> UIButton.lens.titleEdgeInsets .~ .init(topBottom: Styles.grid(1), leftRight: Styles.grid(2))
-    |> (UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode) .~ .byWordWrapping
+private func adaptableStackViewStyle(_ isAccessibilityCategory: Bool) -> (StackViewStyle) {
```
dusi: Using private style because we were only using 50% of the global checkout style which we would use and then had to change 2 props back to default...so this should be simpler

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
dusi: This should probably be tested in our VM to make sure that the output is indeed trimmed. Current tests need better template values since those provided values are always trimmed.

dusi: We can pair on this

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
cdolm92: This could have been suggested in https://github.com/kickstarter/ios-oss/pull/702. Also don't think this is a good name because you cannot pledge after a project is over. You can only view rewards.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
cdolm92: I did not reproduce this but it could have been pointed out in https://github.com/kickstarter/ios-oss/pull/702.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -100,12 +111,31 @@ final class PledgeCTAContainerView: UIView {
...
+    let axis: NSLayoutConstraint.Axis = (isAccessibilityCategory ? .vertical : .horizontal)
+    let spacing: CGFloat = (isAccessibilityCategory ? Styles.grid(1) : 0)
+
+    return stackView
+      |> \.axis .~ axis
+      |> \.spacing .~ spacing
+  }
+}
+
+private func projectStateButtonStyle(
```
cdolm92: I think  `pledgeStateButtonStyle` would be a more accurate name.

justinswart:
Wondering if one can do anything to improve this loading situation - or do we have plans to in the future? (watch the container resize after the call completes).

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
justinswart: I think we want to avoid compression and hugging priorities if possible, seems like a good enough solution to change the `lineBreakMode`

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
dannyalright: I'm wondering if we should to handle Dynamic Type on fixed elements differently. The viewable part of the screen (beneath the `PledgeCTAContainerView`) is getting pretty tight!

dannyalright:
> Wondering if one can do anything to improve this loading situation - or do we have plans to in the future? (watch the container resize after the call completes).
> 
> 

Yeah. One simple thing that could help: the `Manage` button shouldn't grow depending on the left-text content. It should have a fixed height. I think that's the most visually awkward thing transitioning on screen.

PR #725, opened 2019-06-26
[Native Checkout] Pledge Payment Methods Collection View UI

Scollaco:
# 📲 What
- Adds the payment methods section to the `PledgeViewController`.

### Note: 
This PR doesn't include the PaymentMethods cells, so the collection view will be empty. This work will be done in a different PR.

# 🤔 Why
This is the initial work regarding Payment Methods on the Pledge screen. For more details:
-  https://trello.com/c/A5IgCbCA/1294-checkout-pledge-payment-methods-collection-view

# 🛠 How
- Created a new cell `PledgePaymentMethodsCell`
- The cell contains a label that works as a CollectionView header and an empty collection view.
- Added the cell to the `PledgeDataSource`
- Updated `PledgeDataSourceTests`

# 👀 See

| Before 🐛 | After 🦋 | Accessibility ♿️|
| --- | --- | --- |
| Screenshot 2019-06-26 at 16 53 26 | Screenshot 2019-06-26 at 16 31 55  | Screenshot 2019-06-26 at 16 32 36 |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for


Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
```diff
@@ -8,6 +8,7 @@ final class PledgeAmountCell: UITableViewCell, ValueCell {
   // MARK: - Properties
 
   private let viewModel = PledgeAmountCellViewModel()
+  private let dataSource = PledgeDataSource()
```
dusi: How about we put this `dataSource` on the `PledgeTableViewController` and maybe do a bit of cleanup so that there would be 2 data sources:

* `pledgeDataSource` - responsible for summary, amount, shipping, etc.
* `paymentMethodsDataSource` - responsible for payment methods

Then when the cell gets displayed in `tableView(_: UITableView, willDisplay cell: UITableViewCell, forRowAt _: IndexPath)` we establish the `dataSource` connection, maybe something like this:

```swift
let paymentMethodCell = (cell as? PledgePaymentMethodCell)
paymentMethodCell.collectionView.dataSource = self.paymentMethodsDataSource
```

(this way we keep our data sources consistently on view controllers and don't pollute table view cell with it)

Kickstarter-iOS/Views/Controllers/ActivitiesViewController.swift
```diff
@@ -177,7 +177,7 @@ internal final class ActivitiesViewController: UITableViewController {
       }
 
     self.viewModel.outputs.updateUserInEnvironment
-      .observeValues { [weak self] user in
+      .observeValues { user in
```
dusi: Weird, this should have been fixed on `master` 🤔

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-Framework-iOS.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1000"
+   LastUpgradeVersion = "1020"
```
dusi: Same think, this should have been updated on `master` and also propagated to `feature-native-checkout`

Kickstarter-iOS/Views/Controllers/ActivitiesViewController.swift
Scollaco: Hmm.. so maybe the feature-native-checkout branch is outdated?

Kickstarter-iOS/Views/Cells/PledgeAmountCell.swift
Scollaco: Wait... this line shouldn't be there 🤔 I think it was added by mistake, so you can ignore this. Regarding the `PaymentMethodsDataSource`, I was thinking it would be implemented once we create the `PledgeCreditCardCell`, because then we can configure the cell using the actual class using the real types to populate the cells. What do you think?

Kickstarter-iOS/Views/Cells/PledgePaymentMethodsCell.swift
```diff
@@ -0,0 +1,84 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+final class PledgePaymentMethodsCell: UITableViewCell, ValueCell {
+
+  private let cellIdentifier = "Cell"
```
justinswart: You can remove this and just use the cell's `defaultReusableId` property which is swizzled for us.

dusi: Or even better you should dequeue the cell based on its class name like we do for `UITableViewCell`s (might need some additional `UICollectionView` support)

Scollaco: I would go with dequeueing the cell based on its class name, but since this PR doesn't include the creation of the CreditCardCell class, for now I will use `defaultReusableId`. Does it sound good?

Kickstarter-iOS/Views/Cells/PledgePaymentMethodsCell.swift
```diff
@@ -0,0 +1,83 @@
...
+import Library
+import Prelude
+import UIKit
+
+final class PledgePaymentMethodsCell: UITableViewCell, ValueCell {
+
+  // MARK: - Properties
+
+  private lazy var collectionView: UICollectionView = {
+    UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewLayout())
```
justinswart: Seeing as we know we're going to need a `UICollectionViewFlowLayout` here, should we instantiate one with a `horizontal` `scrollDirection`?

Kickstarter-iOS/Views/Cells/PledgePaymentMethodsCell.swift
```diff
@@ -0,0 +1,90 @@
...
+    UICollectionViewFlowLayout()
+      |> \.scrollDirection .~ .horizontal
+  }()
+
+  private lazy var collectionView: UICollectionView = {
+    UICollectionView(frame: .zero, collectionViewLayout: self.layout)
+      |> \.dataSource .~ self
+  }()
+
+  private lazy var headerStackView: UIStackView = { UIStackView(frame: .zero) }()
```
justinswart: Can we call this `rootStackView`, we tend to do that for the top-most stack view, makes it easier to find 👍

PR #726, opened 2019-06-27
💲[Native Checkout] Load default, min and max pledge values properly

dusi:
# 📲 What

Up until now, we have used hard coded values of `min = 10`, `max = 20` and `value = 15`. This PR fixes this and properly loads this from `project` & `reward` combo.

# 🤔 Why

Replacing hard coded values with the real values.

# 🛠 How

Simple change but in order to properly test this I've added a LOT of tests.

# 👀 See

Please note how values are properly reflected between list & detail
Also note that the `total` reflects correct values after including shipping fee.

| List | Detail |
| --- | --- |
| Screen Shot 2019-06-26 at 5 36 05 PM |Screen Shot 2019-06-26 at 5 36 08 PM |
|Screen Shot 2019-06-26 at 5 36 16 PM | Screen Shot 2019-06-26 at 5 36 19 PM |

# ♿️ Accessibility 

Does not affect a11y

# ✅ Acceptance criteria

- [x] Default value is set to `min` value
- [x] `Min` value is set properly based on the reward
- [x] `Max` value is set properly based on the reward
- [x] Stepper increments in values equal to the `min` value (i.e. if `min` value is `15` then each step increments/decrements `15`)
- [x] Changing the amount to be less than `min` makes the `-` button disabled
- [x] Tapping the `+` button after previous step re-enables the `-` button, increments `min` value and updates text field with that value (i.e. if `min` value is `10`, setting the text field to `6` disables the `-` button, tapping `+` button re-enables the `-` button, increments stepper and changes text field value to `11`
- [x] Changing the amount to be greater than `max` makes the `+` button disabled
- [x] Tapping the `-` button after previous step re-enables the `+` button, decrements `max` value and updates text field with that value (i.e. if `max` value is `10_000`, setting the text field to `11_000` disables the `+` button, tapping `-` button re-enables the `+` button, decrements stepper and changes text field value to `9999`
- [x] Changing the amount based on the above criteria calculates and refreshes `total` value
- [x] Deleting the value in text field calculates the `total` as `0`


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -3598,13 +3600,14 @@
 				D015879D1EEB2ED6006E7684 /* CheckoutEnvelopeLenses.swift */,
 				D015879E1EEB2ED6006E7684 /* CommentLenses.swift */,
 				D015879F1EEB2ED6006E7684 /* ConfigLenses.swift */,
+				378CA24522C4449F004E3C86 /* CountryLenses.swift */,
 				D01587A01EEB2ED6006E7684 /* CreatePledgeEnvelopeLenses.swift */,
 				D01587A11EEB2ED6006E7684 /* DiscoveryEnvelopeLenses.swift */,
 				D01587A21EEB2ED6006E7684 /* DiscoveryParamsLenses.swift */,
+				D723708F2118E1F5001EA4CA /* ExportDataEnvelopeLenses.swift */,
```
dusi: Looks like something wasn't ordered alphabetically

KsApi/models/lenses/CountryLenses.swift
```diff
@@ -0,0 +1,29 @@
+import Prelude
+
+extension Project.Country {
```
dusi: In order for us to being able to set `minPledge` & `maxPledge` in tests

Library/ViewModels/PledgeAmountCellViewModelTests.swift
```diff
@@ -35,13 +35,61 @@ internal final class PledgeAmountCellViewModelTests: TestCase {
     self.vm.outputs.textFieldIsFirstResponder.observe(self.textFieldIsFirstResponder.observer)
   }
 
-  func testAmountAndCurrency() {
+  func testAmountCurrencyAndStepper_NoReward() {
```
dusi: There is a lot of variation tests mostly because of the way `min` & `max` is calculated - for more details please see `minAndMaxPledgeAmount `

Library/ViewModels/PledgeAmountCellViewModelTests.swift
```diff
@@ -142,59 +661,41 @@ internal final class PledgeAmountCellViewModelTests: TestCase {
...
 
     self.vm.inputs.stepperValueChanged(10)
     self.generateNotificationWarningFeedback.assertValueCount(2)
 
-    self.vm.inputs.stepperValueChanged(15)
+    self.vm.inputs.stepperValueChanged(11)
     self.generateNotificationWarningFeedback.assertValueCount(2)
   }
 
-  func testStepperMinValue() {
```
dusi: These are now part of the amount & currency so that we don't repeat the setup

KsApi/models/lenses/CountryLenses.swift
```diff
@@ -0,0 +1,29 @@
...
+  public enum lens {
```
justinswart: I think this is fine, we should probably figure out which approach we want to use going forward considering that we changed our other models to just be `public var` properties in order to use keypath lenses 🤔

Library/ViewModels/PledgeAmountCellViewModel.swift
```diff
@@ -42,10 +43,10 @@ public final class PledgeAmountCellViewModel: PledgeAmountCellViewModelType,
       .map(second)
 
     let minAndMax = Signal.combineLatest(project, reward)
-      .map { _ in (10.0, 20.0) }
+      .map(minAndMaxPledgeAmount)
 
-    let initialValue = Signal.combineLatest(project, reward)
-      .map { _ in 15.0 }
+    let initialValue = minAndMax.signal
```
justinswart: Don't think this needs the extra `.signal` at the end.

PR #727, opened 2019-06-27
💲[Native Checkout] Login & Sign-up on the pledge screen (Part 1)

ifbarrera:
# 📲 What

Allows logged out users to log in or sign up before pledging right from the pledge screen.

**Does not include**
- design updates to the login/signup screen, that's coming in Part 2 of this work

# 🤔 Why

You need to be signed into an account in order to back a project. This feature allows users to log in or sign up from within the pledging flow.

# 🛠 How

We already had this functionality in the old `DeprecatedRewardPledgeViewController`, which means the `LoginToutViewController` is already built to support login/signup from multiple contextual places throughout the app. Adding support for this in our new pledge screen was relatively straightforward since the `LoginToutViewController` manages it's own dismissal (this is something we should re-evaluate in the future). The `PledgeTableViewController` listens for `userSessionStarted` notifications which trigger a table view refresh.

# 👀 See

nep1zI4Hb5

# ♿️ Accessibility 

No accessibility changes required.

# ✅ Acceptance criteria

- [x] Logging in from the pledge screen works as expected.
- [x] Signing up for a new account for the pledge screen works as expected.
- [x] Logging in using Facebook works as expected.
- [x] Signing up in using Facebook works as expected.


Kickstarter-iOS/Views/Cells/PledgeContinueCell.swift
```diff
@@ -42,5 +64,11 @@ final class PledgeContinueCell: UITableViewCell, ValueCell {
       |> ksr_constrainViewToMarginsInParent()
 
     self.continueButton.heightAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(8)).isActive = true
+
+    self.continueButton.addTarget(self, action: #selector(self.continueButtonTapped), for: .touchUpInside)
```
justinswart: I think we're taking the convention of `Type.selector` for these, SwiftFormat does automatically insert `self` but if you replace that with `PledgeContinueCell` it will leave it as that.

Library/ViewModels/PledgeContinueCellViewModel.swift
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private let continueButtonTappedProperty = MutableProperty(())
+  public func continueButtonTapped() {
+    self.continueButtonTappedProperty.value = ()
+  }
+
+  public let goToLoginSignup: Signal<Void, Never>
+
+  public var inputs: PledgeContinueCellViewModelInputs {
```
justinswart: We usually put these on one line:

```swift
public var inputs: PledgeContinueCellViewModelInputs { return self }
public var outputs: PledgeContinueCellViewModelOutputs { return self }
```

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -59,22 +62,51 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         (project, reward, isLoggedIn, reward.shipping.enabled, total)
       }
 
+    let userUpdatedReload = data.takeWhen(self.userSessionStartedSignal)
```
justinswart: Can I suggest this, just to keep the mapping of the `true`/`false` for reload in one place?

```swift
let userUpdatedData = data.takeWhen(self.userSessionStartedSignal)
  .map { data in
    (
      project: data.project,
      reward: data.reward,
      isLoggedIn: AppEnvironment.current.currentUser != nil,
      isShippingEnabled: data.isShippingEnabled,
      pledgeTotal: data.pledgeTotal
    )
  }

let initialLoad = data.take(first: 1).map { data in (data, true) }
let silentReload = data.skip(first: 1).map { data in (data, false) }
let userUpdatedReload = userUpdatedData.map { data in (data, true) }

self.pledgeViewDataAndReload = Signal.merge(
  initialLoad,
  silentReload,
  userUpdatedReload
)
```

ifbarrera:
[[Checkout] Pledge Continue Button Cell - User Interaction](https://trello.com/c/uVA3sli0/1291-checkout-pledge-continue-button-cell-user-interaction)

Library/ViewModels/PledgeContinueCellViewModel.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+import ReactiveSwift
+
+public protocol PledgeContinueCellViewModelOutputs {
+  var goToLoginSignup: Signal<Void, Never> { get }
```
justinswart: I think this should be `notifyDelegateDidTapContinue`

Library/ViewModels/PledgeContinueCellViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+}
+
+public protocol PledgeContinueCellViewModelInputs {
+  func continueButtonTapped()
```
justinswart: And then this should be `pledgeContinueCellDidTapContinue`

Library/ViewModels/PledgeViewModel.swift
justinswart: Did you want to still make this change?

ifbarrera: I already made it

PR #730, opened 2019-06-28
Fix dSYMs upload bug

justinswart:
# 📲 What

Fixes an issue in that our dSYMs were silently failing to upload to Fabric/Crashlytics.

# 🤔 Why

tl;dr - We enjoy de-symbolicating stuff.

iOS crash logs only show symbols (memory addresses) in their stack traces unless these files are provided which allows Fabric to de-symbolicate them so that they are legible to mere mortals such as us.

# 🛠 How

Updated:
- Fastlane -> `2.126.0`.
- aws_s3 Fastlane plugin -> `1.6.0` (it kept warning about this).
- Fabric -> `3.13.2`.
  - This included a newer `upload-symbols` binary which appears to be where the change crept in. We keep this binary in our `bin` directory and it appears that the newer version of this handles an API change which was causing ours to break. An added benefit is that this file is significantly smaller than the previous binary 🤷‍♂ 

References:

https://github.com/fastlane/fastlane/issues/14140
https://github.com/fastlane/fastlane/issues/13096
https://docs.fastlane.tools/actions/upload_symbols_to_crashlytics/ (`binary_path`).

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-06-28 at 10 39 58 AM | Screen Shot 2019-06-28 at 10 40 11 AM |

# ✅ Acceptance criteria

- [x] Uploads dSYMs without the `Shell command exited with exit status 8 instead of 0.` error (see steps below).

You can test this locally by doing the follow:

```bash
make fabric #just to ensure the SDK is updated
bundle install #should update fastlane
bundle exec fastlane match_appstore
agvtool new-version -all $(($(date +%s)/100))
bundle exec fastlane itunes_gym
FABRIC_API_KEY=<api_key> bundle exec fastlane upload_dsyms #first make upload_dsyms lane public in Fastfile
```
^ Find the API KEY here: https://fabric.io/settings/organizations/ (select the production app, API key at the top).


justinswart:
[Fix Fastlane dSYM uploading to Crashlytics](https://trello.com/c/o8a2SR0P/1439-fix-fastlane-dsym-uploading-to-crashlytics)

PR #731, opened 2019-06-28
👩‍🔬 [Investment Day] - Test Suite Cleanup

dusi:
# 📲 What

* Updates comment references to `VoiceOver` thorough the codebase
* Initializes template model object with explicit type for faster type inference
* Renames environment variable to only reflect `1Password` support (as oppose to a whole OS version support)

# 🤔 Why

We need to clean up once in a while..

Also the adoption of our `isOSVersionAvailable` API wasn't as big as we assumed mostly due to some limiting factors. The `#available` API from Apple is way better in this regard. We haven't been able to adopt this widely mostly due to the following limitations..

We could not use

```swift
// We could not simply use the following condition
if ApiEnvironment.current.isOSVersionAvailable(12.0) {
  // Only do iOS 12 stuff
} else {
  // Only do iOS 11 stuff
  // The build would fail because the API couldn't be used directly unless it was under `#available` check which completely beats the purpose
}
```

# ✅ Acceptance criteria

- [ ] `1Password` button shows up and allows login using the `1Password` extension on `iOS 11` with `1Password.app` installed
- [ ] Tests raise exception when trying to run tests on anything but iPhone 8 / iOS 12 combo


Library/TestHelpers/TestCase.swift
```diff
@@ -71,11 +69,16 @@ internal class TestCase: FBSnapshotTestCase {
 }
 
 internal func preferredSimulatorCheck() {
```
justinswart: Can we keep this as:

```swift
internal func preferredSimulatorCheck() {
  guard
    let identifier = ProcessInfo().environment["SIMULATOR_MODEL_IDENTIFIER"],
    ["iPhone10,1", "iPhone10,4"].contains(identifier),
    #available(iOS 12.0, *)
  else {
    fatalError("Please only test and record screenshots on an iPhone 8 simulator running iOS 12")
  }
}
```
and then can we add this to `TestCase`:
```swift
override var recordMode: Bool {
  willSet(newValue) {
    if newValue {
      preferredSimulatorCheck()
    }
  }
}
```
This will only perform the `preferredSimulatorCheck()` when we are actually recording screenshots and not just running unit tests.
You can then remove the line calling `preferredSimulatorCheck()` from `setUp()` in `TestCase.swift` 👍

dusi:
@Scollaco @ifbarrera @cdolm92 please note that you can now test on any SIM you want (iPhone SE, iPhone X, etc. without running into `fatalError` as oppose to before where you could only run tests against iPhone 8). Obviously this does not apply to snapshot tests which have to be run against the same device form factor that they were recorded 😎 

Hope this will help make your day to day experience better. Thanks @justinswart for the idea!

PR #732, opened 2019-06-28
👩‍🔬 [Investment Day] - Localize zip/postal/post code based on locale region

dusi:
# 📲 What

This work has been inspired by @krstnfx and it simply localizes our `Zip code` field depending on the `Locale.regionCode`. You can read more [here](https://speakerdeck.com/krstnfx/internationalizing-your-app?slide=37).

# 🤔 Why

It's a nice a11y/inclusivity touch for our users.

# 🛠 How

We're simply checking `Locale.regionCode` and decide which string to show based on that. Since we've previously already localized `Zip code` into our supported languages (which already use their region flavours) this PR simply touches our `English` localization for base or non-supported languages (i.e. English, Swedish, Czech, etc.)

# 👀 See

| 🇺🇸`en_US` | 🌏 `en_GB` |
| --- | --- |
| Screen Shot 2019-06-28 at 3 21 59 PM | Screen Shot 2019-06-28 at 3 24 46 PM |

| 🇨🇦 `en_CA` | 🇨🇦 `fr_CA` |
| --- | --- |
| Screen Shot 2019-06-28 at 3 23 34 PM | Screen Shot 2019-06-28 at 3 26 26 PM |

# ♿️ Accessibility 

Since this only affects the base language the device language must be set to English

- [x] `Zip code` field is properly pronounced by VoiceOver when the device region is set to 🇺🇸
- [x] `Postal code` field is properly pronounced by VoiceOver when the device region is set to 🇨🇦
- [x] `Post code` field is properly pronounced by VoiceOver when the device region is set to anything else

# ✅ Acceptance criteria

Since this only affects the base language the device language must be set to English

- [x] The field shows `Zip code` label when the device region is set to 🇺🇸
- [x] The field shows `Postal code` when the device region is set to 🇨🇦
- [x] The field shows `Post code` when the device region is set to anything else

# ⏰ TODO

- [x] Have @dannyalright weight in on this and maybe test it on device


Kickstarter-iOS/Locales/de.lproj/Localizable.strings
```diff
@@ -322,6 +322,8 @@
 "Pledged_on_date" = "<b>Beitrag geleistet</b> am %{pledge_date}";
 "Popular" = "Beliebt";
 "Popular_Projects" = "Beliebte Projekte";
+"Post_code" = "Postleitzahl";
```
dusi: Since these strings have been previously translated and their locale flavours were taken into account we now unfortunately have 3 strings that are all equal (in other than base localizations). I couldn't think of a way around this problem but am open to any ideas.

Library/LocalizedString.swift
```diff
@@ -1,5 +1,18 @@
...
+public func localizedPostalCode() -> String {
+  // 🙏 Inspired by Kristina Fox
+  // https://speakerdeck.com/krstnfx/internationalizing-your-app?slide=37
+  switch AppEnvironment.current.locale.regionCode {
+  // 🇺🇸
+  case .some("US"): return localizedString(key: "Zip_code", defaultValue: "Zip code")
+  // 🇨🇦
+  case .some("CA"): return localizedString(key: "Postal_code", defaultValue: "Postal code")
+  // 🌍
+  default: return localizedString(key: "Post_code", defaultValue: "Post code")
```
dannyalright: I think _Post code_ should be one word (_postcode_). I've rarely seen it as two words in the wild (in Australia). Big fan of this PR otherwise.

dusi:
I've opened up a PR to address the strings first https://github.com/kickstarter/kickstarter/pull/16343 (so current status is still blocked).

dusi:
Blocked after discussing this with Jazzy - waiting for the translations

dusi:
The strings have been translated (and are actually identical to what we previously had)...should be good to go for final review.

PR #734, opened 2019-07-02
💲[Native Checkout] Reward->Pledge Transition Plumbing

justinswart:
# 📲 What

The initial plumbing for the transition from `RewardCollectionViewController` to `PledgeViewController`.

# 🤔 Why

Decided to break this up into its own PR to make reviewing easier.

# 🛠 How

- Added `CheckoutNavigationController` which will be the `UINavigationController` subclass for Checkout and can be extended to perform additional animations if necessary.
- Added `RewardPledgePushTransitionAnimator` and `RewardPledgePopTransitionAnimator`. I believe it makes sense to split these into their own animators but am open to refactoring down the line.

This approach is largely inspired by this great series of blog posts: https://devsign.co/notes/navigation-transitions-iii

**Note:** Currently there are no animations and to prove that the correct code paths are being called the push and pop animations will complete instantly.

# ✅ Acceptance criteria

- [ ] Tapping on a reward causes it to appear immediately (no animation).
- [ ] Tapping back on the navigation bar causes it to pop immediately (no animation).

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
+++ b/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
@@ -1,6 +1,8 @@
 import UIKit
 
 final class CheckoutNavigationController: UINavigationController {
+  // MARK: - Lifecycle
+  
   override func viewDidLoad() {
     super.viewDidLoad()
     self.delegate = self
--- a/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
+++ b/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
@@ -1,0 +1,26 @@
+import UIKit
+
+final class CheckoutNavigationController: UINavigationController {
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    self.delegate = self
+  }
+}
+
+extension CheckoutNavigationController: UINavigationControllerDelegate {
+  func navigationController(
+    _: UINavigationController,
+    animationControllerFor operation: UINavigationController.Operation,
+    from fromVC: UIViewController,
+    to toVC: UIViewController
+  ) -> UIViewControllerAnimatedTransitioning? {
+    switch (operation, fromVC, toVC) {
+    case (.push, is RewardPledgeTransitionAnimatorDelegate, is PledgeViewController):
+      return RewardPledgePushTransitionAnimator()
+    case (.pop, is PledgeViewController, is RewardPledgeTransitionAnimatorDelegate):
+      return RewardPledgePopTransitionAnimator()
+    default:
+      return nil
+    }
+  }
+}
--- a/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
+++ b/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
@@ -5,6 +5,7 @@ final class CheckoutNavigationController: UINavigationController {
   
   override func viewDidLoad() {
     super.viewDidLoad()
+
     self.delegate = self
   }
 }
--- a/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
+++ b/Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
@@ -2,7 +2,7 @@ import UIKit
 
 final class CheckoutNavigationController: UINavigationController {
   // MARK: - Lifecycle
-  
+
   override func viewDidLoad() {
     super.viewDidLoad()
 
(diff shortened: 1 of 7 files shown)


Kickstarter-iOS/Views/Transitions/RewardPledgePopTransitionAnimator.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
+
+public class RewardPledgePopTransitionAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+  public func transitionDuration(
+    using _: UIViewControllerContextTransitioning?
+  ) -> TimeInterval {
+    return 0.3
```
justinswart: These will be adjusted as we build out the animations.

Kickstarter-iOS/Views/Transitions/RewardPledgePushTransitionAnimator.swift
```diff
@@ -0,0 +1,18 @@
...
+
+public class RewardPledgePushTransitionAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+  public func transitionDuration(
+    using _: UIViewControllerContextTransitioning?
+  ) -> TimeInterval {
+    return 0.3
+  }
+
+  public func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+    guard let toView = transitionContext.view(forKey: .to) else { return }
```
justinswart: In the blog post examples there were delegate and vc properties on the animators but I think that that is unnecessary and will instead `guard let` them here.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -1096,8 +1099,6 @@
 		D7774466217A345D008D679F /* UpdateUserProfileMutation .swift in Sources */ = {isa = PBXBuildFile; fileRef = D77743E2217A2D67008D679F /* UpdateUserProfileMutation .swift */; };
 		D78E038E229305E90043E92F /* PledgeStateCTAType.swift in Sources */ = {isa = PBXBuildFile; fileRef = D78E0355229305E20043E92F /* PledgeStateCTAType.swift */; };
 		D78E039022930DF80043E92F /* PledgeCTAContainerViewViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D78E038F22930DF80043E92F /* PledgeCTAContainerViewViewModelTests.swift */; };
-		D78E4E482188CB4300E99295 /* UIButton+HapticFeedback.swift in Sources */ = {isa = PBXBuildFile; fileRef = D78E4E472188CB4300E99295 /* UIButton+HapticFeedback.swift */; };
```
dusi: Oh no haptic, don't go :D

Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+
+final class CheckoutNavigationController: UINavigationController {
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    self.delegate = self
```
dusi: Just to separate call to super and make it stand out a bit more. Definitely can live without this :) 

```suggestion

    self.delegate = self
```

Kickstarter-iOS/Views/Transitions/RewardPledgePopTransitionAnimator.swift
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  public func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+    guard let toView = transitionContext.view(forKey: .to) else { return }
```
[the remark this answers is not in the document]
justinswart: It'll change quite a bit when I fill these functions in.

Kickstarter.xcodeproj/project.pbxproj
justinswart: Weird 🤔

justinswart:
> Plus wanted to double check that this way (by using custom UINavigationController subclass) we have intentionally disabled swipe to go back gesture?

I'm guessing this is because I'm overriding the animation stuff and that behaviour relies on an `UIPercentDrivenInteractiveTransition`, I don't think the swipe from left will make sense once the transition is full working 🤔 

> Also weirdly the navigation bar still animates ... does this mean it will still have the push animation even though we're trying to present modal-ish style?

Good question - will have to figure this out when I work on the transition but my guess is that I'll be hiding the nav bar or in some other way preventing that standard animation.

Kickstarter.xcodeproj/project.pbxproj
justinswart: Ok, as we discussed - this file was removed before but the project file must have had some stray references to it.

justinswart:
[Reward Screen to Pledge Screen Animation](https://trello.com/c/Euu0uxzB/1371-reward-screen-to-pledge-screen-animation)

PR #736, opened 2019-07-03
💲[Native Checkout] Refactor shipping rules

dusi:
# 📲 What

Big refactor of pledge shipping rules logic

# 🤔 Why

In order for us to being able to hold `shippingRules` in memory and not having to reload it from graph when the cell gets reused as well as being able to later on passing these rules to a shipping rule picker.

# 🛠 How

Lots of shuffling around as well as tests.

# ✅ Acceptance criteria

- [x] Pledge screen shows rewards that don't have shipping rules
- [x] Pledge screen shows rewards that have shipping rules (default rule is applied for now)


Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -93,32 +99,23 @@ final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
...
+      isLoading: value.isLoading,
+      project: value.project,
+      selectedShippingRule: value.selectedShippingRule
+    )
   }
 
-  // MARK: - Public Functions
+  // MARK: - Actions
 
-  func animate(_: Bool) {}
```
dusi: We don't need this for now...there's a card on our sprint board to address the loading state .. in the future this should be driven from the VM not externally

Library/ViewModels/PledgeShippingLocationCellViewModel.swift
```diff
@@ -24,60 +21,26 @@ public protocol PledgeShippingLocationCellViewModelType {
...
-      Strings.We_were_unable_to_load_the_shipping_destinations()
-    }
+    self.shippingLocation = projectAndSelectedShippingRule
+      .map { _, selectedShippingRule in selectedShippingRule.location.localizedName }
   }
 
-  private let projectAndRewardProperty = MutableProperty<(Project, Reward)?>(nil)
-  public func configureWith(project: Project, reward: Reward) {
-    self.projectAndRewardProperty.value = (project, reward)
+  private let configureWithDataProperty = MutableProperty<(Bool, Project, ShippingRule?)?>(nil)
```
justinswart: I think we've generally just called these `configDataProperty`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -31,43 +40,90 @@ public protocol PledgeViewModelType {
...
+      .switchMap { (project, reward) -> SignalProducer<Signal<[ShippingRule], ErrorEnvelope>.Event, Never> in
+        AppEnvironment.current.apiService.fetchRewardShippingRules(projectId: project.id, rewardId: reward.id)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .map(ShippingRulesEnvelope.lens.shippingRules.view)
+          .retry(upTo: 3)
+          .prefix(value: [])
+          .materialize()
+      }
+
+    self.presentShippingRules = shippingRulesEvent.values()
```
justinswart: Is this just a placeholder for now? Perhaps add a comment and set to `.empty`?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -31,43 +40,90 @@ public protocol PledgeViewModelType {
...
+      Strings.We_were_unable_to_load_the_shipping_destinations()
+    }
+
+    let pledgeTotal = Signal.combineLatest(pledgeAmount, shippingAmount).map(+)
 
-    let data = Signal.combineLatest(project, reward, isLoggedIn, total)
-      .map { project, reward, isLoggedIn, total -> PledgeViewData in
-        (project, reward, isLoggedIn, reward.shipping.enabled, total)
+    // swiftlint:disable line_length
+    let data = Signal.combineLatest(project, reward, isLoggedIn, isShippingLoading, selectedShippingRule, pledgeTotal)
```
justinswart: We might be able to improve readability extracting this to a global function in this file?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1,179 +1,444 @@
...
 
-  private let reload = TestObserver<Bool, Never>()
+  private let configureSummaryCellWithDataPledgeTotal = TestObserver<Double, Never>()
+  private let configureSummaryCellWithDataProject = TestObserver<Project, Never>()
+
+  private let pledgeViewDataAndReloadIsLoggedIn = TestObserver<Bool, Never>()
+  private let pledgeViewDataAndReloadIsShippingEnabled = TestObserver<Bool, Never>()
+  private let pledgeViewDataAndReloadIsShippingRulesLoading = TestObserver<Bool, Never>()
+  private let pledgeViewDataAndReloadProject = TestObserver<Project, Never>()
+  private let pledgeViewDataAndReloadReload = TestObserver<Bool, Never>()
```
justinswart: Can I suggest adding a comment here for future Justin and Pavel, something like:

```
/*
`pledgeViewDataAndReload` emits the initial data for this view model along with a `Bool` indicating whether the table view should be reloaded (currently only the first time but probably during login as well). Every time any of the combined data of `PledgeViewData` changes, this signal will emit all of its data in order to ensure that the data source has the current state of the data that the cells have. This ensures that if a cell is recycled it will be reloaded with its most recent data. Given the noise that this frequent emission creates and also that what we really care about is the final emission from this signal, we only test its last value.
*/
```

Maybe this should actually be a comment above the VM output?

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -93,32 +99,23 @@ final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
...
+  @objc func shippingLocationButtonTapped(_: UIButton) {}
```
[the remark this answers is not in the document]
dusi: I agree with the naming and was always not in sync with our existing naming conventions...I'll take a stab an try creating a separate PR to address this thorough the whole codebase

Library/ViewModels/PledgeShippingLocationCellViewModel.swift
```diff
@@ -5,15 +5,12 @@ import ReactiveExtensions
...
 }
 
 public protocol PledgeShippingLocationCellViewModelOutputs {
-  var amount: Signal<NSAttributedString, Never> { get }
-  var location: Signal<String, Never> { get }
-  var selectedShippingRule: Signal<ShippingRule, Never> { get }
-  var shippingIsLoading: Signal<Bool, Never> { get }
-  var shippingRulesError: Signal<String, Never> { get }
+  var amountAttributedText: Signal<NSAttributedString, Never> { get }
+  var shippingLocation: Signal<String, Never> { get }
```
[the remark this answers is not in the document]
dusi: Yep..will fix

Library/ViewModels/PledgeViewModel.swift
dusi: Yup, it's a placeholder...I'm using it right in the follow up PR so would be undoing some unnecessary work (plus altering tests)...let's keep it this way for now?

Library/ViewModels/PledgeViewModelTests.swift
dusi: I'll put it on the VM above the output

Library/ViewModels/PledgeViewModel.swift
justinswart: Sounds good

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -19,8 +25,17 @@ public protocol PledgeViewModelInputs {
 }
 
 public protocol PledgeViewModelOutputs {
-  var configureSummaryCellWithProjectAndPledgeTotal: Signal<(Project, Double), Never> { get }
+  var configureShippingLocationCellWithData: Signal<(Bool, Project, ShippingRule?), Never> { get }
+  var configureSummaryCellWithData: Signal<(Project, Double), Never> { get }
+
+  /// Emits the initial data for this view model along with a `Bool` indicating whether the table view should
```
justinswart: I think we'd typically demarcate a long comment like this with `/* ... */`?

PR #738, opened 2019-07-04
💲[Native Checkout] Shipping Picker - Navigation Workflow - 1/4

dusi:
# 📲 What

Hooks up the shipping location cell button to present shipping locations picker.

Please note that this simply hooks up the button and presents an empty `VC`.

Also note that as soon as you present the modal, you won't be able to dismiss it (just kill the SIM or restart the app)

# 🤔 Why

Adds the ability to show the picker with all available shipping rules.

# 🛠 How

The plumbing was already there, we just needed to connect this to the right user interaction

# 👀 See

| iPhone | iPad |
| --- | --- |
| Screen Shot 2019-07-04 at 1 13 00 PM | Screen Shot 2019-07-04 at 2 25 19 PM |

somethings-fucky

# ✅ Acceptance criteria

For the rewards that support shipping..

- [ ] Tapping the `Shipping location button` will show a `cyan` modal (using our custom sheet style)

# ⏰ TODO

- [ ] Fix modal presentation on iPad
- [ ] Show the list of shipping rules
- [ ] Add the ability to select a different shipping rule
- [ ] Add the ability to search shipping rules


Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -117,10 +119,24 @@ class PledgeTableViewController: UITableViewController {
 
   // MARK: - Actions
 
+  @objc func dismissKeyboard() {
```
dusi: Little bit of reordering #boyScoutRule

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -117,10 +119,24 @@ class PledgeTableViewController: UITableViewController {
...
+    self.present(nc, animated: true)
+  }
+
+  private func presentShippingRules(
+    _: Project, shippingRules _: [ShippingRule], selectedShippingRule _: ShippingRule
+  ) {
+    let vc = UIViewController()
+    vc.view.backgroundColor = UIColor.cyan
+    let nc = UINavigationController(rootViewController: vc)
+    let shieetVC = SheetOverlayViewController(child: nc, offset: self.view.frame.height / 4)
```
dusi: I'm open to any suggestions re this magic number 😄

dusi: Uh, I've also noticed, the naming isn't `ideal` 🤦‍♂

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -130,9 +146,9 @@ extension PledgeTableViewController: PledgeDescriptionCellDelegate {
   }
 }
 
-extension PledgeTableViewController: PledgeSummaryCellDelegate {
-  internal func pledgeSummaryCell(_: PledgeSummaryCell, didOpen helpType: HelpType) {
-    self.presentHelpWebViewController(with: helpType)
+extension PledgeTableViewController: PledgeAmountCellDelegate {
```
dusi: Delegate conformance sorted alphabetically now

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -66,15 +66,20 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    self.presentShippingRules = Signal.combineLatest(
+      project,
+      shippingRulesEvent.values(),
+      defaultShippingRule.skipNil()
     )
+    .takeWhen(self.shippingRuleSignal.signal)
 
     let shippingAmount = Signal.merge(
-      selectedShippingRule.skipNil().map { $0.cost },
+      defaultShippingRule.skipNil().map { $0.cost },
```
dusi: Renaming as this is the `default` value..when we allow selection we will add another signal replacing this.

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
dusi: There was no real need for division...used value that was close to design specs

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
```diff
@@ -101,6 +101,13 @@ final class PledgeShippingLocationCell: UITableViewCell, ValueCell {
 
     self.amountLabel.rac.attributedText = self.viewModel.outputs.amountAttributedText
     self.shippingLocationButton.rac.title = self.viewModel.outputs.shippingLocationButtonTitle
+
+    self.viewModel.outputs.shippingLocationSelected
+      .observeForUI()
+      .observeValues { [weak self] shippingRule in
+        guard let self = self else { return }
+        self.delegate?.pledgeShippingCell(self, didSelectShippingRule: shippingRule)
```
justinswart: Seeing as `ShippingRule` is a type that explains this at the callsite, we can actually make this:

```swift
self.delegate?.pledgeShippingCell(self, didSelect: shippingRule)
```

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -67,6 +73,14 @@ class PledgeTableViewController: UITableViewController {
         }
       }
 
+    self.viewModel.outputs.presentShippingRules
+      .observeForUI()
+      .observeValues { [weak self] project, shippingRules, selectedShippingRule in
+        self?.presentShippingRules(
+          project, shippingRules: shippingRules, selectedShippingRule: selectedShippingRule
```
justinswart: The argument naming here is fine because `selectedShippingRule` is a noun in this case, if we changed this to `selected:` even though we have the `ShippingRule` type it would change the meaning to be an action.

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -3,6 +3,12 @@ import Library
 import Prelude
 import UIKit
 
+private enum Layout {
+  enum Sheet {
+    static let offset: CGFloat = 200
```
justinswart: This fine for different device sizes?

Library/ViewModels/PledgeShippingLocationCellViewModel.swift
```diff
@@ -6,11 +6,13 @@ import ReactiveSwift
 
 public protocol PledgeShippingLocationCellViewModelInputs {
   func configureWith(isLoading: Bool, project: Project, selectedShippingRule: ShippingRule?)
+  func shippingLocationButtonTapped()
 }
 
 public protocol PledgeShippingLocationCellViewModelOutputs {
   var amountAttributedText: Signal<NSAttributedString, Never> { get }
   var shippingLocationButtonTitle: Signal<String, Never> { get }
+  var shippingLocationSelected: Signal<ShippingRule, Never> { get }
```
justinswart: Not `selectedShippingLocation` ?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -74,15 +74,20 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
-      self.presentShippingRules.map(defaultShippingRule(fromShippingRules:))
+      shippingRulesEvent.values().map(defaultShippingRule(fromShippingRules:))
+    )
+
+    self.presentShippingRules = Signal.combineLatest(
+      project,
+      shippingRulesEvent.values(),
+      defaultShippingRule.skipNil()
     )
+    .takeWhen(self.shippingRuleSignal.signal)
```
justinswart: `shippingRuleSignal` doesn't seem like the right name here, this probably just points to our delegate naming perhaps not being explanatory enough. I think we'd want this to indicate that the `selectShippingRuleButton` was tapped and that the currently selected `ShippingRule` was passed in.

So perhaps the delegate should be:

```swift
func pledgeShippingCellWillPresentShippingRules(_: PledgeShippingLocationCell, selectedShippingRule rule: ShippingRule)
// or
func pledgeShippingCellPresentedShippingRules(_: PledgeShippingLocationCell, selectedShippingRule rule: ShippingRule)
```
I think we can then base the view model input on that and it should improve the signal naming.

Kickstarter-iOS/Views/Cells/PledgeShippingLocationCell.swift
justinswart: See later comment, I now realize exactly what the delegate is doing, it's really just presenting the rules, the actual selection will happen elsewhere.

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
dusi: Yeah, it looks like we need enough space to still show the shipping location cell behind the modal.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -141,7 +147,12 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     self.pledgeAmountObserver.send(value: amount)
   }
 
-  private let (shippingRuleSignal, shippingRuleObserver) = Signal<ShippingRule, Never>.pipe()
+  private let presentShippingRulesProperty = MutableProperty<(ShippingRule)?>(nil)
+  public func presentShippingRules(with rule: ShippingRule) {
```
dusi: Added new signal to present shipping rules...the old signal was more appropriate for when we actually select a rule..I've kept it around because there's bunch of test related to that...this should be a plumbing for the selection itself.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -20,6 +20,7 @@ public typealias PledgeViewData = (
 public protocol PledgeViewModelInputs {
   func configureWith(project: Project, reward: Reward)
   func pledgeAmountDidUpdate(to amount: Double)
+  func presentShippingRules(with rule: ShippingRule)
```
justinswart: Typically we want our inputs to represent _what happened_, and our outputs to be named according to _what should happen_, so in this case our output is `presentShippingRules` because that's the side effect, but our input should be more like exactly what happened - `pledgeShippingCellWillPresentShippingRules` which is matches our delegate naming.

PR #737, opened 2019-07-04
💲[Native Checkout] Reward -> Pledge Transition Feature

justinswart:
# 📲 What

Feature branch collecting the PR's for the Reward -> Pledge transition animation.

# 👀 Changelog

## #740 Refactor RewardCell to RewardCardView

Refactored `RewardCell` into `RewardCell`, `RewardCardContainerView`, `RewardCardView`.

## #734 - Reward->Pledge Transition Plumbing

The initial plumbing for the transition from `RewardCollectionViewController` to `PledgeViewController`.

PR #741, opened 2019-07-07
💲[Native Checkout] Reward -> Pledge View transition animation

justinswart:
# 📲 What

A continuation of #737 and #740, this adds the heavily lifting for the transition from `RewardCollectionViewController` to `PledgeViewController`.

**Note A:** While this is ready for a high-level review, please take note of the TODOs at the bottom of this description as there are still a number of tweaks needed. It should just be a matter of finessing the animation into the final product, I'm hoping the hard part is done, these tweaks could be in a separate PR if that makes it easier for review.

**Note B:** Currently this only works reliably in portrait orientation and that can also break if one rotates to landscape and back to portrait while on the pledge view. In reviewing the design pack, it is mentioned that this is only expected to work in portrait so we may just want to lock the orientation of these view controllers.

# 🤔 Why

This transition is key to the design language of taking us from the list (or upcoming grid) of reward cards to the pledge view and maintaining a sense of continuity between these views.

# 🛠 How

- Firstly, all of the moving around of views in #740 was necessary to have everything in right place for snapshotting, etc.
- Filled in the requirements of `RewardPledgeTransitionAnimatorDelegate`.
- Conformed `PledgeTableViewController` (via `PledgeViewController`) and `RewardsCollectionViewController` to `RewardPledgeTransitionAnimatorDelegate` protocol.
- Implemented the transition behaviours in `RewardPledgePushTransitionAnimator` and `RewardPledgePopTransitionAnimator`, both are defined in `RewardPledgeTransitionAnimators.swift` for simplicity and sharing of code.

# 👀 See

| Small card (button remains 👇) | Larger card, button clipped | Larger card, button overlay reappears on return | Scrolled down card, returns to original position |
| --- | --- | --- | --- |
|  |  |  |  |

☝️ Keeping the button is necessary for the two views to have the exact same size at the time of taking snapshots for the transition. If it's important to never show the button in the thumbnail we can work on this a bit more but it was proving to be pretty complicated.

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Tapping a reward in `RewardCollectionView` pushes `PledgeViewController` on to the navigation stack with the transition shown above.
- [x] Tapping the thumbnail pops the view controller with transition shown above.

# ⏰ TODO

- [x] Slide and fade `PledgeViewController` up as the thumbnail settles into place.
- [x] Improve timing and spring tension of the animation.
- [ ] ~~Pop animation is meant to have the `RewardCell`s slide in from the left and right on returning.~~
- [ ] ~~Consider allowing the thumbnail height to be based on the text to the right of it (may improve how it looks for accessibility sizes).~~
- [x] Disable pledge button when view is in thumbnail size.
- [x] Add _expansion_ icon to the top right corner of the thumbnail.
- [x] Add drop-shadow to thumbnail.
- [x] Add gesture and animation to give the effect of _de-pressing_ card on tap.
- [x] Fix the `CGRect` conversion between the views on iPad.


justinswart:
[Reward Screen to Pledge Screen Animation](https://trello.com/c/Euu0uxzB/1371-reward-screen-to-pledge-screen-animation)

justinswart:
@ifbarrera this last merge commit includes the pill cells and state image stuff from #755 #757. Due to the renames and stuff I had to manually move things around but I think it's all there and working.

@dusi please confirm!

Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
```diff
@@ -0,0 +1,29 @@
+import UIKit
+
+final class CheckoutNavigationController: UINavigationController {
```
ifbarrera: We haven't really been using the term `Checkout` with the new work, I wonder if this should be `PledgeNavigationController`?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -1,45 +1,60 @@
...
   }()
 
-  private lazy var descriptionStackView: UIStackView = { UIStackView(frame: .zero) }()
-  private lazy var estimatedDeliveryLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var dateLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var spacerView = UIView(frame: .zero)
-  private lazy var learnMoreTextView: UITextView = { UITextView(frame: .zero) |> \.delegate .~ self }()
+  private let viewModel = PledgeDescriptionViewModel()
+
+  private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
```
ifbarrera: Should this be alphabetize along with the rest of the properties?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -1,45 +1,60 @@
...
 
   // MARK: - Lifecycle
 
   override func viewDidLoad() {
     super.viewDidLoad()
 
     self.configureSubviews()
+
+    self.bindViewModel()
```
ifbarrera: I don't think we need to manually call this, should happen by default from calling `super.viewDidLoad()`

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -74,36 +86,79 @@ final class PledgeDescriptionViewController: UIViewController {
...
+    self.rewardCardContainerShadowViewHeightConstraint = self.rewardCardContainerShadowView.heightAnchor
+      .constraint(equalToConstant: 0)
+
+    let rewardCardContainerShadowViewConstraints = [
+      self.rewardCardContainerShadowViewWidthConstraint,
+      self.rewardCardContainerShadowViewHeightConstraint
+    ]
+    .compact()
 
     NSLayoutConstraint.activate([
```
ifbarrera: Can we add all the constraints to a `setupConstraints` function?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -112,12 +167,47 @@ final class PledgeDescriptionViewController: UIViewController {
...
+      from: actualRect,
+      toRect: thumbnailRect
+    )
+
+    self.view.setNeedsLayout()
+  }
+
+  // MARK: - Actions
+
+  @objc func rewardCardTapped() {
```
ifbarrera: `private`?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -130,12 +220,50 @@ final class PledgeDescriptionViewController: UIViewController {
...
 
-  internal func configureWith(value: Reward) {
-    self.viewModel.inputs.configureWith(reward: value)
+  internal func configureWith(value: (project: Project, reward: Reward)) {
+    self.viewModel.inputs.configureWith(data: value)
+  }
+
+  // MARK: - Depress Transform
+
+  @objc func depress(_ gestureRecognizer: UILongPressGestureRecognizer) {
```
ifbarrera: `private`?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -130,12 +220,50 @@ final class PledgeDescriptionViewController: UIViewController {
...
+      case .changed:
+        return
+      case .began:
+        let scale = CheckoutConstants.RewardCard.Transition.DepressAnimation.scaleFactor
+        transform = CGAffineTransform(scaleX: scale, y: scale)
+      default:
+        transform = .identity
+      }
+
+      self.rewardCardContainerShadowView.transform = transform
```
ifbarrera: Can we use `[weak self]` here?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -168,6 +304,15 @@ private let descriptionStackViewStyle: StackViewStyle = { (stackView: UIStackVie
   stackView
     |> \.axis .~ NSLayoutConstraint.Axis.vertical
     |> \.distribution .~ UIStackView.Distribution.fill
+    |> \.spacing .~ Layout.Description.spacing
```
ifbarrera: Is there a reason this spacing is stored in an enum? Additionally, should we be using `Style.grid` units for this spacing?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -196,6 +341,26 @@ private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) ->
...
+}
+
+private let rewardCardContainerShadowViewStyle: ViewStyle = { (view: UIView) -> UIView in
+  view
+    |> rewardCardShadowStyle
+}
+
+private let rewardCardContainerViewStyle: ViewStyle = { (view: UIView) -> UIView in
+  view
+    |> \.translatesAutoresizingMaskIntoConstraints .~ false
```
ifbarrera: I've noticed that sometimes it's safer to set `. translatesAutoresizingMaskIntoConstraints .~ false` right in the lazy initializer for the view, so that you're guaranteed that this will be set when you configure & activate your constraints. Otherwise it's a little non-deterministic since this style is only applied in `bindStyles` which is called on `viewWillAppear` and `traitCollectionDidChange`.

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -210,3 +375,77 @@ private func attributedLearnMoreText() -> NSAttributedString? {
...
+}
+
+private func rewardCardThumbnailViewSize(
+  with cardContainerViewSize: CGSize,
+  parentWidth: CGFloat
+) -> CGSize {
+  let width = cardContainerViewSize.width
+  let height = cardContainerViewSize.height
+
+  let minWidth = CGFloat(100)
```
ifbarrera: Should this go in our `Layout` enum?

Kickstarter-iOS/Views/Transitions/RewardPledgeTransitionAnimator.swift
```diff
@@ -0,0 +1,298 @@
...
+    else {
+      // something went wrong, fall back to no transition
+      forceTransition(with: transitionContext, operation: operation)
+      return
+    }
+
+    _ = containerView
+      |> \.backgroundColor .~ fromView.backgroundColor
+
+    containerView.addSubview(toView)
```
ifbarrera: Can we use our helper functions `ksr_addSubviewtoParent`?

Library/SharedFunctions.swift
```diff
@@ -188,8 +188,33 @@ public func updatedUserWithClearedActivityCountProducer() -> SignalProducer<User
     .demoteErrors()
 }
 
-internal func classNameWithoutModule(_ klass: AnyClass) -> String {
-  return klass
+public func formattedAmountForRewardOrBacking(
```
ifbarrera: I'm not sure this should be a shared function, it seems pretty specific to the reward view 🤔

Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+  private let currentRewardProperty = MutableProperty<Reward?>(nil)
+  public func currentReward(is reward: Reward) -> Bool {
+    return self.currentRewardProperty.value == reward
+  }
+
+  public var inputs: RewardCardContainerViewModelInputs { return self }
+  public var outputs: RewardCardContainerViewModelOutputs { return self }
+}
+
+// MARK: - Private Helpers
```
ifbarrera: `Functions`? 😄

Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
ifbarrera: Or alternatively `RewardsCollectionNavigationController`

Library/SharedFunctions.swift
ifbarrera: Oh I see now why it's used in two places. I'm still not sure this `SharedFunctions` file is the right  place for it though 🤔 🤔

Kickstarter-iOS/Views/Controllers/CheckoutNavigationController.swift
justinswart: I think the idea was just that this is the navigation controller that will be used for the checkout flow and that it could potentially serve up other animators for different controllers.. not so?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
justinswart: Ah right thanks, just overhang from all the refactoring

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
justinswart: Hmm, don't think we need to - there shouldn't be a retain cycle because the animator is not being retained by the VC

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
justinswart: Good q's - I think in this view initially referred to the spacing in two places so I moved it to a constant. Also in that view it was just `10`. We can't really get to that on the grid spacing but I could make it .grid(2).

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
justinswart: Sounds good

Library/SharedFunctions.swift
justinswart: Suggestions?

Library/ViewModels/RewardCardContainerViewModel.swift
justinswart: lol another refactor left-over! :trollface:

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
ifbarrera: It's a little unclear from the naming what this spacing is for - let's remove it if it's only used here?

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -11,437 +11,134 @@ protocol RewardCellDelegate: AnyObject {
...
 
     self.configureViews()
-    self.setupConstraints()
-    self.bindViewModel()
 
-    self.pledgeButton.addTarget(self, action: #selector(self.pledgeButtonTapped), for: .touchUpInside)
-    self.containerView.addGestureRecognizer(
-      UITapGestureRecognizer(target: self, action: #selector(self.rewardCardTapped))
-    )
+    self.bindViewModel()
```
ifbarrera: Looks like we don't have a view model for this cell anymore, so we don't need this.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -11,437 +11,132 @@ protocol RewardCellDelegate: AnyObject {
...
   private let scrollView = UIScrollView(frame: .zero)
-  private let stateImageView: UIImageView = {
-    UIImageView(frame: .zero)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
-  }()
+  private lazy var longPressGestureRecognizer: UILongPressGestureRecognizer = {
+    let longPressGestureRecognizer = UILongPressGestureRecognizer(
+      target: self, action: #selector(RewardCell.depress(_:))
+    )
+    longPressGestureRecognizer.minimumPressDuration = CheckoutConstants.RewardCard.Transition
```
ifbarrera: Can we use key path setters for all these properties?

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -11,437 +11,132 @@ protocol RewardCellDelegate: AnyObject {
...
-    NSLayoutConstraint.activate([
-      containerConstraints,
-      baseStackViewConstraints,
-      pillCollectionViewConstraints,
-      pledgeButtonConstraints,
-      pledgeButtonLayoutGuideConstraints,
-      stateImageViewContraints
-    ].flatMap { $0 })
-  }
+    self.rewardCardContainerView.widthAnchor.constraint(equalTo: self.contentView.widthAnchor)
```
ifbarrera: Same here, key path setter for `\.isActive`

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -11,437 +11,132 @@ protocol RewardCellDelegate: AnyObject {
...
-    self.viewModel.inputs.pledgeButtonTapped()
+  func currentReward(is reward: Reward) -> Bool {
+    return self.rewardCardContainerView.currentReward(is: reward)
   }
 
-  @objc func rewardCardTapped() {
-    self.viewModel.inputs.rewardCardTapped()
+  func cancelDepress() {
+    self.longPressGestureRecognizer.isEnabled = false
+    self.longPressGestureRecognizer.isEnabled = true
```
ifbarrera: 🤔 🤔 🤔

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -1,45 +1,64 @@
...
   }()
 
-  private lazy var pledgeImageView: UIImageView = {
-    UIImageView(frame: .zero) |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  private lazy var learnMoreTextView: UITextView = { UITextView(frame: .zero) |> \.delegate .~ self }()
+  private lazy var longPressGestureRecognizer: UILongPressGestureRecognizer = {
+    let longPressGestureRecognizer = UILongPressGestureRecognizer(
+      target: self, action: #selector(PledgeDescriptionViewController.depress(_:))
+    )
+    longPressGestureRecognizer.minimumPressDuration = CheckoutConstants.RewardCard.Transition
```
ifbarrera: Can we set these properties using key path setters?

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -112,12 +170,47 @@ final class PledgeDescriptionViewController: UIViewController {
...
+  // MARK: - Actions
+
+  @objc private func rewardCardTapped() {
+    self.viewModel.inputs.rewardCardTapped()
+  }
+
+  // MARK: - Accessors
+
+  public func setThumbnailHidden(_ hidden: Bool) {
+    self.rewardCardContainerShadowView.alpha = hidden ? 0 : 1
```
ifbarrera: Key path setter?

Kickstarter-iOS/Views/Controllers/RewardPledgeNavigationController.swift
```diff
@@ -0,0 +1,29 @@
+import UIKit
+
+final class RewardPledgeNavigationController: UINavigationController {
+  // MARK: - Lifecycle
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.delegate = self
```
ifbarrera: `_ = self |> \.delegate .~ self`?

PR #744, opened 2019-07-08
💲[Native Checkout] Fix Pledge CTA for logged out users

ifbarrera:
# 📲 What

This PR fixes an issue where the `PledgeCTAContainerView` was incorrectly showing "You are a backer!" for logged out users, and hiding the `Back this project` button. This was happening because previously, the `PledgeCTAContainerView` was being configured with a `User`, and since the `User` is `nil` when a user is logged out, it wasn't being configured properly.

Luckily, we actually don't **need** the user object. Previously, it was being used to fetch the backing for a given project, but this network call was actually unnecessary, because the `Project` object has a `Personalization` sub-object that returns the backing if the request is authorized. I've updated it to now read from the `personalization` object instead of making a network request.

This PR also adds an additional check for `featureNativeCheckoutEnabled` for the configuration of the CTA view so we're not configuring it unnecessarily.

# 🤔 Why

The `PledgeCTAContainerView` wasn't behaving properly under logged-out conditions.

# 🛠 How

- updated `PledgeCTAContainerView` to use `project.personalization.backing` instead of making a network request for the backing
- fixed the logic for logged out users
- added filter for the `configurePledgeCTAView` output that checks if the native checkout feature flag is enabled

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - Clone 1 of iPhone 8 - 2019-07-08 at 18 37 25 |Simulator Screen Shot - iPhone 8 - 2019-07-08 at 18 32 31|

## Also see

| Project Ended State | Backed Project State | Backed Project, Project Ended State |
| --- | --- | --- |
|Simulator Screen Shot - iPhone 8 - 2019-07-08 at 18 32 24|Simulator Screen Shot - iPhone 8 - 2019-07-08 at 18 33 09|Simulator Screen Shot - iPhone 8 - 2019-07-08 at 18 33 16|

# ✅ Acceptance criteria

- [x] Logged out: use the debugging tool to toggle the native checkout feature flag to *ON*. Navigate to a **live project**. You should see the `Back this Project` button, and tapping it should take you to rewards.
- [x] Logged out: use the debugging tool to toggle the native checkout feature flag to *ON*. Navigate to  an **ended project**. You should see the `View Rewards` button, and tapping it should take you to rewards.


ifbarrera:
[Project page in logged out state shows that you're a backer](https://trello.com/c/0Btg5PIe/1446-project-page-in-logged-out-state-shows-that-youre-a-backer)

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,7 +76,8 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
     let project = freshProjectAndRefTag
       .map(first)
 
-    self.projectAndUser = Signal.combineLatest(project, user)
+    self.configurePledgeCTAView = project
+      .filter { _ in featureNativeCheckoutEnabled() == true }
```
dusi: Do we need the explicit use of `== true` here?

Library/ViewModels/ProjectPamphletViewModelTests.swift
```diff
@@ -373,4 +375,32 @@ final class ProjectPamphletViewModelTests: TestCase {
...
   }
+
+  func testConfigurePledgeCTAView_featureEnabled() {
+    let config = Config.template |> \.features .~ [Feature.checkout.rawValue: true]
+    let project = Project.template
+
+    withEnvironment(config: config) {
+      self.vm.inputs.configureWith(projectOrParam: .left(project), refTag: .discovery)
+      self.vm.inputs.viewDidLoad()
+      self.vm.inputs.viewWillAppear(animated: false)
```
dusi: Commenting out both `viewWillAppear` + `viewDidAppear` in both of these tests passes...do we still need it then?

ifbarrera: I think we usually consider it good practice to trigger all the lifecycle inputs so it reproduces the sequence that would usually happen in a real app run.

dusi: I see.

I think there's another HWDWWD candidate as I'm pretty sure this has been followed very loosely. I think keeping those calls here makes the test harder to read and actually factually incorrect (because it isn't dependent on those signals) but it does resemble the lifecycle closely so I'll stay undecided for now 😅

dusi: Just chatted with @justinswart and wonder if we can assert that `configurePledgeCTAView` hasn't emitted before calling any inputs? To make sure that it indeed is one of the inputs causing the signal to emit.

PR #747, opened 2019-07-10
💲[Native Checkout] Shipping Picker - Show List - 2/4

dusi:
# 📲 What

- Shows the list of shipping rules
- Adds the ability to dismiss the modal

# 🤔 Why

Showing the actual value and in addition to that adds the ability to dismiss the modal.

# 🛠 How

We're passing a formatted array of strings to the picker data source that shows it in a table view.

# 👀 See

(please note the following is localized)

Screen Shot 2019-07-09 at 5 17 15 PM

# ✅ Acceptance criteria

- [x] Tapping on shipping location button will show the list of all shipping rules
- [x] The list is localized
- [x] The list is formatted properly (currency & plus sign) - please refer to #667 & #691

# ⏰ TODO

- [ ] Add the ability to select a different shipping rule
- [ ] Add the ability to search shipping rules


Kickstarter-iOS/DataSources/ShippingRulesDataSource.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import Library
+import UIKit
+
+final class ShippingRulesDataSource: ValueCellDataSource {
+  // MARK: - Load
+
+  func load(_ values: [String]) {
```
dusi: We shouldn't really need the `ShippingRule` type as we can always find that during selection by the index of selected row (this will be done in the follow up work).

Kickstarter-iOS/Views/Controllers/PledgeTableViewController.swift
```diff
@@ -94,6 +94,18 @@ class PledgeTableViewController: UITableViewController {
       .observeValues { [weak self] project, pledgeTotal in
         self?.pledgeSummaryCell?.configureWith(value: (project, pledgeTotal))
       }
+
+    self.viewModel.outputs.dismissShippingRules
+      .observeForUI()
+      .observeValues { [weak self] in
+        guard
```
dusi: View controller that presented something should also be responsible for dismissing it (it shouldn't be the responsibility of the presented vc)

dusi: Yep. We could make this simpler and hold reference to the view controller but I thought this was less confusing than introducing another property...It's mostly just a best practice to explicitly say what we're dismissing coz otherwise it could lead to unexpected bugs.

justinswart: @dusi and I chatted IRL, we're going to revert this for consistency with the rest of the codebase for now and look to solve this with an agreed upon pattern going forward. One option is perhaps a dismiss closure that we can optionally set on our view controllers which we will hopefully have conforming to a protocol defining all of our programmatic UI functions in the near future.  That way at the time of presenting we can define how this VC should be dismissed.

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+public protocol ShippingRulesViewModelType {
+  var inputs: ShippingRulesViewModelInputs { get }
+  var outputs: ShippingRulesViewModelOutputs { get }
+}
+
+public final class ShippingRulesViewModel: ShippingRulesViewModelType,
+  ShippingRulesViewModelInputs, ShippingRulesViewModelOutputs {
+  public init() {
+    self.values = self.configDataProperty.signal
+      .takeWhen(self.viewDidLoadProperty.signal)
```
justinswart: So this is cool to do, just worth noting that in all the other areas we've used `combineLatest` because then we aren't dependent on the order in which these inputs are called. Using `takeWhen` just means that we _must_ call the `configureWith()` before `viewDidLoad()`. In this case it's fine, just wanted to point that out.

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+import ReactiveExtensions
+import ReactiveSwift
+
+public protocol ShippingRulesViewModelInputs {
+  func configureWith(_ project: Project, shippingRules: [ShippingRule], selectedShippingRule: ShippingRule)
+  func viewDidLoad()
+}
+
+public protocol ShippingRulesViewModelOutputs {
+  var values: Signal<[String], Never> { get }
```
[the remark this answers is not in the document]
dusi: We use various combinations thorough our codebase (i.e. `loadItem`, `loadItemsDataSource`, `items` where `item` could be `Project`, etc.)...I'd leave it like this for now and maybe we can talk about consistency with the rest of the team?

Library/ViewModels/ShippingRulesViewModel.swift
dusi: Oh, that's good to know! Maybe I can change this so that it doesn't bite us later?

Library/ViewModels/ShippingRulesViewModel.swift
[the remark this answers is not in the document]
ifbarrera: I think `values` is way too general, but I'll defer to @justinswart who is the assignee on this PR.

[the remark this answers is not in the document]
ifbarrera: I'd like to call out the `Naming Conventions` portion at the end of our [inputs and outputs documentation](https://github.com/kickstarter/native-docs/blob/master/inputs-outputs.md): 

```
Ideally, the view should be able to infer directly via variable name what 1. triggers the input action and 2. what an output should do.
```
Granted, this documentation is extremely out of date 😅 and worth a revisit, but I think it serves as a good guideline.

Library/ViewModels/ShippingRulesViewModel.swift
justinswart: It may be safer and more consistent to use `combineLatest`, yeah.

Library/ViewModels/ShippingRulesViewModel.swift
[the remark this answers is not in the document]
justinswart: I guess in this case `values` is no more ambiguous than say `pledgeViewData` 🤷‍♂. I think perhaps improving the naming to better-describe the side-effect is what we really want. So we should probably be calling it `loadValuesIntoDataSource` and `loadPledgeViewDataIntoDataSource` in the other sitch 🤔. Naming is hard!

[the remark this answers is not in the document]
justinswart: @dusi and I chatted IRL, there's been very little consistency in how we name these datasource-loading outputs, we went with `loadValues` for now and will look to establish a convention for these.

PR #753, opened 2019-07-15
💲[Native Checkout] Pledge Error State CTA

cdolm92:
# 📲 What

Presents the CTA for a pledge in an error state.

# 🤔 Why

This  PR **shows** backers the CTA to correct their pledge info if there is an error (card expired, etc).

# 🛠 How

The`PledgeCTAContainerView` displays the CTA depending on state. The state is determined in the `PledgeCTAContainerViewViewModel` by either of two functions that return a `PledgeStateCTAType`. `pledgeCTAButton(project: Project) -> PledgeStateCTAType` is called when there is no backing and
`pledgeCTA(project: Project, backing: Backing) -> PledgeStateCTAType` if there is a backing.

# 👀 See

Simulator Screen Shot - iPhone Xs - 2019-07-11 at 10 52 26

# ✅ Acceptance criteria

Because this PR only implements the error CTA UI (not functionality)  I would suggest modifying `private func pledgeCTA(project: Project, backing: Backing) -> PledgeStateCTAType`  in `PledgeCTAContainerViewViewModel` to view it.

```
// line 75.

private func pledgeCTA(project: Project, backing: Backing) -> PledgeStateCTAType {
  switch (project.state, backing.status) {
  case (.live, .errored):
    return .fix
  case (.live, _):
    return .fix
  case (_, _):
    return .viewBacking
  }
}
```
- [x] Run the app. Navigate to Profile and tap on a backed and live project
- [x] You should see the Error CTA
- [x] Check out Abstract to determine that the design matches

# ⏰ TODO

- [x] Design Feedback cc: @dannyalright 
- [x] Acceptance criteria testing.


Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -95,8 +105,15 @@ final class PledgeCTAContainerView: UIView {
   override func bindViewModel() {
     super.bindViewModel()
 
-    self.amountAndRewardTitleStackView.rac.hidden = self.vm.outputs.stackViewIsHidden
-    self.amountOrRewardLabel.rac.text = self.vm.outputs.rewardTitle
+    self.vm.outputs.buttonTitleTextColor
+      .observeForUI()
+      .observeValues { [weak self] textColor in
+        self?.pledgeCTAButton.setTitleColor(textColor, for: .normal)
```
justinswart: Just to note, by setting this we don't have any highlight state colours. Something I mentioned to @dannyalright is that we need highlight state colours for the buttons. This can probably just be like opacity `0.5` -> `0.8` for the textColor and a slightly darker colour for the button background which we can potentially just do programmatically.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -71,9 +71,10 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       }
 
     let project = freshProjectAndRefTag
-      .map(first)
+      .map { project, _ in project }
```
justinswart: did `map(first)` not work here?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -3,17 +3,24 @@ import Library
 import Prelude
 import UIKit
 
+private enum Layout {
+  enum Button {
+    static let height: CGFloat = 48.0
```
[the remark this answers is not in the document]
cdolm92: @justinswart chatted earlier about this. We are going to need a sweep to update all constant names to `minHeight` or `minWidth`  where necessary.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
cdolm92: chatted on slack w/ @justinswart about doing a future PR addressing this

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift
```diff
@@ -122,6 +122,38 @@ internal final class ProjectPamphletViewControllerTests: TestCase {
...
+    let currentUser = User.template
+    let backing = Backing.template
+      |> Backing.lens.status .~ .errored
+    let backedProject = Project.cosmicSurgery
+      |> Project.lens.photo.full .~ ""
+      |> Project.lens.personalization.isBacking .~ true
+      |> Project.lens.personalization.backing .~ backing
+      |> Project.lens.state .~ .live
+
+    [Device.phone4inch, Device.phone5_5inch, Device.phone5_8inch].forEach { device in
```
ifbarrera: Can we use `combos(Language.allLanguages, Device.allCases)` so it's consistent with the rest of the tests in this file? I know it generates tons of screenshots, but this is a high-value piece of UI that would be useful to have fully covered in all languages and screens sizes

dannyalright:
Noticing a slight difference in type size on the simulator. @cdolm92 could you please change the type style for _Check your payment details_ to `callout` `.bolded`? This should be the same for _Manage your pledge_.

| Simulator | Preferred |
| --- | --- |
| simulator | preferred |

PR #755, opened 2019-07-15
💲[Native Checkout] Prepare reward checkmark image view UI

dusi:
# 📲 What

Updates the view hierarchy of our `RewardCell` to include the checkmark `imageView`.

# 🤔 Why

Some of the reward states will have to show this checkmark (with different tint colors) so we would like to have this UI ready to go when we work on the business logic. Currently we only use one tint color (again, the business logic will handle this later).

# 🛠 How

Since this PR only deals with preparing the UI, it does not include any business logic of hiding/showing the checkmark. In order for us not to having to re-record screenshots all the time, I've decided to hide the `imageView` temporarily so if you want to see it render live you need to make the following change to the code

```swift
// In RewardCell.swift comment out `isHidden` property of the `imageViewContainer`
private let imageViewContainer: UIView = {
    UIView(frame: .zero)
//      |> \.isHidden .~ true
  }()
```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-07-15 at 3 47 14 PM | Screen Shot 2019-07-15 at 3 53 56 PM |

# ♿️ Accessibility 

- [x] The checkmark should not be accessible (we should handle this in the cell accessibility label)
- [x] The checkmark does not resize when other font sizes are applied

# ✅ Acceptance criteria

- [x] Uncommenting the code above shows the checkmark on the reward cell


Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -172,13 +193,20 @@ final class RewardCell: UICollectionViewCell, ValueCell {
     _ = (self.pledgeButtonLayoutGuide, self.containerView)
       |> ksr_addLayoutGuideToView()
 
-    _ = ([
-      self.priceStackView, self.rewardTitleLabel, self.includedItemsStackView,
-      self.descriptionStackView
-    ], self.baseStackView)
+    let baseSubviews = [
```
dusi: Extracted to local variable for readability

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewControllerTests.swift
```diff
@@ -31,7 +31,7 @@ final class RewardsCollectionViewControllerTests: TestCase {
         parent.view.frame.size.height = device == .pad ? 1_800 : 1_000
         parent.view.frame.size.width = device == .pad ? 2_300 : 1_800
 
-        FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)", tolerance: 0.01)
+        FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)")
```
dusi: @ifbarrera do you remember what was the problem originally when we decided to use `tolerance`?

For some reason using `tolerance` (and no matter what value) would pass this test even with the checkmark on so clearly that wasn't ideal.

I've re-recorded snapshots with checkmark hidden (seems like no difference at all)..and after removing tolerance tests pass for checkmark hidden, fail for checkmark shown - as it should (we don't want to show it for now - there's no business logic behind it).

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -23,6 +23,17 @@ final class RewardCell: UICollectionViewCell, ValueCell {
   private let descriptionLabel = UILabel(frame: .zero)
   private let descriptionStackView = UIStackView(frame: .zero)
   private let descriptionTitleLabel = UILabel(frame: .zero)
+  private let imageView: UIImageView = {
```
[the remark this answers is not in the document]
dusi: That’s exactly why I tried to keep it generic (because it’s either checkmark or exclamation mark) but let me try thinking about better name. 

Maybe `stateImageView`, `promptImageView`, `statusImageView` ¯\_(ツ)_/¯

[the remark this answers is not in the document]
cdolm92: Oh I see! Well the rename to `stateImageView` is cool. Thanks for making those changes!

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewControllerTests.swift
ifbarrera: I can't remember exactly. Feel free to remove!

PR #757, opened 2019-07-17
💲[Native Checkout] Prepare reward pill collection view UI

dusi:
# 📲 What

Updates the view hierarchy of our `RewardCell` to include the pill style `collection view`.

# 🤔 Why

Required UI for `time-based` reward and `limited` reward states.

# 🛠 How

Since this PR only deals with preparing the UI, it does not include any business logic of hiding/showing the pills. In order for us not to having to re-record screenshots all the time, I've decided to hide the `collectionView` temporarily so if you want to see it render live you need to make the following change to the code

```swift
// In RewardCell.swift make the following changes by commenting out `isHidden` prop
private lazy var pillCollectionView: UICollectionView = {
  UICollectionView(
    frame: .zero,
    collectionViewLayout: PillLayout(
      minimumInteritemSpacing: Styles.grid(1),
      minimumLineSpacing: Styles.grid(1),
      sectionInset: UIEdgeInsets(topBottom: Styles.grid(1))
    )
  )
    |> \.contentInsetAdjustmentBehavior .~ UIScrollView.ContentInsetAdjustmentBehavior.always
    |> \.dataSource .~ self.pillDataSource
    |> \.delegate .~ self
//    |> \.isHidden .~ true
}()
```

# 👀 See

| Before 🐛 | Regular | Extra extra large | Accessibility extra extra large |
| --- | --- | --- | --- |
| Screen Shot 2019-07-16 at 10 31 23 PM | Screen Shot 2019-07-16 at 10 29 56 PM | Screen Shot 2019-07-16 at 10 30 01 PM | Screen Shot 2019-07-16 at 10 30 11 PM |

# ♿️ Accessibility 

- [x] The pills try to fill in the row and if a pill can't fit it will be aligned on the next row

# ✅ Acceptance criteria

- [x] Uncommenting the code above shows `Ends in 3 days` & `16 left` pills on the reward cell


Kickstarter-iOS/Views/Layouts/PillLayout.swift
```diff
@@ -0,0 +1,54 @@
+import Prelude
+import UIKit
+
+class PillLayout: UICollectionViewFlowLayout {
```
dusi: First solution from the following answer https://stackoverflow.com/a/49717759

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -144,6 +158,8 @@ final class RewardCell: UICollectionViewCell, ValueCell {
         _ = self?.containerView
           ?|> \.isUserInteractionEnabled .~ isUserInteractionEnabled
       }
+
+    self.collectionViewController.configure(with: ["Ends in 3 days", "16 left"])
```
dusi: Temporarily hardcoded .. this value will be `output` of our `vm`.

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -272,6 +299,47 @@ final class RewardCell: UICollectionViewCell, ValueCell {
       |> ksr_addArrangedSubviewsToStackView()
   }
 
+  // MARK: - View controller containment
+
+  func addCollectionViewController(to parent: UIViewController) {
+    _ = self.collectionViewController
+      |> UIViewController.lens.view .. UIView.lens.isHidden .~ true
```
dusi: Temporarily hidden for the tests to show properly ... this will be set using an `output` of our `vm`

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -272,6 +299,47 @@ final class RewardCell: UICollectionViewCell, ValueCell {
...
+
+    parent.addChild(self.collectionViewController)
```
dusi: For reusability `PillCollectionViewController` is a view controller and in order for it to being properly added to view hierarchy and view-controller hierarchy (UIViewController containment) it has to be added / removed as a `child` to/from a `parent` (this is best practice).

Library/DataSource/UICollectionView-Extensions.swift
```diff
@@ -1,24 +1,27 @@
 import UIKit
 
 public extension UICollectionView {
+  // MARK: - Registration
+
   func registerCellClass<CellClass: UICollectionViewCell>(_ cellClass: CellClass.Type) {
-    self.register(cellClass, forCellWithReuseIdentifier: cellClass.description())
+    self.register(cellClass, forCellWithReuseIdentifier: classNameWithoutModule(cellClass))
```
dusi: Needed to use some of these convenience functions so in regards to the plan to merge feature branch soon I haven't made these changes against `master` first (only here).

Library/SharedFunctions.swift
```diff
@@ -187,3 +187,11 @@ public func updatedUserWithClearedActivityCountProducer() -> SignalProducer<User
     .skipNil()
     .demoteErrors()
 }
+
+internal func classNameWithoutModule(_ klass: AnyClass) -> String {
```
dusi: Should probably also test this..

Kickstarter-iOS/Views/Cells/RewardCell.swift
dusi: This will obviously not be needed when we migrate to `stack view`

Kickstarter-iOS/Views/Cells/PillCell.swift
```diff
@@ -0,0 +1,63 @@
...
+
+private let contentViewStyle: ViewStyle = { view in
+  view
+    |> checkoutRoundedCornersStyle
+    |> \.backgroundColor .~ UIColor.ksr_green_500.withAlphaComponent(0.06)
+}
+
+private let labelStyle: LabelStyle = { label in
+  label
+    |> \.font .~ UIFont.ksr_body(size: 13).bolded
```
ifbarrera: @dannyalright and I have been trying to stick to using Apple's default text styles, without size overrides. Maybe `ksr_footnote` would work here?

Kickstarter-iOS/Views/Cells/RewardCell.swift
ifbarrera: @dusi do we forsee needing to reuse this UI anywhere? It's seems unnecessary to me to use a view controller since this piece of UI is very static and has no user interaction requirements. I would imagine adding a `UICollectionView` to the `baseStackView` would be sufficient to meet the requirements of this feature.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -232,6 +232,16 @@ extension RewardsCollectionViewController {
   ) {
     if let rewardCell = cell as? RewardCell {
       rewardCell.delegate = self
+      rewardCell.addCollectionViewController(to: self)
```
ifbarrera: Why do you think it's best to configure addition of the VC in `RewardsCollectionViewController`? I think that it's the case that the pills only show in specific cases, so it might make more sense to have the `RewardViewModel` drive the creation & addition of the `PillCollectionViewController` via an output. Something like this:

**In RewardCell**
```
self.viewModel.outputs.addPillCollectionViewControllerWithData
  .observeForUI()
  .observeValues { data in
    self.addCollectionViewController()
    self.reloadCollectionViewControllerWith(data)
}
```

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -232,6 +232,16 @@ extension RewardsCollectionViewController {
...
+    }
+  }
+
+  override func collectionView(
+    _: UICollectionView, didEndDisplaying cell: UICollectionViewCell, forItemAt _: IndexPath
+  ) {
+    if let rewardCell = cell as? RewardCell {
+      rewardCell.delegate = nil
+      rewardCell.removeCollectionViewController()
```
ifbarrera: Can this action not be performed in the `RewardCell`'s `prepareForReuse` function?

Kickstarter-iOS/Views/Cells/PillCell.swift
dusi: Happy to change...just wanted to understand what was the rationale? The size override still scales properly...and supports Dynamic Type.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
dusi: It could, but considering the addition above, pairing these was more explicit of what's happening...it should also definitely happen in `prepareForReuse` so I'll also add it there.

Kickstarter-iOS/Views/Cells/PillCell.swift
ifbarrera: We're hoping to be able to get to a place sometime where we can remove our custom font initialization:

```
  let font = UIFont.preferredFont(forTextStyle: style)
  let descriptor = font.fontDescriptor
    return UIFont(
      descriptor: descriptor,
      size: ceil(font.pointSize / defaultSize * (size ?? defaultSize))
    )
```

... and just use `UIFont.preferredFont(forTextStyle: style)` out of the box for simplicity and 1:1 consistency with Apple's suggested sizes for a given text style.

Kickstarter-iOS/Views/Cells/RewardCell.swift
dusi: I'd be happy to do that.

My reasoning was that for something like this `pill layout` which is easily done with `flex box` but not really with auto layout it seemed like a good opportunity to introduce this as a reusable component.

Any thoughts @justinswart ? If we use a `collection view` directly we can get rid of the VC containment fiddling as well.

Kickstarter-iOS/Views/Cells/PillCell.swift
dannyalright: +1 to @ifbarrera's explanation. Additionally: designers can access Apple's text styles in Sketch. Engineers can see this text style via Abstract/Sketch. Just makes things more straightforward all-round.

dusi: ^^^
CC @cdolm92 @justinswart @garycanino

dannyalright: I learned about Apple's Sketch library recently, so parts of Checkout (including these pills) does not take advantage of it. For those cases (until they're updated by designover time), you can reference [Apple's text styles list](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/) to see what matches most closely to the UI mockup.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
dusi: Yeah that would work too...the problem is that for addition we need to parent VC in order for the containment to work properly...we'd have to store a `weak` reference to it in the cell.

What do you think about that?

Kickstarter-iOS/Views/Cells/PillCell.swift
dusi: The way other font sizes resize is unfortunately not identical to what `body` currently does so I'm not really sure which font would match the closest without explicitly specifying the size. Will do some more testing..

dusi: Alright, resolved this with Danny face to face

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -234,6 +234,14 @@ extension RewardsCollectionViewController {
       rewardCell.delegate = self
     }
   }
+
+  override func collectionView(
```
ifbarrera: Just curious, why is this necessary?

Kickstarter-iOS/Views/Cells/RewardCell.swift
```diff
@@ -219,28 +274,26 @@ final class RewardCell: UICollectionViewCell, ValueCell {
     _ = ([self.priceStackView, self.stateImageViewContainer], self.titleStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    self.setupConstraints()
-
-    self.pledgeButton.addTarget(self, action: #selector(self.pledgeButtonTapped), for: .touchUpInside)
-
-    let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(self.rewardCardTapped))
-    self.containerView.addGestureRecognizer(tapGestureRecognizer)
+    self.pillCollectionView.registerCellClass(PillCell.self)
```
ifbarrera: Because `PillCell` conforms to `ValueCell` you should be able to use `self.pillCollectionView.register(PillCell.self)`, which is what we use for other `ValueCell`- conforming cells. It's confusing that we have two and I think we could refactor these helpers so we only have one way to register cell that works both with `ValueCellDataSource` and without it, but I just wanted to flag that `ValueCellDataSource` relies on the cells being registered with `T.defaultReusableId` as the identifier.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
dusi: I wanted to make sure we remove the reference to the delegate when the cell gets dequeued. The reason I put it side by side `willDisplay` is for better readability but this could also go to `prepareForReuse`. It was entirely for better readability as we set the delegate when the cell gets displayed and `nil` it out when it leaves the screen.

I guess it's not really going to break anything (since the delegate is always `self` for any of these cells) in this use case but I wanted to use this properly...happy to remove.

ifbarrera: I'd say because all the cells have the same delegate there's no harm in leaving it - then you don't need to worry about re-adding it when the cell is configured.

PR #759, opened 2019-07-18
💲[Native Checkout] Prepare reward error image

dusi:
# 📲 What

Adds error state image resource.

# 🤔 Why

Reward cell requires this for its error state.

# 🛠 How

We actually have the image resource for this, just updated it's properties to use single scale in image assets manager.

# ✅ Acceptance criteria

- [ ] CI passes


Kickstarter-iOS/Assets.xcassets/icons/icon--alert.imageset/Contents.json
```diff
@@ -2,16 +2,7 @@
...
-      "scale" : "1x"
-    },
-    {
-      "idiom" : "universal",
-      "scale" : "2x"
-    },
-    {
-      "idiom" : "universal",
-      "scale" : "3x"
+      "filename" : "icon--alert.pdf"
```
dusi: We're now using `Single scale` properly.

Kickstarter-iOS/Views/Storyboards/SettingsAccountWarningCell.xib
```diff
@@ -20,13 +20,17 @@
...
-                            <rect key="frame" x="0.0" y="307.5" width="288" height="20.5"/>
+                            <rect key="frame" x="0.0" y="307.5" width="286" height="20.5"/>
                             <fontDescription key="fontDescription" type="system" pointSize="17"/>
                             <nil key="textColor"/>
                             <nil key="highlightedColor"/>
                         </label>
                         <imageView clearsContextBeforeDrawing="NO" userInteractionEnabled="NO" contentMode="scaleToFill" horizontalHuggingPriority="253" verticalHuggingPriority="253" horizontalCompressionResistancePriority="752" verticalCompressionResistancePriority="752" image="icon--alert" translatesAutoresizingMaskIntoConstraints="NO" id="W3R-uz-66d">
-                            <rect key="frame" x="306" y="306.5" width="22" height="22"/>
+                            <rect key="frame" x="304" y="305.5" width="24" height="24"/>
```
dusi: Since we're now using smaller image we have to tell the `image view` to explicitly size itself. Since it's using `.fill` `contentMode` it will adjust for the difference and the UI will be almost identical (I say almost because the image offset changes by 1 pixel - which should be fine but requires snapshot update).

PR #760, opened 2019-07-19
💲[Native Checkout] Pledge Screen Separators

ifbarrera:
# 📲 What

Adds separators to the description section and summary section.

# 🤔 Why

To visually distinguish between various sections in the pledge screen.

# 🛠 How

This PR adds two separator views: `descriptionSectionSeparator` and `summarySectionSeparator`, and also adds logical view array groupings that define which views should belong to each section. This makes it easier to reason about which views are logically grouped, and will come in handy later when we want to conditionally hide/show different sections. For example, if one wishes to hide the `pledgeSummarySection`, instead of creating two different output signals to hide the `summarySectionSeparator` and the `pledgeSummaryViewController.view`, one could do some thing like the following:

```
self.viewModel.outputs.hideSummarySection
  .observeForUI()
  .observeValues { isHidden in
      _ = self.summarySectionViews
        ||> \.isHidden .~ isHidden
  }
```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|Simulator Screen Shot - iPhone 5s - 2019-07-18 at 16 37 26|Simulator Screen Shot - iPhone 5s - 2019-07-19 at 13 22 23|

See also:

| Logged Out, No Shipping Rules | Logged Out, Shipping Rules |
| --- | --- |
| Simulator Screen Shot - iPhone X - 2019-07-19 at 13 26 22 | Simulator Screen Shot - iPhone X - 2019-07-19 at 13 26 32 |

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

**Logged in**

1. Log in to an account with the native checkout feature flag enabled.
2. Navigate to any project.
3. Select a reward without shipping enabled.
4. **You should the the pledge screen with two separators, without the shipping row, and they should match the designs.**
5. Navigate back to rewards and select a reward with shipping enabled.
6. **You should see the pledge screen with two separators and the shipping row, and they should match the designs.**

**Logged out**

1. When you are logged out, navigate to the feature flag tools and toggle the "Native checkout" feature flag to **ON**.
2. Navigate to any project and select a reward without shipping enabled.
3. **You should see the pledge screen with two separators and the continue button, and they should match the designs.**
4. Navigate back to rewards and select a reward with shipping enabled.
5. **You should see the pledge screen with two separators and the shipping row, and the continue row, and they should match the designs.**


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -15,21 +29,41 @@ final class PledgeViewController: UIViewController {
...
+  private lazy var pledgeInputsSectionViews = {
+    [self.pledgeAmountViewController.view, self.pledgeShippingLocationViewController.view]
+  }()
+
+  private lazy var pledgeLoginSectionViews = {
+    [self.pledgeContinueViewController.view]
+  }()
+
+  private lazy var pledgePaymentMethodsSectionViews = {
+    [self.pledgePaymentMethodsViewController.view]
```
justinswart: Are we doing arrays like this in anticipation of a separator potentially being added later on?

ifbarrera: Yes exactly, but also for consistency so that we carry the concept of "sections" as being arrays of views for every logical grouping in the screen.

ifbarrera:
> Looking good, couple comments and then I'm just wondering:
> 
> 1. Do we get much benefit in overall clarity by making these computed arrays `var`s and not just constructing them once in the function where we use them? I guess my question is - will we ever refer to them in multiple places?
> 2. I know we chatted in passing about the fact that separators may need to be hidden/shown later on when this view's functionality grows. Remind me how that will work?
> 3. There seems to be a constraint warning: Untitled Changing the priority of the constraint on line `142` of `AmountInputView.swift` to `.defaultHigh` appears to fix this.

1. The main reason for storing a reference to these sections at the VC level was so we can easily hide/show groupings of views (ie. sections) like in the example described in the *How* section of the PR description.

2. See the example in the *How* section. The separators will always be hidden/shown as part of their section group.

3. Would you mind drafting a small card for us to fix this constraint warning? 🙏

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
dusi: Just had a quick peek at the naming conventions here...would it make sense to simplify the naming by dropping `pledge` prefix? We already know we're working with `PledgeVC` so it could simplify a lot of naming here?

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -98,6 +145,13 @@ final class PledgeViewController: UIViewController {
     NSLayoutConstraint.activate([
       self.rootStackView.widthAnchor.constraint(equalTo: self.rootScrollView.widthAnchor)
     ])
+
+    self.sectionSeparatorViews.forEach { view in
+      _ = view.heightAnchor.constraint(equalToConstant: 1)
```
dusi: Did the design expect `1` or more hair like separators (which could usually be achieved by using `0.5` on retina devices (`1/UIDevice.screen.scale`)?

Library/Styles/BaseStyles.swift
```diff
@@ -118,6 +118,12 @@ public let separatorStyle: ViewStyle = { (view: UIView) in
     |> \.accessibilityElementsHidden .~ true
 }
 
+public let separatorStyleDark: ViewStyle = { view in
```
dusi: Do we also have `Light` style?

ifbarrera: We do - the one above is called `separatorStyle` which essentially the light style. @dannyalright mentioned that he uses these two styles for separators when designing so I figured it made sense to add it to our `BaseStyles`. I could also refactor & rename `separatorStyle` to `separatorLightStyle` but that felt out of the scope of this PR.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
ifbarrera: The design specified 1 point for these separators.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
ifbarrera: Open to doing that. Thoughts @justinswart ?

justinswart: Hmm, I know it's a bit verbose but I typically like the variable name to reflect the type name.

I'm more interested in these potentially not being lazy vars and constructed where we use them?

justinswart: oh I see what @dusi meant, yeah for things like `pledgePaymentMethodsSectionViews` that could just be `paymentMethodsSectionViews`

justinswart: ah i'm not using github properly, i saw your comment below now about why we store a reference 👍

ifbarrera:
[[Checkout] Section separator](https://trello.com/c/SjubmRvB/1300-checkout-section-separator)

PR #761, opened 2019-07-22
Update Fastlane

justinswart:
# 📲 What

Updates Fastlane and its ruby gems to silence a warning that we're seeing on our repo:

image

# 🤔 Why

Although this doesn't affect us too much because ours is not a rails app, it's still not pleasant to see this warning each time we land on the repo page.

# 🛠 How

Ran `bundle update`.

# ✅ Acceptance criteria

Log in to CircleCI, find the `build-and-cache` job for this branch, re-run it with SSH enabled. Then SSH in and test the following:

- [x] `bundle exec fastlane match_all` (will only work once per session because of the keychain stuff it tries to do, if you need to run it again you have to comment some stuff out in `Fastfile`).
- [x] `bundle exec fastlane itunes_gym` (must be run after `match_all`)
- [x] swiftlint should not be a newer version than the one we use.


justinswart:
[Update fastlane](https://trello.com/c/wMEkEus4/1466-update-fastlane)

PR #764, opened 2019-07-22
Feature native checkout reward pledge states

cdolm92:
# 📲 What

Business logic to determine state of reward cell. 

# 🤔 Why

The reward cell has several states depending on whether the reward has a time limit or is limited,  the project is backed by the user, if the backing errors or if the project is live.

# 🛠 How

To handle these states we have the `RewardCellProjectBackingState` and `RewardState` enums. Each enum has a  state function. `RewardCellProjectBackingState.state(with project: Project, reward: Reward)` determines the `RewardCellProjectBackingState` and the `RewardCellProjectBackingState.RewardState.state(with: reward, project: project)` determines the `RewardState`.


Library/RewardStateType.swift
```diff
@@ -0,0 +1,61 @@
...
+      return .backedError(rewardState: RewardState.state(with: reward, project: project))
+    case(.live, _):
+      return .backed(live: .live, rewardState: RewardState.state(with: reward, project: project))
+    case (_, _):
+      return .backed(live: .nonlive, rewardState: RewardState.state(with: reward, project: project))
+    }
+  }
+
+  public enum RewardState {
+    case limited
```
ifbarrera: What would the `RewardState` be for a reward that is neither limited, nor time based? Can we add a test to see what the state function would return with a reward of type `Reward.noReward`?

cdolm92: A reward that is neither limited, nor time based  would be `inactive`

Library/RewardStateTests.swift
```diff
@@ -168,4 +168,59 @@ class RewardStateTests: TestCase {
       RewardCellProjectBackingState.state(with: project, reward: reward)
     )
   }
+
+  func test_noRewardType_BackedNonLive() {
+    let reward = Reward.noReward
+      |> Reward.lens.remaining .~ 0
```
ifbarrera: I think we shouldn't modify these properties on the `.noReward` template, because the template already mimics what gets returned from the server 🤔.

Library/RewardStateTests.swift
```diff
@@ -168,4 +168,59 @@ class RewardStateTests: TestCase {
...
+      |> Reward.lens.limit .~ 0
+
+    let backing = Backing.template
+      |> Backing.lens.reward .~ reward
+
+    let project = .template
+      |> Project.lens.personalization.isBacking .~ true
+      |> Project.lens.personalization.backing .~ backing
+      |> Project.lens.state .~ .successful
```
ifbarrera: Shouldn't this be `.live` since this test is testing for `noRewardType_backedLive`?

Library/RewardStateTests.swift
```diff
@@ -168,4 +168,59 @@ class RewardStateTests: TestCase {
...
+      .backed(live: .nonlive, rewardState: .inactive),
+      RewardCellProjectBackingState.state(with: project, reward: reward)
+    )
+  }
+
+  func test_noRewardType_NonBacked() {
+    let reward = Reward.noReward
+      |> Reward.lens.remaining .~ 0
+      |> Reward.lens.limit .~ 0
+      |> Reward.lens.endsAt
```
ifbarrera: Afaik "no reward" rewards should never be time based.

Library/RewardStateTests.swift
cdolm92: No, it's correct. Forgot to rename.

PR #767, opened 2019-07-24
🐞 [Bug] NSAttributedString.DocumentType.html main thread crash

justinswart:
# 📲 What

Attempts to fix an escalating crash that we've been seeing in our crash reports which appear to be linked to parsing HTML strings using `NSAttributedString.DocumentType.html` outside of the main thread. Although I have not been able to reproduce this crash myself, this is my suggested fix which I'd like to ship with the next release to see if it brings the reported crashes down.

# 🤔 Why

This crash has been rising in our crash reports (link below).

# 🛠 How

Ensure that functions in `String+SimpleHTML.swift` always execute on the main thread.

# 👀 See

https://fabric.io/kickstarter2/ios/apps/com.kickstarter.kickstarter/issues/5c19e0ddf8b88c2963607faa?time=last-seven-days
https://stackoverflow.com/questions/28915954/nsattributedstring-initwithdata-and-nshtmltextdocumenttype-crash-if-not-on-main

# ✅ Acceptance criteria

- [ ] No regression or noticeable performance issues introduced by this change.


justinswart:
[Fix escalating NSAttributedString crash](https://trello.com/c/m29EQzIs/1443-fix-escalating-nsattributedstring-crash)

Library/String+SimpleHTML.swift
```diff
@@ -21,44 +21,55 @@ public extension String {
...
-      if htmlFont.fontDescriptor.symbolicTraits.contains(.traitBold) {
-        newAttributes = bold
-      } else if htmlFont.fontDescriptor.symbolicTraits.contains(.traitItalic) {
-        newAttributes = italic
-      } else {
-        newAttributes = base
+    if Thread.isMainThread {
+      return parsedHtml()
+    } else {
+      return DispatchQueue.main.sync {
```
ifbarrera: Shouldn't this be `async`? 🤔

Library/String+SimpleHTML.swift
```diff
@@ -93,19 +104,29 @@ public extension String {
...
+      if trimWhitespace {
+        return result?.trimmingCharacters(in: .whitespacesAndNewlines)
+      }
+      return result
+    }
+
+    if Thread.isMainThread {
+      return parsedHtml()
+    } else {
+      return DispatchQueue.main.sync {
```
ifbarrera: Same here, should this be `async`?

Library/String+SimpleHTML.swift
justinswart: We actually do want this to run synchronously on the main thread. The difference between `async` and `sync` here is essentially that the code after the closure won't be executed until the closure returns, this is also why we're able to just do return `return DispatchQueue.main.sync { ... }` as it's synchronously returning a value, `async` would asynchronously execute the closure and return early.

justinswart: I think a better way to say it is that it blocks the current thread, we also first check that whether we're on the main thread because it will crash if we try to dispatch it in this way if we're already on that thread.

Library/String+SimpleHTML.swift
justinswart: No, `sync` is correct as per other answer.

Library/String+SimpleHTML.swift
ifbarrera: Ah okay gotcha 👍

PR #768, opened 2019-07-24
🐞 [Bug] UIPageViewController crash fix

justinswart:
# 📲 What

Attempts to fix an escalating crash that we've been seeing in our crash reports which appear to be linked to `setViewControllers` being called during a page transition. Although I have not been able to reproduce this crash myself, this is my suggested fix which I'd like to ship with the next release to see if it brings the reported crashes down.

# 🤔 Why

This crash has been rising in our crash reports (link below).

# 🛠 How

- Added a `ksr_setViewControllers` extention to `UIPageViewController` which ensures this is always executed on the main thread.

# 👀 See

https://fabric.io/kickstarter2/ios/apps/com.kickstarter.kickstarter/issues/5b8409046007d59fcdf6275f?time=last-seven-days
https://stackoverflow.com/questions/25740245/assertion-failure-in-uipageviewcontroller

# ✅ Acceptance criteria

- [ ] No regression on current functionality in the 3 affected VCs.


justinswart:
[Fix escalating page view controller bug](https://trello.com/c/3bwsg75G/1442-fix-escalating-page-view-controller-bug)

PR #769, opened 2019-07-24
💲[Native Checkout] Sheet Overlay iPad Bugfix

ifbarrera:
# 📲 What

Fixes an issue with using the sheet overlay modal on iPad.

# 🤔 Why

Because we need to be able to show overlays on iPad.

# 🛠 How

Since iPad has a different presentation style than compact devices, the transition calculations (as well as the sizing of the child view controller's view) were off. After chatting with @dannyalright as to the best approach for displaying overlays on iPad (given that the checkout flow is already presented as a `.pageSheet` overlay), we decided that instead of using the custom overlay & transition, we should present overlays using the default `.formSheet` presentation style.

In order to facilitate this logic, I've created a helper function that encompasses the behavior for iPad and other devices.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen_Shot_2019-07-05_at_9 45 07_AM| Simulator Screen Shot - iPad Air 2 - 2019-07-23 at 15 40 25|

# ✅ Acceptance criteria

**Login Modal**
On an iPad device,
Logged out, navigate to the feature flag debug menu and toggle the feature native checkout feature flag *ON*,
Navigate to any project, select any reward,
Tap `Continue` on the pledge screen,
You should see the login modal displayed in the "form sheet" presentation (centered in the middle with a transparent black background).

**Shipping Modal**
On an iPad device,
Navigate to the feature flag debug menu and toggle the feature native checkout feature flag *ON*,
Navigate to any project, select a reward with a shippable item,
Tap `United States` on the shipping view in the pledge screen,
You should see the shipping selection modal displayed in the "form sheet" presentation (centered in the middle with a transparent black background).

**Checkout Modal**
On an iPad device,
Navigate to the feature flag debug menu and toggle the feature native checkout feature flag *ON*,
Navigate to any project, tap `Back this project`,
You should see the reward carousel displayed in the "page sheet" presentation style (with a back background peeking out from the top, and the view having rounded corners).

**Webview Modals**
On an iPad device,
Navigate to the feature flag debug menu and toggle the feature native checkout feature flag *ON*,
Navigate to any project, select any reward,
Tap any of the links in the pledge screen (trust and safety, accountability, etc),
You should see the info webview links displayed in the "page sheet" presentation style (with a back background peeking out from the top, and the view having rounded corners).


Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
```diff
@@ -70,12 +70,10 @@ final class PledgeContinueViewController: UIViewController {
 
   private func goToLoginSignup(with intent: LoginIntent) {
     let loginSignupViewController = LoginToutViewController.configuredWith(loginIntent: intent)
+
     let navigationController = UINavigationController(rootViewController: loginSignupViewController)
-    let sheetOverlayViewController = SheetOverlayViewController(
-      child: navigationController,
-      offset: Layout.Sheet.offset
```
ifbarrera: I realized after doing a bunch of testing that we can't really use the same offset for every usage of the sheet overlay. This offset was way too big for the login modal, and it's also too big for landscape mode. We should address this issue separately, but for the shipping selection modal we'll need to be a bit smarter about how we define the offset.

dannyalright:
thanks for pairing! Looks good.

ifbarrera:
[Sheet Overlay Modal does not present content properly on iPad](https://trello.com/c/ebSUot2x/1454-sheet-overlay-modal-does-not-present-content-properly-on-ipad)

Kickstarter-iOS/Library/UIViewController+Presentation.swift
```diff
@@ -1,11 +1,36 @@
...
     let nc = UINavigationController(rootViewController: vc)
+    nc.modalPresentationStyle = presentationStyle
+
     self.present(nc, animated: true)
   }
+
+  /* A helper for presenting a view controller using the sheet overlay,
+    while also handling iPad behavior
+  */
+  internal func presentViewControllerWithSheetOverlay(_ viewController: UIViewController,
```
justinswart: Slight reshuffle suggestion for one `self.present(_:animated:)` call.
```suggestion
  internal func presentViewControllerWithSheetOverlay(
    _ viewController: UIViewController,
    offset: CGFloat
  ) {
    let vc: UIViewController

    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
      vc = viewController
        |> \.modalPresentationStyle .~ .formSheet
        |> \.modalTransitionStyle .~ .crossDissolve
    } else {
      vc = SheetOverlayViewController(
        child: viewController,
        offset: offset
      )
    }

    self.present(vc, animated: true)
  }
```

justinswart:
@dannyalright was chatting to @ifbarrera on Slack but I think the web links should open in the smaller modal:

| Current  | Suggested |
| image | image |

PR #770, opened 2019-07-24
💲[Native Checkout] Login/Signup design pass

ifbarrera:
# 📲 What

Updated the designs of the signup/login screen. 

This included:
- updated button colors/styles
- updated fonts
- dynamically setting the font of the context label based on whether it's shown with a header or not

As part of this work, I converted `LoginToutViewController.swift` to use programmatic UI, and deleted the corresponding scene from the `Login.storyboard`.

# 🛠 How

- removed the `LoginTout` scene from the `Login` storyboard and converted all relevant UI to programmatic views
- updated the necessary styles both in `ButtonStyles.swift` and in `Styles.swift`
- regenerated all the necessary screenshots

Note that because a lot of these button styles are shared, it resulted in changes to several login-related views including the sign up VC, login VC, forgot password VC, facebook signup confirmation VC and rewards collection VC. These changes are purely visual and should not impact the functionality of these views. Please examine updated screenshots carefully :)

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone Xs Max - 2019-07-24 at 17 28 27 |Simulator Screen Shot - iPhone X - 2019-07-24 at 17 30 11 |

Also see:

| LoginViewController | SignUpViewController | FacebookConfirmationViewController | 
| --- | --- | --- |
|testView_lang_en_device_phone4_7inch@2x|testView_lang_en_device_phone4_7inch@2x|testView_lang_en_device_phone4_7inch@2x|

# ♿️ Accessibility 

# ✅ Acceptance criteria

Please see all updated screenshots.

**Testing login/signup screen:**

Logged out, tap "Signup or login" from the discovery screen.
You you should see the login/signup screen with the new button/font designs.
Tap "Sign up with email" and you should be taken to the sign up screen.
Tap "Log in" and you should be taken to the login screen.
Tap "Log in with Facebook" and you should be prompted to accept the facebook login terms.
Tap the disclaimer text below the "Log in" button and you should see an action sheet with various privacy/terms disclaimer options.


Kickstarter-iOS/Views/Storyboards/Login.storyboard
```diff
@@ -317,7 +157,10 @@
                                                                 </constraints>
                                                             </view>
                                                             <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="JvF-DM-f0p">
-                                                                <rect key="frame" x="0.0" y="524" width="360" height="33"/>
+                                                                <rect key="frame" x="0.0" y="655.5" width="360" height="48"/>
+                                                                <constraints>
+                                                                    <constraint firstAttribute="height" relation="greaterThanOrEqual" constant="48" id="HTg-gw-BIr"/>
```
ifbarrera: I had to add height constraints to all the buttons in these views because we removed the default `contentEdgeInsets` that the old `baseStyles` had, which were basically giving these buttons their "height". Now they have a real height consistent with button heights in the new checkout. @dannyalright lmk if 48 points sounds good or we should use the default 44 points instead.

Library/Styles/ButtonStyles.swift
```diff
@@ -19,6 +19,17 @@ public let baseButtonStyle =
...
   <> UIButton.lens.adjustsImageWhenHighlighted .~ false
 
+// This is a temporary style meant to replace baseButtonStyle in the very near future
+public let baseNewButtonStyle: ButtonStyle = { button in
+  button
+    |> roundedButtonStyle
+    |> UIButton.lens.titleLabel.font .~ UIFont.ksr_callout().bolded
+    |> (UIButton.lens.titleLabel .. UILabel.lens.textAlignment) .~ NSTextAlignment.center
+    |> UIButton.lens.titleEdgeInsets .~ .init(all: Styles.grid(2))
+    |> UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode .~ NSLineBreakMode.byTruncatingMiddle
```
ifbarrera: Buttons will always by default truncate in the middle when the text is too long. cc @dannyalright

Library/Styles/ButtonStyles.swift
```diff
@@ -81,6 +92,13 @@ public let greenButtonStyle =
   <> UIButton.lens.layer.borderColor .~ UIColor.ksr_green_700.cgColor
   <> UIButton.lens.layer.borderWidth .~ 1.0
 
+public let roundedGreenButtonStyle: ButtonStyle = { button in
```
ifbarrera: We should eventually replace `greenButtonStyle` with the style from `roundedGreenButtonStyle` (because all buttons will be rounded in the future. For now, kept them separate.

ifbarrera:
[Design updates for signup/login screen](https://trello.com/c/QrhNtRXW/1445-design-updates-for-signup-login-screen)

Kickstarter-iOS/Views/Storyboards/Login.storyboard
dannyalright: I'm a fan of the 48pt!

Library/Styles/ButtonStyles.swift
dannyalright: A string like `Pledge US$ 124 or more` would be truncated to something like `Pledge US... more`. Can we have an exception for buttons like that? Ones that have the core information in the centre of the string?

I agree with middle-of-the-string truncation otherwise.

Library/Styles/ButtonStyles.swift
dannyalright: Noted in our upcoming button sweep.

Library/Styles/ButtonStyles.swift
ifbarrera: Yep we can have an exception for the certain buttons. I'll update the truncation on the pledge button to be `.byTruncatingTail` 👍

Kickstarter-iOS/Views/RewardCardContainerView.swift
```diff
@@ -48,10 +48,8 @@ public final class RewardCardContainerView: UIView {
       |> \.layoutMargins .~ .init(all: Styles.grid(3))
 
     _ = self.pledgeButton
-      |> checkoutGreenButtonStyle
-
-    _ = self.pledgeButton.titleLabel
-      ?|> checkoutGreenButtonTitleLabelStyle
+      |> roundedGreenButtonStyle
+      |> UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode .~ .byTruncatingTail
```
Scollaco: Here, to avoid the use of lens and to keep consistent with the styling code on `LoginTouViewController`, we can use:
```
_ = self.pledgeButton.titleLabel
      ?|> \.lineBreakMode .~ .byTruncatingTail
```

PR #772, opened 2019-07-26
[Pledge Payment Methods] Replace UICollectionView with UIScrollView

justinswart:
# 📲 What

This is my suggested fix for the problematic self-sizing `UICollectionView` that we've been struggling with in the `PledgePaymentMethodsViewController`. This is not a view that we expect to show a large amount of cells so I don't think we need to be concerned about view recycling. My suggestion is to replace the `UICollectionView` with a `UIScrollView`.

# 🤔 Why

We've been fiddling with the self-sizing `UICollectionView` issue for 2 days and I don't think it's worth our time anymore. This works _very_ reliably and has _zero_ constraint warnings, even when resizing with dynamic type.

# 🛠 How

Replace the `UICollectionView` with a `UIScrollView`, add the card views to a horizontal `UIStackView` and enjoy our lives.

# 👀 See

| Size 1 | Size 2 | Size 3 |
| --- | --- | --- |
| image | image | image |


Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -59,32 +58,17 @@ final class PledgeCreditCardCell: UICollectionViewCell, ValueCell {
...
 
-    let minTouchSizeConstraint = self.selectButton.heightAnchor
-      .constraint(greaterThanOrEqualToConstant: Styles.minTouchSize.height)
-    _ = minTouchSizeConstraint
-      |> \.priority .~ .defaultHigh
-
-    let imageViewWidthConstraint = self.imageView.widthAnchor
-      .constraint(equalToConstant: Layout.ImageView.width)
-    _ = imageViewWidthConstraint
-      |> \.priority .~ .defaultHigh
```
justinswart: None of these priority settings were needed for this approach.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -124,13 +114,25 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
 
-  private func updateCollectionViewConstraints() {
-    self.collectionView.layoutIfNeeded()
+    _ = (cardViews, self.cardsStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+  }
 
-    self.collectionViewHeightConstraint.constant = self.collectionView.contentSize.height
+  private func updateScrollViewHeightConstraint() {
+    self.scrollViewHeightConstraint.constant = self.scrollView.contentSize.height
```
justinswart: So simple.

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -29,17 +29,16 @@ final class PledgeCreditCardCell: UICollectionViewCell, ValueCell {
...
+  private let selectButton: UIButton = { UIButton(type: .custom) }()
 
   // MARK: - Lifecycle
 
   override init(frame: CGRect) {
     super.init(frame: frame)
-    self.configureSubviews()
+
     self.bindViewModel()
+    self.configureSubviews()
```
ifbarrera: 🙏 ty!

Kickstarter-iOS/Views/RewardCardView.swift
```diff
@@ -228,12 +228,16 @@ public final class RewardCardView: UIView {
   }
 
   private func setupConstraints() {
+    let stateImageViewContainerWidthConstraint = self.stateImageViewContainer.widthAnchor
```
ifbarrera: Is this change unrelated?

Kickstarter-iOS/Views/RewardCardView.swift
```diff
@@ -228,12 +228,16 @@ public final class RewardCardView: UIView {
...
+      .constraint(equalToConstant: Styles.grid(5))
+    stateImageViewContainerWidthConstraint.priority = .defaultHigh
```
Scollaco: To be consistent with how we've set priority, could you use this instead?
```
let stateImageViewContainerWidthConstraint = self.stateImageViewContainer.widthAnchor
      .constraint(equalToConstant: Styles.grid(5))
|> \.priority .~ .defaultHigh
```

justinswart: 😄 i was gonna merge it and make that comment on your PR!

j/k i'll update it.

justinswart:
> Tested this out on my iPhone X and it seems ok... I'm still a bit wary of how it will perform on older devices for users with > 10 cards, but given some of the data we've pulled on this I'd be open to giving it a try.

Sure, I don't think anybody's suggesting this is the ultimate long-term solution, it's more just to get us unstuck on this for now. We can certainly iterate on bringing the collection view back but this allows us to have something that _works_ in the interim.

Kickstarter-iOS/Views/RewardCardView.swift
justinswart: It fixes a constraint warning

PR #773, opened 2019-07-29
💲[Native Checkout] Activity Indicator for the Pledge View

ifbarrera:
# 📲 What

Adds an activity indicator to indicate a loading state to the user while the full project is being fetched from the server.

# 🤔 Why

We can't determine the pledge CTA view's state until we have the full project, so instead of showing a blank view we show an activity indicator commonly used to signify content that is being loaded. Once the full project has fetched from the server, we hide the activity indicator and fade in the pledge CTA view with its correct state.

# 🛠 How

Since the project is fetched from the `ProjectPamphletViewModel`, we need to pass the loading info to the `PledgeCTAView` through it's configuration function. This PR extends the configuration `value` passed into the `PledgeCTAView` into a tuple containing a `Project` object and an `isLoading` `Bool` to communicate the loading state of the project.

The `PledgeCTAView` uses this `isLoading` property to hide/show the activity indicator and the `rootStackView` which contains all the subviews of the `PledgeCTAView`. Only once the project is fully loaded (`isLoading == false`) does the `PledgeCTAView` calculate the state of it's subviews and display them.

Note that if the `fetchProject` endpoint fails (due to server error or poor network access), the default `Project` value is used - this means that the default state of the `PledgeCTAView` in case of an error will be the green `Back this Project` button.

# 👀 See

PFikwcYzfJ

# ✅ Acceptance criteria

Use the Network Link Conditioner or Charles Proxy to simulate slower connections. Then do the following:

As a user with the `Native Checkout` feature flag turned **ON**;
Navigate to any project on the discovery screen;
You should see the activity indicator animating before seeing the `Back this project` button fade in.
Tap `Back this project`;
Navigate back to the project screen;
You should see the activity indicator animating before seeing the `Back this project` button fade in again.

As a user with the `Native Checkout` feature flag turned **ON**;
Navigate to any backed project in the backer dashboard;
Tap a backed project;
You should see the activity indicator animating before seeing the `Manage` button fade in.
Tap `Manage`;
Navigate back to the project screen;
You should see the activity indicator animating before seeing the `Manage` button fade in again.

# ⏰ TODO

- [ ] loading announcement copy for Voice Over

### DIFF
--- a/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
+++ b/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
@@ -108,6 +108,16 @@ final class PledgeCTAContainerView: UIView {
         self?.pledgeCTAButton.setTitleColor(textColor, for: .normal)
       }
 
+    self.viewModel.outputs.rootStackViewAnimateIsHidden
+      .observeValues { [weak self] isHidden in
+        let duration = isHidden ? 0.0 : 0.3
+        let alpha: CGFloat = isHidden ? 0.0 : 1.0
+        UIView.animate(withDuration: duration, animations: {
+          _ = self?.rootStackView
+            ?|> \.alpha .~ alpha
+        })
+    }
+
     self.activityIndicator.rac.animating = self.viewModel.outputs.activityIndicatorIsAnimating
     self.titleAndSubtitleStackView.rac.hidden = self.viewModel.outputs.stackViewIsHidden
     self.titleLabel.rac.text = self.viewModel.outputs.titleText
--- a/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
+++ b/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
@@ -93,7 +93,8 @@ final class PledgeCTAContainerView: UIView {
       |> \.numberOfLines .~ 0
 
     _ = self.activityIndicator
-      |> \.color .~ UIColor.ksr_grey_500
+      |> \.color .~ UIColor.ksr_dark_grey_500
+      |> \.hidesWhenStopped .~ true
   }
 
   // MARK: - View model
@@ -118,7 +119,7 @@ final class PledgeCTAContainerView: UIView {
 
   // MARK: - Configuration
 
-  func configureWith(value: (project: Project?, isLoading: Bool)) {
+  func configureWith(value: (project: Project, isLoading: Bool)) {
     self.viewModel.inputs.configureWith(value: value)
   }
 
--- a/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
+++ b/Kickstarter-iOS/Views/PledgeCTAContainerView.swift
@@ -8,6 +8,10 @@ private enum Layout {
     static let minHeight: CGFloat = 48.0
     static let minWidth: CGFloat = 98.0
   }
+
+  enum ActivityIndicator {
+    static let height: CGFloat = 30
+  }
 }
 
 final class PledgeCTAContainerView: UIView {
@@ -116,13 +120,6 @@ final class PledgeCTAContainerView: UIView {
         })
       }
 
-    self.viewModel.outputs.activityIndicatorIsAnimating
-      .observeValues { isAnimating in
-        let announcementMessage = isAnimating ? "Loading project state" : "Loading project state complete"
-
-        UIAccessibility.post(notification: .announcement, argument: announcementMessage)
-      }
-
     self.activityIndicator.rac.animating = self.viewModel.outputs.activityIndicatorIsAnimating
     self.titleAndSubtitleStackView.rac.hidden = self.viewModel.outputs.stackViewIsHidden
     self.titleLabel.rac.text = self.viewModel.outputs.titleText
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -98,31 +89,81 @@ final class PledgeCTAContainerView: UIView {
...
+          _ = self?.rootStackView
+            ?|> \.alpha .~ alpha
+        })
+      }
+
+    self.viewModel.outputs.activityIndicatorIsAnimating
+      .observeValues { isAnimating in
+        let announcementMessage = isAnimating ? "Loading project state" : "Loading project state complete"
+
+        UIAccessibility.post(notification: .announcement, argument: announcementMessage)
```
ifbarrera: @jamielynnroth @dannyalright need some guidance here. What this code does is it posts a notification for users with Voice Over enabled, such that if the activity indicator is loading, it will say "Loading project state". If the activity indicator is not loading (ie. the request has completed) it will say "Loading project state complete".

When testing this, I noticed that the requests are coming back so quickly that only the "Loading project state complete" announcement is actually heard. I'm wondering if this is useful at all then? I imagine it maybe being useful for Voice Over users under bad network conditions, but if it's a very small edge case it might not be worth having this at all. Thoughts?

ifbarrera:
[Indicate loading of rewards in Pledge CTA](https://trello.com/c/hT4roOHp/1471-indicate-loading-of-rewards-in-pledge-cta)

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -98,31 +89,81 @@ final class PledgeCTAContainerView: UIView {
...
+    _ = ([self.titleLabel, self.subtitleLabel], self.titleAndSubtitleStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+
+    _ = ([self.titleAndSubtitleStackView, self.spacer, self.pledgeCTAButton], self.rootStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+  }
+
+  private func setupConstraints() {
+    NSLayoutConstraint.activate([
+      self.activityIndicator.heightAnchor.constraint(equalToConstant: 30),
```
ifbarrera: Anyone have a suggestion for where to put these constants? They're pretty one-off.

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
dannyalright: Hmm. Outside of this immediate question: I think we should have guidelines for when and when _not_ to have VoiceOver.

Does the loading VoiceOver string impede or block the fully-loaded VoiceOver string from being said? If not, I'd be in favour for keeping it how you have it (both _loading_ and _loading complete_).

justinswart: My understanding is that we have a pretty low number of users with voice-over enabled so I wouldn't want to over-engineer this for a tiny minority, but something else we can do is add a delay to this signal of like a second or so when we know that voice over is enabled?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
justinswart: I think a local variable would be fine just so that we can update one value and it's clear that they're linked. Actually... why not rather constrain the width to the height and then just set one of those constants?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
ifbarrera: I think if we want to implement this announcement I'll add a new signal for it, but @dannyalright I'm not sure I understand your question. The loading Voice Over string cannot be heard most of the time because the fetch happens so quickly - so all a user hears is "Loading Complete". This feels a bit out of context to me and not really that useful unless you actually have slow signal and hear the previous announcement of "Loading project state", so I'm inclined to potentially remove this announcement all-together.

dannyalright: Fair enough. Let's get rid of it!

justinswart:
Hmm, perhaps one thing that's a little weird from a UX point of view is that navigating back causes the button to refresh... I know _why_ it's happening, but would we want to suppress that at all?

dannyalright:
> Hmm, perhaps one thing that's a little weird from a UX point of view is that navigating back causes the button to refresh... I know why it's happening, but would we want to suppress that at all?

Agree with @justinswart. Also, has the Activity Indicator always been at that Y position, @ifbarrera? It'd be great to have have the same center Y position as the button. Sorry for missing that.

justinswart:
What if we didn't hide the button but showed the loader on top of it (while disabling the button)? Might give more of a feel that the state of the button/view is reloading. Sorry I don't want to inflate the complexity of this and we can definitely iterate on it.

ifbarrera:
> What if we didn't hide the button but showed the loader on top of it (while disabling the button)? Might give more of a feel that the state of the button/view is reloading. Sorry I don't want to inflate the complexity of this and we can definitely iterate on it.

We could do that but then if the button's state changes (let's say it goes from "Back this project" to the "Manage pledge" state where the button is blue and on the right) the transition will be more jarring. @dannyalright thoughts?

dannyalright:
> > What if we didn't hide the button but showed the loader on top of it (while disabling the button)? Might give more of a feel that the state of the button/view is reloading. Sorry I don't want to inflate the complexity of this and we can definitely iterate on it.
> 
> We could do that but then if the button's state changes (let's say it goes from "Back this project" to the "Manage pledge" state where the button is blue and on the right) the transition will be more jarring. @dannyalright thoughts?

Agree with the jarring call, unfortunately.

PR #776, opened 2019-07-31
Feature native checkout rewards gradient

Scollaco:
# 📲 What

Adds a gradient view on the bottom of the RewardCell.

# 🤔 Why

It gives the user a hint that the content is scrollable.

# 🛠 How
- Adding a `GradientView` to the RewardCardContainerView.
- Changed the function that pins the button to the cell, so it can pin the bottom views (button + gradient view)
- Set the view's constraint based on the size of the button. This way the gradient expands if dynamic type is enabled.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-07-31 at 17 17 09 | Simulator Screen Shot - iPhone 8 - 2019-07-31 at 17 15 56 |


Kickstarter-iOS/Views/RewardCardContainerView.swift
```diff
@@ -125,21 +144,30 @@ public final class RewardCardContainerView: UIView {
...
       self.pledgeButton.rightAnchor.constraint(equalTo: layoutMarginsGuide.rightAnchor),
       self.pledgeButton.bottomAnchor.constraint(lessThanOrEqualTo: layoutMarginsGuide.bottomAnchor),
-      self.pledgeButton.heightAnchor.constraint(greaterThanOrEqualToConstant: Styles.minTouchSize.height)
+      self.pledgeButton.heightAnchor.constraint(greaterThanOrEqualToConstant: minTouchSize)
+    ])
+
+    NSLayoutConstraint.activate([
+      self.gradientView.leftAnchor.constraint(equalTo: layoutMarginsGuide.leftAnchor),
+      self.gradientView.rightAnchor.constraint(equalTo: layoutMarginsGuide.rightAnchor),
+      self.gradientView.topAnchor.constraint(equalTo: self.pledgeButton.topAnchor, constant: -minTouchSize),
```
ifbarrera: Looking at the gradient view with the UI inspector, I'd say it looks a teeeeeny bit big.

image

From my understanding the purpose of the gradient is to fade text as it falls *below* the pledge button, basically to prevent de-emphasizing the pledge button. If that's the case then really we just want the gradient to wrap around the pledge button - my recommendation would be using something like `Styles.grid(2)` as a margin above and below the pledge button so that the gradient is only slightly larger than the pledge button in height.

Thoughts @dannyalright ?

dannyalright: > From my understanding the purpose of the gradient is to fade text as it falls below the pledge button, basically to prevent de-emphasizing the pledge button.

That's correct, although fading one or two lines of the _above_ text helps with this goal. I.e. less-faded text above the green button takes away from the button's emphasis.

I think we should ship as is, but appreciate this deep dive!

PR #777, opened 2019-08-01
💲[Native Checkout] Add New Card View

ifbarrera:
# 📲 What

Adds a new `PledgeAddNewCardView` and appends an instance of this view the stack view that displays cards in `PledgePaymentMethodsViewController`.

# 🤔 Why

So a user can add a new card during the pledge flow.

# 🛠 How

Adds `PledgeAddNewCardView` and appends an instance of this view to the credit cards array in `PledgePaymentMethodsViewController`. The `PledgeAddNewCardView` has an "Add new card button" that triggers the modal presentation of the `AddNewCardViewController`.

Note that currently you can only display and dismiss the Add New Card screen - actually adding a credit card will be done separately.

# 👀 See

The new view:

Simulator Screen Shot - iPhone X - 2019-08-01 at 10 56 42

And in action:

xqFaTE1HVD

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Log into an account. Turn the native checkout feature flag on. Navigate to any project, select any reward. Scroll down to the list of credit cards, then scroll all the way to the end of the credit card list. *You should see a new view "Add new card"* and tapping the "Add New Card" button should display the add new card screen. Tapping "Close" on the add new card screen should dismiss the screen.


ifbarrera:
[[Checkout] Pledge Payment Methods Collection View Cell - Add New Card](https://trello.com/c/SJaWqMJL/1477-checkout-pledge-payment-methods-collection-view-cell-add-new-card)

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -3,21 +3,6 @@ import Library
...
-    static let width: CGFloat = 64
-    static let height: CGFloat = 40
-  }
-
-  enum Button {
-    static let width: CGFloat = 217
-  }
-}
-
 final class PledgeCreditCardView: UIView {
```
Scollaco: Since the name of properties now have the prefix `paymentSource`, should we rename this file to `PledgePaymentSourceView`?

Kickstarter-iOS/Views/PledgeAddNewCardView.swift
```diff
@@ -0,0 +1,123 @@
...
+import Prelude
+import UIKit
+
+protocol PledgeAddNewCardViewDelegate: AnyObject {
+  func pledgeAddNewCardViewDidTapAddNewCard(_ view: PledgeAddNewCardView)
+}
+
+final class PledgeAddNewCardView: UIView {
+  private lazy var addNewCardImageView: UIImageView = {
+    UIImageView(image: UIImage.init(named: "icon--add")?.withRenderingMode(.alwaysOriginal))
```
Scollaco: Is there a reason why we are overriding the renderingMode here instead of in the `image set`?

ifbarrera: Good question... I guess since it's highly unlikely we'll ever reuse this view with a different tint it makes sense to set it as an original image in the asset folder 👍

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
ifbarrera: I renamed these to `cardStyles` so it's more consistent with the naming of our files.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -163,3 +166,27 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
+    let navigationController = UINavigationController.init(rootViewController: addNewCardViewController)
+    let offset = navigationController.navigationBar.bounds.height
+
+    self.presentViewControllerWithSheetOverlay(navigationController, offset: offset)
+  }
+}
+
+extension PledgePaymentMethodsViewController: AddNewCardViewControllerDelegate {
+  func addNewCardViewControllerDismissed(_: AddNewCardViewController) {
+    self.dismiss(animated: true)
```
justinswart: Do we need this or is it just to avoid the pattern of the child dismissing itself?

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -163,3 +166,27 @@ final class PledgePaymentMethodsViewController: UIViewController {
       |> \.textAlignment .~ .center
   }
 }
+
+extension PledgePaymentMethodsViewController: PledgeAddNewCardViewDelegate {
+  func pledgeAddNewCardViewDidTapAddNewCard(_: PledgeAddNewCardView) {
+    let addNewCardViewController = AddNewCardViewController.instantiate()
+      |> \.delegate .~ self
+    let navigationController = UINavigationController.init(rootViewController: addNewCardViewController)
```
justinswart: We don't need `.init` here.

Kickstarter-iOS/Views/PledgeAddNewCardView.swift
```diff
@@ -0,0 +1,123 @@
...
+import Prelude
+import UIKit
+
+protocol PledgeAddNewCardViewDelegate: AnyObject {
+  func pledgeAddNewCardViewDidTapAddNewCard(_ view: PledgeAddNewCardView)
+}
+
+final class PledgeAddNewCardView: UIView {
+  private lazy var addNewCardImageView: UIImageView = {
+    UIImageView(image: UIImage.init(named: "icon--add"))
```
justinswart: Or here... were you fighting autocomplete? 😅

Library/Styles/CheckoutStyles.swift
```diff
@@ -17,6 +17,21 @@ public enum CheckoutConstants {
...
     }
   }
+
+  public enum PaymentSource {
+    public enum Card {
+      public static let width: CGFloat = 240
+    }
+
+    public enum ImageView {
+      public static let width: CGFloat = 64
```
justinswart: Little confused as to when we stopped using the grid for these, I know we have done this a fair amount... but why? 😄

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
ifbarrera: We need this to dismiss the `AddNewCardViewController` when `Cancel` is tapped.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
ifbarrera: Yeah that's the only way I could get the autocomplete 😓

Library/Styles/CheckoutStyles.swift
ifbarrera: I think we only really use the grid to define positions and margins - when it comes to assets they have their own size so we use that 🤷‍♀

PR #779, opened 2019-08-01
💲[Native Checkout] RewardCardContainerView snapshots and view model bindings

justinswart:
# 📲 What

- Adds the view model outputs for binding the reward card UI elements.
- Incorporates the state type introduced in #764 and adds functions to determine the state of discrete UI elements.

# 🤔 Why

Rewards can be in _many_ different states based on:
- whether or not the project is live.
- if the user has backed the project.
- if their payment method has errored.
- the availability of the reward.
- the position of the 43rd moon of Saturn.

# 🛠 How

I found that it was easier to determine that all the states had been covered by recording snapshot tests with all of the combinations, I will add view model tests in a subsequent PR and link it back here.

# 👀 See

Examples of just some of the states (see 45 recorded snapshots for all of them):

| Live, Non-backed, available, timebased | Live, backed | Live, backed, errored | Live, no longer available | Live, backed different reward |
| --- | --- | --- | --- | --- |
| testLive_NonBackedProject_AvailableTimebasedReward_lang_en_device_phone4_7inch@2x | testLive_BackedProject_BackedReward_AvailableTimebasedReward_lang_en_device_phone4_7inch@2x | testLive_BackedProject_BackedReward_Errored_AvailableTimebasedReward_lang_en_device_phone4_7inch@2x | testLive_NonBackedProject_UnavailableTimebasedReward_lang_en_device_phone4_7inch@2x | testLive_BackedProject_NonBackedReward_Errored_NoReward_lang_en_device_phone4_7inch@2x |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

Given that there are just so many combinations of states (approx 45 if I have it correct), I think looking through the snapshots, the naming of the snapshot tests, carefully inspecting the properties set on the templates and confirming that the state in the snapshot makes sense, it _should_ be fairly sufficient as AC. Of course run it on device and look out for strangeness but I'm not sure what else to suggest for full QA test coverage other than going through each of the combinations.

# ⏰ TODO

- [ ] Merge in correct styles once that sweep is completed.
- [ ] Add VM tests (separate PR).


justinswart:
[Reward States](https://trello.com/c/IUvuucUU/1448-reward-states)

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -132,7 +132,7 @@ internal final class BetaToolsViewController: UITableViewController {
     self.navigationController?.dismiss(animated: true)
   }
 
-  // MARK: Private Helper Functions
+  // MARK: - Private Helper Functions
```
ifbarrera: Shouldn't this just be `Functions` 😁

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -56,8 +56,7 @@ final class FeatureFlagToolsViewController: UITableViewController {
   private func updateConfig(with features: Features) {
     guard let config = AppEnvironment.current.config else { return }
 
-    let updatedConfig = config
-      |> \.features .~ features
+    let updatedConfig = config |> \.features .~ features
```
ifbarrera: 🤔

Kickstarter-iOS/Views/RewardCardContainerViewTests.swift
```diff
@@ -0,0 +1,248 @@
...
+          project: project,
+          reward: reward
+        )
+
+        FBSnapshotVerifyView(vc.view, identifier: "\(rewardDescription)_lang_\(language)_device_\(device)")
+      }
+    }
+  }
+
+  func testLive_NonBackedProject() {
```
ifbarrera: One thing to note is that there are actually two different kinds of "live, non-backed project" states - one when the user is logged in, and one when the user is logged out. The difference between these two from the data perspective is the value of `project.personalization.isBacking` and `project.personalization.backing`. If the user is logged out, both of these will be `nil`. If the user is logged in `project.personalization.isBacking` will be `false`, and `project.personalization.backing` will be `nil`.

It might be worthwhile to add a test for the logged-out state.

Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -84,12 +100,80 @@ private func backingReward(fromProject project: Project) -> Reward? {
...
+
+private func pledgeButtonIsEnabled(project: Project, reward: Reward) -> Bool {
+  let isAvailable = rewardIsAvailable(project: project, reward: reward)
+  let isBacking = userIsBacking(reward: reward, inProject: project)
+
+  return isAvailable || isBacking
+}
+
+private func rewardIsAvailable(project _: Project, reward: Reward) -> Bool {
+  let limited = !(reward.remaining == nil && reward.endsAt == nil)
```
ifbarrera: How about:

`let isLimited = reward.remaining != nil || reward.endsAt != nil`

Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -152,10 +177,65 @@ private func backingReward(fromProject project: Project) -> Reward? {
...
+  return reward.title.coalesceWith("")
+}
+
+private func pillStrings(project: Project, reward: Reward) -> [String] {
+  var pillStrings: [String] = []
+
+  if let endsAt = reward.endsAt, project.state == .live, endsAt > 0,
+    endsAt >= AppEnvironment.current.dateType.init().timeIntervalSince1970 {
+    let (time, unit) = Format.duration(
+      secondsInUTC: min(endsAt, project.dates.deadline),
```
ifbarrera: Why would an `endsAt` timestamp ever be greater than the project deadline? 🤔

Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -152,10 +177,65 @@ private func backingReward(fromProject project: Project) -> Reward? {
...
+
+  if let remaining = reward.remaining, reward.limit != nil, project.state == .live {
+    pillStrings.append(Strings.Left_count_left(left_count: remaining))
+  }
+
+  return pillStrings
+}
+
+private func stateIconImageColor(project: Project, reward: Reward) -> UIColor? {
+  if userIsBacking(reward: reward, inProject: project) {
```
ifbarrera: Can we do a guard instead?

`guard userIsBacking(reward: reward, inProject: project) else { return nil }`

Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -152,10 +177,65 @@ private func backingReward(fromProject project: Project) -> Reward? {
...
+    } else {
+      return .ksr_soft_black
+    }
+  }
+
+  return nil
+}
+
+private func stateIconImageName(project: Project, reward: Reward) -> String? {
+  if userIsBacking(reward: reward, inProject: project) {
```
ifbarrera: Same here, a guard is a little cleaner.

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
justinswart: lol thanks

Kickstarter-iOS/Views/RewardCardContainerViewTests.swift
justinswart: Ah thanks will do.

Library/ViewModels/RewardCardViewModel.swift
justinswart: This is just how the code was before... didn't really want to change logic in case it was there for some historical reason.

PR #781, opened 2019-08-02
💲[Native Checkout] Button style sweep

justinswart:
# 📲 What

The button design sweep that @dannyalright and I did similar to the buttons introduced in #770.

Will update the rest of the PR details when I open it.

# 🤔 Why

To align our current and deprecated button styles with our new Checkout button styles.

# 🛠 How

Removed the old button styles, replaced them with new styles which are simpler and use programmatic colour mixing for highlight and disabled states. Updated the places where these are used.

# ✅ Acceptance criteria

- [ ] Check screenshots (also note Danny's comments below).
- [ ] Run the app and test for visual regressions.
- [ ] Enjoy a cup of ☕️


Library/Styles/Colors.swift
```diff
@@ -208,6 +212,11 @@ extension UIColor {
     return .hex(0x282828)
   }
 
+  /// 0x1C1C1C
+  public static var ksr_text_black: UIColor {
```
ifbarrera: Is this different from `soft_black`? Just curious 👀

Library/UIColor+Mixing.swift
```diff
@@ -0,0 +1,32 @@
+import UIKit
+
+extension UIColor {
+  func mixLighter(_ amount: CGFloat) -> UIColor {
```
ifbarrera: Ooooo nice! 😍

Library/Styles/Colors.swift
dannyalright: yep. `text_black` is something I dug up from the pattern library for black text. It'd be good to start using it—especially where text is on white backgrounds—as `soft_black` can be little too light for legibility.

PR #67, opened 2017-01-31
Fix tests to use MockDate

justinswart:
I noticed that master was failing on some screenshot tests that weren't dependent on `MockDate` for their assertions. This meant that what previously said _13**6** days to go_ today said _13**5** days to go_ and would continue to decrement as time passes. I've fixed this to always be _15 days to go_ (see screenshots below).

I've updated our model templates in KsApi to use a constant when setting initial date values, see https://github.com/kickstarter/ios-ksapi/pull/12.

I've also updated tests that I could find that were failing after the change, however there may be more that should be updated to use `MockDate`. Tests are currently passing with these changes.

PR #784, opened 2019-08-05
[Native Checkout] Event tracking & tests

cdolm92:
# 📲 What

First release events that adds tracking events and tests for checkout events in project page and reward page.
# ✅ Acceptance criteria

- [ ] Turn on the `KOALA_TRACKING` environment variable in the scheme. Then launch the app and navigate to a project. You should see the `Back this project` CTA, tap it. You should see an event `"Back this Project Button Clicked"` in the console.
- [ ] Turn on the `KOALA_TRACKING` environment variable in the scheme. Then launch the app and navigate to a project. You should see the `Back this project` CTA, tap it. You should get to the reward screen. Tap on a reward to select it and an event, `"Select Reward Button Clicked"` should appear in the console.


Library/Koala/Koala.swift
```diff
@@ -2048,6 +2087,39 @@ private func properties(
...
+  var props: [String: Any] = [:]
+
+  props["name"] = project.name
+  props["pid"] = project.id
+  props["category"] = project.category.name
+  props["has_video"] = project.video != nil
+  props["location"] = project.location.name
+  props["country"] = project.country.countryCode
+
+  let now = MockDate().timeIntervalSince1970
```
Scollaco: I see that on line 2069 we are using `MockDate()` too, but I think we should be using `Date()` instead. @justinswart Do you know the reason for this?

justinswart: Hmm well-spotted @Scollaco. I can't imagine why `MockDate()` was used in this file, I see it's a change I made in #67 to fix tests, but I think doing it in this file, which is application code and not unit test code, is incorrect.

The only reason we use `MockDate()` in tests is to give ourselves control over the movement of time. It also ensures that tests pass in environments that are in different timezones, etc.

So yes, I think both instances of `MockDate()` here should be `AppEnvironment.current.dateType.init()`. This means that in unit tests it will use `MockDate` and in the app it uses `Date`

This likely means that the value of `hours_remaining` in our tracking properties has always been incorrect 🤔 🤷‍♂

PR #785, opened 2019-08-05
Update Rounding Strategy for Rewards

ifbarrera:
# 📲 What

Updates the rounding strategy for the following amounts:

- reward minimums
- backing amounts
- reward conversion labels
- backing amount conversion labels

This PR also adds more unit tests around the rounding behavior for project goal and pledged amounts. It does not change the rounding strategy for these amounts.

As a reminder, the rounding strategies are updated as follows:

- Reward minimums: `.halfUp`
- Backing amounts: `.halfUp`
- Conversion label amounts: `.up`
- Project goal & pledged amounts: `.down`

Note that this is a temporary improvement to our rounding strategy while we work on migrating this logic to the backend for truly consistent rounding across platforms.

# 🤔 Why

For better consistency for currency values across platforms.

# 🛠 How

By default, the `NumberFormatter` used to format currencies via the `Format.currency` helper functions was set to have a `roundingMode` of `down`. This PR adds `roundingMode` as an optional parameter to the `Format.currency` functions (with the default being `.down`). For the updated amount values, we then simply pass a preferred rounding mode to the `Format.currency` function to get the desired rounding behavior.

# 👀 See

Please examine updated screenshots - the most noticeable change is to the reward conversion labels which were previously casted to an `Int` (essentially rounding `.down`), and instead are now  rounded `.up`.

# ♿️ Accessibility 

This PR should not affect accessibility.

# ✅ Acceptance criteria

- [ ] examine updated screenshots
- [ ] Verify against web-staging:

Choose a project on staging.kickstarter.com that's in a non-US currency;
On the staging environment in the iOS app, find the same project;
Verify that the pledged, goal and conversion amounts are the same;
Verify that the reward minimums, and conversion label amounts are the same*;

* Note that the conversion label amounts *may* differ by $1 due to the updated rounding logic for currency conversion labels.

# ⏰ TODO

- [ ] get confirmation of the rounding approach for conversion labels
- [ ] add rounding strategy updates to the `Feature Native Checkout` files (not yet merged to master)


Library/ViewModels/ProjectPamphletMainCellViewModel.swift
```diff
@@ -347,12 +347,12 @@ private func fundingStatus(forProject project: Project) -> String {
   }
 }
 
-typealias ConvertedCurrrencyProjectData = (pledgedAmount: Int, goalAmount: Int, country: Project.Country)
+typealias ConvertedCurrencyProjectData = (pledgedAmount: Int, goalAmount: Int, country: Project.Country)
```
ifbarrera: Fixed a typo 😳

ifbarrera:
Closing this PR. We've decided on a different strategy - we'll be receiving converted amounts directly from the API for rewards, pledge total, and goal amounts which should remove the need handle our own rounding logic.

PR #787, opened 2019-08-07
💲[Native Checkout] Add Shipping Availability Pill

justinswart:
# 📲 What

Adds a pill to our `RewardCardView` describing the shipping availability of a reward, that is whether it's available anywhere in the world or only in the U.S.

# 🤔 Why

A continuation of the rewards carousel work and necessary information to surface to our international backers.

# 🛠 How

- Made use of the string that is set on `reward.shipping.summary` and created a pill from that to add to the collection view.
- Recorded new snapshots of the available and unavailable state.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

**Note:** This does not appear to be localized by the back-end.

- [x] Rewards that have shipping enabled should show this pill.
- [x] Rewards that don't have shipping enabled should not show this pill.


Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -194,7 +194,9 @@ private func rewardTitle(project: Project, reward: Reward) -> String {
 private func pillStrings(project: Project, reward: Reward) -> [String] {
   var pillStrings: [String] = []
 
-  if let endsAt = reward.endsAt, project.state == .live, endsAt > 0,
```
ifbarrera: Do we we want to remove the check from the `remaining` logic too?

https://github.com/kickstarter/ios-oss/blob/8080091b9f023fb8d1ffb87faec20f217ee1a95e/Library/ViewModels/RewardCardViewModel.swift#L210

justinswart: sure 👍

PR #788, opened 2019-08-08
💲[Native Checkout] Update pledge button copy

justinswart:
# 📲 What

Updates the pledge button copy on the reward cell:
- `Pledge X or more` -> `Select`
- For `NoReward`: `Pledge $1 or more` -> `Make a pledge without a reward`

# 🤔 Why

This is the preferred copy, we're also moving away from displaying dynamic text in buttons.

# 🛠 How

Updated the copy, tests and snapshot tests.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Button says `Select` for all available rewards except `NoReward`.
- [ ] Button says `Make a pledge without a reward` for all `NoReward`s.


Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -150,13 +150,10 @@ private func buttonStyleType(project: Project, reward: Reward) -> ButtonStyleTyp
...
 }
 
-private func nonBackedPledgeButtonTitle(project: Project, reward: Reward) -> String {
-  let minimumFormattedAmount = formattedAmountForRewardOrBacking(
-    project: project,
-    rewardOrBacking: .init(left: reward)
-  )
-
-  return Strings.rewards_title_pledge_reward_currency_or_more(reward_currency: minimumFormattedAmount)
+private func nonBackedPledgeButtonTitle(project _: Project, reward: Reward) -> String {
```
cdolm92: Are we still using `project` here?

Library/ViewModels/RewardCardContainerViewModelTests.swift
```diff
@@ -192,14 +192,14 @@ final class RewardCardContainerViewModelTests: TestCase {
     self.pledgeButtonEnabled.assertValues([true, true, true, true, false, false, false, true])
     self.pledgeButtonHidden.assertValues([false, false, false, false, false, false, false, false])
     self.pledgeButtonTitleText.assertValues([
-      "Pledge £6 or more",
-      "Pledge £6 or more",
-      "Pledge £6 or more",
-      "Pledge £6 or more",
+      "Select",
```
cdolm92: Can we use `Strings.Select()`, ` Strings.No_longer_available()`, and     `Strings.Make_a_pledge_without_a_reward()` instead?

Library/ViewModels/RewardCardContainerViewModelTests.swift
```diff
@@ -234,14 +234,14 @@ final class RewardCardContainerViewModelTests: TestCase {
...
```
cdolm92: Here too.

Library/ViewModels/RewardCardContainerViewModel.swift
justinswart: Ah thanks, I'll remove it!

Library/ViewModels/RewardCardContainerViewModelTests.swift
justinswart: So this is an interesting question. I realize we've been fairly inconsistent in tests with this but I'd like to actually do a sweep sometime to change every `String` assertion to be explicit rather than through `Strings.*`. I think the reason being that if the underlying string changes we'd actually want our tests to fail and for us to validate whether the test needs to be updated 🤔 

cc @Scollaco @ifbarrera for discussion.

PR #789, opened 2019-08-08
💲[Native Checkout] Pledge button cell reuse bugfix

justinswart:
# 📲 What

Fixed a minor issue from #779 in that the pledge button sometimes appears on a recycled cell.

# 🤔 Why

We don't want to see a strange empty button on a reward cell.

# 🛠 How

For the hiding and showing of the button we were activating and deactivating constraints to allow the card view to stretch all the way to the bottom of the cell. In addition to this we also need to hide the button in case it was previously visible on the cell before being recycled.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Navigate to a non-live project that you've backed, observe the cell with the `View your pledge` button, keep scrolling and there shouldn't be any other cells containing and empty button as in the screenshot.


Kickstarter-iOS/Views/RewardCardContainerView.swift
```diff
@@ -89,12 +89,14 @@ public final class RewardCardContainerView: UIView {
     self.viewModel.outputs.pledgeButtonHidden.observeValues { [weak self] hidden in
       guard let self = self else { return }
 
+      self.pledgeButton.isHidden = hidden
```
ifbarrera: Can we use the reactive binding for this? `self.pledgeButton.rac.hidden = self.viewModel.outputs.pledgeButtonHidden`

PR #790, opened 2019-08-09
Pledge Screen - Load User's Stored Cards

ifbarrera:
# 📲 What

Load a user's stored cards from Graph.

# 🤔 Why

So users can select a real stored card to pledge with.

# 🛠 How

This PR does the following:

- loads a user's stored cards
- shows an error `MessageBanner` if the request fails

Note that this PR **does not** include the loading state for the cards.

# 👀 See

| Logged In | Logged Out |
| --- | --- |
| cc0W9qTnzk | zspSB8rj0Z |

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

**Logged In**
- [x] with the `Native Checkout Pledge View` feature flag turned **ON**, navigate to any project and select any reward. 
**PASS** In the pledge screen, you should see your user's stored cards load in, with the "Add New Card" tile at the end.

**Logged Out**
- [x] Make sure you are logged out. Then, with the `Native Checkout Pledge View` feature flag turned **ON**, navigate to any project and select any reward. In the pledge screen, you should see the `Continue` button. Tap continue and sign in. 
**PASS** After you sign in, you should see the sign in overlay dismiss, and your stored cards should load in, with the "Add New Card" tile at the end.

**No Network Connectivity**

- [x] Logged in, make sure you have the `Native Checkout Pledge View` feature flag turned **ON**, navigate to any project. Turn off your internet (you can do this by turning airplane mode on your device to **ON**). Then, select any reward.
**PASS** You should not see any cards loaded in. Then, after a few seconds, you should see an error banner appear with the message "Something went wrong".


justinswart:
Could we remove this type from the `xib`, seems to complain that it doesn't exist:

image

justinswart:
Something else either as part of this or separately but we should probably float the card logo to the top:

ifbarrera:
@justinswart I've added both of those visual styling comments to our ticket for `iOS Checkout Polish`

PR #801, opened 2019-08-15
Native checkout experiment

Scollaco:
# 📲 What
- Implements an A/B experiment that brings users to the new Native Checkout or to the old checkout flow, depending on which bucket they are.

# 🤔 Why
- To investigate the impact the new Native Checkout will have in comparison to the old checkout flow.
And this is the [JIRA ticket](https://dripsprint.atlassian.net/projects/NT/issues/NT-38?filter=allopenissues&orderby=priority%20DESC)

# 🛠 How 
The app now receives the `native_checkout_v1` on the `Config` response.
**Note** The users will only see the checkout if the nativeCheckout feature is enabled AND they are on the `experimental` bucket.

# ✅ Acceptance criteria

One way to test this is first go to `Config.swift` and add a breakpoint to see the values of stringsArray (line 30).
 
If the value of `native_checkout_v1` is `experimental`:
- [x] Go to the Debug menu, turn on the Native Checkout feature flag and select any project. You should see the Native Checkout flow.
- [x] Turn off the feature flag and select any project. You should see the old checkout flow. 

If the value of `native_checkout_v1` is `control`:
- [x] Go to the Debug menu, turn on the Native Checkout feature flag and select any project. You should see the **old** checkout flow.
- [x] Turn off the feature flag and select any project. You should see the old checkout flow. 

If you wanna test all the scenarios, simply go to `ExperimentName +Helpers.swift` and change the value of `return Experiment.Variant(rawValue: variant) == .experimental` to `control`. Here we are just changing the logic simulating a different `bucket`.

GIF


Kickstarter-iOS/DataSources/ProjectPamphletContentDataSourceTests.swift
```diff
@@ -64,10 +65,22 @@ final class ProjectPamphletContentDataSourceTests: TestCase {
     }
   }
 
-  func testRewardsSection_nativeCheckoutFeature_showsWithTurnedOff() {
+  func testRewardsSection_showsWhen_nativeCheckoutFeatureEnabled_checkoutExperimentDisabled() {
```
dusi: Should these tests cover the following combinations? (I think we might be missing one)

1. feature enable + experiment disabled
2. feature disabled + experiment enabled
3. feature disabled + experiment disabled (I think we're missing this one)

KsApi/models/Config.swift
```diff
@@ -1,6 +1,7 @@
 public enum Experiment {
   public enum Name: String {
     case creatorsNameDiscovery = "show_created_by_discovery"
+    case nativeCheckoutV1 = "native_checkout_v1"
```
dusi: Just wondering about the naming of this ... is this named `native` after our team (hence it's a native team feature) or is this named after the feature `native checkout`? 

If it's the first then I think it's fine. 
If it's the later then V1 is technically `web checkout` so not sure if it makes sense but don't feel strongly about it (just wanted to bring this up).

Library/Config+Helpers.swift
```diff
@@ -2,8 +2,8 @@ import Foundation
 import KsApi
 
 extension Experiment.Name {
-  public func isEnabled(in _: Environment) -> Bool {
-    guard let experiments = AppEnvironment.current.config?.abExperiments else { return false }
+  public func isEnabled(in environment: Environment = AppEnvironment.current) -> Bool {
```
dusi: I've noticed this function is not tested...can we add appropriate tests?

Library/Feature.swift
```diff
@@ -5,6 +5,14 @@ public enum Feature: String {
   case nativeCheckoutPledgeView = "ios_native_checkout_pledge_view"
 }
 
+extension Feature {
+  public func isEnabled(in environment: Environment = AppEnvironment.current) -> Bool {
```
dusi: Also looks untested..

Library/ViewModels/ProjectPamphletViewModelTests.swift
```diff
@@ -543,4 +646,11 @@ final class ProjectPamphletViewModelTests: TestCase {
       )
     }
   }
+
```
dusi:
```suggestion

// MARK: - Functions

```

dusi: Not sure if this will swing by SwiftFormat? 🤔

KsApi/models/Config.swift
Scollaco: This name starts with `native` so it becomes clear on the Kickstarter app that it's a native experiment (so no one deletes it by accident).

Library/Config+Helpers.swift
Scollaco: This function stayed untested for over a year! 👀 I will add a test for it.

Library/ExperimentName+HelpersTests.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    let config = Config.template
+      |> \.abExperiments .~ [Experiment.Name.nativeCheckoutV1.rawValue: "control"]
+
+    withEnvironment(config: config) {
+      XCTAssertFalse(nativeCheckoutExperimentIsEnabled())
+    }
+  }
+
+  func testIsEnabled_Experimental() {
```
dusi: Can we name these 2 tests similarly?

Library/ExperimentName+HelpersTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    let config = Config.template
+      |> \.abExperiments .~ [Experiment.Name.nativeCheckoutV1.rawValue: "experimental"]
```
dusi: I'd also suggest we test when `abExperiments` is empty or does not include `nativeCheckoutV1`. What do you think?

Library/ProjectPamphlet+Helpers.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+import KsApi
+
+public func userCanSeeNativeCheckout() -> Bool {
+  return nativeCheckoutExperimentIsEnabled() && featureNativeCheckoutEnabled()
```
dusi: I completely missed the naming the first time around...

What do you say we make this more consistent so that we:
* use `isEnabled` for both
* use the same pattern (either: `nativeCheckoutExperimentIsEnabled` & `nativeCheckoutFeatureIsEnabled` or `experimentNativeCheckoutIsEnabled` & `featureNativeCheckoutIsEnabled`)?

Library/ExperimentName+HelpersTests.swift
Scollaco: Ops!

Library/ExperimentName+HelpersTests.swift
Scollaco: Good idea!

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,10 +80,10 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       }
 
     self.goToRewards = goToRewards
-      .filter { _ in featureNativeCheckoutPledgeViewEnabled() }
```
dusi: Make sure that when merging `master` this is actually using the correct feature flag `Native Checkout` not `Native Checkout Pledge View` ... @justinswart has recently changed it since it was causing issue with deep links 😅

Scollaco: Yeah, I noticed that. Actually we now will use  `userCanSeeNativeCheckout` because of the experiment 😄

PR #803, opened 2019-08-15
[Bugfix] message banner hierarchy and rewards card size on smaller screens

ifbarrera:
# 📲 What

Fixes two small bugs:

- a bug where the message banner was being presented below other view's added on the `PledgeViewController`
- bug where the width of the `RewardCell` is awkwardly large on small (SE) screens

# 🤔 Why

Polish & fixes.

# 🛠 How

- guarantee that the `MessageBanner` is brought to to the front before it's displayed
- add some special logic to reduce the width of the `RewardCell` for screens with smaller bounds

# 👀 See

Simulator Screen Shot - iPhone 5s - 2019-08-15 at 17 29 14

uKm57jKI6W

# ✅ Acceptance criteria

- [x] Logged in, make sure you have the Native Checkout Pledge View feature flag turned ON, navigate to any project. Turn off your internet (you can do this by turning airplane mode on your device to ON). Then, select any reward.
PASS You should not see any cards loaded in. Then, after a few seconds, you should see an error banner appear with the message "Something went wrong".
- [x] on iPhone SE, with the "Native Checkout" feature flag turned **ON** navigate to the Rewards screen. Verify that the first reward is fully visible within the bounds of the screen.


Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -206,10 +206,12 @@ final class RewardsCollectionViewController: UICollectionViewController {
     let adjustedContentInset = collectionView.adjustedContentInset
 
     let topBottomSectionInsets = sectionInsets.top + sectionInsets.bottom
+    let leftRightSectionInsets = sectionInsets.left + sectionInsets.right
     let topBottomContentInsets = adjustedContentInset.top + adjustedContentInset.bottom
 
     let itemHeight = collectionView.frame.height - topBottomSectionInsets - topBottomContentInsets
-    let itemWidth = CheckoutConstants.RewardCard.Layout.width
+    let widthByBounds = self.view.bounds.size.width - leftRightSectionInsets
```
dusi: We could probably simply use `collectionView.frame.width` to not having to call `self` here ... in case you'd still prefer to use `self` it could be shortened to use the getter on `CGRect` like so `self.view.bounds.width`.

PR #804, opened 2019-08-16
View backing when project has ended

ifbarrera:
# 📲 What

This PR fixes two issues:
- users were previously able to navigate to the pledge screen on non-live projects
- users trying to view their pledge on a successful project were instead taken to the update pledge screen

Now, users viewing a project they did not back will not be able to tap rewards if the project has ended. Users who have backed a project and tap "View your pledge" are correctly taken to the "View pledge" screen.

# 🛠 How

# 👀 See

| Viewing project that has ended | Viewing Backed project that has ended |
| --- | --- |
| 6jfODEoaK5 | xtwL5ospjc |

# ✅ Acceptance criteria

**Viewing project that has ended**
- [x] With the *Native Checkout* feature flag turned **ON**, browse through discovery and find a project that is either *Funding Successful* or *Funding Unsuccessful*. Tap "View rewards".
**AC** On the rewards carousel, you should not be able to select/tap any of the rewards.

**Viewing project you backed that has ended**
- [x] With the *Native Checkout* feature flag turned **ON**, log in and go to your backed projects. Tap on a project with tag *Successful* or *Unsuccessful* or *Cancelled*. Tap "View your pledge".
**AC** On the rewards carousel, you should not be able to select/tap any of the rewards that do not have the `View your pledge` button. On the reward you backed, you should see the `View your pledge` button, and tapping it should take you to the "View pledge" screen.


Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -73,10 +74,27 @@ public final class RewardsCollectionViewModel: RewardsCollectionViewModelType,
       selectedRewardFromId,
       refTag
     )
+    .filter { arg in
+      let (project, _, _) = arg
```
justinswart: you probably did this to help the compiler but i'm sure it'll also compile as `.filter { project, _, _ in ... }`

PR #805, opened 2019-08-19
Determine "Apple Pay capable" based on available card types from the project

ifbarrera:
# 📲 What

Previously, we were determining Apple Pay capabilities from hardcoded logic. This PR updates that logic to instead use the `availableCardTypes` field on the project, and also adds the business logic of showing/hiding the Apple Pay button in the new `PledgePaymentMethodsViewController`. The logic for determining supported networks is shared, and is now fully tested.

# 🤔 Why

Part of Apple Pay implementation on the new Pledge screen.

# 🛠 How

We now have a new `availableCardTypes` field on the `Project` model. We can use this array to determine which `supportedNetworks` to request from the method PassKit function `PKPaymentAuthorizationViewController.canMakePayments(usingNetworks:)`. There is still fallback logic which mimics the backend logic, in case `availableCardTypes` is `nil`.

# ✅ Acceptance criteria

- [ ] TBD

# ⏰ TODO

- [x] **IMPORTANT** we're still waiting for the backend to update the `availableCardType` strings that are being returned to match the GraphQL `CreditCardType` schema names so this feature will work


KsApi/models/templates/ProjectTemplates.swift
```diff
@@ -3,6 +3,15 @@ import Prelude
 // swiftlint:disable line_length
 extension Project {
   internal static let template = Project(
+    availableCardTypes: [
+      "AMEX",
+      "MASTERCARD",
+      "VISA",
+      "DISCOVER",
+      "JCB",
+      "DINERS",
```
ifbarrera: Note that `DINERS` is actually not a `supportedNetwork` according to Apple Pay, so it's filtered out.

KsApi/models/Project.swift
```diff
@@ -4,6 +4,7 @@ import Prelude
 import Runes
 
 public struct Project {
+  public var availableCardTypes: [String]?
```
ifbarrera: This is optional because the project JSON object we get from the `/discover` endpoint doesn't have this field (it's only a partial project), and we use the same `Project` model to decode that payload.

Library/PKPaymentAuthorizationViewController+Helpers.swift
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public static func applePayCapable(for project: Project) -> Bool {
+    return PKPaymentAuthorizationViewController.canMakePayments(
+      usingNetworks: PKPaymentAuthorizationViewController.supportedNetworks(for: project)
+    )
+  }
+
+  public static func supportedNetworks(for project: Project) -> [PKPaymentNetwork] {
+    guard let availableCardTypes = project.availableCardTypes else {
```
ifbarrera: `availableCardTypes` should never be `nil` in this context, but decided to add a fallback just in case 🤷‍♀

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -99,12 +99,20 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
       }
+
+    self.applePayButton.rac.hidden = self.viewModel.outputs.applePayButtonHidden
   }
 
   // MARK: - Configuration
 
-  func configureWith(_ user: User) {
-    self.viewModel.inputs.configureWith(user)
+  func configureWith(_ value: (User, Project)) {
```
justinswart: Can this be `func configure(with value: (user: User, project: Project)) {`, reads a bit better than `value.0`

KsApi/models/Project.swift
justinswart: I suppose we could also consider coalescing an empty array if we wanted this to be non-optional 🤔 but I guess having `nil` here is more explicit about _not_ having the information rather than there are no available card types.

Library/PKPaymentAuthorizationViewController+Helpers.swift
```diff
@@ -0,0 +1,72 @@
...
+  public static var merchantIdentifier: String {
+    return Secrets.ApplePay.merchantIdentifier
+  }
+
+  public static var allSupportedNetworks: [PKPaymentNetwork] = [
+    .amex,
+    .masterCard,
+    .visa,
+    .discover,
+    .JCB,
```
justinswart: I would just make it `.jcb` 😅

Library/PKPaymentAuthorizationViewController+Helpers.swift
```diff
@@ -0,0 +1,72 @@
...
+      return PKPaymentNetwork.discover
+    case .jcb:
+      return PKPaymentNetwork.JCB
+    case .mastercard:
+      return PKPaymentNetwork.masterCard
+    case .unionPay:
+      return PKPaymentNetwork.chinaUnionPay
+    case .visa:
+      return PKPaymentNetwork.visa
+    default:
```
justinswart: How about just exhausting the `switch` and returning `nil` for those two? Just gives us the confidence that if another case is added, the compiler will force it to be handled here.

KsApi/models/Project.swift
ifbarrera: Yeah I'd rather keep it `nil` to make it clear that the info is not available.

Library/PKPaymentAuthorizationViewController+Helpers.swift
ifbarrera: This is Apple's choice not ours 😄 it's `PKPaymentNetwork.JCB`

Library/PKPaymentAuthorizationViewController+Helpers.swift
ifbarrera: Okay yeah that makes sense 👍

Library/PKPaymentAuthorizationViewController+Helpers.swift
justinswart: lol oops

Library/PKPaymentAuthorizationViewController+Helpers.swift
```diff
@@ -65,8 +65,10 @@ extension PKPaymentAuthorizationViewController {
       return PKPaymentNetwork.chinaUnionPay
     case .visa:
       return PKPaymentNetwork.visa
-    default:
-      return nil // case .diners and case .generic
+    case .diners:
```
justinswart: Could just be `case .diners, .generic:` but I won't lie in the road for it 😬 😅

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -7,48 +7,72 @@ import ReactiveExtensions_TestHelpers
...
+      self.vm.inputs.viewDidLoad()
+
+      self.scheduler.run()
+
+      self.applePayButtonHidden.assertValues([false])
+      self.reloadPaymentMethods.assertValue(response.me.storedCards.nodes)
+    }
+  }
+
+  func testReloadPaymentMethods_Error_LoggedIn_ApplePayCapable_isFalse() {
```
justinswart: Is there a scenario where we could be `true` for Apple Pay capable but error on fetching the cards? And do we want to test that?

ifbarrera: Sure

PR #806, opened 2019-08-19
♿️ Disable Dynamic Type on buttons

dusi:
# 📲 What

Disables Dynamic Type support globally on our buttons.

# 🤔 Why

Due to quite a few bugs and also the fact that all our buttons resemble `Apple Pay` button very closely we're temporarily disabling Dynamic Type support.

The majority of our buttons did not use multiline labels and therefore supporting Dynamic Type would result in truncations anyways.

# 🛠 How

Simply by using UIKit's `boldSystemFont` API.

# 👀 See

**Project page**

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-08-19 at 2 36 50 PM | Screen Shot 2019-08-19 at 2 40 27 PM |
| Screen Shot 2019-08-19 at 2 36 56 PM | Screen Shot 2019-08-19 at 2 40 29 PM |
| Screen Shot 2019-08-19 at 2 36 58 PM | Screen Shot 2019-08-19 at 2 40 31 PM |
| Screen Shot 2019-08-19 at 2 37 00 PM | Screen Shot 2019-08-19 at 2 40 34 PM |

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-08-19 at 2 43 16 PM | Screen Shot 2019-08-19 at 2 45 02 PM |
| Screen Shot 2019-08-19 at 2 43 18 PM | Screen Shot 2019-08-19 at 2 45 13 PM |
| Screen Shot 2019-08-19 at 2 43 21 PM | Screen Shot 2019-08-19 at 2 45 15 PM |
| Screen Shot 2019-08-19 at 2 43 26 PM | Screen Shot 2019-08-19 at 2 45 18 PM |

**Rewards**

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-08-19 at 2 43 56 PM | Screen Shot 2019-08-19 at 2 45 46 PM |
| Screen Shot 2019-08-19 at 2 44 01 PM | Screen Shot 2019-08-19 at 2 45 50 PM |
| Screen Shot 2019-08-19 at 2 44 05 PM | Screen Shot 2019-08-19 at 2 45 54 PM |
| Screen Shot 2019-08-19 at 2 44 09 PM | Screen Shot 2019-08-19 at 2 45 56 PM |

**Pledge screen**

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-08-19 at 1 52 13 PM | Screen Shot 2019-08-19 at 1 57 51 PM |
| Screen Shot 2019-08-19 at 1 52 14 PM | Screen Shot 2019-08-19 at 1 57 53 PM |
| Screen Shot 2019-08-19 at 1 52 17 PM | Screen Shot 2019-08-19 at 1 57 57 PM |
| Screen Shot 2019-08-19 at 1 52 20 PM | Screen Shot 2019-08-19 at 1 57 59 PM |

To illustrate this slightly better here's how **before** and **after** looks with a bit of blur (that potentially highlights the distinction of each major element on the screen.

| Before 🐛 | After 🦋 |
| --- | --- |
| project-page-before | project-page-after |
| project-page-manage-before | project-page-manager-after |
| rewards-carousel-before | rewards-carousel-after |

# 👆 Notes

Please note that you might need to temporarily do the following changes to `ProjectPamphlet+Helpers.swift` due to an existing problem with our experiments..

```swift
public func userCanSeeNativeCheckout() -> Bool {
  return /* experimentNativeCheckoutIsEnabled() && */ featureNativeCheckoutIsEnabled()
}
```

# ✅ Acceptance criteria

- [x] CTA button on `Project page` does not support Dynamic Type
- [x] CTA button on `Rewards cell` does not support Dynamic Type
- [x] CTA button on `Pledge screen` does not support Dynamic Type
- [ ] Every other button in the app (random sweep should be done) does not support Dynamic Type


Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
```diff
@@ -5,7 +5,7 @@ import Prelude
 final class PledgeContinueViewController: UIViewController {
   // MARK: - Properties
 
-  private let continueButton = MultiLineButton(type: .custom)
+  private let continueButton = UIButton(type: .custom)
```
dusi: We no longer need to use `MultiLineButton` here...in fact since we've previously added trimming we haven't been using it for a while (so this is just a cleanup).

Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
```diff
@@ -25,10 +25,7 @@ final class PledgeContinueViewController: UIViewController {
       |> checkoutBackgroundStyle
 
     _ = self.continueButton
-      |> greenButtonStyle
-      |> UIButton.lens.title(for: .normal) %~ { _ in
-        Strings.Continue()
-      }
+      |> continueButtonStyle
```
dusi: Moving to `// MARK: - Styles` for consistency

Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
```diff
@@ -74,3 +71,15 @@ final class PledgeContinueViewController: UIViewController {
...
+// MARK: - Styles
+
+private let continueButtonStyle: ButtonStyle = { button in
+  button
+    |> greenButtonStyle
+    |> UIButton.lens.title(for: .normal) %~ { _ in
+      Strings.Continue()
+    }
+    |> (UIButton.lens.titleLabel .. UILabel.lens.font) .~ UIFont.boldSystemFont(ofSize: 16)
+    |> (UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode) .~ NSLineBreakMode.byTruncatingMiddle
```
dusi: Using consistent truncation in the middle for all our CTA buttons now

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -74,10 +74,7 @@ final class PledgeCTAContainerView: UIView {
       |> \.spacing .~ Styles.gridHalf(1)
 
     _ = self.pledgeCTAButton
-      |> pledgeCTAButtonStyle(
-        isAccessibilityCategory,
-        amountAndRewardTitleStackViewIsHidden: self.titleAndSubtitleStackView.isHidden
-      )
+      |> pledgeCTAButtonStyle
```
dusi: We no longer need to know whether it's accessibility category or what state we're in

Kickstarter-iOS/Views/RewardCardContainerView.swift
```diff
@@ -104,8 +104,9 @@ public final class RewardCardContainerView: UIView {
     self.viewModel.outputs.pledgeButtonStyleType
       .observeForUI()
       .observeValues { [weak self] styleType in
-        guard let self = self else { return }
-        _ = self.pledgeButton |> styleType.style
+        _ = self?.pledgeButton
```
dusi: Using optional chaining for consistency between files

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -169,27 +168,19 @@ final class PledgeCTAContainerView: UIView {
...
-      ? NSLineBreakMode.byWordWrapping : NSLineBreakMode.byTruncatingTail
-
-    return button
-      |> roundedStyle(cornerRadius: 12)
-      |> UIButton.lens.titleLabel.font .~ UIFont.ksr_headline(size: 15)
-      |> UIButton.lens.layer.borderWidth .~ 0
-      |> (UIButton.lens.titleLabel .. UILabel.lens.textAlignment) .~ NSTextAlignment.center
-      |> (UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode) .~ lineBreakMode
-  }
+private let pledgeCTAButtonStyle: ButtonStyle = { button in
```
justinswart: It shouldn't be necessary to set `roundedStyle` and `borderWidth` because these are included in our base button styles now. If we want all of our buttons to behave this way shouldn't we move these to the base style?

Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
```diff
@@ -76,10 +76,7 @@ final class PledgeContinueViewController: UIViewController {
 
 private let continueButtonStyle: ButtonStyle = { button in
   button
-    |> greenButtonStyle
-    |> UIButton.lens.title(for: .normal) %~ { _ in
-      Strings.Continue()
-    }
-    |> (UIButton.lens.titleLabel .. UILabel.lens.font) .~ UIFont.boldSystemFont(ofSize: 16)
-    |> (UIButton.lens.titleLabel .. UILabel.lens.lineBreakMode) .~ NSLineBreakMode.byTruncatingMiddle
+    |> ButtonStyleType.green.style
```
justinswart: Not necessary to use the `ButtonStyleType` `enum` here, that is really just for our VMs to output a button style as something that we can do equality comparison on in tests. Swift doesn't support equality comparison on function types. So this can just be `greenButtonStyle` 👍

Scollaco:
I've noticed, though, that the buttons on the credit card cells still support dynamic type. Was this intentional or supposed to be part of other PR?

Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
Scollaco: I was thinking that we can start renaming the `Styles` pragma mark above `bindStyles()` to `Styles Binding` or `Bind Styles`, so it differs from the `Styles` on the bottom of the file. what do you think?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -169,27 +164,10 @@ final class PledgeCTAContainerView: UIView {
 
 private func adaptableStackViewStyle(_ isAccessibilityCategory: Bool) -> (StackViewStyle) {
   return { (stackView: UIStackView) in
-    let axis: NSLayoutConstraint.Axis = (isAccessibilityCategory ? .vertical : .horizontal)
     let spacing: CGFloat = (isAccessibilityCategory ? Styles.grid(1) : 0)
 
     return stackView
-      |> \.axis .~ axis
+      |> \.axis .~ NSLayoutConstraint.Axis.horizontal
```
Scollaco: Was the compiler giving you trouble here?  😄 I think that `|> \.axis .~ .horizontal` is more readable and more used on the codebase. Also, we have more styling going on inside the `bindStyles()` function. To keep consistency and avoid doing the same thing in multiple places, I would either move this code to the bindStyles function (not recommended) or do a small refactor bringing the other styles to the bottom. (recommended, since that's how we are approaching styling).

Library/Styles/ButtonStyles.swift
```diff
@@ -15,10 +15,10 @@ public let applePayButtonStyle: ButtonStyle = { button in
 public let baseButtonStyle: ButtonStyle = { button in
   button
     |> roundedStyle(cornerRadius: Styles.grid(2))
-    |> UIButton.lens.titleLabel.font .~ UIFont.ksr_callout().bolded
-    |> (UIButton.lens.titleLabel .. UILabel.lens.textAlignment) .~ NSTextAlignment.center
-    |> UIButton.lens.contentEdgeInsets .~ .init(all: Styles.grid(2))
+    |> UIButton.lens.titleLabel .. UILabel.lens.font .~ UIFont.boldSystemFont(ofSize: 16)
```
Scollaco: Lenses give us access to the `font` property of a UIButton. So this line could simply be `|> UIButton.lens.titleLabel.font .~ UIFont.boldSystemFont(ofSize: 16)` (this syntax is the most commonly used on our codebase)

dusi: I thought we were moving away from lenses?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
dusi: It's not that the compiler would give us trouble but we've been seeing a lot of warnings lately related to Xcode's type checker...therefore at some point we've decided that for styles it would make more sense to be more explicit.

I'll make the styling more consistent by extracting `bindStyles` here.

Library/Styles/ButtonStyles.swift
dusi: Had to update `Prelude` so if you can also check it out and approve if everything looks fine

dusi:
Thanks for catching up the payment methods issue (I've fixed it and went through all the places we refer to base style in the whole app so now all buttons should in fact not support Dynamic Type 🤞 ).

As for the pragma mark I'd defer this to some future work (it would be great to have a whole team agree on this before changing something we've been doing for some time now. `// MARK: - Bind Styles` is currently only used in 1 place (`PledgePaymentMethodsViewController`) and I'm kinda tempted at changing it to `Styles` for consistency 😅

Kickstarter-iOS/Views/Controllers/PledgeContinueViewController.swift
justinswart: I see we're only using this in one place, not time to remove `MultiLineButton` yet?

dusi:
Fun fact, those screenshots did actually not fit on my Desktop (I think I need 15" MBP or the new Pro Display soon) 🤦‍♂

PR #807, opened 2019-08-19
Reward card IA updates

justinswart:
# 📲 What

Adds the IA changes that we wanted to get in for the reward card before shipping v1:

- The `descriptionLabel` is positioned above the `includedItemsStackView`.
- `descriptionTitleLabel` has been removed.
- `sectionTitleLabelStyle` has been updated to `ksr_text_dark_grey_500` to improve accessibility.
- `estimatedDeliveryDateLabel` appears below `includedItemsStackView`.
- Limited reward now display a pill indicating `<remaining> left of <limit>`.
- A backer count is displayed in place of the limited reward remaining pill when:
  - Reward has at least 1 backer AND:
  - Reward has no limit OR:
  - Limit has been reached OR:
  - Project has ended.
- If none of the above are matched, the backer count pill isn't shown.

# 👀 See

[Abstract](https://app.goabstract.com/projects/a94cda80-1e60-11e8-8903-0b9b4e24ea4b/branches/422360c1-542b-4f6c-ba28-31ed4266c855/collections/42941029-98b9-45d1-9738-efa3928d7d09)
[Jira](https://dripsprint.atlassian.net/browse/NT-170)

Some examples:

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |
| image | image |
| image | image |

# ✅ Acceptance criteria

- [ ] Test the backer count displaying/hidden in the states described above.
- [ ] Confirm positioning of items is correct against Abstract designs.


Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -201,8 +201,11 @@ private func rewardTitle(project: Project, reward: Reward) -> String {
 private func pillStrings(project: Project, reward: Reward) -> [String] {
   var pillStrings: [String] = []
 
-  if project.state == .live, let endsAt = reward.endsAt, endsAt > 0,
-    endsAt >= AppEnvironment.current.dateType.init().timeIntervalSince1970 {
+  if project.state == .live,
+    let endsAt = reward.endsAt,
+    endsAt > 0,
+    endsAt >= AppEnvironment.current.dateType.init().timeIntervalSince1970,
+    reward.limit == nil || reward.remaining ?? 0 > 0 {
```
ifbarrera: Can we extract this last piece to a var?

`let isUnlimitedOrAvailable = reward.limit == nil || reward.remaining ?? 0 > 0`

Come to think of it, you probably don't need the first part of that statement because if it's an "unlimited" reward then `reward.remaining` will be `nil` so you'll still get the same outcome. So:

`let isAvailable = reward.remaining ?? 0 > 0` would suffice?

You're the expert on this though 😛

justinswart:
Updated the strings to include `_count` but realized that wasn't enough, the string in the yaml in fact needs to look like this:

```yaml
remaining_count_left_of_limit_count:
  zero: "%{remaining_count} left of %{limit_count}"
  one: "%{remaining_count} left of %{limit_count}"
  two: "%{remaining_count} left of %{limit_count}"
  few: "%{remaining_count} left of %{limit_count}"
  many: "%{remaining_count} left of %{limit_count}"
  other: "%{remaining_count} left of %{limit_count}"
```

This isn't necessary for this particular string as we don't vary the string based on the count. Also I'm not sure that it's built to handle multiple counts in the same string and looking through the rest of the yaml it's fairly inconsistent about that.

🤷‍♂

Library/ViewModels/RewardCardViewModel.swift
justinswart: 😄 I extracted it to a variable for readability.

Actually, we do need the OR, coalescing the nil `remaining` value would assume it's a limited reward. We're saying - return the string if this reward is unlimited OR if it has more than 0 remaining. If the left side is `nil` the second is ignored.

Tidied it up a little though.

PR #808, opened 2019-08-20
Disabling experiment on debug builds

Scollaco:
# 📲 What
- Disables experiments on non-Release builds.

# 🤔 Why
The Native Experiment currently makes the Native Checkout unavailable for users that are part of the `control` bucket, making the work and QA for that project impossible. 

In order to revert this, we should enable experiments exclusively for Release builds, this way the Native Checkout work won't be interrupted.

# 🛠 How

Following @justinswart 's suggestion, we now have a `debugDefault` variable inside the `Experiment.Name` enum that currently enables both existing experiments.

# ✅ Acceptance criteria

- [ ] On `alpha`, `beta` and `debug` builds, the native checkout show be available regardless the value of the bucket of the `native_checkout_v1` experiment. *

* Note: The app should still be sending `control` or `experimental` to db/Looker.


Library/ExperimentName+HelpersTests.swift
```diff
@@ -37,7 +42,7 @@ final class ExperimentName_HelpersTests: TestCase {
     let config = Config.template
       |> \.abExperiments .~ ["unknown": "experimental"]
 
-    withEnvironment(config: config) {
+    withEnvironment(config: config, mainBundle: self.releaseBundle) {
```
ifbarrera: Can we add a test to verify that when the bundle *is not* the `.isRelease` bundle, that the `debugDefault` is returned?

PR #822, opened 2019-08-28
Update config when the session starts and ends

dusi:
# 📲 What

Updates config every time a user logs in or logs out (in addition to launching the app or coming back to the app from background).

# 🤔 Why

We've noticed that the config would only refresh after the app has been backgrounded and brought back to foreground. This caused some confusion when logging in as a different user and logging out because the config wasn't properly updated unless the app was backgrounded and brought back to foreground again. This PR fixes that and should always make sure that the config is always up to date even without leaving the app.

# 🛠 How

Very straightforward, please see the code.

# ✅ Acceptance criteria

In order to properly test this you'll have to observe your device connection with `mitm` tool like `Charles proxy` or `mitmproxy` (there's a guru card with instructions on how to set this up).

- [ ] Launching the app from terminated state should fetch `config`
- [ ] Re-opening the app from background should fetch `config`
- [ ] Logging in should fetch `config`
- [ ] Logging out should fetch `config`


ifbarrera:
Nice! This has been in the backlog for while 🌟

PR #823, opened 2019-09-01
[1/2] Shipping rules - Selection UI

dusi:
# 📲 What

Improves the UI so that it provides UX for selecting shipping rule from a list

# 🤔 Why

First of two PRs allowing users changing shipping rule.

# 👀 See

shipping rule selection

# ✅ Acceptance criteria

- [ ] Tapping on shipping rule that has not been previously selected deselects previous selection and selects the new shipping rule
- [ ] Scrolling newly selected shipping rule off screen and then scrolling it back on shows the proper rule selected (this proves that data source has been updated properly)
- [ ] Tapping on shipping rule that has been previously selected doesn't do anything
- [ ] The modal opens with selected shipping rule scrolled to top
- [ ] The modal opens and flashes scroll indicators

# ⏰ TODO

- [ ] Dismiss shipping rule modal on selection (will be done in a separate PR)


Kickstarter-iOS/Views/Controllers/ShippingRulesTableViewController.swift
```diff
@@ -48,4 +77,12 @@ final class ShippingRulesTableViewController: UITableViewController {
       selectedShippingRule: selectedShippingRule
     )
   }
+
+  // MARK: - UITableViewDelegate
+
+  override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
+    self.viewModel.inputs.didSelectShippingRule(at: indexPath.row)
+
+    self.tableView.deselectRow(at: indexPath, animated: true)
```
ifbarrera: Can we add this to the observer for the `deselectCellAtIndex` output?

Library/ViewModels/ShippingRuleCellViewModel.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+import KsApi
+import Prelude
+import ReactiveExtensions
+import ReactiveSwift
+
+public protocol ShippingRuleCellViewModelInputs {
+  func configureWith(_ values: ShippingRuleData)
```
ifbarrera: Should this be `configureWith(_ value:)` since usually we reserve the plural for arrays?

Library/ViewModels/ShippingRulesViewModelTests.swift
```diff
@@ -2,31 +2,133 @@ import Foundation
...
+
+    self.flashScrollIndicators.assertValueCount(1)
+    self.reloadDataWithShippingRulesData.assertValueCount(1)
+    self.reloadDataWithShippingRulesReload.assertValues([true])
+    self.scrollToCellAtIndex.assertValues([1])
+
+    self.vm.inputs.didSelectShippingRule(at: 0)
+
+    self.flashScrollIndicators.assertValueCount(1)
+    self.reloadDataWithShippingRulesData.assertValueCount(2)
```
ifbarrera: Should we also assert the actual values of `reloadDataWithShippingRulesData`? Mostly to check that `ShippingRuleData` is updated with the correct `selectedShippingRule`.

Library/ViewModels/ShippingRulesViewModelTests.swift
```diff
@@ -2,31 +2,133 @@ import Foundation
...
+        ShippingRuleData(
+          project: project, selectedShippingRule: selectedShippingRule, shippingRule: shippingRule3
+        )
+      ]
+    ])
+
+    self.flashScrollIndicators.assertValueCount(1)
+    self.reloadDataWithShippingRulesData.assertValueCount(1)
+    self.reloadDataWithShippingRulesReload.assertValues([true])
+    self.scrollToCellAtIndex.assertValues([1])
```
ifbarrera: Should we check the value of `selectCellAtIndex`?

Library/ViewModels/ShippingRulesViewModelTests.swift
```diff
@@ -2,31 +2,133 @@ import Foundation
...
+    self.vm.inputs.configureWith(
+      .template,
+      shippingRules: [.template, .template, selectedShippingRule, .template],
+      selectedShippingRule: selectedShippingRule
+    )
     self.vm.inputs.viewDidLoad()
 
-    self.loadValues.assertValues([["Canada (+CA$ 25)"]])
+    self.deselectCellAtIndex.assertDidNotEmitValue()
+    self.selectCellAtIndex.assertValues([2])
```
ifbarrera: Oh I see you're testing them here. Cool cool.

Kickstarter-iOS/Views/Controllers/ShippingRulesTableViewController.swift
dusi: I tried this but for some unknown reason it doesn't work. I believe this is why the rest of the codebase also deselects row directly in `didSelectRowAt` ¯\_(ツ)_/¯

Library/ViewModels/ShippingRulesViewModelTests.swift
dusi: Yep, I can do that!

Kickstarter-iOS/Views/Controllers/ShippingRulesTableViewController.swift
ifbarrera: Gotcha!

PR #824, opened 2019-09-01
[2/2] Shipping rules - Dismiss on selection

dusi:
# 📲 What

Dismisses shipping rules modal on shipping rule selection

# 🤔 Why

Based on design

# 👀 See

dismissal

# ✅ Acceptance criteria

- [ ] Tapping on shipping rule that's different from current selection selects the rule and dismisses the modal after a small delay
- [ ] Tapping on currently selected shipping rule does nothing
- [ ] Selecting shipping rule properly updates `shipping amount` on the pledge screen and `total` value


Kickstarter-iOS/Views/Controllers/PledgeShippingLocationViewController.swift
```diff
@@ -136,32 +136,37 @@ final class PledgeShippingLocationViewController: UIViewController {
...
-    self.viewModel.inputs.dismissShippingRulesButtonTapped()
-  }
-
   // MARK: - Functions
 
   private func presentShippingRules(
     _ project: Project, shippingRules: [ShippingRule], selectedShippingRule: ShippingRule
   ) {
     let viewController = ShippingRulesTableViewController.instantiate()
-      |> \.navigationItem.leftBarButtonItem .~ UIBarButtonItem(
```
dusi: I've moved this to be a responsibility of `ShippingRulesTableViewController`

Library/ViewModels/PledgeShippingLocationViewModel.swift
```diff
@@ -89,25 +89,30 @@ public final class PledgeShippingLocationViewModel: PledgeShippingLocationViewMo
     self.shippingLocationButtonTitle = self.notifyDelegateOfSelectedShippingRule
       .map { $0.location.localizedName }
 
-    self.dismissShippingRules = self.dismissShippingRulesButtonTappedProperty.signal
+    self.dismissShippingRules = Signal.merge(
+      self.shippingRulesCancelButtonTappedProperty.signal,
+      self.shippingRuleUpdatedSignal.signal
+        .ignoreValues()
+        .ksr_debounce(.milliseconds(300), on: AppEnvironment.current.scheduler)
```
dusi: We're using a small delay...using `ksr_debounce` also allows us to test this properly

Library/ViewModels/PledgeShippingLocationViewModelTests.swift
```diff
@@ -18,7 +18,7 @@ private let shippingRules = [
 final class PledgeShippingLocationViewModelTests: TestCase {
   private let vm: PledgeShippingLocationViewModelType = PledgeShippingLocationViewModel()
 
-  private let amountAttributedText = TestObserver<NSAttributedString, Never>()
+  private let amountText = TestObserver<String, Never>()
```
dusi: Since we're not really interested in any attributes I'm simply mapping the `string` value for more readable tests

Library/ViewModels/PledgeShippingLocationViewModelTests.swift
```diff
@@ -53,24 +53,27 @@ final class PledgeShippingLocationViewModelTests: TestCase {
...
-      XCTAssertEqual(self.amountAttributedText.values.last?.string, "+$0.00")
-      self.shippingLocationButtonTitle.assertValues([])
       self.notifyDelegateOfSelectedShippingRule.assertDidNotEmitValue()
+      self.shippingLocationButtonTitle.assertValues([])
 
-      self.scheduler.run()
+      self.scheduler.advance()
 
-      let defaultShippingRule = shippingRules.first(where: { $0.location == .brooklyn })
+      guard let defaultShippingRule = shippingRules.first(where: { $0.location == .brooklyn }) else {
```
dusi: Some cleanup so that we don't have to force unwrap

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -38,10 +39,18 @@ public final class ShippingRulesViewModel: ShippingRulesViewModelType,
     .map(second)
     .skipNil()
 
-    let dataSelected = Signal.combineLatest(
+    let selectedIndex = Signal.combineLatest(
       dataInitial,
       self.didSelectShippingRuleAtIndexProperty.signal.skipNil()
     )
+    .map { data, newIndex in (data.1.firstIndex(of: data.2), newIndex) }
```
Scollaco: I wonder if the would be more readable to unpack `dataInitial` by doing 
```
let selectedIndex = Signal.combineLatest(
      dataInitial.map(unpack),
      self.didSelectShippingRuleAtIndexProperty.signal.skipNil()	      
    )
```

And use explicit names here
```
.map { data, newIndex in (data.1.firstIndex(of: data.2), newIndex) }
```

So it would become
```
.map { _, shippingRules, selectedRule, newIndex in 
    (shippingRules.firstIndex(of: selectedRule), newIndex) 
}
```

Library/ViewModels/PledgeShippingLocationViewModel.swift
dusi: Yes, exactly...we've paired on this with Danny and decided we need a bit of a delay before dismissing so that the user has a perception of what they've selected.

It should still be testable..

```swift
self.vm.inputs.shippingRuleUpdated(to: selectedShippingRule)

self.scheduler.advance(by: .milliseconds(300))

self.dismissShippingRules.assertValueCount(1)
```

Library/ViewModels/ShippingRulesViewModel.swift
dusi: Good idea...I'd love to do that but it won't compile ... not really sure what's wrong.

Screen Shot 2019-09-03 at 4 18 15 PM

Library/ViewModels/PledgeShippingLocationViewModel.swift
Scollaco: Awesome. Thanks for clarifying!

Library/ViewModels/ShippingRulesViewModel.swift
Scollaco: Yeah, the compiler must be confused due some type inference. I'm ok to left as it is then, unless you think it's worth it investigating a little more.

PR #827, opened 2019-09-03
💸 [Native Checkout] Create Backing Mutation

cdolm92:
# 📲 What

Adding `createBacking` Mutation to the codebase. 

# 🤔 Why

This is needed so that users can complete their pledge on the app.

# ▶️ ▶️ ▶️ Note
**Testing procedure to trigger call:**

- [ ] Add the following code:

```
// PledgeViewModel.swift

// L: 21
var createBackingSuccess: Signal<String, Never> { get }
var createBackingFailure: Signal<String, Never> { get }

// L: 78
let createBackingOnLoad = projectAndReward
      .map { CreateBackingInput(
        projectId: toBase64($0.0), amount: "200", locationId: "\($0.0.location.id)",
        rewardId: toBase64($0.1), paymentSourceId: "source", paymentType: "card"
      ) }
      .switchMap {
        AppEnvironment.current.apiService.createBacking(input: $0)
          .materialize()
      }

    self.createBackingFailure = createBackingOnLoad.errors()
      .map {
        $0.localizedDescription
      }
    self.createBackingSuccess = createBackingOnLoad.values().map { $0.createBacking.checkout.state.rawValue }

// L: 112
 public let createBackingSuccess: Signal<String, Never>
 public let createBackingFailure: Signal<String, Never>
```

```
// PledgeViewController.swift

// L: 188
self.viewModel.outputs.createBackingFailure
   .observeForControllerAction()
    .observeValues { [weak self] errorMessage in
       self?.present(UIAlertController.genericError(errorMessage), animated: true, completion: nil)
    }

self.viewModel.outputs.createBackingSuccess
   .observeForUI()
    .observeValues { [weak self] state in
      print("\(state)")
    }
```

# ✅ Acceptance Criteria
- [ ] Enable the `Native Checkout Pledge View` feature flag in the debug menu. The call will trigger when the `PledgeViewController` loads
- [ ]  Navigate to a project tap `Back this Project`, select a reward the pledge screen loads, check the console log to see the checkout state printed. It should say `VERIFYING`


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -111,3 +111,15 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
   public var inputs: PledgeViewModelInputs { return self }
   public var outputs: PledgeViewModelOutputs { return self }
 }
+
+private func toBase64(_ project: KsApi.Project) -> String {
```
ifbarrera: I don't think we actually need this. We have a protocol `GraphIDBridging` that defines:

```
extension GraphIDBridging {
  public var graphID: String {
    return Data("\(type(of: self).modelName)-\(self.id)".utf8).base64EncodedString()
  }
}
```

Looks like `Project` already conforms to it, you just need to add it to `Reward`.

KsApi/models/Project.swift
```diff
@@ -25,6 +25,10 @@ public struct Project {
   public var urls: UrlsEnvelope
   public var video: Video?
 
+  public static func decode(id: Int) -> String {
```
Scollaco: Can you add test for this function?

KsApi/models/Reward.swift
```diff
@@ -22,6 +22,10 @@ public struct Reward {
     return self.id == Reward.noReward.id
   }
 
+  public static func decode(id: Int) -> String {
```
Scollaco: And for this too

KsApi/models/Project.swift
ifbarrera: I think from my comment [here](https://github.com/kickstarter/ios-oss/pull/827#discussion_r320845807) we actually don't need this either because the `GraphIdBridging` protocol takes care of it for us.

KsApi/mutations/inputs/CreateBackingInput.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+public struct CreateBackingInput: GraphMutationInput {
+  let projectId: String
+  let amount: String
+  let locationId: String
```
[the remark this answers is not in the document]
ifbarrera: @cdolm92 can we also write some tests for this input class? We want to verify that the output of `toInputDictionary()` is what we expect.

KsApi/models/CreateBackingEnvelope.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public struct CreateBackingEnvelope: Decodable {
+  public var createBacking: CreateBacking
```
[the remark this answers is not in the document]
ifbarrera: Nevermind, looks like we need this since the structure of the response from graph is:

```
{
  "data": {
    "createBacking": {
      "clientMutationId": null,
      "checkout": {
        "state": "VERIFYING"
      }
    }
  }
}
```

KsApi/models/Project.swift
ifbarrera: @cdolm92 can we remove this?

KsApi/mutations/inputs/CreateBackingInputTests.swift
```diff
@@ -0,0 +1,25 @@
+@testable import KsApi
+import Prelude
+import XCTest
+
+final class CreateBackingInputTests: XCTestCase {
+  func testCreateBackingInputDictionary() {
```
ifbarrera: Can we add a test for the case where `locationId` and `rewardId` are `nil`?

PR #828, opened 2019-09-04
[Apple Pay] Part 1 - Create Apple Pay Backing Mutation

ifbarrera:
# 📲 What

Adds the `createApplePayBacking` mutation and mutation input object, with tests. Also, in preparation for sending formatted amounts to the mutation, I've added a new currency formatter that formats currency amounts into doubles with the correct number of fraction digits. This formatting function isn't in use in this PR, but will be used in Part II of this work.

# 🤔 Why

To allow users to pledge using  Apple Pay without going through the legacy API routes.

# 🛠 How

Adds `CreateApplePayBackingMutation` and `CreateApplePayBackingInput`. Also adds `decimalCurrency(for amount:)` that formats `Double`s into currency strings that can be sent to the API as an input parameter.

# ✅ Acceptance criteria

Note: this mutation cannot be tested with the existing code. It will be tested in a follow up PR once we integrate the Apple Pay button with the mutation.

# ⏰ TODO

- [ ] waiting on a refactor of the Graph mutation for consistent naming

### DIFF
--- a/KsApi/mutations/inputs/CreateApplePayBackingInputTests.swift
+++ b/KsApi/mutations/inputs/CreateApplePayBackingInputTests.swift
@@ -6,25 +6,25 @@ final class CreateApplePayBackingInputTests: XCTestCase {
   func testCreateApplePayBacking_toInputDictionary_noNilValues() {
     let input = CreateApplePayBackingInput(
       amount: "10.00",
-      locationId: 123,
+      locationId: "123",
       paymentInstrumentName: "instrumentName",
       paymentNetwork: "paymentNetwork",
-      projectId: 12_345,
-      rewardId: 321,
+      projectId: "12345",
+      rewardId: "321",
       stripeToken: "stripeTokenXYZ",
       transactionIdentifier: "transactionId"
     )
 
     let inputDictionary = input.toInputDictionary()
 
-    XCTAssertEqual(inputDictionary["amount"] as! String, "10.00")
-    XCTAssertEqual(inputDictionary["locationId"] as! String, "123")
-    XCTAssertEqual(inputDictionary["paymentInstrumentName"] as! String, "instrumentName")
-    XCTAssertEqual(inputDictionary["paymentNetwork"] as! String, "paymentNetwork")
-    XCTAssertEqual(inputDictionary["projectId"] as! String, "12345")
-    XCTAssertEqual(inputDictionary["rewardId"] as! String, "321")
-    XCTAssertEqual(inputDictionary["token"] as! String, "stripeTokenXYZ")
-    XCTAssertEqual(inputDictionary["transactionIdentifier"] as! String, "transactionId")
+    XCTAssertEqual(inputDictionary["amount"] as? String, "10.00")
+    XCTAssertEqual(inputDictionary["locationId"] as? String, "123")
+    XCTAssertEqual(inputDictionary["paymentInstrumentName"] as? String, "instrumentName")
+    XCTAssertEqual(inputDictionary["paymentNetwork"] as? String, "paymentNetwork")
+    XCTAssertEqual(inputDictionary["projectId"] as? String, "12345")
+    XCTAssertEqual(inputDictionary["rewardId"] as? String, "321")
+    XCTAssertEqual(inputDictionary["token"] as? String, "stripeTokenXYZ")
+    XCTAssertEqual(inputDictionary["transactionIdentifier"] as? String, "transactionId")
   }
 
   func testCreateApplePayBacking_toInputDictionary_withNilValues() {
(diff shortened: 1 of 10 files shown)


KsApi/mutations/inputs/CreateApplePayBackingInputTests.swift
```diff
@@ -0,0 +1,53 @@
...
+      paymentNetwork: "paymentNetwork",
+      projectId: 12_345,
+      rewardId: 321,
+      stripeToken: "stripeTokenXYZ",
+      transactionIdentifier: "transactionId"
+    )
+
+    let inputDictionary = input.toInputDictionary()
+
+    XCTAssertEqual(inputDictionary["amount"] as! String, "10.00")
```
dusi: I know that these are unit tests but I would be more careful force-unwrapping things like this...because if something happens the test would not fail, but rather the test would crash. This might be during deployment, etc. so being a little bit more safe and intentional we can use the following approach (which should fail the test if conditions change, but not crash it).

IMO failing test is usually better more understandable than crash in console reported by CircleCI.

```swift
XCTAssertEqual(inputDictionary["amount"] as? String, "10.00")
...
```

What do you think?

Library/Format.swift
```diff
@@ -456,15 +475,27 @@ private struct NumberFormatterConfig {
...
     currencySymbol: "$"
   )
 
+  fileprivate static let defaultDecimalCurrencyConfig = NumberFormatterConfig(
+    numberStyle: .decimal,
+    roundingMode: .down,
+    maximumFractionDigits: 2,
+    minimumFractionDigits: 2,
+    generatesDecimalNumbers: false,
+    locale: Locale(identifier: "en"), // Always format decimal currency amounts using En locale
```
dusi: Can you elaborate on this?

dusi: I've also noticed these `static` configurations are not sorted alphabetically so was just wondering if we could put `defaultDecimalCurrencyConfig` and `defaultCurrencyConfig` next to each other, because they both deal with currency?

Library/FormatTests.swift
```diff
@@ -94,6 +94,44 @@ final class FormatTests: TestCase {
...
+    XCTAssertEqual(Format.decimalCurrency(for: 10.00), "10.00")
+    XCTAssertEqual(Format.decimalCurrency(for: 10.50), "10.50")
+    XCTAssertEqual(Format.decimalCurrency(for: 10.5555), "10.55", "Rounds down to 2 fraction digits")
+    XCTAssertEqual(Format.decimalCurrency(for: 10.511), "10.51", "Rounds down to 2 fraction digits")
+  }
+
+  func testAttributedCurrency() {
+    withEnvironment(locale: Locale(identifier: "en")) {
+      withEnvironment(countryCode: "US") {
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .us)!.string, "$1,000.00")
```
dusi: Again, I would suggest not using force unwrap here so `XCTAssertTrue` API might be considered as an alternative.

```swift
XCTAssertTrue(Format.attributedCurrency(1_000, country: .us)?.string == "$1,000.00")
...
```

Library/Format.swift
ifbarrera: > Can you elaborate on this?

Since we're using this formatter to format currency amounts into `String`s for the API, we need to always use the `.en` locale because the API doesn't support accepting amounts that are formatted using a different locale. For example: `10.00` vs `10,00`.

dusi: Could we possibly expand that comment and explain this is for the purposes of API/Graph?

Library/FormatTests.swift
```diff
@@ -94,6 +94,44 @@ final class FormatTests: TestCase {
...
+    withEnvironment(locale: Locale(identifier: "en")) {
+      withEnvironment(countryCode: "US") {
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .us)?.string, "$1,000.00")
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .ca)?.string, " CA$ 1,000.00")
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .gb)?.string, "£1,000.00")
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .dk)?.string, " DKK 1,000.00")
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .de)?.string, "€1,000.00")
+        XCTAssertEqual(Format.attributedCurrency(1_000, country: .jp)?.string, "¥1,000.00")
+
+        XCTAssertEqual(Format.attributedCurrency(
```
dusi: Any chance we could fit this on one line like so? (it's really hard to read the test)

```swift
XCTAssertEqual(
  Format.attributedCurrency(1_000, country: .ca, omitCurrencyCode; true)?.string, " CA$ 1,000.00")
)
```

PR #829, opened 2019-09-05
[Apple Pay] Hide Apple Pay for unsupported project countries

ifbarrera:
# 📲 What

Apple Pay is only supported a subset of countries that Kickstarter has launched in, so we should hide the Apple Pay button if the project is in a country that is not supported.

# 🤔 Why

Apple Pay.

# 🛠 How

Apple Pay is only supported in a subset of countries that Kickstarter has launched in. We get this list of countries from the `/config` endpoint. In addition to checking whether the user's device has the Apple Pay capability, as also determine whether Kickstarter supports Apple Pay in the project's country.

# ✅ Acceptance criteria

**With the Native Checkout and Native Pledge Screen feature flags turned ON**

- [x] Run the app on a device that has Apple Pay set up. Navigate to a project in the US. You should see the Apple Pay button on the new pledge screen.
- [x] Run the app on a device that has Apple Pay set up. Navigate to a project in Mexico (not supported). You should **not** see the Apple Pay button on the new pledge screen.


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -38,7 +38,8 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
       }
 
     self.applePayButtonHidden = configureWithValue
-      .map { $0.applePayCapable }
+      .map { ($0.project, $0.applePayCapable) }
+      .map { project, applePayCapable in showApplePayButton(for: project, applePayCapable: applePayCapable) }
```
Scollaco: We could use the point free syntax here
```
.map(showApplePayButton(for:applePayCapable:))
```

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -117,4 +118,35 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
       self.notifyDelegateLoadPaymentMethodsError.assertDidNotEmitValue()
     }
   }
+
+  func testApplePayButton_applePayCapable_unsupportedProjectCountry_isHidden() {
```
Scollaco: I think that the name `func testApplePayButton_isHidden_applePayCapable_unsupportedProjectCountry()` would describe better what we are testing here.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -117,4 +118,35 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+
+    withEnvironment(config: mockConfig) {
+      self.vm.inputs.configureWith((User.template, project, true))
+      self.vm.inputs.viewDidLoad()
+
+      self.applePayButtonHidden.assertValues([true])
+    }
+  }
+
+  func testApplePayButton_applePayCapable_supportedProjectCountry_isShown() {
```
Scollaco: Same here

PR #830, opened 2019-09-09
[NT-164] Consume Rewards converted minimum from v1

Scollaco:
# 📲 What 
- Shows Rewards converted minimum based on the value of `convertedMinimum` returned by the API.

# 🤔 Why
- This allows us to have one source of truth and avoids the calculation of conversions to be done on the client side.

# 🛠 How
- Added a new property `convertedMinimum` to our Reward model.
- Changed logic to present the value of that property instead of calculating conversion value based on project conversion rate.
- Updated tests.

# ✅ Acceptance criteria

- [ ] iOS and web should show the same values for conversion minimum on Reward cells.

### DIFF
--- a/KsApi/models/Reward.swift
+++ b/KsApi/models/Reward.swift
@@ -5,7 +5,7 @@ import Runes
 
 public struct Reward {
   public let backersCount: Int?
-  public let convertedMinimum: Double?
+  public let convertedMinimum: Double
   public let description: String
   public let endsAt: TimeInterval?
   public let estimatedDeliveryOn: TimeInterval?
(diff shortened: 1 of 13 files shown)


KsApi/models/Reward.swift
```diff
@@ -5,6 +5,7 @@ import Runes
 
 public struct Reward {
   public let backersCount: Int?
+  public let convertedMinimum: Double
```
ifbarrera: Is `convertedMinimum` guaranteed to always be returned? If no conversion is required (ie. project's currency is the same as the user's currency) is the `convertedMinimum` just the same as the `minimum`?

KsApi/models/templates/RewardTemplates.swift
```diff
@@ -24,6 +25,7 @@ extension Reward {
 
   public static let noReward = Reward(
     backersCount: nil,
+    convertedMinimum: 0,
```
ifbarrera: I know this is just a template, but I think a `convertedMinimum` would never be 0 🤔 Maybe make this `1` so it's more realistic?

KsApi/models/Reward.swift
Scollaco: Exactly

KsApi/models/templates/RewardTemplates.swift
Scollaco: I set `convertedMinimum` to  `0` so it matches the `minimum`. For this template, all properties are either `nil` or `0`.

PR #831, opened 2019-09-09
[NT-280] Manage pledge screen from CTA container

Scollaco:
# 📲 What
 - Presents the ManageViewPledge screen (blank for now) from the cta container.

# 🤔 Why
- This was decided to make the access to the screen easier, compared to what was proposed previously.

# 🛠 How
- Moved the code from `RewardsCollectionViewController` to `ProjectPamphletViewController`
- Moved tests to the classes to be tested

# 👀 See

| Managing 🐛 | Viewing Pledge 🦋 |
| --- | --- |
| manage_pledge_cta | view_pledge_cta |

# ✅ Acceptance criteria

## With `featureNativeCheckoutPledgeView` enabled. 
- [ ] Select a non-backed project and tap the `Back this project` button.  The Rewards Carousel should be presented.
- [ ] Select a backed project and tap the `Manage` button. The  `ManageViewPledgeViewController` (blank) should be presented.
- [ ] Selected a backed and ended project and tap the `View your pledge` button. The  `ManageViewPledgeViewController` (blank) should be presented.
### Note: 
By tapping to view your pledge, note that the title shown is `Manage your pledge`. The logic to change the title based of the pledge and project states will be implemented in a different PR.

## With `featureNativeCheckoutPledgeView` disabled. 
On the Rewards Carousel screen:
- [ ] Select a non-backed project and tap the `Back this project` button.  The Rewards Carousel should be presented.
- [ ] Select a backed project and tap the `Manage` button. The `DeprecatedRewardPledgeViewController` should be presented.
- [ ] Selected a backed and ended project and tap the `View your pledge` button. The `BackingViewController` should be presented.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
@@ -135,7 +135,7 @@ public final class ProjectPamphletViewController: UIViewController {
         self?.goToRewards(project: project, refTag: refTag)
       }
 
-    self.viewModel.outputs.goToManagePledge
+    self.viewModel.outputs.goToManageViewPledge
       .observeForControllerAction()
       .observeValues { [weak self] (params) in
         let (project, reward, refTag) = params
@@ -210,7 +210,12 @@ public final class ProjectPamphletViewController: UIViewController {
     let managePledgeViewController = ManagePledgeViewController.instantiate()
     managePledgeViewController.configureWith(project: project, reward: reward)
 
-    self.navigationController?.pushViewController(managePledgeViewController, animated: true)
+    let nav = UINavigationController(rootViewController: managePledgeViewController)
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      _ = nav
+        |> \.modalPresentationStyle .~ .formSheet
+    }
+    self.present(nav, animated: true)
   }
 
   private func goToDeprecatedPledge(project: Project, reward: Reward, refTag _: RefTag?) {
@@ -218,14 +223,31 @@ public final class ProjectPamphletViewController: UIViewController {
       .configuredWith(
         project: project, reward: reward
     )
+    pledgeViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(
+      image: image(named: "icon--cross", tintColor: .ksr_navy_600),
+      style: .plain,
+      target: pledgeViewController,
+      action: #selector(DeprecatedRewardPledgeViewController.closeButtonTapped)
+    )
 
-    self.navigationController?.pushViewController(pledgeViewController, animated: true)
+    let nav = UINavigationController(rootViewController: pledgeViewController)
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      _ = nav
+        |> \.modalPresentationStyle .~ .formSheet
+    }
+    self.present(nav, animated: true, completion: nil)
   }
 
   private func goToViewBacking(project: Project, user: User?) {
-    let backingViewController = BackingViewController.configuredWith(project: project, backer: user)
+    let backingViewController = BackingViewController.configuredWith(project: project, backer: nil)
 
-    self.navigationController?.pushViewController(backingViewController, animated: true)
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      let nav = UINavigationController(rootViewController: backingViewController)
+        |> \.modalPresentationStyle .~ .formSheet
+      self.present(nav, animated: true, completion: nil)
+    } else {
+      self.navigationController?.pushViewController(backingViewController, animated: true)
+    }
   }
 
   private func updateContentInsets() {
--- a/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
@@ -135,6 +135,14 @@ public final class ProjectPamphletViewController: UIViewController {
         self?.goToRewards(project: project, refTag: refTag)
       }
 
+    self.viewModel.outputs.goToManagePledge
+      .observeForControllerAction()
+      .observeValues { [weak self] (params) in
+        let (project, reward, refTag) = params
+
+        self?.goToManagePledge(project: project, reward: reward, refTag: refTag)
+    }
+
     self.viewModel.outputs.configureChildViewControllersWithProject
       .observeForUI()
       .observeValues { [weak self] project, refTag in
@@ -186,6 +194,24 @@ public final class ProjectPamphletViewController: UIViewController {
     self.present(vc, animated: true)
   }
 
+  private func goToManagePledge(project: Project, reward: Reward, refTag _: RefTag?) {
+    let managePledgeViewController = ManagePledgeViewController.instantiate()
+    managePledgeViewController.configureWith(project: project, reward: reward)
+
+    let nav = UINavigationController(rootViewController: managePledgeViewController)
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      _ = nav
+        |> \.modalPresentationStyle .~ .formSheet
+    }
+    self.present(nav, animated: true)
+  }
+
+  private func goToViewBacking(project: Project, user: User?) {
+    let backingViewController = BackingViewController.configuredWith(project: project, backer: user)
+
+    self.navigationController?.pushViewController(backingViewController, animated: true)
+  }
+
   private func updateContentInsets() {
     let buttonSize = self.pledgeCTAContainerView.pledgeCTAButton.systemLayoutSizeFitting(
       UIView.layoutFittingCompressedSize
@@ -203,8 +229,9 @@ public final class ProjectPamphletViewController: UIViewController {
 }
 
 extension ProjectPamphletViewController: PledgeCTAContainerViewDelegate {
-  func pledgeCTAButtonTapped() {
-    self.viewModel.inputs.backThisProjectTapped()
+
+  func pledgeCTAButtonTapped(_ state: PledgeStateCTAType) {
+    self.viewModel.inputs.pledgeCTAButtonTapped(state)
   }
 }
 
--- a/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
@@ -143,6 +143,18 @@ public final class ProjectPamphletViewController: UIViewController {
         self?.goToManagePledge(project: project, reward: reward, refTag: refTag)
     }
 
+    self.viewModel.outputs.goToDeprecatedViewBacking
+      .observeForControllerAction()
+      .observeValues { [weak self] project, user in
+        self?.goToViewBacking(project: project, user: user)
+    }
+
+    self.viewModel.outputs.goToDeprecatedManagePledge
+      .observeForControllerAction()
+      .observeValues { [weak self] project, reward, refTag in
+        self?.goToDeprecatedPledge(project: project, reward: reward, refTag: refTag)
+    }
+
     self.viewModel.outputs.configureChildViewControllersWithProject
       .observeForUI()
       .observeValues { [weak self] project, refTag in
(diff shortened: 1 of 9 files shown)


Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -135,6 +135,26 @@ public final class ProjectPamphletViewController: UIViewController {
         self?.goToRewards(project: project, refTag: refTag)
       }
 
+    self.viewModel.outputs.goToManageViewPledge
```
dusi: I'm seeing a lot of inconsistencies between the naming of VMs `output` signal and the `private` action on the controller...do you think it would be worth it cleaning this up..

So for example:

`self.viewModel.outputs.goToManageViewPledge` would trigger `self?.goToManageViewPledge`, etc...

For the `deprecated` ones I'd suggest we simply add it as a suffix:

`self.viewModel.outputs.goToViewBackingDeprecated` links to `self?.goToViewBacking`...

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -186,6 +206,50 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    }
+    self.present(nav, animated: true)
+  }
+
+  private func goToDeprecatedPledge(project: Project, reward: Reward, refTag _: RefTag?) {
+    let pledgeViewController = DeprecatedRewardPledgeViewController
+      .configuredWith(
+        project: project, reward: reward
+      )
+    pledgeViewController.navigationItem.leftBarButtonItem = UIBarButtonItem(
```
dusi: Is there a reason we've decided to set this externally on `DeprecatedRewardPledgeViewController`?

If we do this inside of the controller and set `self.navigationItem` (like we do everywhere else we will be able to remove this code as well the identical code in `ProjectPamphletContentViewController`.

But maybe I'm missing the reason here? 🤔

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -186,6 +206,50 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    self.present(nav, animated: true, completion: nil)
+  }
+
+  private func goToViewBacking(project: Project, user _: User?) {
+    let backingViewController = BackingViewController.configuredWith(project: project, backer: nil)
+
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      let nav = UINavigationController(rootViewController: backingViewController)
+        |> \.modalPresentationStyle .~ .formSheet
+      self.present(nav, animated: true, completion: nil)
```
dusi: Can we omit `completion` handler?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -186,6 +206,50 @@ public final class ProjectPamphletViewController: UIViewController {
...
+    } else {
+      self.navigationController?.pushViewController(backingViewController, animated: true)
```
dusi: So, I know we've previously discussed the fact that we're presenting manage screen over the previous context using `.formSheet`...but do we really want to `push` on non-iPad devices? Just trying to understand the decision behind this?

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -4,7 +4,7 @@ import Prelude
 import UIKit
 
 protocol PledgeCTAContainerViewDelegate: AnyObject {
-  func pledgeCTAButtonTapped()
+  func pledgeCTAButtonTapped(_ state: PledgeStateCTAType)
```
dusi: Maybe `with` state?

```swift
func pledgeCTAButtonTapped(with state: PledgeStateCTAType)
```

then we call it with more context like so

```swift
self?.delegate?.pledgeCTAButtonTapped(with: state)
```

Dunno which one is better ¯\_(ツ)_/¯

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -34,6 +34,12 @@ public protocol ProjectPamphletViewModelOutputs {
   /// Emits a (project, isLoading) tuple used to configure the pledge CTA view
   var configurePledgeCTAView: Signal<(Either<Project, ErrorEnvelope>, Bool), Never> { get }
 
+  var goToDeprecatedManagePledge: Signal<PledgeData, Never> { get }
+
+  var goToDeprecatedViewBacking: Signal<(Project, User?), Never> { get }
```
dusi: I know we're deprecating stuff but looking at these parameters I wonder if we can improve this by declaring `BackingData` typealias, etc. for `(Project, User?)` tuple

Just a suggestion

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,13 +86,36 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
 
     let ctaButtonTapped = freshProjectAndRefTag
-      .takeWhen(self.backThisProjectTappedProperty.signal)
-      .map { project, refTag in
-        (project, refTag)
+      .takePairWhen(self.pledgeCTAButtonTappedProperty.signal)
+      .map(unpack)
+
+    let goToManagePledge = ctaButtonTapped
+      .filter(canShowManageViewPledgeScreen(_:_:state:))
```
dusi: Looks like `canShowManageViewPledgeScreen` does not need `refTag`? Could omit and have more readable call site `canShowManageViewPledgeScreen(_:state:)`

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,13 +86,36 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
+      .map { (project, refTag, _) -> PledgeData in
+        let r = reward(from: project.personalization.backing, inProject: project)
```
dusi: Could we name this `reward` (it should fit on the screen and be more descriptive)?

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,13 +86,36 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
+        return PledgeData(project: project, reward: r, refTag: refTag)
+      }
+
+    self.goToManageViewPledge = goToManagePledge
+      .filter { _ in featureNativeCheckoutPledgeViewIsEnabled() }
+
+    self.goToDeprecatedManagePledge = ctaButtonTapped
+      .filter(shouldGoToDeprecatedManagePledge(_:_:state:))
```
dusi: Same thing with `refTag`...is this because we'll need it in the future but don't need it just now?

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,13 +86,36 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
+      .map { (project, refTag, _) -> PledgeData in
+        let r = reward(from: project.personalization.backing, inProject: project)
```
dusi: Same for the naming

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
Scollaco: It makes sense to change the `private` actions to match the name of the output. Will change that.

Regarding the `deprecated` naming, I'm following the naming convention that we adopted for deprecated classes (`DeprecatedCheckoutViewModel`, `DeprecatedRewardPledgeViewController`, `DeprecatedRewardCell` etc)

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
Scollaco: On a rush to unblock the other tickets I had copied the code from somewhere else. Good catch.

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
Scollaco: The from `BackingViewController`, we can access Messages (the vc's navigation controller pushes the MessagesViewController). If we present the `BackingViewController` modally instead, we are no longer able to see the messages, since it requires a navigation controller to work

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
Scollaco: Me gusta :)

Library/ViewModels/ProjectPamphletViewModel.swift
Scollaco: Good idea

Library/ViewModels/ProjectPamphletViewModel.swift
Scollaco: The problem with that is `ctaButtonTapped` emits (project, refTag, state). If we omit the refTag, the compiler will complain because the types emitted don't match the function arguments, thus the point free syntax will no longer be possible.

Library/ViewModels/ProjectPamphletViewModel.swift
Scollaco: I tried, but keep getting the error `Variable used within its own initial value`

Library/ViewModels/ProjectPamphletViewModel.swift
dusi: I see!

Thanks for clarifying this...I don't know whether a stylistic thing like being able to call something with `point-free` expression would overweight the code debt of passing argument that's not necessary.

In this case, I'd consider whether non-point-free expression would be more suitable (less code + less debt).

```swift
.filter { (project, _, state) in canShowManageViewPledgeScreen(project, state:state) }
```

or

```swift
.filter { canShowManageViewPledgeScreen($0.0, state:$0.2) }
```

Library/ViewModels/ProjectPamphletViewModel.swift
dusi: Right, good point!

How about something like this:

```swift
.map { (project, refTag, _) -> PledgeData in
        PledgeData(
          project: project,
          reward: reward(from: project.personalization.backing, inProject: project),
          refTag: refTag
        )
      }
```

or

```swift
.map {
        PledgeData(
          project: $0.0, reward: reward(from: $0.0.personalization.backing, inProject: $0.0), refTag: $0.1
        )
      }
```
(but that might be more than confusing due to the repetition of those variables)

dusi: Since this is a language limitation I'd be OK with leaving it as is .. your decision! 🥂

Library/ViewModels/ProjectPamphletViewModel.swift
dusi: Yeah, I'd probably try to carry over as little unused things as possible..this makes it easier to reason about code and will be less confusing in the future (read in a week when we forget WTF was this) :D

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -284,3 +319,52 @@ private func fetchProject(projectOrParam: Either<Project, Param>, shouldPrefix:
...
+  _: RefTag?,
+  state: PledgeStateCTAType?
+) -> Bool {
+  guard let state = state else {
+    return false
+  }
+  return userCanSeeNativeCheckout() && (state == .pledge || state == .viewRewards)
+}
+
+private func canShowManageViewPledgeScreen(
```
dusi: After we remove the unused params, these function names should fit on one line making it a bit more readable IMO.

```swift
private func canShowRewardsScreen(state: PledgeStateCTAType?) -> Bool { .. }
private func canShowManageViewPledgeScreen(_ project: Project, state: PledgeStateCTAType?) -> Bool { .. }
private func shouldGoToDeprecatedViewBacking(_ project: Project, state: PledgeStateCTAType?) -> Bool { .. }
private func shouldGoToDeprecatedManagePledge(_ project: Project, state: PledgeStateCTAType?) -> Bool { .. }
```

Library/ViewModels/ProjectPamphletViewModelTests.swift
```diff
@@ -686,7 +862,7 @@ final class ProjectPamphletViewModelTests: TestCase {
       self.goToRewardsProject.assertDidNotEmitValue()
       self.goToRewardsRefTag.assertDidNotEmitValue()
 
-      self.vm.inputs.backThisProjectTapped()
+      self.vm.inputs.pledgeCTAButtonTapped(.pledge)
```
dusi: See, I think this is where a better naming convention would help us a little `Tapped(with: .pledge)`..What do you think?

Scollaco: Oh right. I changed in one place but missed that one. Will change it

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -80,13 +89,38 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
...
 
     let ctaButtonTapped = freshProjectAndRefTag
-      .takeWhen(self.backThisProjectTappedProperty.signal)
-      .map { project, refTag in
-        (project, refTag)
+      .takePairWhen(self.pledgeCTAButtonTappedProperty.signal)
+      .map(unpack)
+
+    let goToManagePledge = ctaButtonTapped
+      .filter { canShowManageViewPledgeScreen($0.0, state:$0.2) }
```
dusi: Looking at this in action, it might actually be more readable to name those params. I'll leave this out to you, I'm OK with both - but wanted to raise it in case you were in favour of the other option.

`.filter { project, _, state in canShowManageViewPledgeScreen(project, state: state) }`

(this of course would apply to other places in this PR.

PR #832, opened 2019-09-10
NT-124 [Apple Pay] Apple Pay Button Interaction

ifbarrera:
# 📲 What

Adds full interaction of backing a project with Apple Pay on the new pledge screen.

# 🤔 Why

So that users can back projects using Apple Pay.

# 🛠 How

Apple Pay works by doing the following:

1. Generating a PKPaymentRequest which includes details of what will be charged.
2. Displaying the PKPaymentAuthorizationViewController
3. Generating a `stripeToken` by passing an encrypted, authorized payment (`PKPayment`) to the `.createToken` function on the Stripe SDK.
4. Passing this `stripeToken`, along with some details from the `PKPayment` object to our `createApplePayBacking` graph mutation.
5. On success, we take users to the `Thanks` screen. If there's an error, we display a message banner with details of the error.

# 👀 See

Simulator Screen Shot - iPhone 8 - 2019-09-10 at 11 51 58

# ✅ Acceptance criteria

Note: Apple Pay can only be tested end-to-end on a device.

**Apple Pay without shipping**
1. Make sure you have both `Native Checkout` and `Native Pledge Screen.
2. Select a project, then choose a reward without shipping (ex. "Make a Pledge without a Reward")
3. Tap "Apple Pay" button: 
- [x] You should see the Apple Pay sheet with the pledge amount in the project's currency
- [x] There should be no shipping item
- [x] The line item should be the project's name.

**Apple Pay with shipping**
1. Make sure you have both `Native Checkout` and `Native Pledge Screen.
2. Select a project, then choose a reward with shipping enabled.
3. Tap "Apple Pay" button: 
- [x] You should see the Apple Pay sheet with the pledge amount in the project's currency
- [x] There should be a shipping item with a cost that corresponds to the default shipping rule cost
- [x] The line item should be the reward title
4. Change your shipping rule, then tap the Apple Pay button:
- [x] The updated shipping rule cost should be reflected, and the total should be correct
- [x] Both should be in the project's currency
- [x] the line item description should be the reward title

# ⏰ TODO

- [x] end-to-end testing. Right now the mutation is returning an error.


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -222,13 +253,79 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
...
+    let paymentNetworkName = payment.token.paymentMethod.network?.rawValue
+    let transactionId = payment.token.transactionIdentifier
+
+    self.viewModel.inputs.paymentAuthorizationDidAuthorizePayment(paymentData: (
+      paymentDisplayName,
+      paymentNetworkName,
+      transactionId
+    ))
+
+    STPAPIClient.shared().createToken(with: payment) { [weak self] token, error in
```
justinswart: So we _could_ consider mocking this and adding it to the `Environment` for testing 🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -1,21 +1,44 @@
...
+  selectedShippingRule: ShippingRule?, merchantIdentifier: String
+)
+public typealias PKPaymentData = (displayName: String, network: String, transactionIdentifier: String)
+
 public protocol PledgeViewModelInputs {
+  func applePayButtonTapped()
   func configureWith(project: Project, reward: Reward)
+  func paymentAuthorizationDidAuthorizePayment(
+    paymentData:
+    (displayName: String?, network: String?, transactionIdentifier: String)
```
justinswart: This can be on the same line as the param name.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -74,13 +97,174 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+        $0.reward,
+        $0.pledgeAmount,
+        $0.selectedShippingRule,
+        PKPaymentAuthorizationViewController.merchantIdentifier
+      ) as PaymentAuthorizationData }
+
+    self.goToApplePayPaymentAuthorization = paymentAuthorizationData
+      .takeWhen(self.applePayButtonTappedSignal)
+
+    let stripeToken: Signal<String, Never> = self.tokenOrErrorSignal
```
justinswart: For simplicity seeing as we only care about this being `.left(token)` we could lose the `switch` and do:

```swift
let stripeToken: Signal<String, Never> = self.tokenOrErrorSignal
  .map { tokenOrError -> String? in
    if case let .left(token) = tokenOrError {
      return token
    }

    return nil
  }
  .skipNil()
```

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -74,13 +97,174 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+        default:
+          return nil
+        }
+      }.skipNil()
+
+    let stripeTokenError: Signal<Error?, Never> = self.tokenOrErrorSignal
+      .filter {
+        if case .right = $0 {
+          return true
+        } else {
```
justinswart: No need for the `else` here we can just `return false`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -74,13 +97,174 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+          return false
+        }
+      }
+      .map { tokenOrError -> Error? in
+        switch tokenOrError {
```
justinswart: Similar suggestion to above.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -74,13 +97,174 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+          backingData.reward,
+          backingData.pledgeAmount,
+          backingData.selectedShippingRule,
+          pkPaymentData,
+          stripeToken
+        )
+      }
+
+    let createApplePayBackingEvent = createApplePayBackingData
+      .map(createApplePayBackingInput(
```
justinswart: This fits as:
```swift
.map(
  createApplePayBackingInput(for:reward:pledgeAmount:selectedShippingRule:pkPaymentData:stripeToken:)
)
```

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -100,6 +107,15 @@ final class PledgePaymentMethodsViewController: UIViewController {
         self.messageDisplayingDelegate?.pledgeViewController(self, didErrorWith: errorMessage)
       }
 
+    self.viewModel.outputs.notifyDelegateApplePayButtonTapped
+      .observeForUI()
+      .observeValues { [weak self] in
+        guard let self = self else { return }
+
+        self.delegate?
```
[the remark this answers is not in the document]
ifbarrera: This is weird I don't remember doing this. Would SwiftFormat do this?

Library/PKPaymentRequest+Helpers.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    paymentSummaryItems.append(
+      PKPaymentSummaryItem(
+        label: reward.title ?? project.name,
+        amount: NSDecimalNumber(value: pledgeAmount),
+        type: .final
+      )
+    )
+
+    if let selectedShippingRule = selectedShippingRule, selectedShippingRule.cost != 0.0 {
```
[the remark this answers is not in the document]
ifbarrera: I just copied and pasted what was previously in `DeprecatedRewardPledgeViewModel` here. I think for the purposes of shipping cost, `!= 0.0` and ` > 0 ` are essentially equivalent, but I can change it to `> 0 `.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
ifbarrera: I think it's not necessary to do that as long as the outputs from the completion block all go directly to the view model so they can be tested. With that in mind, I think I'm actually going to refactor this and send the `token` and `error` from this completion block directly to the view model instead of turning them into an `Either<String, Error?>` type.

It's also nice from the dependency tree perspective to keep all our interactions with 3rd-party APIs at the `Framework` level, which minimizes dependencies in the `Library` framework, theoretically making it easier to extract to a shared library in the future if necessary.

Library/ViewModels/PledgeViewModel.swift
ifbarrera: SwiftFormat is doing all of this.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: Sounds good!

PR #834, opened 2019-09-10
[NT-145] Shipping rules search

dusi:
# 📲 What

Adds the ability to search shipping rules picker.

# 🤔 Why

Shipping rules picker usually comes with a very long list of locations that we want our users to be able to filter down based on the search term.

# 🛠 How

The VM had to be completely re-written in order to support search & selection.

😱 Please note that this PR is marked as Big, because there's a lot of tests coverage. The tests should be very descriptive and include a lot of comments.

# 👀 See

Untitled

# ✅ Acceptance criteria

- [ ] Selecting a shipping rule from the list (that has previously NOT been filtered) updates selected shipping rule properly and dismisses the picker
- [ ] Selecting shipping rule from the list (that has been previously filtered/searched) updates selected shipping rule properly and dismisses the picker
- [ ] Searching the list filters down the list to only the shipping rules that start with the searched query
- [ ] Searching the list scrolls selected shipping rule to the top (if it's included in the filtered list)
- [ ] Opening up the picker always scrolls to the initial shipping rule (no matter where in array it was (first, middle, last)
- [ ] Clearing up the search input restores the original (unfiltered) list
- [ ] Tapping the `Cancel` button dismisses the picker


Kickstarter-iOS/Views/Controllers/ShippingRulesTableViewController.swift
```diff
@@ -50,8 +60,8 @@ final class ShippingRulesTableViewController: UITableViewController {
     self.viewModel.outputs.deselectCellAtIndex
       .map { IndexPath(row: $0, section: 0) }
       .observeForUI()
-      .observeValues { [weak self] indexPath in
-        self?.tableView.cellForRow(at: indexPath)?.accessoryType = .none
+      .observeValues { [weak self] _ in
+        self?.tableView.visibleCells.forEach { $0.accessoryType = .none }
```
dusi: This simplified a lot of logic, because we don't have to keep in track previously selected index (this gets really complicated in case the list is filtered causing that previous index to be out of sync).

To simplify this behaviour we simply make sure to deselect cells first, then selecting the latest index.

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
 
-    let selectedIndex = Signal.combineLatest(
-      dataInitial,
-      self.didSelectShippingRuleAtIndexProperty.signal.skipNil()
-    )
-    .map { data, newIndex in (data.1.firstIndex(of: data.2), newIndex) }
-    .filter { oldIndex, newIndex in oldIndex != newIndex }
-    .map(second)
+    let searchText = self.searchTextDidChangeProperty.signal
+      .ksr_debounce(.milliseconds(100), on: AppEnvironment.current.scheduler)
```
dusi: Adding a little bit of delay so that this functionality doesn't cause a bottle neck on slower devices when typing really fast.

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
+      reloadDataFiltered,
+      reloadDataSelected
     )
-    .skipRepeats()
 
-    self.deselectCellAtIndex = self.selectCellAtIndex
-      .combinePrevious()
-      .map(first)
+    self.scrollToCellAtIndex = initialData
+      .take(first: 1)
```
ifbarrera: Just curious, is there ever a scenario where `initialData` would fire multiple times?

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
-    .map { data, index in (data.0, data.1, index) }
-    .filter { _, shippingRules, index in index >= 0 && index < shippingRules.count }
-    .map { project, shippingRules, index in (project, shippingRules, shippingRules[index]) }
+    let selectedIndex = self.didSelectShippingRuleAtIndexProperty.signal
+      .skipNil()
 
-    let reloadDataWithShippingRulesInitial = dataInitial
-      .map(shippingRuleData(for:shippingRules:selectedShippingRule:))
-      .map { ($0, true) }
+    self.deselectCellAtIndex = selectedIndex
```
ifbarrera: It looks like in `ShippingRulesTableViewController` where we observe `deselectCellAtIndex` we don't actually use the index anymore and just deselect all cells. Can we rename this output to reflect that? Maybe  something like `let deselectCells: Signal<Void, Never>`?

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
 
-    self.deselectCellAtIndex = self.selectCellAtIndex
-      .combinePrevious()
-      .map(first)
+    self.scrollToCellAtIndex = initialData
+      .take(first: 1)
+      .map { _, shippingRules, selectedShippingRule in shippingRules.firstIndex(of: selectedShippingRule) }
+      .skipNil()
+
+    self.selectCellAtIndex = self.deselectCellAtIndex
```
ifbarrera: This is very confusing. Can't it just be `self.selectedCellAtIndex = selectedIndex`?

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
-    self.scrollToCellAtIndex = Signal.combineLatest(
-      selectedShippingRuleIndexInitial,
-      reloadDataWithShippingRulesInitial
+    let reloadDataSelected = Signal.merge(
+      initialData,
+      filteredData
     )
-    .map(first)
+    .takePairWhen(selectedIndex)
+    .map { data, selectedIndex in (data.0, data.1, selectedIndex) }
```
ifbarrera: It's a little hard to follow what's happening when the tuples elements aren't named. It might be helpful to `typealias` the tuple as `ShippingRulesData` with named elements so this would read:

`.map { data, selectedIndex in (data.project, data.shippingRules, selectedIndex) }`

Library/ViewModels/ShippingRulesViewModel.swift
```diff
@@ -32,70 +33,73 @@ public protocol ShippingRulesViewModelType {
...
 
-    self.scrollToCellAtIndex = Signal.combineLatest(
-      selectedShippingRuleIndexInitial,
-      reloadDataWithShippingRulesInitial
+    let reloadDataSelected = Signal.merge(
+      initialData,
+      filteredData
     )
-    .map(first)
+    .takePairWhen(selectedIndex)
```
ifbarrera: I noticed that above you define `selectedShippingRule`. I think you then might be able to simplify this:

```
    let reloadDataSelected = Signal.merge(
      initialData,
      filteredData
    )
    .takePairWhen(selectedShippingRule)
    .map { data, selectedShippingRule in (data.0, data.1, selectedShippingRule) }
    .map(shippingRulesData(project:shippingRules:selectedShippingRule:))
    .map { ($0, false) }
```

Library/ViewModels/ShippingRulesViewModel.swift
dusi: It's not, I was being paranoid...I can remove

Library/ViewModels/ShippingRulesViewModel.swift
dusi: Oh, nice! 😉

PR #836, opened 2019-09-11
[NT-146] Shipping rules search visual tweaks

dusi:
# 📲 What

We've noticed that the search modal looks slightly different from the specs

# 🤔 Why

Design, amirite :)

# 🛠 How

We had to
* Disable row separators
* Set navigation bar's background color
* Set navigation controller's view tint color (in order to affect both the `Cancel` button and the `checkmark` accessory views on the table view cell

# 👀 See

Please note the differences are very subtle so you might need to zoom in.

* The navigation bar background color should now be white
* The `Cancel` button and `Checkmark` should now use the right green color
* The table should no longer use row separators

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-09-11 at 2 33 11 PM | Screen Shot 2019-09-11 at 2 30 17 PM |

# ✅ Acceptance criteria

- [x] The navigation bar background color uses the right color (white)
- [x] The `Cancel` button and `Checkmark` should now use the right green color (matching the new pledge screen green)
- [x] The table should no longer use row separators


Kickstarter-iOS/Views/Controllers/ShippingRulesTableViewController.swift
```diff
@@ -41,11 +41,19 @@ final class ShippingRulesTableViewController: UITableViewController {
         action: #selector(ShippingRulesTableViewController.dismissViewController)
       )
 
+    _ = self.navigationController?.navigationBar
+      ?|> \.barTintColor .~ UIColor.white
+      ?|> \.isTranslucent .~ false
+
+    _ = self.navigationController?.view
+      ?|> \.tintColor .~ UIColor.ksr_green_500
```
dusi: I'm starting to thing we should set this globally in `AppDelegate` on the `window` object since we now have parts of the app using wrong tint color...thoughts: @justinswart @ifbarrera @Scollaco @cdolm92 ?

PR #838, opened 2019-09-13
Manage pledge configuration outputs

Scollaco:
# 📲 What
 Adds outputs on the `ManageViewPledgeViewController` in order to configure each section (subview) separately.

# 🤔 Why
This will avoid merge conflicts (or make them less bad) when more than one engineer start working on that screen.

# 🛠 How
- Implementing an output for each subview
- Adding tests for the outputs


Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -3,15 +3,15 @@ import Library
 import Prelude
 import UIKit
 
-final class ManagePledgeViewController: UIViewController {
+final class ManageViewPledgeViewController: UIViewController {
```
ifbarrera: I think the old name is better 😬 what made you change it to `ManageViewPledgeViewController`?

Scollaco: The main reason is that we will use the same screen to Manage and View pledge.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -93,33 +93,37 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .map(unpack)
 
     let goToManagePledge = ctaButtonTapped
-      .filter { canShowManageViewPledgeScreen($0.0, state:$0.2) }
+      .filter { canShowManageViewPledgeScreen($0.0, state: $0.2) }
```
Scollaco: The changes on this file were made by SwiftFormat.

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
ifbarrera: Okay makes sense! 👍

Library/ViewModels/ManageViewPledgeViewModel.swift
```diff
@@ -0,0 +1,66 @@
...
+
+public protocol ManageViewPledgeViewModelType {
+  var inputs: ManageViewPledgeViewModelInputs { get }
+  var outputs: ManageViewPledgeViewModelOutputs { get }
+}
+
+public final class ManageViewPledgeViewModel:
+  ManageViewPledgeViewModelType, ManageViewPledgeViewModelInputs, ManageViewPledgeViewModelOutputs {
+  public init() {
+    let projectAndReward = self.projectAndRewardSignal
```
ifbarrera: I think we _usually_ use `.combineLatest` here to ensure that both signals have fired at least once.

Library/ViewModels/ManageViewPledgeViewModelTests.swift
```diff
@@ -0,0 +1,83 @@
...
+
+  func testNavigationBarTitle_LiveProject() {
+    self.title.assertDidNotEmitValue()
+
+    let project = Project.template
+    self.vm.inputs.configureWith(project, reward: .template)
+
+    self.vm.inputs.viewDidLoad()
+
+    self.title.assertValue(Strings.Manage_your_pledge())
```
ifbarrera: Can we use the actual values of these strings in here? It's better to be more explicit about what you're testing against:

```
self.title.assertValue("Manage your pledge")
```

Library/ViewModels/ManageViewPledgeViewModelTests.swift
```diff
@@ -0,0 +1,83 @@
...
+  func testNavigationBarTitle_FinishedProject() {
+    self.title.assertDidNotEmitValue()
+
+    let finishedProject = Project.template
+      |> \.state .~ .successful
+    self.vm.inputs.configureWith(finishedProject, reward: .template)
+
+    self.vm.inputs.viewDidLoad()
+
+    self.title.assertValue(Strings.View_your_pledge())
```
ifbarrera: Here too 🙏

Library/ViewModels/ManageViewPledgeViewModel.swift
Scollaco: The problem with `.combineLatest` here is that will change the variable's type by combining the signals type. So instead of ` Signal<(Project, Reward), Never>`, it would become ` Signal<((Project, Reward), Void), Never>`. The `.takeWhen` guarantees that the .`projectAndRewardSignal` only emits when `viewDidLoad` after emit without changing the original type. Does it make sense?

ifbarrera: It does change the signature, but that's not usually an issue because we can do `.map(first)`.

I _think_ problem with using `.takeWhen` is it that it doesn't wait for both signals to fire at least once 🤔 from the description of `.takeWhen`: `"Emits the most recent value of `self` when `other` emits."`.

If the most recent of `self` (ie. `projectAndRewardSignal`) isn't there because that signal hasn't fired once, then it wouldn't work. Most of the time it would probably be fine, but we don't control when `viewDidLoad` gets called because it's a lifecycle function - I'm concerned there could be a race condition where somehow `viewDidLoad` gets called before we can call `configureWith(project:reward)`.

cc @justinswart any feedback/insight on this?

justinswart: Yes, @ifbarrera is correct here in that we would want to use `combineLatest` because that negates the concern about the order in which these signals emit. Using `takeWhen` we must ensure that `configure(with:)` is called first so that it was a value to _take_ when `viewDidLoad` emits. By combining the two we're allowed to call these inputs in any order. Our preference and the pattern we typically use for this is `combineLatest` 👍

ifbarrera: @Scollaco can you please make sure to make this change in a subsequent PR?

PR #841, opened 2019-09-14
[NT-308] Update strings and nokogiri gem

justinswart:
# 📲 What

Just some upkeep:

- Performs a general `make strings` and records new snapshots.
- Updates the `nokogiri` gem which `Danger` uses to silence the security warning on the repo.

# 🤔 Why

Now that we only include related string changes in our PRs we do need to run `make strings` from time-to-time to ensure that we're seeing the latest translations in the app.

I've also updated the `nokogiri` gem as the current version has a security vulnerability.

# 🛠 How

- `make strings`, recorded new screenshots.
- `bundle update nokogiri`.

# ✅ Acceptance criteria

- [ ] Review new snapshots.
- [ ] Danger works as before.


justinswart:
@dusi this appears to be related to this issue: https://github.com/nicklockwood/SwiftFormat/issues/368

I think SwiftFormat is failing because `Strings.swift` is the only file that was modified but is also an ignored file.

According to the issue I linked to this was fixed in SwiftFormat `0.39.4`, I've updated the binary to `0.40.11` but it still seems to be happening.

Does Danger definitely make use of the `swiftformat` executable in our `/bin` directory?

justinswart:
Lol, nevermind, the other linked issue on that issue is the same problem and describes a workaround: https://github.com/nicklockwood/SwiftFormat/issues/470#issuecomment-517936600

Dangerfile
```diff
@@ -9,5 +9,6 @@ swiftlint.lint_files inline_mode: true
 
 # SwiftFormat
 swiftformat.binary_path = "bin/swiftformat"
-swiftformat.additional_args = "--config .swiftformat --swiftversion 4.2"
+swiftformat.additional_args = "--config .swiftformat --swiftversion 5"
```
justinswart: Not sure why this was still on `4.2`.

Also added our two generated files to be excluded by Danger here to avoid SwiftFormat failing saying that it has nothing to format.

justinswart:
Side-note: seeing this warning on Circle too:

image

justinswart:
@ifbarrera well-spotted. Looks like possibly an encoding issue, let me check it out 🤔

ifbarrera:
@justinswart I think you need to run swiftformat though

PR #844, opened 2019-09-17
NT-206 - [1/4] Reward received toggle - UI

dusi:
# 📲 What

Basic UI for the reward received toggle.

# 🤔 Why

Step one in showing the toggle is to actually have something to show 😉 so this is the UI itself.

# 🛠 How

It's been pretty challening making this work with a stack view - apparently multi-line labels and stack views are [hard](https://useyourloaf.com/blog/stack-views-and-multi-line-labels) - who'd knew 🤦‍♂.

Because of that we're simply using auto layout and we're not adjusting axis for accessibility traits. Even though this causes the lack of axis adjustment often seen in the Settings.app (see bellow) where the toggle would align on the next line it's a small trade off to make. The toggle's width makes it acceptable staying at it's position at all times while adjusting just the label. We can potentially address this in the future with more customized control.

IMG_1D53F745316F-1

# 👀 See

| `.large` | `.extraExtraExtraLarge` | `.accessibilityExtraExtraExtraLarge` |
| --- | --- | --- |
| Screen Shot 2019-09-16 at 4 26 14 PM | Screen Shot 2019-09-16 at 4 26 29 PM | Screen Shot 2019-09-16 at 4 26 33 PM |

| `.large` | `.extraExtraExtraLarge` | `.accessibilityExtraExtraExtraLarge` |
| --- | --- | --- |
| Screen Shot 2019-09-16 at 4 27 03 PM | Screen Shot 2019-09-16 at 4 27 22 PM | Screen Shot 2019-09-16 at 4 27 25 PM |

| `.large` | `.extraExtraExtraLarge` | `.accessibilityExtraExtraExtraLarge` |
| --- | --- | --- |
| Screen Shot 2019-09-16 at 4 30 14 PM | Screen Shot 2019-09-16 at 4 30 21 PM | Screen Shot 2019-09-16 at 4 30 23 PM |

# ♿️ Accessibility 

`VoiceOver` accessibility will be addressed in a follow up PR since it showed up quite challenging.

- [x] Dynamic type resizes properly

# ✅ Acceptance criteria

- [x] Rewards received toggle UI is showing when navigating to manage pledge screen (`Profile > Backed project > Manage`

# ⏰ TODO

(will be addressed in follow up PRs)
- [ ] VoiceOver support
- [ ] The actual business logic
- [ ] Strings


Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewController.swift
```diff
@@ -0,0 +1,18 @@
+import Library
+import Prelude
+import UIKit
+
+final class ManageViewPledgeRewardReceivedViewController: ToggleViewController {
```
dusi: Since we're using `ToggleViewController` as a base class it could be reused in the future and this controller should be able to provide it's own logic (which API call will be called) + styling.

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewControllerTests.swift
```diff
@@ -0,0 +1,35 @@
...
+  func testView() {
+    let devices = [Device.phone4_7inch, Device.phone5_8inch, Device.pad]
+    combos([Language.en], devices).forEach { language, device in
+      withEnvironment(language: language) {
+        let controller = ManageViewPledgeRewardReceivedViewController.instantiate()
+        controller.toggle.setOn(true, animated: false)
+
+        let (parent, _) = traitControllers(device: device, orientation: .portrait, child: controller)
+
+        parent.view.frame.size.height = 60
```
dusi: More real time scenario and should make the snapshot smaller.

Kickstarter-iOS/Views/Controllers/ToggleViewController.swift
```diff
@@ -0,0 +1,62 @@
...
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  // MARK: - Lifecycle
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    _ = self.titleLabel
+      |> ksr_setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
```
dusi: Makes sure that the `titleLabel` does not push the switch off screen and rather pushes its own text on next line.

Kickstarter-iOS/Views/Controllers/ToggleViewControllerTests.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    super.tearDown()
+  }
+
+  func testView() {
+    let devices = [Device.phone4_7inch, Device.pad]
+    combos([Language.en], devices).forEach { language, device in
+      withEnvironment(language: language) {
+        let controller = ToggleViewController.instantiate()
+        controller.titleLabel.text = "Title for testing purposes only"
```
dusi: Since the base class doesn't provide default title or ON/OFF state we're setting it here simply for the purposes of our tests.

Kickstarter-iOS/Views/Controllers/ToggleViewControllerTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        let (parent, _) = traitControllers(device: device, orientation: .portrait, child: controller)
+
+        parent.view.frame.size.height = 60
+
+        FBSnapshotVerifyView(parent.view, identifier: "lang_\(language)_device_\(device)")
+      }
+    }
+  }
+
+  func testViewSupportsDynamicType() {
```
dusi: I've also decided to add Dynamic type snapshot in order to keep a regression test for the resizing of the label. This should only be done on the base class (no need for subclasses to handle this)

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -98,6 +132,27 @@ final class ManageViewPledgeViewController: UIViewController {
...
+      _ = ([viewController.view], self.rootStackView)
+        |> ksr_addArrangedSubviewsToStackView()
+
+      viewController.didMove(toParent: self)
+    }
+  }
+
+  private func setupConstraints() {
+    NSLayoutConstraint.activate([
+      self.rootStackView.widthAnchor.constraint(equalTo: self.rootScrollView.widthAnchor)
```
ifbarrera: You *usually* need to also set the width of the scroll view to be the equal to the width of the parent VC 🤔

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -123,7 +178,18 @@ final class ManageViewPledgeViewController: UIViewController {
 
 // MARK: Styles
 
-public let viewStyle: ViewStyle = { (view: UIView) in
+private let rootScrollViewStyle = { (scrollView: UIScrollView) in
+  scrollView
+    |> \.alwaysBounceVertical .~ true
+}
+
+private let rootStackViewStyle = { (stackView: UIStackView) in
```
ifbarrera: In `PledgeViewController` we also have a `rootStackViewStyle`: https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/Views/Controllers/PledgeViewController.swift#L402

I wonder if it be worthwhile to extract this style into our `CheckoutStyles.swift` file so they can be shared? I'd imagine that the pledge screen and the manage pledge screen should probably have consistent root stack view styles regardless.

Kickstarter-iOS/Views/Controllers/ToggleViewController.swift
```diff
@@ -0,0 +1,62 @@
...
+    _ = (self.toggle, self.view)
+      |> ksr_addSubviewToParent()
+
+    NSLayoutConstraint.activate([
+      self.titleLabel.leftAnchor.constraint(equalTo: self.view.leftAnchor),
+      self.titleLabel.topAnchor.constraint(equalTo: self.view.topAnchor),
+      self.titleLabel.bottomAnchor.constraint(equalTo: self.view.bottomAnchor),
+      self.titleLabel.centerYAnchor.constraint(equalTo: self.toggle.centerYAnchor),
+      self.titleLabel.rightAnchor.constraint(lessThanOrEqualTo: self.toggle.leftAnchor),
+      self.toggle.rightAnchor.constraint(equalTo: self.view.rightAnchor)
```
ifbarrera: Does the toggle need a vertical position constraint? 🤔

Library/Styles/CheckoutStylesTests.swift
```diff
@@ -39,6 +39,15 @@ final class CheckoutStylesTests: XCTestCase {
     XCTAssertEqual(9, stackView.spacing)
   }
 
+  func testCheckoutSwitchControlStyle() {
```
ifbarrera: I really don't think these tests are necessary 😬 Anytime we decide to change a small aspect of the checkout UI these tests break - it seems like it adds unnecessary maintenance for something that's already covered with other tests (snapshot tests) and offers only a marginal increase in test coverage.

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -68,6 +93,15 @@ final class ManageViewPledgeViewController: UIViewController {
     _ = self.closeButton
       |> \.accessibilityLabel %~ { _ in Strings.Dismiss() }
       |> \.width .~ Styles.minTouchSize.width
+
+    _ = self
```
Scollaco: The title should be handled by the view model, because it depends on the project state (live vs finished)

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewControllerTests.swift
Scollaco: I wonder if would make sense for us to wait until the manage/view pledge screen is finished in order to create snapshot test for just that viewController instead of one for each child. This way we would avoid testing the same screen twice 🤔

Library/Styles/CheckoutStyles.swift
```diff
@@ -109,6 +109,12 @@ public let checkoutStackViewStyle: StackViewStyle = { (stackView: UIStackView) i
     |> \.spacing .~ (Styles.grid(1) + Styles.gridHalf(1))
 }
 
+public let checkoutSwitchControlStyle: SwitchControlStyle = { switchControl in
+  switchControl
+    |> \.onTintColor .~ UIColor.ksr_green_500
+    |> \.tintColor .~ UIColor.ksr_green_500
```
Scollaco: Is the `tintColor` supposed to be green? Our app usually adopts grey for the `off` state. Maybe a `ksr_grey_500` or darker would fit better?

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewControllerTests.swift
dusi: Yeah, I'm not sure which side I lean towards more...I think each controller should be testable by itself but agree that as soon as we add manage pledge screen tests it will be a duplicate. 

We already do this with `SortPagerViewControllerTests` or `ProjectNavBarViewControllerTests` which test just the menu itself and then we also have a test for the container `ProjectPamphletContentViewControllerConversionTests`.

I believe testing each controller is more correct, since its parent can additionally style it or decide what it needs to do with it.

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
dusi: Yeah, I know we've done it before but it didn't seem necessary this time ... same on `PledgeVC` 🤔 🐒

Kickstarter-iOS/Views/Controllers/ToggleViewController.swift
dusi: `centerY`?

Library/Styles/CheckoutStylesTests.swift
dusi: **Deleted**

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
dusi: LOL, there's so many repeated styles..I'll try to consolidate these into something more reusable 😄

Library/Styles/CheckoutStyles.swift
dusi: Yeah, the design was kinda weird because it wasn't using the native switch component...let me run this by Nneka

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
dusi: Apologies, completely missed this

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -112,7 +112,6 @@ final class PledgeCreditCardView: UIView {
 
 private let adaptableStackViewStyle: StackViewStyle = { stackView in
   stackView
-    |> \.backgroundColor .~ UIColor.white
```
dusi: From the docs...

> UIStackView is a non-drawing view, meaning that  drawRect() is never called and its background color is ignored. If you desperately want a background color, consider placing the stack view inside another UIView and giving that view a background color.

dusi: Basically all these `backgroundColor` styles were obsolete so I've removed them

ifbarrera:
> Trying to tidy up the styles into more reusable styles, I've discovered that some of the child controller (sub-controllers on the pledge screen) use different spacing for their sub-components (like stepper, buttons, labels, etc.)
> 
> I wonder if we should simplify this and make it consistent...This would mean the vertical spacing (in points) would be the same for these sections (highlighted bellow in red).
> 
> Screen Shot 2019-09-18 at 3 45 29 PM copy
> 
> Before and after would look something like this:
> 
> Before	After
> Screen Shot 2019-09-18 at 3 45 29 PM	Screen Shot 2019-09-18 at 3 51 15 PM
> This is something we would have probably spotted if we had a designer. It's always a good practice to keep things consistent between components that construct the same page because it's easier to thing about these in the future and makes changes easier to do.
> 
> Cc: @nnekab @jamielynnroth

Yeah I would vote we make them consistent.

Scollaco:
Yeah, I agree that we should be consistent. This would help us to define baseStyles and make the styling process much simpler without the need to tweak the spacing for each subview.

PR #845, opened 2019-09-17
[1/2] Pledge summary section

Scollaco:
# 📲 What
- Adds the Summary View section to the `ManageViewPledgeViewController`.

### Note:
- This PR doesn't contain the final strings. They will be added on the 2/2 part of this work.

[JIRA ticket](https://app.goabstract.com/projects/a94cda80-1e60-11e8-8903-0b9b4e24ea4b/branches/11b62619-b4c6-44c0-be45-e000aee1dbaa/collections/38089804-badd-474b-9658-001ea32d3df3?collectionLayerId=5d260584-6d56-46cb-9364-cd42063c6672)

# 🤔 Why

- This is the first part of the `ManagePledge` feature. More to come.

# 🛠 How
- Created a new `ManagePledgeSummaryView` class
- Created viewModel to handle the views outputs
- Created viewModel tests
- Added the new view to `ManageViewPledgeViewController`

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-09-17 at 14 50 41 | Simulator Screen Shot - iPhone 8 - 2019-09-17 at 14 48 59 |

# ♿️ Accessibility 
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

### With NativeCheckoutPledgeView feature enabled:
- [ ] Go to `Profile` and choose a backed project.
- [ ]  The Backer number should match the real value
- [ ]  The Baking date number should match the real date the user backed the project
- [ ]  The Pledge amount should match the real pledged value
- [ ]  The Shipping cost should match the real shipping cost
- [ ]  If backing has non-shippable reward, no reward or no location available, the Shipping location should be hidden
- [ ]  The Total amount should match the real total amount value


Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+  public var inputs: ManagePledgeSummaryViewViewModelInputs { return self }
+  public var outputs: ManagePledgeSummaryViewViewModelOutputs { return self }
+}
+
+private func shouldHideShippingLocationStackView(_ project: Project) -> Bool {
+  guard let backing = project.personalization.backing,
+    let _ = backing.rewardId else {
+    return true
+  }
+  if let reward = backing.reward {
```
Scollaco: Sometimes a backing can have `rewardId` but the `reward` can still be `nil`

Library/Format.swift
```diff
@@ -144,6 +144,12 @@ public enum Format {
     return NSAttributedString(string: Strings.plus_shipping_cost(shipping_cost: ""), attributes: attributes)
   }
 
+  public static func attributedAmount(
```
Scollaco: I created this function to return the attributed amount without the `+` sign

Kickstarter-iOS/Views/Controllers/ManagePledgeSummaryView.swift
```diff
@@ -0,0 +1,196 @@
...
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.minimumScaleFactor .~ 0.5
+}
+
+private let totalLabelStyle: LabelStyle = { label in
+  label
+    |> \.textColor .~ UIColor.black
+    |> \.font .~ UIFont.ksr_subhead().bolded
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.text %~ { _ in "Total amount" }
```
justinswart: Best to use `localizedString` directly until we have the string added.

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+  var shippingLocationText: Signal<String, Never> { get }
+  var totalAmountText: Signal<NSAttributedString, Never> { get }
+}
+
+public protocol ManagePledgeSummaryViewViewModelType {
+  var inputs: ManagePledgeSummaryViewViewModelInputs { get }
+  var outputs: ManagePledgeSummaryViewViewModelOutputs { get }
+}
+
+public class ManagePledgeSummaryViewViewModel: ManagePledgeSummaryViewViewModelType,
```
justinswart: I'm not sure when we started calling VM's VVM's 😅. I don't think we should be naming them `ViewViewModel`, the fact that this VM is used for a View and not a ViewController doesn't matter, we should just call it `ManagePledgeSummaryViewModel`, I think we should also add an investment day task to fix these names in the codebase for other VVMs.

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+  public init() {
+    let backing = self.projectSignal
+      .map { $0.personalization.backing }
+      .skipNil()
+
+    let projectAndBacking = self.projectSignal
+      .zip(with: backing)
+
+    self.backerNumberText = backing
+      .map { "Backer #\($0.sequence)" }
```
justinswart: Please use `localizedString` until this is translated.

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+  if let reward = backing.reward {
+    return !reward.shipping.enabled || reward.isNoReward
+  }
+
+  return backing.locationId.isNil
+}
+
+private func formattedPledgeDate(_ backing: Backing) -> String {
+  let formattedDate = Format.date(secondsInUTC: backing.pledgedAt, dateStyle: .long, timeStyle: .none)
+  return "As of \(formattedDate)"
```
justinswart: `localizedString`

Library/Format.swift
justinswart: I can't think of a reason that we'd want to do this 🤔

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+      country: project.country,
+      omitCurrencyCode: project.stats.omitUSCurrencyCode,
+      defaultAttributes: defaultAttributes,
+      superscriptAttributes: superscriptAttributes
+    ) else { return nil }
+
+  let combinedAttributes = defaultAttributes
+    .withAllValuesFrom(superscriptAttributes)
+
+  return Format.attributedAmount(attributes: combinedAttributes) + attributedCurrency
```
justinswart: Seeing as we don't need the `+` sign, we can just do:

```swift
private func attributedCurrency(with project: Project, amount: Double) -> NSAttributedString? {
  let defaultAttributes = checkoutCurrencyDefaultAttributes()
    .withAllValuesFrom([.foregroundColor: UIColor.ksr_green_500])
  let superscriptAttributes = checkoutCurrencySuperscriptAttributes()
  guard
    let attributedCurrency = Format.attributedCurrency(
      amount,
      country: project.country,
      omitCurrencyCode: project.stats.omitUSCurrencyCode,
      defaultAttributes: defaultAttributes,
      superscriptAttributes: superscriptAttributes
    ) else { return nil }

  return attributedCurrency
}
```

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+  guard
+    let attributedCurrency = Format.attributedCurrency(
+      shippingRuleCost,
+      country: project.country,
+      omitCurrencyCode: project.stats.omitUSCurrencyCode,
+      defaultAttributes: defaultAttributes,
+      superscriptAttributes: superscriptAttributes
+    ) else { return nil }
+
+  let combinedAttributes = defaultAttributes.merging(superscriptAttributes) { _, new in new }
```
justinswart: Using `.withAllValuesFrom(superscriptAttributes)` is a bit neater and does the same thing 👍

justinswart:
Do we want the date to truncate at these large sizes?

image

justinswart:
By the way, I noticed there's an autolayout warning when presenting the alert controller, it seems to be a UIKit bug: http://openradar.appspot.com/49289931

justinswart:
I think that we should use the `RewardPledgeNavigationController` to present this view, that already has the styling applies to its navigation bar.

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
Scollaco: lol it makes sense. I will rename it

Library/ViewModels/ManagePledgeSummaryViewViewModel.swift
Scollaco: Much nicer than my solution :)

Kickstarter-iOS/Views/Controllers/ManagePledgeSummaryView.swift
Scollaco: Actually after merging master I noticed that most of the strings were already there!

Library/ViewModels/ManagePledgeSummaryViewModel.swift
```diff
@@ -0,0 +1,133 @@
...
+    self.totalAmountText = projectAndBacking
+      .combineLatest(with: shippingAmount)
+      .map(unpack)
+      .map { project, backing, shippingAmount in
+        attributedCurrency(with: project, amount: backing.amount + shippingAmount)
+      }
+      .skipNil()
+
+    self.shippingLocationText = backing.ignoreValues()
+      .map { Strings.Shipping() + ": " + "Australia" }
```
Scollaco: Here the `Australia` is hardcoded because we will use a Backing property called `locationName`. This is done in another branch (for the 2/2 PR)

Library/Format.swift
justinswart: We can probably remove this function now, right?

Library/ViewModels/ManagePledgeSummaryViewModel.swift
justinswart: Ok, I think we want to add a new string here:

`"Shipping: %{country_name}"`

Although it seems very much like what you have, by including the `:` in the string we give the translator the ability to move it around if they need to.

Scollaco: Yeah, before I implemented this, I checked out the Android codebase and that's how it was implemented. I can create a new string on the second part of this work, but I'm not sure if it would make a big difference.
cc @ginarovirosa

justinswart:
Will wait to hear back from @ginarovirosa on that string.

What about this comment? Is there a max number of lines set? I'm surprised it doesn't wrap on space.

Scollaco:
> Will wait to hear back from @ginarovirosa on that string.
> 
> What about [this comment](https://github.com/kickstarter/ios-oss/pull/845#issuecomment-533323868)? Is there a max number of lines set? I'm surprised it doesn't wrap on space.

It didn't have a number of lines set. I had set `adjustsFontSizeToFitWidth = true` instead. This is the result:

| lines = 0 | adjustsFontSizeToFitWidth = true |
| --- | --- | 
| Simulator Screen Shot - iPhone 8 - 2019-09-24 at 10 38 52 | Simulator Screen Shot - iPhone 8 - 2019-09-24 at 10 30 32

Seeing the difference I'm leaning towards the lines = 0 solution, because then it won't compromise the font size too much depending on the text.

Library/ViewModels/ManagePledgeSummaryViewModel.swift
ginarovirosa: hello @Scollaco @justinswart. Indeed having the `Shipping: %{country_name}` format makes it easier for the translators to work around country name prefixes.  Fwiw, yt does seem that the `Shipping: %{country_name}` variable is an old one. We use `%{country}` now. As you can see in this string for instance: `en -> pledges -> checkout_summary -> Shipping_to_country `

justinswart: Thanks @ginarovirosa!

PR #846, opened 2019-09-19
[NT-278] Manage pledge menu

ifbarrera:
# 📲 What

When tapping the "menu" button on the top right, an action sheet with various options is displayed.

# 🤔 Why

To allow a user to navigate to various other screens that allow you to modify your pledge.

# 🛠 How

Depending on the current state of the project, various options are shown.

# 👀 See

| Live Project | Ended Project | iPad |
| --- | --- | --- |
|Simulator Screen Shot - iPhone 8 - 2019-09-19 at 10 25 57|Simulator Screen Shot - iPhone 8 - 2019-09-19 at 10 25 49|Simulator Screen Shot - iPad Pro (12 9-inch) - 2019-09-19 at 14 40 36|

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Logged in, with both *Native Checkout* and *Native Pledge Screen* feature flags turned on, navigate to a live project that you have backed. Tap the "Manage" CTA on the project screen, then tap the navigation button on the top right of the "Manage" screen. You should see the action sheet pop up with 5 options: update pledge, change payment method, choose another reward, contact creator and cancel pledge. Tapping the "Cancel" button should dismiss the action sheet.
- [x] Logged in, with both *Native Checkout* and *Native Pledge Screen* feature flags turned on, navigate to an *ended* project that you have backed. Tap the "Manage" CTA on the project screen, then tap the navigation button on the top right of the "Manage" screen. You should see the action sheet pop up with 1 options: contact creator. Tapping the "Cancel" button should dismiss the action sheet.

# TODOS

- [x] get confirmation on the accessibility label for the menu navigation bar button


Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -68,6 +68,9 @@ final class ManageViewPledgeViewController: UIViewController {
     _ = self.closeButton
       |> \.accessibilityLabel %~ { _ in Strings.Dismiss() }
       |> \.width .~ Styles.minTouchSize.width
+
+    _ = self.menuButton
+      |> \.accessibilityLabel %~ { _ in Strings.Manage_your_pledge() }
```
dusi: I'm still wondering whether `Menu, button` would be better because you're already on a screen called `Manage your pledge` where you navigated by tapping `Manage` button 😄

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -104,15 +113,44 @@ final class ManageViewPledgeViewController: UIViewController {
...
+    self.viewModel.inputs.menuButtonTapped()
+  }
 
-    actionSheet.addAction(
-      UIAlertAction(title: Strings.Contact_creator(), style: .default)
+  private func showActionSheetMenuWithOptions(_ options: [ActionSheetMenuOption]) {
+    let actionSheet = UIAlertController.alert(
+      title: Strings.Select_an_option(),
+      preferredStyle: .actionSheet,
+      barButtonItem: self.menuButton
```
dusi: Cool, this works and makes it properly relocate when rotating the device 👍

Library/ViewModels/ManageViewPledgeViewModel.swift
```diff
@@ -3,15 +3,25 @@ import Prelude
 import ReactiveExtensions
 import ReactiveSwift
 
+public enum ActionSheetMenuOption: CaseIterable {
```
dusi: This naming seems like it's very high up the domain level (maybe even on a UIKit level)...how about we make it more local `ManagePledgeAlertAction`? This naming makes it explicitly clear that this is `Managing pledge` (since it's `public`) and also makes it clearer to UIKit component we're setting it on to `UIAlertAction`. What do you think?

dusi: I don't feel strongly about the `MenuOption` naming, just wanted to give another suggestion.

Library/ViewModels/ManageViewPledgeViewModel.swift
```diff
@@ -38,13 +48,30 @@ public final class ManageViewPledgeViewModel:
 
     self.configureRewardSummaryView = projectAndReward
       .map(second)
+
+    let project = projectAndReward.map(first)
+
+    self.showActionSheetMenuWithOptions = project
+      .takeWhen(self.menuButtonTappedSignal)
+      .map { project -> [ActionSheetMenuOption] in
```
dusi: This almost makes me think this should be an extension (maybe `private` extension on `Project`)...something along these lines..

```swift
private extension Project {
  func managePledgeOptions() -> [ActionSheetMenuOption] {
    switch self.state {
    case: .live: .allCases
    default: [.contactCreator]
    }
  }
}
```

Then we'd be able to do

```swift
.map(project.managePledgeOptions) // not sure if it does point free - wasn't using compiler when writing this
```

Dunno which one is better (this is pretty easy to read so I'd be fine with it).

ifbarrera: I'm not sure I agree with this - the list of alert options is very view-specific, I would be hesitant to couple it so directly with `Project` model. I'm going to leave it in the view model if that's ok.

PR #835, opened 2019-09-11
[  💳 Native Checkout ] Refresh Card List

cdolm92:
# 📲 What

User can now add new credit card from the pledge screen.
# 🤔 Why

Users can now add a preferred credit card without leaving the pledge flow.

# 🛠 How
We have a method called  `insertNewCard(_ newCard: GraphUserCreditCard.CreditCard)` that inserts the card in the first position of the `cardStackView`. We also made changes with how the card is configured. In the cards `configureWith()`  we now have the parameter `isNew: Bool` so that it is in the selected state when added.

# 👀 See
add-new-card

# ✅ Acceptance criteria

**Note:** Test this on staging. Make sure the recommended projects setting is disabled and use a test card from here: https://stripe.com/docs/testing

- [x] Navigate to pledge screen and tap on `add new card`
- [x] Enter test card details and tap `Done` when done.
- [x] You should now see the newly added card in a selected state

**Further Testing:**
- [x] You will need to add at least **two** test cards. Tap `Select` on more than one card. Only last card tapped should remain selected.
- [x] Deselect/Select a card

PR #847, opened 2019-09-19
[NT-303] Pledge Button

ifbarrera:
# 📲 What

Adds a `Pledge` button below the list of stored payment methods. This button is initially disabled until the user selects a credit card. Once a card is selected, the button becomes enabled.

# 🤔 Why

To allow users to pledge, and to simplify the pledging experience.

# 🛠 How

Some things to keep in mind:

- right now the experience is a bit jarring because the credit cards load in dynamically. This will be fixed with a loading state for the stored cards in a subsequent PR.
- right now the pledge button is initially disabled until a card is selected. A future PR will address this and automatically select the first card in the list.
- right now tapping `Select` on any of the cards will enable the `Pledge` button even though the card's `selected` state doesn't change. This is because toggling the card view's `selected` state is currently being worked on separately as part of another ticket

# 👀 See

8UpWVPUsVP

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Signed in, with the `Native Pledge Screen` feature flag turned ON, navigate to any project and select any reward. Scroll down and you should see the `Pledge` button below the list of stored cards, and it should be disabled. Tap `Select` on any of the stored cards - the `Pledge` button should become enabled.


Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -130,12 +152,16 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
 
-  // MARK: - Actions
+  // MARK: - Accessors
 
   @objc private func applePayButtonTapped() {
     self.viewModel.inputs.applePayButtonTapped()
   }
 
+  func updatePledgeButton(_ enabled: Bool) {
+    self.viewModel.inputs.updatePledgeButtonEnabled(isEnabled: enabled)
```
ifbarrera: The real "brains" of the pledge button enabled state is actually the `PledgeViewModel`. This is because we expect the pledge button's enabled state to be driven by a variety of factors: whether the reward minimum is met, whether a shipping location has been selected, etc. Since the `PledgeViewModel` is the only source of truth for this information, it needs to be the one to drive the pledge button's enabled state.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -53,6 +62,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
     self.notifyDelegateLoadPaymentMethodsError = storedCardsEvent
       .errors()
       .map { $0.localizedDescription }
+
+    self.notifyDelegateCreditCardSelected = self.creditCardSelectedSignal
```
ifbarrera: Right now this is very simple - has a credit card been selected? An upcoming PR will add the rest of the requirements for the pledge button being enabled: reward minimum is met, shipping rule selected, etc.

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -3,18 +3,25 @@ import Library
 import Prelude
 import UIKit
 
+protocol PledgeCreditCardViewDelegate: AnyObject {
+  func pledgeCreditCardViewSelected(
```
justinswart: This _might_ overlap with some of the work happening in #835.

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -3,18 +3,25 @@ import Library
...
-
   private let adaptableStackView: UIStackView = { UIStackView(frame: .zero) }()
+  weak var delegate: PledgeCreditCardViewDelegate?
   private let expirationDateLabel: UILabel = { UILabel(frame: .zero) }()
   private let imageView: UIImageView = { UIImageView(frame: .zero) }()
   private let labelsStackView: UIStackView = { UIStackView(frame: .zero) }()
   private let lastFourLabel: UILabel = { UILabel(frame: .zero) }()
   private let rootStackView: UIStackView = { UIStackView(frame: .zero) }()
   private let selectButton: UIButton = { UIButton(type: .custom) }()
+  private let viewModel: CreditCardCellViewModelType = CreditCardCellViewModel()
```
justinswart: I mentioned in #835 that we should probably rename this 😬

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -17,6 +21,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
   private lazy var cardsStackView: UIStackView = { UIStackView(frame: .zero) }()
   internal weak var delegate: PledgePaymentMethodsViewControllerDelegate?
   internal weak var messageDisplayingDelegate: PledgeViewControllerMessageDisplaying?
+  private lazy var pledgeButton: UIButton = { UIButton.init(type: .system) }()
```
justinswart: Think we want `UIButton(type: .custom)` to avoid the built-in flash animation.

ifbarrera:
> Cool cool, I think important to just note that there is payment method selection stuff being done in this PR and in #835, so we would want to align on that.
> 
> Also, the spacing here seems quite tight?
> 
> 
> And a final thought - should this button be in `PledgePaymentMethodsViewController` or `PledgeViewController` as its own nested VC? 🤔 😱

- the spacing is consistent with the spacing set on the root stack view. In this context I think it's ok because from the "information architecture" perspective, the "Pledge" button belongs in the same group as the credit cards and the `Other payment methods` label. Happy to double check with a designer though
- I guess it *could* be it's own nested VC, but since all of the business logic around the pledge button's state is happening in the `PledgeViewModel`, it didn't seem necessary to create a new VC for such a simple view with no business logic of it's own. In the future if we wanted to reuse the `Pledge` button I could see a case for extracting it into it's own VC/VM pair, but at this stage it didn't seem necessary. Additionally, the visual positioning of the button needs to be part of the `Other payment methods` "group", and I felt that the easiest way to accomplish that would be to have them be part of the same root stack view.

Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
ifbarrera: Yeah I agree - I really don't think we should have used the same view model as the one in the `Settings` screen at all though :\

justinswart: Oh I had no idea it was being used in both places.

PR #848, opened 2019-09-20
[NT-262] Remember this card UI

justinswart:
# 📲 What

Adds a _Remember this card_ switch to `AddNewCardViewController` when being presented from `PledgeViewController`.

**Note:** VM binding and GraphQL requests will be added in upcoming PR.

# 🤔 Why

This allows the user to indicate that a card used during checkout should be stored for future backings.

# 🛠 How

- Added `AddNewCardIntent`.
- Show and hide the reusable card switch based on this value.

# 👀 See

| Settings | PledgeViewController |
| --- | --- |
| image | image |

# ♿️ Accessibility 

- [ ] Works with VoiceOver

# ✅ Acceptance criteria

- [x] Switch is shown when `AddNewCardViewController` is presented from `PledgeViewController`.
- [x] Switch is hidden when `AddNewCardViewController` is presented from Settings.

# ⏰ TODO

- [ ] VM bindings and GraphQL mutations in upcoming PR.


Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -136,11 +141,22 @@ internal final class AddNewCardViewController: UIViewController,
...
       |> \.returnKeyType .~ .done
+
+    _ = self.reusableCardLabel
+      |> \.text %~ { _ in Strings.Remember_this_card() }
+
+    _ = self.reusableCardStackView
+      |> \.alignment .~ .center
+
+    _ = self.reusableCardSwitch
+      |> baseSwitchControlStyle
```
dusi: Ideally we would only have 1 accessible element (the toggle - whose focus area would be the whole cell) but because we're faking a table view cell here we don't get that functionality...so because of that I think we should also add VoiceOver label on the switch that will simply duplicate the label but since it's a `UISwitch` it will also use `double tap to toggle` hint by default.

```suggestion
      |> baseSwitchControlStyle
      |> \.accessibilityLabel %~ { _ in Strings.Remember_this_card() }
```
(not sure if this will pass SwiftFormat)

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -136,11 +141,22 @@ internal final class AddNewCardViewController: UIViewController,
...
   }
 
   override func bindViewModel() {
     super.bindViewModel()
 
+    self.reusableCardStackViewContainer.rac.hidden = self.viewModel.outputs.reusableCardSwitchIsHidden
```
dusi: Should we order this alphabetically (so before the `saveButtonIsEnabled` signal)?

Library/AddNewCardIntent.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public enum AddNewCardIntent: Equatable {
+  case pledgeView
```
dusi: Do we even need the `View` suffix? Simply just `pledge` 🤔

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -36,6 +37,7 @@ public protocol AddNewCardViewModelOutputs {
   var dismissKeyboard: Signal<Void, Never> { get }
   var paymentDetails: Signal<PaymentDetails, Never> { get }
   var paymentDetailsBecomeFirstResponder: Signal<Void, Never> { get }
+  var reusableCardSwitchIsHidden: Signal<Bool, Never> { get }
```
dusi: Does this name not match `reusableCardStackViewContainer` intentionally for it being too long?

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
dusi: Also note that there seems to be an issue with `UISwitch` and right alignment..I've ran into it in my PR - https://github.com/kickstarter/ios-oss/pull/850

Basically even if you align right edges the switch still overlaps a little. Checked this out in view inspector.

Screen Shot 2019-09-23 at 11 09 36 AM

Basically what helped was to align left edge of the switch to the right edge of superview minus the switch width.

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -22,6 +22,10 @@ internal final class AddNewCardViewController: UIViewController,
   @IBOutlet private var creditCardTextField: STPPaymentCardTextField!
   @IBOutlet private var creditCardValidationErrorLabel: UILabel!
   @IBOutlet private var creditCardValidationErrorContainer: UIView!
```
cdolm92: alphabetize

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -23,7 +23,7 @@ internal final class PaymentMethodsViewModelTests: TestCase {
 
     self.vm.outputs.editButtonIsEnabled.observe(self.editButtonIsEnabled.observer)
     self.vm.outputs.editButtonTitle.observe(self.editButtonTitle.observer)
-    self.vm.outputs.goToAddCardScreen.observe(self.goToAddCardScreen.observer)
+    self.vm.outputs.goToAddCardScreenWithIntent.observe(self.goToAddCardScreenWithIntent.observer)
```
cdolm92: alphabetize

PR #849, opened 2019-09-21
[NT-312] Pledge amount invalid input color

dusi:
# 📲 What

* When the input value is lower than the minimum or higher than the maximum the text turns red
* When user dismisses keyboard by tapping anywhere on the screen (and the input is still invalid) it stays this way (instead of clamping it to either min or max)

# 🤔 Why

Some UX improvements to amount input

# 🛠 How

See inline comments

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-09-20 at 11 17 03 PM | Screen Shot 2019-09-20 at 11 18 46 PM |
| Screen Shot 2019-09-20 at 11 17 19 PM | Screen Shot 2019-09-20 at 11 18 57 PM |

# ✅ Acceptance criteria

**TextField input**

- [ ] Entering value that is smaller than minimum pledge value should make the text red
- [ ] Entering value that is larger than maximum pledge value should make the text red
- [ ] Entering `0` disables the decrement button of the stepper
- [ ] Entering `Double max` disables the increment button of the stepper (this will be harder to test so it might be sufficient to review snapshots and integration tests)

**Stepper input**

- [ ] Entering value that is smaller than minimum pledge value should make the text red
- [ ] Entering value that is larger than maximum pledge value should make the text red
- [ ] Entering `0` disables the decrement button of the stepper
- [ ] Entering `Double max` disables the increment button of the stepper (this will be harder to test so it might be sufficient to review snapshots and integration tests)

It's possible that the last criteria will be better tested when we remove decimal support in some of the upcoming work.

# ⏰ TODO

- [ ] Update Cartfile when https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/79 gets merged
- [ ] Accessibility will be handled separately


Kickstarter-iOS/TestHelpers/TraitCollection.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
+
+extension UITraitCollection {
+  static let allCases: [UITraitCollection] = [
```
dusi: Just a test helper for allowing us testing large text better.

Kickstarter-iOS/Views/AmountInputView.swift
```diff
@@ -112,15 +112,13 @@ private let labelStyle: LabelStyle = { (label: UILabel) in
     |> \.font .~ UIFont.ksr_body()
     |> \.adjustsFontForContentSizeCategory .~ true
     |> \.textAlignment .~ NSTextAlignment.right
-    |> \.textColor .~ UIColor.ksr_green_500
```
dusi: These are now set using VM's outputs

Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -21,11 +21,7 @@ final class PledgeAmountViewController: UIViewController {
   private lazy var amountInputView: AmountInputView = { AmountInputView(frame: .zero) }()
   private lazy var titleLabel: UILabel = { UILabel(frame: .zero) }()
   private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
-  private lazy var spacer: UIView = {
-    UIView(frame: .zero)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
-  }()
-
+  private lazy var horizontalSpacer: UIView = { UIView(frame: .zero) }()
```
dusi: No real need for auto layout anymore...we're using spacing (where we have previously used auto layout)...this should be simpler and more readable.

Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -40,16 +36,14 @@ final class PledgeAmountViewController: UIViewController {
       |> ksr_addSubviewToParent()
       |> ksr_constrainViewToEdgesInParent()
 
-    _ = ([self.titleLabel, self.adaptableStackView], self.rootStackView)
+    _ = ([self.titleLabel, self.adaptableStackView, UIView(frame: .zero)], self.rootStackView)
```
dusi: This makes sure that the vertical stack view will always align it's subviews to top (vertically)

Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -40,16 +36,14 @@ final class PledgeAmountViewController: UIViewController {
...
       |> ksr_addArrangedSubviewsToStackView()
 
-    _ = ([self.stepper, self.spacer, self.amountInputView], self.adaptableStackView)
+    _ = ([self.stepper, self.horizontalSpacer, self.amountInputView], self.adaptableStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
     self.amountInputView.textField.delegate = self
 
-    self.spacer.widthAnchor.constraint(greaterThanOrEqualToConstant: Styles.grid(3)).isActive = true
```
dusi: We have previously used this width as the minimum space between the stepper and the input field (otherwise when the text field grows it would almost touch the stepper)...now this is done using `spacing` on the stack view which should be much simpler and easier to reason about.

Kickstarter-iOS/Views/Controllers/PledgeAmountViewControllerTests.swift
```diff
@@ -0,0 +1,139 @@
+@testable import Kickstarter_Framework
+@testable import Library
+@testable import KsApi
+import Prelude
+import UIKit
+
+final class PledgeAmountViewControllerTests: TestCase {
```
dusi: I felt like it was a perfect opportunity to add testing...not only can we test large text but also invalid input values, currency symbol, stepper increment/decrement button etc.

Kickstarter-iOS/Views/Controllers/PledgeViewControllerTests.swift
```diff
@@ -16,4 +18,35 @@ final class PledgeViewControllerTests: TestCase {
...
+        let (parent, _) = traitControllers(device: device, orientation: .portrait, child: controller)
+
+        self.scheduler.run()
+
+        FBSnapshotVerifyView(parent.view, identifier: "lang_\(language)_device_\(device)")
+      }
+    }
+  }
+
+  func testView_WithShippingLocation() {
```
dusi: TODO: Rename to something more meaningful

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -69,11 +74,8 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
     self.currency = project
       .map { currencySymbol(forCountry: $0.country).trimmed() }
 
-    self.stepperMinValue = minAndMax.signal
-      .map(first)
-
-    self.stepperMaxValue = minAndMax.signal
-      .map(second)
+    self.stepperMinValue = minValue.mapConst(0)
+    self.stepperMaxValue = minValue.mapConst(Double.greatestFiniteMagnitude)
```
dusi: I've made a decision to use 0...Double.max but we can chose something else as well (I'm open to discussing this).

The reason we're able to do 0...Double.max is because we want to allow the user to actually decrement/increment bellow/above pledge's min/max value...in that case we simply change the input text to red to indicate this out of bounds state.

Cartfile.resolved
```diff
@@ -4,7 +4,7 @@ github "Alamofire/AlamofireImage" "c41f8b0acfbb3180fe045df73596e4332c338633"
 github "ReactiveCocoa/ReactiveSwift" "6.0.0"
 github "facebook/facebook-objc-sdk" "v5.0.2"
 github "kickstarter/Kickstarter-Prelude" "d03f8831dcec9ad157db4f1c5f769ec17e1a3766"
-github "kickstarter/Kickstarter-ReactiveExtensions" "aa994f0921dca1965ae537fe77c2edc798eef230"
+github "kickstarter/Kickstarter-ReactiveExtensions" "99c3013e22fd6d4ea29d9d4c7a4a7b10bf80a99c"
```
dusi: It's weird that it builds on Circle and not on my machine. I wonder what's the real problem here. 🤔

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -107,25 +109,30 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
...
-      clampedStepperValue
+      minValue,
+      textFieldValue
     )
 
     self.textFieldIsFirstResponder = self.doneButtonTappedProperty.signal
       .mapConst(false)
+
+    self.textFieldTextColor = textColor
+      .wrapInOptional()
```
dusi: Unfortunately it seemed necessary to pass `textColor` as an optional to a `UITextField` (which is how it's also defined in UIKit)

Cartfile.resolved
justinswart: The hash in this file should match the one in the `Cartfile`.

PR #851, opened 2019-09-23
[NT-204] Manage pledge payment method section

Scollaco:
# 📲 What
- Adds the selected payment method to the `ManageViewPledgeViewController`

### Note
- This PR doesn't include Apple Pay information. We don't have designs for that yet and this should be handled in a different PR

# 🤔 Why
- This will allow the user, by visiting the `Manage/View pledge screen`, to see which payment method was used to back a project.

# 🛠 How
- Created a new `ManagePledgePaymentMethodView` class
- Added a new property `paymentSource` to the Backing model
- Reused `GraphUserCreditCard.CreditCard` as `paymentSource`  property type
- Made the `GraphUserCreditCard.CreditCard` conform to `Argo.Decodable`. This was necessary in order to be able to reuse `GraphUserCreditCard.CreditCard` as paymentSource
- Added and fixed tests

# 👀 See

| Default 🐛 | Dynamic type 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-09-23 at 16 11 34 | Simulator Screen Shot - iPhone 8 - 2019-09-23 at 16 11 55 |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

### With Native Checkout Pledge View feature enabled:

- [x] Select a backed project and go Manage/View pledge screen. The Payment Method section should be visible
- [x] The Payment Method section should show the values of the card used to back the project


Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -102,6 +102,8 @@ final class PledgeCreditCardView: UIView {
       |> checkoutCardStackViewStyle
   }
 
+  // MARK: - View model
```
dusi: Can we also add `// MARK: - Configuration`??

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+
+  private lazy var adaptableStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var cardImageView: UIImageView = { UIImageView(frame: .zero) }()
+  private lazy var cardLabelsStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var expirationDateLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var lastFourLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var titleLabel: UILabel = { UILabel(frame: .zero) }()
+
+  private let viewModel = CreditCardCellViewModel()
```
dusi: We've been inconsistent in this lately but shouldn't we define `vm` as CreditCardCellViewModelType` before we initialize it? This way we will be told by the compiler if we try to access any variables directly (skipping `inputs`/`outputs`).

Could you maybe also add refactoring ticket in JIRA to do the same for all the other 17 occurrences of `let viewModel = ` where we let the compiler infer the type without being explicitly told so?

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+
+  // MARK: - Lifecycle
+
+  override init(frame: CGRect) {
+    super.init(frame: frame)
+
+    self.configureViews()
+    self.setupConstraints()
+    self.bindViewModel()
```
dusi: Any update on this @justinswart ? Thought we talked about swizzling some swizzlers to make this automatic :)

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+
+  // MARK: - Configuration
+
+  public func configure(with card: GraphUserCreditCard.CreditCard) {
+    self.viewModel.inputs.configureWith(creditCard: card)
+  }
+
+  private func configureViews() {
+    _ = self
+      |> \.accessibilityElements .~ self.subviews
```
dusi: Should we exclude the image view here?

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // MARK: - View model
+
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.expirationDateLabel.rac.text = self.viewModel.outputs.expirationDateText
+    self.lastFourLabel.rac.text = self.viewModel.outputs.cardNumberTextShortStyle
+    self.viewModel.outputs.cardImage
```
dusi: Can we add empty line here to separate `rac` and `output` signals?

dusi: On a second thought, would it be worth it adding reactive extension on the `UIImageView` to being able to bind `image` property? 🤔

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+          ?|> \.image .~ image
+      }
+  }
+
+  // MARK: - Functions
+
+  private func setupConstraints() {
+    NSLayoutConstraint.activate([
+      self.cardImageView.widthAnchor.constraint(
+        equalToConstant: CheckoutConstants.PaymentSource.ImageView.width
```
dusi: Isn't this being set by the `image`'s intrinsicSize?

If we're setting this explicitly to prevent from a weird UI jump could we also set height or aspect ratio constraint (just to be safe)?

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+      )
+    ])
+  }
+}
+
+// MARK: - Styles
+
+private let adaptableStackViewStyle: StackViewStyle = { stackView in
```
dusi: In this case I would a be a bit more specific and name it `paymentMethodAdaptableStackViewStyle` since it's super easy to confuse reader when scanning the code (made me think you were simply using the global `adaptableStackViewStyle` (without the parameter).

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+  stackView
+    |> \.spacing .~ Styles.grid(2)
+}
+
+private let cardExpirationDateLabelStyle: LabelStyle = { label in
+  label
+    |> checkoutTitleLabelStyle
+    |> \.font .~ UIFont.ksr_caption1().bolded
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.textColor .~ .ksr_text_dark_grey_500
```
dusi: Can we be explicit when defining styles and use full types? `UIColor.` this helps the compiler with the type inference and prevents from potential type-check warnings. This has been something we've mostly seen with styles so this should only be done when dealing with styles. We can be more on ease for other operations.

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+}
+
+private let cardLabelsStackViewStyle: StackViewStyle = { stackView in
```
dusi: Can we simply use `verticalStackViewStyle` without defining a wrapper style?

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -0,0 +1,147 @@
...
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.textColor .~ .ksr_soft_black
+}
+
+private let titleLabelStyle: LabelStyle = { label in
+  label
+    |> \.textColor .~ UIColor.black
+    |> \.font .~ UIFont.ksr_subhead().bolded
+    |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.text %~ { _ in Strings.Payment_method() }
```
dusi: I think we'll also need to use `numberOfLines .~ 0` here to prevent from trimming the title

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -114,7 +109,9 @@ final class ManageViewPledgeViewController: UIViewController {
 
     self.viewModel.outputs.configurePaymentMethodView
       .observeForUI()
-      .observeValues { _ in }
+      .observeValues { [weak self] in
```
dusi: Can we name the closure variable for better readability? To better describe `$0` in this case. Feel like we've been pretty consistent about that everywhere else.

Library/ViewModels/ManageViewPledgeViewModel.swift
```diff
@@ -77,7 +79,7 @@ public final class ManageViewPledgeViewModel:
     self.viewDidLoadObserver.send(value: ())
   }
 
-  public let configurePaymentMethodView: Signal<Project, Never>
+  public let configurePaymentMethodView: Signal<GraphUserCreditCard.CreditCard, Never>
```
dusi: Similar to my comment here https://github.com/kickstarter/ios-oss/pull/835#discussion_r328359310

Would communicate the change with Christella to prevent from merge conflicts

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
Scollaco: That's a good idea. I can create a ticket for that to be done in a different PR.

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
Scollaco: The constraint is being set here so the views are placed correctly in the stackView. Also, the imageView's contentMode is `.scaleAspectFit`, so the aspect ratio should be maintained by default.

Library/ViewModels/ManageViewPledgeViewModel.swift
Scollaco: So was the typealias created yet? I think that creating it in this PR will add a ton of new diffs, though.

PR #852, opened 2019-09-25
[NT-199] Manage menu navigation

ifbarrera:
# 📲 What

Adds view model handling for various menu options.

# 🤔 Why

To handle users selecting a menu option from the `Manage Pledge` flow.

# 🛠 How

Note that this PR is mostly just wiring up the view model to handle the various menu options. It doesn't include actually creating the required VCs or handling navigational changes required for each VC (back button vs close button).

# ✅ Acceptance criteria

- [ ] From the Manage Pledge view, tap the top right "Menu" button. Select "Choose another reward" - this should take you to the rewards carousel.

Please note that updating the left navigation item on the rewards carousel  to a back arrow is not addressed as part of this PR.


Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
```diff
@@ -221,6 +253,30 @@ final class ManageViewPledgeViewController: UIViewController {
...
+
+  // MARK: - Functions
+
+  private func goToRewards(_ project: Project) {
+    let rewardsVC = RewardsCollectionViewController.instantiate(with: project, refTag: nil)
+
+    self.navigationController?.pushViewController(rewardsVC, animated: true)
+  }
+
+  private func goToUpdatePledge(_: Project) {
```
Scollaco: Are we omitting the variable name here because of SwiftFormat?

Library/ViewModels/ManageViewPledgeViewModelTests.swift
```diff
@@ -110,4 +120,64 @@ internal final class ManageViewPledgeViewModelTests: TestCase {
...
+  func testGoToRewards() {
+    self.vm.inputs.configureWith(.template, reward: .template)
+    self.vm.inputs.viewDidLoad()
+
+    self.goToRewards.assertDidNotEmitValue()
+
+    self.vm.inputs.menuButtonTapped()
+    self.vm.inputs.menuOptionSelected(with: .chooseAnotherReward)
+
+    self.goToRewards.assertValues([.template])
```
Scollaco: I think for the `testGoToRewards` and `testGoToUpdatePledge`, would be more readable if the type of `.template` was explicit, like:
```
self.goToRewards.assertValues([Project.template])
```
```
self.goToUpdatePledge.assertValues([Project.template])
```
This would also avoid confusion, since the `configureWith` function is taking 2 different types of `.template`.

Other option would be to simply create variables and use them instead: 
```
let project = Project.template
let reward = Reward.template
```

Kickstarter-iOS/Views/Controllers/ManageViewPledgeViewController.swift
ifbarrera: Hmm good question I'm not sure actually. I'll add it back.

ifbarrera: Ok yeah I think because it's not being used SwiftFormat removes it. This'll be fixed once we consume the `project` argument.

PR #853, opened 2019-09-25
[NT-203] (2/2) Pledge summary section

Scollaco:
# 📲 What

- This is the second part of a 2-part work to build the Pledge summary section of the  Manage Pledge screen.

# 🤔 Why

Because the work has to be done 😎 .
 
# 🛠 How

- Added a new property `locationName` to the Backing model.
- Updated strings
- Added missing tests

# 👀 See
It shouldn't have difference between the work done here and and this [PR](https://github.com/kickstarter/ios-oss/pull/845#issue-318485091).


KsApi/models/Backing.swift
```diff
@@ -9,6 +9,7 @@ public struct Backing {
   public let backerCompleted: Bool?
   public let id: Int
   public let locationId: Int?
+  public let locationName: String?
```
justinswart: Just wondering if we prefer this over just passing the `Reward` to this VC? That already has all of the shipping info and would save us a lot of code.

Scollaco:
> Hey I have the one question and also wondering if we could rename `ManageViewPledgeViewController` -> `ManagePledgeViewController` in this PR?

Yeah, since the PR is small and simple, could be a good opportunity to rename it

KsApi/models/Backing.swift
justinswart: Cool and what about this? 😄

justinswart: We chatted on Slack 👍

Scollaco: From Slack: ` The reward contains shipping rules, but it doesn’t contain the shipping rule you chose`.

PR #856, opened 2019-09-26
[NT-262] Remember this card added to mutation

justinswart:
# 📲 What

- Fixes a bug from #835 in that the delegate was not notified of the initially selected card.
- Adds the `reusable` boolean to `CreatePaymentSourceInput` so that the payment source can be stored for reuse.

# 🤔 Why

- An oversight in the #835 meant that the delegate would not be informed of the initially selected card when the payment methods were reloaded.
- We provide the option during checkout for cards to be single-use or stored for reuse.

# 🛠 How

- Modified the signals so that the delegate would be notified.
- Added the `reusable` boolean to the `CreatePaymentSourceInput` mutation.

# ✅ Acceptance criteria

With native pledge view enabled, Tap on add new card from `PledgeViewController`.

- [ ] Remember this card toggle is shown and _disabled_ by default.
- [ ] Add a card and observe in the console that the mutation's input contains `reusable: false`.
- [ ] Rinse and repeat with the toggle turned on, observe in the console that the mutation's input contains `reusable: true`.


Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -273,12 +284,25 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
...
   public let setStripePublishableKey: Signal<String, Never>
   public let zipcodeTextFieldBecomeFirstResponder: Signal<Void, Never>
 
   public var inputs: AddNewCardViewModelInputs { return self }
   public var outputs: AddNewCardViewModelOutputs { return self }
 }
+
+// MARK: - Functions
+
+private func createPaymentSourceInput(
```
ifbarrera: I've been putting these helper functions into their own files as `CreatePaymentSourceInput+Constructor` - that way they can be tested separately.

justinswart: Ah cool, thanks!

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -123,19 +125,35 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
...
+    .map(CreatePaymentSourceInput.input(fromToken:stripeCardId:reusable:))
+    .switchMap {
+      AppEnvironment.current.apiService.addNewCreditCard(input: $0)
+        .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+        .map { (envelope: CreatePaymentSourceEnvelope) in envelope.createPaymentSource }
+        .materialize()
+    }
+
+    self.newCardAddedWithMessage = addNewCardEvent
+      .map { $0.value?.paymentSource }
```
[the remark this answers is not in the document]
justinswart: Hmm, yeah seems it was that way, I just moved it around a little.

Thanks have updated!

PR #857, opened 2019-09-27
[NT-296] Round and format amount input based on some rules

dusi:
# 📲 What

What the title says

# 🤔 Why

It was noticed that rounding was a bit off on the pledge amount field so this work should fix that.

# 🛠 How

In order to provide great user experience we do have couple rules that we not only round the value but also update the input field

* If user enters number that ends with decimal separator we remove the separator (i.e `10.0` => `10)
* If user enters number that ends with trailing zeros we remove zeros and decimal separator
* If user enters number that has more than 2 decimal numbers we round it down and trim it to 2

This is so that users don't have to manually delete trailing zeros to change the value itself. As soon as there's a legit decimal number we keep it (since that one was most likely intended)

# ✅ Acceptance criteria

**Live user input**
- [x] If the value being entered is lower than `min` value the `Done` button on the keyboard is disabled
- [x] If the value being entered is lower than `min` value the text color is changed to red
- [x] If the value being entered is higher than `max` value the `Done` button on the keyboard is disabled
- [x] If the value being entered is higher than `max` value the text color is changed to red

**Dismissing keyboard (either by tapping the `Done` button or tapping anywhere else on the screen) while entering this input results to the following output**
- [x] Input `10` (or similar - test with other values) produces `10`
- [x] Input `10.` (or similar - test with other values) produces `10`
- [x] Input `10.0` (or similar - test with other values) produces `10`
- [x] Input `10.00` (or similar - test with other values) produces `10`
- [x] Input `10.000` (or similar - test with other values) produces `10`
- [x] Input `10.1` (or similar - test with other values) produces `10.10`
- [x] Input `10.01` (or similar - test with other values) produces `10.01`
- [x] Input `10.99` (or similar - test with other values) produces `10.99`
- [x] Input `10.999` (or similar - test with other values) produces `11`
- [x] Input `10.129` (or similar - test with other values) produces `10.13`
- [x] Input `10.123456789` (or similar - test with other values) produces `10.12`

**Stepper**
- [x] At any time for the sample input in the table above, tapping stepper's `+` or `-` button increments/decrements the value properly


Kickstarter-iOS/Views/Controllers/PledgeAmountViewControllerTests.swift
```diff
@@ -77,13 +77,13 @@ final class PledgeAmountViewControllerTests: TestCase {
     }
   }
 
-  func testView_StepperIncrementButtonDisabled_WhenStepperValueSetToMaximum() {
+  func testView_StepperIncrementButtonDisabled_WhenStepperValueSetToMaximumStepperValue() {
```
dusi: Renamed couple tests for better understanding of what min/max are we referring to (since there is stepper min/max value, pledge min/max value, etc.)

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -68,14 +69,25 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
...
+        let numberOfDecimalPlaces = value.truncatingRemainder(dividingBy: 1) == 0 ? 0 : 2
+        return String(format: "%.\(numberOfDecimalPlaces)f", value)
+      }
 
     self.currency = project
       .map { currencySymbol(forCountry: $0.country).trimmed() }
 
     self.stepperMinValue = minValue.mapConst(0)
-    self.stepperMaxValue = minValue.mapConst(Double.greatestFiniteMagnitude)
+    self.stepperMaxValue = minValue.mapConst(1_000_000_000)
```
dusi: I'm not entirely sure what to do about this...we have to set `stepper.maxValue` since the default value is `100` but I'm not sure to what value. Since we allow inputs bellow pledge minimum, we should also allow inputs above pledge maximum in order to change the text color to red when the input is above that value and still allow stepper / text interactions.

Library/ViewModels/PledgeAmountViewModelTests.swift
```diff
@@ -122,19 +122,19 @@ internal final class PledgeAmountViewModelTests: TestCase {
     self.doneButtonIsEnabled.assertValues([true])
 
     self.vm.inputs.stepperValueChanged(2)
-    self.doneButtonIsEnabled.assertValues([true, true])
```
dusi: Mostly `skipRepeats` for these signals...should have done that before, only noticed now.

Library/ViewModels/PledgeAmountViewModel.swift
justinswart: I think this is fine... it's an arbitrarily high number.

Kickstarter-iOS/Views/Controllers/PledgeAmountViewControllerTests.swift
```diff
@@ -137,3 +176,9 @@ final class PledgeAmountViewControllerTests: TestCase {
     }
   }
 }
+
+// MARK: - Functions
+
+private func maximumValue() -> Double {
```
justinswart: Could we have this as a static property on an enum so that we can ensure it's shared between the tests and VM? Something like:

```swift
enum PledgeAmountStepperConstants {
  static let min: Double = 0
  static let max: Double = 1_000_000_000
}
```

PR #859, opened 2019-09-27
[NT-327] Disable pledge button for invalid inputs

dusi:
# 📲 What

If the amount input is invalid (`<` min pledge value OR `>` max pledge value) the `Pledge` button will be disabled.

# 🤔 Why

This should reflect our UI/UX approach for validating data on this screen and further actions users can perform. They should not be able to try pledging if the input is locally invalid since this would only result in failed requests.

# 🛠 How

Small refactor has been done to not only report back amount value, but also it's validity within the `min`/`max` bounds.

In addition to that I've tweaked the way initial payment method is selected (which previously was not reporting back to the view model - and was only set visually in UI). Now when the user fetches payment methods and the first payment method is set (as the default option) the vm reports back which allows us to enable/disable the `Pledge` button.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-09-27 at 11 50 41 AM | Screen Shot 2019-09-27 at 11 47 46 AM |

# ✅ Acceptance criteria

- [ ] Setting amount value to an invalid number (smaller than pledge min value) should disable 
`Pledge` button
- [ ] Setting amount value to an invalid number (larger than pledge max value) should disable `Pledge` button
- [ ] Setting amount value to a valid number (larger than or equal to pledge min value OR lower than or equal to pledge max value) should enable the `Pledge` button (if it was previously disabled)
- [ ] `Pledge` button is enabled initially if there is at least one payment method and that method is indicating of being selected


Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -7,7 +7,7 @@ import UIKit
 protocol PledgeAmountViewControllerDelegate: AnyObject {
   func pledgeAmountViewController(
     _ viewController: PledgeAmountViewController,
-    didUpdateAmount amount: Double
+    didUpdate amount: (value: Double, isValid: Bool)
```
dusi: Using a tuple to wrap both value and its validity now

Library/ViewModels/PledgeCreditCardViewModel.swift
```diff
@@ -65,10 +65,18 @@ public final class PledgeCreditCardViewModel: PledgeCreditCardViewModelInputs,
...
-    self.notifyDelegateOfCardSelected = creditCard
+    let creditCardInitial = creditCard
+      .map { $0.id }
+
+    let creditCardSelected = creditCard
       .takeWhen(self.selectButtonTappedProperty.signal)
       .map { $0.id }
 
+    self.notifyDelegateOfCardSelected = Signal.merge(
+      creditCardInitial,
```
dusi: We now make sure to notify delegate when payment methods get loaded initially and default card gets selected (if there's at least one card).

ifbarrera: This is also fixed in https://github.com/kickstarter/ios-oss/pull/856

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -23,7 +23,7 @@ public protocol PledgeViewModelInputs {
     paymentData: (displayName: String?, network: String?, transactionIdentifier: String)
   )
   func paymentAuthorizationViewControllerDidFinish()
-  func pledgeAmountDidUpdate(to amount: Double)
+  func pledgeAmountDidUpdate(_ amount: (value: Double, isValid: Bool))
```
ifbarrera: Should we also update the parameter name here from `amount` to maybe `value`?

dusi: Yeah, I'd be happy to...any ideas what to replace `value` to?

`pledgeAmountDidUPdate(_ value: (???: Double, isValid: Bool))`

Library/ViewModels/PledgeCreditCardViewModel.swift
dusi: Good to know...I can see it looks slightly different to my approach...I can either way for the other PR to make it in or try reflecting those changes here.

Library/ViewModels/PledgeViewModel.swift
justinswart: I suppose if we introduced a typealias like:
```swift
typealias PledgeAmountData = (amount: Double, isValid: Bool)
```
Then that would allow us to change the delegate to something like:
```swift
func pledgeAmountViewController(
  _ viewController: PledgeAmountViewController,
  didUpdateWith data: PledgeAmountData
)
```
Then it'd also help name our VM input on `PledgeViewModel`:
```swift
```

PR #860, opened 2019-09-30
[NT-226] Cancel backing navigation workflow

ifbarrera:
# 📲 What

Creates `CancelPledgeViewController` along with `CancelPledgeViewModel` and test setup. Configures `ManagePledgeViewController` to push this new VC onto the stack when `Cancel pledge` is selected from the action sheet menu.

# 🤔 Why

Setup for cancel pledge.

# 👀 See

Ou6VVgKR9j

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] With the *Native Checkout* and *Native Pledge Screen* feature flags turned *ON*, navigate to a backed project. Tap the menu button on the top right. Select "Cancel pledge" - you should be taken to a new screen with the title set to "Cancel pledge". The back button should appear on the top left.


Kickstarter-iOS/Views/Controllers/RewardPledgeNavigationController.swift
```diff
@@ -32,8 +32,6 @@ extension RewardPledgeNavigationController: UINavigationControllerDelegate {
     case (.pop, is PledgeViewController, is RewardPledgeTransitionAnimatorDelegate):
       return RewardPledgePopTransitionAnimator()
     default:
-      // We're not performing any custom transition, bring back the standard divider line
-      self.navigationBar.shadowImage = nil
```
ifbarrera: Removed this line in favor of keeping the shadow for all VCs managed by `RewardPledgeNavigationController`. This will cause some funkiness with colors in the deprecated pledge VC and possibly the Thanks page, but I think we can easily fix those during a design pass - ideally we can get to a place where there's visual consistency between all the VCs in the checkout flow.

Library/ViewModels/CancelPledgeViewModel.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import KsApi
+import ReactiveSwift
+
+public protocol CancelPledgeViewModelOutputs {}
+
+public protocol CancelPledgeViewModelInputs {
+  func configure(with project: Project, backing: Backing)
+  func viewDidLoad()
+}
```
cdolm92: The order here is different. `Inputs` should go before the `Outputs`

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -0,0 +1,34 @@
...
+  private let viewModel: CancelPledgeViewModelType = CancelPledgeViewModel()
+
+  // MARK: - Lifecycle
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    _ = self
+      |> \.title %~ { _ in Strings.Cancel_pledge()
+      }
```
dusi: I know this is a styling issue but could we possibly put this on the previous line...It looks like another limitation of `SwiftFormat` 😅

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -0,0 +1,34 @@
...
+
+  // MARK: - Configuration
+
+  internal func configure(with _: Project, backing _: Backing) {}
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = self.view
+      |> \.backgroundColor .~ UIColor.ksr_grey_400
```
dusi: Should we use `checkoutBackgroundStyle` to stay consistent?

Kickstarter-iOS/Views/Controllers/RewardPledgeNavigationController.swift
dusi: Can you describe what funkiness do you have in mind? 😄

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
ifbarrera: It's not the right color 😭 we really need to do a sweep of all our background colors and align on just one if possible.

Kickstarter-iOS/Views/Controllers/RewardPledgeNavigationController.swift
ifbarrera: I believe what might happen is the line won't come back in the `Thanks` VC after checkout. What I think we should do is do a light colors pass on the `Thanks` screen to make it visually consistent with the other screens in the new checkout flow.

PR #861, opened 2019-09-30
[NT-204] Additional fixes

dusi:
# 📲 What

Fixes some of the issues found in #851

# 🤔 Why

Since @Scollaco is away we want to merge his PR addressing some of the last issues found on it.

# ♿️ Accessibility 

- [ ] Works with VoiceOver (the traits are set properly - header for `Payment method`, static text for the other two labels)
- [ ] Supports Dynamic Type (given the short text in those labels, please test on iPhone SE or similar in order to see that labels are supporting multiple lines)

# ✅ Acceptance criteria

- [ ] Addresses outstanding issues found in #851

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
+++ b/Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
@@ -112,17 +112,17 @@ final class ManagePledgePaymentMethodView: UIView {
 
 private let expirationDateLabelStyle: LabelStyle = { label in
   label
-    |> checkoutTitleLabelStyle
-    |> \.font .~ UIFont.ksr_caption1().bolded
     |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.font .~ UIFont.ksr_caption1().bolded
+    |> \.numberOfLines .~ 0
     |> \.textColor .~ UIColor.ksr_text_dark_grey_500
 }
 
 private let lastFourDigitsLabelStyle: LabelStyle = { label in
   label
-    |> checkoutTitleLabelStyle
-    |> \.font .~ UIFont.ksr_subhead().bolded
     |> \.adjustsFontForContentSizeCategory .~ true
+    |> \.font .~ UIFont.ksr_subhead().bolded
+    |> \.numberOfLines .~ 0
     |> \.textColor .~ UIColor.ksr_soft_black
 }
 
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -9,7 +9,7 @@ final class ManagePledgePaymentMethodView: UIView {
   private lazy var cardImageView: UIImageView = { UIImageView(frame: .zero) }()
   private lazy var cardLabelsStackView: UIStackView = { UIStackView(frame: .zero) }()
   private lazy var expirationDateLabel: UILabel = { UILabel(frame: .zero) }()
-  private lazy var lastFourLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var lastFourDigitsLabel: UILabel = { UILabel(frame: .zero) }()
```
dusi: Renamed for better clarity

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -110,19 +110,19 @@ final class ManagePledgePaymentMethodView: UIView {
 
 // MARK: - Styles
 
-private let cardExpirationDateLabelStyle: LabelStyle = { label in
+private let expirationDateLabelStyle: LabelStyle = { label in
```
dusi: `expirationDateLabel` & `lastFourDigitsLabel` styles make it simpler when they don't inherit from `checkoutTitleLabelStyle` because these styles were overriding more properties than they were setting...so we're setting these explicitly here per style.

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -133,9 +133,6 @@ private let paymentMethodAdaptableStackViewStyle: StackViewStyle = { stackView i
 
 private let titleLabelStyle: LabelStyle = { label in
   label
-    |> \.textColor .~ UIColor.black
-    |> \.font .~ UIFont.ksr_subhead().bolded
-    |> \.adjustsFontForContentSizeCategory .~ true
+    |> checkoutTitleLabelStyle
```
dusi: On the other hand `titleLabelStyle` is identical to the one we've been using in other places.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -183,14 +183,17 @@ final class ManagePledgeViewController: UIViewController {
...
-    ]
-
-    _ = (childViews, self.rootStackView)
+    _ = ([self.pledgeSummaryView, self.paymentMethodView], self.rootStackView)
       |> ksr_addArrangedSubviewsToStackView()
+
+    [self.rewardReceivedViewController].forEach { viewController in
+      self.addChild(viewController)
+
+      _ = ([viewController.view], self.rootStackView)
```
dusi: We were previously not adding the VC to the containment hierarchy properly..this should fix it.

PR #862, opened 2019-09-30
[NT-206] Add accessibility label to the toggle

dusi:
# 📲 What

As the title says

# 🤔 Why

The toggle wasn't very accessible since it didn't provide much context to the VoiceOver about what's being toggled ON/OFF

# ♿️ Accessibility 

Navigate to `Manage pledge screen`

- [x] VoiceOver reads the reward received toggle as `Reward received OFF, double tap to toggle 
setting` if the toggle is OFF
- [x] VoiceOver reads the reward received toggle as `Reward received ON, double tap to toggle setting` if the toggle is ON

# ⏰ TODO
- [ ] Localization will be done in a follow up PR


dusi:
Unfortunately I didn't have a real device to build on so please test on a real device.

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewController.swift
```diff
@@ -14,5 +14,8 @@ final class ManageViewPledgeRewardReceivedViewController: ToggleViewController {
 
     _ = self.toggle
       |> checkoutSwitchControlStyle
+      |> \.accessibilityLabel %~ {
+        _ in localizedString(key: "Reward_received", defaultValue: "Reward received")
```
ifbarrera: Are we still waiting for translations or can we use `Strings.Reward_received`? 🤔

dusi: Tried pulling new strings ... still not translated ... I thought we're not going to block PRs because of localizations but I'm happy to hold on for now

dusi:
I believe so..I was under the impression that these are section headers (please see the attachment)

Screen Shot 2019-10-01 at 3 58 55 PM

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewController.swift
ifbarrera: I can't remember what we decided about localizations 😭but I think we should at least wait for the key to be available in the `Strings` file?

dusi:
Waiting for base strings

dusi:
Made https://github.com/kickstarter/kickstarter/pull/17573

Will wait for it to be deployed and I think we should not hold a11y by localization so I'll add a TODO to my follow up PR to address localization separately

dusi:
@ifbarrera I've updated the strings (they're not localized yet)...but I think it should not hold this PR...I've also updated the ticket to include localization of reward received toggle logic in JIRA so that we can keep track of it and not sign off on it before it's localized.

Kickstarter-iOS/Locales/fr.lproj/Localizable.strings
```diff
@@ -406,6 +406,7 @@
 "Reward_Surveys.zero" = "%{reward_survey_count} questionnaires";
 "Reward_delivered" = "Récompense livrée ?";
 "Reward_estimated_for_delivery_in_date" = "<b>Récompense</b> prévue pour %{delivery_date}";
+"Reward_received" = "Récompense reçue";
```
dusi: Lol, actually already translated in 🇫🇷 🤦‍♂

PR #863, opened 2019-09-30
[NT-330] Set initial amount from Backing

justinswart:
# 📲 What

Follow-up to #858, this sets the initial pledge amount to that of the `Backing` if it is set in the `Project`'s `Personalization`.

# ✅ Acceptance criteria

- [x] Navigate to _Pledge_ for a live project, the initial amount should be that of the reward that you selected.
- [x] Back the project and navigate to _Update Pledge_, the initial pledge amount should be the amount that you backed it for.
- [x] The pledge amount that is shown does not include the shipping amount.


Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -207,3 +215,9 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
 private func rounded(_ value: Double) -> Double {
   return round(value * 100) / 100
 }
+
+private func amountFromBacking(_ backing: Backing?) -> Double? {
+  guard let amount = backing?.amount else { return nil }
+
+  return amount - Double(backing?.shippingAmount ?? 0)
```
ifbarrera: I think it's usually better when doing arithmetic using money types to use `Decimal`:

```
let shippingAmount: Double = backing?.shippingAmount ?? 0
let pledgeAmount = Decimal(amount) - Decimal(shippingAmount)

return (pledgeAmount as NSDecimalNumber).doubleValue
```

What do you think?

justinswart: Hmm yeah good point - although we seem to have been a little inconsistent with this 🤦‍♂

https://github.com/kickstarter/ios-oss/blob/master/Library/ViewModels/BackingViewModel.swift#L176

But I'll change that here!

KsApi/models/BackingTests.swift
```diff
@@ -43,4 +44,15 @@ final class BackingTests: XCTestCase {
     XCTAssertEqual(1, backing.value?.sequence)
     XCTAssertEqual(Backing.Status.pledged, backing.value?.status)
   }
+
+  func testPledgeAmount() {
+    let backing = Backing.template
+      |> Backing.lens.reward .~ Reward.postcards
+      |> Backing.lens.rewardId .~ Reward.postcards.id
+      |> Backing.lens.shippingAmount .~ 116
```
dusi: I'd really chose some easier numbers here for better understanding of what's going on in here 😄

PR #864, opened 2019-10-01
[NT-227] Cancel Pledge UI

ifbarrera:
# 📲 What

Builds the UI components of the cancel pledge screen. Note that this PR does not include the text field functionality, that will be added as part of the implementation for the full cancel pledge UX.

# 🤔 Why

For the Cancel Pledge feature.

# 🛠 How

Adds all UI elements as part of a root stack view. Notice that here I've preferred to use `setCustomSpacing(after)` to tweak spacing between certain views, instead of creating sub-stackviews. Open to feedback on that approach.

Note that the following will be addressed in a follow-up PR:

- accessibility label on the cancellation reason text field
- highlight/disabled states of the `redButtonStyle` are TBD

# 👀 See

| iPhone 8 | Dynamic Type | Landscape |
| --- | --- | --- |
| Yw0lzWq6qo | Simulator Screen Shot - iPhone 8 - 2019-10-01 at 13 03 42 | am9FEHVUT5 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Navigate to a backed project with the "Native Checkout" and "Native Checkout Pledge View" feature flags turned on. Tap the menu button on the top right. Select "Cancel Pledge". You should be taken to the cancel pledge screen. Tap "No, go back". You should be taken back to the Manage Pledge screen.


Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -14,20 +33,190 @@ final class CancelPledgeViewController: UIViewController {
...
+      |> UIButton.lens.title(for: .normal) %~ { _ in Strings.No_go_back() }
   }
 
   override func bindViewModel() {
     super.bindViewModel()
+
+    self.viewModel.outputs.cancellationDetailsTextLabelValue
+      .observeForUI()
+      .observeValues { [weak self] amount, projectName in
+        self?.setCancellationDetailsAttributedText(with: amount, projectName: projectName)
```
dusi: Shouldn't this attributed text be returned from the `vm` so that we can simply use reactive binding?

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -14,20 +33,190 @@ final class CancelPledgeViewController: UIViewController {
...
+    attributedString.addAttributes(boldFontAttribute, range: rangeAmount)
+    attributedString.addAttributes(boldFontAttribute, range: rangeProjectName)
+
+    _ = self.cancellationDetailsTextLabel
+      |> \.attributedText .~ attributedString
+  }
+
+  private func contentInsetsFor(traitCollection: UITraitCollection) -> UIEdgeInsets {
+    return traitCollection.preferredContentSizeCategory.isAccessibilityCategory
+      ? .init(top: 0) : .init(top: self.view.frame.size.height / 4)
```
dusi: Doesn't `.init(top: 0)` equal `.zero`?

Also you can simply use the getter of the frame `self.view.frame.height`

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -14,20 +33,190 @@ final class CancelPledgeViewController: UIViewController {
...
+  }
+}
+
+// MARK: - Styles
+
+private let cancellationDisclaimerLabelStyle: LabelStyle = { label in
+  label
+    |> \.font .~ .ksr_caption1()
+    |> \.textColor .~ UIColor.ksr_text_navy_600
+    |> \.lineBreakMode .~ NSLineBreakMode.byWordWrapping
```
dusi: This could be removed since it's a default value

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -14,20 +33,190 @@ final class CancelPledgeViewController: UIViewController {
...
+private let cancellationReasonTextFieldStyle: TextFieldStyle = { textField in
+  textField
+    |> formFieldStyle
+    |> \.backgroundColor .~ UIColor.white
+    |> \.borderStyle .~ UITextField.BorderStyle.roundedRect
+}
+
+private let cancellationDetailsTextLabelStyle: LabelStyle = { label in
+  label
+    |> \.lineBreakMode .~ NSLineBreakMode.byWordWrapping
```
dusi: Same here

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -14,20 +33,190 @@ final class CancelPledgeViewController: UIViewController {
...
+    |> \.textAlignment .~ NSTextAlignment.center
+    |> \.numberOfLines .~ 0
```
dusi: These could technically be reused for `multi-line centered style` ¯\_(ツ)_/¯

Library/Styles/ButtonStyles.swift
```diff
@@ -34,6 +34,14 @@ public let apricotButtonStyle = baseButtonStyle
   <> UIButton.lens.backgroundColor(for: .highlighted) .~ UIColor.ksr_apricot_500.mixDarker(0.12)
   <> UIButton.lens.backgroundColor(for: .disabled) .~ UIColor.ksr_apricot_500.mixLighter(0.36)
 
+// MARK: - Red
+
+public let redButtonStyle = baseButtonStyle
```
dusi: My apologies, I removed this previously since we didn't use it

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -5,6 +5,25 @@ import Prelude
 import UIKit
 
 final class CancelPledgeViewController: UIViewController {
+  // MARK: - Properties
```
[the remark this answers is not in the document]
ifbarrera: Sure.

ifbarrera:
> Couple minor suggestions / questions and than was wondering about UX when a user might have font set to a larger size (maybe on a smaller device) which would result into keyboard overlapping the input field...so the question is - should we remove the top space when the keyboard is on as well (not only when the traits are accessibility+)?
> 
> See bellow
> 
> Screen Shot 2019-10-02 at 3 18 25 PM

I'd hate to have a bunch of custom UI configuration for dynamic sizing + smaller devices etc. Although it overlaps the text field on smaller devices with larger sizes set, the `contentInset.bottom` is still updated to accommodate the presence of the keyboard, which means all a user would have to do is scroll the text field into place. This feels acceptable enough for such a small edge case. What do you think?

dusi:
> I'd hate to have a bunch of custom UI configuration for dynamic sizing + smaller devices etc. Although it overlaps the text field on smaller devices with larger sizes set, the `contentInset.bottom` is still updated to accommodate the presence of the keyboard, which means all a user would have to do is scroll the text field into place. This feels acceptable enough for such a small edge case. What do you think?

I see your point Isabel and I don't have strong feelings so I'd be OK with this approach...one other edge case will be landscape mode where basically the textfield will be completely hidden...I mentioned this because the implementation is in my opinion trivial to achieve.

PR #854, opened 2019-09-25
[NT-330] Update Pledge UI

justinswart:
# 📲 What

Adds `UpdatePledgeViewController`, the UI and wires up of the VM outputs.

**Note:** Upcoming PR's will add the `updateBacking` mutation and the currency conversion below the total shown on [Abstract](https://app.goabstract.com/projects/a94cda80-1e60-11e8-8903-0b9b4e24ea4b/branches/master/collections/09e55cf7-092d-4126-bc77-f41f430d4578?collectionLayerId=3a41848b-988c-44eb-9c55-6ca845c88e77&mode=design).

# 🤔 Why

This is the UI work for what will eventually allow the user to update their pledge.

# 🛠 How

This does duplicate some functionality from `PledgeViewModel` unfortunately and I'm still somewhat in two minds about whether we should rather try to configure `PledgeViewModel` with a state/context that would hide and show the things that we don't want to see on this screen.

@ifbarrera and I chatted about this and leaned towards the new view model just because there's a ton of stuff in `PledgeViewModel` that is not needed here and also most of the views are nested VCs so the view itself is fairly straightforward.

Inviting thoughts on this!

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

With the native pledge view feature flag enabled, navigate to a project you've backed which is still live, tap on `Manage`, select `Update Pledge` from the top-right context menu.

- [ ] The `UpdatePledgeViewController` is displayed.
- [ ] The stepper adjusts the amount.
- [ ] The amount can be adjusted via the text field.
- [ ] The shipping location can be changed and the total reflects this.
- [ ] The shipping location is not shown for rewards that do not have shipping enabled.
- [ ] Confirmation text appears above the `Confirm` button including the project's deadline.

PR #865, opened 2019-10-01
[NT-331] UpdateBacking mutation

justinswart:
# 📲 What

Following on from the work in #854 and #863, this adds the `UpdateBacking` mutation insofar as we anticipate it to look when it's completed.

# 🤔 Why

This will allow a user to update an existing pledge before the project reaches its deadline.

The API team is currently underway with the work to implement this mutation and we're just getting ahead of that slightly by building out most of it in the mean time.

# 🛠 How

- Added `UpdateBacking` mutation, response envelope and tests.
- Added an extension to `Encodable<String, Any>` which should allow us to more easily generate the dictionary needed for the mutation.

# ✅ Acceptance criteria

- [ ] Review tests and approach (until this is implemented on the back-end we can't really test this end-to-end).


KsApi/mutations/inputs/UpdateBackingInput.swift
```diff
@@ -0,0 +1,49 @@
...
+    self.amount = amount
+    self.applePay = applePay
+    self.id = id
+    self.locationId = locationId
+    self.paymentSourceId = paymentSourceId
+    self.rewardId = rewardId
+  }
+
+  public func toInputDictionary() -> [String: Any] {
+    return self.dictionaryRepresentation ?? [:]
```
justinswart: `dictionaryRepresentation` just allows us to generate this dictionary without having to manually construct it and strip out keys for things that are nil.

KsApi/models/Checkout.swift
```diff
@@ -2,22 +2,17 @@ import Foundation
 
 public struct Checkout: Decodable {
```
justinswart: Updated `Checkout` to receive the new properties. Currently it's shared with `CreateBacking` but that mutation is set to change so I figured we could do this for now.

KsApi/mutations/UpdateBackingMutation.swift
```diff
@@ -0,0 +1,25 @@
...
+  public init(input: T) {
+    self.input = input
+  }
+
+  public var description: String {
+    return """
+    mutation updateBacking($input: UpdateBackingInput!) {
+      updateBacking(input: $input) {
+        checkout {
+          state
```
eoji: Hey Justin! I just tested @cj-dimaggio's changes in `createBacking` and I know this is `updateBacking` but since the mutations will now be synchronous, we should be checking that the state is `successful`

justinswart: Cool thanks!

KsApi/mutations/inputs/UpdateBackingInput.swift
```diff
@@ -0,0 +1,49 @@
...
+
+public struct UpdateBackingInput: GraphMutationInput, Encodable {
+  let amount: String?
+  let applePay: ApplePay?
+  let id: String
+  let locationId: String?
+  let paymentSourceId: String?
+  let rewardId: String?
+
+  public struct ApplePay: Encodable {
```
ifbarrera: Since this will be shared in `CreateBackingInput`, can we take it out of the `UpdateBackingInput` scope?

KsApi/models/Checkout.swift
```diff
@@ -2,22 +2,17 @@ import Foundation
...
-
-extension Checkout {
-  enum CodingKeys: String, CodingKey {
-    case state
-  }
 
-  public init(from decoder: Decoder) throws {
-    let values = try decoder.container(keyedBy: CodingKeys.self)
-    self.state = try values.decode(State.self, forKey: .state)
+  public struct Backing: Decodable {
```
ifbarrera: Wondering why not just add these two properties to our existing `Backing` object 🤔 are you thinking that it would be too much work to conform `Backing` to `Swift.Decodable`? I'm just worried about having multiple `Backing` objects in different scopes.

KsApi/mutations/inputs/UpdateBackingInput.swift
ifbarrera: Nice ⭐️

KsApi/models/Checkout.swift
justinswart: We could do that, I guess we'd just have to query for a few things here in `Backing` that we don't really need in this scope?

Also, I think it's ok to have something like `Checkout.Backing` and `Backing`. Being nested within `Checkout` makes it fairly clear that it's not the same thing.

But open to the idea.

KsApi/mutations/inputs/UpdateBackingInput.swift
justinswart: Sure thing 👍

KsApi/models/Checkout.swift
justinswart: Thinking let's not do this right now?

KsApi/GraphMutation.swift
```diff
@@ -11,3 +11,9 @@ protocol GraphMutation: CustomStringConvertible {
 }
 
 public struct GraphMutationEmptyResponseEnvelope: Decodable {}
+
+extension GraphMutationInput where Self: Encodable {
+  public func toInputDictionary() -> [String: Any] {
+    return self.dictionaryRepresentation ?? [:]
+  }
+}
```
justinswart: Adds automatic conformance to `GraphMutationInput` when also confirming to `Encodable`, foresee any issues doing this?

ifbarrera: I think this is fine.

KsApi/models/UpdateBackingEnvelopeTests.swift
```diff
@@ -0,0 +1,31 @@
...
+    }
+    """
+
+    let data = Data(jsonString.utf8)
+
+    do {
+      let envelope = try JSONDecoder().decode(UpdateBackingEnvelope.self, from: data)
+      XCTAssertEqual(envelope.updateBacking.checkout.state, .verifying)
+      XCTAssertEqual(envelope.updateBacking.checkout.backing.requiresAction, false)
+      XCTAssertEqual(envelope.updateBacking.checkout.backing.clientSecret, "super-secret")
```
ifbarrera: Since `clientSecret` can be `null` I'm not sure if you want to maybe add another test to make sure it decodes properly in that scenario.

KsApi/mutations/UpdateBackingMutation.swift
ifbarrera: @eoji @cj-dimaggio do you mean that if there's an error the checkout state will be `.failed`? Would it make more sense to return an error via the `errors` array in that case? 🤔

KsApi/mutations/inputs/ApplePayParamsTests.swift
```diff
@@ -0,0 +1,20 @@
+@testable import KsApi
+import XCTest
+
+final class ApplePayParamsTests: XCTestCase {
+  func test() {
```
ifbarrera: `testApplePayParams`? Or something more descriptive 😅

KsApi/mutations/inputs/UpdateBackingInput.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public struct UpdateBackingInput: GraphMutationInput, Encodable {
+  let amount: String?
+  let applePay: ApplePayParams?
+  let id: String
```
ifbarrera: Can you add a comment here to clarify what this `id` is? Is it the `backing.id`? Also should it be the `graphId` or the regular `id`? Come to think of it, it would be useful to have comments on all the `id`s, though I know we haven't done that for other mutation inputs.

KsApi/mutations/inputs/ApplePayParamsTests.swift
justinswart: Figured seeing as it was one test, but happy to!

KsApi/mutations/inputs/UpdateBackingInput.swift
justinswart: Added documentation to the init

justinswart:
Merging with caveat that some stuff is subject to change with @cj-dimaggio's final implementation on the back-end.

PR #866, opened 2019-10-01
[NT-206] Reward received toggle business logic

dusi:
# 📲 What

Adds the business logic to
* Hide reward received toggle section if either is true (reward = `NoReward`, backing status != `collected`
* Show the initial value properly
* Allow toggling the value by also calling our endpoint

# 🤔 Why

 Now that we have UI & a11y it's time to hook up the business logic

# ✅ Acceptance criteria

**Default value**

- [x] Navigating to manage pledge screen should `HIDE` `reward received toggle` section if the project backing status is `NOT` `collected`
- [x] Navigating to manage pledge screen should `HIDE` `reward received toggle` section if the 
reward is `NoReward`
- [x] Navigating to manage pledge screen should `SHOW` `reward received toggle` section if both the project backing status `IS` `collected` and the reward is `NOT` a `NoReward`

(this could be verified against the backend by going to https://www.kickstarter.com/profile/backings?ref=user_menu and comparing the `Got it!` column for backed projects.

**Toggling value**

- [x] Toggling the value ON updates the value on the backend so that leaving this screen and navigating back again will persist the value properly
- [x] Toggling the value ON updates the value on the backend so that leaving this screen and navigating back again will persist the value properly

(instead of navigating back and forth this could be verified against the backend by going to https://www.kickstarter.com/profile/backings?ref=user_menu


Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -6,6 +6,8 @@ import UIKit
 final class ManagePledgeViewController: UIViewController {
   // MARK: - Properties
 
+  private let viewModel: ManagePledgeViewModelType = ManagePledgeViewModel()
```
dusi: I've moved the `viewModel` to the top of Properties section. This is consistent with the rest of the codebase and makes it easier to glance at the same place where we would in other VC

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -44,15 +46,6 @@ final class ManagePledgeViewController: UIViewController {
       |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  private let viewModel: ManagePledgeViewModelType = ManagePledgeViewModel()
```
dusi: We've mostly used two step configuration
1. Instantiate
2. Configure

So I've gotten rid of this because it was effectively introducing pattern we've been trying to move away from and making the `configureWith` function obsolete.

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -93,6 +96,9 @@ public final class ManagePledgeViewModel:
     self.goToChangePaymentMethod = self.menuOptionSelectedSignal
       .filter { $0 == .changePaymentMethod }
       .ignoreValues()
+
+    self.rewardReceivedViewControllerViewIsHidden = projectAndReward
+      .map { project, reward in reward.isNoReward || project.personalization.backing?.status != .collected }
```
dusi: Hopefully this is the same logic @eoji was trying to explain to me ... please let me know if I'm off here

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -195,4 +199,160 @@ internal final class ManagePledgeViewModelTests: TestCase {
     self.goToUpdatePledgeProject.assertValues([Project.template])
     self.goToUpdatePledgeReward.assertValues([Reward.template])
   }
+
+  func testRewardReceivedViewControllerIsHidden_NoReward_Canceled() {
```
dusi: These tests could potentially only test when the view is `NOT` hidden but since tests are cheap it should be not a problem for testing all the variations here.

Library/ViewModels/ManagePledgeViewModel.swift
eoji: :chefs_kiss: lgtm!

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewController.swift
```diff
@@ -15,4 +47,16 @@ final class ManageViewPledgeRewardReceivedViewController: ToggleViewController {
...
+
+  // MARK: - View model
+
+  override func bindViewModel() {
+    super.bindViewModel()
+
+    self.viewModel.outputs.rewardReceived
+      .observeForUI()
+      .observeValues { [weak self] isOn in
+        self?.toggle.isOn = isOn
```
justinswart: Didn't you add a RAC binding for this? `self.toggle.rac.on`?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -207,14 +207,17 @@ public final class ProjectPamphletViewController: UIViewController {
...
-    let managePledgeViewController = ManagePledgeViewController.instantiate(with: project, reward: reward)
+    let vc = ManagePledgeViewController.instantiate()
+    vc.configureWith(project: project, reward: reward)
+
+    let nc = RewardPledgeNavigationController(rootViewController: vc)
 
-    let nav = RewardPledgeNavigationController(rootViewController: managePledgeViewController)
     if AppEnvironment.current.device.userInterfaceIdiom == .pad {
-      _ = nav
+      _ = nc
```
justinswart: The diff would probably be small with less renaming here 😬

Kickstarter-iOS/Views/Controllers/ManageViewPledgeRewardReceivedViewController.swift
dusi: I didn't end up using it...let me re-open it and use it for this PR! Will require one more review pass

PR #867, opened 2019-10-02
Handling for Unsupported Cards

cdolm92:
# 📲 What

Users now know why they cannot add an unsupported card when adding a new card from the pledge flow.

# 🤔 Why
Users will not add an unsupported card and know to add a different card.

# 🛠 How

When adding a new card from the pledge flow add new card screen is now configured with the project. Also removed card brand validation logic from the `AddNewCardViewController` .

# 👀 See

handling-unsupported-cards

# ✅ Acceptance criteria

- [ ] Navigate to a non-US project location. Back a project using a test `JCB`, `Discovery`,  or` Diners Club` [Stripe Test Cards](https://stripe.com/docs/testing)
- [ ] Error should appear stating the project location

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -265,7 +265,9 @@ internal final class AddNewCardViewController: UIViewController,
       .observeForUI()
       .observeValues { [weak self] cardNumber, projectCountry in
         guard let _self = self else { return }
-        _self.cardBrandIsSupported(projectCountry: projectCountry, cardNumber: cardNumber, supportedCardBrands: _self.unsupportedCardBrands)
+        _self.cardBrandIsSupported(projectCountry: projectCountry,
+                                   cardNumber: cardNumber,
+                                   supportedCardBrands: _self.unsupportedCardBrands)
     }
 
     Keyboard.change
@@ -355,7 +357,8 @@ internal final class AddNewCardViewController: UIViewController,
     }
   }
 
-  private func cardBrandIsSupported(projectCountry: Location, cardNumber: String, supportedCardBrands _: [STPCardBrand]) {
+  private func cardBrandIsSupported(projectCountry: Location,
+                                    cardNumber: String, supportedCardBrands _: [STPCardBrand]) {
         let country = projectCountry.country
         let brand = STPCardValidator.brand(forNumber: cardNumber)
 
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -363,16 +363,16 @@ internal final class AddNewCardViewController: UIViewController,
     projectLocation: Location,
     cardNumber: String, supportedCardBrands _: [STPCardBrand]
   ) {
-    let country = projectLocation.country
-    let brand = STPCardValidator.brand(forNumber: cardNumber)
-
-    if country != "US", self.unsupportedCardBrands.contains(brand) {
-      self.viewModel.inputs.cardBrand(isValid: false)
-    } else if country != "US", self.supportedCardBrands.contains(brand) {
-      self.viewModel.inputs.cardBrand(isValid: true)
-    } else if country == "US", self.allCardBrands.contains(brand) {
-      self.viewModel.inputs.cardBrand(isValid: true)
-    }
+//    let country = projectLocation.country
+//    let brand = STPCardValidator.brand(forNumber: cardNumber)
+//
+//    if country != "US", self.unsupportedCardBrands.contains(brand) {
+//      self.viewModel.inputs.cardBrand(isValid: false)
+//    } else if country != "US", self.supportedCardBrands.contains(brand) {
+//      self.viewModel.inputs.cardBrand(isValid: true)
+//    } else if country == "US", self.allCardBrands.contains(brand) {
+//      self.viewModel.inputs.cardBrand(isValid: true)
+//    }
   }
 
   private func dismissAndPresentMessageBanner(with message: String) {
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -142,7 +142,6 @@ internal final class AddNewCardViewController: UIViewController,
     _ = self.creditCardValidationErrorLabel
       |> settingsDescriptionLabelStyle
       |> \.textColor .~ .ksr_red_400
-      |> \.text %~ { _ in Strings.Unsupported_card_type() }
 
     _ = self.scrollView
       |> \.alwaysBounceVertical .~ true
@@ -172,6 +171,7 @@ internal final class AddNewCardViewController: UIViewController,
 
   override func bindViewModel() {
     super.bindViewModel()
+    self.creditCardValidationErrorLabel.rac.text = self.viewModel.outputs.unsupportedCardError
 
     self.rememberThisCardToggleViewControllerContainer.rac.hidden =
       self.viewModel.outputs.rememberThisCardToggleViewControllerContainerIsHidden
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -265,10 +265,12 @@ internal final class AddNewCardViewController: UIViewController,
       .observeForUI()
       .observeValues { [weak self] cardNumber, projectCountry in
         guard let _self = self else { return }
-        _self.cardBrandIsSupported(projectCountry: projectCountry,
-                                   cardNumber: cardNumber,
-                                   supportedCardBrands: _self.unsupportedCardBrands)
-    }
+        _self.cardBrandIsSupported(
+          projectCountry: projectCountry,
+          cardNumber: cardNumber,
+          supportedCardBrands: _self.unsupportedCardBrands
+        )
+      }
 
     Keyboard.change
       .observeForUI()
@@ -357,18 +359,20 @@ internal final class AddNewCardViewController: UIViewController,
     }
   }
 
-  private func cardBrandIsSupported(projectCountry: Location,
-                                    cardNumber: String, supportedCardBrands _: [STPCardBrand]) {
-        let country = projectCountry.country
-        let brand = STPCardValidator.brand(forNumber: cardNumber)
+  private func cardBrandIsSupported(
+    projectCountry: Location,
+    cardNumber: String, supportedCardBrands _: [STPCardBrand]
+  ) {
+    let country = projectCountry.country
+    let brand = STPCardValidator.brand(forNumber: cardNumber)
 
-        if country != "US" && self.unsupportedCardBrands.contains(brand) {
-          self.viewModel.inputs.cardBrand(isValid: false)
-        } else if country != "US" && self.supportedCardBrands.contains(brand) {
-          self.viewModel.inputs.cardBrand(isValid: true)
-        } else if country == "US" && self.allCardBrands.contains(brand) {
-          self.viewModel.inputs.cardBrand(isValid: true)
-      }
+    if country != "US", self.unsupportedCardBrands.contains(brand) {
+      self.viewModel.inputs.cardBrand(isValid: false)
+    } else if country != "US", self.supportedCardBrands.contains(brand) {
+      self.viewModel.inputs.cardBrand(isValid: true)
+    } else if country == "US", self.allCardBrands.contains(brand) {
+      self.viewModel.inputs.cardBrand(isValid: true)
+    }
   }
 
   private func dismissAndPresentMessageBanner(with message: String) {
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -30,28 +30,6 @@ internal final class AddNewCardViewController: UIViewController,
   @IBOutlet private var stackView: UIStackView!
   @IBOutlet private var zipcodeView: SettingsFormFieldView!
 
-  private let allCardBrands: [STPCardBrand] = [
-    .amex,
-    .dinersClub,
-    .discover,
-    .JCB,
-    .masterCard,
-    .unionPay,
-    .visa
-  ]
-
-  private let supportedCardBrands: [STPCardBrand] = [
-    .amex,
-    .masterCard,
-    .visa
-  ]
-
-  private let unsupportedCardBrands: [STPCardBrand] = [
-    .dinersClub,
-    .discover,
-    .JCB
-  ]
-
   private var saveButtonView: LoadingBarButtonItemView!
 
   internal var messageBannerViewController: MessageBannerViewController?
@@ -261,17 +239,6 @@ internal final class AddNewCardViewController: UIViewController,
         self?.zipcodeView.textField.becomeFirstResponder()
       }
 
-    self.viewModel.outputs.cardNumberAndProjectLocation
-      .observeForUI()
-      .observeValues { [weak self] cardNumber, location in
-        guard let _self = self else { return }
-        _self.cardBrandIsSupported(
-          projectLocation: location,
-          cardNumber: cardNumber,
-          supportedCardBrands: _self.unsupportedCardBrands
-        )
-      }
-
     Keyboard.change
       .observeForUI()
       .observeValues { [weak self] change in
@@ -357,22 +324,6 @@ internal final class AddNewCardViewController: UIViewController,
         self.viewModel.inputs.stripeError(error)
       }
     }
-  }
-
-  private func cardBrandIsSupported(
-    projectLocation: Location,
-    cardNumber: String, supportedCardBrands _: [STPCardBrand]
-  ) {
-//    let country = projectLocation.country
-//    let brand = STPCardValidator.brand(forNumber: cardNumber)
-//
-//    if country != "US", self.unsupportedCardBrands.contains(brand) {
-//      self.viewModel.inputs.cardBrand(isValid: false)
-//    } else if country != "US", self.supportedCardBrands.contains(brand) {
-//      self.viewModel.inputs.cardBrand(isValid: true)
-//    } else if country == "US", self.allCardBrands.contains(brand) {
-//      self.viewModel.inputs.cardBrand(isValid: true)
-//    }
   }
 
   private func dismissAndPresentMessageBanner(with message: String) {
--- a/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
@@ -30,7 +30,7 @@ internal final class AddNewCardViewController: UIViewController,
   @IBOutlet private var stackView: UIStackView!
   @IBOutlet private var zipcodeView: SettingsFormFieldView!
 
-  private let supportedCardBrands: [STPCardBrand] = [
+  private let allCardBrands: [STPCardBrand] = [
     .amex,
     .dinersClub,
     .discover,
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -261,13 +276,7 @@ extension PledgePaymentMethodsViewController: PledgeCreditCardViewDelegate {
...
 extension PledgePaymentMethodsViewController: PledgeAddNewCardViewDelegate {
   func pledgeAddNewCardView(_: PledgeAddNewCardView, didTapAddNewCardWith intent: AddNewCardIntent) {
-    let addNewCardViewController = AddNewCardViewController.instantiate()
-      |> \.delegate .~ self
-    addNewCardViewController.configure(with: intent)
-    let navigationController = UINavigationController.init(rootViewController: addNewCardViewController)
-    let offset = navigationController.navigationBar.bounds.height
-
-    self.presentViewControllerWithSheetOverlay(navigationController, offset: offset)
+    self.viewModel.inputs.addNewCardTapped(with: intent)
```
ifbarrera: Thank you for adding this 🙏

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -2,6 +2,7 @@ import KsApi
 import Prelude
 import ReactiveExtensions
 import ReactiveSwift
+import Stripe
```
ifbarrera: We should avoid coupling 3rd party SDKs with our `Library` framework to limit dependencies. Ideally you would pass the info needed from the VC to the VM so you don't have to include this import. One way to do that would be to extend the `CardDetails` typealias to take a `cardBrand: GraphUserCreditCard.CreditCardType`, and then pass the card type from the `STPPaymentCardTextFieldDelegate` function:

```
// From AddNewCardViewController
  internal func paymentCardTextFieldDidChange(_ textField: STPPaymentCardTextField) {
    self.viewModel.inputs.paymentInfo(isValid: textField.isValid)

    guard let cardnumber = textField.cardNumber else {
      return
    }

    let stpCardBrand = STPCardValidator.brand(forNumber: cardnumber)

    let cardDetails = (cardnumber,
                       textField.expirationMonth,
                       textField.expirationYear,
                       textField.cvc,
                       stpCardBrand.creditCardType)

    self.viewModel.inputs.creditCardChanged(cardDetails: cardDetails)
  }
```

You'll want to somehow map `STPCardBrand` -> `GraphUserCreditCard.CreditCardType`, but you could probably do that in an extension.

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -76,16 +77,32 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
     let zipcode = self.zipcodeProperty.signal.skipNil()
     let zipcodeIsValid: Signal<Bool, Never> = zipcode.map { !$0.isEmpty }
 
+    let project = self.addNewCardIntentAndProjectProperty.signal.skipNil()
+      .map { $0.1 }
```
ifbarrera: Using `.map(first)` would keep this point-free.

ifbarrera: Sorry, this would be `.map(second)`

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -288,8 +300,39 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
...
+    .masterCard,
+    .unionPay,
+    .visa
+  ]
+
+  let unsupportedCardBrands: [STPCardBrand] = [
+    .dinersClub,
+    .discover,
+    .JCB
+  ]
```
ifbarrera: I think there was a small misunderstanding as to the requirements of this ticket. The supported card types should actually be coming from `project.availableCardTypes`, not hard-coded on the client anymore. You should be able to transform `project.availableCardTypes` from type `[String]` to type `[GraphUserCreditCard.CreditCardType]`, and from there just check if the card brand is contained in the array of supported cards. Something like this:

```
private func cardBrandIsSupported(project: Project, cardBrand: GraphUserCreditCard.CreditCardType) -> Bool {
  guard let availableCardTypes = project.availableCardTypes else {
    return true
  }

  let availableCreditCardTypes = availableCardTypes
    .compactMap { GraphUserCreditCard.CreditCardType(rawValue :$0) }

  return availableCreditCardTypes.contains(cardBrand)
}
```

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -42,6 +42,7 @@ public protocol AddNewCardViewModelOutputs {
   var rememberThisCardToggleViewControllerIsOn: Signal<Bool, Never> { get }
   var saveButtonIsEnabled: Signal<Bool, Never> { get }
   var setStripePublishableKey: Signal<String, Never> { get }
+  var unsupportedCardBrandError: Signal<String, Never> { get }
```
ifbarrera: I feel like `unsupportedCardBrandErrorText` would be more descriptive as the name for this output, since it's observed by a text label. What do you think?

Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift
```diff
@@ -429,3 +412,18 @@ private let creditCardTextFieldStyle: PaymentCardTextFieldStyle = { (textField:
     |> \.cursorColor .~ .ksr_green_700
     |> \.placeholderColor .~ .ksr_text_dark_grey_400
 }
+
+extension STPCardBrand {
```
cdolm92: Should this go here? 🤔

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -177,64 +180,52 @@ internal final class AddNewCardViewModelTests: TestCase {
...
     self.vm.inputs.paymentInfo(isValid: true)
-    self.vm.inputs.cardBrand(isValid: true)
     self.vm.inputs.zipcodeChanged(zipcode: "123")
 
     self.saveButtonIsEnabled.assertValues([false, true], "Enabled when form is valid.")
-
-    self.vm.inputs.paymentInfo(isValid: true)
-    self.vm.inputs.cardBrand(isValid: false)
-
-    self.saveButtonIsEnabled.assertValues([false, true, false], "Disabled if card brand is invalid")
```
ifbarrera: These assertions seem important. We need to verify that even if the form is valid (and the save button is enabled), if the user goes back and enters an invalid card brand, the save button needs to become disabled.

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -340,41 +329,43 @@ internal final class AddNewCardViewModelTests: TestCase {
...
-    self.vm.inputs.cardBrand(isValid: false)
-    self.vm.inputs.creditCardChanged(cardDetails: ("123", nil, nil, nil))
+    self.vm.inputs.creditCardChanged(cardDetails: ("123", nil, nil, nil, .generic))
 
     self.creditCardValidationErrorContainerHidden
       .assertValues([true, false], "Unsupported card message shows")
   }
 
   func testUnsupportedCardMessage_hidesWithValidCardBrand_AndExistingCardNumber() {
+    let project = Project.template
```
ifbarrera: We need to test `unsupportedCardMessage` for both the `.pledge` intent and the `.settings` intent, since they have slightly different logic.

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -76,16 +83,32 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
...
       )
       .map { (brandValid, cardNumber) -> Bool in
-        // If card number is insufficiently long, always return "valid card brand" behaviour
         brandValid || cardNumber.count < 2
       }
 
+    self.unsupportedCardBrandErrorText = Signal.combineLatest(cardBrandIsValid, project.skipNil())
+      .map { _, project in
+        Strings.You_cant_use_this_credit_card_to_back_a_project_from_project_country(
+          project_country: project.location.displayableName
```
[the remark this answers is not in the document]
cdolm92: That only seems to return what I think is the country code..I think that's why on Android we're  using `displayableName`. 
Simulator Screen Shot - iPhone XS - 2019-10-03 at 17 45 54
Simulator Screen Shot - iPhone XS - 2019-10-03 at 17 45 20

Library/ViewModels/AddNewCardViewModelTests.swift
cdolm92: Well cardBrand is no longer an input here. But I'll test the assertion with unavailable cards

Library/ViewModels/AddNewCardViewModel.swift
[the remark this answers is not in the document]
ifbarrera: Oh okay! Thanks for checking that.

Library/ViewModels/AddNewCardViewModel.swift
```diff
@@ -288,8 +312,34 @@ public final class AddNewCardViewModel: AddNewCardViewModelType, AddNewCardViewM
...
+    .mastercard,
+    .unionPay,
+    .visa
+  ]
+
+  guard let project = project else {
+    return supportedCardBrands.contains(cardBrand)
+  }
+
+  guard let availableCardTypes = project.availableCardTypes else {
```
ifbarrera: I think we should group this logic together with line 333:

```
guard let availableCardTypes = project?.availableCardTypes else {
  return supportedCardBrands.contains(cardBrand)
}
```

That will cover both the `project` being `nil` and `availableCardTypes` being `nil`.

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -177,64 +177,61 @@ internal final class AddNewCardViewModelTests: TestCase {
...
-    self.vm.inputs.cardBrand(isValid: false)
     self.saveButtonIsEnabled.assertDidNotEmitValue()
 
     self.vm.inputs.zipcodeChanged(zipcode: "")
-    self.saveButtonIsEnabled.assertValues([false], "Disabled form is incomplete")
 
     self.vm.inputs.cardholderNameChanged("Native Squad")
-    self.vm.inputs.paymentInfo(isValid: true)
-    self.vm.inputs.cardBrand(isValid: true)
+    self.vm.inputs.creditCardChanged(cardDetails: ("6200 0000 0000 0005", 11, 99, "123", .unionPay))
```
ifbarrera: I noticed there's a couple of places in the tests where we call `paymentInfo(isValid)` after calling `creditCardChanged(cardDetails)`. We should try to keep our tests as close to what happens irl as we can - it looks like in the `STPPaymentCardTextFieldDelegate` we always call `creditCardChanges(cardDetails)` _after_ calling `paymentInfo(isValid:)`.

image

So this should be:

```
    self.vm.inputs.zipcodeChanged(zipcode: "")

    self.vm.inputs.cardholderNameChanged("Native Squad")
    self.vm.inputs.paymentInfo(isValid: false) // input called before .creditCardChanged(cardDetails:)
    self.vm.inputs.creditCardChanged(cardDetails: ("6200 0000 0000 0005", 11, 99, "123", .unionPay))
    self.saveButtonIsEnabled.assertValues([false], "Disabled form is incomplete")

    self.vm.inputs.zipcodeChanged(zipcode: "123") // missing field is entered
    self.vm.inputs.paymentInfo(isValid: true)
    self.vm.inputs.creditCardChanged(cardDetails: ("6200 0000 0000 0005", 11, 99, "123", .unionPay))

    self.saveButtonIsEnabled.assertValues([false, true], "Enabled when form is valid.")
```

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -177,64 +177,61 @@ internal final class AddNewCardViewModelTests: TestCase {
...
+    self.saveButtonIsEnabled.assertValues([false], "Disabled form is incomplete")
+    self.vm.inputs.paymentInfo(isValid: false)
     self.vm.inputs.zipcodeChanged(zipcode: "123")
 
-    self.saveButtonIsEnabled.assertValues([false, true], "Enabled when form is valid.")
+    self.saveButtonIsEnabled.assertValues([false], "Enabled when form is valid.")
 
+    self.vm.inputs.cardholderNameChanged("Native Squad")
+    self.vm.inputs.creditCardChanged(cardDetails: ("4242 4242 4242 4242", 11, 99, "123", .visa))
     self.vm.inputs.paymentInfo(isValid: true)
```
ifbarrera: Same here this should go before line 230.

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -177,64 +177,61 @@ internal final class AddNewCardViewModelTests: TestCase {
...
     self.vm.inputs.cardholderNameChanged("Native Squad")
-    self.vm.inputs.paymentInfo(isValid: true)
-    self.vm.inputs.cardBrand(isValid: true)
+    self.vm.inputs.creditCardChanged(cardDetails: ("6200 0000 0000 0005", 11, 99, "123", .unionPay))
+    self.saveButtonIsEnabled.assertValues([false], "Disabled form is incomplete")
+    self.vm.inputs.paymentInfo(isValid: false)
     self.vm.inputs.zipcodeChanged(zipcode: "123")
 
-    self.saveButtonIsEnabled.assertValues([false, true], "Enabled when form is valid.")
+    self.saveButtonIsEnabled.assertValues([false], "Enabled when form is valid.")
```
ifbarrera: This assertion doesn't make sense with what the message is 🤔

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -247,10 +244,11 @@ internal final class AddNewCardViewModelTests: TestCase {
 
   func testDismissKeyboard() {
     self.vm.inputs.viewDidLoad()
+    self.vm.inputs.configure(with: .settings, project: nil)
 
     self.vm.inputs.cardholderNameChanged("Native Squad")
+    self.vm.inputs.creditCardChanged(cardDetails: ("4242 4242 4242 4242", 11, 99, "123", .visa))
     self.vm.inputs.paymentInfo(isValid: true)
```
ifbarrera: Here too can we swap the order of calling these inputs?

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -340,41 +335,43 @@ internal final class AddNewCardViewModelTests: TestCase {
 
   func testUnsupportedCardMessage_showsWithInvalidCardBrand_AndExistingCardNumber() {
     self.vm.inputs.viewDidLoad()
+    self.vm.inputs.configure(with: .pledge, project: Project.template)
```
ifbarrera: Can you clarify what `.intent` is being tested here? It seems like this should be configured with `project: nil`. We should also update the test name to describe what intent is being tested.

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -340,41 +335,43 @@ internal final class AddNewCardViewModelTests: TestCase {
...
 
     self.creditCardValidationErrorContainerHidden
       .assertValues([true], "Unsupported card message is hidden on viewDidLoad")
 
-    self.vm.inputs.cardBrand(isValid: false)
-    self.vm.inputs.creditCardChanged(cardDetails: ("123", nil, nil, nil))
+    self.vm.inputs.creditCardChanged(cardDetails: ("123", nil, nil, nil, .generic))
 
     self.creditCardValidationErrorContainerHidden
```
ifbarrera: Can we also add an assertion for the `unsupportedCardBrandErrorText` output in this test to check that the outputted text is correct for the given intent?

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -340,41 +335,43 @@ internal final class AddNewCardViewModelTests: TestCase {
...
       .assertValues([true, false], "Unsupported card message shows")
   }
 
   func testUnsupportedCardMessage_hidesWithValidCardBrand_AndExistingCardNumber() {
```
ifbarrera: Can you update the test name to describe what `.intent` is being tested here?

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -340,41 +335,43 @@ internal final class AddNewCardViewModelTests: TestCase {
...
   func testUnsupportedCardMessage_hidesWithEmptyOrInvalidCardNumber() {
     self.vm.inputs.viewDidLoad()
+    self.vm.inputs.configure(with: .settings, project: nil)
 
     self.creditCardValidationErrorContainerHidden
       .assertValues([true], "Unsupported card message is hidden on viewDidLoad")
 
-    self.vm.inputs.cardBrand(isValid: false)
-    self.vm.inputs.creditCardChanged(cardDetails: ("", nil, nil, nil))
+    self.vm.inputs.creditCardChanged(cardDetails: ("", nil, nil, nil, nil))
```
ifbarrera: A better test here would be to call this input with `cardDetails: ("1", nil, nil, nil, .generic)` since we're trying to verify that even is a card brand is invalid (as `.generic` is), the error message should hide if the card number character count is < 2.

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -405,9 +402,53 @@ internal final class AddNewCardViewModelTests: TestCase {
...
     self.rememberThisCardToggleViewControllerIsOn.assertDidNotEmitValue()
 
-    self.vm.inputs.configure(with: .settings)
+    self.vm.inputs.configure(with: .settings, project: nil)
     self.vm.inputs.viewDidLoad()
 
     self.rememberThisCardToggleViewControllerIsOn.assertValues([true])
   }
+
+  func testUnsupportedCardBrandsError() {
```
ifbarrera: Can we update this test name to clarify which `.intent` is being tested?

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -405,9 +402,53 @@ internal final class AddNewCardViewModelTests: TestCase {
...
+    self.vm.inputs.viewDidLoad()
+    self.vm.inputs.configure(with: .pledge, project: project)
+
+    self.creditCardValidationErrorContainerHidden
+      .assertValues([true], "Unsupported card message is hidden on viewDidLoad")
+
+    self.vm.inputs.creditCardChanged(cardDetails: ("3566", nil, nil, nil, .jcb))
+
+    self.creditCardValidationErrorContainerHidden
+      .assertValues([true, false], "Unsupported card message hides with a valid card brand")
```
ifbarrera: I think this message is incorrect. Shouldn't this be "Unsupported card message shows with invalid card brand" 🤔

Library/ViewModels/AddNewCardViewModelTests.swift
```diff
@@ -405,9 +402,53 @@ internal final class AddNewCardViewModelTests: TestCase {
...
+    self.unsupportedCardBrandErrorText
+      .assertValues(
+        ["You can’t use this credit card to back a project from Hastings, UK."],
+        "Card is unsupported"
+      )
+  }
+
+  func testUnsupportedCardBrandsError_fromSettings() {
```
ifbarrera: I think this test might be redundant since we already have the following tests for this intent:

`func testUnsupportedCardMessage_HiddenOnViewDidLoad()` (line 329)
`func testUnsupportedCardMessage_showsWithInvalidCardBrand_AndExistingCardNumber()` (line 336)
`func testUnsupportedCardMessage_hidesWithValidCardBrand_AndExistingCardNumber()` (line 349)
`func testUnsupportedCardMessage_hidesWithEmptyOrInvalidCardNumber()` (line 364)

The test names just need to be updated to reflect the intent that is being tested (`.settings`), and you should be able to remove this test.

cdolm92: That test is testing the `cardBrandErrorText` when an invalid card is entered from the `.settings` intent. `func testUnsupportedCardMessage_HiddenOnViewDidLoad()` (line 329)

Library/ViewModels/AddNewCardViewModelTests.swift
cdolm92: Not sure what your feedback is here. .pledge intent is being tested here and that needs to be configured with a `project`

Library/ViewModels/AddNewCardViewModelTests.swift
cdolm92: Might delete this test as there already is one for it ` func testUnsupportedCardBrandsError_withPledgeIntent()` on `line 439`

PR #873, opened 2019-10-03
[NT-329] Show alert when using invalid pledge amount with Apple Pay

dusi:
# 📲 What

If a user enters incorrect pledge amount (bellow `min` or above `max`) and taps the `Apple Pay` button we now show an alert with more instructions.

# 🤔 Why

Since Apple is not fond of disabling `Apple Pay` button we have to have a way of local validation of the input when user goes through `Apple Pay` path.

# 🛠 How

Simply checking the input & its min, max values when evaluating the `Apple Pay` button tap logic.

Please note that this PR is mostly tests since we're now passing more values in the `amount` tuple so our tests suite had to be updated to account for those.

# 👀 See

Screen Shot 2019-10-03 at 3 50 07 PM

# ✅ Acceptance criteria

On device that supports `Apple Pay` (has `Apple Pay` setup in the `Wallet.app`) go to the new pledge screen and

- [x] Enter pledge amount value that is `smaller` than the `min` value (initial value) > this should result into alert being shown when tapping the `Apple Pay` button
- [x] Enter pledge amount value that is `larger` than the `max` value (usually 10k but depends on the project) > this should result into alert being shown when tapping the `Apple Pay` button
- [x] Enter pledge amount value that is between `min` and `max` value > this should result into the native Apple Pay sheet being shown

# ⏰ TODO

- [ ] Update to use generated strings

### DIFF
--- a/Library/ViewModels/PledgeAmountViewModel.swift
+++ b/Library/ViewModels/PledgeAmountViewModel.swift
@@ -18,7 +18,7 @@ public protocol PledgeAmountViewModelInputs {
 }
 
 public protocol PledgeAmountViewModelOutputs {
-  var amount: Signal<(Double, Bool), Never> { get }
+  var amount: Signal<(Double, Double, Double, Bool), Never> { get }
   var currency: Signal<String, Never> { get }
   var doneButtonIsEnabled: Signal<Bool, Never> { get }
   var generateSelectionFeedback: Signal<Void, Never> { get }
@@ -133,11 +133,11 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
 
     self.amount = updatedValue
       .map { min, max, value in
-        (rounded(value), min <= value && value <= max)
+        (rounded(value), min, max, min <= value && value <= max)
       }
 
     let isValueValid = self.amount
-      .map(second)
+      .map { $0.3 }
       .skipRepeats()
 
     self.doneButtonIsEnabled = isValueValid
@@ -167,7 +167,7 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
 
     self.stepperValue = Signal.merge(
       minValue,
-      self.amount.map(first)
+      self.amount.map { $0.0 }
     )
     .skipRepeats()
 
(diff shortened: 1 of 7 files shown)


Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -18,7 +18,7 @@ public protocol PledgeAmountViewModelInputs {
 }
 
 public protocol PledgeAmountViewModelOutputs {
-  var amount: Signal<(Double, Bool), Never> { get }
+  var amount: Signal<(Double, Double, Double, Bool), Never> { get }
```
justinswart: Could we just change the type of this output to `Signal<PledgeAmountData, Never>`?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -219,8 +223,37 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         PKPaymentAuthorizationViewController.merchantIdentifier
       ) as PaymentAuthorizationData }
 
-    self.goToApplePayPaymentAuthorization = paymentAuthorizationData
+    let goToApplePayPaymentAuthorization = pledgeAmountIsValid
+      .takeWhen(self.applePayButtonTappedSignal)
+      .filter { isTrue($0) }
```
justinswart: Can be `filter(isTrue)`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -219,8 +223,37 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+
+    let showApplePayAlert = pledgeAmountIsValid
       .takeWhen(self.applePayButtonTappedSignal)
+      .filter { isFalse($0) }
```
justinswart: Can be `filter(isFalse)`.

justinswart: 😄 and dis?

dusi: gif

PR #874, opened 2019-10-04
[NT-331] Perform Update Backing Mutation Request

justinswart:
# 📲 What

- Wires up the confirm button to make an `UpdateBacking` request when the pledge amount or shipping location has changed.

# ✋ What Not

- Does not entirely address loading state of this view.
  - The confirm button will be disabled while the request is in-flight but fields will still be tappable. This should be addressed separately along with the loading state of the button.
- Does not confirm that the mutation fully works. Work on this is still underway by @cj-dimaggio.
- Does not address updating of payment method or Apple Pay.

# 🤔 Why

Brings the functionality to allow the user to update their pledge amount or shipping location.

# 🛠 How

- Added validation to confirm that a value has in fact changed for the confirm button to be enabled.
- Added `UpdateBackingData` typealias to gather the information that will be needed to update the backing.

# 👀 See

| Enabling on values changed | Success (faked) | Error |
| --- | --- | --- |
|  |  |  |

# ✅ Acceptance criteria

Navigate Update Pledge on a project that you have backed that is still live.

- [ ] Changing the pledge amount or shipping location should enable the confirm button. Changing them back to their original values should disable the pledge button.
- [ ] Tapping confirm on the enabled button should perform the request (this will currently fail).

# ⏰ TODO

- [ ] Possible changes once back-end work is completed.


KsApi/models/Backing.swift
```diff
@@ -79,3 +79,9 @@ extension Backing {
     return (pledgeAmount as NSDecimalNumber).doubleValue
   }
 }
+
+extension Backing: GraphIDBridging {
+  public static var modelName: String {
+    return "Backing"
```
justinswart: Need to confirm this name.

KsApi/models/Location.swift
```diff
@@ -31,3 +31,9 @@ extension Location: Equatable {}
 public func == (lhs: Location, rhs: Location) -> Bool {
   return lhs.id == rhs.id
 }
+
+extension Location: GraphIDBridging {
+  public static var modelName: String {
+    return "Location"
```
justinswart: Need to confirm this name.

Library/UpdateBackingInput+Constructor.swift
```diff
@@ -0,0 +1,38 @@
...
+      applePay: nil,
+      id: backingId,
+      locationId: locationId,
+      paymentSourceId: nil,
+      rewardId: rewardId
+    )
+  }
+}
+
+func pledgeAmountString(withAmount amount: Double, shippingRule: ShippingRule?) -> String {
```
justinswart: Will move this to a shared place once we implement the new `CreateBackingInput`. Also not sure if we're going the route of splitting out amount and shipping rule for that.

KsApi/models/Location.swift
ifbarrera: I think the `locationId` mutation input is actually not a `graphID`. In other words you should just be able to use `location.id`.

Library/UpdateBackingInput+Constructor.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension UpdateBackingInput {
+  internal static func input(
+    from updateBackingData: UpdateBackingData
+  ) -> UpdateBackingInput {
+    let pledgeAmount: String? = updateBackingData.pledgeAmount.flatMap { pledgeAmount in
+      pledgeAmountString(withAmount: pledgeAmount, shippingRule: updateBackingData.shippingRule)
+    }
+    let backingId: String = updateBackingData.backing.graphID
+    let locationId: String? = updateBackingData.shippingRule.flatMap { $0.location.graphID }
```
ifbarrera: Like I mentioned I think this is actually just the `location.id`

Library/UpdateBackingInput+Constructor.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+import KsApi
+
+extension UpdateBackingInput {
+  internal static func input(
+    from updateBackingData: UpdateBackingData
+  ) -> UpdateBackingInput {
+    let pledgeAmount: String? = updateBackingData.pledgeAmount.flatMap { pledgeAmount in
```
ifbarrera: Is pledge amount not required? 🤔

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -99,6 +101,10 @@ public final class ManagePledgeViewModel:
 
     self.rewardReceivedViewControllerViewIsHidden = projectAndReward
       .map { project, reward in reward.isNoReward || project.personalization.backing?.status != .collected }
+
+    self.showSuccessBannerWithMessage = self.pledgeViewControllerDidUpdatePledgeSignal.mapConst(
```
ifbarrera: Is the idea behind putting this string in `ManagePledgeViewModel` instead of `PledgeViewModel` (where the update pledge logic happens) that we're anticipating reusing this string for other "update" types? (ie. update payment method, update reward).

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -21,13 +21,26 @@ public enum PledgeViewContext {
...
 
 public typealias StripeConfigurationData = (merchantIdentifier: String, publishableKey: String)
 public typealias CreateBackingData = (
   project: Project, reward: Reward, pledgeAmount: Double,
   selectedShippingRule: ShippingRule?, refTag: RefTag?
 )
+public typealias UpdateBackingData = (
+  backing: Backing,
+  reward: Reward,
+  pledgeAmount: Double?,
```
ifbarrera: How come this is optional? 🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +361,70 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     let createBackingTransactionSuccess = project.takeWhen(createBackingEventSuccess)
 
     self.goToThanks = Signal.merge(applePayTransactionCompleted, createBackingTransactionSuccess)
+
+    let amountChanged = Signal.combineLatest(
+      project.map { $0.personalization.backing?.pledgeAmount },
+      self.pledgeAmountDataSignal.map { $0.0 }
```
ifbarrera: `self.pledgeAmountDataSignal.map(first)`? Might also be valuable to unpack the backing seperately, for clarity:

```
let backing = project.map { $0.personalization.backing }.skipNil()

let pledgeAmount = Signal.combineLatest(
  backing.map { $0.pledgeAmount }, 
  self.pledgeAmountDataSignal.map(first))
```

Looks like `backing` would also come in useful below in a couple places.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +361,70 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    )
+    .map(!=)
+
+    let shippingRuleChanged = Signal.merge(
+      self.viewDidLoadProperty.signal.mapConst(false),
+      Signal.combineLatest(
+        project.map { $0.personalization.backing?.locationId },
```
ifbarrera: Here you could use `backing.map { $0.locationId }` can't wait for key paths to be supported in `.map` 😅

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +361,70 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    .map { amountChanged, shippingRuleChanged, _ -> Bool in
+      [amountChanged, shippingRuleChanged].contains(true)
+    }
+
+    self.title = context.map { $0.title }
+
+    // MARK: Update Backing
+
+    let updateBackingData = Signal.combineLatest(
+      project.map { $0.personalization.backing }.skipNil(),
```
ifbarrera: Could use `backing` here too if you go that route.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +361,70 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+        project.map { $0.personalization.backing?.locationId },
+        self.shippingRuleSelectedSignal.map { $0.location.id }
+      )
+      .map(!=)
+    )
+
+    let valuesChanged = Signal.combineLatest(
+      amountChanged,
+      shippingRuleChanged,
+      self.viewDidLoadProperty.signal
```
ifbarrera: I _think_ you don't need to combine with `viewDidLoad` because `amountChanged` and `shippingRuleChanged` both already depend on `viewDidLoad` 🤔. You could easily test this by removing `viewDidLoad` from here - your tests should still pass.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1071,4 +1092,173 @@ final class PledgeViewModelTests: TestCase {
...
+  func testUpdateBacking_Success() {
+    let reward = Reward.postcards
+
+    let project = Project.cosmicSurgery
+      |> Project.lens.state .~ .live
+      |> Project.lens.personalization.isBacking .~ true
+      |> Project.lens.personalization.backing .~ (
+        .template
+          |> Backing.lens.paymentSource .~ GraphUserCreditCard.amex
+          |> Backing.lens.status .~ .errored
```
ifbarrera: Any reason this status is `.errored`? Maybe `.pledged` would be more consistent with what we'd be likely to see during this flow?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1071,4 +1092,173 @@ final class PledgeViewModelTests: TestCase {
...
+    withEnvironment(apiService: mockService, currentUser: .template) {
+      self.vm.inputs.configureWith(project: project, reward: reward, refTag: nil, context: .update)
+      self.vm.inputs.viewDidLoad()
+
+      self.notifyDelegateUpdatePledgeDidSucceed.assertDidNotEmitValue()
+      self.updatePledgeFailedWithError.assertDidNotEmitValue()
+      self.confirmButtonEnabled.assertDidNotEmitValue()
+
+      self.vm.inputs.pledgeAmountViewControllerDidUpdate(
+        with: (amount: 10.0, min: 25.0, max: 10_000.0, isValid: true)
```
ifbarrera: This is a little confusing because the min is `25`, the amount is `10` but it says `isValid: true` 🤔

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1071,4 +1092,173 @@ final class PledgeViewModelTests: TestCase {
...
+
+    self.confirmButtonHidden.assertValues([false])
+    self.confirmButtonEnabled.assertDidNotEmitValue()
+
+    self.vm.inputs.pledgeAmountViewControllerDidUpdate(
+      with: (amount: 690, min: 25.0, max: 10_000.0, isValid: true)
+    )
+
+    self.confirmButtonHidden.assertValues([false])
+    self.confirmButtonEnabled.assertValues([false])
```
ifbarrera: Is it possible to add an assertion message here? Took me a while to figure out why the confirm button isn't enabled in this case 😅 (it's because the `amount` is actually just the same as was already pledged, right?).

Library/UpdateBackingInput+Constructor.swift
justinswart: No because we might not be updating it.

Library/ViewModels/ManagePledgeViewModel.swift
justinswart: No, just that this is a side-effect from the delegate being notified about the success of the request so that the banner is visible when the VC is popped off the stack.

Library/ViewModels/PledgeViewModel.swift
justinswart: Because it might not be changed.

Library/ViewModels/PledgeViewModel.swift
justinswart: Can't do `self.pledgeAmountDataSignal.map(first)` because the tuple is 4 values and our versions of `first`, `second`, `third` only support up to 3.

Library/ViewModels/PledgeViewModelTests.swift
justinswart: I probably copy-pasted this from somewhere 😄

Library/ViewModels/PledgeViewModelTests.swift
justinswart: Sure will do :)

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +363,69 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      )
+      .map(!=)
+    )
+
+    let valuesChanged = Signal.combineLatest(
+      amountChanged,
+      shippingRuleChanged
+    )
+    .map { amountChanged, shippingRuleChanged -> Bool in
+      [amountChanged, shippingRuleChanged].contains(true)
```
dusi: Good old `contains` 😄 👍

Library/ViewModels/PledgeViewModel.swift
ifbarrera: Alas

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -156,6 +160,14 @@ final class ManagePledgeViewController: UIViewController {
       .observeValues { [weak self] project, backing in
         self?.goToCancelPledge(project: project, backing: backing)
       }
+
+    self.viewModel.outputs.showSuccessBannerWithMessage
+      .observeForUI()
```
dusi: Sometimes we use `.observeForUI` other times we use `.observeForControllerAction()` is there any difference?

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -277,11 +299,21 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
...
 
     self.confirmationLabel.rac.attributedText = self.viewModel.outputs.confirmationLabelAttributedText
 
+    self.viewModel.outputs.title
+      .observeForUI()
+      .observeValues { [weak self] title in
+        guard let self = self else { return }
+
+        _ = self
+          |> \.title %~ { _ in title }
```
dusi: Wish we had a RAC extension on `UIViewController` for dis ¯\_(ツ)_/¯

Library/UpdateBackingInput+Constructor.swift
```diff
@@ -0,0 +1,38 @@
...
+      applePay: nil,
+      id: backingId,
+      locationId: locationId,
+      paymentSourceId: nil,
+      rewardId: rewardId
+    )
+  }
+}
+
+func pledgeAmountString(withAmount amount: Double, shippingRule: ShippingRule?) -> String {
```
dusi: Should we either make this `private` and/or `test` it? 🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -65,11 +80,15 @@ public protocol PledgeViewModelOutputs {
   var descriptionViewHidden: Signal<Bool, Never> { get }
   var goToApplePayPaymentAuthorization: Signal<PaymentAuthorizationData, Never> { get }
   var goToThanks: Signal<Project, Never> { get }
+  var notifyDelegateUpdatePledgeDidSucceedWithMessage: Signal<String, Never> { get }
   var paymentMethodsViewHidden: Signal<Bool, Never> { get }
+  var popViewController: Signal<(), Never> { get }
```
ifbarrera: So something I noticed while testing is it seems like the `popViewController` and `notifyDelegateUpdatePledgeDidSucceedWithMessage` outputs are firing at the same time, which means that the pop animation and the message banner animation overlap. I think it would be _nice_ to only show the message banner once the pop animation has completed. 🤔 Maybe setting a delay on the `notifyDelegateUpdatePledgeDidSucceedWithMessage` signal could help? I'm not sure if it's worth the trouble but just a thought.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
justinswart: My understanding on the difference between these two is that there was some bug in the ReactiveSwift scheduler before that required this distinction. I should actually did in to see whether that was since fixed.

For consistency though I'll update this.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: That would be nice.

Library/UpdateBackingInput+Constructor.swift
justinswart: Going to make it private for now and consider it tested by way of this constructor being tested.

This will become a shared function when we add the newer `CreateBacking` mutation and then I'll break it out and write tests for it specifically.

Library/ViewModels/PledgeViewModel.swift
ifbarrera: Lol might I suggest `amountChanged || shippingRuleChanged` 😅nothing against `contains` I just think an `OR` seems to illustrate the logic more clearly.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -338,6 +363,70 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    self.notifyDelegateUpdatePledgeDidSucceedWithMessage = updateBackingEvent.values()
+      .mapConst(Strings.Got_it_your_changes_have_been_saved())
+    self.updatePledgeFailedWithError = updateBackingEvent.errors()
+      .map { $0.localizedDescription }
+
+    self.popViewController = self.notifyDelegateUpdatePledgeDidSucceedWithMessage.ignoreValues()
+
+    let valuesChangedAndValid = Signal.combineLatest(
+      valuesChanged,
+      self.pledgeAmountDataSignal.map { $0.3 }
```
ifbarrera: Can we use the parameter name here instead? I think it's `$0.isValid`?

Library/ViewModels/PledgeViewModel.swift
justinswart: I've added this to the JIRA ticket in which we'll address the refreshing of that view.

PR #877, opened 2019-10-04
👩‍🔬 [Investment Day] - Sort feature flags alphabetically

dusi:
# 📲 What

As the title says.

# 🤔 Why

Dealing with feature flags it's often confusing the two feature flags change order (this is because we're not sorting the returned array form the JSON). This should fix that.

# 🛠 How

Sorting based on the `description` field.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screen Shot 2019-10-04 at 1 22 44 PM | Screen Shot 2019-10-04 at 1 22 45 PM |

# ✅ Acceptance criteria

- [ ] Feature flags are sorted alphabetically


Library/ViewModels/FeatureFlagToolsViewModelTests.swift
```diff
@@ -7,7 +7,7 @@ import ReactiveSwift
 import XCTest
 
 final class FeatureFlagToolsViewModelTests: TestCase {
-  private let vm = FeatureFlagToolsViewModel()
+  private let vm: FeatureFlagToolsViewModelType = FeatureFlagToolsViewModel()
```
dusi: This is so that we don't have direct access to `VM`'s properties but have to go through `input`/`output`

cdolm92: Thanks for changing this! 👍🏾

PR #884, opened 2019-10-09
[NT-226] Cancel Backing Integration

ifbarrera:
# 📲 What

Integrates the cancel backing flow with the mutation. Note that this work *does not* include error handling.

# 🤔 Why

For the cancel backing flow.

# 🛠 How

- adds the `CancelBackingMutation` file and `CancelBackingInput` file, and adds a `cancelBacking` function to our `ServiceType`
- hooks up the cancel backing mutation to the `cancelPledgeButton`, plus adds handling for dismissing the keyboard on return
- sets up delegates to handle when the cancel mutation succeeds
- adds enabling/disabling behavior for the cancel pledge button

Note: the recording didn't properly capture the message banner animation. If you run it on a device the animation is working as expected.

# 👀 See

| Cancel Pledge | Cancel Pledge Button Disabled |
| --- | --- |
| VrGImUty6P | SvFIROdfIz |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] On staging, back a project *that has been created on staging*. Then navigate to your backing, and on the Manage Pledge screen navigate to "Cancel your pledge". Enter a cancellation reason. Then tap "Yes, cancel it". You should see in the console that the input parameters for the mutation should include the field "note" with your cancellation reason. You should then be taken back to the project page and you should see the message banner with the message "You've cancelled your pledge.".

- [x] Repeat the test above *without* entering a cancellation reason. Everything should work as expected.


Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -4,18 +4,30 @@ import Library
 import Prelude
 import UIKit
 
+protocol CancelPledgeViewControllerDelegate: AnyObject {
+  func cancelPledgeViewController(
+    _ viewController: CancelPledgeViewController,
+    didCancelPledgeWith message: String
+  )
```
justinswart: I think in this case we'd want like:
```swift
func cancelPledgeViewController(
  _ viewController: CancelPledgeViewController,
  didCancelPledgeWithMessage message: String
)
```
Just because it isn't obvious at the callsite what `String` is.

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -161,12 +194,28 @@ final class CancelPledgeViewController: UIViewController {
   // MARK: - Accessors
 
   @objc private func dismissKeyboard() {
-    self.view.endEditing(true)
+    self.viewModel.inputs.textFieldShouldReturn()
```
justinswart: Do we want to represent this event separately from `textFieldShouldReturn`?

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
```diff
@@ -161,12 +194,28 @@ final class CancelPledgeViewController: UIViewController {
...
+  }
+}
+
+extension CancelPledgeViewController: UITextFieldDelegate {
+  func textFieldDidEndEditing(_ textField: UITextField) {
+    self.viewModel.inputs.textFieldDidEndEditing(with: textField.text)
+  }
+
+  func textFieldShouldReturn(_: UITextField) -> Bool {
+    self.viewModel.inputs.textFieldShouldReturn()
```
justinswart: We could consider returning the boolean from the viewModel and writing a test for it.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -3,7 +3,17 @@ import Library
 import Prelude
 import UIKit
 
+protocol ManagePledgeViewControllerDelegate: AnyObject {
+  func managePledgeViewController(
+    _ viewController: ManagePledgeViewController,
+    shouldDismissAndShowSuccessBannerWith message: String
```
justinswart: Maybe here too, but tbh I'm not sure how consistent we've been with this 🤷‍♂

KsApi/mutations/inputs/CancelBackingInput.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+/**
```
justinswart: Think we'd usually put the comment about the `init`.

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
ifbarrera: Since we aren't doing any validation here, I'm not sure it's necessary because we always return `true`. If we were to actually add validation where the result of the validation would dictate the response, then yeah for sure that would go through the view model.

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
ifbarrera: Ok you got me there.

justinswart: lol that's not the goal of my PR reviews 😂

Kickstarter-iOS/Views/Controllers/CancelPledgeViewController.swift
justinswart: That's fair 👍

PR #887, opened 2019-10-10
[NT-357][NT-358] Change Payment Method UI

justinswart:
# 📲 What

An unfortunately large PR that adds _Change payment method_ UI to `PledgeViewController`.

**Note:** This is UI-only, we'll add the mutation request in a subsequent PR, as well as the ability to update Apple Pay.

# 🤔 Why

Allows us to update the payment method for a pledge.

# 🛠 How

This repurposes the `PledgeViewController` to configure its views based on a third `PledgeViewContext`, `.changedPaymentMethod`. It also removes the pledge button from `PledgePaymentMethodsViewController` so that there is a single `submitButton` which triggers different events based on the context. We could consider doing this for the _continue_ button too.

- Added `changePaymentMethod` case to `PledgeViewContext` enum.
- Added convenience vars to `PledgeViewContext` which can be more easily map'd over in `PledgeViewModel`.
- Tests for `PledgeViewContext`.
- Tests for `PledgeViewModel` for the new context.

# 👀 See

# ✅ Acceptance criteria

Navigate to _Manage Pledge_ on a project that you've backed, select `Change payment method` from the context menu.

- [ ] You should see the amount that you backed.
- [ ] If the reward had shipping enabled, this should be shown below the amount.
- [ ] The total should be shown and the T&Cs.
- [ ] The payment methods scrollview should be visible and you should be able to select a new payment method.
- [ ] The confirm button should enable/disable based on whether you've selected the existing payment method for the reward. **Note:** It appears this isn't currently working because the comparison needs to be made using a GraphID (will fix).

# ⏰ TODO

- [ ] Wire up `submitButton` to perform the `updateBacking` mutation in a subsequent PR.
- [ ] Wire up ability to update Apple Pay payment.


Kickstarter-iOS/Views/Controllers/PledgeAmountSummaryViewController.swift
```diff
@@ -0,0 +1,128 @@
+import KsApi
+import Library
+import Prelude
+import Prelude_UIKit
+import UIKit
+
+final class PledgeAmountSummaryViewController: UIViewController {
```
justinswart: This is just copy-pasting the required views from `ManagePledgeSummaryView`.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -68,19 +59,12 @@ final class PledgePaymentMethodsViewController: UIViewController {
       action: #selector(PledgePaymentMethodsViewController.applePayButtonTapped),
       for: .touchUpInside
     )
-
-    self.pledgeButton.addTarget(
```
justinswart: This is replaced by `submitButton` in `PledgeViewController`.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -60,6 +59,10 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
 
   internal var messageBannerViewController: MessageBannerViewController?
 
+  private lazy var pledgeAmountSummaryViewController: PledgeAmountSummaryViewController = {
+    PledgeAmountSummaryViewController.instantiate()
```
justinswart: Not the most amazing name for this VC, open to suggestions.

Library/PledgeViewContext.swift
```diff
@@ -0,0 +1,92 @@
+import Foundation
+
+public enum PledgeViewContext {
+  case pledge
+  case update
+  case changePaymentMethod
+
+  var confirmationLabelHidden: Bool {
```
justinswart: These booleans basically map directly in `PledgeViewModel`, it just helped for readability to do this here. They're tested in both places.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -573,3 +578,63 @@ private func attributedConfirmationString(with project: Project) -> NSAttributed
...
+  return backing.paymentSource?.id != paymentSourceId
+}
+
+private func allValuesChangedAndValid(
+  amountValid: Bool,
+  shippingRuleValid: Bool,
+  paymentSourceValid: Bool,
+  context: PledgeViewContext
+) -> Bool {
+  if context.isUpdating {
```
justinswart: In an updating context we enable the submit button if anything has changed and his valid. In a creating context we require all values to be valid.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1223,46 +1273,92 @@ final class PledgeViewModelTests: TestCase {
...
 
-    self.confirmButtonHidden.assertValues([false])
-    self.confirmButtonEnabled.assertValues(
-      [false, true, false, true, false], "Amount and shipping rule unchanged"
+    self.submitButtonEnabled.assertValues(
+      [false, false, false, true, false, true, false], "Amount and shipping rule unchanged"
+    )
+  }
+
+  func testUpdatingConfirmButtonEnabled_NoShipping() {
```
justinswart: Note to self that I need to add a test here for `testUpdatingConfirmButtonEnabled_ChangePaymentMethod` 😅

Kickstarter-iOS/Views/Controllers/PledgeAmountSummaryViewController.swift
justinswart: As discussed on Zoom, I'll create a follow-up PR to incorporate this as as subview of `ManagePledgeSummaryViewController` to promote reuse.

Library/ViewModels/PledgeAmountSummaryViewModel.swift
```diff
@@ -0,0 +1,131 @@
...
+  public var inputs: PledgeAmountSummaryViewModelInputs { return self }
+  public var outputs: PledgeAmountSummaryViewModelOutputs { return self }
+}
+
+private func shouldHideShippingLocationStackView(_ project: Project) -> Bool {
+  guard let backing = project.personalization.backing,
+    let _ = backing.locationName else {
+    return true
+  }
+  if let reward = backing.reward {
```
ifbarrera: This is interesting. Do you think we need this extra layer of logic? I'd imagine if the `backing.locationName` and `backing.locationId` are not nil that you could assume that the backing was created with a shipping address. 🤔

I wonder if this could cause a bug in a situation where the reward's shipping eligibility is modified by the creator *after* the user has created a pledge (not sure if that's even possible).

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -160,15 +162,19 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
   func testGoToChangePaymentMethod() {
-    self.vm.inputs.configureWith(Project.template, reward: .template)
+    let project = Project.template
+    let reward = Reward.template
+
+    self.vm.inputs.configureWith(project, reward: reward)
     self.vm.inputs.viewDidLoad()
 
-    self.goToChangePaymentMethod.assertDidNotEmitValue()
+    self.goToChangePaymentMethodProject.assertDidNotEmitValue()
```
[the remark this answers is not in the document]
justinswart: Ah yes, thanks!

Library/ViewModels/PledgeAmountSummaryViewModel.swift
```diff
@@ -0,0 +1,131 @@
...
+      }
+      .skipNil()
+
+    let shippingAmount = backing
+      .map { Double($0.shippingAmount ?? 0) }
+
+    self.shippingAmountText = project
+      .combineLatest(with: shippingAmount)
+      .map { project, shippingAmount in
+        shippingValue(with: project, with: shippingAmount)
```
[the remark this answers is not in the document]
justinswart: Oh, sure, this was a copy-paste 😄

Library/ViewModels/PledgeAmountSummaryViewModel.swift
justinswart: Another copy-paste, I'll check it out, thanks!

Library/ViewModels/PledgeAmountSummaryViewModel.swift
```diff
@@ -0,0 +1,131 @@
...
+public class PledgeAmountSummaryViewModel: PledgeAmountSummaryViewModelType,
+  PledgeAmountSummaryViewModelInputs, PledgeAmountSummaryViewModelOutputs {
+  public init() {
+    let project = Signal.combineLatest(
+      self.projectSignal,
+      self.viewDidLoadProperty.signal
+    )
+    .map(first)
+
+    let backing = project
```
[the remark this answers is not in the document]
justinswart: I think seeing as the `Backing` is from the `Project` I'm going to leave it as is for now.

Library/ViewModels/PledgeAmountSummaryViewModel.swift
justinswart: Confirmed with team R&F that this cannot occur, so checking that `locationId` is not `nil` on the `Backing` is sufficient. Good catch!

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -11,10 +11,9 @@ protocol PledgeViewControllerDelegate: AnyObject {
 final class PledgeViewController: UIViewController, MessageBannerViewControllerPresenting {
   // MARK: - Properties
 
-  private lazy var confirmButton: UIButton = { UIButton(type: .custom) }()
```
dusi: Yeah this will really mess up my loading button but it should be fairly simple to fix

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -75,8 +78,14 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
       |> \.delegate .~ self
   }()
 
+  private lazy var submitButton: UIButton = { UIButton(type: .custom) }()
```
dusi: Is `init with type` the designated initializer?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -40,7 +36,8 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
     let configureWithValue = Signal.combineLatest(
       self.viewDidLoadProperty.signal,
       self.configureWithValueProperty.signal.skipNil()
-    ).map(second)
+    )
+    .map(second)
```
dusi: https://www.youtube.com/watch?v=SsoOG6ZeyUI

😂

Kickstarter-iOS/Views/Controllers/PledgeAmountSummaryViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+    ], self.rootStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+  }
+}
+
+// MARK: Styles
+
+private let amountLabelStyle: LabelStyle = { label in
+  label
+    |> \.adjustsFontForContentSizeCategory .~ true
```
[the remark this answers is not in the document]
justinswart: `pledgeAmountLabel` uses an attributed string so the font setting isn't necessary here.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: Yeah we want this to be a `custom` button to not have the highlight animation that `system` buttons get.

Kickstarter-iOS/Views/Controllers/PledgeAmountSummaryViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+
+  public func configureWith(_ project: Project) {
+    self.viewModel.inputs.configureWith(project)
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.configureSubviews()
+    self.bindViewModel()
```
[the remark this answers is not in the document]
justinswart: Indeed, thanks.

justinswart:
@dusi Good catch on the snapshots - This is because the template `Reward`'s `ShippingRule` had an amount of `5` but the `Backing` on the `Project` had a shipping amount of `10`. I _think_ this is ok because in reality when we're changing the payment method these things should line up, but let's test this logic a bit more to be sure.

PR #888, opened 2019-10-10
[NT-204] Manage pledge payment section apple pay

Scollaco:
# 📲 What 
- Shows ApplePay flag on ManagePledgeViewController if project was pledged using ApplePay

# 🤔 Why
- So the user can see not only the card infos (number, expiration, etc), but also if the pledge was done through ApplePay.

# 🛠 How
- Creating a `PaymentType` enum
- Adding the ApplePay icon provided by Apple
- Updating tests

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# 📖 Story

NT-204

# ✅ Acceptance criteria

Choose a project backed using ApplePay and go to the ManagePledge screen :
- [ ] The ApplePay icon and the infos of the card used should be visible.
- [ ] If a project was backed using credit card (not using ApplePay), the card brand icon and its infos should be visible.


Kickstarter-iOS/Assets.xcassets/icons/credit cards/icon--apple_pay.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "icon--apple_pay.pdf"
```
ifbarrera: I think we've typically preferred dashes over underscores for asset names, so this should be `icon--apple-pay` instead of `icon--apple_pay`

KsApi/models/BackingTests.swift
```diff
@@ -14,7 +14,7 @@ final class BackingTests: XCTestCase {
         "expiration_date": "2019-09-23",
         "id": 20,
         "last_four": "1234",
-        "payment_type": "CREDIT_CARD",
+        "payment_type": "APPLE_PAY",
```
ifbarrera: Is it necessary to change this?

KsApi/models/GraphUserCreditCard.swift
```diff
@@ -27,6 +27,11 @@ public struct GraphUserCreditCard: Swift.Decodable {
     }
   }
 
+  public enum PaymentType: String, Swift.Decodable, CaseIterable {
+    case applePay = "APPLE_PAY"
+    case card = "CREDIT_CARD"
```
ifbarrera: I wonder if `creditCard` would be more explicit here so we can differentiate between credit cards and debit cards in the future.

KsApi/models/BackingTests.swift
Scollaco: Not really, I had changed it just to make sure everything was working. I will revert it 👍

PR #892, opened 2019-10-11
[NT-394] - Loading Button

dusi:
# 📲 What

Adds a `LoadingButton` class that could be used to display a button that has a loading state.

# 🤔 Why

Our CTA buttons will be performing network requests so this came up as a design tweak to our checkout workflow.

# 🛠 How

The button has a `Bool` property `isLoading` that could be toggled ON/OFF in order to show the activity indicator or hide it. This in addition requires storing titles for couple states into a temporary dictionary so that these could be restored later when the loading is stopped.

# 👀 See

**Pledge**

pledge

**Update pledge**

update

# ♿️ Accessibility 

- [ ] When the spinner is visible, its accessibility label says `Loading` and it's localized (this could be verified by slowing down your connection using Network Link Conditioner and using VoiceOver to tap on the button
- [ ] After the spinner is stopped, its accessibility label is using the title again (this could be verified by manually disabling push to thank you screen or the pop back on update manage pledge screen) and using VoiceOver to tap on the button

# ✅ Acceptance criteria

- [ ] The `Pledge` button shows a spinner when tapped and during the network call on the pledge screen
- [ ] The `Confirm` button shows a spinner when tapped and during the network call on the update pledge screen
- [ ] Button's user interaction is disabled when loading is ON
- [ ] Button's user interaction is re-enabled when loading finishes (this could be verified by manually disabling push to thank you screen or the pop back on update manage pledge screen)


Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -25,7 +25,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
   private lazy var cardsStackView: UIStackView = { UIStackView(frame: .zero) }()
   internal weak var delegate: PledgePaymentMethodsViewControllerDelegate?
   internal weak var messageDisplayingDelegate: PledgeViewControllerMessageDisplaying?
-  private lazy var pledgeButton: UIButton = { UIButton.init(type: .custom) }()
+  private(set) lazy var pledgeButton: LoadingButton = { LoadingButton(frame: .zero) }()
```
dusi: I can expose this another way but thought this was simplest for what we need to do (RAC binding)

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -11,7 +11,7 @@ protocol PledgeViewControllerDelegate: AnyObject {
 final class PledgeViewController: UIViewController, MessageBannerViewControllerPresenting {
   // MARK: - Properties
 
-  private lazy var confirmButton: UIButton = { UIButton(type: .custom) }()
+  private lazy var confirmButton: LoadingButton = { LoadingButton(frame: .zero) }()
```
dusi: Changing to `LoadingButton` as well as using designated initializer

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
+import Library
+import Prelude
+import UIKit
+
+final class LoadingButton: UIButton {
+  // MARK: - Properties
+
+  private lazy var activityIndicator: UIActivityIndicatorView = {
+    UIActivityIndicatorView(style: .white)
```
dusi: Style is something that could potentially be injected in the initializer in the future but for now I think this style will suffice

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+      |> \.hidesWhenStopped .~ true
```
dusi: This is the magic `prop` that does not require us to add/remove it from the view hierarchy...it will automatically show/hide the component depending on the `isAnimating` property.

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+
+  override init(frame: CGRect) {
+    super.init(frame: frame)
+
+    self.addSubview(self.activityIndicator)
+    self.activityIndicator.centerXAnchor.constraint(equalTo: self.centerXAnchor).isActive = true
+    self.activityIndicator.centerYAnchor.constraint(equalTo: self.centerYAnchor).isActive = true
+  }
+
+  required init?(coder: NSCoder) {
```
dusi: Init with coder so that this supports XIBs

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+    super.init(coder: coder)
+
+    self.addSubview(self.activityIndicator)
+    self.activityIndicator.centerXAnchor.constraint(equalTo: self.centerXAnchor).isActive = true
+    self.activityIndicator.centerYAnchor.constraint(equalTo: self.centerYAnchor).isActive = true
+  }
+
+  override func setTitle(_ title: String?, for state: UIControl.State) {
+    // Do not allow changing the title while the activity indicator is animating
+    guard !self.activityIndicator.isAnimating else { return }
```
dusi: Setting a title while the activity indicator is loading would display both the indicator and the title so I think this is the right approach here..we can potentially add logic to stop animating if we're setting title but I think due to our RAC bindings it could cause some flickering

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+    self.originalTitles[highlightedState.rawValue] = self.title(for: highlightedState)
+    self.originalTitles[normalState.rawValue] = self.title(for: normalState)
+    self.originalTitles[selectedState.rawValue] = self.title(for: selectedState)
+
+    self.setTitle(nil, for: disabledState)
+    self.setTitle(nil, for: highlightedState)
+    self.setTitle(nil, for: normalState)
+    self.setTitle(nil, for: selectedState)
+
+    _ = self
```
dusi: This makes sure that during the loading phase it's still accessible..

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+    self.setTitle(self.originalTitles[highlightedState.rawValue], for: highlightedState)
+    self.setTitle(self.originalTitles[normalState.rawValue], for: normalState)
+    self.setTitle(self.originalTitles[selectedState.rawValue], for: selectedState)
+
+    self.originalTitles[disabledState.rawValue] = nil
+    self.originalTitles[highlightedState.rawValue] = nil
+    self.originalTitles[normalState.rawValue] = nil
+    self.originalTitles[selectedState.rawValue] = nil
+
+    _ = self
```
dusi: And this ensures that after loading is done it will again use title for the accessibility label

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,115 @@
...
+      } else {
+        self.stopLoading()
+      }
+
+      _ = self
+        |> \.isUserInteractionEnabled .~ !self.isLoading
+    }
+  }
+
+  private var originalTitles: [UInt: String] = [:]
```
justinswart: Is there a reason you chose to do this without a view model?

dusi: I simply tried to go with the simplest approach but can back this by a VM to be more consistent with the rest of the codebase 👍

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,159 @@
...
+
+  // MARK: - Titles
+
+  private func removeTitle() {
+    let disabledState = UIControl.State.disabled
+    let highlightedState = UIControl.State.highlighted
+    let normalState = UIControl.State.normal
+    let selectedState = UIControl.State.selected
+
+    if let disabledTitle = self.title(for: disabledState) {
```
dusi: Extracting the title first will make sure that if for some reason we want to start the label with a loading state, we don't just clean the titles, but rather only cache previous titles if there were any.

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,159 @@
...
+
+  private func stopLoading() {
+    self.activityIndicator.stopAnimating()
+
+    self.restoreTitle()
+  }
+
+  // MARK: - Titles
+
+  private func removeTitle() {
```
justinswart: What do you think about something like this for DRYness:

```swift
  private func removeTitle() {
    let states: [UIControl.State] = [.disabled, .highlighted, .normal, .selected]

    states.compactMap { state -> (String, UIControl.State)? in
      guard let title = self.title(for: state) else { return nil }
      return (title, state)
    }
    .forEach { title, state in
      self.originalTitles[state.rawValue] = title
      self.setTitle(nil, for: state)
    }

    _ = self
      |> \.accessibilityLabel %~ { _ in Strings.Loading() }

    UIAccessibility.post(notification: .layoutChanged, argument: self)
  }
```

Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -0,0 +1,159 @@
...
+      self.setTitle(nil, for: selectedState)
+    }
+
+    _ = self
+      |> \.accessibilityLabel %~ { _ in Strings.Loading() }
+
+    UIAccessibility.post(notification: .layoutChanged, argument: self)
+  }
+
+  private func restoreTitle() {
```
justinswart: You could probably do this here too.

justinswart:
PS. Thanks for everything! 🎉

PR #893, opened 2019-10-11
[NT-359] Integrate Updating Backing mutation

justinswart:
# 📲 What

A continuation of the work in #887 and #890, this PR is also based off #887's branch.

This wires up the `UpdateBacking` mutation to allow one to change their payment method and Apple Pay backing.

# 🤔 Why

This is a continuation of the work to allow a user to manage their pledge.

# 🛠 How

- Added `paymentSourceId` and `ApplePayParams` to `UpdateBackingData`.
- Created somewhat duplicate signals to handle the result of the apple pay calls. I think when we do the work for the `CreateBacking` mutation, I may have a way to make this even more reusable due to that mutation and `UpdateBacking` being so similar.

# 👀 See

**Note:** Currently the manage pledge view does not refresh automatically, this is a known issue that will be addressed in a subsequent PR. Also, currently the credit card that is the current payment source for the backing is meant to be selected by default and ordered first. This will also be addressed in an upcoming PR. So, at the moment, the experience is that when you land on _Change payment method_, you may see the incorrect card selected, but the behaviour of the submit button enabling/disabling should still be correct and allow you to change it.

# ✅ Acceptance criteria

Be sure to test this against the native dev environment. The payment method that you use and the project that you back should all be created on that environment. Navigate to _Change payment method_ in the _Manage pledge view_.

- [ ] Select a new payment method and hit confirm you should see a success message and your change reflected (with the caveats mentioned above).
- [ ] Select Apple Pay, the backing should change to an Apple Pay backing*

\* I was unable to test this unfortunately as I didn't have an iOS 12 device with me at the time and, although I think it should work, Apple Pay doesn't seem to do anything on the simulator. This is possible a Stripe issue? 🤔


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -276,14 +283,19 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       applePayStatusFailure
     )
 
-    // MARK: - Apple Pay
+    // MARK: - Create Apple Pay Backing
+
+    let willCreateApplePayBacking = Signal.combineLatest(
```
justinswart: This and `willUpdateApplePayBacking` are to ensure that upon completion of these asynchronous events we show the appropriate messaging/view to the user.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -428,6 +431,82 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       createBackingEvent.filter { $0.isTerminating }.mapConst(true),
       updateBackingEvent.filter { $0.isTerminating }.mapConst(true)
     )
+
+    // MARK: - Success/Failure Create
+
+    let createPaymentAuthorizationDidFinishSignal = Signal.zip(
```
justinswart: `zip` allows us to ensure that each of these signals will emit at least once before this signal emits. So we know that we started out creating here and not updating because we `zip` with the `willCreateApplePayBacking` signal.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1441,4 +1441,192 @@ final class PledgeViewModelTests: TestCase {
...
+
+      self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
+      self.updatePledgeFailedWithError.assertDidNotEmitValue()
+      self.submitButtonEnabled.assertValues([false])
+      self.createBackingError.assertDidNotEmitValue()
+      self.goToThanks.assertDidNotEmitValue()
+
+      XCTAssertEqual(
+        PKPaymentAuthorizationStatus.success,
+        self.vm.inputs.stripeTokenCreated(token: "stripe_token", error: nil)
```
ifbarrera: We might want to add an additional test for failure in creating a stripe token - so we would pass an error here and assert that the status is `PKPaymentAuthorizationStatus.failure`, and that none of the signals emit after running the scheduler (ie. the network request for `updateBacking` never goes out).

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -428,6 +431,82 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    // MARK: - Success/Failure Update
+
+    let updatePaymentAuthorizationDidFinishSignal = Signal.zip(
+      willUpdateApplePayBacking,
+      self.paymentAuthorizationDidFinishSignal
+    )
+
+    let updateBackingEventErrors = Signal.zip(
+      updateBackingEvent.errors(),
+      self.submitButtonTappedSignal
```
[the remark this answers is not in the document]
justinswart: I this case I don't think we need `zip`, thanks. 👍

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -428,6 +431,82 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+
+    let updateApplePayBackingSuccess = updateBackingEvent.values()
+      .takeWhen(updatePaymentAuthorizationDidFinishSignal)
+    let updateApplePayBackingError = updateBackingEvent.errors()
+      .takeWhen(updatePaymentAuthorizationDidFinishSignal)
+
+    let updateBackingDidCompleteApplePay = updateApplePayBackingSuccess
+      .ignoreValues()
+    let updateBackingDidComplete = Signal.zip(
+      self.submitButtonTappedSignal,
```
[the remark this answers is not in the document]
justinswart: I'm no longer convinced it would do what I need it to here, so I'll consider an alternative.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1441,4 +1443,276 @@ final class PledgeViewModelTests: TestCase {
...
+      self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
+      self.updatePledgeFailedWithError.assertDidNotEmitValue()
+      self.submitButtonEnabled.assertValues([false])
+      self.createBackingError.assertDidNotEmitValue()
+      self.goToThanks.assertDidNotEmitValue()
+
+      self.scheduler.run()
+
+      self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
+      self.updatePledgeFailedWithError.assertValues([
```
ifbarrera: Hmm so if the request errors before the payment authorization VC has dismissed, the error banner might show below the payment authorization VC 🤔 that's probably not ideal?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1256,7 +1258,7 @@ final class PledgeViewModelTests: TestCase {
     }
   }
 
-  func testUpdatingConfirmButtonEnabled_ShippingEnabled() {
+  func testUpdatingSubmitButtonEnabled_ShippingEnabled() {
```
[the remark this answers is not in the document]
justinswart: Yes, it's not always a confirm button anymore, sometimes a pledge button. We should also remove the continue button and just use this one.

PR #894, opened 2019-10-12
[NT-363] Select a new reward

justinswart:
# 📲 What

Continues work in #887, #893, #890. This is based off the branch of #893.

Adds yet another context to `PledgeViewModel`, `updateReward`. Attempts to treat updating a reward the same as creating a new pledge, save for the fact that we make use of the `UpdateBacking` mutation.

# 🤔 Why

This allows backers to update the reward that they've selected for a project.

# 🛠 How

- Added `updateReward` to `PledgeViewContext`.
- Changed the behaviour of `PledgeShippingLocationViewModel` to only return the currently selected `ShippingRule` _if_ the new reward also shared that location.
- Changed the behaviour of `PledgeAmountViewModel` to return the initial value as being the minimum for the newly selected reward.
- Updated the validation rules in `PledgeViewModel` to ignore values being the same when a new reward is being selected.

# ✅ Acceptance criteria

Back a project using the correct environment and with the stars aligned, navigate to manage pledge, select _Choose another reward_ from the context menu.

- [ ] The `RewardsCollectionViewController` is presented and allows you to select a different reward taking you to the `PledgeViewController`.
- [ ] The amount should be the minimum amount of the new reward.
- [ ] The shipping location should be the one that was previously selected _if_ the new reward also provides that location as an option.

\* As mentioned in #893 the payment method does not have the correct default selection behaviour, although this may be the way that we'd prefer it for updating a reward.


justinswart:
@ifbarrera I'll update this to hide the payment methods. I think we'll still need the additional context for that because it looks like we do show the summary view with the thumbnail.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1423,6 +1468,45 @@ final class PledgeViewModelTests: TestCase {
...
+
+    self.vm.inputs.pledgeAmountViewControllerDidUpdate(
+      with: (amount: 690, min: 25.0, max: 10_000.0, isValid: true)
+    )
+
+    self.submitButtonEnabled.assertValues([false, false], "Amount unchanged")
+
+    self.vm.inputs.shippingRuleSelected(.init(cost: 1, id: 1, location: .brooklyn))
+
+    self.submitButtonEnabled.assertValues(
```
ifbarrera: Should we also test for the behavior when the submit button is tapped? 🤔

Library/ViewModels/PledgeShippingLocationViewModel.swift
```diff
@@ -163,8 +163,13 @@ private func shippingValue(of project: Project, with shippingRuleCost: Double) -
 private func determineShippingRule(with project: Project, shippingRules: [ShippingRule]) -> ShippingRule? {
   guard
     let backing = project.personalization.backing,
-    let selectedRule = shippingRules.filter({ $0.location.id == backing.locationId }).first
+    let selectedRule = shippingRules.filter({ $0.location.id == backing.locationId }).first,
+    locationId(selectedRule.location.id, isValidIn: shippingRules)
```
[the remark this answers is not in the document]
justinswart: Let's say you picked Reward A and selected Canada as your shipping location. Now if you're changing to Reward B it will pre-select Canada, but only if Reward B _also_ ships to Canada. So the `locationId(_:isValidIn:)` check is this second bit of logic. If it fails it returns the default shipping rule.

Library/ViewModels/PledgeViewModelTests.swift
justinswart: Ah yes, thanks.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1423,6 +1468,83 @@ final class PledgeViewModelTests: TestCase {
...
+      self.submitButtonTitle.assertValues(["Confirm"])
+      self.submitButtonEnabled.assertValues([false])
+      self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
+      self.updatePledgeFailedWithError.assertDidNotEmitValue()
+
+      self.vm.inputs.pledgeAmountViewControllerDidUpdate(
+        with: (amount: 690, min: 25.0, max: 10_000.0, isValid: true)
+      )
+
+      self.submitButtonEnabled.assertValues([false, false], "Amount unchanged")
```
ifbarrera: I might be missing something with this test, but why is the submit button disabled here? I would expect that because the VM is being configured with `reward` (which is different from `project.backing.reward`) that the submit button would be enabled because the pledge amount is between the min and max _of the new reward_. What am I missing here?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1453,42 +1575,46 @@ final class PledgeViewModelTests: TestCase {
       with: (amount: 690, min: 25.0, max: 10_000.0, isValid: true)
     )
 
+    self.submitButtonEnabled.assertValues([false, false], "Amount unchanged")
+
     self.vm.inputs.shippingRuleSelected(.init(cost: 1, id: 1, location: .brooklyn))
 
-    self.submitButtonEnabled.assertValues([false, false], "Shipping rule and amount unchanged")
+    self.submitButtonEnabled.assertValues([false, false, false], "Shipping rule and amount unchanged")
```
ifbarrera: Should we consider throwing a `skipRepeats()` on this output? 😅

Library/ViewModels/PledgeViewModelTests.swift
justinswart: The reward has shipping enabled so the way the validation rules are working is that a shipping rule needs to be selected in order to change the reward.

Library/ViewModels/PledgeViewModelTests.swift
justinswart: Maybe, but not right now 😅

PR #895, opened 2019-10-15
[NT-422] Payment source refactor

Scollaco:
# 📲 What 
- Creates `PaymentSource` to be used on the manage pledge context.

# 🤔 Why
- We were reusing the `GraphUserCreditCard` model, but due some differences between the payloads returned by `v1` and `graphQL`, we've decided to create a new model to avoid work arounds.

# 🛠 How
- Creating `PaymentSource` and making it Argo decodable.
- Removing unnecessary properties of GraphUserCreditCard model
- Creating `ManagePledgePaymentMethodViewModel`
- Updating, fixing and adding tests

# 👀 See
- No differences should be noticed.

# ✅ Acceptance criteria

- [ ] No differences should be noticed on `ManagePledgeViewController` or `ManagePledgePaymentMethodView`


KsApi/models/Backing.swift
```diff
@@ -21,6 +21,29 @@ public struct Backing {
   public let shippingAmount: Int?
   public let status: Status
 
+  public struct PaymentSource {
+    public let expirationDate: String?
+    public let id: String?
+    public let lastFour: String?
+    public let paymentType: PaymentType?
```
ifbarrera: I think this should be non-optional.

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,100 @@
+import KsApi
+import Prelude
+import ReactiveExtensions
+import ReactiveSwift
+import UIKit
```
ifbarrera: Can we get rid of the `UIKit` import here and just return the `imageName` as a string? I think we generally try to keep our VMs decoupled from UIKit.

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+
+public protocol ManagePledgePaymentMethodViewModelInputs {
+  /// Call to configure payment method section with payment source values.
+  func configureWith(value: Backing.PaymentSource)
+}
+
+public protocol ManagePledgePaymentMethodViewModelOutputs {
+  /// Emits the card's image.
+  var cardImage: Signal<UIImage?, Never> { get }
```
ifbarrera: Here this could be `cardImageName: Signal<String?, Never>`

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+
+public final class ManagePledgePaymentMethodViewModel: ManagePledgePaymentMethodViewModelInputs,
+  ManagePledgePaymentMethodViewModelOutputs, ManagePledgePaymentMethodViewModelType {
+  public init() {
+    self.cardImage = self.paymentSourceSignal
+      .map(cardImage(for:))
+
+    self.cardNumberAccessibilityLabel = self.paymentSourceSignal
+      .map {
+        [$0.type?.description, Strings.Card_ending_in_last_four(last_four: $0.lastFour ?? .init())]
```
ifbarrera: Is it possible to set this up such that if `type` and `lastFour` are nil, we don't send anything back in the output? It doesn't seem like we'd ever want a situation where the label reads `" , ending in "` 🤔

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+          .joined(separator: ", ")
+      }
+
+    self.cardNumberTextShortStyle = self.paymentSourceSignal
+      .map { $0.lastFour }
+      .skipNil()
+      .map { Strings.Ending_in_last_four(last_four: $0) }
+
+    self.expirationDateText = self.paymentSourceSignal
+      .map { String($0.expirationDate?.dropLast(3) ?? "") }
```
ifbarrera: Same here, I think we probably shouldn't have the signal output when the `expirationDate` is nil:

```
self.expirationDateText = self.paymentSourceSignal
  .map { $0.expirationDate }
  .skipNil()
  .map { String($0.dropLast(3)) }
... etc
```

KsApi/models/Backing.swift
Scollaco: I was following the description of this [ticket](https://dripsprint.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=NT-422&assignee=5d41b601f14e7c0da562b03f)

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
Scollaco: Good point. I will change that

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
Scollaco: I missed that one. Good catch!

KsApi/models/Backing.swift
ifbarrera: I think the ticket says it should be non-optional? 🤔 
image

Scollaco: OMG! paymentType is non-optional, but type is! Got lost between types 😅

KsApi/models/Backing.swift
```diff
@@ -68,12 +91,12 @@ extension Backing: Argo.Decodable {
  on that environment. This is a workaround to allow us to test on Staging and should be deleted once the
  data is being returned normally.
  */
-private func tryDecodePaymentSource(_ json: JSON?) -> Decoded<GraphUserCreditCard.CreditCard?> {
+private func tryDecodePaymentSource(_ json: JSON?) -> Decoded<Backing.PaymentSource?> {
```
justinswart: Do we still need this in this PR?

KsApi/models/Backing.swift
```diff
@@ -91,8 +114,27 @@ extension Backing: EncodableType {
   }
 }
 
+extension Backing.PaymentSource: Argo.Decodable {
```
justinswart: I think we could have this only conform to Swift's `Decodable` and do a similar function like before to return an `Argo.Decodable` `success` or `failure`?

justinswart: Just a thought, perhaps this PR is not the place to start doing that, but I think that can be an approach to help us to start moving away from Argo.

KsApi/models/Backing.swift
Scollaco: We do. I tried removing it, but same issue still happens on Staging.

KsApi/models/Backing.swift
```diff
@@ -91,8 +114,27 @@ extension Backing: EncodableType {
...
+  public static func decode(_ json: JSON) -> Decoded<Backing.PaymentSource?> {
+    return curry(Backing.PaymentSource.init)
+      <^> json <|? "expiration_date"
+      <*> (json <|? "id" <|> (json <| "id" >>- intToString))
```
ifbarrera: CJ merged the change to return a `String` instead of an `Int` so you should be able to remove this now! 🎉

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,109 @@
...
+import ReactiveSwift
+
+public protocol ManagePledgePaymentMethodViewModelInputs {
+  /// Call to configure payment method section with payment source values.
+  func configureWith(value: Backing.PaymentSource)
+}
+
+public protocol ManagePledgePaymentMethodViewModelOutputs {
+  /// Emits the card's image.
+  var cardImage: Signal<String, Never> { get }
```
ifbarrera: Maybe `cardImageName` would be more accurate here?

Library/ViewModels/ManagePledgePaymentMethodViewModelTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    self.cardNumberAccessibilityLabel.assertLastValue("Visa, Card ending in 1111")
+    self.cardNumberTextShortStyle.assertLastValue("Ending in 1111")
+    self.expirationDateText.assertLastValue("Expires 09/2019")
+  }
+
+  func testApplePay() {
+    self.vm.inputs.configureWith(value: Backing.PaymentSource.applePay)
+
+    self.cardImage.assertValue("icon--apple-pay")
+    self.cardNumberTextShortStyle.assertLastValue("Ending in 1111")
```
ifbarrera: Are we missing an assertion here for the accessibility label? And actually on second thought, do we want to update the accessibility label to indicate that the card is associated with Apple Pay?

Scollaco: Regarding the accessibility label, it's a good idea, but I think this ends up being out of scope for this PR.

PR #897, opened 2019-10-16
[NT-364]Unsupported cards carousel

cdolm92:
# 📲 What

We have disabled the select button, changed the button title to `No longer available` and provided copy below the unavailable credit card so that users understand that they cannot pledge with an unavailable card type.

# 🤔 Why

Users want to clearly and easily see which of their saved payment methods may not be used to back a particular project in the Pledge screen so that they know to select or add a different payment method to complete their pledge. 

# 🛠 How

We determine unavailable card types by checking if the user card is included in `project.availableCardTypes`.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone XS - 2019-10-16 at 16 57 59|Simulator Screen Shot - iPhone XS - 2019-10-16 at 16 45 58|

# ✅ Acceptance criteria

- [ ]  Navigate to a non-USD project. Scroll through the available payment options, all unacceptable payment options should have a disabled button with the title text that reads `No longer available`
- [ ]  Copy below the Cardview should read `You can’t use this credit card to back a project from [project location].` **Note: this will change to project country in an upcoming PR**

# ✍🏾 TO DO
- [x] Snapshot testing for cards


Library/ViewModels/PledgeCreditCardViewModel.swift
```diff
@@ -4,9 +4,14 @@ import ReactiveExtensions
 import ReactiveSwift
 import UIKit
 
+public typealias PledgeCreditCardValue = (
+  card: GraphUserCreditCard.CreditCard,
+  isEnabled: Bool, projectCountry: String?
```
ifbarrera: `projectCountry` probably doesn't have to be optional.

Library/ViewModels/PledgeCreditCardViewModel.swift
```diff
@@ -83,10 +101,19 @@ public final class PledgeCreditCardViewModel: PledgeCreditCardViewModelInputs,
...
+    self.unavailableCardText = self.creditCardProperty.signal.skipNil()
+      .map { Strings.You_cant_use_this_credit_card_to_back_a_project_from_project_country(
+        project_country: $0.projectCountry ?? ""
+      )
+      }
   }
 
-  fileprivate let creditCardProperty = MutableProperty<GraphUserCreditCard.CreditCard?>(nil)
-  public func configureWith(value: GraphUserCreditCard.CreditCard) {
+  fileprivate let creditCardProperty = MutableProperty<PledgeCreditCardValue?>(nil)
```
ifbarrera: Can we update the name of this property to match the fact that we now configure with `PledgeCreditCardValue`? Maybe `pledgeCreditCardValueProperty`?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -56,12 +63,20 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
     )
     .scan([]) { current, new in new + current }
 
+    let cardValues = Signal.combineLatest(
+      card,
+      availableCardTypes,
+      project.map { $0.location.displayableName }
+    ).map { $0 as CardViewValues }
+
+    self.reloadPaymentMethods = cardValues
```
ifbarrera: Here is where we could move even more logic out of the view controller – something we could do is instead of returning `CardViewValues`, return an array of `PledgeCreditCardValue`:

```
    self.reloadPaymentMethods = Signal.combineLatest(
      cards,
      availableCardTypes,
      project.map { $0.location.displayableName }
    ).map { cards, availableCardTypes, projectLocation in
      return cards.map { card in
        let isEnabled = availableCardTypes.contains(card.type.rawValue)

        return (card, isEnabled, projectLocation)
      }
    }
```

Then we can test that given a list of stored cards, and a list of `availableCardTypes` in a project, that the output is exactly what we're feeding into each `PledgeCreditCardViewModel` to configure the card view.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -14,7 +14,12 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
   private let notifyDelegateApplePayButtonTapped = TestObserver<Void, Never>()
   private let notifyDelegateCreditCardSelected = TestObserver<String, Never>()
   private let notifyDelegateLoadPaymentMethodsError = TestObserver<String, Never>()
-  private let reloadPaymentMethods = TestObserver<[GraphUserCreditCard.CreditCard], Never>()
+  private let notifyDelegatePledgeButtonTapped = TestObserver<Void, Never>()
+  private let pledgeButtonEnabled = TestObserver<Bool, Never>()
```
ifbarrera: Are these two outputs still supposed to be here? They might have been removed on master.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -38,19 +50,30 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+      self.reloadPaymentMethodsProjectCountry.assertDidNotEmitValue()
 
       self.vm.inputs.configureWith((User.template, Project.template, false))
       self.vm.inputs.viewDidLoad()
 
       self.scheduler.run()
 
-      self.reloadPaymentMethods.assertValue(response.me.storedCards.nodes)
+      self.reloadPaymentMethodsCards.assertValue(response.me.storedCards.nodes)
+      self.reloadPaymentMethodsAvailableCardTypes.assertValues([
```
ifbarrera: It's a little difficult to follow these assertions when they're split up. Is it possible to do the assertions on the tuples? Something like this:
```
self.reloadPaymentMethodsCardsAndAvailable.assertValues([(GraphUserCreditCard.visa, true), (GraphUser CreditCard.masterCard, true)...etc])
```

Library/ViewModels/PledgeCreditCardViewModelTests.swift
```diff
@@ -24,12 +28,16 @@ internal final class PledgeCreditCardViewModelTests: TestCase {
     self.vm.outputs.cardNumberTextShortStyle.observe(self.cardNumberTextShortStyle.observer)
     self.vm.outputs.expirationDateText.observe(self.expirationDateText.observer)
     self.vm.outputs.notifyDelegateOfCardSelected.observe(self.notifyDelegateOfCardSelected.observer)
+    self.vm.outputs.selectButtonEnabled.observe(self.selectButtonEnable.observer) //
```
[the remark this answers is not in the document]
cdolm92: oh right, I still need to test those

PR #901, opened 2019-10-17
[NT-438] Create backing with free shipping

ifbarrera:
# 📲 What

Adds support for creating a pledge with a reward that has free shipping.

# 🤔 Why

This was missed during the create backing implementation – often rewards ship for free, so we shouldn't be checking on the cost of shipping being greater than 0.

# 🛠 How

Also includes the following:
- refactoring shared logic between CreateApplePay, UpdateBacking and CreateBacking
- we now show the free shipping summary item in the Apple Pay share sheet, something we didn't used to show

# 👀 See

| Create Backing (Free shipping) | Update Backing (Free Shipping) |
| --- | --- |
| zLmNUuAEp5 | WdoGY9AdxU |

Also see the new free shipping summary item on the Apple Pay share sheet.

Simulator Screen Shot - iPhone 8 - 2019-10-17 at 16 47 22

# ✅ Acceptance criteria

- [x] Navigate to a pledge screen for a reward that has free shipping. Create a pledge - the pledge should succeed.
- [x] Navigate to update backing for a backing that has free shipping. Change your amount or shipping location. Confirm the backing update. This should succeed.


Library/SharedFunctions.swift
```diff
@@ -216,3 +216,27 @@ internal func classNameWithoutModule(_ class: AnyClass) -> String {
     .dropFirst()
     .joined(separator: ".")
 }
+
+typealias FormattedPledgeParameters = (pledgeTotal: String, rewardId: String?, locationId: String?)
```
justinswart: When I read `FormattedPledgeParameters` I think about this going through a formatter, I'm wondering if `SanitizedPledgeParameters` or `SanitizedPledgeParams` describes what we're doing more accurately?

PR #902, opened 2019-10-18
[NT-286][NT-360] SCA Support

justinswart:
# 📲 What

Adds SCA support to our create and update pledge flows.

# 🤔 Why

From the Stripe [docs](https://stripe.com/docs/strong-customer-authentication):

> Strong Customer Authentication (SCA), a rule in effect as of September 14, 2019, as part of PSD2 regulation in Europe, requires changes to how your European customers authenticate online payments. Card payments require a different user experience, namely 3D Secure, in order to meet SCA requirements. Transactions that don’t follow the new authentication guidelines may be declined by your customers’ banks.

# 🛠 How

- Added an output to `PledgeViewModel` to trigger the SCA flow.
- Added an input to `PledgeViewModel` to handle the response (succeeded/failed/canceled).
- During this process I noticed that we weren't being very DRY about handling the create and update signals. Their patterns are exactly the same so I refactored things to be more reusable. Tests passed without any changes except for the SCA stuff that was added.
- Also introduced `StripePaymentHandlerActionStatus` (an enum) and `StripePaymentHandlerActionStatusType` (a protocol) that allows us to more easily mock the Stripe response enum in tests.

**Note:** I've renamed `createApplePayBackingEvent` to `deprecatedCreateApplePayBackingEvent` in the expectation that it will be removed soon when we've updated our `CreateBacking` mutation to accept `ApplePayParams`.

# ✅ Acceptance criteria

Using the test cards listed [here](https://stripe.com/docs/testing#three-ds-cards):

**Note:** Please be aware of the caveats mentioned in [this thread](https://kickstarter.slack.com/archives/C3LS41US0/p1571406004140400) when testing.

- [x] Create a new pledge, the SCA test modal should be displayed.
- [ ] Update a pledge, the SCA test modal should be displayed.
- [x] Change payment method from a non-SCA card to an SCA card, the SCA test modal should be displayed.
- [x] When creating a pledge, confirming in the SCA modal should dismiss and take you to the thanks page.
- [x] When updating a pledge, confirming in the SCA modal should dismiss with success and return to manage plege.
- [ ] In any of the above scenarios producing an error from the modal (by way of an error test card), should dismiss and display the error in the error banner.
- [x] In any of the above scenarios dismissing the modal shouldn't do anything.
- [ ] No Apple Pay pledges should trigger this SCA flow.


Library/StripeTypes.swift
```diff
@@ -0,0 +1,37 @@
...
+  public var requiresSCAFlow: Bool {
+    return self.createBacking.checkout.backing.requiresAction
+  }
+
+  public var clientSecret: String? {
+    return self.createBacking.checkout.backing.clientSecret
+  }
+}
+
+public enum StripePaymentHandlerActionStatus {
```
justinswart: This type maps to `STPPaymentHandler.ActionStatus` and makes it easier for us to mock in tests without having to link the Stripe SDK to our Library framework.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -33,6 +33,7 @@ public protocol PledgeViewModelInputs {
   )
   func paymentAuthorizationViewControllerDidFinish()
   func pledgeAmountViewControllerDidUpdate(with data: PledgeAmountData)
+  func scaFlowCompleted(with result: StripePaymentHandlerActionStatusType, error: Error?)
```
justinswart: We pass in our protocol type here.

Library/StripeTypes.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import KsApi
+
+public protocol StripeSCARequiring {
```
justinswart: This protocol helps to merge the signals in `PledgeViewModel` for simplification.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      .takeWhen(updateCreateApplePayBackingCompleted)
 
-    let updateApplePayBackingSuccess = updateBackingEvent.values()
-      .takeWhen(updateApplePayBackingCompleted)
+    let creatingContext = context.filter { $0.isCreating }
 
-    let updateBackingDidCompleteApplePay = updateApplePayBackingSuccess
-      .ignoreValues()
+    #warning("To be removed once new CreateBacking is implemented to handle Apple Pay")
+    let deprecatedCreateApplePayBackingCompleted = Signal.combineLatest(
```
justinswart: Self-explanatory but this should be removed once we're handling Apple Pay through `CreateBacking`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -636,17 +667,23 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
   public let popViewController: Signal<(), Never>
   public let sectionSeparatorsHidden: Signal<Bool, Never>
   public let shippingLocationViewHidden: Signal<Bool, Never>
+  public let showErrorBannerWithMessage: Signal<String, Never>
```
justinswart: Errors are all emitted through one output now.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -379,6 +379,23 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
   @objc private func dismissKeyboard() {
     self.view.endEditing(true)
   }
+
+  private func beginSCAFlow(withClientSecret secret: String) {
+    STPPaymentHandler.shared().confirmSetupIntent(
+      withParams: .init(clientSecret: secret),
+      authenticationContext: self
+    ) { [weak self] status, _, error in
```
ifbarrera: Just curious what's this middle parameter here that we're ignoring?

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -379,6 +379,23 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
...
+      self?.viewModel.inputs.scaFlowCompleted(with: status, error: error)
+    }
+  }
+}
+
+// MARK: STPAuthenticationContext
```
ifbarrera: Can you add the `-` here

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    let paymentAuthorizationDidFinish = didInitiateApplePayBacking
+      .takeWhen(self.paymentAuthorizationDidFinishSignal)
 
-    self.createBackingError = Signal.merge(
-      createApplePayBackingError,
-      createBackingEventError
+    let createOrUpdateApplePayBackingCompleted = Signal.combineLatest(
+      Signal.merge(
+        updateBackingEvent.filter { $0.isTerminating }.ignoreValues(),
+        createBackingEvent.filter { $0.isTerminating }.ignoreValues()
```
ifbarrera: Should this be `deprecatedCreateApplePayBackingEvent`? 🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
 
-    let createApplePayBackingError = createApplePayBackingEventErrors
-      .takeWhen(createApplePayBackingCompleted)
+    let paymentAuthorizationDidFinish = didInitiateApplePayBacking
+      .takeWhen(self.paymentAuthorizationDidFinishSignal)
 
-    self.createBackingError = Signal.merge(
-      createApplePayBackingError,
-      createBackingEventError
+    let createOrUpdateApplePayBackingCompleted = Signal.combineLatest(
```
ifbarrera: Wondering if this should be a `zip` to account for multiple groups of events?:

1. User opens payment sheet and pledges - payment sheet dismisses, but pledge fails.
2. User opens up payment sheet a second time and pledges - because the `createBackingEvent.filter` event has already fired once before, the next `paymentAuthorizationDidFinish` event will cause the `createOrUpdateApplePayBackingCompleted` to fire without waiting for the next request to terminate.

🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    self.beginSCAFlowWithClientSecret = createOrUpdateBackingEventValuesRequiresSCA.map { $0.clientSecret }
+      .skipNil()
 
-    let updateApplePayBackingCompleted = Signal.combineLatest(
-      updateBackingEvent
-        .filter { $0.isTerminating }
-        .ignoreValues(),
-      updatePaymentAuthorizationDidFinishSignal
-    )
+    let didCompleteApplePayBacking = createOrUpdateBackingEventValues
```
ifbarrera: Curious why do we need to take `createOrUpdateBackingEventValues` here? That seems to require either `createBackingEvent.values()` or `updateBackingEvent.values()`.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1102,31 +1117,242 @@ final class PledgeViewModelTests: TestCase {
...
+      self.goToThanks.assertDidNotEmitValue()
+      self.showErrorBannerWithMessage.assertDidNotEmitValue()
+
+      self.scheduler.run()
+
+      self.beginSCAFlowWithClientSecret.assertDidNotEmitValue()
+      self.submitButtonEnabled.assertValues([false, true, false, true])
+      self.goToThanks.assertDidNotEmitValue()
+      self.showErrorBannerWithMessage.assertValues(
+        ["The operation couldn’t be completed. (KsApi.GraphError error 5.)"]
```
ifbarrera: This is a strange error message. Shouldn't this be "Something went wrong." which is the `localizedDescription` of `GraphError.invalidInput`?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1102,31 +1117,242 @@ final class PledgeViewModelTests: TestCase {
...
+      self.beginSCAFlowWithClientSecret.assertValues(["client-secret"])
+      self.submitButtonEnabled.assertValues([false, true, false, true])
+      self.goToThanks.assertDidNotEmitValue()
+      self.showErrorBannerWithMessage.assertDidNotEmitValue()
+
+      self.vm.inputs.scaFlowCompleted(
+        with: MockStripePaymentHandlerActionStatus(status: .succeeded), error: nil
+      )
+
+      self.beginSCAFlowWithClientSecret.assertValues(["client-secret"])
```
ifbarrera: This should probably be above the call to `inputs.scaFlowCompleted` no?

ifbarrera: Oh I see it's already on line 1205, apologies!

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
 
-    let updateBackingDidComplete = submitButtonTapped
-      .takeWhen(updateBackingEvent.values())
-      .filter(isTrue)
-      .ignoreValues()
+    self.goToThanks = project.takeWhen(createBackingCompletionEvents)
 
-    let updateBackingEventErrors = updateBackingEvent.errors()
+    let createUpdateBackingEventErrors = Signal.merge(
+      deprecatedCreateApplePayBackingEvent.errors().map { $0 as Error },
```
ifbarrera: I don't _think_ you should have to do this casting to `Error` 🤔

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -318,7 +315,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       )
     }
 
-    let createApplePayBackingEvent = createApplePayBackingData.map(
+    let deprecatedCreateApplePayBackingEvent = createApplePayBackingData.map(
```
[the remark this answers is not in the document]
justinswart: That's correct, we should be able to remove this completely once that work is done.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: That's the `SetupIntent` object. I don't believe we need it for anything.

Library/ViewModels/PledgeViewModel.swift
justinswart: I'll try to write a failing test for it and then prove that scenario 👍

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    let deprecatedCreateApplePayBackingCompleted = Signal.combineLatest(
+      deprecatedCreateApplePayBackingEvent.filter { $0.isTerminating }.ignoreValues(),
+      paymentAuthorizationDidFinish
+    )
 
-    let submitButtonTapped = Signal.merge(
-      self.submitButtonTappedSignal.mapConst(true),
-      willUpdateApplePayBacking.mapConst(false)
+    let deprecatedCreateApplePayBackingCompletedError = deprecatedCreateApplePayBackingCompleted
+      .withLatest(from: deprecatedCreateApplePayBackingEvent.errors())
```
Scollaco: For curiosity, why are we using `withLatest(from:)` and not `zip` here? Shouldn't `deprecatedCreateApplePayBackingCompletedError` only emit when both `deprecatedCreateApplePayBackingCompleted` and `deprecatedCreateApplePayBackingEvent.errors()` emit?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
     )
 
-    let updateBackingDidComplete = submitButtonTapped
-      .takeWhen(updateBackingEvent.values())
-      .filter(isTrue)
-      .ignoreValues()
+    self.goToThanks = project.takeWhen(createBackingCompletionEvents)
 
-    let updateBackingEventErrors = updateBackingEvent.errors()
+    let createUpdateBackingEventErrors = Signal.merge(
```
Scollaco: Could you rename this to `createUpdateBackingEventErrors` to match the `createOrUpdateBacking...` convention adopted above?

Library/ViewModels/PledgeViewModel.swift
Scollaco: I think that's because `self.showErrorBannerWithMessage` merges different types of errors (`GraphError` and `Error`). What if here we simplify to:
```
let createUpdateBackingEventErrors = Signal.merge(
      deprecatedCreateApplePayBackingEvent.errors(),
      createBackingEvent.errors(),
      updateBackingEvent.errors()
    ).map { $0 as Error }
```

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      deprecatedCreateApplePayBackingEvent.errors().map { $0 as Error },
+      createBackingEvent.errors().map { $0 as Error },
+      updateBackingEvent.errors().map { $0 as Error }
+    )
 
-    let updateApplePayBackingError = updateApplePayBackingCompleted
-      .withLatest(from: updateBackingEventErrors)
+    let createUpdateApplePayBackingError = createOrUpdateApplePayBackingCompleted
```
Scollaco: Could you rename this variable too?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -449,87 +446,115 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      .withLatest(from: createUpdateBackingEventErrors)
       .map(second)
 
-    let updateBackingError = submitButtonTapped
-      .takePairWhen(updateBackingEventErrors)
+    let createUpdateBackingError = submitButtonTapped
```
Scollaco: Same here

Library/ViewModels/PledgeViewModel.swift
justinswart: Moved the mapping to after the merge. I think I was moving things around and did this to satisfy the compiler at some point.

Once we've removed `deprecatedCreateApplePayBackingCompletedError` this should be a little simpler.

justinswart: Moved this around to better support Graph vs SCA errors 👍

Library/ViewModels/PledgeViewModel.swift
justinswart: Yes this is future proof for when `createBacking` supports Apple Pay.

Library/ViewModels/PledgeViewModel.swift
justinswart: I wanted to avoid `zip` just because it can often cause sneaky problems, but that's possible true for this case.

I'll look into this along with https://github.com/kickstarter/ios-oss/pull/902#discussion_r337591752.

Library/ViewModels/PledgeViewModel.swift
justinswart: That is handled on its own so that it'll be easy to remove later. This future-proofs for when `CreateBacking` supports Apple Pay.

Library/ViewModels/PledgeViewModel.swift
justinswart: Ok @ifbarrera @Scollaco I've taken a new approach which clears the values for the create and update backing events each time a new flow is started. I think this gives us more confidence that we won't have overlapping values between flows being interrupted and new ones starting.

The tests are passing (all several thousand lines of them 😓).

Note right now I don't think the _deprecatedCreateApplePayBacking_ needs this treatment because it's somewhat separate, but I didn't want to spend too much time on that given that we should remove it in the next PR.

Library/ViewModels/PledgeViewModel.swift
justinswart: I don't think this is going to be necessary - https://github.com/kickstarter/ios-oss/pull/902#discussion_r337832977.

Also we did `withLatest(from:)` here because the errors emit but we have a race condition between the Apple Pay sheet dismissing and the events emitting, which is why wait for `deprecatedCreateApplePayBackingCompleted`, which is a combined signal to emit and then take the most recently emitted errors.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1388,41 +1414,37 @@ final class PledgeViewModelTests: TestCase {
...
-      [false, false, false, false, true, false, true], "Shipping rule changed"
-    )
+    self.submitButtonEnabled.assertValues([false, true, false, true], "Shipping rule changed")
 
     self.vm.inputs.shippingRuleSelected(.init(cost: 1, id: 1, location: .brooklyn))
 
-    self.submitButtonEnabled.assertValues(
-      [false, false, false, false, true, false, true, false], "Amount and shipping rule unchanged"
-    )
+    self.submitButtonEnabled.assertValues([false, true, false, true, false], "Shipping rule changed")
```
ifbarrera: Is the comment a little off here?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1934,58 +1958,1317 @@ final class PledgeViewModelTests: TestCase {
...
       self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
-      self.updatePledgeFailedWithError.assertValues([
-        "Something went wrong."
+      self.showErrorBannerWithMessage.assertValues(["Something went wrong."])
+      self.submitButtonEnabled.assertValues([false, true])
+      self.goToThanks.assertDidNotEmitValue()
+    }
+  }
+
+  func testApplePayBackingFails_ThenSucceeds_SignalsDoNotOverlap() {
```
ifbarrera: Looks like this is testing with the `.update` context. Can we update (see what I did there) the name of this test to reflect that?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1934,58 +1958,1317 @@ final class PledgeViewModelTests: TestCase {
...
+        "Got it! Your changes have been saved."
       ])
+      self.showErrorBannerWithMessage.assertValues(["Something went wrong."])
       self.submitButtonEnabled.assertValues([false, true])
-      self.createBackingError.assertDidNotEmitValue()
       self.goToThanks.assertDidNotEmitValue()
     }
   }
+
+  func testApplePayBackingFails_ThenStoredCardSucceeds_SignalsDoNotOverlap() {
```
ifbarrera: Same here the context seems to be `.update`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -459,90 +457,139 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       updateButtonIsLoading
     )
 
-    // MARK: - Success/Failure Create
+    let submitButtonTapped = Signal.merge(
```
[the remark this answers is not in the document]
justinswart: Good call!

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -459,90 +457,139 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      .map(second)
 
-    let submitButtonTapped = Signal.merge(
-      self.submitButtonTappedSignal.mapConst(true),
-      willUpdateApplePayBacking.mapConst(false)
+    let createBackingCompletionEvents = Signal.merge(
+      deprecatedCreateApplePayBackingEvent.values()
+        .combineLatest(with: deprecatedCreateApplePayBackingCompleted)
+        .ignoreValues(),
+      didCompleteApplePayBacking.combineLatest(with: willCreateApplePayBacking).ignoreValues(),
```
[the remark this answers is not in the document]
justinswart: Yeah naming is hard. In this particular case the combined signal is to differentiate between whether this updated or created an Apple Pay Backing 🤕 .

PR #904, opened 2019-10-18
[NT-450] Deprecate iOS 11

Scollaco:
# 📲 What 
- Removes support for iOS 11.
- Sets our minimum deployment target to iOS 12.0.
- Replaced some deprecated methods

🤔 Why
- We no longer have enough iOS 11 in order to keep supporting this version.

🛠 How
- Setting minimum deployment target to iOS 12.0
- Removed #available(iOS 12.0, *)
- Updated some functions that were deprecated on iOS 12
 
✅ Acceptance criteria
- [ ] App doesn't build in iOS 11 devices
- [ ] Everything works as expected, specially our `koalaTrackingClient`, since the `save()` and `load()` functions were using code that are now deprecated.

# 🤔 Why
- To keep us moving forward 😄 .


Library/SharedFunctions.swift
```diff
@@ -157,10 +157,7 @@ internal func is1PasswordButtonHidden(_ isHidden: Bool) -> Bool {
 }
 
 public func ksr_is1PasswordSupported() -> Bool {
-  if #available(iOS 12.0, *) {
-    return false
-  }
-  return true
+  return false
```
dusi: Can probably nuke 1Password extension from the app completely 🤔

Library/TestHelpers/TestCase.swift
```diff
@@ -78,7 +78,7 @@ internal func preferredSimulatorCheck() {
   let supportedModels = ["iPhone10,1", "iPhone10,4"] // iPhone 8
   let modelKey = "SIMULATOR_MODEL_IDENTIFIER"
 
-  guard #available(iOS 12.0, *), supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
+  guard supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
```
dusi: I think this will now allow both iOS 12 & 13 (better be explicit)

dusi:
💤 😄

PR #899, opened 2019-10-17
[NT-361] Select new reward polish

Scollaco:
# 📲 What 
- Changes the title of the `RewardsCollectionViewController` based on the RewardsCollectionViewContext.
- Centralizes horizontally the backed reward card when presenting screen from ManagePledgeViewController.

# 🤔 Why

To match the behavior and design shown [here](https://app.goabstract.com/projects/a94cda80-1e60-11e8-8903-0b9b4e24ea4b/branches/11b62619-b4c6-44c0-be45-e000aee1dbaa/collections/38089804-badd-474b-9658-001ea32d3df3?collectionLayerId=225e4c09-0c6c-41e3-b018-908d92721270&mode=design).

# 🛠 How
- Creating `RewardsCollectionViewContext` enum. I thought of reusing `RewardCardViewContext`, but while I was developing this, I noticed that the code was getting harder to understand.
- Adding output that emits title based of the context
- Adding output that emits IndexPath of backed reward. This will allow the collection view to centralize the reward when presenting the carousel from the ManagePledge screen.
- Adding and fixing tests

# 👀 See

| Backing Project | Managing Pledge |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-10-03 at 15 46 44 |  |

# ✅ Acceptance criteria

- [ ] Select a `non-backed` project and open the Rewards Carousel. The `No reward` should be centralized and the title should show `Back this project`
- [ ] Select a `backed` and `live` project and open the Rewards Carousel by selecting `Choose another reward` from the update pledge menu. The collection view should centralize the backed reward and the title should show `Choose another reward`

PR #905, opened 2019-10-22
[NT-425] Manage Pledge Refresh

ifbarrera:
# 📲 What

Refreshes the manage pledge screen when a user:

- updates their pledge amount
- changes their payment method
- selects a new reward

# 🤔 Why

So that the updated pledge is reflected in the manage pledge screen when a change has been made.

# 🛠 How

Initially I was thinking the best approach would be to just refresh the backing when any change is updated. But after discussing with the Android team, it was brought up that it's nice to refresh the whole project in case rewards have changed. I decided to go with that approach based on that point.

# 👀 See

pSIe2lFmlU

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] Test update pledge: navigate to a backing that was created in the same environment that you are testing on. Then, navigate to "Update pledge" from the Manage Pledge screen. Change your pledge amount and tap "Confirm". **You should be taken back to the Manage Pledge Screen and the amount should be updated.**
- [x] Test update payment method: navigate to a backing that was created in the same environment that you are testing on. Then, navigate to "Change Payment Method" from the Manage Pledge screen. Change your payment method and tap "Confirm". **You should be taken back to the Manage Pledge Screen and the payment source should be updated.**
- [ ] Test choose a new reward: navigate to a backing that was created in the same environment that you are testing on. Then, navigate to "Choose a new reward" from the Manage Pledge screen. Change your selected reward tap "Confirm". **You should be taken back to the Manage Pledge Screen and the reward should be updated.**


Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -93,13 +93,6 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
     [self.pledgeSummarySectionSeparator, self.paymentMethodSectionSeparator, self.rewardSectionSeparator]
   }()
 
-  static func instantiate(with project: Project, reward: Reward) -> ManagePledgeViewController {
```
ifbarrera: This `instantiate` function was not being used.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -245,8 +238,8 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
 
   // MARK: - Configuration
 
-  func configureWith(project: Project, reward: Reward) {
-    self.viewModel.inputs.configureWith(project, reward: reward)
+  func configureWith(project: Project) {
```
ifbarrera: We can actually grab the `reward` directly from the `backing` and the project.

ifbarrera: This centralizes how we retrieve the reward when the project refreshes (the reward is now always taken from the `backing`)

Library/SharedFunctions.swift
```diff
@@ -21,6 +21,21 @@ internal func userIsBacking(reward: Reward, inProject project: Project) -> Bool
...
+/**
+ Returns a reward from a backing in a given project
+
+ - parameter backing: A backing
+ - parameter project: A project
+
+ - returns: A reward
+ */
+
+internal func reward(from backing: Backing?, inProject project: Project) -> Reward {
```
justinswart: Ah nice 👍 

I suppose this could also be a computed `var` on `Project` like `backedReward: Reward?`.

That said - if nothing is backed, it'll return `Reward.noReward` which doesn't seem correct, so I think the return type should be optional for this?

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -99,13 +99,7 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
 
     let goToManagePledge = ctaButtonTapped
       .filter { canShowManageViewPledgeScreen($0.0, state: $0.2) }
```
justinswart: I think this (and other) signals here would read better with something like:

```swift
.map { project, _, state in (project, state) }
.filter(canShowManageViewPledgeScreen)
```

Library/SharedFunctions.swift
ifbarrera: I copied and pasted this straight out of `ProjectPamphletViewModel`, but can update. I think ideally it would take a non-optional `Backing` and return a non-optional `Reward`.

ifbarrera: Also just realized I forgot to add tests for this 😅

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -46,7 +46,7 @@ public protocol ProjectPamphletViewModelOutputs {
 
   var goToDeprecatedViewBacking: Signal<BackingData, Never> { get }
 
-  var goToManageViewPledge: Signal<PledgeData, Never> { get }
+  var goToManageViewPledge: Signal<Project, Never> { get }
```
justinswart: This output and related TestObservers really should just be `goToManagePledge`, I think we fixed this naming elsewhere but forgot to update it here 😬

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -241,18 +261,24 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
 
   func testRewardReceivedViewControllerIsHidden_NoReward_Canceled() {
     let backing = Backing.template
       |> Backing.lens.status .~ .canceled
-
-    let project = Project.template
+      |> Backing.lens.reward .~ Reward.noReward
+
+    let project = Project.cosmicSurgery
+      |> Project.lens.rewards %~ { rewards in
```
justinswart: If we just want to switch out the reward at index `0` we could write this here and in other places as:

```swift
|> Project.lens.rewards .~ ([Reward.noReward] + Project.cosmicSurgery.rewards.suffix(from: 1))
```

Library/ViewModels/ProjectPamphletViewModel.swift
ifbarrera: I refactored all this 😅

ifbarrera:
> Nice! This appears to work great. Approved with some minor suggestions.
> 
> Also, I was unable to test the third AC because that is something I'm currently fixing separately and that depends on work being done in #899.
> 
> Oh, also - of course this doesn't handle errors from the refreshing. I _think_ that's fine for this? One would have to lose connectivity immediately after saving changes which is kind've an edge case. However if something did go wrong with the refreshing endpoint we wouldn't know about it.
> 
> Let me know if you think we should add error messaging for the refresh and I can re-review.

Yeah this was a conscious decision – I figured that since we already show the `"Got, it your pledge has been updated"` message when the update request succeeds, it would be really confusing as a user to then see something like `"Failed to get project"` or `"Something went wrong"` because there's no context for the user around what actually failed. I felt it was best to fail silently, but open to discussing it. Fwiw Android also fails silently on refreshing the project.

Library/ViewModels/ProjectPamphletViewModel.swift
ifbarrera: I can update it 👍

PR #907, opened 2019-10-23
[NT-449] Update pledge disclaimer

cdolm92:
# 📲 What

We are now showing a disclaimer when updating pledge in different currency.
# 🤔 Why

So that users understand that they are about to be charged in a foreign currency and foreign transaction fees may apply. 

# 🛠 How

Checking to see if the user's currency matches the project's. If not we show the conversion label otherwise we don't.
# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone Xs - 2019-10-23 at 11 11 01|Simulator Screen Shot - iPhone Xs - 2019-10-23 at 11 08 57|

# ✅ Acceptance criteria

- [ ] Go to a project w/ a currency different than the preferred, navigate to update pledge, you should see the conversion label


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -141,13 +141,15 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     }
     .skipNil()
 
-    self.confirmationLabelAttributedText = Signal.merge(
+    let initialProjectAndProject = Signal.merge(
```
Scollaco: Here the `initialProjectAndProject` are the same project, but emitted in different times and this can end up becoming misleading. Maybe something like this would be better?
```
 let projectAndPledgeTotal = Signal.merge(
      project,
      project.takeWhen(self.traitCollectionDidChangeSignal)
    )
    .combineLatest(with: pledgeTotal)

self.confirmationLabelAttributedText = projectAndPledgeTotal
    .ksr_debounce(.milliseconds(10), on: AppEnvironment.current.scheduler)
    .map(attributedConfirmationString(with:pledgeTotal:))
    .skipNil()
```

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -665,38 +667,44 @@ private func requiresSCA(_ envelope: StripeSCARequiring) -> Bool {
...
 
-  guard let attributedString = try? NSMutableAttributedString(
-    data: Data(string.utf8),
-    options: [
-      .documentType: NSAttributedString.DocumentType.html,
-      .characterEncoding: String.Encoding.utf8.rawValue
-    ],
-    documentAttributes: nil
-  ) else { return nil }
+  if project.stats.currentCurrency == project.stats.currency {
```
Scollaco: To make this function a bit smaller, we could extract this to a new function that returns a String based on the currencies.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -665,38 +667,44 @@ private func requiresSCA(_ envelope: StripeSCARequiring) -> Bool {
...
+    )
+  } else {
+    fullString =
+      Strings.If_the_project_reaches_its_funding_goal_you_will_be_charged_total_on_project_deadline(
+        total: pledgeTotal,
+        project_deadline: date
+      )
+  }
+
+  let attributedString: NSMutableAttributedString = NSMutableAttributedString.init(string: fullString)
```
Scollaco: For consistency, the syntax without `.init` would be preferred `(NSMutableAttributedString(string: fullString))`

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -293,6 +295,55 @@ final class PledgeViewModelTests: TestCase {
     }
   }
 
+  func testUpdateContext_ConfirmationLabelShowsTotalAmount() {
+    let dateComponents = DateComponents()
+      |> \.month .~ 11
+      |> \.day .~ 1
+      |> \.year .~ 2_019
+      |> \.timeZone .~ TimeZone.init(secondsFromGMT: 0)
```
Scollaco: I know this code was already here, but could you get rid of the `.init`s here too?

Library/ViewModels/PledgeViewModel.swift
cdolm92: Oh, I forgot to change that 🤦🏾‍♀️. Thanks for catching!

PR #908, opened 2019-10-24
[NT-241]  Decimal Pledge Amount  in Manage Pledge CTA Fix

cdolm92:
# 📲 What

Users can now view their pledge amount correctly if they pledged w/ a decimal value.

# 🤔 Why

# 🛠 How

`func formattedPledge(amount: Double, project: Project) -> String`  formats the backing amount. If the user backed with a whole amount we drop the decimal places.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|Simulator Screen Shot - iPhone Xs - 2019-10-24 at 15 32 10|Simulator Screen Shot - iPhone Xs - 2019-10-24 at 15 28 33|

# ✅ Acceptance criteria

- [ ] Navigate to a live and pledged project. Make sure that the pledge amount is formatted correctly. You should see a decimal value if you pledge a decimal amount or a whole number if you did not.

### DIFF
--- a/Library/ViewModels/PledgeCTAContainerViewViewModel.swift
+++ b/Library/ViewModels/PledgeCTAContainerViewViewModel.swift
@@ -162,7 +162,9 @@ private func formattedPledge(amount: Double, project: Project) -> String {
   let numberOfDecimalPlaces = amount.truncatingRemainder(dividingBy: 1) == 0 ? 0 : 2
   let formattedAmount = String(format: "%.\(numberOfDecimalPlaces)f", amount)
 
-  return Format.formattedCurrency(formattedAmount,
-                                  country: project.country,
-                                  omitCurrencyCode: project.stats.omitUSCurrencyCode)
+  return Format.formattedCurrency(
+    formattedAmount,
+    country: project.country,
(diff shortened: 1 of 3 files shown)


Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -160,3 +157,14 @@ private func subtitle(project: Project, pledgeState: PledgeStateCTAType) -> Stri
   guard let rewardTitle = reward.title else { return "\(amount)" }
   return "\(amount) • \(rewardTitle)"
 }
+
+private func formattedPledge(amount: Double, project: Project) -> String {
```
justinswart: Hey @cdolm92, was there a reason for not using an `NSNumberFormatter` here?

cdolm92: @justinswart Are we using it to format decimal values elsewhere?

cdolm92: ✅ Convo resolved on slack. The formatter will be added in separate pr.

PR #909, opened 2019-10-24
[NT-472] Fix payment method selection from backing

justinswart:
# 📲 What

- Addresses an issue in which the payment method used on a backing is not selected by default when managing your pledge.
- Slight refactor.
- Adds tests.

# 🤔 Why

Currently this has the appearance to the user of a payment method change not being stored. Navigating back to to the view after changing one's payment method may not have that method selected initially and this also breaks the validation for enabling/disabling the submit button.

# 🛠 How

- Refactored things slightly as there seemed to be a redundant `typealias` that I figured we didn't need.
- Removed the workarounds for comparing payment method IDs now that we are de-serializing this correctly.
- Added tests for the initial payment method selection.

# ✅ Acceptance criteria

With an account that has multiple payment methods:

- [ ] Back a project and navigate to _change payment method_ in manage pledge. The payment method that you used should be selected and the submit button disabled.
- [ ] Selecting another payment method should enable the submit button.
- [ ] Adding a new payment method should still order that new card first in the list and select it.


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -5,10 +5,6 @@ import ReactiveSwift
 import UIKit
 
 public typealias PledgePaymentMethodsValue = (user: User, project: Project, applePayCapable: Bool)
-public typealias CardViewValues = (
```
justinswart: This type seemed very similar to `PledgeCreditCardValue`, removed this one and renamed that to `PledgeCreditCardViewData`.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -60,25 +57,23 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
-    let cardValues = Signal.combineLatest(
-      cardsAndAvailable,
-      project.map { $0.location.displayableName }
-    ).map { CardViewValues(cardAndIsAvailableCardType: $0, projectCountry: $1) }
-
-    self.reloadPaymentMethods = cardValues
+    self.reloadPaymentMethodsAndSelectCard = Signal.combineLatest(cards, availableCardTypes, project)
+      .map(pledgeCreditCardViewData(with:availableCardTypes:project:))
+      .map { pledgeCreditCardViewData in
+        (pledgeCreditCardViewData, pledgeCreditCardViewData.first?.card)
```
justinswart: Selects the first card by default, optional in case there are no cards stored.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -132,28 +127,37 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
 private func showApplePayButton(for project: Project, applePayCapable: Bool) -> Bool {
   return applePayCapable &&
     AppEnvironment.current.config?.applePayCountries.contains(project.country.countryCode) ?? false
 }
+
+private func cards(
+  _ cards: [GraphUserCreditCard.CreditCard],
+  orderedBy backing: Backing?
+) -> [GraphUserCreditCard.CreditCard] {
+  return cards.sorted { card1, _ in card1.id == backing?.paymentSource?.id }
```
justinswart: Simple sort that places the card first if it's ID matches that of the payment source on the `Backing`.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -96,12 +96,12 @@ final class PledgePaymentMethodsViewController: UIViewController {
   override func bindViewModel() {
     super.bindViewModel()
 
-    self.viewModel.outputs.reloadPaymentMethods
+    self.viewModel.outputs.reloadPaymentMethodsAndSelectCard
       .observeForUI()
-      .observeValues { [weak self] cardValues in
+      .observeValues { [weak self] cardValues, selectedCard in
```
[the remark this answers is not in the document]
ifbarrera: Although maybe that would complicate the flow of selecting a card 🤔

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -106,7 +121,83 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+        card1, _ in card1 == GraphUserCreditCard.visa
+      }
+
+      self.reloadPaymentMethodsCards.assertValues([visaFirstCards])
+      self.reloadPaymentMethodsAvailableCardTypes.assertValues([
+        [true, true, true, true, true, true, false, true]
+      ])
+      self.reloadPaymentMethodsProjectCountry.assertValues([
+        (0...response.me.storedCards.nodes.count - 1).map { _ in "Hastings, UK" }
+      ], "One card is unavailable")
```
ifbarrera: Is this message meant to go in the assertion above on line 173?

justinswart: I added this mostly to indicate why I'm doing the range as `0...response.me.storedCards.nodes.count - 1`, the existing tests didn't have any messages describing the assertions really. I could add the same message to all of the `reloadPaymentMethodsAvailableCardTypes` but I think it's pretty clear what's going on?

justinswart:
> The Confirm button is correctly enabled, because the card is different from the card that was used the pledge. My only concern is that it might be confusing for users because it makes it seem like the card that was used to pledge was the first one, when in fact it was a different card.

Yeah this is an unfortunate side-effect of us always defaulting to the first card being "selected". Only way I think we could fix this would be to not select a default when changing payment method _if_ that payment method no longer exists. I'll create a follow-up ticket, I think it'll need a bit more consideration.

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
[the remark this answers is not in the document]
justinswart: As per Slack convo - let's revisit this in future.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
ifbarrera: Ah gotcha!

PR #910, opened 2019-10-25
[NT-424] Migrate HockeyApp to AppCenter

justinswart:
# 📲 What

Replaces HockeyApp with Microsoft's new AppCenter SDK.

**Note:** So far I have only migrated our alpha target in order to test this end-to-end. Once we're satisfied we should migrate our beta and production targets.

# 🤔 Why

HockeyApp is [being sunset](https://hockeyapp.net/transition/) on Nov 16 2019. We're required to migrate to AppCenter to continue using the platform for distributing our internal alpha and beta builds.

# 🛠 How

- Removed all references to Hockey in our codebase, secrets, circle.yml, Fastfile.
- Configured AppCenter as closely as possible to how we had been configuring Hockey. However, some configuration properties that we had before don't seem to be available on AppCenter.
- Added the AppCenter Fastlane plugin for distributing alpha and beta builds via CircleCI.
- Added api token to secrets: https://github.com/kickstarter/native-secrets/pull/61

# ✅ Acceptance criteria

Distribute an alpha locally via `fastlane` using the following commands:
```bash
agvtool new-version -all $(($(date +%s)/100))
bundle exec fastlane alpha_gym
bundle exec fastlane alpha_appcenter
```
- [ ] This should upload an Alpha build to AppCenter and it should be visible in the dashboard.
- [ ] Push to `alpha-dist-<something>`. This will generate an ad hoc alpha build on CircleCI and distribute that to AppCenter.

# ⏰ TODO

- [ ] Move Beta and Production apps over to AppCenter and generate their API tokens.


.circleci/config.yml
```diff
@@ -44,15 +41,6 @@ jobs:
           keys:
             - fabric-sdk-cache-{{ checksum "fabric_version.txt" }}
 
-      - run:
-          name: Store Hockey SDK Version
```
justinswart: The AppCenter SDK is managed via Carthage, no need to cache this separately.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -144,22 +147,28 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
-        manager.configure(withIdentifier: data.appIdentifier)
-        manager.crashManager.crashManagerStatus = .disabled
-        manager.isUpdateManagerDisabled = data.disableUpdates
-        manager.userID = data.userId
-        manager.userName = data.userName
-        manager.start()
-        manager.authenticator.authenticateInstallation()
+      .observeValues { data in
+        let customProperties = MSCustomProperties()
+        customProperties.setString(data.userName, forKey: "userName")
```
justinswart: Added this because there doesn't appear to be a similar `userName` property like `BITHockeyManager` had before.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -144,22 +147,28 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        MSAppCenter.setCustomProperties(customProperties)
+
+        MSCrashes.setDelegate(self)
+
+        MSAppCenter.start(
+          data.appSecret,
+          withServices: [
+            MSAnalytics.self,
+            MSCrashes.self,
+            MSDistribute.self
```
justinswart: Need to confirm that `MSDistribute` works correctly for us. In the set up [docs](https://docs.microsoft.com/en-us/appcenter/sdk/distribute/ios) it mentions adding the app secret to `Info.plist` which is obviously problematic for us. So I didn't do it but I'm not sure if it will cause issues.

Kickstarter-iOS/Info.plist
```diff
@@ -31,6 +31,7 @@
 		<dict>
 			<key>CFBundleURLSchemes</key>
 			<array>
+				<string>appcenter-$(APPCENTER_$(APPCENTER_SCHEME_NAME)_API_TOKEN)</string>
```
justinswart: @Scollaco @ifbarrera This accesses the environment var for the API token based on an environment var for this configuration.

So for our alpha target the concatenated environment var is `APPCENTER_ALPHA_API_TOKEN` and in this var name `ALPHA` is a variable based on `APPCENTER_SCHEME_NAME` which is a user-defined variable in our build settings.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -6754,11 +6756,12 @@
...
 				VALIDATE_PRODUCT = YES;
 			};
-			name = "Hockey Alpha";
+			name = "AppCenter Alpha";
 		};
-		D083B1041F27EF0400BEC742 /* Hockey Alpha */ = {
+		D083B1041F27EF0400BEC742 /* AppCenter Alpha */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPCENTER_SCHEME_NAME = ALPHA;
```
justinswart: Exhibit A

Kickstarter-iOS/Info.plist
justinswart: [These docs](https://docs.microsoft.com/en-us/appcenter/sdk/distribute/ios#23-modify-the-projects-infoplist) mention that we need to have the secret in the `Info.plist`

justinswart:
@Scollaco this is ready for re-review. Just note, where previously I was calling it an API token per app, it's actually just one API token and rather a secret per app. I've also since removed the production target from being submitted to AppCenter. We will only use it for alpha and beta distribution.

PR #911, opened 2019-10-25
Pledge maximum adding shipping rules

Scollaco:
# 📲 What
- Includes shipping cost on maximum pledge amount.
- Move minimum pledge label to above stepper.
- Add an error label that appears when pledge amount is greater the maximum pledge amount.

# 🤔 Why
- To make it easier for the user to see what's the maximum possible amount to pledge.

[JIRA ticket](https://dripsprint.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=NT-435&assignee=5d41b601f14e7c0da562b03f)

# 🛠 How
- Passing the value of the shipping cost to the `PledgeAmountViewModel` and updating the max pledge by subtracting the currency's max pledge and shipping cost
- Added a new Label to `PledgeAmountViewController` that appears whenever the amount is greater than the new max pledge amount

# 👀 See

max_pledge

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

Project with no Shipping Rules:
- [ ] Select a project with no Shipping Rules and type an amount greater than the maximum pledge amount  (e.g $10000 for US projects). A label should appear with the text `Please enter a pledge amount between [reward min] and [currency max pledge].`
- [ ] The `Done` button should be `disabled.

Project with Shipping Rules:
- [ ] Select a project with Shipping Rules and type an amount equals to the `currency max pledge amount` (e.g $10000 for US projects) minus `shipping Cost`. Then select a new shipping rule with greater shipping cost.  A label should appear with the text `Please enter a pledge amount between [reward min] and [new max]`
- [ ] The `Done` button should be `disabled.


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -68,6 +68,8 @@ public protocol PledgeViewModelOutputs {
   var submitButtonIsLoading: Signal<Bool, Never> { get }
   var submitButtonTitle: Signal<String, Never> { get }
   var title: Signal<String, Never> { get }
+  var updateMaximumPledgeAmount: Signal<Double, Never> { get }
+  var updatePledgeFailedWithError: Signal<String, Never> { get }
```
ifbarrera: Was this output re-added during a merge? I believe it's now removed on master.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -156,6 +160,10 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     let pledgeAmountIsValid = self.pledgeAmountDataSignal
       .map { $0.isValid }
 
+    pledgeAmountIsValid.observeValues { v in
```
ifbarrera: Should we remove this?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -408,6 +400,22 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     )
     .map(shippingRuleValid)
 
+    self.showApplePayAlert = Signal.combineLatest(
+      project,
+      self.pledgeAmountDataSignal
+    )
+    .takeWhen(Signal.merge(showApplePayAlert))
```
ifbarrera: Do we still need this `.merge`?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -575,6 +583,8 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
 
     self.notifyDelegateUpdatePledgeDidSucceedWithMessage = updateBackingCompletionEvents
       .mapConst(Strings.Got_it_your_changes_have_been_saved())
+    self.updatePledgeFailedWithError = updateBackingEvent.errors()
```
ifbarrera: I think this is removed on master.

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -15,6 +15,7 @@ public protocol PledgeAmountViewModelInputs {
   func stepperValueChanged(_ value: Double)
   func textFieldDidEndEditing(_ value: String?)
   func textFieldValueChanged(_ value: String?)
+  func updateMaximumPledgeAmount(with shippingCost: Double)
```
justinswart: The input should be named based on the event that occurred, so in this case I think it's `selectedShippingAmountChanged(to amount: Double)`.

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -59,8 +62,17 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
...
 
+    let shippingCost = Signal.merge(
+      self.projectAndRewardProperty.signal.mapConst(0),
+      self.shippingCostProperty.signal
+    )
+
     let maxValue = minAndMax
       .map(second)
+      .combineLatest(with: shippingCost)
+      .map { max, shippingCost in
```
justinswart: Swift is so fancy this can actually be `.map(-)` 😄

Library/ViewModels/PledgeViewModel.swift
justinswart: lol who dis?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -706,6 +716,8 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
   public let submitButtonIsLoading: Signal<Bool, Never>
   public let submitButtonTitle: Signal<String, Never>
   public let title: Signal<String, Never>
+  public let updateMaximumPledgeAmount: Signal<Double, Never>
+  public let updatePledgeFailedWithError: Signal<String, Never>
```
justinswart: ☝️

Library/ViewModels/PledgeViewModel.swift
Scollaco: Used for debug purposes and missed it during cleanup. 🤦‍♂

Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -156,6 +163,10 @@ final class PledgeAmountViewController: UIViewController {
   @objc func textFieldDidChange(_ textField: UITextField) {
     self.viewModel.inputs.textFieldValueChanged(textField.text)
   }
+
+  func selectedShippingAmountChanged(to amount: Double) {
```
justinswart: Can we put this under `// MARK: - Accessors`

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -117,6 +118,8 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
 
     let pledgeTotal = Signal.combineLatest(pledgeAmount, shippingCost).map(+)
 
+    self.selectedShippingAmountChanged = shippingCost
```
justinswart: So typically we'd want this sort of information to be sent via a delegate but in this case our `PledgeViewController` will have two delegates 🤔. I think this is a bit of an anti-pattern and points to a larger architectural problem that we have with our VCs talking directly to each other. I have thoughts on how we can improve this but it's outside the scope of this work.

In any case, I would suggest at least renaming this output to make what this is doing a bit clearer - something like `notifyPledgeAmountViewControllerShippingAmountChanged`.

We're likely going to want to do some refactoring here down the line but I don't want to hold this up to do that right now.

Please make this change before merging though.

Scollaco: Yes, I agree with that and I think it will be a fun problem to solve. A discussion in this topic with the team would be worth it. But for now, thanks for you input 😄

PR #913, opened 2019-10-25
[NT-411, NT-429] Show total amount currency conversion label

ifbarrera:
📲 What

Shows a currency conversion label beneath the `Total` amount in the `PledgeSummaryViewController` if the user's preferred currency is different from the project's currency. If the user does not have a preferred currency set, we use `USD` to determine whether a conversion label is needed. The conversion label will read `About $10.50` and display the converted amount to 2 precision points.

I also removed the `+` icon from in front of the total amount as part of NT-429.

# 🤔 Why

To give users a better idea of how much they are pledging, if the project's currency is different from their preferred currency.

# 🛠 How

- Added `roundingMode`, `maximumFractionDigits` and `minimumFractionDigits` parameters to our `Format.currency` function so we can customize these parameters when we format the converted amount.
- added a new `totalConversionLabel` to `PledgeSummaryViewController`

# 👀 See

| Pledge | Update | Logged Out |
| --- | --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-10-25 at 10 26 48 | Simulator Screen Shot - iPhone 8 - 2019-10-25 at 10 28 49 | Simulator Screen Shot - iPhone 8 - 2019-10-25 at 16 56 25 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [ ] Navigate to a project with the same currency as your preferred currency. Select any reward. You should **not** see a conversion label on the pledge screen.
- [ ] Navigate to a project with a different currency than your preferred currency. Select any reward. You should **see a conversion label** below the total on the pledge screen.
- [ ] Log out of your account. Then, navigate to a project with a currency that is not USD. Select any reward. You should **see a conversion label** below the total on the pledge screen.


Library/ViewModels/PledgeSummaryViewModel.swift
```diff
@@ -1,13 +1,18 @@
 import Foundation
+import KsApi
 import Prelude
 import ReactiveSwift
 
 public protocol PledgeSummaryViewModelInputs {
   func tapped(_ url: URL)
+  func configureWith(_ project: Project, total: Double)
+  func viewDidLoad()
```
cdolm92: alphabetize

PR #914, opened 2019-10-25
[NT-462] Return to Manage Pledge after selecting new reward

justinswart:
# 📲 What

Returns the user to Manage Pledge after changing their reward.

# 🤔 Why

Currently they found themselves back on the rewards carousel which is the incorrect behaviour.

# 🛠 How

Quickest fix for this was to set the `PledgeViewController` delegate when presenting it from `RewardsCollectionViewController`. This isn't really ideal but it basically looks like:

`ManagePledgeViewController` -> `RewardsCollectionViewController.pledgeViewDelegate` -> `PledgeViewController.delegate`.

So, upon an update completion, `PledgeViewController` informs `ManagePledgeViewController` so that it refreshes.

`PledgeViewController` then pops to the root VC which works for all of our update flows.

**Note:** This change also meant that we needed to push `RewardsCollectionViewController` on the navigation stack when changing a reward rather than presenting it modally. This allows us to pop back to the `ManagePledgeViewController` after changing the reward.

# 👀 See

# ✅ Acceptance criteria

- [x] Changing a reward should return to the Manage Pledge VC on completion.
- [x] All other pledge update flows should work as before.


Kickstarter-iOS/Views/Controllers/ProjectPamphletViewController.swift
```diff
@@ -343,6 +343,19 @@ private func rewardsCollectionViewController(
   let rewardsCollectionViewController = RewardsCollectionViewController
     .instantiate(with: project, refTag: refTag, context: .createPledge)
 
+  let closeButton = UIBarButtonItem(
```
justinswart: The closeButton is set at the point of presentation now so that we don't see it when pushing onto the navigation controller stack.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -6,8 +6,15 @@ import Prelude
 final class RewardsCollectionViewController: UICollectionViewController {
   // MARK: - Properties
 
+  private var collectionViewBottomConstraintSuperview: NSLayoutConstraint?
```
justinswart: Just alphabetized some properties here.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -50,6 +50,7 @@ final class PledgeViewModelTests: TestCase {
   private let paymentMethodsViewHidden = TestObserver<Bool, Never>()
   private let pledgeAmountViewHidden = TestObserver<Bool, Never>()
   private let pledgeAmountSummaryViewHidden = TestObserver<Bool, Never>()
+  private let popToRootViewController = TestObserver<(), Never>()
```
justinswart: Seems we weren't testing `popViewController` before so added tests for it now as `popToRootViewController`.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -340,13 +340,15 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
...
+    let vc = RewardsCollectionViewController.instantiate(
       with: project,
       refTag: nil,
       context: .managePledge
     )
 
-    self.navigationController?.pushViewController(rewardsVC, animated: true)
+    vc.pledgeViewDelegate = self
+
+    self.show(vc, sender: nil)
```
ifbarrera: Any reason to use `show` instead of our regular `pushVC`? From the docs it seems that the `show` function is only really useful if you're using something like `UISplitViewController`, and elsewhere in the app we tend to use `push`. 🤔

> You use this method to decouple the need to display a view controller from the process of actually presenting that view controller onscreen. Using this method, a view controller does not need to know whether it is embedded inside a navigation controller or split-view controller. It calls the same method for both.

 Seems to me like it might be breaking with what we're using elsewhere without a clear benefit, but I might be missing something.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1787,6 +1811,8 @@ final class PledgeViewModelTests: TestCase {
       self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertValues([
         "Got it! Your changes have been saved."
       ])
+      self.popToRootViewController.assertValueCount(1)
+      self.popToRootViewController.assertValueCount(1)
```
ifbarrera: Should this line be duplicated?

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
justinswart: I noticed it was used in a couple other places in this VC but yeah it seems it's only really useful if we are using `UISplitViewController` to abstract away which control the VC is embedded in. I'll revert to push here.

Library/ViewModels/PledgeViewModelTests.swift
justinswart: Nope, thanks!

PR #917, opened 2019-10-28
[NT-484] Tracking event for Cancel Pledge

cdolm92:
# 📲 What
Improvements in monitoring user behavior for v1 checkout. This tracks canceled pledges.

# 🤔 Why

We now have the `Cancel Pledge Button Clicked` event

# ✅ Acceptance criteria

- [ ] Navigate to a live and backed project. Go to the manage menu and tap `Cancel Pledge`.  Tap on `Yes, Cancel it` button and you should see the event triggered.

### DIFF
--- a/Library/ViewModels/CancelPledgeViewModelTests.swift
+++ b/Library/ViewModels/CancelPledgeViewModelTests.swift
@@ -222,4 +222,19 @@ final class CancelPledgeViewModelTests: TestCase {
       self.cancelPledgeError.assertValues(["You can't cancel your pledge right now."])
     }
   }
+
+  func testTrackingEvents() {
+    let mockService = MockService(cancelBackingResult: .init(success: .init()))
+
+    withEnvironment(apiService: mockService) {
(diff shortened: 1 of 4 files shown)


Library/ViewModels/CancelPledgeViewModelTests.swift
```diff
@@ -222,4 +222,19 @@ final class CancelPledgeViewModelTests: TestCase {
...
   }
+
+  func testTrackingEvents() {
+    let mockService = MockService(cancelBackingResult: .init(success: .init()))
+
+    withEnvironment(apiService: mockService) {
+      self.vm.inputs.configure(with: .template, backing: .template)
+      self.vm.inputs.viewDidLoad()
+
+      self.cancelPledgeButtonEnabled.assertValues([true])
```
Scollaco: The cancelPledgeButton status is already being tested in other place and it's not so useful here since you're only testing the tracking. Instead, I would replace this line with `XCTAssertEqual([], self.trackingClient.events)` to prove that the event is only tracked after the button is tapped.

PR #918, opened 2019-10-28
[NT-463] Improve error handling for graph requests

ifbarrera:
# 📲 What

This PR makes some changes to how we make Graph requests, and how we parse responses from Graph requests.

# 🤔 Why

To improve error parsing/handling for graph requests and make our requests more robust to changes in the JSON structure coming back from the API. Also, to make graph requests more consistent with how our `v1` requests are made.

# 🛠 How

Previously, we made all graph requests through a single `performRequest` function that started a data task within a producer and parsed errors and data in the completion block of that task. This approach was a little inconsistent with our approach for `v1` requests, which splits up error handling and model decoding by running all error handling through `rac_*` helper functions defined in an extension of `URLSession`. These helpers also make use of Reactive Swift's `data(with request:)` function which is specially designed to return a `SignalProducer` which `"performs work associated with an NSURLSession"`. It seemed like a good idea for Graph requests to follow a similar pattern of doing error handling in a similar `rac_graphDataResponse` function, and then running the resulting `Data` through a decoding function (`decodeGraphModel`). 

Now, `v1` requests are made like this (no change here):

```
 return Service.session.rac_JSONResponse(preparedRequest(forURL: paginationUrl))
      .flatMap(self.decodeModel)
```

... and Graph requests are made like this:

```
return Service.session.rac_graphDataResponse(request)
        .flatMap(self.decodeGraphModel)
```

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

Setup: Using Charles Proxy, set up a breakpoint such that all requests hitting the Graph endpoint are intercepted.

- [x] Test that aborted requests correctly show a request error.
- [x] Test that removing the "data" field from a request with a validation error (ex. entering the wrong password when trying to change your email) does not affect parsing of the returned error. The message banner should still show the validation error returned from the API.
- [x] Test that errors returned from the API (ex. entering the wrong password when changing your email) are correctly parsed and displayed.
- [x] Test that modifying the response body to an invalid object correctly produces a decoding error


KsApi/GraphSchema.swift
```diff
@@ -3,7 +3,10 @@ import Prelude
 // MARK: - Graph Response
 
 public struct GraphResponse<T: Decodable>: Decodable {
-  let data: T?
+  let data: T
```
ifbarrera: Here we decouple `data` and `errors` by removing the `errors` from `GraphResponse` and giving them their own `GraphResponseErrors` struct. This allows us to decode errors separately from `data`.

KsApi/Service+RequestHelpers.swift
```diff
@@ -43,45 +43,23 @@ extension Service {
       .map { json in decode(json) as M? }
   }
 
-  private func performRequest<A: Swift.Decodable>(request: URLRequest) -> SignalProducer<A, GraphError> {
```
ifbarrera: The `performRequest` function has been removed in favor of a `decodeGraphModel` function that decodes `Data` into objects, and an `rac_graphDataResponse` function which is responsible for actually making the request and parsing any resulting errors.

KsApi/extensions/NSURLSession.swift
```diff
@@ -12,6 +12,65 @@ private func parseJSONData(_ data: Data) -> Any? {
...
+      }
+      .flatMap(.concat) { data, response -> SignalProducer<Data, GraphError> in
+        guard let response = response as? HTTPURLResponse else { fatalError() }
+
+        guard
+          (200..<300).contains(response.statusCode),
+          let headers = response.allHeaderFields as? [String: String],
+          let contentType = headers["Content-Type"], contentType.hasPrefix("application/json")
+        else {
+          print("🔴 [KsApi] HTTP Failure \(self.sanitized(request))")
```
justinswart: Seems to be an extra parenthesis at the end of this string.

KsApi/extensions/NSURLSession.swift
```diff
@@ -12,6 +12,65 @@ private func parseJSONData(_ data: Data) -> Any? {
...
+            error: .jsonDecodingError(
+              responseString: String(data: data, encoding: .utf8),
+              error: nil
+            )
+          )
+
+          return self.decodeGraphErrorsOrCoalesce(from: data, coalesce: producer)
+        }
+
+        print("🔵 [KsApi] HTTP Success \(self.sanitized(request))")
```
justinswart: Here too.

KsApi/extensions/NSURLSession.swift
```diff
@@ -12,6 +12,65 @@ private func parseJSONData(_ data: Data) -> Any? {
...
+
+        let producer = SignalProducer<Data, GraphError>(value: data)
+
+        return self.decodeGraphErrorsOrCoalesce(from: data, coalesce: producer)
+      }
+  }
+
+  private func decodeGraphErrorsOrCoalesce(
```
justinswart: Lol I know I kind've pieced this together when we paired, but I think this would be better as just:

```swift
private func tryDecodeGraphErrors(from data: Data) -> SignalProducer<Data, GraphError> {
  let decodedErrors = try? JSONDecoder().decode(GraphResponseErrors.self, from: data)

  guard let error = decodedErrors?.errors?.first else { return .init(value: data) }

  print("🔴 [KsApi] Graph Error \(error.message)")

  return .init(error: GraphError.decodeError(error))
}
```
^ So if it successfully decodes an error envelope is will error out the stream, else it'll just return the data that it received satisfying the type requirements.

The call sites would then look like:
```swift

return self.tryDecodeGraphErrors(from: data)
  .flatMap { data in
    SignalProducer<Data, GraphError>(
      error: .jsonDecodingError(
        responseString: String(data: data, encoding: .utf8),
        error: nil
      )
    )
  }
```
and:
```swift
return self.tryDecodeGraphErrors(from: data)
  .flatMap { data in
    SignalProducer<Data, GraphError>(value: data)
  }
```
I also don't think we need to use the `.concat` strategy on these inner producers if the encompassing producer is already using it 🤔 .

KsApi/extensions/NSURLSession.swift
ifbarrera: I think actually it has the correct number of parenthesis 🤔

ifbarrera:
> This all checks out! Only thing to mention is that during testing, it correctly produced a decoding error by way of entering that part of the code, but the message is still just "Something went wrong". Was that your intention?
> 
> If so - :shipit:

By "decoding error" are you referring to the error that results if the structure of the data doesn't match our expectation? If so, then yes this is the expected behavior.

justinswart:
> By "decoding error" are you referring to the error that results if the structure of the data doesn't match our expectation? If so, then yes this is the expected behavior.

Yes. The console of course still said that it was a decoding error.

PR #919, opened 2019-10-30
[NT-394] Fix loading button title restoration

justinswart:
# 📲 What

Fixes a small bug which left our button title empty after the loader was hidden.

# 🤔 Why

This was just a small oversight in #892.

# 🛠 How

Fixed what was essentially just a copy-paste error.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] Update your pledge, after loading completes the button title should be set again. Use Charles proxy with breakpoints or slow down animations to confirm.

### DIFF
--- a/Kickstarter-iOS/Views/LoadingButton.swift
+++ b/Kickstarter-iOS/Views/LoadingButton.swift
@@ -115,7 +115,7 @@ final class LoadingButton: UIButton {
     let states: [UIControl.State] = [.disabled, .highlighted, .normal, .selected]
 
     states.compactMap { state -> (String, UIControl.State)? in
-      guard let title = self.title(for: state) else { return nil }
+      guard let title = self.originalTitles[state.rawValue] else { return nil }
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/LoadingButton.swift
```diff
@@ -115,7 +115,7 @@ final class LoadingButton: UIButton {
     let states: [UIControl.State] = [.disabled, .highlighted, .normal, .selected]
 
     states.compactMap { state -> (String, UIControl.State)? in
-      guard let title = self.title(for: state) else { return nil }
```
justinswart: The bug was just that this line was copy-pasted from the block above whereas it should in fact be restoring the title from the array and not `self.title(for:)`.

PR #921, opened 2019-10-31
[NT-502] Fix pledge total & update pledge amount for No Reward backings

ifbarrera:
# 📲 What

Fixes a bug where the "Update pledge amount" and "Change payment method" screens were displaying the incorrect pledge amount and totals on No Reward backings that have pledge amounts that are different from the reward minimum ($1).

# 🤔 Why

Need to fix the 🐛s

# 🛠 How

There was a bug in the function in `PledgeAmountViewModel` that determines whether a user is backing a given reward, because of the strange structure of No Reward objects.

Updating this logic to use our shared function `userIsBacking(reward:inProject)` fixed this issue.

I also added more robust tests.

# 👀 See

**Change Payment Method screen**

| After 🦋 | Before 🐛 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-10-31 at 11 26 49 | Simulator Screen Shot - iPhone 8 - 2019-10-31 at 13 20 13 |

**Update Pledge Amount screen**

| After 🦋 | Before 🐛 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-10-31 at 13 26 02 | Simulator Screen Shot - iPhone 8 - 2019-10-31 at 13 20 21 | 

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] Backing with no shipping
	Change payment method screen:
		- Pledge amount should reflect the backing amount
		- total should reflect the backing amount
	Update pledge amount
		- pledge amount should start at the backing amount

- [x] Backing with shipping
	Change payment method screen:
		- Pledge amount should reflect the backing amount, not including shipping
		- shipping amount should reflect the selected shipping method amount
		- total should reflect the total backing amount including shipping
	Update pledge amount
		- pledge amount should start at the backing amount
		- shipping amount should start at the original shipping amount

- [x] No reward backing
	Change payment method screen:
		- Pledge amount should reflect the backing amount
		- total should reflect the backing amount
	Update pledge amount
		- pledge amount should start at the backing amount

- [x] No reward backing where pledged amount is higher than min
	Change payment method screen:
		- Pledge amount should reflect the backing amount
		- total should reflect the backing amount
	Update pledge amount
		- pledge amount should start at the backing amount


Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -19,7 +21,6 @@ public protocol PledgeAmountViewModelInputs {
 }
 
 public protocol PledgeAmountViewModelOutputs {
-  var amount: Signal<PledgeAmountData, Never> { get }
```
ifbarrera: Renamed this output to something more descriptive.

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -4,6 +4,8 @@ import Prelude
 import ReactiveExtensions
 import ReactiveSwift
 
+public typealias PledgeAmountData = (amount: Double, min: Double, max: Double, isValid: Bool)
```
ifbarrera: This typealias was previously in `PledgeViewModel` for some reason. I thought it made more sense for it to live here.

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -185,11 +187,7 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
       )
     }
 
-    self.stepperValue = Signal.merge(
-      minValue,
-      self.amount.map { $0.0 }
-    )
-    .skipRepeats()
+    self.stepperValue = self.notifyDelegateAmountUpdated.map { $0.0 }.skipRepeats()
```
ifbarrera: There's no real need to `merge` with `minValue` - this is already taken into account in `notifyDelegateAmountUpdated`.

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -261,8 +259,9 @@ private func rounded(_ value: Double) -> Double {
   return round(value * 100) / 100
 }
 
-private func initialPledgeAmount(from backing: Backing?, reward: Reward, minValue: Double) -> Double {
-  guard let backing = backing, reward.id == backing.rewardId else { return minValue }
```
ifbarrera: Here, the bug was essentially that `No Reward`'s `rewardId` in `backing.rewardId` comes back as `nil`, even though the "No reward" `Reward` object has an `id` of 0 ¯\_(ツ)_/¯

Our shared function `userIsBacking(reward)` accounts for this, so using that fixed the issue.

justinswart: Ah very silly, thanks!

PR #924, opened 2019-10-31
[NT-521] Fix No Reward minimum conversion bug

ifbarrera:
# 📲 What

Fixes a bug where if a user pledged to a project with No Reward, they would not see the correct reward minimum conversion value.

# 🤔 Why

🐛 s

# 🛠 How

Our shared function `reward(fromBacking:inProject)` which is used to access the full reward object from a backing's `rewardId`, had a bug where if the backing's reward was No Reward (and consequently had a `rewardId` of `nil`), we should return our local `Reward.noReward` object in that function. This template object should actually never be returned, because it doesn't have an accurate `minimum` or `convertedMinimum` value (these values depend on the project). The way to solve this is actually to search for the `No Reward` reward in the project's list of rewards (by searching for the reward with an `id` of 0). This is *actually* the No Reward `Reward` that we want, with the correct `minimum` and `convertedMinimum` values for the backed project.

So, the fix was to tweak that shared function to search for the No Reward `Reward` in the list of rewards from the project.

# 👀 See

| After 🦋 |  Before 🐛 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-10-31 at 15 06 54 | Simulator Screen Shot - iPhone 8 - 2019-10-31 at 16 39 31 |

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [ ] Back a project that's in a different currency than your preferred currency. Make sure to back it by selecting the "No Reward" option. Navigate to the manage/view pledge screen. You should see the reward minimum conversion label in the "Selected reward" section, and the conversion should not be $0 (it should be the correct conversion value).


Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -115,7 +115,24 @@ internal final class SharedFunctionsTests: XCTestCase {
...
+      |> Project.lens.rewards .~ [
+        noReward, // No Reward "reward" is available
+        Reward.template
+          |> Reward.lens.id .~ 456
+      ]
+
+    XCTAssertEqual(noReward, reward(from: backing, inProject: project))
+  }
+
+  func testRewardFromBackingWithProject_WhenRewardNotPresent_NoRewardNotPresent() {
```
Scollaco: Is it possible to have a project with no `NoReward`? 🤔

Scollaco: According to @ifbarrera , it's probable that there's no project with no `NoReward`, but this test would cover this weird case if it happens.

PR #926, opened 2019-11-01
[NT-320] Improve experience on SheetOverlay for vertically compact environments

ifbarrera:
# 📲 What

This PR makes some improvements to our custom `SheetOverlayViewController` to better support vertically compact environments.

This PR also updates the iPad presentation style of `ManagePledgeViewController` to `.pageSheet` so it's more consistent with how the `PledgeViewController` is presented on iPad.

# 🤔 Why

On landscape mode on some devices, the shipping location selector on the pledge screen was very difficult to use, because it was presented inside the `SheetOverlayViewController` with a very large offset. This created a very frustrating user experience.

# 🛠 How

The root issue is that on landscape mode (in vertically compact environments), the offset on the `SheetOverlayViewController` is too large, even though it works for portrait or vertically regular environments. To get around this, we override `traitCollectionDidChange` on `SheetOverlayViewController` to check if the new trait collection is vertically compact. If it is, we simply update the offset to a more reasonable number for that trait collection: the height of the navigation bar, or a hard-coded similar value in the case where there is no navigation bar. If the user then rotates back to a vertically regular environment, the originally defined `offset` is used.

# 👀 See

| After 🦋 | Before 🐛 |
| --- | --- |
| Simulator Screen Shot - iPhone 5s - 2019-11-01 at 11 13 17 | Screen Shot 2019-10-11 at 11 42 38 AM |

**iPhone X**

Simulator Screen Shot - iPhone X - 2019-11-01 at 11 22 20

**iPad**

Simulator Screen Shot - iPad Pro (12 9-inch) - 2019-11-01 at 11 27 50

**Manage Pledge presented using Page Sheet on iPad**

Simulator Screen Shot - iPad Pro (12 9-inch) - 2019-11-01 at 11 31 30

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] Navigate to the pledge screen and tap the shipping location button to display a list of shipping locations. Then, transition to landscape mode. You should see a smaller offset in landscape mode, and returning to portrait mode should show the original offset.


cdolm92:
ACs met, Tested on: iPhone XS simulator

PR #930, opened 2019-11-04
[NT-505] Project page creators header

Scollaco:
# 📲 What 
- Adds a header to allow the creator to easily see dashboard/progress of the project she/he created.
**Note:** This PR doesn't handle the case when contributors visit the project page. This will be done in the future.

# 🤔 Why
This task is the first part of a series of tickets that will prevent creators of backing their own project. 
JIRA ticket

# 🛠 How
- Created `ProjectPamphletCreatorHeaderCell` and its viewModel.
- Created `ProjectPamphletCreatorHeaderCellDelegate` to notify the `ProjectPamphletContentViewController` when the user taps the `View progress/dashboard` button.
- The top margin on the `Video` storyboard was a little off and it was preventing the cell to be presented correctly.
- Added translations
- Added tests for viewModel, dataSource, and viewController.

# 👀 See

| Non Creator 🐛 | Creator 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-11-04 at 16 46 44 | Simulator Screen Shot - iPhone 8 - 2019-11-04 at 16 42 53 |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

** Logged-in using the `therealnativesquad` account:
- [ ] Select any non-created project from Discovey. The header should be hidden.
- [ ] Search for a project created using that account (e.g `Help me transform this pile of wood` and select it. The Project page should be presented with the header containing a label indicating when the project was created and a button showing `View progress` if project is live or `View dashboard` if project is no longer live.


Kickstarter-iOS/Views/Storyboards/ProjectPamphlet.storyboard
```diff
@@ -19,7 +19,7 @@
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
                             <containerView opaque="NO" contentMode="scaleToFill" insetsLayoutMarginsFromSafeArea="NO" translatesAutoresizingMaskIntoConstraints="NO" id="tIm-aj-d8Q">
-                                <rect key="frame" x="0.0" y="52" width="400" height="1348"/>
+                                <rect key="frame" x="0.0" y="55" width="400" height="1346"/>
```
Scollaco: Had to move the ProjectPamphletContent container down a little bit because it was overlapping the top cell.

Kickstarter-iOS/Views/Cells/ProjectPamphletCreatorHeaderCell.swift
```diff
@@ -0,0 +1,134 @@
...
+}
+
+protocol ProjectPamphletCreatorHeaderCellDelegate: class {
+  func projectPamphletCreatorHeaderCellDidTapButton(
+    _ cell: ProjectPamphletCreatorHeaderCell,
+    project: Project
+  )
+}
+
+final class ProjectPamphletCreatorHeaderCell: UITableViewCell, ValueCell {
```
Scollaco: Still not sure about this name 🤔  Would `ProjectPamphletViewProgressCell` be better?

Library/ViewModels/ProjectPamphletCreatorHeaderCellViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+  public var outputs: ProjectPamphletCreatorHeaderCellViewModelOutputs { return self }
+}
+
+private func title(for project: Project) -> String {
+  return project.state == .live ? Strings.View_progress() : Strings.View_dashboard()
+}
+
+private func attributedLaunchDateString(with project: Project)
+  -> NSAttributedString? {
+  let date = Format.date(secondsInUTC: project.dates.launchedAt, dateStyle: .long, timeStyle: .none)
```
ifbarrera: We _might_ want to pass a timezone of UTC to this date formatter so that the date doesn't get converted to the device's timezone? 🤔

Library/ViewModels/ProjectPamphletCreatorHeaderCellViewModelTests.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    self.buttonTitle.assertDidNotEmitValue()
+
+    self.vm.inputs.configure(with: project)
+
+    self.buttonTitle.assertValue("View dashboard")
+  }
+
+  func testLaunchDateLabelAttributedText() {
+    let project = Project.template
```
ifbarrera: Can we set the `.launchedAt` variable on the project to a date we've defined so we can see that it's actually displaying the correct date? Otherwise this test feels a little hand-wavey.

Kickstarter-iOS/Views/Cells/ProjectPamphletCreatorHeaderCell.swift
```diff
@@ -0,0 +1,134 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+private enum Layout {
+  enum Button {
+    static let height: CGFloat = 48
```
justinswart: Where are these constants from?

Library/ViewModels/ProjectPamphletCreatorHeaderCellViewModelTests.swift
Scollaco: Thats a good idea!

Kickstarter-iOS/Views/Cells/ProjectPamphletCreatorHeaderCell.swift
Scollaco: This comes from Abstract.

Kickstarter-iOS/Views/Cells/ProjectPamphletCreatorHeaderCell.swift
```diff
@@ -0,0 +1,134 @@
...
+  // MARK: - View model
+
+  override func bindViewModel() {
+    super.bindViewModel()
+    self.launchDateLabel.rac.attributedText = self.viewModel.outputs.launchDateLabelAttributedText
+    self.viewProgressButton.rac.title = self.viewModel.outputs.buttonTitle
+
+    self.viewModel.outputs.notifyDelegateViewProgressButtonTapped
+      .observeForUI()
+      .observeValues { [weak self] in
```
ifbarrera: Can we name the argument here? We typically name them when observing the values in `bindViewModel` to make it clear what the value is.

Library/ViewModels/ProjectPamphletCreatorHeaderCellViewModelTests.swift
ifbarrera: I meant more something like:

```
   // Set the date to Nov 1, 2019
    let dateComponents = DateComponents()
      |> \.month .~ 11
      |> \.day .~ 1
      |> \.year .~ 2_019
      |> \.timeZone .~ TimeZone.init(secondsFromGMT: 0)

      let date = AppEnvironment.current.calendar.date(from: dateComponents)

      let project = Project.template |> \.dates.launchedAt .~ date.timeIntervalSince1970
...
    // Assert that the date is Nov 1, 2019, which is the date you set in the "launchedAt" property
    self.launchDateLabelAttributedText.assertValue("You launched this project on November 1, 2019.")
```
You might also want to set the locale in `withEnvironment` to ensure that the date is always formatted in the same way.

Scollaco: Ohh I see!

PR #931, opened 2019-11-05
[NT-489] Default card selection improvement

justinswart:
# 📲 What

Fixes an issue where the first stored card was always selected in any context where the payment method could be set even when it was not used to back the project.

# 🤔 Why

Our logic for handling this was a bit shortsighted in that it would simply select the first card in the list every time. Instead we need to confirm that the card was in fact used to back the project and hasn't since been removed.

# 🛠 How

Updated our logic for this to take into account the `PaymentSource` of the `Backing` and if it's set, check that the card is still an available payment source.

# ✅ Acceptance criteria

- [ ] Back a project and delete the payment method used to back the project. When navigating to change your payment source, there shouldn't be any payment source selected by default and the `Confirm` button should be disabled.
- [ ] Back a project using Apple Pay. Similarly there shouldn't be any default payment method selected when changing.


Library/ViewModels/PledgeCreditCardViewModel.swift
```diff
@@ -96,16 +96,20 @@ public final class PledgeCreditCardViewModel: PledgeCreditCardViewModelInputs,
       .takeWhen(Signal.merge(cardConfiguredAsSelected, self.selectButtonTappedProperty.signal))
       .map { $0.id }
 
-    let cardIsSelected = cardAndSelectedCard
-      .map(==)
+    let cardIsSelected = Signal.merge(
+      creditCard.mapConst(false),
```
justinswart: Emitting `false` here to ensure that the card has a title even though we're not setting a selected card.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -136,18 +133,32 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+) -> ([PledgeCreditCardViewData], GraphUserCreditCard.CreditCard?) {
+  let data = cards.compactMap { card -> PledgeCreditCardViewData? in
     guard let cardBrand = card.type?.rawValue else { return nil }
 
     let isAvailableCardType = availableCardTypes.contains(cardBrand)
 
     return (card, isAvailableCardType, project.location.displayableName)
   }
+
+  guard let backing = project.personalization.backing else {
```
justinswart: If this project hasn't been backed then we do just select the first card.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -136,18 +133,32 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+
+  var backedCard: GraphUserCreditCard.CreditCard?
+
+  let cardIds = cards.map { $0.id }
+
+  if let cardId = backing.paymentSource?.id, cardIds.contains(cardId) {
+    backedCard = cards.first
+  }
+
+  return (data, backedCard)
```
Scollaco: Here I think we could simply use 
```
let backedCard = cards.first(where: { $0.id == backing.paymentSource?.id })
```

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -265,6 +265,55 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
     }
   }
 
+  func testReloadPaymentMethods_LoggedIn_ApplePayCapable_isTrue_BackedCardRemoved() {
+    let filteredCards = GraphUserCreditCard.template.storedCards.nodes
+      .filter { $0.id != GraphUserCreditCard.visa.id
+      }
```
Scollaco: Could you put the `}` on the same line?

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -265,6 +265,55 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+
+      self.vm.inputs.configureWith((User.template, project, true))
+      self.vm.inputs.viewDidLoad()
+
+      self.scheduler.run()
+
+      self.applePayButtonHidden.assertValues([false])
+      self.reloadPaymentMethodsCards.assertValue(response.me.storedCards.nodes)
+      self.reloadPaymentMethodsAvailableCardTypes.assertValues([
+        [true, true, true, true, true, false, true]
```
Scollaco: Why there's an unavailable card? 🤔

Library/ViewModels/PledgePaymentMethodsViewModel.swift
justinswart: lol thanks that's much better

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
justinswart: This is just the Project template's available card types, I copied this test from the one above it.

Scollaco: Oh cool!

PR #934, opened 2019-11-07
[NT-558][NT-508][NT-532] Indicate Backing state on Manage Pledge View

justinswart:
# 📲 What

Adds text describing the `Backing`'s state to `ManagePledgeViewController`.

# 🤔 Why

This provides some additional context to our backers that may be useful for them when viewing their pledge in the app.

# 🛠 How

- Added strings for each state.
- Added tests.
- Repurposes `attributedConfirmationString` from `PledgeViewModel` to be a bit more reusable.
- Updated snapshots for this view.
  - Didn't record new snapshots for each state as I feel the view model tests are covering this.

**Note:** Once we are sure that we have the final copy for all of the strings I will add them for translation and rebuild the strings.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ♿️ Accessibility

- [x] Works with VoiceOver
- [x] Supports Dynamic Type

# 🏎 Performance

# ✅ Acceptance criteria

**Note:** It may be easiest to test by manipulating the JSON response using breakpoints with Charles proxy.

For both backer and creator:

- [ ] Back a project, navigate to _Manage Pledge_. The copy that you see for each state should match the [JIRA ticket](https://dripsprint.atlassian.net/browse/NT-508).
- [ ] Note that if a project has a `canceled` or `failed` state, we show different copy also noted on the ticket.

# ⏰ TODO

- [x] Add strings for translation.


Library/ViewModels/PledgeStatusLabelViewModel.swift
```diff
@@ -0,0 +1,129 @@
...
+    )
+  case .live, .purged, .started, .submitted, .suspended, .successful:
+    return nil
+  }
+
+  return string
+}
+
+private func creatorStatusLabelText(with _: Project, status _: Backing.Status) -> NSAttributedString? {
+  return nil
```
justinswart: Will add this in the next PR for the creator context.

justinswart:
Apologies, this PR got a lot bigger than initially intended.

Kickstarter-iOS/Views/Controllers/ManagePledgeSummaryViewController.swift
```diff
@@ -12,6 +12,10 @@ final class ManagePledgeSummaryViewController: UIViewController {
   private lazy var backerNumberLabel: UILabel = { UILabel(frame: .zero) }()
   private lazy var backingDateLabel: UILabel = { UILabel(frame: .zero) }()
 
+  private lazy var pledgeStatusLabelViewController: PledgeStatusLabelViewController = {
```
ifbarrera: Just curious, why did you opt for using a VC here instead of a `UIView`? 🤔

Library/ViewModels/PledgeStatusLabelViewModelTests.swift
```diff
@@ -0,0 +1,240 @@
...
+
+        let project = Project.cosmicSurgery
+          |> Project.lens.creator .~ creator
+          |> Project.lens.state .~ .successful
+          |> Project.lens.personalization.backing .~ (
+            Backing.template
+              |> Backing.lens.status .~ $0
+          )
+
+        self.vm.inputs.configure(with: project)
```
ifbarrera: Do you need to call `self.vm.inputs.viewDidLoad()` somewhere?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -707,47 +711,34 @@ private func requiresSCA(_ envelope: StripeSCARequiring) -> Bool {
...
-    NSAttributedString.Key.foregroundColor: UIColor.ksr_text_dark_grey_500
+  let attributes = [
+    NSAttributedString.Key.paragraphStyle: paragraphStyle
   ]
-  let boldFontAttribute = [NSAttributedString.Key.font: UIFont.ksr_caption1().bolded]
 
-  attributedString.addAttributes(regularFontAttribute, range: fullRange)
-  attributedString.addAttributes(boldFontAttribute, range: rangePledgeTotal)
-  attributedString.addAttributes(boldFontAttribute, range: rangeProjectDeadline)
+  if project.stats.currentCurrency == project.stats.currency {
```
ifbarrera: Here too we could use `project.stats.needsConversion`

Kickstarter-iOS/Views/Controllers/ManagePledgeSummaryViewController.swift
justinswart: I just had a good feeling about it 😄

Library/ViewModels/PledgeStatusLabelViewModel.swift
```diff
@@ -0,0 +1,135 @@
...
+  let foregroundColor = UIColor.ksr_text_black
+
+  let paragraphStyle = NSMutableParagraphStyle()
+  paragraphStyle.alignment = .center
+
+  let attributes = [
+    NSAttributedString.Key.paragraphStyle: paragraphStyle
+  ]
+
+  if project.stats.currentCurrency == project.stats.currency {
```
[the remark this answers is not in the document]
justinswart: oh cool thanks!

Library/ViewModels/PledgeStatusLabelViewModelTests.swift
justinswart: 😬 😬

Library/ViewModels/PledgeViewModel.swift
justinswart: This change seemed to break snapshots. Will dig in more tomorrow.

PR #936, opened 2019-11-07
[NT-421] Improvements to the pledge CTA View on failure to load project

ifbarrera:
# 📲 What

This PR makes some improvements to the UI of the pledge CTA view when the project fails to load.

This PR also fixes a bug where the content inset on the project pamphlet page wasn't being correct set when the size of the pledge CTA view changes due to either an increase in dynamic type size or wrapping text.

# 🤔 Why

To improve the user experience.

# 🛠 How

Previously, we would show a button with some very long copy explaining that the content failed to load in the CTA. This didn't make it super obvious that the copy you were seeing was in fact an actionable button that you could tap to retry the request. Instead, we've split out the "Content isn't loading right now" copy into a label on the left, and put a more actionable button on the right with a CTA of "Retry".

# 👀 See

| After 🦋 | Before 🐛 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-11-07 at 15 38 20 |  Simulator Screen Shot - Clone 1 of iPhone 8 - 2019-11-07 at 15 35 19 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

N/A

# ✅ Acceptance criteria

Pre-requisites: Using Charles proxy, set up break points for responses so you can abort requests.
- [x] Navigate to any project and abort the request -- you should see the new "Content isn't loading view", and tapping "Retry" should prompt a new request to be triggered.


Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -130,10 +144,16 @@ final class PledgeCTAContainerView: UIView {
         self?.animateView(self?.pledgeCTAButton, isHidden: isHidden)
       }
 
+    self.viewModel.outputs.retryStackViewIsHidden
+      .observeForUI()
+      .observeValues { [weak self] isHidden in
+        self?.updateRetryStackViewIsHidden(isHidden)
```
ifbarrera: This is a work-around for the fact that the height of the `retryStackView` is *usually* (depending on the device) larger than the height of the other views in the pledge CTA view (manage button, "Back this project" button etc.). Because the height was larger, simply hiding the `retryStackView` meant that the pledge CTA view appeared to be too tall for the "Back this project" button, because the parent view is accommodating for the height of the hidden `retryStackView`. Another alternative to this approach would be to only set the text of the `retryDescriptionLabel` when the retry stack view is being shown – it didn't feel like a worthwhile enough reason to add more complexity to the view model, so I went with this approach instead. Open to comments in case I missed something.

PS. iterating through the `retryStackView`'s subviews and hiding them also worked, but this seemed like a cleaner approach 🤷‍♀

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
```diff
@@ -242,16 +277,20 @@ private let titleLabelStyle: LabelStyle = { label in
...
-    |> UIButton.lens.image(for: .normal) %~ { _ in image(named: "icon--refresh-small") }
-    |> UIButton.lens.image(for: .highlighted) %~ { _ in image(named: "icon--refresh-small", alpha: 0.66) }
-    |> UIButton.lens.title(for: .normal) %~ { _ in Strings.Content_isnt_loading_right_now() }
+private let retryStackViewStyle: StackViewStyle = { stackView in
+  stackView
+    |> \.axis .~ .horizontal
+    |> \.alignment .~ .center
+    |> \.spacing .~ Styles.grid(3)
+    |> \.isLayoutMarginsRelativeArrangement .~ true
+    |> \.layoutMargins .~ UIEdgeInsets.init(topBottom: Styles.grid(3), leftRight: Styles.grid(3))
```
justinswart: Could just do `UIEdgeInsets(all: Styles.grid(3))`

Kickstarter-iOS/Views/PledgeCTAContainerView.swift
justinswart: Is there a reason you chose to add the `retryStackView` with Auto Layout and not just to the `rootStackView`. I think if it were a child of a stack view then hiding/showing it should have the desired behaviour.

ifbarrera: 🤔 let me try that

ifbarrera: That worked 🌟 brilliant

PR #937, opened 2019-11-08
[NT-551] Crash on logout fix

Scollaco:
# 📲 What 
- Fixes a bug where the app crashes when user tries to logout.
- The crash started happening after the [PR](https://github.com/kickstarter/ios-oss/pull/933) that navigates from Project Page was merged. We were returning a `viewController` of type `T` from the function `viewControllerAndParam<T, P>(with :param:) -> (T, P)?` when we should be returning nil.

# 🤔 Why
- Crashes are a no no.

# 🛠 How
- Determining on `RootTabBarViewController` if we are switching to Dashboard. 

**If yes:** 
- We try to cast the `viewController[index]` to `UINavigationController` and then cast its first child to type `T`.

**If no:**
- We try to cast the `viewController[index]` to  type `T` directly.

We return `nil` otherwise. 

# 👀 See
logout


Kickstarter-iOS/Views/Controllers/RootTabBarViewController.swift
```diff
@@ -184,11 +184,11 @@ public final class RootTabBarViewController: UITabBarController {
   }
 
   private func viewControllerAndParam<T, P>(with index: RootViewControllerIndex, param: P) -> (T, P)? {
-    guard
-      let vcs = self.viewControllers,
+    guard let vcs = self.viewControllers,
```
justinswart: Can we make this small change for readability and to reuse the existing function to extract the VC:
```swift
guard
  let vcs = self.viewControllers,
  let vc = extractViewController(vcs[clamp(0, vcs.count - 1)(index)]) as? T
else { return nil }
```

PR #938, opened 2019-11-11
[NT-534] Show User Name and Avatar on the Manage Pledge screen

ifbarrera:
# 📲 What

This PR adds the user's avatar and name on the manage pledge screen if the current user is the backer of that project.

Also adds a placeholder image for remote avatar URL loading so the transition is more subtle.

# 🤔 Why

More visual context on the manage pledge screen.

# 🛠 How

We check whether the current user is the backer of the project, and if they are, we show their avatar and name on the Manage/View pledge screen. If the current user *is not* the backer of the project, we hide the avatar and name labels.

# 👀 See

| Current User Is Backer | Current User is Not Backer |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-11-08 at 13 18 09 | Simulator Screen Shot - iPhone 8 - 2019-11-08 at 16 29 25 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

- [x] Optimized Blended Layers

# ✅ Acceptance criteria

- [ ] Navigate to the manage pledge screen on a backed project. You should see your avatar and your name.


Kickstarter-iOS/Views/Cells/ProjectPamphletCreatorHeaderCell.swift
```diff
@@ -10,7 +10,7 @@ private enum Layout {
   }
 }
 
-protocol ProjectPamphletCreatorHeaderCellDelegate: class {
+protocol ProjectPamphletCreatorHeaderCellDelegate: AnyObject {
```
ifbarrera: This was done by `swiftformat`, seems it was missed in a recent PR.

Library/Styles/CheckoutStyles.swift
```diff
@@ -105,6 +105,11 @@ public let checkoutBackgroundStyle: ViewStyle = { (view: UIView) in
     |> \.backgroundColor .~ UIColor.ksr_grey_300
 }
 
+public let checkoutLabelStyle: LabelStyle = { label in
```
ifbarrera: To prevent color blending on labels.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewControllerTests.swift
```diff
@@ -19,12 +19,17 @@ final class ManagePledgeViewControllerTests: TestCase {
     super.tearDown()
   }
 
-  func testView() {
+  func testView_CurrentUser_IsBacker() {
```
justinswart: Is this just future-proofing? As I understand it from our discussions the user will always be the backer? Which is why I removed the creator strings in #934 🤔

Library/ViewModels/ManagePledgeSummaryViewModel.swift
```diff
@@ -38,6 +42,31 @@ public class ManagePledgeSummaryViewModel: ManagePledgeSummaryViewModelType,
...
+
+    let userBackingProject = userAndIsBackingProject
+      .filter(second >>> isTrue)
+      .map(first)
+
+    self.backerNameText = userBackingProject
+      .map(\.name)
+
+    self.backerImageURLAndPlaceholderImageName = userBackingProject
+      .map(\.avatar.small)
```
justinswart: 😮 🤯

Library/ViewModels/ManagePledgeSummaryViewModel.swift
```diff
@@ -38,6 +42,31 @@ public class ManagePledgeSummaryViewModel: ManagePledgeSummaryViewModelType,
...
+      .map(URL.init)
+      .skipNil()
+      .map { ($0, "avatar--placeholder") }
```
justinswart: Wonder if we could get this placeholder to show up in our snapshot tests 🤔

ifbarrera: Ah! I think maybe we just need to pass the bundle 🤔let me try it

Kickstarter-iOS/Views/Controllers/ManagePledgeViewControllerTests.swift
ifbarrera: Yeah sort of - although there's currently no way to get to this state, the VM does support it so I created a screenshot so we can verify that it visually matches what we expect.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewControllerTests.swift
```diff
@@ -41,11 +46,42 @@ final class ManagePledgeViewControllerTests: TestCase {
     }
   }
 
+  func testView_CurrentUser_IsNotBacker() {
```
Scollaco: I think the name can cause confusion in the future, since only creators and backers will have access to this screen (and for now, we're not showing this screen to creators). So I think we could either delete it or maybe rename and update it to test view when user is creator of project?

Library/ViewModels/ManagePledgeSummaryViewModel.swift
```diff
@@ -9,8 +9,12 @@ public protocol ManagePledgeSummaryViewModelInputs {
 }
 
 public protocol ManagePledgeSummaryViewModelOutputs {
+  var backerImageURLAndPlaceholderImageName: Signal<(URL, String), Never> { get }
+  var backerNameText: Signal<String, Never> { get }
```
Scollaco: alpha is off here 🙈

Library/ViewModels/ManagePledgeSummaryViewModel.swift
```diff
@@ -61,8 +90,12 @@ public class ManagePledgeSummaryViewModel: ManagePledgeSummaryViewModelType,
     self.viewDidLoadObserver.send(value: ())
   }
 
+  public let backerImageURLAndPlaceholderImageName: Signal<(URL, String), Never>
+  public let backerNameText: Signal<String, Never>
```
Scollaco: And here too

Kickstarter-iOS/Views/Controllers/ManagePledgeSummaryViewController.swift
```diff
@@ -77,8 +99,22 @@ final class ManagePledgeSummaryViewController: UIViewController {
         self?.pledgeAmountSummaryViewController.configureWith(project)
       }
 
+    self.viewModel.outputs.backerImageURLAndPlaceholderImageName
+      .observeForUI()
+      .on(event: { [weak self] _ in
+        self?.circleAvatarImageView.af_cancelImageRequest()
```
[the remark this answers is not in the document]
ifbarrera: Hmm is it possible that the user could navigate away from the VC before the request has completed, and then navigates back to it? I figure it doesn't hurt 🤔

Kickstarter-iOS/Views/Controllers/ManagePledgeViewControllerTests.swift
ifbarrera: I think it's helpful to have this screenshot as a reference for how the view should look if the current user is not the backer 🤷‍♀

Library/ViewModels/ManagePledgeSummaryViewModel.swift
ifbarrera: Hmm that didn't work. I think it might have something to do with how alamofire animates the placeholder in. We should spend more time looking into it outside of this PR.

PR #939, opened 2019-11-11
[NT-514] "View your rewards" creator CTA

ifbarrera:
# 📲 What

Updates the Pledge CTA to `View your rewards` if the current user is the creator of the project. This CTA will display regardless of whether the project is live or not.

# 🤔 Why

So that creators can view their rewards, but not pledge to their own project.

# 🛠 How

Added a `.viewYourRewards` state to the `PledgeStateCTAType` enum and corresponding visual treatment. When the CTA button is tapped, the user is taken to the `RewardsCollectionViewController`.

# 👀 See

ey9CtJQud7

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] As a creator, navigate to the project page of a project you created. You can do this via Search or via the Dashboard. You should see the CTA button on the project page; it should be black, and say "View your rewards". Tapping the "View your rewards" button should take you to the rewards carousel.


Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -112,7 +112,7 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .skipNil()
 
     let shouldGoToRewards = ctaButtonTappedWithType
-      .filter { $0 == .pledge || $0 == .viewRewards }
+      .filter { [.pledge, .viewRewards, .viewYourRewards].contains($0) }
```
justinswart: `.contains`! my favourite 😅

PR #948, opened 2019-11-14
[NT-570] Update Fastlane, fix command in circle config

justinswart:
# 📲 What

Updates Fastlane and its plugins to suppress some gem security warnings on our repo. Also renames the command `itunes_match_gym_deliver_appcenter` -> `itunes_match_gym_deliver` in `.circleci/config.yml` so that it matches what we have in our Fastfile.

# 🤔 Why

Our app store builds were failing due to the command name being incorrect.

# 🛠 How

- `bundle update fastlane`
- `bundle exec fastlane update_plugins`
- Rename command.

# 👀 See

CircleCI error:

image

# ✅ Acceptance criteria

- [ ] Renamed command matches that in the `Fastfile` and we're able to submit builds to the app store via `make deploy`.
- [ ] Fastlane jobs run as before.


justinswart:
Just noticed this is currently failing on Circle because of the ruby version:

image

PR #953, opened 2019-11-18
[NT-548] Stepper min value bug fix

Scollaco:
# 📲 What
- Fixes a bug where the `-` button of the stepper kept enabled even after typing a value below the reward minimum on the `PledgeAmountViewController`.

# 🤔 Why
- By design, the button should stay disabled if the amount typed is below Reward minimum.

# 🛠 How
- The stepper in value was being set to `0`, so I just needed to change it to match the reward `minimum`.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2019-11-18 at 14 44 20 | Simulator Screen Shot - iPhone 8 - 2019-11-18 at 14 41 15 |

# ✅ Acceptance criteria

- [ ] Select a reward and go to the Pledge screen. The `-` button of the stepper should be disabled.
- [ ] Tap the `+` until the amount is greater than the reward `minimum`. The `-` should become enabled.
- [ ] Using the textField, tap any amount below the minimum. The `-` button of the stepper should be disabled.
- [ ] Tapping any amount greater than the `minimum` should re-enable the `-` button.


Scollaco:
> Nice @Scollaco I think this looks good. Just one question.
> 
> If I set the value to below the minimum and invalidate the form. If I then tap the `+` it sets the value to `min + 1`, I'm wondering if it wouldn't be better to bring the value up to the minimum? In this case `35` instead of `36`.
> 
> I think it might be unrelated to the change in this ticket but is possibly a usability issue to consider.
> 
> Example:
> 
> 

Interesting! I remember having a conversation with Danny about this and he said that the field should show the `minimum`. I wonder when this behaviour changed 🤔

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -111,7 +110,24 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
...
+      maxValue,
+      Signal.merge(
+        stepperValue,
+        textFieldValue.signal
+      )
+    )
+
+    self.stepperMinValue = updatedValue
+      .map { ($0.0, $0.2) }
+      .map(stepperMinValue(with:value:))
```
justinswart: You can just use the built-in `min` function here:

```swift
self.stepperMinValue = updatedValue
  .map { ($0.0, $0.2) }
  .map(min)
```

PR #954, opened 2019-11-18
[NT-526] Max pledge error handling polish

Scollaco:
# 📲 What 
- Updates the error message when user tries to back a project with an amount greater than the maximum allowed.
- Changes label's textAlignment to `.right`

# 🤔 Why
- To match the designs.
[JIRA ticket](https://dripsprint.atlassian.net/browse/NT-526)

# 🛠 How
- Replacing string
- Changing `textAlignment` of label to `.right`
- Updating tests and snapshots

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone Xʀ - 2019-11-18 at 16 44 27 | Simulator Screen Shot - iPhone Xʀ - 2019-11-18 at 15 39 54 |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Select a reward from any project and type an amount greater than the `maximum` value. A label with the text `The maximum pledge is [currency] [maximum pledge amount]`.
- [x] The error message should appear below the textField and aligned to the right.


Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -158,11 +158,10 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
       .map { ($0.0, $0.1) }
```
cdolm92: Should we just map to the max amount because we're not using the minimum amount anymore.

PR #955, opened 2019-11-19
[NT-581] Re-set visitor cookie before fetching config

justinswart:
# 📲 What

- Ensures that `identifierForVendor` is set in our visitor cookie before fetching the config.
- Adds a `no_cache` query param to the request that fetches the config to ensure that it is not cached.

# 🤔 Why

We've found that in some scenarios, logging out in particular, we refresh the config without a visitor cookie being set. This causes unexpected behaviour in the way our A/B experiments work as the back-end depends on the identifier in this cookie in order to place the user into the experimental or control group.

We've also noticed that we are sometimes receiving `304 Not Modified` responses to our requests for the user config. We always want a fresh config when we request one.

# 🛠 How

- Moved the code that sets our visitor cookie to just before we fetch the user config.
  - I added a test to ensure that this does not modify existing cookies but only sets this new value.
- Added a query param called `no_cache` to our `config` `Route` which passes in the current timestamp as its value. This ensures that the backend sees each request as unique.

# ✅ Acceptance criteria

Using Charles proxy, observe the responses on the `/v1/app/ios/config` endpoint.

- [x] You should see a query param `no_cache` with the current timestamp as its value and this value should change with each request.
- [x] There should _always_ be a JSON response from this endpoint.
- [x] Run the app logged out, observe that you are (or aren't) in the experiment group. Log in, observe that your experiment membership didn't change. Log out, again observe that it didn't change.
- [x] Delete the app, install and run it again. Repeat the previous steps. Try to do this for both scenarios where you initially begin in either the control group or the experimental group and remain in those across login and logout events.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -190,13 +190,19 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
       .filter { $0.ksrCode == .AccessTokenInvalid }
       .ignoreValues()
 
-    self.updateConfigInEnvironment = Signal.merge([
-      self.applicationWillEnterForegroundProperty.signal,
-      self.applicationLaunchOptionsProperty.signal.ignoreValues(),
-      self.userSessionEndedProperty.signal,
-      self.userSessionStartedProperty.signal
-    ])
-      .switchMap { AppEnvironment.current.apiService.fetchConfig().demoteErrors() }
+    self.updateConfigInEnvironment = Signal.merge(
+      [
+        self.applicationWillEnterForegroundProperty.signal,
+        self.applicationLaunchOptionsProperty.signal.ignoreValues(),
+        self.userSessionEndedProperty.signal,
+        self.userSessionStartedProperty.signal
+      ]
+    )
+    .switchMap { _ -> SignalProducer<Config, Never> in
+      visitorCookies().forEach(AppEnvironment.current.cookieStorage.setCookie)
+
+      return AppEnvironment.current.apiService.fetchConfig().demoteErrors()
+    }
 
     self.postNotification = self.currentUserUpdatedInEnvironmentProperty.signal
       .mapConst(Notification(name: .ksr_userUpdated, object: nil))
@@ -547,12 +553,6 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
 
     self.crashManagerDidFinishSendingCrashReportProperty.signal
       .observeValues { AppEnvironment.current.koala.trackCrashedApp() }
-
-    self.applicationLaunchOptionsProperty.signal
-      .take(first: 1)
-      .observeValues { _ in
-        visitorCookies().forEach(AppEnvironment.current.cookieStorage.setCookie)
-      }
 
     Signal.combineLatest(
       performShortcutItem.enumerated(),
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -548,12 +554,6 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     self.crashManagerDidFinishSendingCrashReportProperty.signal
       .observeValues { AppEnvironment.current.koala.trackCrashedApp() }
 
-    self.applicationLaunchOptionsProperty.signal
```
justinswart: This event is in the merged signal that triggers the config fetch so we didn't need this anymore. One thing to note is that this did have `take(first: 1)` so it was explicitly only meant to happen once per app session. I can't really think of any reason that it would be problematic to do this many times per session (backrounding/foregrounding) but worth considering!

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1355,7 +1355,7 @@ final class AppDelegateViewModelTests: TestCase {
     )
   }
 
-  func testVisitorCookies() {
+  func testVisitorCookies_ApplicationDidFinishLaunching() {
```
justinswart: This test is unchanged just to prove that existing functionality wasn't affected.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1368,6 +1368,99 @@ final class AppDelegateViewModelTests: TestCase {
     )
   }
 
+  func testVisitorCookies_ApplicationWillEnterForeground() {
+    let existingCookie = HTTPCookie(
```
justinswart: In these other tests I've added an existing cookie to test that setting the visitor cookie multiple times does not affect any other cookies we may have set.

KsApi/lib/Route.swift
```diff
@@ -121,7 +121,7 @@ internal enum Route {
       return (.GET, url, [:], nil)
 
     case .config:
-      return (.GET, "/v1/app/ios/config", [:], nil)
+      return (.GET, "/v1/app/ios/config", ["no_cache": "\(Date().timeIntervalSince1970)"], nil)
```
ifbarrera: I wonder if in the future we could update this to instead use `URLRequest`'s `cachePolicy` of `reloadIgnoringLocalAndRemoteCacheData` 🤔

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -1368,6 +1368,99 @@ final class AppDelegateViewModelTests: TestCase {
...
+
+    AppEnvironment.current.cookieStorage.setCookie(existingCookie!)
+
+    self.vm.inputs.applicationWillEnterForeground()
+
+    XCTAssertEqual(
+      ["existing-cookie", "vis", "vis"],
+      AppEnvironment.current.cookieStorage.cookies!.map { $0.name }.sorted()
+    )
+    XCTAssertEqual(
```
ifbarrera: Should we add an assertion for the cookie domain? I think we should have two different domains for the two `vis` cookies.

justinswart: Sure, can do!

KsApi/lib/Route.swift
justinswart: Yeah good point - this would require a bit more work because it would mean having a separate `URLSession` for just this request. I think this should be fine for now and hopefully we can remove this when we disable caching for this endpoint at the backend.

PR #956, opened 2019-11-19
[NT-573] Pull to refresh functionality on the Manage/View Pledge screen

ifbarrera:
# 📲 What

Adds a `UIRefreshControl` to support pull-to-refresh functionality on the Manage/View pledge screen.

# 🤔 Why

So that users have a way to refresh the screen and get the most up to date information regarding their backing.

# 🛠 How

- adds a `UIRefreshControl` to the root scroll view
- hooks up the `.valueChanged` action on the `UIRefreshControl` to fetching an updated project in the view model

# 👀 See

EPMwCeQVAY

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] navigate to the manage pledge screen and pull down to see the refresh icon. The screen should refresh every time you pull it down.


Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -175,6 +177,12 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
         self?.rewardView.configure(with: $0)
       }
 
+    self.viewModel.outputs.endRefreshing
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.rootScrollView.refreshControl?.endRefreshing()
```
justinswart: I guess this could reference `self.refreshControl` directly 🤷‍♂ - or seeing as we don't customize it and are able to refer to it through the scrollView we could just set it on the scrollView and not have a property referencing it. Up to you.

Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
```diff
@@ -294,6 +311,10 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
     self.viewModel.inputs.menuButtonTapped()
   }
 
+  @objc private func handleRefreshControl() {
```
justinswart: Could just be `refresh()` or `beginRefresh()`

PR #957, opened 2019-11-19
[NT-577] Unavailable Card Selection Fix

cdolm92:
# 📲 What

When the first card is an unavailable card, it should not be selected and the Pledge button should appear disabled.

# 🤔 Why

Before the pledge button was enabled even when the first card was unavailable and therefore disabled.
 
# 🛠 How

We check to see if the first card is an available card type before selection.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|Simulator Screen Shot - iPhone Xs - 2019-11-19 at 17 18 56|Simulator Screen Shot - iPhone Xs - 2019-11-19 at 17 16 44|

# ✅ Acceptance criteria

- [x] Navigate to the pledge screen to back a project. If you have an unavailable card listed first you the pledge button should be disabled


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -179,8 +179,16 @@ private func pledgeCreditCardViewDataAndSelectedCard(
     return (card, isAvailableCardType, project.location.displayableName)
   }
 
-  // If there is no backing, simply select the first card in the list.
+  // If there is no backing, simply select the first card in the list when it is an available card type.
   guard let backing = project.personalization.backing else {
+    if let firstCardBrand = cards.first?.type?.rawValue {
```
ifbarrera: I think we could actually use the `data` that was computed in the line above, since we've already computed the first card's availability. This will save us from having to iterate through `availableCardTypes` to find the card brand again:

```
    guard let cardData = data.first, cardData.isEnabled else {
      return (data, nil)
    }

    return (data, cards.first)
```

Issue #949, opened 2019-11-14
Show/Hide Password button overridden by orientation change 


robjaq:
System: 12.4
Device: iPhone X
Version: 3.20.1
Steps: Type into password field > tap eye icon to unhide > rotate > password field autohides > then you have to double tap the icon to show again

![pic1](https://user-images.githubusercontent.com/30357661/68901987-f6727980-06eb-11ea-8d61-ac74b6a26068.gif)

I would like to submit a pr if that's cool. This project is pretty dope =D
![pic2](https://user-images.githubusercontent.com/30357661/68902170-6e40a400-06ec-11ea-9e7a-168bd461cc83.gif)

justinswart:
hey @robjaq thanks for letting us know and for digging into it! I believe you've spoken to @ifbarrera about this separately - feel free to open a PR from a fork, our tests should run on it and we'll review your PR to help you get it merged 👍

justinswart:
Thanks! Closing this for now, we can track is on the PR that you opened.

PR #959, opened 2019-11-20
Add traitCollection input to LoginVM

robjaq:
Wazzzzzzaaaap! :P Here is a proposed solution.
# 📲 What

Issue #949

# 🤔 Why

Eye icon and secure text loses its style during orientation change. 

# 🛠 How

Add a trait collection input to the LoginViewModel and fire an event from the LoginVC for a trait collection change.

# 👀 See

Issue #949


Library/ViewModels/LoginViewModel.swift
```diff
@@ -42,6 +42,9 @@ public protocol LoginViewModelInputs {
 
   /// Call when the view will appear.
   func viewWillAppear()
+  
+  /// Call when a trait collection is applied.
+  func traitCollectionApplied()
```
justinswart: We usually like to keep the input names as close as possible to the way it's named outside of the view model, so in this case it would be:

You can rename the input along with its backed properties, etc.

```swift
/// Call when a trait collection did change.
func traitCollectionDidChange()
```

Library/ViewModels/LoginViewModel.swift
```diff
@@ -171,7 +174,10 @@ public final class LoginViewModel: LoginViewModelType, LoginViewModelInputs, Log
     self.logIntoEnvironment
       .observeValues { _ in AppEnvironment.current.koala.trackLoginSuccess(authType: Koala.AuthType.email) }
 
-    self.showHidePasswordButtonToggled = self.shouldShowPasswordProperty.signal
+    self.showHidePasswordButtonToggled = Signal.combineLatest(
```
justinswart: In this case let's consider what we're wanting to express with the `Signal`. We're saying:

1. Emit `showHidePasswordButtonToggled` when `shouldShowPasswordProperty` emits its value.
1. _Also_ emit that same value when `traitCollectionDidChange` emits.

So by using `comineLatest` here, we're requiring that both of the signals emit in order for `showHidePasswordButtonToggled` to emit, which isn't really necessary. Instead I think what we want is to `merge` like this:

```swift
let shouldShowPassword = self.shouldShowPasswordProperty.signal

self.showHidePasswordButtonToggled = Signal.merge(
  shouldShowPassword,
  shouldShowPassword.takeWhen(self.traitCollectionDidChangeProperty.signal)
)
```

This captures it a little more clearly by saying that we want to _take_ the value of `shouldShowPassword` _when_ `self.traitCollectionDidChangeProperty.signal` emits.

Library/ViewModels/LoginViewModelTests.swift
```diff
@@ -314,6 +314,7 @@ final class LoginViewModelTests: TestCase {
   func testShowPassword() {
     self.vm.inputs.viewWillAppear()
     self.vm.inputs.viewDidLoad()
+    self.vm.inputs.traitCollectionApplied()
```
justinswart: After you make the previous change I suggested then you won't need this in this particular test because we will have changed the `combineLatest` to a `merge`.

Please add an additional test which tests the `traitCollectionDidChange` input.

justinswart:
@robjaq thanks for addressing that feedback. I have opened another PR #964 which, once we've merged into this one, should hopefully get all of our workflows passing on your branch.

In the mean time, please run `bin/format.sh .` in the project root, that will format your code according to our formatting rules. It should just strip out some whitespace and move some parentheses around.

robjaq:
👌

justinswart:
@robjaq hey, please pull the latest from `master` and update your fork, i think that should get all of the jobs passing.

robjaq:
THat is the weekend!
yeahdawg
lol seriously no prob and thank you for the feedback and insight!

Also feel free to merge it, since I do not have write access

PR #961, opened 2019-11-21
[NT-594] Add EditorialProjectsViewController

justinswart:
# 📲 What

Adds `EditorialProjectsViewController` which embeds `DiscoveryPageViewController` for displaying a particular collection of projects.

# 🤔 Why

This was just an easy way to reuse all of functionality baked into `DiscoveryPageViewController` in order to show projects tagged for the _Go Rewardless_ campaign.

# 🛠 How

- Embedded `DiscoveryPageViewController` in `EditorialProjectsViewController`.
- Modified `DiscoveryPageViewController` to be configured with a `preferredBackgroundColor`.
  - This allows us to configure the `UITableView` with a `clear` `backgroundColor` in order for it to slide over the header view that we are adding in `EditorialProjectsViewController`.
- Set the `contentOffset` of `DiscoveryPageViewController`'s `UITableView` so that the project cards are positioned below the header view.

**Note:** Some strangeness around how the `UIRefreshControl` behaves in `DiscoveryPageViewController` made it necessary to tweak things slightly so that the `projectsAreLoadingAnimated` output of `DiscoveryPageViewModel` emits on the next run loop. Without this it was causing the `UITableView` to arbitrarily inset itself by the height of the `UIRefreshControl` 🤷‍♂. Thanks @Scollaco for narrowing it down to the `UIRefreshControl`. #636 seems to be related.

# 👀 See

**Note:** Just using a placeholder blue colour and height for the header view currently:

image

# ♿️ Accessibility 

- [ ] Works with VoiceOver

# ✅ Acceptance criteria

Put the following at the bottom of `AppDelegate`'s `application(_:didFinishLaunchingWithOptions:)`:

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
  let vc = EditorialProjectsViewController.instantiate()

  self.window?.rootViewController?.present(vc, animated: true)
}
```

- [ ] You should see the view controller in the screenshot above.
- [ ] Scrolling up should scroll the project cards over the header view.
- [ ] If you are able to scroll to the end of the project list the bottom of the view extending beyond the safe area should be white. Pro tip: Use tag ID `250` for a short list of projects.

# ⏰ TODO

- [ ] Accessibility strings for close button.


Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -98,6 +95,11 @@ internal final class DiscoveryPageViewController: UITableViewController {
 
     _ = self
       |> baseTableControllerStyle(estimatedRowHeight: 200.0)
+
+    if let preferredBackgroundColor = self.preferredBackgroundColor {
```
justinswart: `baseTableControllerStyle` always sets the background to `white`, this was necessary for us to control it from the parent.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+  // MARK: - Properties
+
+  private lazy var closeButton: UIButton = {
+    UIButton(type: .custom)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private lazy var discoveryPageViewController: DiscoveryPageViewController = {
+    DiscoveryPageViewController.configuredWith(sort: .newest)
+      |> \.preferredBackgroundColor .~ .clear
```
justinswart: Setting this here and not in `bindStyles` because the property must be set when the VC is instantiated so that it will be available at the time that `bindStyles` is called in `DiscoveryPageViewController`.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+    self.setupConstraints()
+
+    self.closeButton.addTarget(
+      self,
+      action: #selector(EditorialProjectsViewController.closeButtonTapped),
+      for: .touchUpInside
+    )
+
+    // Will be moved to be a testable output of VC that presents this one
+    self.viewModel.inputs.configure(with: .goRewardless)
```
justinswart: As the comment says - once we have the navigation flow for this we would move this configuration and write a test for that output.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+    self.viewModel.inputs.viewDidLoad()
+  }
+
+  public override func viewDidLayoutSubviews() {
+    super.viewDidLayoutSubviews()
+
+    let currentTableViewInsets = self.discoveryPageViewController.tableView.contentInset
+
+    self.discoveryPageViewController.tableView.contentInset = currentTableViewInsets
+      |> UIEdgeInsets.lens.top .~ (self.headerView.frame.height - self.view.safeAreaInsets.top)
```
justinswart: @ifbarrera we were using the correct values all along!

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    self.discoveryPageViewController.tableView.scrollIndicatorInsets =
+      self.discoveryPageViewController.tableView.contentInset
+  }
+
+  public override var preferredStatusBarStyle: UIStatusBarStyle {
+    return .lightContent
```
justinswart: This works for the dark blue on the header but not so much when we scroll the cards up 🤷‍♂

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+  // MARK: - Styles
+
+  public override func bindStyles() {
+    super.bindStyles()
+
+    _ = self.view
+      |> \.backgroundColor .~ UIColor.white
+
+    _ = self.headerView
+      |> \.backgroundColor .~ UIColor.hex(0x00007D)
```
justinswart: Temp colour, should be updated in subsequent PR that fills in the header.

Library/ViewModels/DiscoveryPageViewModelTests.swift
```diff
@@ -31,6 +31,8 @@ internal final class DiscoveryPageViewModelTests: TestCase {
   internal override func setUp() {
     super.setUp()
 
+    self.vm = DiscoveryPageViewModel()
```
justinswart: Moved this here so that the VM is always created after the environment is set up, it was causing the scheduler advancement to fail in the test below.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewControllerTests.swift
```diff
@@ -0,0 +1 @@
+import Foundation
```
justinswart: Placeholder file for snapshot tests once we've completed the header.

PR #962, opened 2019-11-21
[NT-590] Editorial "go rewardless" header

ifbarrera:
# 📲 What

Adds an editorial header at the top of the discovery screen, below the login/onboarding cell and above activity updates.

# 🤔 Why

So we can show editorial content at the top of the discovery screen 🎉 

# 🛠 How

Right now, the header is configured basically on the initial load of `DiscoveryPageViewController`, as long as the `.sort` is `magic` and the feature flag is turned on. 

This means that only the initial value of the feature flag will be taken into account -- if an updated config loads some time later, the editorial header won't update based on the updated config value. In an effort to reduce the size of this PR, I've decided to handle observing changes in the config in a subsequent PR.

# 👀 See

Simulator Screen Shot - iPhone 8 - 2019-11-21 at 12 02 47

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [x] verify that the header appears logged in: The easiest way to test this is toggle the `ios_go_rewardless` feature flag on staging, then restart the app twice for the editorial header to show. This is because the cached value of the config is what's used to determine the feature flag value, so we need an updated `config` to be cached before the correct value is used. As mentioned above, this behavior will be improved in a subsequent PR.
- [x] verify that the header appears logged out: toggle the feature flag on the debug menu, then restart the app - you should see the editorial header


Kickstarter-iOS/Views/Cells/DiscoveryEditorialCell.swift
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// MARK: - Styles
+
+private let containerViewStyle: ViewStyle = { view in
+  view
+    |> roundedStyle(cornerRadius: Styles.grid(2))
+    |> \.backgroundColor .~ .ksr_trust_700
+    |> \.isAccessibilityElement .~ true
+    |> \.accessibilityTraits .~ [UIAccessibilityTraits.button]
```
ifbarrera: Open to other suggestions on how best to make this header accessible.

Kickstarter-iOS/Views/Cells/DiscoveryEditorialCell.swift
```diff
@@ -0,0 +1,184 @@
...
+    _ = (self.editorialImageView, self.containerView)
+      |> ksr_addSubviewToParent()
+
+    _ = ([self.editorialTitleLabel, self.editorialSubtitleLabel], self.rootStackView)
+      |> ksr_addArrangedSubviewsToStackView()
+
+    let tapGestureRecognizer = UITapGestureRecognizer(
+      target: self,
+      action: #selector(DiscoveryEditorialCell
+        .editorialCellTapped)
```
justinswart: Can this go on the previous line pls

Kickstarter-iOS/Views/Cells/DiscoveryEditorialCell.swift
justinswart: I think `button` is closest to what we want.

Kickstarter-iOS/Views/Cells/DiscoveryEditorialCell.swift
```diff
@@ -0,0 +1,184 @@
...
+}
+
+private let rootStackViewStyle: StackViewStyle = { stackView in
+  stackView
+    |> \.axis .~ .vertical
+    |> \.spacing .~ Styles.grid(2)
+    |> \.isLayoutMarginsRelativeArrangement .~ true
+    |> \.layoutMargins .~ UIEdgeInsets.init(
+      top: Styles.grid(3), left: Styles.grid(3), bottom: 0,
+      right: Styles.grid(3)
```
justinswart: Formatting a little strange here.

Kickstarter-iOS/Views/Cells/DiscoveryOnboardingCell.swift
```diff
@@ -27,8 +27,11 @@ internal final class DiscoveryOnboardingCell: UITableViewCell, ValueCell {
       |> baseTableViewCellStyle()
       |> DiscoveryOnboardingCell.lens.contentView.layoutMargins %~~ { layoutMargins, cell in
         cell.traitCollection.isRegularRegular
-          ? .init(topBottom: Styles.grid(8), leftRight: Styles.grid(30))
-          : .init(topBottom: Styles.grid(6), leftRight: layoutMargins.left)
+          ? UIEdgeInsets.init(top: Styles.grid(5), left: Styles.grid(30), bottom: 0, right: Styles.grid(30))
+          : UIEdgeInsets.init(
+            top: Styles.grid(3), left: layoutMargins.left, bottom: 0,
```
justinswart: And here. Also none of the `UIEdgeInsets.init`s need `.init`.

Library/ViewModels/DiscoveryEditorialViewModel.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import ReactiveSwift
+
+public typealias DiscoveryEditorialCellValue = (
+  title: String,
+  subtitle: String, imageName: String, tag: String, refTag: RefTag
```
justinswart: I would do all on the same line or all on a new line 😬

Library/ViewModels/DiscoveryEditorialViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+  var inputs: DiscoveryEditorialViewModelInputs { get }
+  var outputs: DiscoveryEditorialViewModelOutputs { get }
+}
+
+public final class DiscoveryEditorialViewModel: DiscoveryEditorialViewModelType,
+  DiscoveryEditorialViewModelInputs, DiscoveryEditorialViewModelOutputs {
+  public init() {
+    let configureWithValue = self.configureWithValueProperty.signal.skipNil()
+
+    self.imageName = configureWithValue.map { $0.imageName }
```
justinswart: Not using keypaths? No matter, just asking.

ifbarrera: It's a tuple, so no key path support 😔

PR #963, opened 2019-11-21
[NT-555] Always show Apple Pay for logged in Users

cdolm92:
# 📲 What

We want to show users the apple pay button even when it's not set up.

# 🤔 Why

So that users know it is a payment option.

# 🛠 How

Deleted some code that set the stackview where the button is in to hidden.
# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| IMG_0056|IMG_0055|

# ✅ Acceptance criteria

- [ ] Test on a device w/o apple pay setup. Navigate to a project, tap `Back this project`. You should see the pledge screen with the apple pay button shown. 

*Test for non Apple Pay countries*
- [ ] Search `Marz live`. This project's location is Mexico, Mexico. Tap `Back this project`, you should not see the `Apple Pay` button.


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -53,8 +53,7 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
       }
 
     self.applePayStackViewHidden = configureWithValue
-      .map { ($0.project, $0.applePayCapable) }
-      .map(showApplePayButton(for:applePayCapable:))
+      .map { showApplePayButton(for: $0.project) }
```
ifbarrera: Hmmm I think we actually still need an `applePayCapable` value, because we need to know whether the device itself supports Apple Pay at all. The way we can determine this is by using `PKPaymentAuthorizationViewController.applePayDevice()`, which would passed in here: https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift#L170 .

Additionally, we might also want to include this behavior on `DeprecatedRewardPledgeViewController`, so that it's consistent between the two experiences: https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/Views/Controllers/DeprecatedRewardPledgeViewController.swift#L67

We should also be able to remove all of the custom logic in `PKPaymentMethodsViewController+Helpers` now that we're only using `.applePayDevice()` and no longer care about which project is being passed in.

cdolm92: Will check w/ product to make sure we want to include that on the `DeprecatedRewardPledgeViewController` and then file a separate ticket/PR. Removing the custom logic can also go in that separate ticket/PR.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -202,8 +202,8 @@ private func pledgeCreditCardViewDataAndSelectedCard(
   return (data, backedCard)
 }
 
-private func showApplePayButton(for project: Project, applePayCapable: Bool) -> Bool {
-  return applePayCapable &&
+private func showApplePayButton(for project: Project, applePayDevice: Bool) -> Bool {
+  return applePayDevice &&
```
nativeksr: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/kickstarter/ios-oss/blob/5b8f50da8451b55c1820ae8ec887b2a7e5064f91/Library/ViewModels/PledgePaymentMethodsViewModel.swift#L206"></span>Line should be 110 characters or less: currently 122 characters<br />
<code>line_length</code> <code>PledgePaymentMethodsViewModel.swift:206</code></td>
    </tr>
  </tbody>
</table>

Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -167,7 +167,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
     let pledgePaymentMethodsValue = PledgePaymentMethodsValue(
       user: value.user,
       project: value.project,
-      applePayCapable: PKPaymentAuthorizationViewController.applePayCapable(for: value.project)
+      applePayDevice: PKPaymentAuthorizationViewController.applePayDevice()
```
Scollaco: By usin `applePayDevice`, one can assume that we are returning a device with ApplePay, while we are returning a `Bool`. Maybe `deviceIsApplePayCapable` can be more self explanatory? 🤔

PR #964, opened 2019-11-22
[NT-620] Skip danger for external contributors

justinswart:
# 📲 What

Does not run Danger on PRs from external contributors.

# 🤔 Why

We received an external contribution recently and the CircleCI workflow that runs Danger is failing because it does not have access to `DANGER_GITHUB_API_TOKEN`.

# 🛠 How

- Moved the command that we previously used to execute Danger into a bash script and updated it so that it will not run if this environment variable is not available.
  - This is the solution described in Danger's docs for not exposing environment vars - https://danger.systems/guides/troubleshooting.html#i-want-to-only-run-danger-for-internal-branches
- CircleCI also recommends [explicitly disabling](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests) passing of secrets to builds that run on forks.

# 👀 See

image

# ✅ Acceptance criteria

- [ ] Danger still runs on our own branches.
- [ ] Once this is merged into #949, all CircleCI workflows should pass.

### DIFF
--- a/bin/danger.sh
+++ b/bin/danger.sh
@@ -1,3 +1,7 @@
 #!/bin/sh
 
-[ ! -z $DANGER_GITHUB_API_TOKEN ] && bundle exec danger --verbose || echo "Skipping Danger for External Contributor"
+if [ ! -z $DANGER_GITHUB_API_TOKEN ]; then
+  bundle exec danger --verbose;
+else
+  echo "Skipping Danger for External Contributor";
+fi
--- a/bin/danger.sh
+++ b/bin/danger.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+[ ! -z $DANGER_GITHUB_API_TOKEN ] && bundle exec danger --verbose || echo "Skipping Danger for External Contributor"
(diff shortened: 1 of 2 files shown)


bin/danger.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+[ ! -z $DANGER_GITHUB_API_TOKEN ] && bundle exec danger --verbose || echo "Skipping Danger for External Contributor"
```
ifbarrera: Would this print "Skipping Danger For External Contributor" if `bundle exec danger --verbose` fails for any reason? 🤔 My bash scripting skills are very weak I admit.

Scollaco: They don't say much about it on their website, but I _suspect_ that it means "If you don't have the token `and` execute this command, print that"...

ifbarrera: I guess I'm more wondering about the case where the token is present, but `bundle exec danger` returns a `false` value for some reason.

justinswart: Good catch @ifbarrera, I've updated the script to fail under normal circumstances with:

image

Otherwise, yes - it would just say "Skipping Danger For External Contributor" for any failure reason.

PR #965, opened 2019-11-22
[NT-593] Editorial Project Navigation

justinswart:
# 📲 What

Fills in the navigation from `DiscoveryEditorialCell` -> `DiscoverPageViewController` -> `EditorialProjectsViewController` -> `DiscoverPageViewController` -> `ProjectNavigatorViewController` and ensures that the correct `RefTag` makes it through to the end.

# 🤔 Why

Continuation of the works for our Go Rewardless editorial campaign.

# 🛠 How

- Updated delegate methods to pass `DiscoveryParams.TagID` value.
- Inferred the `RefTag` from `DiscoveryParams` as it had been done before.
- Updated tests.

# ✅ Acceptance criteria

- [x] Tap on the Go Rewardless Editorial cell on Discovery, select a project from the list, using a breakpoint or Charles, verify that you're seeing the correct `RefTag`.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
@@ -210,8 +210,8 @@ internal final class DiscoveryPageViewController: UITableViewController {
 
     self.viewModel.outputs.goToEditorialProjectList
       .observeForControllerAction()
-      .observeValues { [weak self] tag, refTag in
-        self?.goToEditorialProjectList(using: tag, refTag: refTag)
+      .observeValues { [weak self] tagId in
+        self?.goToEditorialProjectList(using: tagId)
       }
   }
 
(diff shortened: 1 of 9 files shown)


Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -424,6 +424,10 @@ private func saveSeen(activities: [Activity]) {
 }
 
 private func refTag(fromParams params: DiscoveryParams, project _: Project) -> RefTag {
+  if params.tagId != nil, params.tagId == .goRewardless {
```
justinswart: It seems `RefTag`s are inferred in this function. I realize that `Project` is not needed but I am still trying to change as little as possible 😅

ifbarrera: Any reason you've chosen to hard-code the `tagId` type here? Couldn't you just pass in the `tagId` to the `.projectCollection` refTag?

```
if let tagId = params.tagId {
   return .projectCollection(tagId)
}
```

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -210,8 +210,8 @@ internal final class DiscoveryPageViewController: UITableViewController {
 
     self.viewModel.outputs.goToEditorialProjectList
       .observeForControllerAction()
-      .observeValues { [weak self] tag, refTag in
-        self?.goToEditorialProjectList(using: tag, refTag: refTag)
+      .observeValues { [weak self] tagId in
+        self?.goToEditorialProjectList(using: tagId)
```
ifbarrera: We have a tracking event for tapping the editorial header that'll need the `refTag` passed in 🤔 Do you think we could just generate it dynamically by doing something like:

```
 AppEnvironment.current.koala.trackEditorialHeaderTapped(refTag: RefTag.projectCollection(.goRewardless))
```

justinswart: Yeah I think that should be fine?

PR #966, opened 2019-11-23
[NT-594] Editorial Modal Header

justinswart:
# 📲 What

Continuation of #961, this adds in the header details. Please note TODOs below.

# 🛠 How

Added the views and styles similar to what was added in #962.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type (* does something weird to the `contentInset` when changing with the inspector, but if the view loads at a particular text size then it's fine from there on).

# ✅ Acceptance criteria

Put the following at the bottom of `AppDelegate`'s `application(_:didFinishLaunchingWithOptions:)` (unless #965 has merged):

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
  let vc = EditorialProjectsViewController.instantiate()

  self.window?.rootViewController?.present(vc, animated: true)
}
```

- [ ] Confirm that header looks [correct](https://www.figma.com/file/lgmP2ZzXj5sj2JQuPVqTyD/GoRewardless?node-id=0%3A1).

# ⏰ TODO

- [ ] This doesn't look quite right on iPad just yet, we're figuring it out.
- [ ] Going to add the "These projects can use your support" cell in a separate PR (it appears just below the header).
- [ ] Also in a separate PR will add a slight transform to the image as well as change the `closeButton` and status bar to be a darker colour when the list is scrolled up.


Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -68,7 +83,21 @@ public final class EditorialProjectsViewController: UIViewController {
       |> \.backgroundColor .~ UIColor.white
 
     _ = self.headerView
-      |> \.backgroundColor .~ UIColor.hex(0x00007D)
+      |> UIView.lens.layoutMargins %~~ { _, view in
```
justinswart: I added this in to made the paragraph narrower on iPad, but we still need to think about what to do with the image. This approach worked for the cell but with the edge-to-edge header it looks strange. We might just have to make the header _look_ like the cell and give it rounded corners?

Library/ViewModels/EditorialProjectsViewModel.swift
```diff
@@ -51,7 +56,21 @@ public class EditorialProjectsViewModel: EditorialProjectsViewModelType,
...
+
+private func editorialImageName(for tagId: DiscoveryParams.TagID) -> String {
+  switch tagId {
+  case .goRewardless: return "go-rewardless-home"
+  }
+}
+
+private func editorialTitleLabelText(for tagId: DiscoveryParams.TagID) -> String {
+  switch tagId {
+  case .goRewardless: return Strings.This_holiday_season_support_a_project_for_no_reward()
```
justinswart: This _may_ be the wrong string: https://github.com/kickstarter/android-oss/pull/683#discussion_r349847955

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -200,7 +200,7 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
       isRefreshing.map { ($0, true) },
       projectsLoadingNoRefresh.map { ($0, false) },
       self.viewWillAppearProperty.signal.take(first: 1).mapConst((true, false))
-        .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+        .ksr_delay(.nanoseconds(0), on: AppEnvironment.current.scheduler)
```
justinswart: Decided to make this explicitly `.nanoseconds(0)` as it shouldn't be tied to the `apiDelayInterval` it's really just to put this emission on the next run loop.

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -200,7 +200,6 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
       isRefreshing.map { ($0, true) },
       projectsLoadingNoRefresh.map { ($0, false) },
       self.viewWillAppearProperty.signal.take(first: 1).mapConst((true, false))
-        .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
```
justinswart: Decided to go with `DispatchQueue.main.async { ... }` in the VC instead. It was the only way to get the snapshots to reliably record correctly 🤷‍♂.

Library/ViewModels/DiscoveryPageViewModelTests.swift
```diff
@@ -7,7 +7,7 @@ import UIKit
 import XCTest
 
 internal final class DiscoveryPageViewModelTests: TestCase {
-  fileprivate var vm: DiscoveryPageViewModelType!
+  fileprivate var vm: DiscoveryPageViewModelType = DiscoveryPageViewModel()
```
justinswart: No longer needed if we're using `DispatchQueue.main.async { ... }`

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -136,3 +215,23 @@ public final class EditorialProjectsViewController: UIViewController {
...
 }
+
+// MARK: - Styles
+
+private let headerViewStyle: ViewStyle = { view in
+  view
+    |> \.clipsToBounds .~ true
+    |> \.backgroundColor .~ .ksr_trust_700
+    |> \.isAccessibilityElement .~ true
+    |> \.accessibilityTraits .~ [UIAccessibilityTraits.button]
```
ifbarrera: Don't think you need this because it's a static view 🤔

justinswart:
> Lgtm! If you're keen, go ahead and merge this and we can open a separate PR to address updates for iPad.

Cool, will open a polish PR, I have a few other tweaks too 👍

PR #967, opened 2019-11-23
[NT-615] Respect config feature flag values when showing the editorial header

ifbarrera:
# 📲 What

- editorial header should now respond to config changes
- editorial header is now only shown on the default `DiscoveryParams` filter value

# 🤔 Why

To make it easier to test feature flag changes on the `go-rewardless` feature.

# 🛠 How

- adds debouncing to the signal that drives the configuration of the `showEditorialHeader` output, whereby the signal will wait for 1 second before it emits either the cached config feature flag value, or the updated config feature flag value fetched from the server
- converts `DiscoveryEditorialCellValue` to a struct for out-of-the-box equatable conformance
- way more test cases

# 👀 See

ZZmRpW8kKj

# ♿️ Accessibility 

N/Z

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] Turning the `Go rewardless` feature flag ON in the debug feature flag menu shows the editorial header on the default category.
- [x] Turning the `Go rewardless` feature flag OFF in the debug feature flag menu hides the editorial header on the default category.
- [x] The `Go rewardless` header should only show on the default category: if you are logged in and have `Recommendations` turned ON, you should *only* see the editorial header on the `Recommended` category. All other categories should not show the header.
- [x] The `Go rewardless` header should only show on the default category: if you're logged out, and have the `Go rewardless` feature flag turned on, you should only see the editorial header on the "All projects" (default) category. All other categories should not show the header.


Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -305,25 +308,53 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
+        sort == .magic
+      }
+      .map { args -> Bool in
+        let (sort, filterParams) = args
+
+        return sort == .magic && filterParams == DiscoveryViewModel.initialParams()
+      }
+
+    let updatedConfig = self.configUpdatedProperty.signal.skipNil()
+      .logEvents(identifier: "***updated config***")
```
ifbarrera: I'll remove these once my testing is complete 👍

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -305,25 +308,53 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
+      .filter { sort, _ in
+        sort == .magic
+      }
+      .map { args -> Bool in
+        let (sort, filterParams) = args
+
+        return sort == .magic && filterParams == DiscoveryViewModel.initialParams()
+      }
+
+    let updatedConfig = self.configUpdatedProperty.signal.skipNil()
```
justinswart: Would something like this help to ensure a config is read when this view appears? 🤔 

```swift
let updatedConfig = self.configUpdatedProperty.signal.skipNil()
let currentConfig = self.viewWillAppearProperty.signal.map { _ in AppEnvironment.current.config }
  .skipNil()
  .take(until: updatedConfig.ignoreValues())

let config = Signal.merge(
  updatedConfig,
  currentConfig
)
```

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -62,6 +62,9 @@ public protocol AppDelegateViewModelInputs {
   /// Call when the app delegate gets notice of a successful notification registration.
   func didRegisterForRemoteNotifications(withDeviceTokenData data: Data)
 
+  /// Call when the config has been updated the AppEnvironment
+  func didUpdateConfig(_ config: Config?)
```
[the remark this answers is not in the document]
ifbarrera: Hmmm no good reason I think 😅

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -4,6 +4,9 @@ import ReactiveExtensions
 import ReactiveSwift
 
 public protocol DiscoveryPageViewModelInputs {
+  /// Call when the Config has been updated in the AppEnvironment
+  func configUpdated(config: Config?)
```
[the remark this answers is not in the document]
ifbarrera: I think they're a little different. In `AppDelegate` we're responding to a side effect (`AppEnvironment.updateConfig`) that's actually happening in the `AppDelegate` and informing the VM that the side effect has happened - `didUpdateConfig`.

In the case of `DiscoveryPageViewController`, (and all future observers for the `.ksr_configUpdated` notification), the VC is letting the VM know that somewhere, the config has been updated, not that *it* specifically (the VC) updated the config: hence `configUpdated`. I know it's super nuanced, but I'd say all future observers of the `ksr_configUpdated` notification should use the input name `configUpdated`, but it's ok that the AppDelegate's is slightly different.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -4,14 +4,18 @@ import Prelude
 import UIKit
 
 internal final class DiscoveryPageViewController: UITableViewController {
+  fileprivate let viewModel: DiscoveryPageViewModelType = DiscoveryPageViewModel()
+  fileprivate let shareViewModel: ShareViewModelType = ShareViewModel()
+
+  // MARK: - Properties
+
+  private var configUpdatedObserver: Any?
```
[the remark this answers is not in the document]
ifbarrera: Ah yes thank you.

Library/ViewModels/DiscoveryPageViewModel.swift
ifbarrera: I added debouncing, lmk what you think.

justinswart: If the aim is to ensure that this emits _after_ the `UITableView` has reloaded I just wonder if it wouldn't be better to have an input informing us that it's reloaded instead of arbitrarily debouncing this emission 🤔

ifbarrera: Oh so actually that's not the issue here - the issue is that there's sometimes a race condition between the config updating and the VC being initialized, whereby if the `ksr_configUpdated` notification fires _before_ the VC has been recreated (say after login or logout) then this signal doesn't emit. However, we also don't want it to always emit twice - once with the cached value, and again with the updated value. The debounce attempts to solve this by "waiting" for 1 second for the updated config notification: if the notification arrives, then only the updated config is taken to determine the `showEditorialHeader` value. If we miss the notification, then the signal will emit with the cached value.

PR #968, opened 2019-11-25
[NT-545] Return to project page after backing

Scollaco:
# 📲 What 
- Changes the way we present the Thanks page (modally instead of using a push animation).
- Brings users back to the ProjectPage once they back a project.
**Note:** The Thanks page is being dismissed without animation due a workaround that needed to be done in order to bring the ProjectPage back to the stack. In a conversation with @jamielynnroth and @colleenmacd , we decided that this could work like that for now, since changes will be made on the ThanksPage.

# 🤔 Why
- The transition changed because the animation was inconsistent, where the ViewController was being presented with a push animation but being dismissed with a modal animation.
- To improve UX, user can now see right away the Project Page updated with the status of their backing.

[JIRA ticket](https://dripsprint.atlassian.net/browse/NT-545)

# 🛠 How
- Instead of calling the `dismiss` function from the ProjectPage, the function is now handled by the parent. This was necessary because the view hierarchy can change slightly. 
   . Before, the ProjectPage could only be showed from the Thanks Page when the user tapped in a recommended project. Now, it will be also presented once the `X` button is tapped. 
- To present the ThanksPage we now use `navigationController.present(_:animated:)` and now `navigationController.pushViewController(_:animated:)`

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| thanks-page-before | thanks_page |

# ✅ Acceptance criteria

- [ ] Back a project and on the Thanks Page, tap the `X` button. The ThanksPage should be dismissed (with no animation) and you should be brought to the Project Page.
- [ ] The CTA button of the ProjectPage should be updated according to the current status of the backing (e.g. `Manage`)


Kickstarter-iOS/Views/Controllers/ThanksViewController.swift
```diff
@@ -107,12 +107,6 @@ internal final class ThanksViewController: UIViewController, UITableViewDelegate
 
     self.backedLabel.rac.attributedText = self.viewModel.outputs.backedProjectText
 
-    self.viewModel.outputs.dismissToRootViewController
-      .observeForControllerAction()
-      .observeValues { [weak self] in
-        self?.view.window?.rootViewController?.dismiss(animated: true, completion: nil)
```
justinswart: Hey I'm still not entirely sure what the problem was and why the workaround was required. I found that by changing this line to `self?.dismiss(animated: true)` it achieves what we want? Let me know if I'm missing something!

PR #970, opened 2019-11-26
[NT-621] Editorial header tracking

ifbarrera:
#📲 What

Adds required tracking to the editorial header feature.

# 🤔 Why

So we can track how this feature performs.

# 🛠 How

- adds `Editorial Card Clicked` tracking event when the editorial header on the Discovery screen is tapped
- cleans up our usage of the `Config` in `Koala`
- added missing tests for `current_variant` default properties
- added `enabled_feature_flags` to the `defaultProperties` for better tracking of features controlled via feature flag

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] In `Edit Schemes` turn on the `KOALA_TRACKING` argument, then run the app and toggle the "Go rewardless" feature flag to _enabled_. Then, from the discovery screen, tap the "Go rewardless" header -- you should see a Koala tracking event in the console `[Koala Track]: Editorial Card Clicked`
- [x] for all Koala tracking events, you should now see a default property `enabled_feature_flags` with an array of all enabled feature flags

# ⏰ TODO

- [ ] wait for #965 to merge and resolve conflicts

### DIFF
--- a/Library/Koala/Koala.swift
+++ b/Library/Koala/Koala.swift
@@ -491,7 +491,6 @@ public final class Koala {
    */
   public func trackDiscovery(params: DiscoveryParams, page: Int) {
     var props = properties(params: params).withAllValuesFrom(["page": page])
-    props["current_variants"] = AppEnvironment.current.config?.abExperimentsArray
 
     self.track(event: "Loaded Discovery Results", properties: props)
 
@@ -594,6 +593,13 @@ public final class Koala {
    */
   public func trackDiscoveryPullToRefresh() {
     self.track(event: "Triggered Refresh")
+  }
+
+  /**
+   Call when the user taps the editorial header at the top of Discovery
+ */
+  public func trackEditorialHeaderTapped(refTag: RefTag) {
+    self.track(event: "Editorial Card Clicked", properties: ["refTag": refTag.stringTag])
   }
 
   // MARK: - Checkout Events
@@ -1366,7 +1372,6 @@ public final class Koala {
     var props = properties(project: project, loggedInUser: self.loggedInUser)
     props["ref_tag"] = refTag?.stringTag
     props["referrer_credit"] = cookieRefTag?.stringTag
-    props["current_variants"] = AppEnvironment.current.config?.abExperimentsArray
 
     // Deprecated event
     self.track(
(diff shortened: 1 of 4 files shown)


Library/Koala/Koala.swift
```diff
@@ -491,7 +491,6 @@ public final class Koala {
    */
   public func trackDiscovery(params: DiscoveryParams, page: Int) {
     var props = properties(params: params).withAllValuesFrom(["page": page])
-    props["current_variants"] = AppEnvironment.current.config?.abExperimentsArray
```
ifbarrera: ` current_variants` is already being sent as part of the `defaultProperties`, so this was redundant.

Library/Koala/Koala.swift
```diff
@@ -2043,10 +2048,16 @@ public final class Koala {
...
+    let enabledFeatureFlags = self.config?.features
+      .filter { _, v in v == true }
+      .keys
+      .sorted()
+
     props["manufacturer"] = "Apple"
     props["app_version"] = self.bundle.infoDictionary?["CFBundleVersion"]
     props["app_release"] = self.bundle.infoDictionary?["CFBundleShortVersionString"]
-    props["current_variants"] = AppEnvironment.current.config?.abExperimentsArray
+    props["current_variants"] = self.config?.abExperimentsArray.sorted()
```
ifbarrera: Whenever values from the config are accessed in other parts of this class, the object's reference of `config` was used. Grabbing the `config` from the `AppEnvironment` was a deviation from this pattern, so I updated it to be in-line with how the `config` is accessed elsewhere in this class. Please note that anytime the config is updated, it is also updated in the `Koala` instance stored in `AppEnvironment` so we don't have to worry about out-of-date config values: https://github.com/kickstarter/ios-oss/blob/master/Library/AppEnvironment.swift#L68

Library/Koala/Koala.swift
```diff
@@ -2043,10 +2048,16 @@ public final class Koala {
   private func defaultProperties() -> [String: Any] {
     var props: [String: Any] = [:]
 
+    let enabledFeatureFlags = self.config?.features
+      .filter { key, value in key.starts(with: "ios_") && value }
+      .keys
+      .sorted()
```
Scollaco: Why does the keys need to be sorted here? 🤔

Library/Koala/Koala.swift
```diff
@@ -2043,10 +2048,16 @@ public final class Koala {
...
+
     props["manufacturer"] = "Apple"
     props["app_version"] = self.bundle.infoDictionary?["CFBundleVersion"]
     props["app_release"] = self.bundle.infoDictionary?["CFBundleShortVersionString"]
-    props["current_variants"] = AppEnvironment.current.config?.abExperimentsArray
+    props["current_variants"] = self.config?.abExperimentsArray.sorted() ?? []
```
Scollaco: And here...

ifbarrera: I originally just had it as an optional but thought it might be nicer on the consuming side to see the property as empty rather than not see it at all 🤔

Scollaco: Hm I see what you mean... For other optional properties we don't send anything, though. But this is definitely not a blocker. :)

PR #971, opened 2019-11-27
[NT-594] Editorial modal header polish

justinswart:
# 📲 What

Adds some final touches to the editorial modal header to add some interaction and to ensure that the close button and status bar are always visible.

**Note:** TODOs 😬 

# 🤔 Why

- The transforms and text alpha just add a bit of interaction to the header which seemed fun.
- I found that the status bar and close button would get lost against the feed's white background so the quickest fix for now was to flip this to black. A longer term fix would be to give the close button a semi-transparent background.

# 🛠 How

- Apply transforms to the image and label on scrolling up and down.
- Fade the text as we scroll up.
- Flip the status bar and close button colour from white to black once we scroll beyond `(0,0)`.

# 👀 See

Had to break it up into a few GIFs because they kept going over 10mb, try it out on sim/device.

| Before 🐛 | Status bar and Close button | Image transform |
| --- | --- | --- |
|  |  |  |

# ✅ Acceptance criteria

- [ ] Run the app, panning up and down should do as described (close button and status bar are black when the cells have moved beyond the top of the screen). There is a slight transform and alpha adjustment applies to the image and label as you scroll over the header.
- [ ] Modal looks correct on iPad (see TODO).

# ⏰ TODO

- [x] Get larger image and adjust header for iPad.
- [x] Add cell above projects with "These projects could use your support" text.
- [x] Hide login cell when editorial content in shown.


Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -11,6 +11,7 @@ internal final class DiscoveryPageViewController: UITableViewController {
 
   private var configUpdatedObserver: Any?
   private var currentEnvironmentChangedObserver: Any?
+  public var contentOffsetChanged: ((CGPoint) -> ())?
```
justinswart: Feel like we can use more closures, less delegates 🤔 I'll understand if there's resistance to this deviation at this stage and can probably update this to a delegate.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -59,7 +63,7 @@ public final class EditorialProjectsViewController: UIViewController {
     let currentTableViewInsets = self.discoveryPageViewController.tableView.contentInset
 
     self.discoveryPageViewController.tableView.contentInset = currentTableViewInsets
-      |> UIEdgeInsets.lens.top .~ self.headerView.frame.height
+      |> UIEdgeInsets.lens.top .~ (self.headerView.frame.height - self.view.safeAreaInsets.top)
```
justinswart: 🤷‍♂

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -217,6 +239,18 @@ public final class EditorialProjectsViewController: UIViewController {
   @objc private func closeButtonTapped() {
     self.viewModel.inputs.closeButtonTapped()
   }
+
+  private func applyViewTransforms(withY y: CGFloat) {
```
justinswart: These values are somewhat arbitrary which is why I just did the one output from the VM and I'm not really testing them... Maybe I'm just lazy, I'm just not sure what we'd gain from testing them necessarily. But if there's a desire to do so I can add tests.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -137,6 +140,25 @@ public final class EditorialProjectsViewController: UIViewController {
...
+      }
+
+    self.viewModel.outputs.setNeedsStatusBarAppearanceUpdate
+      .observeForUI()
+      .observeValues { [weak self] in
+        self?.setNeedsStatusBarAppearanceUpdate()
+      }
+
+    self.viewModel.outputs.applyViewTransformsWithY
+      .observeForControllerAction()
```
justinswart: Snapshot tests simply wouldn't pass without this. This VC is doing something really strange with the refreshControl and it's breaking a number of things, I think.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
ifbarrera: Yeah I'd say we probably shouldn't break from convention for this one case 😅

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -98,7 +102,6 @@ public final class EditorialProjectsViewController: UIViewController {
 
     _ = self.closeButton
       |> UIButton.lens.title(for: .normal) .~ nil
-      |> UIButton.lens.image(for: .normal) .~ image(named: "icon--cross", tintColor: .white)
```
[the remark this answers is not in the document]
justinswart: Seems not, unfortunately. The `UIImage` is created with the tint color.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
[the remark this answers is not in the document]
justinswart: We don't need the `x` of the point though, so how about adding the word _offset_?:

```swift
func applyViewTransforms(withYOffset y: CGFloat) { ... }
```

Library/ViewModels/EditorialProjectsViewModel.swift
```diff
@@ -37,6 +42,20 @@ public class EditorialProjectsViewModel: EditorialProjectsViewModelType,
     self.titleLabelText = tagId.map(editorialTitleLabelText)
 
     self.dismiss = self.closeButtonTappedProperty.signal
+
+    self.preferredStatusBarStyleProperty <~ self.contentOffsetChangedProperty.signal
```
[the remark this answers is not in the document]
justinswart: Yeah that's cool, thanks!

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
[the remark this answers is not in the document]
justinswart: Apologies, does work!

PR #975, opened 2019-11-29
[NT-642] Migrate SurveyResponseViewController to WKWebView

justinswart:
# 📲 What

Migrates `SurveyResponseViewController` to use `WKWebView`.

# 🤔 Why

`SurveyResponseViewController` is the last of our controllers (other than the deprecated checkout controllers) that still makes use of `UIWebView`. In order for us to move to iOS 13 we have to migrate this controller to use `WKWebView`.

Although this began as part of the iOS 11 deprecation work in #904, migrating this to `WKWebView` is actually not needed to deprecate iOS 11 but it will be needed for us to build against iOS 13 so I chose to split this out into a separate PR.

**Note:** Although it seems iOS 13 may still compile with `UIWebView` being linked, folks are [receiving warnings](https://stackoverflow.com/questions/57722616/itms-90809-deprecated-api-usage-apple-will-stop-accepting-submissions-of-app) when submitting builds to the app store (we haven't yet received one of these).

Response from Apple: https://forums.developer.apple.com/thread/125753#393941

# 🛠 How

_How_ is indeed the question. It turns out that there is a [bug](https://bugs.webkit.org/show_bug.cgi?id=167131) in `WKWebView` that has been around for some time which causes the `httpBody` property to be `nil` in `WKNavigationDelegate`'s `webView(_:decidePolicyFor:decisionHandler:)` callback. This breaks our existing form submission that worked with `UIWebView`.

Googling turned up various hacks and workarounds but [the one](https://github.com/lionheart/openradar-mirror/issues/1558#issuecomment-341262718) I settled on was to run a small bit of JavaScript which returns the form's input fields and data. We then prepare a new request injecting this data into the `httpBody` property and instruct our `webView` to load that request. This appears to work and although it's not the most elegant of solutions, it may have to do if we want to build against iOS 13 without rebuilding survey responses natively.

# ✅ Acceptance criteria

Create a test survey on staging (ping me on Slack for instructions). Open this survey in the app (I manually triggered a deep-link to get to it). Use Charles to inspect the raw post data being sent.

- [ ] Do the above against `master`. Observe that the survey results were received.
- [ ] Repeat against this branch, results should be the same.


Kickstarter-iOS/Views/Controllers/SurveyResponseViewController.swift
```diff
@@ -60,7 +61,15 @@ internal final class SurveyResponseViewController: DeprecatedWebViewController {
...
       .observeValues { [weak self] request in
-        self?.webView.loadRequest(request)
+        self?.webView.load(request)
+      }
+
+    self.viewModel.outputs.extractFormDataWithJavaScript
+      .observeForUI()
+      .observeValues { [weak self] js in
+        self?.webView.evaluateJavaScript(js) { result, _ in
+          self?.viewModel.inputs.didEvaluateJavaScriptWithResult(result)
```
justinswart: We run a small bit of JS in the `webView` and call the `viewModel` with the result.

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -68,33 +83,47 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
...
     self.showAlert = redirectAfterPostRequest
       .mapConst(Strings.Got_it_your_survey_response_has_been_submitted())
 
     self.title = self.viewDidLoadProperty.signal
       .mapConst(Strings.Survey())
 
+    self.extractFormDataWithJavaScript = surveyResponse
+      .takeWhen(postRequest.filter { $0.httpBody == nil })
+      .map { surveyResponse in
+        "$('#edit_survey_response_\(surveyResponse.id)').serialize()"
```
justinswart: This is the little bit of JavaScript. For a `SurveyResponse` with an `ID` of `12345` this would be:

```javascript
$('#edit_survey_response_12345').serialize()
```
It makes the following assumptions:
1. jQuery is available on the page.
1. There is a form on the page with the ID of `edit_survey_response_<survey_response_id>`.

It returns the serialized form input fields like:

```

survey_response[answers_attributes][1][survey_question_id] | 2610391
survey_response[answers_attributes][1][response] |  
survey_response[answers_attributes][2][survey_question_id] | 2610392
survey_response[answers_attributes][2][response] | small
survey_response[answers_attributes][3][survey_question_id] | 2610393
survey_response[answers_attributes][3][response] |  
survey_response[answers_attributes][4][survey_question_id] | 2610394
survey_response[answers_attributes][4][response][] |  
survey_response[answers_attributes][4][response][] | red
```

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -142,3 +174,11 @@ private func isSurvey(request: URLRequest) -> Bool {
   guard case (.project(_, .survey, _))? = Navigation.match(request) else { return false }
   return true
 }
+
+private func requestInjectedWithFormData(with request: URLRequest, result: Any?) -> URLRequest {
+  var newRequest = request
+  if let formData = result as? String {
+    newRequest.httpBody = formData.data(using: .utf8)
```
justinswart: We cast the form data from `Any?` to `String` and set it as the raw `httpBody` of our request which we then prepare with our authentication headers.

Library/ViewModels/SurveyResponseViewModel.swift
justinswart: Note there is an alternative to this `jQuery` snippet that alleviates at least one of these assumptions, but it looks like this 😬 

```swift
self.extractFormDataWithJavaScript = surveyResponse
  .takeWhen(postRequest.filter { $0.httpBody == nil })
  .map { surveyResponse in
    serializeFormDataJavaScript(withSurveyResponseId: surveyResponse.id)
  }

private func serializeFormDataJavaScript(withSurveyResponseId id: Int) -> String {
  let serialize = """
  var serialize=function(form){var serialized=[];for(var i=0;i<form.elements.length;i+=1){var field=form.elements[i];
  if(!field.name||field.disabled||field.type==='file'||field.type==='reset'||field.type==='submit'||field.type==='button'){continue}
  if(field.type==='select-multiple'){for(var n=0;n<field.options.length;n+=1){if(!field.options[n].selected){continue}
  serialized.push(encodeURIComponent(field.name)+"="+encodeURIComponent(field.options[n].value))}}
  else if((field.type!=='checkbox'&&field.type!=='radio')||field.checked)
  {serialized.push(encodeURIComponent(field.name)+"="+encodeURIComponent(field.value))}}return serialized.join('&')};
  """
  // swiftlint:enable line_length

  return "\(serialize)\nserialize(document.getElementById('edit_survey_response_\(id)'));"
}
```
This uses a minified `serialize()` function which does the same as jQuery's `serialize()` except we won't be affected by the possible (albeit unlikely) removal of jQuery from the survey response page.

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -43,21 +51,29 @@ public protocol SurveyResponseViewModelType: SurveyResponseViewModelInputs, Surv
...
         guard let url = URL(string: surveyResponse.urls.web.survey) else { return nil }
         return URLRequest(url: url)
       }
       .skipNil()
 
-    let postRequest = self.shouldStartLoadProperty.signal.skipNil()
+    let requestAndNavigationType = self.policyForNavigationActionProperty.signal.skipNil()
+      .map { action in (action.request, action.navigationType) }
+
+    let postRequest = requestAndNavigationType
```
ifbarrera: `surveyPostRequest` might be clearer naming here?

PR #980, opened 2019-12-04
[NT-669, NT-496] Shimmer Loading Extension

justinswart:
# 📲 What

Adds a protocol extension `ShimmerLoading` for the purpose of adding a shimmering loading state to views.

# 🤔 Why

- This is a continuation of the UI improvements from our native checkout work.
- A shimmer loading state hints at the shape of the view that is being loaded rather than just using a native `UIActivityIndicatorView`.

# 🛠 How

- Added `ShimmerLoading` protocol extension.
  - Views that conform to this protocol are extended with `startLoading()` and `stopLoading()` functions.
  - All `UIView`s are given a boolean property `shimmersWhenLoading` which, when `true`, means that the shimmering gradient layer is applied to them if their parent conforms to `ShimmerLoading` and `startLoading()` is called.
  - **Note:** I would like to think of a way to do this without extending _all_ `UIView`s to have that boolean.
- Created a `PledgeShippingLocationShimmerLoadingView` example implementation.
  - Technically we could just conform the `rootStackView` of `PledgeShippingLocationViewController` to `ShimmerLoading` and tell it which subviews to apply the shimmer to. While this would work, the shape of the views weren't quite that of the designs so I chose to create a separate view instead.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

To have the view appear for longer add this to line `56` of `PledgeShippingLocationViewModel`:
```swift
.ksr_debounce(.seconds(2), on: AppEnvironment.current.scheduler)
```
We will probably want to make this permanent so that the shimmer is not too brief to be jarring.

- [ ] Navigate to the pledge view of a reward that has shipping. The shimmer loading views should appear in the position of the shipping location section.


Kickstarter-iOS/Views/Controllers/PledgeShippingLocationViewController.swift
```diff
@@ -117,6 +121,13 @@ final class PledgeShippingLocationViewController: UIViewController {
         )
       }
 
+    self.viewModel.outputs.isLoading
```
justinswart: This output was still dangling but I'll add separate outputs for hiding and showing these.

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
```diff
@@ -0,0 +1,174 @@
...
+  static var shimmerLayers = "shimmerLayers"
+}
+
+protocol ShimmerLoading: AnyObject {
+  func startLoading()
+  func stopLoading()
+}
+
+extension UIView {
+  var shimmersWhenLoading: Bool {
```
justinswart: Would like to think of a way to not do this to _all_ `UIView`s.

justinswart: Actually a simpler way of doing this is probably just to have the conforming view return an array of views to add the shimmer to. I’ll make that change.

ifbarrera: Agree this would be nicer 👍

Kickstarter-iOS/Views/ShimmerLoadingViews/PledgeShippingLocationShimmerLoadingView.swift
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  required init?(coder _: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  override func layoutSubviews() {
+    super.layoutSubviews()
+
+    self.layoutGradientLayers()
```
ifbarrera: It's always going to be necessary to call this on the `ShimmerLoading`-conforming view, right?

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
```diff
@@ -0,0 +1,174 @@
...
+        &AssociatedKeys.shimmersWhenLoading,
+        newValue,
+        objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC
+      )
+    }
+  }
+}
+
+extension ShimmerLoading where Self: UIView {
+  private var isLoading: Bool {
```
ifbarrera: Is this private `isLoading` property really needed, or can we just pass it into `updateAnimation` directly?

```
func startLoading() {
    self.updateAnimation(isLoading: true)
}
```

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
```diff
@@ -0,0 +1,174 @@
...
+
+        gradientLayer.add(animationGroup, forKey: animation.keyPath)
+
+        view.layer.addSublayer(gradientLayer)
+        self.shimmerLayers.append(gradientLayer)
+      }
+  }
+}
+
+private func allSubViews(of view: UIView) -> [UIView] {
```
ifbarrera: You won't need this if you store `var shimmerViews: [UIView]` instead right?

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
```diff
@@ -0,0 +1,174 @@
...
+  private func updateAnimation() {
+    guard self.isLoading else {
+      self.shimmerLayers.forEach { $0.removeFromSuperlayer() }
+      return
+    }
+
+    allSubViews(of: self)
+      .filter { $0.shimmersWhenLoading }
+      .forEach { view in
+        let gradientLayer = newGradientLayer(with: view.bounds)
```
ifbarrera: Would something like `shimmerGradientLayer`, `shimmerAnimation` and `shimmerAnimationGroup` be clearer naming for these funcs?

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
ifbarrera: Also the fewer associated objects the better 😅

Kickstarter-iOS/Views/UIView+ShimmerLoading.swift
justinswart: Yip, all gone!

justinswart:
> I think it would be nice not to have to create these sort of shimmer dummy views like PledgeShippingLocationShimmerLoadingView, I wonder if we just set greaterThanOrEqualTo width and height constraints on the views we want the shimmer to appear on if that would be sufficient 🤔

I'll look into this, it might just mean having to adjust some constraints on the view when it's loading/not loading but yeah that could work!

Kickstarter-iOS/Views/ShimmerLoadingViews/PledgeShippingLocationShimmerLoadingView.swift
justinswart: Hmm, yeah because we can't override `layoutSubviews` in an extension 🤔

justinswart:
Going to merge this and look into possibly not using a dummy view separately 👍

PR #981, opened 2019-12-05
[NT-649] Errored pledges query

Scollaco:
# 📲 What
 - Adds a query to fetch errored pledges from graphQL.

# 🤔 Why
- We need this query to show a list of errored pledges on the `ActivitiesViewController` and allow users to fix them.

# 🛠 How
- Created a new model `GraphBacking`
- Added the required fields on the GraphSchema
- Added output on `ActivityViewModel` that emits the list of errored pledges
- Added tests

[JIRA ticket](https://dripsprint.atlassian.net/browse/NT-649)

# ✅ Acceptance criteria
*Testing with Charles*
- [ ]  Open the app and tap the Activity tab. A list of errored backings (if any), should be visible on the response.

*If  there's no errored pledges*
- [ ] Add a breakpoint to intercept the response once the `ActivitiesViewController` is open and change the `status` of backings to `errored`. The `erroredBackings` output on ActivitiesViewModel should emit a list with the errored pledges.


KsApi/models/GraphBackingTests.swift
```diff
@@ -0,0 +1,45 @@
...
+              "id": "UHJvamVjdC0yMDQ4MTExNDEw",
+              "name": "Cool project",
+              "slug": "/cool-project"
+            }
+          }
+        ]
+      }
+    }
+    """
+    let data = jsonString.data(using: .utf8)
```
justinswart: If we use the non-failable `Data` initializer then we don't need to force-unwrap later on:

```swift
let data = Data(jsonString.utf8)
```

KsApi/models/GraphBacking.swift
```diff
@@ -0,0 +1,22 @@
...
+  public var project: Project?
+  public var status: Status
+
+  public struct Project: Swift.Decodable, Equatable {
+    public var id: String
+    public var name: String
+    public var slug: String
+  }
+
+  public enum Status: String, CaseIterable, Swift.Decodable {
```
justinswart: I see the back-end also has a status `authentication_required` 🤔

PR #982, opened 2019-12-05
[NT-636] "Fix" button on Manage Pledge screen for errored backings

ifbarrera:
# 📲 What

Adds a "Fix" button in the payment method section of the "Manage Pledge" screen when the backing has errored due to collection issues.

# 🤔 Why

So users have the option to update their payment source when collection has failed after the project has ended.

# 🛠 How

Added a "Fix" button that only shows if the backing is in an `errored` state. If you tap the button, we push the `Change Payment Method` screen. Note that right now, the Change Payment Method screen isn't configured to show the "Fix payment method" state.

# 👀 See

Simulator Screen Shot - iPhone X - 2019-12-05 at 11 45 46

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

To test this feature, add `.mapConst(false)` to line 84 in `ManagePledgePaymentMethodViewModel.swift`:

image

- [ ] Navigate to a project that you backed that has ended. Tap "View pledge". You should see the red "Fix" button in the payment method section. If you tap this button, you should be navigated to the Change Payment Method screen.


Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -107,6 +140,10 @@ final class ManagePledgePaymentMethodView: UIView {
       )
     ])
   }
+
+  @objc private func fixButtonTapped() {
+    self.delegate?.managePledgePaymentMethodViewDidTapFixButton(self)
```
justinswart: I think you mean to call the VM input here.

Kickstarter-iOS/Views/Controllers/ManagePledgePaymentMethodView.swift
```diff
@@ -3,12 +3,22 @@ import Library
...
 final class ManagePledgePaymentMethodView: UIView {
+  weak var delegate: ManagePledgePaymentMethodViewDelegate?
+
   // MARK: - Properties
 
   private lazy var cardLabelsStackView: UIStackView = { UIStackView(frame: .zero) }()
   private lazy var cardNumberLabel: UILabel = { UILabel(frame: .zero) }()
   private lazy var expirationDateLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var fixButton: UIButton = { UIButton(type: .custom)
+    |> \.translatesAutoresizingMaskIntoConstraints .~ false
```
justinswart: Don't think we need this when we're adding directly to a stackview.

ifbarrera: We need it cuz we set a height and width in `setupConstraints`

justinswart: Oh right but it seems that simply by adding the view to a stackview it sets `translatesAutoresizingMaskIntoConstraints` to `false`. Anyway, not a big deal but if you did feel like removing it everything will still work.

PR #983, opened 2019-12-05
[NT-685] Abstract TrackingClient to be configurable

justinswart:
# 📲 What

Renames our existing `KoalaTrackingClient` to `TrackingClient` and adds `TrackingClientConfiguration` which is a struct with which we can initialize `TrackingClient` so that we can use it for more than one tracking endpoint.

**Note:** We could even create a `TrackingClient` to forward our events to something like `Crashlytics` for our breadcrumbs. Currently we do this with a callback at the time of tracking, but I think we should move that to be its own `TrackingClient`.

**Also note:** `Koala` itself will be renamed separately too. Currently it really just behaves as an interface for sending the events to the clients.

# 🤔 Why

We're going to be sending tracking events to two endpoints until we're able to fully deprecate `Koala`, this allows us to do so in a somewhat controlled manner and still keeps things testable.

# 🛠 How

- Abstracted any endpoint-specific functionality from `TrackingClient` to `TrackingClientConfiguration` either via simple properties or closures.
- Updated `Koala`'s initializer so that it can be initialized with two tracking clients.
  - `Koala` will send all tracking events to both clients, we will be doing some filtering of whitelisted events one we know what they are. I would suggest that we do this filtering in `Koala` so that we can test it through mocked clients.
- Wrote tests for these endpoint-specific parts of `TrackingClientConfiguration`, this actually gives us a bit more test coverage than before because these parts of `KoalaTrackingClient` weren't tested.

# ✅ Acceptance criteria

With the simulator set to staging, return `true` for `isKoalaTrackingEnabled` (to be renamed) in `EnvironmentVariables.swift`. Use Charles Proxy to inspect traffic.

- [ ] Observe that tracking events are sent to both the `DataLake` and `Koala` endpoint.
- [ ] Observe that both requests return `200` and are in the correct structure for each endpoint (e.g. `POST` for `Koala`, `PUT` for `DataLake`, etc.).

# ⏰ TODO

- [ ] Additional tests once we have whitelisted events.
- [ ] Additional clean-up noted in comments.


Library/TestHelpers/TestCase.swift
```diff
@@ -13,11 +13,12 @@ internal class TestCase: FBSnapshotTestCase {
   internal let cache = KSCache()
   internal let config = Config.config
   internal let cookieStorage = MockCookieStorage()
+  internal let dataLakeTrackingClient = MockTrackingClient()
   internal let dateType = MockDate.self
   internal let mainBundle = MockBundle()
   internal let reachability = MutableProperty(Reachability.wifi)
   internal let scheduler = TestScheduler(startDate: MockDate().date)
-  internal let trackingClient = MockTrackingClient()
+  internal let trackingClient = MockTrackingClient() // TODO: Rename
```
justinswart: Renaming this resulted in ~700 changes so I will do that separately.

Library/TestHelpers/TestCase.swift
```diff
@@ -54,7 +55,11 @@ internal class TestCase: FBSnapshotTestCase {
       debounceInterval: .seconds(0),
       device: MockDevice(),
       isVoiceOverRunning: { false },
-      koala: Koala(client: self.trackingClient, loggedInUser: nil),
+      koala: Koala(
+        dataLakeClient: self.dataLakeTrackingClient,
```
justinswart: There is a mock client set for this by default in `TestCase` but I think we can add tests once we whitelist events.

Library/Tracking/Koala.swift
```diff
@@ -355,15 +356,17 @@ public final class Koala {
 
   public init(
     bundle: NSBundleType = Bundle.main,
-    client: TrackingClientType,
+    dataLakeClient: TrackingClientType = TrackingClient(.dataLake), // TODO: Remove default value
```
justinswart: Also to reduce the amount of line changes in this PR I'm passing in a default value here. We can tackle this separately.

Library/Tracking/Koala.swift
```diff
@@ -2028,7 +2031,12 @@ public final class Koala {
 
     self.logEventCallback?(event, props)
 
-    self.client.track(
```
justinswart: Currently we just send all events to both clients. We will whitelist before this point.

Library/Tracking/TrackingClientConfiguration.swift
```diff
@@ -0,0 +1,132 @@
...
+  guard
+    let baseUrl = config.url(environmentType),
+    var components = URLComponents(url: baseUrl, resolvingAgainstBaseURL: false)
+  else { return nil }
+
+  let dataString = data.base64EncodedString(options: [])
+
+  if dataString.count >= 10_000 {
+    // swiftlint:disable:next line_length
+    print("\(config.identifier.emoji) [\(config.identifier) Error]: Base64 payload is longer than 10,000 characters.")
```
justinswart: This was just something we printed at this point previously, not sure how important it is to keep.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #987, opened 2019-12-12
[NT-616] Add Qualtrics SDK

justinswart:
# 📲 What

Adds the Qualtrics SDK to the app.

**Note:** Follow-up PRs will configure the SDK and present the survey.

# 🤔 Why

We intend to use Qualtrics to run some user surveys.

# 🛠 How

Added the SDK to our `Cartfile`, linked it in the project and added a line of code to test that the project compiles correctly.

# 👀 See

https://api.qualtrics.com/docs/getting-started-with-the-mobile-app-sdk-on-ios

# ✅ Acceptance criteria

- [x] Qualtrics SDK downloads successfully via `carthage bootstrap --platform iOS`.
- [x] App compiles.


Carthage-xcfilelist/app-output-files.xcfilelist
```diff
@@ -2,16 +2,13 @@
 
 $(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Alamofire.framework
 $(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/AlamofireImage.framework
-$(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/AppCenter.framework
```
justinswart: Seems we removed this from the input files list but forgot to remove it from the output files list when we added the AppCenter SDK. This shouldn't be required because we don't use AppCenter for our app store builds.

ifbarrera: Thanks for cleaning this up 👍

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -43,6 +44,9 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     UIImageView.appearance(whenContainedInInstancesOf: [UITabBar.self])
       .accessibilityIgnoresInvertColors = true
 
+    // FIXME: Replace with actual values, trigger via VM output
+    Qualtrics.shared.initialize(brandId: "BRAND ID", zoneId: "PROJECT ID", interceptId: "INTERCEPT ID")
```
ifbarrera: Do we want to wait until we implement the required code to add this?

PR #990, opened 2019-12-13
[NT-645] Clean user properties & session properties

ifbarrera:
#📲 What

Adds clean user properties and session properties, with tests.

# 🤔 Why

Part of the data cleanup project.

# 🛠 How

- cleans up `defaultProperties` and rename to `sessionProperties`
- adds `userProperties` to decouple user props from session props

Now, instead of having user properties in the `defaultProperties` function, we build them separately and join all properties together with:

```
let props = self.sessionProperties()
    .withAllValuesFrom(userProperties())
    .withAllValuesFrom(eventProperties)
```

Note that unfortunately, we're not in a good place to test some of the session properties that don't have mocks. I've made a note and filed a separate tech debt ticket to address this so we can be sure to test all session properties in the near future.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [ ] Verify that all the tests in `KoalaTests.swift` are correct

Verify that the following session properties are sent on every tracking event:
- [x] `apple_pay_capable`
- [x] `apple_pay_device`
- [x] `cellular_connection`
- [x] `client_type`
- [x] `current_variants`
- [x] `device`
- [x] `device_fingerprint`
- [x] `device_format`
- [x] `device_manufacturer`
- [x] `device_model`
- [x] `device_orientation`
- [x] `distinct_id`
- [x] `enabled_features`
- [x] `iphone_uuid`
- [x] `is_voiceover_running`
- [x] `mp_lib`
- [x] `os`
- [x] `os_version`
- [x] `time`
- [x] `app_version`
- [x] `app_release` // might re-name to app_build
- [x] `screen_width`
- [x] `user_agent`
- [x] `user_logged_in`
- [x] `wifi_connection`
- [x] `client_platform`

Verify that the following user properties are sent on every tracking event:
- [x] `user_is_admin` // only if logged in
- [x] `user_backed_projects_count` // only if logged in
- [x] `user_country`
- [x] `user_facebook_account` // only if logged in
- [x] `user_watched_projects_count` // only if logged in
- [x] `user_uid` // only if logged in

# ⏰ TODO

- [x] finalize exact native session props

### DIFF
--- a/Library/Tracking/Koala.swift
+++ b/Library/Tracking/Koala.swift
@@ -1940,7 +1940,7 @@ public final class Koala {
 
   // Private tracking method that merges in default properties.
   private func track(event: String, properties: [String: Any] = [:]) {
-    let props = self.defaultProperties()
+    let props = self.sessionProperties()
       .withAllValuesFrom(userProperties(for: self.loggedInUser, config: self.config))
       .withAllValuesFrom(properties)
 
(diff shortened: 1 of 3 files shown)


Library/Tracking/Koala.swift
```diff
@@ -1963,53 +1964,46 @@ public final class Koala {
...
+    props["client_type"] = "native"
+    props["current_variants"] = self.config?.abExperimentsArray.sorted()
+
+    props["device"] = self.device.name
     props["device_fingerprint"] = self.distinctId
+    props["device_format"] = self.deviceFormat
+    props["device_manufacturer"] = "Apple"
+    props["device_model"] = Koala.deviceModel
+    props["device_orientation"] = self.deviceOrientation
+    props["distinct_id"] = self.distinctId
```
justinswart: As discussed on Slack this, `device_fingerprint` and `iphone_uuid` all appear to be the same value.

Library/Tracking/KoalaTests.swift
```diff
@@ -25,51 +25,62 @@ final class KoalaTests: TestCase {
...
-    XCTAssertEqual(
-      bundle.infoDictionary?["CFBundleShortVersionString"] as? String,
-      properties?["app_release"] as? String
-    )
-    XCTAssertNotNil(properties?["model"])
-    XCTAssertEqual(device.systemName, properties?["os"] as? String)
-    XCTAssertEqual(device.systemVersion, properties?["os_version"] as? String)
-    XCTAssertEqual(UInt(screen.bounds.width), properties?["screen_width"] as? UInt)
-    XCTAssertEqual(UInt(screen.bounds.height), properties?["screen_height"] as? UInt)
+    // FIXME: The following properties cannot be reliably tested
```
justinswart: Best to keep the `FIXME:` here or track in a Jira ticket?

ifbarrera: It's in both, I can remove it from here if you think it's too noisy.

Library/Tracking/Koala.swift
ifbarrera: I'll bring it up with insights and remove one property in a follow-up PR if they're ok with it.

justinswart: Not remove two props?

ifbarrera: Sorry yes. Consolidate to one prop.

PR #992, opened 2019-12-13
[NT-696] Display Qualtrics Survey via Local Notification

justinswart:
# 📲 What

Adds configuration of the Qualtrics SDK, targeting logic evaluation and handling of the local notification to display a Qualtrics Survey in the app.

# 🤔 Why

This is a continuation of #987, we are adding surveys to the app to improve our collection of user feedback data.

# 🛠 How

- Configured the SDK with the `brandId`, `zoneId` and `interceptId` values, as well as a custom `Properties` object with key/value pairs.
  - Currently we only send the `bundle_id` along as a custom property to assist with filtering the targeting logic. This is likely to change.
- Ask the Qualtrics SDK to evaluate targeting logic and if it returned `true` from `TargetingResult.passed()` we instruct it to display the survey.
- Because our Intercept is configured to display a local notification _creative_, when we then instruct Qualtrics to display the survey, it triggers a local notification.
- Tapping this notification activates the survey modal.

# 👀 See

#### Qualtrics docs

[SDK docs](https://api.qualtrics.com/docs/getting-started-with-the-mobile-app-sdk-on-ios)
[Page Views & Impressions](https://api.qualtrics.com/docs/page-view-and-impressions-with-the-mobile-sdk)

# ✅ Acceptance criteria

- [ ] Run the app, 2 seconds after starting up a local notification should appear. Tapping the notification should launch the Survey.


Kickstarter-iOS/Library/QualtricsTypes.swift
```diff
@@ -0,0 +1,30 @@
...
+  public let stringProperties: [String: String]
+}
+
+enum QualtricsIntercept {
+  case survey
+
+  var interceptId: String {
+    switch AppEnvironment.current.environmentType {
+    case .production:
+      return "SI_eUKtARDK3785TQF"
```
justinswart: These intercept IDs are not sensitive because they can easily be observed on the web by inspecting traffic between the JS SDK and Qualtrics API.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -15,7 +15,6 @@ public enum NotificationAuthorizationStatus {
   case authorized
   case denied
   case notDetermined
-  @available(iOS 12, *)
```
justinswart: This should have been removed in our iOS 11 deprecation PR.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -600,6 +614,28 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
     .filter(isTrue)
     .mapConst(0)
+
+    self.configureQualtrics = self.applicationLaunchOptionsProperty.signal.map { _ in
+      return .init(
+        brandId: Secrets.Qualtrics.brandId,
+        zoneId: Secrets.Qualtrics.zoneId,
+        interceptId: QualtricsIntercept.survey.interceptId,
+        stringProperties: [
+          "bundle_id": AppEnvironment.current.mainBundle.bundleIdentifier.coalesceWith("")
```
justinswart: Currently we only sent the `bundle_id` for filtering in the Intercept's configuration. We will likely be adding values here.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -600,6 +614,28 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
+        ]
+      )
+    }
+
+    self.evaluateQualtricsTargetingLogic = self.qualtricsInitializedWithResultProperty.signal
+      .skipNil()
+      .filter { $0.passed() }
+      .ksr_delay(.seconds(2), on: AppEnvironment.current.scheduler)
```
justinswart: Arbitrarily delaying the display by 2 seconds for now - will probably change.

Kickstarter-iOS/Library/QualtricsTypes.swift
justinswart: Also - this is the one value that we expect to change which is also why it isn't in Secrets. Later we may even obtain this dynamically and configure our surveys a bit differently. It seems a little strange to me that the `interceptId` forms part of the configuration data for the singleton but 🤷‍♂ that appears to be how it works.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -11,12 +11,15 @@ import UserNotifications
...
-  let vm: AppDelegateViewModelType = AppDelegateViewModel()
+  var vm: AppDelegateViewModelType!
 
   fileprivate let applicationIconBadgeNumber = TestObserver<Int, Never>()
   fileprivate let configureAppCenterWithData = TestObserver<AppCenterConfigData, Never>()
   fileprivate let configureFabric = TestObserver<(), Never>()
+  fileprivate let configureQualtrics = TestObserver<QualtricsConfigData, Never>()
   fileprivate let didAcceptReceivingRemoteNotifications = TestObserver<(), Never>()
+  fileprivate let displayQualtricsSurvey = TestObserver<(), Never>()
+  private let evaluateQualtricsTargetingLogic = TestObserver<(), Never>()
```
ifbarrera: Can we make this `fileprivate` so it's consistent with the rest of the observers in this test class?

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -11,12 +11,15 @@ import UserNotifications
 import XCTest
 
 final class AppDelegateViewModelTests: TestCase {
-  let vm: AppDelegateViewModelType = AppDelegateViewModel()
+  var vm: AppDelegateViewModelType!
```
ifbarrera: Can we make this `fileprivate` too? And why the need to re-initialize the VM on `setup`?

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
justinswart: Renamed everything to `private`, we should only use `fileprivate` when we need to.

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
justinswart: Needed to do this in order to ensure that the scheduler is replaced in the environment before the test runs. We've had to do this in a few VMs. We can probably undo this depending how we change the survey display logic.

justinswart:
> I think we might want to consider moving when we call evaluateTargetingLogic to some time after the user has spent time in the session, otherwise might be limiting the criteria we can select on.

Agreed, not sure what the desired display logic is just yet!

PR #994, opened 2019-12-16
[NT-703] Pledge CTA A/B Experiment

cdolm92:
# 📲 What

Using Optimizely we have set up an A/B test to modify the copy in the "Back this Project" button to read "See Rewards" 

# 🤔 Why

We want to use Optimizely to start setting up experiments quickly and easily. This will be the first of many.
 
# 🛠 How

We used the `variation(for experiment: OptimizelyExperiment.Key ) -> String` which returns the variation the user is in. We are calling this in `pledgeCTA(project: Project, backing: Backing) -> PledgeStateCTAType` so that we get the `.seeRewards` , `PledgeStateCTAType` if the user is in the experimental variation/group.

# 👀 See
If user is in the experiment group they see this screen:

Simulator Screen Shot - iPhone Xs - 2019-12-16 at 14 26 29

# ✅ Acceptance criteria

- [x]  Search Optimizely on the log and check to see which bucket user is in. If they're in the experimental variation she should see `See Rewards` otherwise they should see `Back this project`.


Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -131,11 +131,22 @@ public final class PledgeCTAContainerViewViewModel: PledgeCTAContainerViewViewMo
 // MARK: - Functions
 
 private func pledgeCTA(project: Project, backing: Backing?) -> PledgeStateCTAType {
+  let uuid = (AppEnvironment.current.device.identifierForVendor
+    ?? UUID()).uuidString // TODO: move to environment
```
justinswart: Move to environment in order to mock a UUID string?

If you just want to test whether a UUID string was returned you can do:
```swift
XCTAssertNotNil(UUID(uuidString: uuidString))
```
It'll be `nil` if it's not a valid UUID string.

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -131,11 +131,22 @@ public final class PledgeCTAContainerViewViewModel: PledgeCTAContainerViewViewMo
...
+  let optimizelyVariant = AppEnvironment.current.optimizelyClient?
+    .variant(for: OptimizelyExperiment.Key.pledgeCTACopy, userId: uuid)
+
   guard let projectBacking = backing, project.personalization.isBacking == .some(true) else {
     if currentUserIsCreator(of: project) {
       return PledgeStateCTAType.viewYourRewards
     }
 
+    if let variant = optimizelyVariant,
```
justinswart: This can be slightly simplified as:
```swift
if OptimizelyExperiment.Variant.experimental == optimizelyVariant, project.state == .live {
  return PledgeStateCTAType.seeRewards
}
```

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
ifbarrera: Hmm maybe a helper function would be better here. I'll add one.

PR #995, opened 2019-12-16
[NT-707] Naming changes and white-listing clean events

ifbarrera:
# 📲 What

Updates some of the naming of certain props to reflect latest changes from insights team. Also adds white-listing on "clean" events, so that only white-listed events are sent to the data lake.

# 🤔 Why

More event cleanup and ensuring that only clean events go to the data lake.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

Testing white-listing:
- [x] Ensure that `KOALA_TRACKING_ENABLED` is set to `true`, then run the app. On app launch, you should see the event `App Open` track on the Koala client. You should *not* see it track on the data lake client.
- [x] Navigate to a project page. You should the event `Project Page Viewed` track on **both** the Koala and Lake clients

Testing discovery props changes:
- [x] navigate to a subcategory on discovery (ex. Ceramics). On the `Explore Page Viewed` event, you should see: 
    - [x] the property `discover_category_name`, and it should say `Art` (the parent category)
    - [x] the property `discover_subcategory_name` and it should say `Ceramics` (the current category)
   - [x] the properties `discover_category_id` and `discover_subcategory_id`
- [x] navigate to the "Projects We Love" filter. the `Explore Page Viewed` event should fire with a property `discover_pwl: true`
- [x] navigate to the "Saved" filter. The `Explore Page Viewed` event should fire with a property `discover_watched: true`

### DIFF
--- a/Library/Tracking/Koala.swift
+++ b/Library/Tracking/Koala.swift
@@ -620,9 +620,11 @@ public final class Koala {
     screen: CheckoutContext
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
-      .withAllValuesFrom(["screen": screen.trackingString,
-                          "backer_reward_minimum": reward?.minimum,
-                          "pledge_total": backing?.amount])
+      .withAllValuesFrom([
+        "screen": screen.trackingString,
+        "backer_reward_minimum": reward?.minimum,
+        "pledge_total": backing?.amount
+      ])
 
     self.track(event: "Select Reward Button Clicked", properties: props)
   }
@@ -2182,7 +2184,7 @@ private func discoveryProperties(
   let parentCategoryProps = params.category?.parent.map { properties(category: $0) }
 
   result = result.withAllValuesFrom(categoryProps ?? [:])
-  result = result.withAllValuesFrom(parentCategoryProps ?? [:] )
+  result = result.withAllValuesFrom(parentCategoryProps ?? [:])
 
   result["everything"] = result.isEmpty
   result["sort"] = params.sort?.rawValue
--- a/Library/Tracking/Koala.swift
+++ b/Library/Tracking/Koala.swift
@@ -434,8 +434,10 @@ public final class Koala {
 
   /// Call when the activities screen is shown.
   public func trackActivities(count: Int) {
-    self.track(event: DataLakeWhiteListedEvents.activityFeedViewed.rawValue,
-               properties: ["activities_count": count])
+    self.track(
+      event: DataLakeWhiteListedEvents.activityFeedViewed.rawValue,
+      properties: ["activities_count": count]
+    )
   }
 
   // MARK: - Application Lifecycle
@@ -502,8 +504,10 @@ public final class Koala {
   public func trackTabBarClicked(_ tabBarItemLabel: TabBarItemLabel) {
     let label = tabBarItemLabel.trackingString
 
-    self.track(event: DataLakeWhiteListedEvents.tabBarClicked.rawValue,
-               properties: ["ios_tab_bar_label": label])
+    self.track(
+      event: DataLakeWhiteListedEvents.tabBarClicked.rawValue,
+      properties: ["ios_tab_bar_label": label]
+    )
   }
 
   // MARK: - Discovery Events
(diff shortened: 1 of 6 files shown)


Library/Tracking/Koala.swift
```diff
@@ -2127,39 +2135,6 @@ private func projectProperties(
     .prefixedKeys(prefix)
 }
 
-private func properties(
```
ifbarrera: This was almost completely redundant with `projectProperties`. Any actually unique props were moved up to the actual tracking function.

Library/ViewModels/SettingsAccountViewModelTests.swift
```diff
@@ -72,21 +72,17 @@ internal final class SettingsAccountViewModelTests: TestCase {
...
 
-    withEnvironment(koala: Koala(dataLakeClient: dataLakeClient, client: koalaClient)) {
+    withEnvironment(koala: Koala(client: koalaClient)) {
       XCTAssertEqual([], koalaClient.events)
-      XCTAssertEqual([], dataLakeClient.events)
 
       self.vm.inputs.viewDidAppear()
 
       XCTAssertEqual(["Viewed Account"], koalaClient.events)
-      XCTAssertEqual(["Viewed Account"], dataLakeClient.events)
```
ifbarrera: Don't need this test anymore since we're testing white-listing events now.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/Tracking/Koala.swift
```diff
@@ -21,6 +21,24 @@ public final class Koala {
   private var preferredContentSizeCategoryObserver: Any?
   private let screen: UIScreenType
 
+  private enum DataLakeWhiteListedEvents: String, CaseIterable {
```
justinswart: Any reason for the enum vs just an array of event names? I guess an enum could help us in a situation where we want to be sure that same string is used in two places. Should we considering using the `rawValue` of the corresponding case in that event's tracking function too? Something like:

```swift
public func trackDiscovery(params: DiscoveryParams) {
  let props = discoveryProperties(from: params)

  self.track(event: DataLakeWhiteListedEvents.explorePageViewed.rawValue, properties: props)
}
```
🤔 I'm not sure this is what we want, just thinking out loud.

Library/Tracking/Koala.swift
```diff
@@ -2198,34 +2173,34 @@ private func discoveryProperties(
...
-  result["staff_picks"] = params.staffPicks
-  result["starred"] = params.starred
+  result["pwl"] = params.staffPicks
+  result["watched"] = params.starred
   result["tag"] = params.tagId?.rawValue
-  result = result.withAllValuesFrom(params.category.map(properties(category:)) ?? [:])
+  let categoryProps = params.category.map { properties(category: $0, prefix: "subcategory_") }
+  let parentCategoryProps = params.category?.parent.map { properties(category: $0) }
+
+  result = result.withAllValuesFrom(categoryProps ?? [:])
```
justinswart: Maybe this reads slightly better here:
```swift
result = result
  .withAllValuesFrom(categoryProps ?? [:])
  .withAllValuesFrom(parentCategoryProps ?? [:])
```

Library/Tracking/Koala.swift
ifbarrera: Yeah I'd be in favor of doing that. Now's the time to do it too, since we're getting rid of a lot of events. I'm happy to do that in this PR if you're open to it.

ifbarrera: I think it's nice to have an enum of "clean" events that we can add to, and get auto-complete in the multiple places that we use the event names: in the "track" function, and in the corresponding test. It's less error prone and also easier to re-name, which I've had to do more than once lately 😅

justinswart: Yeah sounds good. Perhaps once we've fully removed the Koala client with plain strings we could change the `track` method signature to receive the `enum` type and only extract the `rawValue` at the point of tracking.

Anyway - one last suggestion before you merge - can you rename it to singular `DataLakeWhiteListedEvent`?

PR #997, opened 2019-12-17
[NT-692] Prepare Data Lake Request with auth headers

justinswart:
# 📲 What

Prepares requests sent to our Data Lake tracking endpoint with authentication headers.

# 🛠 How

Pass the request to our `Service.preparedRequest()` function before returning it to the client.

# ✅ Acceptance criteria

Inspect requests in Charles:

- [x] Observe that Data Lake requests include the `client_id` and `Authorization` header.


eoji:
Hello, it's me with an annoying question 👀the `client_id` is a query param and **not** a header, right? Wanted to make sure I'm doing the same when I open my PR

justinswart:
That's correct, `client_id` is a query param!

PR #999, opened 2019-12-19
[NT-580] Optimizely Integration Updates

ifbarrera:
# 📲 What

This PR makes some updates to our Optimizely integration:

- links the Optimizely SDK with the main app target and the `Kickstarter-Framework` instead of `Library`
- moves the input and output file locations to our `.xcfilelist` files
- removes `KSOptimizely` class in favor of a protocol extension on `OptimizelyClientType` that can be fully tested

# 🤔 Why

For our initial pass at an Optimizely SDK integration. We're anticipating tweaking the configuration values in the near future.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

We can test the `variant(for experiment:)` functionality by adding the following testing code to `PledgeCTAContainerViewViewModel.swift`:

```
    let experimentVariant = self.projectOrErrorProperty.signal
      .filterMap { _ -> (OptimizelyClientType, User)? in
        guard let optimizelyClient = AppEnvironment.current.optimizelyClient,
          let user = AppEnvironment.current.currentUser else {
          return nil
        }

        return (optimizelyClient, user)
      }.map { client, user in client.variant(for: .pledgeCTACopy, userId: String(user.id)) }
      .observeValues { variant in
        print("**** Optimizely variant: \(variant)")
      }
```

- [x] Run the app, and observe the console log `"🔮 Optimizely SDK Successfully Configured"`. Then, navigate to a project page. In the console, you should see your experimental variant printed.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -299,6 +299,28 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+  private func configureOptimizely(with key: String) {
+    let optimizelyClient = OptimizelyClient(sdkKey: key)
+
+    optimizelyClient.start { result in
+      var shouldUpdateClient: Bool
+
+      switch result {
+      case .success:
+        shouldUpdateClient = self.viewModel.inputs.optimizelyConfigured(isSuccess: true)
```
justinswart: I know we spoke about it but I'd like to still use a type here so that we don't have to do the switch outside of the view model. I think we can do this by extending Swift's `Result<T>` type:

```swift
extension Result where Failure == Optimizely.OptimizelyError {
  init(_ optimizelyResult: Optimizely.OptimizelyResult<Value>) {
    switch optimizelyResult {
    case .success(let value): self = .success(value)
    case .failure(let error): self = .failure(error)
    }
  }
}
```
^ we can add a simple test to just prove that this initializer returns the corresponding `Result<Value>` type for a given `Optimizely.OptimizelyResult<Value>` type.

and then this function can just be:
```swift
private func configureOptimizely(with key: String) {
  let optimizelyClient = OptimizelyClient(sdkKey: key)

  optimizelyClient.start { [weak self] result in
    if self?.viewModel.inputs.optimizelyConfigured(Result(result)) {
      print("🔮 Optimizely SDK Successfully Configured")
      AppEnvironment.updateOptimizelyClient(optimizelyClient)
    }
  }
}
```
^ I added `[weak self]` here too. Probably not that necessary considering we won't `deinit` our `AppDelegate` but 🤷‍♂.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/OptimizelyClientType.swift
```diff
@@ -1,9 +1,20 @@
...
 
-public protocol OptimizelyClientType: class {
-  func activate(experimentKey: String, userId: String, attributes: OptimizelyAttributes?) throws -> String
+public protocol OptimizelyClientType: AnyObject {
+  func activate(experimentKey: String, userId: String, attributes: [String: Any?]?) throws -> String
 }
 
-extension OptimizelyClient: OptimizelyClientType {
+extension OptimizelyClientType {
+  public func variant(
```
justinswart: Just to avoid coalescing twice we can do:

```swift
public func variant(
  for experiment: OptimizelyExperiment.Key,
  userId: String
) -> OptimizelyExperiment.Variant {
  guard
    let variation = try? self.activate(experimentKey: experiment.rawValue, userId: userId, attributes: nil),
    let experiment = OptimizelyExperiment.Variant(rawValue: variation)
  else {
    return .control
  }

  return experiment
}
```

PR #1001, opened 2019-12-19
[NT-717] Add Qualtrics custom properties and feature flag

justinswart:
# 📲 What

Adds `language` and `logged_in` custom properties as well handling `ios_qualtrics` feature flag.

# 🤔 Why

We need additional custom properties in order to do more granular filtering for Qualtrics intercepts. We only want to show surveys to logged in, English-language users.

# 🛠 How

- Added the custom properties.
- Added the feature flag.
- Added tests.

# ✅ Acceptance criteria

- [x] As a logged in user the Qualtrics SDK should be configured (observe console).
- [ ] As a logged out user the Qualtrics SDK should be _not_ configured (observe console).

# ⏰ TODO

- [ ] Merge Feature Flag (https://github.com/kickstarter/kickstarter/pull/18486).

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -615,13 +615,20 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     .filter(isTrue)
     .mapConst(0)
 
-    self.configureQualtrics = self.applicationLaunchOptionsProperty.signal.map { _ in
+    self.configureQualtrics = Signal.zip(
+      self.applicationLaunchOptionsProperty.signal,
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -615,13 +615,20 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     .filter(isTrue)
     .mapConst(0)
 
-    self.configureQualtrics = self.applicationLaunchOptionsProperty.signal.map { _ in
+    self.configureQualtrics = Signal.zip(
```
justinswart: Hmm, I made this zip because I didn't want this to emit again if the config updates, but perhaps we do want that if the user logs in during the app session?

PR #1004, opened 2019-12-23
[NT-719] Pledge properties

ifbarrera:
#📲 What

Adds pledge properties as per the new specs.

# 🤔 Why

Event cleanup.

# 🛠 How

Created a new `pledgeProperties` function to match the updated specs.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

When testing events, make sure the environment variable `KOALA_TRACKING` is set to `true`.

Testing `No Reward`
Navigate to any project, and select the "No reward" option from the rewards carousel. In the console, you should see the event `Select Reward Button Clicked` tracked with the following properties:

- [x] `pledge_backer_reward_id` // should be 0 for no reward
- [x] `pledge_backer_reward_has_items` // should be `false` for no reward
- [x] `pledge_backer_reward_is_limited_quantity` // should be `false` for no reward
- [x] `pledge_backer_reward_is_limited_time` // should be `false` for no reward
- [x] `pledge_backer_reward_minimum`
- [x] `pledge_backer_reward_shipping_enabled` // should be `false` for no reward

Testing `Reward`
Navigate to any project, and select a reward option from the rewards carousel that not `No reward`. In the console, you should see the event `Select Reward Button Clicked` tracked event with the following properties:

- [x] `pledge_backer_reward_id`
- [x] `pledge_backer_reward_has_items`
- [x] `pledge_backer_reward_is_limited_quantity` // should be `true` if reward is limited in quantity
- [x] `pledge_backer_reward_is_limited_time` // should be `true` if reward is limited in time
- [x] `pledge_backer_reward_minimum` // should be the reward's minimum in the project's currency
- [x] `pledge_backer_reward_shipping_enabled`
- [x] `pledge_backer_reward_shipping_preference` // could be `none`, `restricted` or `unrestricted` depending on the reward's shipping preferences

### DIFF
--- a/Library/ViewModels/RewardCardContainerViewModel.swift
+++ b/Library/ViewModels/RewardCardContainerViewModel.swift
@@ -84,7 +84,6 @@ public final class RewardCardContainerViewModel: RewardCardContainerViewModelTyp
         AppEnvironment.current.koala.trackRewardClicked(
           project: project,
           reward: reward,
-          backing: nil,
           screen: .backThisPage
         )
       }
--- a/Library/ViewModels/RewardCardContainerViewModel.swift
+++ b/Library/ViewModels/RewardCardContainerViewModel.swift
@@ -78,14 +78,13 @@ public final class RewardCardContainerViewModel: RewardCardContainerViewModelTyp
       .map { $0.id }
 
     // Tracking
-    projectAndRewardOrBacking
+    projectAndReward
       .takeWhen(self.pledgeButtonTappedProperty.signal)
-      .observeValues { projectAndRewardOrBacking in
-        let (project, rewardOrBacking) = projectAndRewardOrBacking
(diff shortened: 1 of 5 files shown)


Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -78,14 +78,12 @@ public final class RewardCardContainerViewModel: RewardCardContainerViewModelTyp
       .map { $0.id }
 
     // Tracking
-    projectAndRewardOrBacking
+    projectAndReward
       .takeWhen(self.pledgeButtonTappedProperty.signal)
-      .observeValues { projectAndRewardOrBacking in
```
ifbarrera: I think we should actually refactor this view model to only accept a `reward` for configuration -- right now it accepts an `Either<Reward, Backing>` type, although we never actually use the `Backing` configuration.

PR #996, opened 2019-12-17
[NT-681] CTA button reload fix

Scollaco:
# 📲 What 
- Changes the behaviour of the CTA button on which reloaded everytime the Project Page appeared.

# 🤔 Why

- While is important to always show the latest project's data, having the button reloading on viewWillAppear ended up being overkill and having the button "flashing" all the time wasn't a great UX.

# 🛠 How

- Creating a notification `projectBacked` that will send a message to the Project Page once the user taps the `X` button on the Thanks Page.
- The function ManagePledgeViewControllerDelegate now sends messages to its observers once a pledge is cancelled or updated (before it was being used only for cancelling).
- Added and updated tests

# ✅ Acceptance criteria

The CTA button should load when:
- [ ] Project Page is presented (viewDidLoad)
- [ ] A pledge is updated (After closing the PledgeViewController)
- [ ] A pledge is cancelled
- [ ] A project is backed

The CTA button should not load when:
- [ ] Returning from screens such as `Creator's bio`, `Campaign page`, `Comments` and `Updates`

GIF

PR #1005, opened 2019-12-24
[NT-721] Updating Project Page Viewed event timing

ifbarrera:
# 📲 What

Updates the logic of when the `Project Page Viewed` event fires so that we can track all project properties.

# 🤔 Why

Previously, we were firing the `Project Page Viewed` event right when the project page is configured. Because we prefix the `fetchProject` request with a partial project object, the `Project Page Viewed` event was being sent with incomplete `projectProperties`. 

Now, we wait for the `fetchProject` request to complete with the full project object before we track the `Project Page Viewed` event -- this allows us to send the event with all project properties.

# 🛠 How

Shuffling around the view model logic for the `Project Page Viewed` event.

# ✅ Acceptance criteria

Make sure you have the `KOALA_TRACKING` environment variable set to `true`

- [x] Navigate to any project. You should eventually see the `Project Page Viewed` event fire with the following properties populated:
    - `project_rewards_count` // should not be zero
    - `project_updates_count`
    - `project_comments_count`
- [x] Tap "Back this project" to navigate to the rewards carousel, then tap the X to close the rewards carousel. You should see the project page again and the `Project Page Viewed` event should fire again.

### DIFF
--- a/Library/ViewModels/ProjectPamphletViewModel.swift
+++ b/Library/ViewModels/ProjectPamphletViewModel.swift
@@ -189,14 +189,17 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       }
       .take(first: 1)
 
-    Signal.combineLatest(
-      freshProjectAndRefTag,
-      cookieRefTag,
+    Signal.zip(
+      freshProjectAndRefTag.skip(first: 1),
       self.viewDidAppearAnimated.signal.ignoreValues()
     )
-    .map { (project: $0.0, refTag: $0.1, cookieRefTag: $1, _: $2) }
-    .take(first: 1)
-    .observeValues { project, refTag, cookieRefTag, _ in
+    .map(unpack)
+    .map { project, refTag, _ in
+      let cookieRefTag = cookieRefTagFor(project: project) ?? refTag
+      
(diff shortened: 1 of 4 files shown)


Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -189,14 +189,17 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       }
       .take(first: 1)
 
-    Signal.combineLatest(
-      freshProjectAndRefTag,
-      cookieRefTag,
+    Signal.zip(
```
ifbarrera: Note that we'll need to change this to a `combineLatest` once https://github.com/kickstarter/ios-oss/pull/996 merges, because the `fetchProject` request will no longer fire on `viewDidAppear`.

Library/ViewModels/ProjectPamphletViewModelTests.swift
```diff
@@ -840,30 +876,6 @@ final class ProjectPamphletViewModelTests: TestCase {
     }
   }
 
-  func testBackThisProjectButton_Tracking() {
```
ifbarrera: I removed this test because it wasn't really testing anything that we weren't already verifying in other tests 🤔

PR #1007, opened 2020-01-02
Configure OptimizelyLogLevel based on bundle type

ifbarrera:
# 📲 What

Configures the Optimizely SDK's log level based on the bundle type. For debug builds, the log level is `.debug`, which gives us as much debugging information as possible. For release builds, the log level is set to `.error` which is the highest possible log level, only logging severe errors.

# 🤔 Why

To prevent excessive logging on release builds.

# 🛠 How

Added an additional output type `KSROptimizelyLogLevel` to the `configureOptimizely` output which will either be `.debug` or `.error` depending on the main bundle type.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] Run the app on a debug build. Observe in the console the following log: `[OPTIMIZELY] [DEBUG] SDK Version: 3.1.0`
- [x] Run the app on a release build. The console should not log any messages related to Pptimizely, with the exception of `🔮 Optimizely SDK Successfully Configured`

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -19,6 +19,11 @@ public enum NotificationAuthorizationStatus {
   case provisional
 }
 
+public enum KSROptimizelyLogLevel {
+  case error
+  case debug
+}
+
 public protocol AppDelegateViewModelInputs {
   /// Call when the application is handed off to.
   func applicationContinueUserActivity(_ userActivity: NSUserActivity) -> Bool
@@ -95,7 +100,7 @@ public protocol AppDelegateViewModelOutputs {
   var configureFabric: Signal<(), Never> { get }
 
   /// Emits when the application should configure Optimizely
-  var configureOptimizely: Signal<String, Never> { get }
+  var configureOptimizely: Signal<(String, KSROptimizelyLogLevel), Never> { get }
 
   /// Return this value in the delegate method.
   var continueUserActivityReturnValue: MutableProperty<Bool> { get }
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -19,6 +19,11 @@ public enum NotificationAuthorizationStatus {
   case provisional
 }
 
+public enum KSROptimizelyLogLevel {
```
justinswart: Any reason you chose this instead of `OptimizelyLogLevelType` in its own file? It seems similar to `OptimizelyClientType` and `OptimizelyResultType`.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -954,15 +959,22 @@ extension ShortcutItem {
   }
 }
 
-private func optimizelySDKKey(for environmentType: EnvironmentType) -> String {
+private func optimizelyData(for environment: Environment) -> (String, KSROptimizelyLogLevel) {
+  let environmentType = environment.environmentType
+  let logLevel = environment.mainBundle.isDebug ? KSROptimizelyLogLevel.debug : KSROptimizelyLogLevel.error
+
+  var sdkKey: String
```
justinswart: This can be a `let` too.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
ifbarrera: I agree `OptimizelyLogLevelType` sounds better 👍

PR #1003, opened 2019-12-20
[NT-503] Xcode 11/Swift 5.1 support

justinswart:
# 📲 What

Updates the app to compile in Xcode 11 and Swift 5.1 with some caveats:

- The app is currently compiling in Xcode 11 and Swift 5.1 but snapshot tests are failing. We will address this in a separate PR and merge that into this one.
- We can't merge `master` just yet because we first need a Swift 5.1.3 compatible version of the Qualtrics SDK.

# 🤔 Why

We know that we'll need to compile against iOS 13 and Swift 5.1 in the near future and want to get ahead of that requirement from the App Store as well as to keep our codebase up to date and compiling in the latest version of Xcode.

# 🛠 How

- Combed through the code and updated anything that wouldn't compile, mainly:
  - `Double`s that were being inferred as `Int` in many places.
  - Lenses that were causing ambiguities.
- Updated our Formatter tests due to some changes introduced by iOS 13 (will note in the code).

# ✅ Acceptance criteria

For now, this PR addresses everything except the tests in `Kickstarter_Framework`. I expect that it will be a fair bit of work to fix all of the UI and snapshot regressions and will open a separate PR based off this one to address those.

- [ ] KsApi tests pass.
- [ ] Library tests pass.
- [ ] Formatter test changes are 👍 

# ⏰ TODO

- [x] Fix UI regressions and snapshot tests in separate PR which we'll merge into this one.
- [x] Merge https://github.com/kickstarter/Kickstarter-Prelude/pull/96.
- [x] Merge https://github.com/kickstarter/Kickstarter-ReactiveExtensions/pull/81.

PR #1009, opened 2020-01-03
[NT-722] Xcode 11 snapshots

justinswart:
# 📲 What

Records new snapshots in Xcode 11 and iOS 13.3.

**Note:** This is based off #1003.

# 🤔 Why

We use snapshot testing to verify visual changes to our views. Unfortunately Xcode 11 and iOS 13 introduces many changes, some subtle (font anti-aliasing, etc.) and some not-so-subtle (broken constraints, views mis-aligned).

We will need to review all of these snapshots for regressions.

# 🛠 How

- Updated our snapshot testing framework - https://github.com/uber/ios-snapshot-test-case/releases/tag/6.2.0
- Updated calls to `FBSnapshotVerifyView` and removed `tolerance` values.
  - We weren't using `tolerance` in that many places and figured we can add it back if necessary.
- Removed some calls to `setNeedsLayout()` which were causing snapshot tests to get into an infinite loop.
- Re-recorded _all_ of our snapshots.

# 🤷‍♂️ What now?

- We're going to have to review all of the snapshots for regressions and create bug tickets to fix them.
- We need to perform full regression testing on the app itself and create bug tickets for things that we discover during that process too.


justinswart:
Closing as per our plan to break this up into multiple PRs.

PR #1015, opened 2020-01-08
[NT-720] Add ApplePayCapableType to Environment, add Checkout Properties

ifbarrera:
# 📲 What

In order to better access and better test Apple Pay capabilities, this PR creates `ApplePayCapableType` and adds this type as a property on the `Environment`. This allows us to easily access Apple Pay capabilities from the environment, and to mock out various Apple Pay capabilities during testing.

# 🤔 Why

Events cleanup.

# 🛠 How

This refactor made it possible to access `applePayCapableFor(project)` from the `PledgeViewModel` without needing to inject this value into the configuration.

Additionally, this PR adds the following "Checkout Properties:
- `checkout_amount` // including shipping
- `checkout_payment_type` // APPLE_PAY or CREDIT_CARD
- `checkout_reward_id`
- `checkout_reward_title`
- `checkout_revenue_in_usd_cents`
- `checkout_shipping_amount`
- `checkout_reward_estimated_delivery_on`
- `checkout_reward_shipping_enabled`
- `checkout_user_has_eligibile_stored_apple_pay_card`

To test these checkout properties, this PR adds the "Checkout Completed" event.

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [ ] verify that checkout out with Apple Pay is working normally

# ⏰ TODO

- [x] add `checkout_revenue_in_usd_cents`


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/Extensions/Double.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+extension Double {
```
justinswart: Could we rename the file to `Double+Currency.swift` or something to make it a bit clearer that it's an extension and the functionality that it's adding?

Library/ApplePayCapable.swift
```diff
@@ -0,0 +1,111 @@
+import Foundation
+import KsApi
+import PassKit
+
+public protocol ApplePayCapableType {
```
justinswart: Just wondering about the name seeing as it does a little more than just tell us whether the device is capable of Apple Pay. Anyway, no super strong feelings about it, but just `ApplePayType` would probably be fine too 👍

ifbarrera: Yeah the term I was going for here is _capabilities_ to encompass all Apple-Pay-related capabilities. But `ApplePayCapabilitiesType` didn't quite have the same ring to it 😅. `ApplePayType` and then `ApplePay` seems a little vague to me.

Library/ViewModels/DeprecatedRewardPledgeViewModel.swift
```diff
@@ -497,7 +497,13 @@ public final class DeprecatedRewardPledgeViewModel: Type, Inputs, Outputs {
     )
     .map { ($0.0, $0.1, $1, $2, $3) }
     .takeWhen(Signal.merge(applePayEventAfterLogin, loggedInUserTappedApplePayButton))
-    .map(PKPaymentRequest.paymentRequest(for:reward:pledgeAmount:selectedShippingRule:merchantIdentifier:))
+    .map { PKPaymentRequest.paymentRequest(
```
Scollaco: Just for curiosity, is there any reason why you decided to pass the values explicitly?

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -3523,4 +3573,83 @@ final class PledgeViewModelTests: TestCase {
...
+        with: (amount: 25.0, min: 10.0, max: 10_000.0, isValid: true)
+      )
+      self.vm.inputs.submitButtonTapped()
+
+      self.scheduler.run()
+
+      XCTAssertEqual(
+        ["Pledge Screen Viewed", "Pledge Button Clicked"],
+        self.trackingClient.events
+      )
```
Scollaco: Hm.. I think the properties are not being tested here.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -3523,4 +3573,83 @@ final class PledgeViewModelTests: TestCase {
...
+      XCTAssertEqual(
+        PKPaymentAuthorizationStatus.success,
+        self.vm.inputs.stripeTokenCreated(token: "stripe_token", error: nil)
+      )
+
+      self.vm.inputs.paymentAuthorizationViewControllerDidFinish()
+
+      self.scheduler.run()
+
+      XCTAssertEqual(["Pledge Screen Viewed"], self.trackingClient.events)
```
Scollaco: Here too... which properties are involved in this test? 🤔

Library/ApplePayCapable.swift
Scollaco: I personally like `ApplePayCababilities`, because it's more general. For me, `ApplePayCapable` refers to something with a boolean type (function, var...).

Library/ViewModels/DeprecatedRewardPledgeViewModel.swift
ifbarrera: The compiler needed help 😓

Library/ViewModels/PledgeViewModelTests.swift
ifbarrera: Ah right, we actually don't need this test anymore because we don't have the `Completed Checkout` event anymore. Will remove!

PR #1016, opened 2020-01-08
[NT-744] Snapshots generation

Scollaco:
📲 What
- Regenerated some snapshots that were failing due some font changes made on iOS 13.

🤔 Why
- So everyone gets happy again.

✅ Acceptance criteria
- [ ] Nothing looks really broken.


Scollaco:
> All snapshots look good the only one that looks a little different is https://github.com/kickstarter/ios-oss/pull/1016/files?short_path=bd985ae#diff-78b558d62788ec17b885b688575e8bab

Hmm Interesting. If I inspect the diff using the `swipe` mode, the difference is hardly noticed. I wonder if seeing the diff side by side affects the scale of the images.

PR #1021, opened 2020-01-08
[NT-747] New snapshots

justinswart:
# 📲 What

Records new snapshots for `PledgeViewController` and `ProjectNotificationsViewController`.

# 🤔 Why

A continuation of the work in #1003, these VCs appear to have experience minor font-size changes and just needed to be re-recorded.

# ✅ Acceptance criteria

- [ ] Review new snapshots.


justinswart:
@cdolm92 thanks for flagging - it seems that this is the default _off_ state tint color now, so I think it's ok given that the switches will remain consistent with the rest of the app and OS. I think the alternative would be to specify a tint colour and have to manage that everywhere which I think we'd wand to avoid.

PR #1024, opened 2020-01-09
[NT-777] Fix project modal dismissal

justinswart:
# 📲 What

Disables our custom dismissal of the project modal on iOS 13.

# 🤔 Why

This behaviour appears to be broken on iOS 13, views are positioned incorrectly and after dismissal the screen just goes blank. While there may be other ways to work around this, the built-in modal behaviour on iOS 13 works well and is arguably an improvement on what we have. I think that the easiest approach is to keep our current behaviour for iOS 12 and to disable it for 13+.

# 🛠 How

- Return `nil` for the animator object on iOS 13.
- Don't call the view model input with the panning gesture values on iOS 13.

# 👀 See

| iOS 13 with custom animation 🐛 | iOS 12 (current behaviour) | iOS 13+ 🦋 |
| --- | --- | --- |
|  |  |  |

# ✅ Acceptance criteria

Run the app in Xcode 11 on the 12.4 and 13.3 simulators.

- [x] Dismissal behaviour is unchanged on iOS 12.
- [x] Project modal is presented using the new card-style modal on iOS 13 and the user is still able to pan down to dismiss the modal.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
@@ -9,7 +9,11 @@ internal protocol ProjectNavigatorDelegate: AnyObject {
 }
 
 internal final class ProjectNavigatorViewController: UIPageViewController {
-  fileprivate let transitionAnimator = ProjectNavigatorTransitionAnimator()
+  fileprivate lazy var transitionAnimator: ProjectNavigatorTransitionAnimator? = {
+    guard #available(iOS 13.0, *) else { return ProjectNavigatorTransitionAnimator() }
+    return nil
+  }()
+
   fileprivate weak var navigatorDelegate: ProjectNavigatorDelegate?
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/Controllers/ProjectNavigatorViewController.swift
```diff
@@ -169,6 +173,8 @@ extension ProjectNavigatorViewController: ProjectPamphletViewControllerDelegate
     _: ProjectPamphletViewController,
     panGestureRecognizerDidChange recognizer: UIPanGestureRecognizer
   ) {
+    if #available(iOS 13.0, *) { return }
```
justinswart: I considered placing this check in the VM and trying to test it but we ran into a couple of issues trying to test these availability checks in the past and removed it in #731 so I'm thinking this is ok for this?

Scollaco: I think that in this case it's fine to have the check here.

PR #1025, opened 2020-01-09
[NT-699] Align Reward Close Button w/ Project Close Button

cdolm92:
# 📲 What

In rewards carousel, X in the top bar is in a different location than the project page so we're moving it to the same location as rewards carousel page.

# 🤔 Why

So that transitioning between these screens is smoother.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone X - 2020-01-09 at 17 28 39| Simulator Screen Shot - iPhone X - 2020-01-09 at 17 04 23|

# ✅ Acceptance criteria

- [ ] `X` buttons in Project Page and Rewards Carousel appears in the same position


Scollaco:
I noticed that the `X` on the Project Page is still a little misaligned. This should be just a matter of moving it a bit more:

x_buttons_alignment

PR #1026, opened 2020-01-10
[NT-757] Record new snapshots and improve UIView swizzling

justinswart:
# 📲 What

- Records new `ProjectActivityViewController` snapshots.
- Improves our swizzling of `UIView` in order for `bindStyles()` to be called at the appropriate time.

# 🤔 Why

Since the changes introduced to when `traitCollectionDidChange(_:)` is called on iOS 13 (described in #1003) our swizzled implementation that ensured `bindStyles()` would be called for every `UIView` stopped working correctly.

In #1003 I tried moving this to `layoutSubviews()` but this seemed to only improve it in some views, I believe it is being called too late in the view's lifecycle. Storyboard views give us the handy `awakeFromNib()` callback and placing `bindStyles()` there appears to have fixed the layout issues we were seeing in snapshots for this view, `ProjectActivityViewController`. I believe it is likely to fix other storyboard views too.

For our programmatic views, however, there isn't really anything exactly like `awakeFromNib()` so I'm hoping that `didMoveToSuperview()` can perform this function for us. Apple [docs](https://developer.apple.com/documentation/uikit/uiview/1622433-didmovetosuperview) say:

> `didMoveToSuperview()` Tells the view that its superview changed.
> The default implementation of this method does nothing. Subclasses can override it to perform additional actions whenever the superview changes.

Given that the default implementation does nothing, and that we're swizzling (🤷‍♂), this seems like a good place to do this. It also appears to be called earlier than `layoutSubviews()` (citation needed, although that makes sense).

# 🛠 How

- Removed our swizzling of `layoutSubviews()`
- Added `bindStyles()` to our swizzled `awakeFromNib()` for storyboard/nib views.
- Swizzled `didMoveToSuperview()` for programmatic views.

# ✅ Acceptance criteria

- [ ] The newly recorded snapshots are correct.
- [ ] We're willing to try this `UIView` swizzling approach and don't observe any regressions in other views.


justinswart:
> Interesting, good idea to try this! Let me just checkout your branch and play around with it to see if there's any weirdness, but overall this seems better than using `layoutSubviews` because as far as I can tell it'll only get called once per time that the view is added as a subview of it's parent. That seems closest to what we want.

Yeah, it's also called before `layoutSubviews` which makes sense for our use case so that the views are configured/styled before the layout pass.

In fact it's just occurred to me that we probably don't need this in `awakeFromNib` as well considering that those views will "move to superview" too so we'd be binding styles twice unnecessarily.

Let me remove that and see if these snapshots still pass.

PR #1027, opened 2020-01-10
[NT-635] CTA fix button navigation

Scollaco:
# 📲 What 
- Fixes the navigation from the cta button on Project Page (errored state) to the ManagePledgeViewController.

# 🤔 Why
- This was part of the requirements on this [ticket](https://kickstarter.atlassian.net/browse/NT-635).

# 🛠 How
 The button wasn't firing because we were only checking if the `PledgeStateCTAType` was `.viewBacking` or `manage`. To fix that, we now also check if the button state is equal to  `.fix`.

# 👀 See
fix_pledge_navigation

# ✅ Acceptance criteria
If you have an errored pledge, select the project with an errored pledge and tap the `Manage` button.
- [ ] The ManagePledge screen should be presented.

If you don't have an errored pledge:
Select one of your successful and backed project. With Charles, add a breakpoint to intercept the response when the Project Page is presented (v1/projects) and change the backing status to errored. Tap the `Manage` button.
- [ ] The ManagePledge screen should be presented.

### DIFF
--- a/Library/ViewModels/ProjectPamphletViewModel.swift
+++ b/Library/ViewModels/ProjectPamphletViewModel.swift
@@ -358,5 +358,5 @@ private func fetchProject(projectOrParam: Either<Project, Param>, shouldPrefix: 
 }
 
 private func shouldGoToManagePledge(with type: PledgeStateCTAType) -> Bool {
-  return type == .viewBacking || type == .manage || type == .fix
+  return type.isAny(of: .viewBacking, .manage, .fix)
 }
--- a/Library/ViewModels/ProjectPamphletViewModel.swift
+++ b/Library/ViewModels/ProjectPamphletViewModel.swift
@@ -127,7 +127,7 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .filter(featureNativeCheckoutPledgeViewIsEnabled >>> isFalse)
 
     let shouldGoToManagePledge = ctaButtonTappedWithType
-      .filter { $0 == .viewBacking || $0 == .manage }
+      .filter(shouldGoToManagePledge(with:))
       .ignoreValues()
       .filter(featureNativeCheckoutPledgeViewIsEnabled)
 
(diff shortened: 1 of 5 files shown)


Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -127,7 +127,7 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .filter(featureNativeCheckoutPledgeViewIsEnabled >>> isFalse)
 
     let shouldGoToManagePledge = ctaButtonTappedWithType
-      .filter { $0 == .viewBacking || $0 == .manage }
+      .filter(shouldGoToManagePledge(with:))
```
Scollaco: The compiler didn't let me simply add `$0 == .fix` here, so I had to extract the logic to a separated function.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -356,3 +356,7 @@ private func fetchProject(projectOrParam: Either<Project, Param>, shouldPrefix:
 
   return projectProducer
 }
+
+private func shouldGoToManagePledge(with type: PledgeStateCTAType) -> Bool {
+  return type == .viewBacking || type == .manage || type == .fix
```
justinswart: Instead of all the ORs I usually prefer writing this sort of thing as:

```swift
return [.viewBacking, .manage, .fix].contains(type)
```
But I know that some folks don't like to read this because it reverses things a bit and can be confusing. But I recall seeing [this tip](https://www.swiftbysundell.com/tips/expressively-comparing-a-value-with-a-list-of-candidates/) before.

So perhaps we should add something similar:

```swift
private func shouldGoToManagePledge(with type: PledgeStateCTAType) -> Bool {
  return type.isAny(of: .viewBacking, .manage, .fix)
}

extension Equatable {
  func isAny(of candidates: Self...) -> Bool {
    return candidates.contains(self)
  }
}
```
What do you think?

Scollaco: Ohh I really like that! It becomes much easier to read.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/EquatableHelpersTests.swift | blankLinesAtStartOfScope, sortedImports |


</p>

PR #1032, opened 2020-01-13
[NT-754, NT-769, NT-772] Record new snapshots

justinswart:
# 📲 What

Records new snapshots for:

`BackerDashboardProjectsViewController`
`ToggleViewController`
`PledgeAmountViewController`

# 🤔 Why

A continuation of the snapshot re-recording in #1028.

# ✅ Acceptance criteria

- [ ] Check out the branch locally, scan through the snapshots for the above VCs and confirm that they are correct.


Library/Styles/Fonts.swift
```diff
@@ -123,7 +123,15 @@ extension UIFont {
     default: defaultSize = 17
     }
 
-    let font = UIFont.preferredFont(forTextStyle: style)
+    let font: UIFont
+    if #available(iOS 13.0, *) {
+      font = UIFont.preferredFont(
+        forTextStyle: style,
+        compatibleWith: .current
```
justinswart: Seems this is now required in order for the dynamic type changes to take effect during snapshot tests.

Scollaco:
All snapshots look fine, but I noticed this diff:

Screenshot 2020-01-13 at 16 22 44

The functionality is still there, but I wonder why the views were moved down... 🤔

justinswart:
> All snapshots look fine, but I noticed this diff:
> 
> Screenshot 2020-01-13 at 16 22 44
> 
> The functionality is still there, but I wonder why the views were moved down... 🤔

It looks like the Title wasn't in the previous recording which I'm sure is incorrect, so I think the new snapshot fixes that.

PR #1033, opened 2020-01-13
[NT-730] Pledge screen header

Scollaco:
# 📲 What
- Replaces the Reward description session on the PledgeViewController.

# 🤔 Why
- To match the new [design](https://www.figma.com/file/DMwhdWji5zRLFAoIBDzVji/Checkout?node-id=86%3A137).

# 🛠 How
- Disabling the custom animation from the rewards carousel to the pledge screen
- Removing any code related to the old header
- Added logic to show/hide estimated delivery date stackView if reward doesn't have deliverable items
- Added new `NSAttributedString` extension to make it easier the process to add links (not HTML) to strings 
- Added tests and snapshots

# 👀 See

| No Reward | Reward w/o deliverables | Reward w/ deliverables | Dynamic type
| --- | --- | --- | --- |
| Simulator Screen Shot - iPhone 8 - 2020-01-13 at 12 29 44 | Simulator Screen Shot - iPhone 8 - 2020-01-13 at 12 30 08 | Simulator Screen Shot - iPhone 8 - 2020-01-13 at 12 29 54 | Simulator Screen Shot - iPhone 8 - 2020-01-13 at 12 58 27 |

# ♿️ Accessibility 
- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria
Go to a live project and choose:
- [ ] A `NoReward` reward. The header title should say "Back it because you believe in it`, delivery info is `hidden`.
- [ ] A reward **without** deliverable items. The header should show the rewards title and the delivery info is `hidden`.
- [ ] A reward **with** deliverable items. The header should show the estimated delivery info above the rewards title.


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -483,28 +483,15 @@ extension PledgeViewController: PledgeShippingLocationViewControllerDelegate {
...
 
 extension PledgeViewController: RewardPledgeTransitionAnimatorDelegate {
-  func beginTransition(_ operation: UINavigationController.Operation) {
-    self.descriptionViewController.beginTransition(operation)
-  }
-
-  func snapshotData(withContainerView view: UIView) -> RewardPledgeTransitionSnapshotData? {
-    return self.descriptionViewController.snapshotData(withContainerView: view)
-  }
+  func beginTransition(_: UINavigationController.Operation) {}
```
Scollaco: Instead of deleting all of this code, I decided to keep this structure here since we discussed in standup that we may revisit the custom animation in an Investment Day.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/PledgeDescriptionViewController.swift
```diff
@@ -344,100 +191,46 @@ private let learnMoreTextViewStyle: TextViewStyle = { (textView: UITextView) ->
...
 
-  let newWidth = min(maxWidth, max(width / 3, Layout.RewardThumbnail.minWidth))
-  let newHeight = newWidth * aspectRatio
+  let storeString = Strings.Kickstarter_is_not_a_store_Its_a_way_to_bring_creative_projects_to_life()
+  let accountabilityString = Strings.Learn_more_about_accountability()
+  let fullString = storeString + " " + accountabilityString
 
-  return .init(width: newWidth, height: newHeight)
-}
+  guard let attributedString = checkoutAttributedLink(with: fullString) else { return nil }
```
justinswart: I don't think we still want to call `checkoutAttributedLink` here because that function is expecting an HTML string.

Where do we set the attributes for the link text, colour etc?

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: On second thought I think we should remove the animation, I think it's unlikely we'll make it work with this new layout.

Library/ViewModels/PledgeDescriptionViewModel.swift
```diff
@@ -32,10 +33,15 @@ public final class PledgeDescriptionViewModel: PledgeDescriptionViewModelType,
...
+      .skipNil()
+      .map(second)
+      .map { $0.estimatedDeliveryOn.isNil }
+
     self.presentTrustAndSafety = self.learnMoreTappedProperty.signal
-    self.configureRewardCardViewWithData = self.configDataProperty.signal
+    self.rewardTitle = self.configDataProperty.signal
       .skipNil()
-      .map { project, reward in (project, .init(left: reward)) }
+      .map { _, reward in reward.title ?? Strings.Back_it_because_you_believe_in_it() }
```
justinswart: Does it make sense to check if title is `nil` and also whether its `noReward`? Just to be sure 🤔

Scollaco: The `NoReward` only contains some of the properties of the `Reward` object, and it doesn't have the `title`. So the title, in this case, will always be `nil`.

Scollaco:
> Cool cool, removing the use of `checkoutAttributedLink` seems to have changed the font style, is this the correct style now? Perhaps we should confirm with @colleenmacd.
> 
> There is another file that is no longer need now since we removed the animation called `UISpringTimingParameters+Convenience.swift`.

Thanks, @justinswart ! I made a small change on the NSAttributedString and now the style seems correct again.

Scollaco:
> Looks good! In the reward container, is the padding between the edge of container and the text 18 dp? Looks a little large but might be my eyes @Scollaco

It is... I double checked it, just in case.

PR #1036, opened 2020-01-16
[NT-789] Additional Qualtrics custom properties

justinswart:
# 📲 What

Adds additional custom properties for us to send to Qualtrics.

# 🤔 Why

These additional properties allow us to better-filter our targeting logic for surveys.

# 🛠 How

- Added `distinct_id` to distinctly identify the device.
- Added `user_uid` to be able to de-dupe survey responses if necessary.
- Added `first_app_session` which we set to `0` after the first evaluation of the targeting logic in order to have a way of knowing that the app has been used at least once before prompting the user to complete the survey. 

# ✅ Acceptance criteria

- [ ] This PR should just be code-reviewed, we will test the behaviour with Qualtrics Intercepts separately in an alpha build.
- [x] Testing passed on alpha build.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -384,9 +384,8 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
   // MARK: - Qualtrics Configuration
 
   private func configureQualtrics(with config: QualtricsConfigData) {
-    let properties = Properties()
```
justinswart: This clearly wasn't correct - I'm not sure how we got here 🤔

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1059,3 +1049,21 @@ private func visitorCookies() -> [HTTPCookie] {
...
+    zoneId: Secrets.Qualtrics.zoneId,
+    interceptId: QualtricsIntercept.survey.interceptId,
+    stringProperties: [
+      "bundle_id": AppEnvironment.current.mainBundle.bundleIdentifier,
+      "language": AppEnvironment.current.language.rawValue,
+      "logged_in": "\(AppEnvironment.current.currentUser != nil)",
+      "distinct_id": AppEnvironment.current.device.identifierForVendor?.uuidString,
+      "user_uid": AppEnvironment.current.currentUser.flatMap { $0.id }.map(String.init),
+      "hours_since_joined": AppEnvironment.current.currentUser?
+        .hoursSinceJoined(AppEnvironment.current.dateType.init().date).flatMap(String.init)
```
eoji: does `String.init` mean you're sending all these properties as strings? should `hoursSinceJoined` be a number?

eoji: resolving this since we no longer have this property

Kickstarter-iOS/AppDelegate.swift
ifbarrera: Ha it wasn't doing anything before 🤦‍♀. Do you think it matters that we set the properties on the shared instance _before_ initializing it? 🤔

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -55,8 +55,10 @@ public protocol AppDelegateViewModelInputs {
   /// Call when the user taps "OK" from the contextual alert.
   func didAcceptReceivingRemoteNotifications()
 
-  /// Call with the result of evaluating Qualtrics Targeting Logic
-  func didEvaluateQualtricsTargetingLogic(with result: QualtricsResultType)
+  /// Call with the result of evaluating Qualtrics Targeting Logic and current Qualtrics.Properties
+  func didEvaluateQualtricsTargetingLogic(
+    with result: QualtricsResultType, properties: QualtricsPropertiesType
```
ifbarrera: Can we put the first argument on the same line as the definition 😄 

```
func didEvaluateQualtricsTargetingLogic(with result: QualtricsResultType, 
                                        properties: QualtricsPropertiesType)
```

Kickstarter-iOS/AppDelegate.swift
justinswart: This way seems to work.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: I can do:

```swift
  public func didEvaluateQualtricsTargetingLogic(
    with result: QualtricsResultType,
    properties: QualtricsPropertiesType
  ) {
```

PR #1038, opened 2020-01-16
[NT-678] Remove old checkout code

justinswart:
# 📲 What

Removes the deprecated checkout code.

**Note:** This is based off the Xcode 11 branch of #1003.

**Also note:** See TODOs below for some lingering code that needs a little more consideration before removing.

# 🤔 Why

We've rebuilt this flow natively and this code is no longer needed. We thank it and hope that it sparks joy in another code base.

# 🛠 How

- Removed all deprecated checkout files.
- Removed old checkout related Storyboards.
- Did a sweep for functions that are no longer in use and removed those.
- Removed redundant tests that depended on disabled feature flags or experiments.
- Removed v1 endpoints that we aren't using anymore.

# ✅ Acceptance criteria

- [ ] Confirm that all of the deleted code was correctly removed and is no longer needed.
- [ ] Test affected areas of the app for regressions (Project page, Checkout, Manage Pledge, Backing VC, etc).

# ⏰ TODO

- [ ] ~~Remove `BackingViewController` if possible.~~ Unlikely to be possible soon.
- [x] Refactor `FeatureFlagToolsViewModelTests` to not depend on actual Feature flags as we will remove these from time to time.
- [x] Remove native checkout feature flags once our tests don't depend on them.
- [ ] ~~Run [periphery](https://github.com/peripheryapp/periphery) on the code base to be sure I found all of the unused code.~~ Unfortunately there are [issues](https://github.com/peripheryapp/periphery/issues/49) running this with Xcode 11.
- [x] Remove old checkout-related tracking functions.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/Feature.swift
```diff
@@ -1,5 +1,7 @@
 import Foundation
 
+// TODO: remove nativeCheckout, nativeCheckoutPledgeView once we've refactored FeatureFlagToolsViewModelTests
```
justinswart: The tests in `FeatureFlagToolsViewModelTests` depend on the native checkout feature flags being defined. We would need to refactor that VM to either run its tests on different flags (although this will change every time we add/remove feature flags) or abstract it in some way to not be dependent on the concrete flags. The tests are mostly just proving that we order the flags alphabetically.

Kickstarter-iOS/Views/Cells/DeprecatedRewardCell.swift
```diff
@@ -1,270 +0,0 @@
-import KsApi
-import Library
-import Prelude
-
-internal protocol DeprecatedRewardCellDelegate: AnyObject {
```
Scollaco: I noticed that `Nib.swift` still has a `case DeprecatedRewardCell`. Could you remove it?

Kickstarter-iOS/Views/Controllers/ProjectPamphletViewControllerTests.swift
```diff
@@ -316,60 +271,10 @@ internal final class ProjectPamphletViewControllerTests: TestCase {
...
-
-      FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)")
-    }
-  }
-
-  // MARK: - Native Checkout Enabled, Error fetching project
+  // MARK: - Error fetching project
 
-  func testProjectPamphletViewController_NativeCheckoutEnabled_ErrorFetchingProject() {
+  func testProjectPamphletViewController() {
```
Scollaco: Should we rename this to `testProjectPamphletViewController_ErrorFetchingProject`? I think it's more descriptive to what we are testing here.

Library/ViewModels/ProjectPamphletContentViewModel.swift
```diff
@@ -150,7 +148,7 @@ public final class ProjectPamphletContentViewModel: ProjectPamphletContentViewMo
   public let goToRewardPledge: Signal<(Project, Reward), Never>
   public let goToUpdates: Signal<Project, Never>
   public let loadMinimalProjectIntoDataSource: Signal<Project, Never>
-  public let loadProjectIntoDataSource: Signal<(Project, Bool), Never>
+  public let loadProjectIntoDataSource: Signal<Project, Never>
   public let rewardTitleCellVisible: Signal<Bool, Never>
```
Scollaco: Do we still need this?

Kickstarter-iOS/Views/Controllers/FeatureFlagToolsViewController.swift
```diff
@@ -35,7 +35,7 @@ final class FeatureFlagToolsViewController: UITableViewController {
     self.viewModel.outputs.reloadWithData
       .observeForUI()
       .observeValues { [weak self] features in
-        self?.features = features
+        self?.features = featureEnabledFromDictionaries(features)
```
Scollaco: Hm, I wonder if it wouldn't be better if the viewModel returned the results of `featureEnabledFromDictionaries(features)`, so we wouldn't need to access its function directly. Is there any reason why you preferred this approach?

justinswart: Yeah this is intentional. The tests for this VC `FeatureFlagToolsViewControllerTests.swift` were testing that the ordering of the returned flags was correct. The problem (I felt) with those tests is that every time flags are removed those tests would break. When in fact that is not really what the tests were proving. So I changed the tests to just test simple dictionaries which meant that we needed an additional function to prove that we don't create `Feature` `enums` from flags that we don't support. For this purpose there is an [additional test](https://github.com/kickstarter/ios-oss/pull/1038/files#diff-ec92627879e52efe83c947c4f52aac09R102) which tests the function above.

Scollaco: Oh ok. Thanks for explaining it.

PR #1041, opened 2020-01-17
[NT-795] Saved projects refresh bug fix

Scollaco:
# 📲 What
- Fix a bug on the BackerDashBoardViewController where the list wasn't being refreshed after watching/unwatching a project on iOS 13.

# 🤔 Why
- Because bugs are no good. [JIRA ticket](https://kickstarter.atlassian.net/browse/NT-795)

# 🛠 How
- On iOS 13, the `viewWillAppear` is no longer being called due changes made on the default modal presentation. So to be notified that the list and starred projects count should refresh, the existing `.ksr_projectSaved` notification is used.
- The notification is posted only if, after tapping the save button, the project was successfully updated on the server side.

# 👀 See

refresh_project_bug_fix

# ✅ Acceptance criteria

- [x] Using a iOS 13 device/simulator, go to Profile and tap the `Saved` button to see a list of saved projects. Select on project and tap the heart icon to "unwatch" the project. Dismiss the viewController. The project should be removed from the list and the saved projects count should be updated.


Library/ViewModels/BackerDashboardViewModel.swift
```diff
@@ -103,13 +106,16 @@ public final class BackerDashboardViewModel: BackerDashboardViewModelType, Backe
...
-    let fetchedUserEvent = self.viewWillAppearProperty.signal.ignoreValues()
-      .switchMap { _ in
-        AppEnvironment.current.apiService.fetchUserSelf()
-          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
-          .prefix(SignalProducer([AppEnvironment.current.currentUser].compact()))
-          .materialize()
-      }
+    let fetchedUserEvent = Signal.merge(
+      self.projectSavedProperty.signal.ignoreValues(),
+      self.viewWillAppearProperty.signal.ignoreValues()
```
ifbarrera: If we do a `merge`, won't the `fetchUserSelf` network request fire twice for < iOS 13 devices? I wonder if it would be better to use `self.viewDidLoadProperty` so that the network request is made once on initialization, and then again only when a project is saved/unsaved.

Library/ViewModels/BackerDashboardViewModelTests.swift
```diff
@@ -96,6 +96,18 @@ internal final class BackerDashboardViewModelTests: TestCase {
       // Signals that emit just once because they rely on the datasource tab index to exist first.
       self.pinSelectedIndicatorToTab.assertValues([.backed])
       self.pinSelectedIndicatorToTabAnimated.assertValues([false])
+
+      let user2 = user
+        |> \.name .~ "Updated user"
+        |> \.stats.starredProjectsCount .~ 60
+
+      withEnvironment(apiService: MockService(fetchUserSelfResponse: user2)) {
+        self.vm.inputs.projectSaved()
```
ifbarrera: Do you think this should maybe go in its own test? Something like `testProjectSavedRefreshesData`.

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -102,10 +102,8 @@ public final class WatchProjectViewModel: WatchProjectViewModelType,
       .switchMap { project, shouldWatch in
         watchProjectProducer(with: project, shouldWatch: shouldWatch)
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
-          .map { $0.watchProject.project.isWatched }
           .map { _ in (project, project.personalization.isStarred ?? false, success: true) }
           .flatMapError { _ in .init(value: (project, !shouldWatch, success: false)) }
-          .take(until: saveButtonTapped.ignoreValues())
```
ifbarrera: What was the motivation for removing this line? I'm not super clear on why it was there at all, maybe @justinswart can shed some light on this.

Library/ViewModels/WatchProjectViewModel.swift
```diff
@@ -171,8 +169,7 @@ public final class WatchProjectViewModel: WatchProjectViewModelType,
     self.saveButtonAccessibilityValue = self.saveButtonSelected
       .map { starred in starred ? Strings.Saved() : Strings.Unsaved() }
 
-    self.postNotificationWithProject = project
-      .takeWhen(saveButtonTapped)
+    self.postNotificationWithProject = projectOnSaveButtonToggleSuccess
```
ifbarrera: So now we're only firing the notification when the request actually succeeds. Does this have any user experience implications for other parts of the app where the save icon is shown? Was this intentionally done this way previously, perhaps for "optimistic UI" reasons?

Library/ViewModels/BackerDashboardViewModelTests.swift
Scollaco: I thought about it but then decided to follow the convention of the file (what I don't like at all, because it makes the tests too big and hard to see everything that's being tested)... But after a second thought I think I should do a separated test.

Library/ViewModels/WatchProjectViewModel.swift
Scollaco: When I saw this code I couldn't understand either why we had those lines, and removing them seems that didn't change anything. @justinswart any idea why we had those?

Library/ViewModels/WatchProjectViewModel.swift
Scollaco: Using `project` was sometimes causing a behaviour where after un-saving a project, the `Saved` list would not remove the project. Instead, it would simply hide the heart icon. I noticed though that this doesn't happen often. So I guess we could revert this. What do you think?

Library/ViewModels/BackerDashboardViewModel.swift
Scollaco: Actually  `viewDidLoadProperty` won't work in some scenarios, for example:
1-  User opened the open the list of `Saved` projects
2-  Go to Discovery and save another project 
3- Go back to the `Saved` list.

In this case, the list won't refresh (because the notification is posted on the ProjectNavBarViewController).

Library/ViewModels/WatchProjectViewModel.swift
justinswart: Hmm yeah this is some time ago, line `105` seems strange, I'm not sure why I was mapping that value and then ignoring it 🤷‍♂so you can probably remove that. I imagine the `take(until:)` was there because I wanted the producer to immediately complete if the button was tapped again. You may look at the code and think that we don't need that because the `switchMap` would take care of that but take note of the debouncing that's happening. It will take 500ms before the switchMap cancels the in-flight request so I think we do still want that `take(until:)` so that it completes/cancels immediately.

Nevertheless, making any changes here is going to require some regression testing!

PR #1042, opened 2020-01-21
[NT-807] Add new card event

ifbarrera:
# 📲 What

Adds a clean event "Add New Card Button Clicked" to the data lake with pledge properties, project properties, user properties, and session properties.

# 🤔 Why

Part of the "Back this project" event cleanup work.

# 🛠 How

Updates the `Add New Card Button Clicked` event so that it's sent to the data lake and contains the appropriate property groups. The event now includes:
- `projectProperties`
- `pledgeProperties`
- `userProperties`
- `sessionProperties`

# ✅ Acceptance criteria

- [x] With the `KOALA_ENABLED` flag set to `true`, navigate to any project and select any reward. Tap the "Add new card" button on the pledge screen. You should see the `Add New Card Button Clicked` event fire on the data lake, with all project properties, pledge properties, session properties and user properties.


Library/AppEnvironment.swift
```diff
@@ -158,7 +158,7 @@ public struct AppEnvironment: AppEnvironmentType {
       Environment(
         apiService: apiService,
         apiDelayInterval: apiDelayInterval,
-        applePayCapable: applePayCapable,
+        applePayCapabilities: applePayCapabilities,
```
ifbarrera: Missed some renames in my last PR 😅

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -652,13 +678,15 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
 
-    self.vm.inputs.configureWith((User.template, project, true))
     self.vm.inputs.viewDidLoad()
+    self.vm.inputs.configure(with: (User.template, project, Reward.template))
 
     XCTAssertEqual([], self.trackingClient.events)
 
     self.vm.inputs.addNewCardTapped(with: .pledge)
 
+    let props = self.trackingClient.properties.last
```
Scollaco: This variable is not being used. Did you write this to test the properties sent with the event?

ifbarrera: Hm so the properties are tested in `KoalaTests.swift`. I can remove this line 👍

PR #1045, opened 2020-01-23
[NT-780] Add device identifier to prepared request headers

justinswart:
# 📲 What

Adds the device identifier to our prepared request headers for requests to v1 endpoints and GraphQL queries/mutations.

# 🤔 Why

In order to properly attribute our _Completed Checkout_ server-side events to Optimizely experimental variants, we need to send the `device_distinct_id` to the server so that Optimizely tracking calls can be made using the `device_distinct_id` instead of the user's ID.

# 🛠 How

Added a header, `Kickstarter-iOS-App-UUID` which will be set using the device's `identifierForVendor` value.

**Note:** As mentioned in #1044, according to the Apple [docs](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor), the `identifierForVendor` _can_ be `nil` but this is only when the device has been restarted and not yet unlocked. There should not be any scenario in which we're instantiating our `Service` struct when this value is `nil` but if that were to happen then this value would just coalesce to a random UUID string.

# ✅ Acceptance criteria

Run the app in the simulator and using Charles Proxy observe the following:

- [x] The `Kickstarter-iOS-App-UUID` header is present in all requests to the v1 API and its value is the device's identifier.
- [x] The `Kickstarter-iOS-App-UUID` header is present in all requests to GraphQL and its value is the device's identifier.
- [x] Deleting and reinstalling the app generates a new value.

### DIFF
--- a/KsApi/Service.swift
+++ b/KsApi/Service.swift
@@ -19,6 +19,7 @@ public struct Service: ServiceType {
   public let language: String
   public let currency: String
   public let buildVersion: String
+  public let deviceIdentifier: String
 
   public init(
     appId: String = Bundle.main.bundleIdentifier ?? "com.kickstarter.kickstarter",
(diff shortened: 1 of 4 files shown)


KsApi/Service.swift
```diff
@@ -19,21 +19,24 @@ public struct Service: ServiceType {
   public let language: String
   public let currency: String
   public let buildVersion: String
+  public let deviceIdentifier: String
```
justinswart: In order to not make tons of changes I haven't alphabetized these properties for now. I have plans to do this when we do some refactoring of this type soon.

PR #1046, opened 2020-01-23
[NT-781] Voiceover bug fix

cdolm92:
# 📲 What

Corrected a bug where Voiceover on skips straight to the add card button from the other payment options heading with no interactable elements in between.

# 🤔 Why

Voiceover should be able to read to the user their available payment methods.

# 🛠 How

Set the subviews as accessibility elements in the bindStyles(). There was already an available a11y output that was not being used, `cardNumberAccessibilityLabel`.  This is not the `accessibilityLabel` for the `lastFourLabel` because it reads the last four digits of that card as well as the card brand.

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] When VO is on it should read Labels in the credit card view. The labels should tell user what the card brand is, the last four digits and the expiration date of the card. 
- [ ] VO should also read the Button on the view and inform if it is selected or not and if it is unavailable.
- [ ] If a card is unavailable VO should read the text below.


Kickstarter-iOS/Views/Cells/PledgeCreditCardView.swift
```diff
@@ -108,6 +105,9 @@ final class PledgeCreditCardView: UIView {
   override func bindStyles() {
     super.bindStyles()
 
+    _ = self
```
Scollaco: I don't even think we need these lines. Noticed that we usually don't set `self.subViews` as accessibilityElements. I tested without them and it worked just fine 🤷🏻‍♂️

PR #1051, opened 2020-01-24
[NT-830] Removing user props

ifbarrera:
# 📲 What

We're removing any user props that could potentially give us data that is not up to date, in favor of sending that data from the backend.

# 🤔 Why

We want to populate the user attributes directly from the db instead of relying on the client to have the most up-to-date user information. As a result, we don't need the clients to send these user properties anymore.

# 🛠 How

Just delete them!

# ✅ Acceptance criteria

- [x] verify that the removed user properties no longer send


eoji:
https://github.com/kickstarter/android-oss/pull/727 twinning!

PR #1052, opened 2020-01-24
[NT-776] Add Optimizely Tracking Events

justinswart:
# 📲 What

Adds `Project Page Rewards CTA Tapped` and `Pledge Screen Viewed` to be tracked via Optimizely.

# 🤔 Why

In order to track the success of our _See rewards_ experiment.

# 🛠 How

- Adds these two events and their respective attributes and event tags.
- Added `track(eventKey:userId:attributes:eventTags:)` to `OptimizelyClientType` in order to test using our `MockOptimizelyClient`.

# ✅ Acceptance criteria

- [ ] Confirm that the attributes are correct.
- [ ] Confirm that these events are being tracked to Optimizely.


Kickstarter-iOS/Library/CookieRefTagFunctions.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+import KsApi
+
+private let cookieSeparator = "?"
+private let escapedCookieSeparator = "%3F"
+
+// Extracts the ref tag stored in cookies for a particular project. Returns `nil` if no such cookie has
+// been previously set.
+public func cookieRefTagFor(project: Project) -> RefTag? {
```
justinswart: Moved these functions to a shared place now that we need to access `cookieRefTagFor` in two places. Now that they're shared we may want to add tests for them specifically but I figured we could tackle them separately as this doesn't reduce our test coverage currently.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/OptimizelyClientType.swift
```diff
@@ -32,3 +35,28 @@ extension OptimizelyClientType {
...
+    "location": user?.location?.country.lowercased(),
+    "os_version": AppEnvironment.current.device.systemVersion,
+    "logged_in": user != nil,
+    "chosen_currency": project.stats.currentCurrency,
+    "locale": AppEnvironment.current.locale.identifier
+  ]
+  .compact()
+
+  let eventTags: [String: Any] = [
+    "project_subcategory": project.category.name,
```
justinswart: This _might_ not be the property that we want for this. The spreadsheet mentions "What project category has the user pledged previously?" - I've asked for confirmation.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -217,6 +217,25 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .map(cookieFrom(refTag:project:))
       .skipNil()
       .observeValues { AppEnvironment.current.cookieStorage.setCookie($0) }
+
+    Signal.combineLatest(project, refTag)
+      .takeWhen(ctaButtonTappedWithType.filter { $0 == .pledge })
+      .filter { project, _ in project.personalization.backing == nil }
```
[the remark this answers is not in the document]
justinswart: Yeah probably redundant!

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -130,19 +130,18 @@ public final class PledgeCTAContainerViewViewModel: PledgeCTAContainerViewViewMo
...
-      userId: uuid,
-      isAdmin: AppEnvironment.current.currentUser?.isAdmin ?? false
-    )
-
   guard let projectBacking = backing, project.personalization.isBacking == .some(true) else {
     if currentUserIsCreator(of: project) {
       return PledgeStateCTAType.viewYourRewards
     }
 
+    let optimizelyVariant = AppEnvironment.current.optimizelyClient?
```
ifbarrera: Moved the activation of the experiment to _after_ the `guard` that checks whether the project has already been backed or not - this prevents us from activating the experiment when we know it won't be shown.

Library/ViewModels/ProjectPamphletViewModel.swift
```diff
@@ -217,6 +217,27 @@ public final class ProjectPamphletViewModel: ProjectPamphletViewModelType, Proje
       .map(cookieFrom(refTag:project:))
       .skipNil()
       .observeValues { AppEnvironment.current.cookieStorage.setCookie($0) }
+
+    let shouldTrackCTATappedEvent = ctaButtonTappedWithType
+      .filter { [.pledge, .seeTheRewards, .viewTheRewards].contains($0) }
```
ifbarrera: We weren't actually triggering the event on the experimental variants. Added the here and added tests to make sure we're tracking taps to the button in all three cases.

justinswart: Do you think we should do this or just filter on the project not being backed yet? That way we don't need to update this when variants change 🤔

ifbarrera: I think we need to be explicit about when we track an "impression" for an experimental variant being acted on. If we don't know for sure that a variant has been seen/activated, then we shouldn't track an event in Optimizely.

PR #1057, opened 2020-02-03
[NT-732] Consume category name from v1 API

justinswart:
# 📲 What

Improves deserialization of the parent category from our v1 api so that it includes the category name.

# 🤔 Why

This is information that we would like to include in our tracking properties.

# 🛠 How

- Removed the old workaround for deserializing our `Swift.Decodable` `Category` model within our `Argo.Decodable` `Project` model.
- Split out the v1 `Category` info into a `Project.Category` model.
- Also added `tryDecodable(_:)` described in [the RFC](https://github.com/kickstarter/rfcs/blob/master/rfcs/0069-native-deprecate-argo.md) for deprecating Argo for future use.

# ✅ Acceptance criteria

- [ ] No regressions introduced in current deserialization of `Category` (discover feed, discover filter).
- [ ] Parent category name is deserialized correctly.

# ⏰ TODO

- [ ] Add the parent category name to our tracking properties.


justinswart:
Blocking this as it's still WIP

KsApi/models/Category.swift
```diff
@@ -87,15 +87,57 @@ extension Category {
...
     self.name = try values.decode(String.self, forKey: .name)
-    self.parentId = try? values.decode(String.self, forKey: .parentId)
-    self._parent = try? values.decode(ParentCategory.self, forKey: ._parent)
+    self.parentId = try? Category.tryParentId(decoder)
+    self._parent = try? Category.tryParentCategory(decoder)
     self.subcategories = try? values.decode(SubcategoryConnection.self, forKey: .subcategories)
     self.totalProjectCount = try? values.decode(Int.self, forKey: .totalProjectCount)
   }
+
+  /* Tries to decode the parent category from the expected GraphQL structure
```
justinswart: Not sure how I'm feeling about this. It seems to work but I'm going to take another look tomorrow at what would be involved to just split these two models out completely.

justinswart: Ok, started down a different path and don't really feel like using separate models would be significantly better. Let me know what you think @Scollaco @ifbarrera.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/lib/TryDecodable.swift
```diff
@@ -0,0 +1,22 @@
+import Argo
+import Foundation
+
+public func tryDecodable<T: Swift.Decodable>(_ json: JSON) -> Argo.Decoded<T> {
```
Scollaco: Are we planning to use this in the future? It seems that it's not being used currently.

justinswart: Yeah good call out! I know we don't typically like to do this (and if you feel strongly, I'll remove it), but given that we have a test for it and it's very much our intention to start deprecating Argo using this function I'd like to get it into the codebase. What if I open a PR tomorrow that uses it to deprecate at least one `Argo.Decodable` model? 😄

Scollaco: A new PR sounds good. I can review it. Good Job here!

PR #1063, opened 2020-02-07
Revert session, project, pledge properties for Koala events

ifbarrera:
# 📲 What

Reverts the updates to the session, pledge and project properties for old Koala events. This ensures that Looker will continue to work with old Koala properties as we transition events to Amplitude.

# 🤔 Why

Turns out that changing _any_ property names for events going to Koala means that the event becomes completely unavailable in Looker. This is not ideal for some of our Looker dashboards which we want to continue to work until we migrate our DataLake pipeline to send data to Looker.

# 🛠 How

For events going only to Koala, this PR reverts any session, project or pledge property changes such that the events have exactly the same properties  & property names as they used to. We are also now no longer sending white-listed events to Koala:

- white-listed events will _only_ to go the data lake
- only old events will go to Koala with the same properties as before

# ✅ Acceptance criteria

TBD

- [ ] Environment to test on
- [ ] Issues to look out for


ifbarrera:
We decided to move onwards and upwards by not re-introducing these deprecated properties. We'll be cleaning up all unused events over the next few weeks and migrating all important events to the data lake.

PR #1066, opened 2020-02-12
[NT-858] Replace .filter { ... }.first with .find(where: { ... })

justinswart:
# 📲 What

Replaces instances of `.filter { ... }.first` with `first(where: { ... })`.

# 🤔 Why

This is just a simple optimization to not do unnecessary work when picking the first item from collections. In the case of `filter` the entire collection is traversed and each element is evaluated against the condition whereas `.first(where:)` stops once it finds a match and returns that element.

# 🛠 How

Replaced use of `filter` with `first(where:)` where it made sense. 

For example, there are some places where we do something like:

```swift
```
This could be replaced with:
```swift
.first { $0 is UIPageViewController } as? UIPageViewController
```
I felt that the sacrifice in readability and no material performance gain here didn't justify making those sorts of changes.

# 👀 See

Swift's [Sequence](https://developer.apple.com/documentation/swift/sequence) type.

# ✅ Acceptance criteria

- [ ] No regressions introduced by these changes.

### DIFF
--- a/Library/ViewModels/UpdateDraftViewModel.swift
+++ b/Library/ViewModels/UpdateDraftViewModel.swift
@@ -212,7 +212,7 @@ public final class UpdateDraftViewModel: UpdateDraftViewModelType, UpdateDraftVi
 
     self.showRemoveAttachmentConfirmation = addedAttachments
       .takePairWhen(self.attachmentTappedProperty.signal)
-      .map { attachments, id in attachments.filter { $0.id == id }.first }
+      .map { attachments, id in attachments.first { $0.id == id } }
       .skipNil()
 
     let removeAttachmentEvent: Signal<Signal<UpdateDraft.Image, ErrorEnvelope>.Event, Never> = draft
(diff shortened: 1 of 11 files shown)


Library/ViewModels/UpdateDraftViewModel.swift
```diff
@@ -212,7 +212,7 @@ public final class UpdateDraftViewModel: UpdateDraftViewModelType, UpdateDraftVi
 
     self.showRemoveAttachmentConfirmation = addedAttachments
       .takePairWhen(self.attachmentTappedProperty.signal)
-      .map { attachments, id in attachments.filter { $0.id == id }.first }
+      .map { attachments, id in attachments.first { $0.id == id } }
```
Scollaco: I noticed that in some places we use `.first(where: { .... })` and in others  simply `.first { ... }`. Should we adopt only one to keep it consistent? I don't have strong feelings about it though.

justinswart: Yeah that's intentional, when using `first` in a `guard` statement the compiler isn't able to figure it out unless we add the `where:` argument and parentheses.

PR #1070, opened 2020-02-14
[NT-782] Add context_location to whitelisted events

Scollaco:
# 📲 What
 
Adds `context_location` property to whitelisted DataLake events.

# 🤔 Why

For tracking purposes.
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-782)

# 🛠 How

- Created an `enum` based on this [spreadsheet](https://docs.google.com/spreadsheets/d/1YGDmyCMO5IJQT9Xoso5n6VIHmad4YYW14rKoo_nKGZM/edit#gid=0).
- Added location for each event
- Added tests

# 👀 See

N/A

# ✅ Acceptance criteria
- [ ] With `KOALA_TRACKING` enabled, navigate the app and make sure that `context_location` is being sent with data lake events.

Note: the `Tab Bar Clicked` event is not sending location, but this is intentional.


Library/Tracking/Koala.swift
```diff
@@ -22,38 +22,57 @@ public final class Koala {
...
-    case addNewCardButtonClicked = "Add New Card Button Clicked"
 
     static func allWhiteListedEvents() -> [String] {
       return DataLakeWhiteListedEvent.allCases.map { $0.rawValue }
     }
   }
 
+  /// Determines the screen from which the event is sent.
+  public enum LocationContext: String {
+    case activities = "activity_feed_screen" // ActivitiesViewController <-
```
justinswart: What are the `<-`'s for?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -124,7 +124,8 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
           project: project,
           reward: reward,
           context: pledgeContext,
-          refTag: refTag
+          refTag: refTag,
+          location: .pledgeAddNewCard
```
justinswart: Alphabetize?

Scollaco: Do we need to alphabetize function arguments? 🤔

ifbarrera:
> Hey just a question and comment.
> 
> Was just wondering if this made more sense as an additional argument to each tracking function or if we could just append it to the dictionary of properties that are passed in with specific tracking events. Did we make it an argument because it gets the `context_` prefix? Thoughts @ifbarrera?

Hmmm yeah I feel like because it's part of the `contextProperties` group that it makes sense to pass it in as a `location` argument to the `track()` function. Also I think ideally we should be moving away from manually appending properties to the props dictionary - with the new spec we should really never have to do that because all properties should belong to a property group. It should help keep things neater, more predictable, and consistent throughout all of our tracking events.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
justinswart: I think we've done this as far as possible? Usually if we add a property to a struct we would alphabetize that which in turn would reorder the initializer's arguments. Anyway, if it's not too much hassle it would be good to do here 👍

Library/Tracking/Koala.swift
Scollaco: Oh, I was using this to keep track of locations that had been tested and didn't clean it up 🤦🏻‍♂️.

PR #1080, opened 2020-02-24
[NT-889] (1/3) Landing page ui

Scollaco:
# 📲 What 

. This is the first of a 3 parts work to create our new LandingPage.

# 🤔 Why

We want to tell new users a little more about Kickstarter during their first session.
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-889)

# 🛠 How

- Added LandingPageViewController
- Added LandingPageViewModel and its tests (empty for now)

**Note:** There's no logic to present the viewController yet. This will be done in a different PR. So, for now, the viewController will be presented every time the app launches.

# 👀 See

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] The layout matches the design from [Figma](https://www.figma.com/file/C3gkbRATsDy0vFIUk3Cf8D/FTB-Landing-Screen-Experiment?node-id=44%3A10) (without scrollView).


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -369,6 +372,17 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
 
+    self.goToLandingPage = self.applicationLaunchOptionsProperty.signal.ignoreValues()
+      .takeWhen(self.optimizelyConfiguredWithResultProperty.signal
+        .skipNil()
+        .filter { $0.isSuccess }
+        .ignoreValues()
+      )
+      .observeForUI()
+      .map {
+        LandingPageViewController()
```
Scollaco: This is temporary and will be changed once the experiment and the "first session" logic are implemented (in a different PR).

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -134,6 +134,9 @@ public protocol AppDelegateViewModelOutputs {
   /// Emits when the root view controller should navigate to the creator dashboard.
   var goToDiscovery: Signal<DiscoveryParams?, Never> { get }
 
+  /// Emits when the root view controller should present the Landing Page for new users.
+  var goToLandingPage: Signal<LandingPageViewController, Never> { get }
```
justinswart: This can just be a `Void` or `()` signal and we can instantiate the VC outside of the VM.

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,227 @@
...
+    |> \.translatesAutoresizingMaskIntoConstraints .~ false
+}
+
+private let descriptionLabelStyle: LabelStyle = { label in
+  label
+    |> \.numberOfLines .~ 0
+    |> \.textColor .~ UIColor.ksr_text_dark_grey_500
+    |> \.font .~ UIFont.ksr_callout()
+    |> \.textAlignment .~ .center
+    |> \.text %~ { _ in Strings.Pledge_to_projects_and_view_all_your_saved_and_backed_projects_in_one_place()
```
justinswart: Can move `Strings.Pledge...` onto the next line to match the braces.

Library/ViewModels/LandingPageViewModelTests.swift
```diff
@@ -0,0 +1,18 @@
...
+import Prelude
+import ReactiveExtensions_TestHelpers
+import ReactiveSwift
+import UIKit
+import XCTest
+
+internal final class LandingPageViewModelTests: TestCase {
+  fileprivate let vm: LandingPageViewModelType = LandingPageViewModel()
+
+  fileprivate let landingPageCards = TestObserver<[UIView], Never>()
```
justinswart: I also don't think we should emit actual `UIView`s here, rather the data that will be used to configure the views or an `enum` type representing them.

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,227 @@
...
+    let spacer1 = UIView()
+    let spacer2 = UIView()
+    let spacer3 = UIView()
+
+    _ = ([
+      spacer1,
+      self.labelsStackView,
+      self.cardsStackView,
+      spacer2,
+      spacer3,
```
justinswart: Why the two spacers after each other? Are these just views that need to stretch?

Library/ViewModels/LandingPageViewModelTests.swift
Scollaco: Yes. sorry, I forgot to mention that this was also temporary. On the second PR (containing the cards, the signal emits `[LandingPageCardType]`). So for now, I will delete the property from the view model test 👍

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
Scollaco: The distribution of views of this layout is a bit different of what we usually do in the app (extra spacing on top and bigger space between cards and the button). I tried configuring the views in different ways, but the result was never satisfactory. I'm open to retry if you have strong feelings about it.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
justinswart: Oh I guess I'm just not sure what the two spacers adjacent to each other would do that a single spacer view can't do? How are the spacers receiving their height?

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,228 @@
+import Library
+import Prelude
+import ReactiveSwift
+import UIKit
+
+private enum Layout {
+  enum Button {
+    static let height: CGFloat = 54
```
ifbarrera: Any reason we're not using our `Styles.minTouchSize.height` for buttons here?

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let width: CGFloat = 226
+  }
+
+  enum Card {
+    static let height: CGFloat = 122
+  }
+}
+
+public final class LandingPageViewController: UIViewController {
+  // Properties
```
ifbarrera: `MARK`?

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  private func setupConstraints() {
+    NSLayoutConstraint.activate([
+      self.scrollView.heightAnchor.constraint(greaterThanOrEqualToConstant: Layout.Card.height),
+      self.cardViewsStackView.heightAnchor.constraint(equalTo: self.scrollView.heightAnchor)
+    ])
+  }
+
+  @objc private func ctaBUttonTapped() {
```
ifbarrera: Typo here, should be `ctaButtonTapped`

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+  scrollView
+    |> \.bounces .~ false
+    |> \.isPagingEnabled .~ true
+}
+
+private let titleLabelStyle: LabelStyle = { label in
+  label
+    |> \.text %~ { _ in Strings.Bring_creative_projects_to_life() }
+    |> \.font .~ UIFont.ksr_title3().bolded
+    |> \.textAlignment .~ .center
```
ifbarrera: Do we want to this to wrap with larger font sizes? Right now it's truncating.

Simulator Screen Shot - iPhone X - 2020-02-26 at 13 14 47

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
Scollaco: Oh good catch!

Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift
Scollaco: Without the other spacer, the top view would be too low on iPhone 5s. The heights are being calculated by the stackView

PR #1085, opened 2020-02-26
[NT-871] (3/3) Landing page first session

Scollaco:
# 📲 What

- It adds the logic to only show `LandingPageViewController` on the first app session **_and_** if the user part of `variant-1` or `variant-2` of the Optimizely `native_onboarding_series_new_backers` experiment.

# 🤔 Why

- This is the last part of the Landing Page work.
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-871)

# 🛠 How

- Created the `goToLandingPage` output on the AppDelegateViewModel that emits the experiment variants to determine if the LandingPage should be presented.
- Added the `hasSeenLandingPage` to the KeyValueStore in order to keep track if the user has seen the landing page.
- Added tests

# ✅ Acceptance criteria

- [ ] On a fresh build, open the app and inspect with Charles to see which variant you got.
- [ ] If you got `control`, the Landing Page shouldn't be presented
- [ ] If you got `variant-1` or `variant-2`, the LandingPage should be presented
- [ ] If you saw the LandingPage, tap the `Get started` button. The ViewController should be dismissed and the Landing Page should no longer be shown.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -1089,7 +1089,13 @@ private func qualtricsConfigData() -> QualtricsConfigData {
 }
 
 private func shouldGoToLandingPage() -> Bool {
-  guard AppEnvironment.current.currentUser == nil else {
+  let hasNotSeenLandingPage = [
+    AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
+    AppEnvironment.current.userDefaults.hasSeenLandingPage
+  ]
+  .allSatisfy(isFalse)
+
+  guard AppEnvironment.current.currentUser == nil, hasNotSeenLandingPage else {
     AppEnvironment.current.ubiquitousStore.hasSeenLandingPage = true
     AppEnvironment.current.userDefaults.hasSeenLandingPage = true
 
@@ -1109,14 +1115,10 @@ private func shouldGoToLandingPage() -> Bool {
       isAdmin: AppEnvironment.current.currentUser?.isAdmin ?? false,
       userAttributes: userAttributes
     )
-  
+
   switch optimizelyVariant {
   case .variant1, .variant2:
-    return [
-      AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
-      AppEnvironment.current.userDefaults.hasSeenLandingPage
-    ]
-    .allSatisfy(isFalse)
+    return hasNotSeenLandingPage
   case .control, nil:
     return false
   }
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -1090,13 +1090,11 @@ private func qualtricsConfigData() -> QualtricsConfigData {
 
 private func shouldGoToLandingPage() -> Bool {
   let hasNotSeenLandingPage = [
-    AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
     AppEnvironment.current.userDefaults.hasSeenLandingPage
   ]
   .allSatisfy(isFalse)
 
   guard AppEnvironment.current.currentUser == nil, hasNotSeenLandingPage else {
-    AppEnvironment.current.ubiquitousStore.hasSeenLandingPage = true
     AppEnvironment.current.userDefaults.hasSeenLandingPage = true
 
     return false
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -1116,8 +1116,11 @@ private func shouldGoToLandingPage() -> Bool {
 
   switch variant {
   case .variant1, .variant2:
-    return !AppEnvironment.current.ubiquitousStore.hasSeenLandingPage &&
-      !AppEnvironment.current.userDefaults.hasSeenLandingPage
+    return [
+      AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
+      AppEnvironment.current.userDefaults.hasSeenLandingPage
+    ]
+    .allSatisfy(isFalse)
   case .control:
     return false
   }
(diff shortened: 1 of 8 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1083,3 +1087,38 @@ private func qualtricsConfigData() -> QualtricsConfigData {
...
+      userAttributes: userAttributes
+    )
+
+  guard let variant = optimizelyVariant else {
+    return false
+  }
+
+  switch variant {
+  case .variant1, .variant2:
+    return !AppEnvironment.current.ubiquitousStore.hasSeenLandingPage &&
```
justinswart: Nice way of writing this could be:
```swift
  case .variant1, .variant2:
    return [
      AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
      AppEnvironment.current.userDefaults.hasSeenLandingPage
    ]
    .allSatisfy(isFalse)
```

Scollaco: Oh nice!

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/LandingPageViewController.swift | indent |
| Library/ViewModels/LandingPageViewModel.swift | blankLinesAtStartOfScope, indent |


</p>

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1083,3 +1087,41 @@ private func qualtricsConfigData() -> QualtricsConfigData {
...
+  }
+
+  switch variant {
+  case .variant1, .variant2:
+    return [
+      AppEnvironment.current.ubiquitousStore.hasSeenLandingPage,
+      AppEnvironment.current.userDefaults.hasSeenLandingPage
+    ]
+    .allSatisfy(isFalse)
+  case .control:
```
justinswart: For interest sake - if you wanted to you could lose the `guard` above and change this to `case .control, nil` to cover the `nil` case.

But fine either way 👍

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1083,3 +1087,41 @@ private func qualtricsConfigData() -> QualtricsConfigData {
     .compact()
   )
 }
+
+private func shouldGoToLandingPage() -> Bool {
+  guard AppEnvironment.current.currentUser == nil else {
```
ifbarrera: Won't `AppEnvironment.current.currentUser` be nil if the user is logged out? In which case we might be unnecessarily activating an experiment for users that have already seen the landing page?

We might want to check the ubiquitous store as the first step so we can break early if we know the user has already seen the landing page, in which case we don't need to check the optimizely experiment.

Library/ViewModels/LandingPageViewModel.swift
```diff
@@ -22,15 +24,27 @@ public final class LandingPageViewModel: LandingPageViewModelType, LandingPageVi
       .map(cards)
       .skipNil()
 
+    self.dismissViewController = self.ctaButtonTappedSignal
```
ifbarrera: If the user closes out of the app without closing the landing page, the user defaults value will never be set, so they'll see the landing page again the next time they open the app. I'm not sure this is the desired behavior - we might want to immediately set the user defaults value as soon as the landing page screen loads. You might want to confirm with product on this.

Library/ViewModels/LandingPageViewModelTests.swift
```diff
@@ -6,17 +6,41 @@ import UIKit
...
+
+    self.viewModel.inputs.ctaButtonTapped()
+
+    self.dismissViewController.assertValueCount(1)
+  }
+
+  func testUserDefaultsUpdates_OnButtonTap() {
+    let userDefaults = MockKeyValueStore()
+
+    withEnvironment(currentUser: nil, userDefaults: userDefaults) {
```
ifbarrera: Should we also test that the ubiquitous store value has been set?

ifbarrera: Actually, come to think of it, should we even be setting the ubiquitous store value? From my understanding, the ubiquitous store will persist across app installs, which means if the user sees the landing page, then deletes the app, they will never see the landing page again even when they re-install the app. Is this desired behavior?

Library/ViewModels/LandingPageViewModel.swift
Scollaco: Yes, I thought about that while developing. I'd decided to set the values on the button action to make sure that the users saw and interacted with the page (tapped the button). I will check with product, though.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
Scollaco: Hmm good point. I will change that 👍

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift | indent |


</p>

Library/ViewModels/LandingPageViewModelTests.swift
Scollaco: The experiment briefing says that the Landing Page should be only presented on the first session of the user. I guess this is another decision for product. Will keep you updated :)

Scollaco: I removed the piece of code that persists the value on the ubiquitous store. Now we are considering UserDefaults.

Library/ViewModels/LandingPageViewModel.swift
Scollaco: After standup today, we've decided to set `hasSeenLandingPage = true` when the view is presented.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1083,3 +1087,37 @@ private func qualtricsConfigData() -> QualtricsConfigData {
     .compact()
   )
 }
+
+private func shouldGoToLandingPage() -> Bool {
+  let hasNotSeenLandingPage = [
+    AppEnvironment.current.userDefaults.hasSeenLandingPage
+  ]
+  .allSatisfy(isFalse)
```
ifbarrera: You probably don't need the array now that we're only checking on one thing.

Library/ViewModels/LandingPageViewModel.swift
```diff
@@ -22,15 +24,30 @@ public final class LandingPageViewModel: LandingPageViewModelType, LandingPageVi
...
+
     self.numberOfPages = self.landingPageCards
       .map(\.count)
+
+    // Koala
```
ifbarrera: I don't think we need this comment?

PR #1092, opened 2020-03-04
[NT-946] Personalization/onboarding experiment logic

ifbarrera:
# 📲 What

Adds the logic in the `AppDelegateViewModel` for displaying the personalization/onboarding screen if the user is eligible and in the experimental variant for the experiment.

# 🤔 Why

Since this onboarding flow will be run as an experiment, we want to check whether the user is in the experimental variant and only then show the onboarding screen. Additionally, the onboarding screen should only be shown once.

# 🛠 How

- added an output for showing the onboarding/personalization screen
- logic checks user defaults or previously set flag, and then checks Optimizely to get the user's experiment variant
- all deeplinks triggered via notification/handoff/shortcut/universal links are intercepted. If the user has not seen the onboarding page but is eligible to see it, the deeplink will not resolve in order to allow the onboarding page to show

# ✅ Acceptance criteria

In `Service.swift`, set the default serverConfig value to `.staging`. This is required because when the app launches clean for the first time, it needs to be set to staging so that we can access the staging Optimizely configuration.

image

- [x] on a fresh build (no previous build installed on device/sim), run the app. You should see the category selection screen presented after a short moment. If you run the app again, the screen should not display.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -125,6 +125,9 @@ public protocol AppDelegateViewModelOutputs {
   /// Emits when the root view controller should navigate to activity.
   var goToActivity: Signal<(), Never> { get }
 
+  /// Emits when the root view controller should navigate to the onboarding flow
+  var goToCategoryPersonalizationOnboarding: Signal<Void, Never> { get }
+
   /// Emits when application should navigate to the creator's message thread
   var goToCreatorMessageThread: Signal<(Param, MessageThread), Never> { get }
 
@@ -148,9 +151,6 @@ public protocol AppDelegateViewModelOutputs {
 
   /// Emits a URL when we should open it in the safari browser.
   var goToMobileSafari: Signal<URL, Never> { get }
-
-  /// Emits when the root view controller should navigate to the onboarding flow
-  var goToOnboarding: Signal<Void, Never> { get }
 
   /// Emits when the root view controller should navigate to search.
   var goToSearch: Signal<(), Never> { get }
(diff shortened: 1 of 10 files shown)


Library/OptimizelyExperiment.swift
```diff
@@ -3,6 +3,7 @@ import Foundation
 public enum OptimizelyExperiment {
   public enum Key: String {
     case pledgeCTACopy = "pledge_cta_copy"
+    case categoryPersonalization = "onboarding_category_personalization_flow"
```
justinswart: With our other experiments we've camelCased the snake_cased value as the case name.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1087,3 +1094,31 @@ private func qualtricsConfigData() -> QualtricsConfigData {
...
+  let hasSeenCategoryPersonalization = AppEnvironment.current.userDefaults.hasSeenCategoryPersonalizationFlow
+
+  if isLoggedIn || hasSeenCategoryPersonalization {
+    // Currently logged-in users should not see the onboarding flow
+    AppEnvironment.current.userDefaults.hasSeenCategoryPersonalizationFlow = true
+
+    return false
+  }
+
+  let variant = AppEnvironment.current.optimizelyClient?.variant(
```
justinswart: So this will still load async and we may briefly see the discover feed?

ifbarrera: That's right!

Library/OptimizelyExperiment.swift
ifbarrera: It's a little verbose, but ok.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #1095, opened 2020-03-04
[NT- 956] CreatorByline View Model and Tests

cdolm92:
# 📲 What

Implements the `CreatorBylineViewModel` and tests to display correct creator stats.

# 🤔 Why

This work is part of our experimentation to test whether we can increase conversion from the campaign details page.

# 🛠 How

We are now using the `ProjectCreatorDetailsEnvelope` to display the # of projects the creator backed and created.

# 👀 See

Simulator Screen Shot - iPhone 11 2 - 2020-03-05 at 11 11 29

# ✅ Acceptance criteria

**NOTE:** Remember to display the `creatorBylineView` in the `configureWith()` of the  `ProjectPamphletMainCell`

- [ ] Navigate to a project, make sure the creator details are displayed correctly.


Kickstarter-iOS/Views/CreatorBylineView.swift
```diff
@@ -5,6 +5,7 @@ import UIKit
 
 private enum Layout {
   enum ImageView {
+    static let minHeight: CGFloat = 36.0
```
justinswart: This is unused now.

Kickstarter-iOS/Views/CreatorBylineView.swift
```diff
@@ -94,6 +96,7 @@ final class CreatorBylineView: UIView {
 
     _ = self.verifiedCheckmarkImageView
       |> \.clipsToBounds .~ true
+      |> \.contentMode .~ .scaleToFill
```
justinswart: Don't think this is needed, think we removed it previously?

KsApi/models/templates/ProjectCreatorDetailsEnvelopeTemplate.swift
```diff
@@ -0,0 +1,9 @@
+import Prelude
+
+extension ProjectCreatorDetailsEnvelope {
+  internal static let template = ProjectCreatorDetailsEnvelope(
+    backingsCount: 10,
+    id: "",
```
justinswart: Thanks for adding the template, we could probably give it some arbitrary value for `id` like `"123"`.

Library/ViewModels/CreatorBylineViewModel.swift
```diff
@@ -23,21 +26,32 @@ public final class CreatorBylineViewModel: CreatorBylineViewModelType,
...
     let project = self.projectProperty.signal.skipNil()
+    let creatorDetails = self.creatorDetailsProperty.signal.skipNil()
+      .map { $0 }
 
     self.creatorLabelText = project.map {
       Strings.project_creator_by_creator(creator_name: $0.creator.name)
     }
 
     self.creatorImageUrl = project.map { URL(string: $0.creator.avatar.small) }
+    self.creatorStatsText = creatorDetails.map { creatorDetails in
```
justinswart: I think if the launched projects count is `1` this text should emit "First-time creator" for the first part of the string.

cdolm92: The ACs changed we are no longer using that copy.

colleenmacd: correct! we are using "X created • X backed" due to Vicki's recommendation. A slack convo changed the AC's and i tried to update all of the tickets I found to reflect this

PR #1103, opened 2020-03-11
[NT-905] Category Pill Selection

ifbarrera:
# 📲 What

Adds the pill selection functionality that allows users to select and de-select category pills from the `CategorySelectionViewController`.

# 🤔 Why

We want users to be able to select and de-select the categories that interest them, so that we may display relevant, curated projects to them during this onboarding flow.

# 🛠 How

This PR also:
- moves the logic that determines whether the category selection flow has been "seen" from the `CategorySelectionViewModel` to the `LandingViewModel` since that is the first view controller in the flow
- adds screenshot tests for `CategorySelectionViewController`
- refactors the category selection pills out into their own class `CategoryPillCell` to reduce complexity. This also removes the need to have `PillCellStyle` which cleans up the view model tests quite a bit.
- updates the `CategoryPillCell` to use a button instead of a `UILabel` for better out-of-the-box accessibility support and tap-states

# 👀 See

NwwNEz0j2k

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

In the `AppDelegateViewModel.swift` file, comment out the body of the `shouldSeeCategoryPersonalization` function and replace it with an early `return true`:

image

This ensures that you'll always see the category personalization flow an app launch.

- [ ] Launch the app, and tap "Get started" to navigate to the category selection screen. You should be able to tap categories to select them, and tap again to de-select. Scrolling should remember any previously selected categories.


Kickstarter-iOS/DataSources/CategorySelectionDataSource.swift
```diff
@@ -5,23 +5,40 @@ import UIKit
 
 internal final class CategorySelectionDataSource: ValueCellDataSource {
   private var categorySectionTitles: [String] = []
-  func load(_ sectionTitles: [String], categories: [[(String, PillCellStyle)]]) {
+
+  weak var collectionView: UICollectionView?
```
ifbarrera: This reference is used to set the `width` constant on the pill cells.

Library/ViewModels/CategorySelectionViewModel.swift
```diff
@@ -23,9 +26,6 @@ public final class CategorySelectionViewModel: CategorySelectionViewModelType,
   CategorySelectionViewModelInputs, CategorySelectionViewModelOutputs {
   public init() {
     let categoriesEvent = self.viewDidLoadProperty.signal
-      .on { _ in
```
ifbarrera: This is moved to `LandingViewModel.swift` now.

Library/ViewModels/LandingViewModel.swift
```diff
@@ -15,7 +16,12 @@ public protocol LandingViewModelType {
 }
 
 public final class LandingViewModel: LandingViewModelType, LandingViewModelInputs, LandingViewModelOutputs {
+  private let hasSeenCategoryPersonalizationFlowProperty = MutableProperty(false)
```
ifbarrera: This property is just here to make sure that our side-effect closure doesn't get deallocated. Since it's not used outside the scope of this view model, it's marked as `private`.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/LandingViewModel.swift
Scollaco: I think here you can simply do: 
```
self.viewDidLoadProperty.signal
      .observeValues {
        AppEnvironment.current.userDefaults.hasSeenCategoryPersonalizationFlow = true
      }
```

Kickstarter-iOS/Views/Cells/CategoryPillCell.swift
```diff
@@ -0,0 +1,113 @@
...
+
+  private func configureSubviews() {
+    _ = (self.button, self.contentView)
+      |> ksr_addSubviewToParent()
+      |> ksr_constrainViewToEdgesInParent(priority: .defaultHigh)
+
+    self.buttonWidthConstraint = self.button.widthAnchor.constraint(lessThanOrEqualToConstant: 0)
+
+    NSLayoutConstraint.activate([
+      self.buttonWidthConstraint!,
```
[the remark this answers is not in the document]
ifbarrera: Added `compact()` so we don't have to unwrap.

Kickstarter-iOS/Views/Controllers/CategorySelectionViewController.swift
```diff
@@ -213,16 +235,15 @@ extension CategorySelectionViewController: UICollectionViewDelegateFlowLayout {
...
+
+extension CategorySelectionViewController: CategoryPillCellDelegate {
+  func categoryPillCell(_: CategoryPillCell, didTapAtIndex index: IndexPath, action: (Bool) -> ()) {
+    self.viewModel.inputs.categorySelected(at: index)
+
+    let shouldSelectCell = self.viewModel.outputs.shouldSelectCell(at: index)
 
-    _ = pillCell.label
-      |> \.preferredMaxLayoutWidth .~ collectionView.bounds.width
+    action(shouldSelectCell)
```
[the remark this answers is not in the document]
ifbarrera: You're right! Since we pass the cell's reference to the delegate we can just call `setIsSelected(shouldSelectCell)` directly on it. Will update!

PR #1115, opened 2020-03-19
[NT-954] Personalization Explore Card

ifbarrera:
# 📲 What

Displays a callout card at the top of the Explore feed prompting users to view curated projects based on their chosen categories during the onboarding flow.

# 🤔 Why

To give users a way to return to their curated projects based on their preferred categories selected during the onboarding flow.

# 🛠 How

- adds `PersonalizationCell` that is populated into the `.personalization` section in the Discovery feed
- the section is only shown on the default filter: for logged out users, that's the `All Projects` filter, and for logged in users it's either the `Recommended for you` filter or the `All Projects` filter depending on your recommendation preferences
- the section is displayed only if you've completed the onboarding flow, and if you haven't dismissed the card: both of these preferences are stored in `.userDefaults`
- the section is displayed only if you're in `.variant1` of the experiment
- tapping the cell presents the `CuratedProjectsViewController`
- tapping the `X` on the cell will dismiss it and it will not show again unless you re-install the app

# 👀 See

| Dismissing the cell | Screenshot |
| --- | --- |
| c0Ea4sDfq0 | Simulator Screen Shot - iPhone 8 - 2020-03-19 at 11 50 21 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

Prerequisites:
- comment out the `shouldSeeCategoryPersonalization` function body and instead `return true`: this will guarantee that you will be put into the onboarding flow on app launch
image
- delete the app from the simulator and launch a fresh build

- Go through the onboarding flow, selecting 5 categories from the category selection screen. Hit "Continue", and the "Done" on the curated projects screen.
  - [x] You should see the personalization cell appear in Explore
  - [x] tapping the cell should present the curated projects screen
  - [x] if you switch filters, the cell should not show
  - [x] tapping the "X" dismisses the cell, and on app-relaunch it does not re-appear

# ⏰ TODO

- [ ] string translations


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/OptimizelyClientType.swift
```diff
@@ -56,7 +56,7 @@ public func optimizelyTrackingAttributesAndEventTags(
 }
 
 public func optimizelyUserAttributes(
-  with user: User? = nil,
+  with user: User? = AppEnvironment.current.currentUser,
```
justinswart: Yeah was thinking this probably doesn't need to be an argument, we can just reference the environment directly. Also passing in `isAdmin` in the `variant` function is probably unnecessary. But we can address separately.

Library/ViewModels/CategorySelectionViewModelTests.swift
```diff
@@ -377,6 +379,38 @@ final class CategorySelectionViewModelTests: TestCase {
...
+
+      self.vm.inputs.categorySelected(with: (artIndexPath, artId))
+      self.vm.inputs.categorySelected(with: (illustrationIndexPath, illustrationId))
+
+      self.postNotification.assertDidNotEmitValue()
+      XCTAssertFalse(mockKVStore.hasCompletedCategoryPersonalizationFlow)
+
+      self.vm.inputs.continueButtonTapped()
+
+      self.postNotification.assertValueCount(1)
```
justinswart: Can we test which notification was posted?

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -351,6 +372,34 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
+    )
+
+    self.showPersonalization = Signal.merge(
+      viewReady,
+      viewReady.takeWhen(self.onboardingCompletedProperty.signal)
+    )
+    .map(second)
+    .map { shouldShowHeader in
+      if shouldShowHeader {
+        return shouldShowPersonalization()
```
justinswart: What if we passed the `shouldShowHeader` boolean into the `shouldShowPersonalization()` function and then just had a single `return`?

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -499,6 +571,36 @@ private func saveSeen(activities: [Activity]) {
...
+  let variant = AppEnvironment.current.optimizelyClient?
+    .variant(
+      for: .onboardingCategoryPersonalizationFlow,
+      userId: deviceIdentifier(uuid: UUID()),
+      isAdmin: AppEnvironment.current.currentUser?.isAdmin ?? false,
+      userAttributes: userAttributes
+    )
+
+  switch variant {
+  case .control:
```
justinswart: Can we do `case .control, .variant2:`?

Library/ViewModels/CategorySelectionViewModelTests.swift
ifbarrera: Ah yeah thanks for catching!

Library/ViewModels/DiscoveryPageViewModel.swift
ifbarrera: They're kinda separate things - `shouldShowPersonalization` deals purely with the personalization experiment logic, while `shouldShowHeader` is more generally about whether we're in a discovery category that allows headers (even editorial headers) to be shown.

ifbarrera: I could maybe simplify this to `return shouldShowHeader && shouldShowPersonalization()` though.

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -499,6 +567,36 @@ private func saveSeen(activities: [Activity]) {
...
+      isAdmin: AppEnvironment.current.currentUser?.isAdmin ?? false,
+      userAttributes: userAttributes
+    )
+
+  switch variant {
+  case .control, .variant2:
+    return false
+  case .variant1:
+    return true
+  default:
```
justinswart: Don't think you need the `default` case anymore.

PR #1117, opened 2020-03-19
[NT-962] Curated Projects Query

Scollaco:
# 📲 What

- Loads a list of projects from selected categories.

# 🤔 Why

- This is part of the Onboarding Personalization experiment.
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-962) 

# 🛠 How

- On `CategorySelectionViewController`, changed the type of `goToCuratedProjects` signal to emit an array of categories instead of an array of IDs. (We need the Category object in order to set the right DiscoverParams before calling the `fetchDiscovery` function).
- On `CuratedProjectsViewModel`, I created 2 functions
   - One that gets the array of categories and returns `[SignalProducer<[Project], Never>] `
   - And a second one that combine the latest values of the function above. This will make sure that the function will only return once we have the projects for each category:
```
return SignalProducer.combineLatest(producers(from: categories))
```
- Added tests

# 👀 See

curated_projects

# ✅ Acceptance criteria

On `AppDelegateViewModel`, add `return true` below the `shouldSeeCategortyPersonalization()` function signature. This will force the Onboard flow to be presented.

Screenshot 2020-03-18 at 15 08 12

- [x]  Select up to 5 categories on the `CategorySelectionViewController` and tap `Continue`. A list of projects should appear.

**Note:** The total of projects is always 30 if all the calls succeeds and have enough projects. If any call fails of doesn't have enough live projects, we return an empty array for that category.


Library/ViewModels/CuratedProjectsViewModelTests.swift
```diff
@@ -0,0 +1,47 @@
...
+  private let loadProjects = TestObserver<[Project], Never>()
+  private let viewModel: CuratedProjectsViewModelType = CuratedProjectsViewModel()
+
+  override func setUp() {
+    super.setUp()
+    self.viewModel.outputs.dismissViewController.observe(self.dismissViewController.observer)
+    self.viewModel.outputs.loadProjects.observe(self.loadProjects.observer)
+  }
+
+  func testDismissViewController_OnBUttonTap() {
```
ifbarrera: Typo here, should be `_OnButtonTap`

Library/ViewModels/CuratedProjectsViewModelTests.swift
```diff
@@ -0,0 +1,47 @@
...
+      .template
+        |> Project.lens.id .~ $0
+        |> Project.lens.category .~ category
+    }
+    let envelope = .template
+      |> DiscoveryEnvelope.lens.projects .~ projects
+
+    let apiService = MockService(fetchDiscoveryResponse: envelope)
+    withEnvironment(apiService: apiService) {
+      self.viewModel.inputs.configure(with: [Category.art])
```
ifbarrera: Can we add a test for passing in multiple categories? Also one for the case where one request fails?

Library/ViewModels/CuratedProjectsViewModel.swift
```diff
@@ -0,0 +1,77 @@
...
+public protocol CuratedProjectsViewModelType {
+  var inputs: CuratedProjectsViewModelInputs { get }
+  var outputs: CuratedProjectsViewModelOutputs { get }
+}
+
+public final class CuratedProjectsViewModel: CuratedProjectsViewModelType, CuratedProjectsViewModelInputs,
+  CuratedProjectsViewModelOutputs {
+  public init() {
+    let curatedProjects: Signal<[Project], Never> = self.categoriesSignal
+      .takeWhen(self.viewDidLoadSignal.ignoreValues())
```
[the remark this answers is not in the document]
Scollaco: It makes sense. Will change that

Library/ViewModels/CuratedProjectsViewModelTests.swift
Scollaco: Yes, I will add passing multiple categories. We have a ticket to handle errors, so the other test case can be added on that PR.

Library/ViewModels/CategorySelectionViewModel.swift
```diff
@@ -216,3 +218,20 @@ private func updatedSelectedValues<T: Hashable>(selectedValues: Set<T>, currentV
...
+
+private func selectedCategories(_ categories: [KsApi.Category], with ids: Set<Int>) -> [KsApi.Category] {
+  var selectedCategories: [KsApi.Category] = []
+  let subcategories = categories
+    .compactMap { $0.subcategories?.nodes }
+    .flatMap { $0 }
+
+  ids.forEach { id in
+    if let selectedCategory = categories.first(where: { $0.intID == id }) {
+      selectedCategories.append(selectedCategory)
```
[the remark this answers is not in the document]
Scollaco: I ended up changing the `forEach` to use a `for in`. This way I could use `continue` inside the loop.

Library/ViewModels/CategorySelectionViewModel.swift
```diff
@@ -70,8 +70,10 @@ public final class CategorySelectionViewModel: CategorySelectionViewModelType,
 
     self.goToCuratedProjects = selectedCategoryIds
       .takeWhen(self.continueButtonTappedProperty.signal)
+      .map { categories, ids -> [KsApi.Category] in
```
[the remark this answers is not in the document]
Scollaco: Weird that the last code I push is: 
```
self.goToCuratedProjects = orderedCategories
      .combineLatest(with: selectedCategoryIds)
      .takeWhen(self.continueButtonTappedProperty.signal)
      .map { categories, ids -> [KsApi.Category] in
        selectedCategories(categories, with: ids)
      }
      .sort { $0.name < $1.name }
```

Library/ViewModels/CuratedProjectsViewModelTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    let envelope = .template
+      |> DiscoveryEnvelope.lens.projects .~ (artProjects + gamesProjects)
+
+    let apiService = MockService(fetchDiscoveryResponse: envelope)
+    withEnvironment(apiService: apiService) {
+      self.viewModel.inputs.configure(with: [Category.art, Category.tabletopGames])
+
+      self.viewModel.inputs.viewDidLoad()
+
+      self.loadProjects.assertValues([projects, projects + projects])
```
ifbarrera: Why is `loadProjects` firing twice here? Shouldn't it just fire once with the sum of the two individual network requests?

Library/ViewModels/CategorySelectionViewModel.swift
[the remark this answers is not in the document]
Scollaco: Weird that the changes were not appearing here! ☝️ that's what we want

Library/ViewModels/CuratedProjectsViewModelTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    let gamesProjects = (1...5).map {
+      .template
+        |> Project.lens.id .~ $0
+        |> Project.lens.category .~ Project.Category.games
+    }
+    let projects = (artProjects + gamesProjects)
+
+    let envelope = .template
+      |> DiscoveryEnvelope.lens.projects .~ (artProjects + gamesProjects)
```
ifbarrera: This doesn't exactly reflect what we expect to happen. What we expect to happen is for two independent network requests to be made with the responses being `projects .~ artProjects` and `projects .~ gamesProjects`. Since we don't have a way currently to mock out different responses on the same endpoint, I would say it's enough to check that the number of projects returned to `loadProjects` is correct, because that will tell us how many times the request was made. In that case, all you need is something like this:

```    
  let projects = (1...5).map {
      .template
        |> Project.lens.id .~ $0
    }

    let envelope = .template
      |> DiscoveryEnvelope.lens.projects .~ projects

    let apiService = MockService(fetchDiscoveryResponse: envelope)

    withEnvironment(apiService: apiService) {
      self.viewModel.inputs.configure(with: [Category.art, Category.tabletopGames])

      self.loadProjects.assertValues([projects + projects]) // 2x the projects, shows request was made 2x
  }
```

Library/ViewModels/CuratedProjectsViewModelTests.swift
Scollaco: That's because `scan` was emitting partial values. Replacing it with `reduce` seems to fix this.

PR #1119, opened 2020-03-24
[NT-963] Curated projects loading indicator

Scollaco:
# 📲 What

Shows a loading indicator on the `CuratedProjectsViewController` while fetching projects.

# 🤔 Why

To improve UX by showing to users that a request is being made.
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-963)

# 🛠 How

- Added a indicator to the view hierarchy.
- Created an output `isLoading` on the view model.
- Sets the `isLoading` output to `true` on viewDidLoad and to `false` when we get a response back from the server.
- Added tests

# 👀 See

cur_proj_loading

# ✅ Acceptance criteria
On `AppDelegateViewModel`, add `return true` below the `shouldSeeCategortyPersonalization()` function signature. This will force the Onboard flow to be presented.

Screenshot 2020-03-18 at 15 08 12

- [ ] Select few categories and tap the `Continue` button. The CuratedProjectsViewController should be presented with an activity indicator. 
- [ ] The indicator should disappear once the projects are loaded


Library/ViewModels/CuratedProjectsViewModel.swift
```diff
@@ -32,6 +33,12 @@ public final class CuratedProjectsViewModel: CuratedProjectsViewModelType, Curat
     self.loadProjects = curatedProjects
 
     self.dismissViewController = self.doneButtonTappedSignal
+
+    self.isLoading = Signal.merge(
+      self.viewDidLoadSignal.mapConst(true),
+      curatedProjects.mapConst(false)
+        .skipRepeats()
```
ifbarrera: Don't you want the `skipRepeats` to be outside of the `.merge`?

ifbarrera: Actually why do you need the `.skipRepeats`? 🤔

Scollaco: Oh that was because when I used the `scan` function it was emitting more than once. Now that I replaced it with `reduce`, it's no longer necessary. Good catch!

PR #1120, opened 2020-03-24
[NT-877] Read more about this campaign tracking bugfix

justinswart:
# 📲 What

Fixes a tracking event that wasn't being sent to Optimizely when `Read more about this campaign` was tapped.

# 🤔 Why

Tracking this event is necessary for our metrics.

# 🛠 How

The bug was that we were checking if `project.personalization.isBacking` was `false`. In this case `isBacking` can also be `nil` so comparing a `nil` value to `false` fails. We could coalesce this `nil` to `false` but I figured we can also just check to see whether the actual `backing` property is `nil` for the same result.

# ✅ Acceptance criteria

Place yourself in variant 1 or 2 for `native_project_page_campaign_details`. Open Charles Proxy.

- [x] Navigate to a project page, tap on the `Read more about this campaign` button. Observe that an event was tracked for `Campaign Details Button Clicked`.

### DIFF
--- a/Library/ViewModels/ProjectPamphletMainCellViewModel.swift
+++ b/Library/ViewModels/ProjectPamphletMainCellViewModel.swift
@@ -328,9 +328,8 @@ public final class ProjectPamphletMainCellViewModel: ProjectPamphletMainCellView
 
     let shouldTrackCTATappedEvent = projectAndRefTag
       .takeWhen(self.readMoreButtonTappedProperty.signal)
-      .logEvents(identifier: "***")
       .filter { project, _ in
-        project.state == .live && project.personalization.backing == nil
+        project.state == .live && userIsBackingProject(project) == false
       }
 
     // optimizely tracking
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ProjectPamphletMainCellViewModel.swift
```diff
@@ -328,7 +328,10 @@ public final class ProjectPamphletMainCellViewModel: ProjectPamphletMainCellView
 
     let shouldTrackCTATappedEvent = projectAndRefTag
       .takeWhen(self.readMoreButtonTappedProperty.signal)
-      .filter { project, _ in project.state == .live && project.personalization.isBacking == false }
+      .logEvents(identifier: "***")
+      .filter { project, _ in
+        project.state == .live && project.personalization.backing == nil
```
ifbarrera: Ahhh this is such a common pitfall for us. Do you think it could be worthwhile to add a helper function to `SharedFunctions.swift` to handle all the logical cases for this? Something like

```
func currentUserHasBacked(project: Project) {
    return project.personalization.backing != nil || project.personalization.isBacking == .some(true)
}
```
To be honest I'm not entirely sure that `personalization.isBacking` means what we think it does even though we use it all over the place 🤔.

justinswart: Sure I could do that. It looks like I also missed a `logEvents` still in the code there 😂

PR #1122, opened 2020-03-25
[NT-1019] Cache selected categories in NSCache

ifbarrera:
# 📲 What

Caches categories selected during the onboarding flow in NSCache. Also refactors how we keep track of selected categories by passing in the whole `Category` object into the cells so they can be passed back with the delegate callback. This improves performance because we don't have to iterate through nested arrays.

# 🤔 Why

The `CuratedProjectsViewController` needs to be configured with an array of `KsApi.Category`s. This is because the `DiscoveryParams` object which is used to populate the query parameters of the `/discover` requests takes a `Category` object instead of a `categoryId`.

When presenting the `CuratedPrjoectsViewController` from the explore screen, we need to pass in an array of previously selected `Category` objects. This array will now come from an `NSCache`.

# 🛠 How

- cache the selected array of `Category`s in our `NSCache` when the user taps "Continue" on the `CategorySelectionViewController`
- retrieve the cached array in `DiscoveryPageViewModel` when the user taps the personalization card
- also refactors the data source of the `CategorySelectionViewController` screen to take an array of `(displayName: String, category: KsApi.Category)` so that the category can be passed back to the `CategorySelectionViewModel` upon selection. This means we don't need a function to transform selected `categoryIds` into category objects, which improves performance
- changes the way we check the Optimizely experiment variant in `DiscoveryPageViewModel`: instead of calling `variant(for:)`, which uses `activate` under the hood, this PR adds a new helper function called `getVariation` which calls `getVariation` on the Optimizely SDK under the hood. This will not trigger an experiment impression.

# 👀 See

DFjOjRPo7O

# ✅ Acceptance criteria

- [x] Delete any old builds from the simulator. Then, run the app. You should see the new onboarding screen. Navigate to the category selection screen and select a few categories. Tap "Continue". Then tap "Done" on the curated projects screen. You should see the personalization cell at the top of discovery. Tap the cell, and you should see the curated projects screen presented, with projects populated from the categories you selected during the onboarding flow.

### DIFF
--- a/Library/OptimizelyClientType.swift
+++ b/Library/OptimizelyClientType.swift
@@ -36,6 +36,31 @@ extension OptimizelyClientType {
 
     return variant
   }
+
+  /*
+   Calls `getVariation` on the Optimizely SDK for the given experiment,
+   using the default attributes and deviceId
+
+   Does *not* record an Optimizely impression. If you wish to record an experiment impression, use
+   `variant(for experiment)`
+   */
+
+  public func getVariation(for experiment: OptimizelyExperiment.Key) -> OptimizelyExperiment.Variant {
+    let userId = deviceIdentifier(uuid: UUID())
+    let attributes = optimizelyUserAttributes(with: AppEnvironment.current.currentUser)
+    let variationString = try? self.getVariationKey(
+      experimentKey: experiment.rawValue, userId: userId, attributes: attributes
+    )
+
+    guard
+      let variation = variationString,
+      let variant = OptimizelyExperiment.Variant(rawValue: variation)
+    else {
+      return .control
+    }
+
+    return variant
+  }
 }
 
 public func optimizelyTrackingAttributesAndEventTags(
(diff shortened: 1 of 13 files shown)


Library/OptimizelyClientType.swift
```diff
@@ -36,6 +36,31 @@ extension OptimizelyClientType {
...
+
+  /*
+   Calls `getVariation` on the Optimizely SDK for the given experiment,
+   using the default attributes and deviceId
+
+   Does *not* record an Optimizely impression. If you wish to record an experiment impression, use
+   `variant(for experiment)`, which calls `activate` on the Optimizely SDK
+   */
+
+  public func getVariation(for experiment: OptimizelyExperiment.Key) -> OptimizelyExperiment.Variant {
```
justinswart: The docs are unclear on whether `activate` is idempotent or not 🤔 
> Use Activate when the visitor actually sees the experiment. Use Get Variation when you need to know which bucket a visitor is in before showing the visitor the experiment.

This seems to suggest calling `getVariation` _before_ `activate` but I understand that in your scenario you are using it because the experiment is _already_ activated.

If it _is_ idempotent, do we care?

ifbarrera: I think this change is mainly a safety measure to minimize the chances of getting double-charged for impressions. In terms of our actual experiment, calling `activate` instead of `getVariation` would not affect the results, because like you said it's already been activated during the onboarding flow.

PR #1125, opened 2020-03-26
[NT-886] Category Personalization Experiment Tracking

ifbarrera:
# 📲 What

Adds a few Optimizely tracking events to the category personalization flow.

# 🤔 Why

So we can evaluate the effectiveness of the feature in helping new users find projects they'll be interested in.

# 🛠 How

- adds `Get Started Button Clicked` event
- adds `Continue Button Clicked` event
- adds `Skip Button Clicked` event

# ✅ Acceptance criteria

To force the category personalization flow to present on app launch, comment out the body of the `shouldSeeCategoryPersonalization()` function in `AppDelegateViewModel.swift` and instead `return true`.

- [x] Launch the app. You should see the onboarding flow appear. Tap "Get Started". You should see the Optimizely SDK log that an event was fired. Choose some categories. Then tap "Continue". You should see the Optimizely SDK log that an event called `Continue Button Clicked` was fired.
- [x] Launch the app. You should see the onboarding flow appear. Tap "Get Started". You should be navigated to the category selection screen. Tap "Skip" on the top right. You should see the Optimizely SDK log that an event called `Skip Button Clicked` was fired.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/CategorySelectionViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/CategorySelectionViewController.swift
@@ -184,6 +184,10 @@ public final class CategorySelectionViewController: UIViewController {
         self?.present(UIAlertController.genericError(message), animated: true)
       }
 
+    self.viewModel.outputs.dismiss
+      .observeForControllerAction()
+      .observeValues { [weak self] in self?.dismiss(animated: true) }
+
     self.warningLabel.rac.hidden = self.viewModel.outputs.warningLabelIsHidden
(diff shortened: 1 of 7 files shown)


Scollaco:
For some reason the `Continue Button Clicked` doesn't appear on the logs 🤔 The other 2 events are firing as expected, though.

ifbarrera:
> For some reason the `Continue Button Clicked` doesn't appear on the logs 🤔 The other 2 events are firing as expected, though.

Hey @Scollaco, that's because the event was incorrectly named on Optimizely! I've fixed it, it should be firing now 🙏

PR #1127, opened 2020-03-27
[NT-1023] Add Temporary Checkout Completed event

justinswart:
# 📲 What

- Adds `Temporary Checkout Completed` to be sent to Optimizely once a checkout completes.
- Refactors some of our Optimizely tracking functions to reduce repetition at callsites.

# 🤔 Why

- There are currently _reasons_ that the `Checkout Completed` event is not tracking correctly from the back-end. We are using this event as a temporary workaround for _reasons_.
- Decided to tidy up these functions to instead get user-related info from the environment and to also not _require_ some arguments. This tidies up most of the callsites.

# 🛠 How

- Added the event and tags.
- Refactored some code.

# 👀 See

Example JSON that is posted to Optimizely for this event:

```json
"snapshots": [
  {
    "events": [
      {
        "tags": {
          "category": "Publishing",
          "checkout_amount": 1,
          "project_user_has_watched": false,
          "revenue": 130,
          "project_subcategory": "Publishing",
          "project_country": "gb",
          "checkout_payment_type": "CREDIT_CARD",
          "currency": "GBP"
        },
        "key": "Temporary Completed Checkout",
        "timestamp": 1585254754335,
        "revenue": 130,
        "entity_id": "17851500012",
        "uuid": "2A85F177-3328-4EC6-BF16-8D62EDA5EB02"
      }
    ]
  }
],
"visitor_id": "5AC5FC52-9EFD-4521-A38F-3DA7A5127748"
```

# ✅ Acceptance criteria

Place yourself in an experimental variant on Optimizely, complete a checkout with a credit card and with Apple Pay. Observe the following for both:

- [ ] In the console, debug info displaying JSON containing similar data to the above (`checkout_payment_type` changes to `APPLE_PAY` when using that payment method).
- [ ] Verify this info on Optimizely's back-end (I was not yet able to confirm this).

### DIFF
--- a/Library/ViewModels/PledgeViewModelTests.swift
+++ b/Library/ViewModels/PledgeViewModelTests.swift
@@ -1149,7 +1149,12 @@ final class PledgeViewModelTests: TestCase {
       Result.success(CreateBackingEnvelope(createBacking: createBacking))
     )
 
-    withEnvironment(apiService: mockService, currentUser: .template) {
+    let optimizelyClient = MockOptimizelyClient()
+      |> \.experiments .~ [
+        OptimizelyExperiment.Key.pledgeCTACopy.rawValue: OptimizelyExperiment.Variant.variant1.rawValue
+      ]
+
+    withEnvironment(apiService: mockService, currentUser: .template, optimizelyClient: optimizelyClient) {
       let project = Project.template
       let reward = Reward.noReward
         |> Reward.lens.minimum .~ 5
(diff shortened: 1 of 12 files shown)


Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1202,6 +1207,40 @@ final class PledgeViewModelTests: TestCase {
         ["Checkout Payment Page Viewed"],
         self.trackingClient.events
       )
+
+      XCTAssertEqual(optimizelyClient.trackedUserId, "DEADBEEF-DEAD-BEEF-DEAD-DEADBEEFBEEF")
```
ifbarrera: I wonder if it might be nicer to create a separate test for the optimizely events? That way when we remove the experiment we can just delete the test 🤔

justinswart: Can do!

PR #1131, opened 2020-03-27
[NT-864] Continue with apple UI

Scollaco:
# 📲 What

- New SignUp screen with the `Continue with Apple` button for iOS 13+ users

# 🤔 Why
[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-864)

- This is part of the work that implements the new signup flow required by apple.

# 🛠 How

- Added a `ASAuthorizationAppleIDButton` to the view hierarchy
- Added availability checks to hide the `Continue with Apple` button if the app is running on iOS 12
- Added a new background image
- Generated new Snapshots

# 👀 See

| Before 🐛 | After 🦋iOS 12 | After 🦋iOS 13+ |
| --- | --- | --- |
|  |  |  |

**Note:** After talking to Jimmy on Slack, we decided to keep the space between the Facebook and Signup buttons.

- [ ] Works with VoiceOver

# ✅ Acceptance criteria

- [x] Running the app on iOS 13, the `Cotinue with Apple` button should be visible on the Login screen. It's hidden on iOS 12.


Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift
```diff
@@ -85,6 +88,11 @@ internal final class LoginToutViewController: UIViewController, MFMailComposeVie
       self.navigationItem.leftBarButtonItem = .close(self, selector: #selector(self.closeButtonPressed))
     }
 
+    _ = self.navigationController?.navigationBar
+      ?|> \.barTintColor .~ .clear
```
ifbarrera: I created a helper function `configureTransparentNavigationBar()` as an extension of `UINavigationBar` that you can use to set the navigation bar to be transparent. It's currently in the `feature-onboarding-personalization` branch, but maybe make a note here to update this once that's merged?

Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift
```diff
@@ -440,7 +486,20 @@ private let baseStackViewStyle: StackViewStyle = { stackView in
...
+
+private let logoImageViewStyle: ImageViewStyle = { imageView in
+  imageView
+    |> \.image .~ image(named: "kickstarter-logo")?.withRenderingMode(.alwaysTemplate)
+    |> \.tintColor .~ .ksr_green_500
+    |> \.contentMode .~ .scaleAspectFit
+    |> \.translatesAutoresizingMaskIntoConstraints .~ false
+}
+
+private let separatorViewStyle: ViewStyle = { view in
```
ifbarrera: We already have `separatorStyle` in `BaseStyles.swift`. We should try to avoid having many different styles for the same component, would the one we already have work?

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -58,6 +55,9 @@ public protocol LoginToutViewModelOutputs {
   /// Emits an access token envelope that can be used to update the environment.
   var logIntoEnvironment: Signal<AccessTokenEnvelope, Never> { get }
 
+  /// Emits an image to be used as the navigation bar background.
+  var navigationBarBackgroundImage: Signal<UIImage?, Never> { get }
```
ifbarrera: Why is this needed? I noticed when testing that when scrolling, the navigation bar cuts off the scroll content view. Is that the desired behavior or do we want the navigation bar to fully transparent?

Screen Shot 2020-03-30 at 3 03 49 PM

Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift
Scollaco: Oh good to know! I wonder if I could cherry-pick the commit

Library/ViewModels/LoginToutViewModel.swift
Scollaco: The `navigationBarBackgroundImage` to change color to `white` when going to signup and login screens (view with white background). I think that using your extension should fix this and this output shouldn't be necessary anymore.

Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift
Scollaco: Actually I copied and pasted the extension code and this is what I got:



Another solution would be to hide the navigation bar and add the buttons on the top corners. What do you think?

ifbarrera: @nino I think you need to remove the `baseControllerStyle` in order for it to work. `baseControllerStyle` is applying it's own navigation bar style that's overriding our setup.

nino: (Just FYI I don't think you meant to write my name. (I mean, I know _some_ Swift, but I really couldn't tell you if removing the `baseControllerStyle` would help))

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift | indent |


</p>

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/LoginToutViewController.swift | indent |


</p>

Kickstarter-iOS/Views/Controllers/LoginToutViewControllerTests.swift
```diff
@@ -33,6 +33,7 @@ internal final class LoginToutViewControllerTests: TestCase {
     let intent = LoginIntent.generic
     let controller = LoginToutViewController.configuredWith(loginIntent: intent)
 
-    XCTAssertNotNil(controller.view.subviews.first as? UIScrollView)
+    // Due to the new design, a background imageView is now added as subview before the scroll.
+    XCTAssertNotNil(controller.view.subviews[1] as? UIScrollView)
```
ifbarrera, line 37: Hmm what's this test for? Does this change mean that the scroll to top functionality won't work for this screen?

Scollaco: Correct. The problem here is that in order to match the designs, the view hierarchy had to change, since now a imageView was added to stack before the scrollView. As an option, I tried adding the background image view to the scrollView instead, but the scrolling looks weird with the background image moving.

Scollaco: Oh one sec. I think I found a solution.

PR #1135, opened 2020-03-31
[NT-1040] Project Summary Carousel UI

justinswart:
# 📲 What

Adds the UI and basic data source for the Project Summary Carousel experiment.

**Note:** Subsequent PRs will retrieve the items from GraphQL, add detail to the cells and also perform the max height calculations.

# 🤔 Why

We are planning to run an experiment to display horizontally scrollable project summary cells on the project page.

# 🛠 How

- Added a collection view to the `ProjectPamphletMainCell`.
- Restructured the view to allow this collection view to scroll edge-to-edge.
- Added a basic data source which will be updated when we are retrieving the summary items from GraphQL.

# 👀 See

# ✅ Acceptance criteria

Uncomment lines 111-112 in `ProjectPamphletMainCell` to populate the datasource and display empty cells.

- [ ] Navigate to any project, you should see the empty cells below the project blurb.
- [ ] You should not observe any other UI regression on the project page.

PR #1093, opened 2020-03-04
[NT-863] Creator Details Experiment

justinswart:
# 📲 What

Feature branch to track the work completed in #1086, #1088, #1089, #1095, #1097, #1098 which introduces an A/B experiment into the app to display trust indicators and creator details on the project page.

# 🤔 Why

As part of an experiment that we are running we would like to display some additional data to surface trust indicators to users.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image |  |

PR #1137, opened 2020-04-01
[NT-1039, NT-1042] ProjectSummary query and experiment activation

justinswart:
# 📲 What

- Adds the `native_me_project_summary` experiment and activation.
- Adds the GraphQL query and de-serialization for the fetching of project summary items.

# 🤔 Why

A continuation of the work in #1135, we are planning to run an experiment to display horizontally scrollable project summary cells on the project page.

# 🛠 How

- Added deserialization for `ProjectSummaryEnvelope` which contains an array of `ProjectSummaryItem`s.
- Added some safe-guards to ignore `ProjectSummaryItem`s that might have unsupported questions. These will just be discarded during de-serialization.
- Added activation of the `native_me_project_summary` experiment. This behaves similarly to #1093 in that we first attempt to load the project summary items and if that request fails we don't activate the experiment.

# ✅ Acceptance criteria

This is a little tricky to test because the UI is not completely wired up. I will do that in a subsequent PR to keep this one from growing.

Best bet is to create a project on staging that has project summary items. Navigate to that project and observe using `logEvents` statements that the following are true.

- [ ] If you are in the experimental variant _and_ project summary items are being returned, these are returned from `loadProjectPamphletContentDataIntoDataSource` in `ProjectPamphletContentViewModel`. The experiment is activated.
- [ ] If you are in the control group _and_ project summary items are being returned, `loadProjectPamphletContentDataIntoDataSource` emits and empty array for the project summary items. The experiment is activated.
- [ ] If you are in the experimental variant _and_ project summary items are not being returned, `loadProjectPamphletContentDataIntoDataSource` emits and empty array for the project summary items. The experiment is not activated.
- [ ] If you are in the experimental variant _and_ the project summary call errors, `loadProjectPamphletContentDataIntoDataSource` emits and empty array for the project summary items. The experiment is not activated.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/ProjectPamphletContentViewModel.swift
```diff
@@ -104,11 +109,48 @@ public final class ProjectPamphletContentViewModel: ProjectPamphletContentViewMo
...
+    let projectSummaryHasLoadedValues = projectSummaryRequestValues.filter { $0.isEmpty == false }
+
+    let projectSummaryExperimentValues = projectAndRefTag
+      .takePairWhen(projectSummaryHasLoadedValues)
+      .map { projectAndRefTag, projectSummaryItems in
+        (projectAndRefTag.0, projectAndRefTag.1, projectSummaryItems)
+      }
+      .map { project, refTag, projectSummaryItems -> [ProjectSummaryEnvelope.ProjectSummaryItem] in
+        guard
+          projectPageConversionProjectSummaryExperiment(project: project, refTag: refTag) != .control
```
ifbarrera, line 135: Hmm I'm wondering if we can move this optimizely experiment variant check higher up in the logic chain. It seems a bit wasteful to make the network request for the project summary items before we know whether we'll even display them. 🤔

Library/ViewModels/ProjectPamphletContentViewModelTests.swift
```diff
@@ -698,5 +702,225 @@ final class ProjectPamphletContentViewModelTests: TestCase {
...
+      self.loadProjectIntoDataSourceProjectSummaryItems.assertValues([[]])
+      self.loadProjectIntoDataSourceProject
+        .assertValues([project], "Nothing new emits when the view is done.")
+      self.loadProjectIntoDataSourceRefTag.assertValues(
+        [.discovery], "Nothing new emits when the view is done."
+      )
+      self.loadMinimalProjectIntoDataSource.assertValues([], "Nothing new emits when the view is done.")
+
+      // Simulate a new version of the project coming through
+      self.vm.inputs.configureWith(value: (project, .discovery))
```
ifbarrera, line 759: Does this ever happen IRL?

Library/ViewModels/ProjectPamphletContentViewModelTests.swift
```diff
@@ -698,5 +702,225 @@ final class ProjectPamphletContentViewModelTests: TestCase {
...
+      )
+      self.loadProjectIntoDataSourceRefTag.assertValues(
+        [.discovery, .discovery], "The new refTag is loaded into data source"
+      )
+      self.loadMinimalProjectIntoDataSource.assertValues([], "Nothing new emits when the view is done.")
+
+      self.scheduler.advance()
+
+      self.loadProjectIntoDataSourceProjectSummaryItems.assertValues(
+        [[], [], []], "[] is returned"
```
ifbarrera, line 848: I know we've cursed using `skipRepeats` in the past, but I wonder if it could be a performance improvement to use it here -- any values triggered by the `projectSummaryValues` signal will cause `loadProjectPamphletContentDataIntoDataSource` to fire (since we're using `combineLatest`). Every time that signal fires, the whole project table view will reload.

Library/ViewModels/ProjectPamphletContentViewModel.swift
justinswart: Yeah we did this with this experiment and the creator details one. I guess the only risk is that we could activate the experiment without knowing if the request for the data succeeded or not. So here we're flipping that by first ensuring that we have the data to display before activating 🤷‍♂

Library/ViewModels/ProjectPamphletContentViewModelTests.swift
justinswart: Yeah the project page is first configured with a minimal version and then with the updated project that includes more data.

Library/ViewModels/ProjectPamphletContentViewModelTests.swift
justinswart: Ok I've added a filter to prevent any empty arrays being emitted from `projectSummaryRequestValues` after the prefixed empty array to prevent unnecessary reloading.

Library/ViewModels/ProjectPamphletContentViewModel.swift
ifbarrera: You _could_ call `getVariation` which wouldn't activate the experiment. 🤔

PR #1140, opened 2020-04-02
[NT-686] add pledge collection date query

cdolm92:
# 📲 What

Adds pledge collection date to the query and displays it in the Activities Screen when the user has an errored pledge.

# 🤔 Why

So that users know how much time they have to correct their pledge before it is dropped.

# 🛠 How

Added `finalCollectionDate` to query 

# 👀 See
Simulator Screen Shot - iPhone 11 2 - 2020-04-03 at 13 22 53

# ✅ Acceptance criteria

- [ ] Navigate to Activities, if there are errored pledges you should see the final collection date.

# ⏰ TODO

- [x] Confirming copy


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Cells/ErroredBackingView.swift | anyObjectProtocol, blankLinesBetweenScopes |
| Library/ViewModels/ErroredBackingViewViewModel.swift | indent |


</p>

KsApi/models/GraphBacking.swift
```diff
@@ -21,3 +22,20 @@ public struct GraphBacking: Swift.Decodable, Equatable {
     case preauth
   }
 }
+
+extension GraphBacking.Project {
```
justinswart, line 26: I don't think you need this extension, `Decodable` should be able to infer the keys and the values.

We should however update `GraphBackingTests` to include the property that was added.

Library/ViewModels/ErroredBackingViewViewModel.swift
```diff
@@ -37,9 +45,65 @@ public final class ErroredBackingViewViewModel: ErroredBackingViewViewModelType,
...
 
+  public let finalCollectionDateText: Signal<String, Never>
   public let notifyDelegateManageButtonTapped: Signal<GraphBacking, Never>
   public let projectName: Signal<String, Never>
 
   public var inputs: ErroredBackingViewViewModelInputs { return self }
   public var outputs: ErroredBackingViewViewModelOutputs { return self }
 }
+
+private func timeLeftString(date: String) -> String {
```
justinswart, line 56: I think the return type here should be optional, a `nil` value better represents it not having returned anything than an empty string.

Library/ViewModels/ErroredBackingViewViewModel.swift
```diff
@@ -37,9 +45,65 @@ public final class ErroredBackingViewViewModel: ErroredBackingViewViewModelType,
...
+  let dateFormatter = ISO8601DateFormatter()
```
justinswart, line 57: We should cache date formatters as they're pretty expensive to create so may have an adverse effect on performance.

This can be done similarly to how we cache our formatters in [Format.swift](https://github.com/kickstarter/ios-oss/blob/master/Library/Format.swift#L408).

Library/ViewModels/ErroredBackingViewViewModelTests.swift
```diff
@@ -30,16 +32,23 @@ final class ErroredBackingViewViewModelTests: TestCase {
...
     let backing = GraphBacking.template
       |> \.project .~ project
 
     self.projectName.assertDidNotEmitValue()
+    self.finalCollectionDateText.assertDidNotEmitValue()
 
     self.vm.inputs.configure(with: backing)
 
     self.projectName.assertValue("Awesome tabletop collection")
+    self.finalCollectionDateText.assertValue("4 days left")
```
justinswart, line 52: Could we add tests for the hours and less than 1 hour cases?

PR #1143, opened 2020-04-03
[NT-1526] Codecov Integration

yatriks:
# 📲 What

Adds code coverage reports to our PRs and displays a badge in our `README.md`.

# 🤔 Why

This was some work started by @yatriks and @ifbarrera to improve our workflow and surface test coverage information. This information is good to keep track of and can also alert us to a possible inadvertant loss of test coverage.

# 🛠 How

@yatriks and @ifbarrera performed the initial setup to generate reports and upload them to codecov. I've since added an intermediary step which generates xml reports using a tool called [Slather](https://github.com/SlatherOrg/slather). This helps to discard coverage reports for unit tests themselves which would inflate our coverage percentages (see more [here](https://github.com/codecov/swift-standard#coverage-inflation)). Slather also aims to protect against codecov failing to parse Xcode's `.xccoverage` files.

# 👀 See

[codecov](https://codecov.io)
[Slather](https://github.com/SlatherOrg/slather)

Also, see below for the comment generated by codecov.


justinswart:
@yatriks let's set up some time to get this merged 👍

.circleci/config.yml
```diff
@@ -110,10 +113,25 @@ jobs:
...
+            BUNDLE_JOBS: 4
+            BUNDLE_RETRY: 3
+
+      - run:
+          name: Generate Slather coverage report
+          command: bundle exec slather coverage -x --scheme Kickstarter-Framework-iOS ./Kickstarter.xcodeproj
+
+      - run:
+          name: Upload report to codecov
+          command: bash <(curl -s https://codecov.io/bash) -f ./cobertura.xml
```
yatriks, line 134: So since we're using the CircleCI Orb for Codecov, we get a sugary syntax for effectively this command that looks like:
```
- codecov/upload:
    file: cobertura.xml
```
You can see how the orb just translates this into a `curl` command.
I wasn't sure why this was uploading without `CODECOV_TOKEN` set on the environment, but it's because this repo is public (which [codecov is fine with](https://docs.codecov.io/docs/about-the-codecov-bash-uploader#upload-token)) 😄

justinswart: oh nice, thanks i'll update that. I should still run the step just before to generate with slather though, right?

yatriks: Yep sounds right!

justinswart (approved):

PR #1146, opened 2020-04-06
[NT-1055] LandingPage deeplink handling bug fix

Scollaco:
# 📲 What

Fixes a bug where a LandingPage is presented and immediately dismissed if the app is launched through deeplink.

# 🤔 Why

It's a bug and we don't like bugs.

# 🛠 How

- This was caused by a race condition between the `goToLandingPage` and the `deeplinkActivated`. So in order to solve this, a `take(until:)` function was added to prevent the `deeplinkActivated` from emitting if `goToLandingPage` already emitted.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| deeplink_bug | deeplink_fix |

# ✅ Acceptance criteria

In order to test this, follow the steps:
1. Set Launch setting to `Wait for executable to be launched`
2. On line `1116`, return `true` immediately after the open bracket:
   ```
   private func shouldGoToLandingPage() -> Bool { return true
```
3. Add a 2 sec delay to the `openUrl` property:
```
  let openUrl = self.applicationOpenUrlProperty.signal
      .delay(2, on: AppEnvironment.current.scheduler)
      .skipNil()
```
4. Build the app on a device in the release configuration
5. Click the link to open the app through deeplink. I tested using [this link]( https://www.kickstarter.com/projects/chelsea-punk/chelsea-punk-band-the-final-album))

- [ ] The landing page should be presented on the first time.
- [ ] Close the app and tap the link again. The app should show the Project Page and the Landing Page should **not** be presented.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
@@ -2348,6 +2348,31 @@ final class AppDelegateViewModelTests: TestCase {
       self.goToLandingPage.assertDidNotEmitValue()
     }
   }
+
+  func testDeeplink_DoesNotActivateIf_GoToLandingPageEmits() {
+    let optimizelyClient = MockOptimizelyClient()
+      |> \.experiments .~
+      [OptimizelyExperiment.Key.nativeOnboarding.rawValue: OptimizelyExperiment.Variant.variant1.rawValue]
+
+    let userDefaults = MockKeyValueStore()
+      |> \.hasSeenLandingPage .~ false
+
+    withEnvironment(currentUser: nil, optimizelyClient: optimizelyClient, userDefaults: userDefaults) {
+      self.goToLandingPage.assertDidNotEmitValue()
+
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift [hunk 1]
```diff
@@ -2348,6 +2348,31 @@ final class AppDelegateViewModelTests: TestCase {
...
+    let userDefaults = MockKeyValueStore()
+      |> \.hasSeenLandingPage .~ false
+
+    withEnvironment(currentUser: nil, optimizelyClient: optimizelyClient, userDefaults: userDefaults) {
+      self.goToLandingPage.assertDidNotEmitValue()
+
+      self.vm.inputs.applicationDidFinishLaunching(application: UIApplication.shared, launchOptions: nil)
+      self.vm.inputs.didUpdateOptimizelyClient(MockOptimizelyClient())
+
+      _ = self.vm.inputs.applicationOpenUrl(
```
ifbarrera, line 2366: Can we actually run this test with a project deeplink? The output that was being triggered during the bug was `presentViewController`, which first dismisses any previously presented VC.

PR #1149, opened 2020-04-07
[NT-1100, NT-1113] Move confirmation label

justinswart:
# 📲 What

Moves the funding confirmation label below the total on `PledgeViewController`.

**Note:** The removed T&Cs label will be replaced in a subsequent PR when the new button drawer is added.

# 🤔 Why

This is part of some changes we are making to the pledge view, in this case the T&Cs text will replace this text below the submit button in a subsequent PR.

# 🛠 How

- Moved the confirmation label from `PledgeViewController` to `PledgeSummaryViewController`.
- Restructured the stackview in `PledgeSummaryViewController` view to accommodate it.
- Moved tests.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ♿️ Accessibility 

- [X] Supports Dynamic Type 

# ✅ Acceptance criteria

Navigate to the pledge view.

- [ ] Observe that the confirmation summary label is now positioned below the total.
- [ ] Behaviour of this label should be unchanged - that is, if the project currency differs from that of the user, the label should include the total.


Kickstarter-iOS/Views/Controllers/PledgeSummaryViewController.swift
```diff
@@ -145,18 +151,6 @@ private let rootStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
     |> \.spacing .~ Styles.gridHalf(3)
 }
 
-private let termsTextViewStyle: TextViewStyle = { (textView: UITextView) -> UITextView in
```
justinswart, line 148: @cdolm92 note the removed code in this file - you will need it when you add this label back when adding the button drawer.

Library/Tracking/Koala.swift
```diff
@@ -1374,9 +1374,11 @@ public final class Koala {
    - parameter params: The optional Discover params if the project is being watched from Discover
    */
 
-  public func trackWatchProjectButtonClicked(project: Project,
-                                             location: LocationContext,
-                                             params: DiscoveryParams? = nil) {
+  public func trackWatchProjectButtonClicked(
```
justinswart, line 1377: Looks like just some formatting happened here.

Library/ViewModels/PledgeSummaryViewModelTests.swift
```diff
@@ -58,9 +66,97 @@ internal final class PledgeSummaryViewModelTests: TestCase {
...
       |> Project.lens.stats.currentCurrencyRate .~ nil
 
-    self.vm.inputs.configureWith(project, total: 10)
+    self.vm.inputs.configure(with: (project, total: 10, false))
     self.vm.inputs.viewDidLoad()
 
     self.totalConversionLabelText.assertDidNotEmitValue()
   }
+
+  func testUpdateContext_ConfirmationLabel() {
```
justinswart, line 75: These tests were just carried over from `PledgeViewModelTests`.

justinswart:
The snapshots here will be re-recorded several times as we work on this feature, we should chat about whether that makes sense 🤔

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/PledgeSummaryViewController.swift
```diff
@@ -69,10 +71,10 @@ final class PledgeSummaryViewController: UIViewController {
       |> ksr_addSubviewToParent()
       |> ksr_constrainViewToEdgesInParent()
 
-    _ = ([self.titleLabel, self.adaptableStackView], self.rootStackView)
+    _ = ([self.adaptableStackView, self.confirmationLabel], self.rootStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    _ = ([self.termsTextView, self.totalStackView], self.adaptableStackView)
+    _ = ([self.titleLabel, self.totalStackView], self.adaptableStackView)
```
cdolm92, line 77: Should we rename this title label to total? It was kind of confusing knowing what it was.

colleenmacd:
@cdolm92 thanks for tagging me! i realize this was another edge case I was missing for iOS. I think we should add ~18px of padding between the conversion string and the next piece of content.

Design file can be found here: https://www.figma.com/file/DMwhdWji5zRLFAoIBDzVji/Checkout?node-id=542%3A140

Kickstarter-iOS/Views/Controllers/PledgeSummaryViewController.swift
justinswart: I think in this case it refers to the title of the section. I think it's better to name a UI element according to what it is rather than what it says because the latter can change.

PR #1150, opened 2020-04-08
[NT-969] Show full-screen "processing" state during create/update backing requests

ifbarrera:
# 📲 What

Adds a full-screen "processing" view that displays when a create or update backing request is in flight, to prevent users from interacting with the app during critical requests.

# 🤔 Why

We want to guarantee that the user can't submit another backing request while one is in flight.

# 🛠 How

- added `ProcessingView` which is added to the app's `mainWindow` when either the `createBacking` or `updateBacking` requests are in flight
- the `ProcessingView` is removed from the hierarchy when the request terminates

# 👀 See

| Pledging with Card | Apple Pay | iPad | Update Backing
| --- | --- | --- | --- |
| 5mZW7yzcRD | XLqBEGNOtB | oDsk4GR0GW-2 | BzMYkIAf9j |

# ✅ Acceptance criteria

- [ ] Run the app on the simulator, switch to staging. Back any project using a stored card. You should see the "Processing..." state momentarily before seeing the Thank You screen.
- [ ] Run the app on the simulator, switch to staging. Back any project using Apple Pay. You should see the "Processing..." state momentarily before seeing the Thank You screen.
- [ ] Navigate to a backed project, and tap "Manage". Tap "Change your payment method" and update your payment method. You should see the "Processing..." screen momentarily before seeing the success banner.

# ⏰ TODO

- [ ] update "Processing..." string


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -412,6 +426,24 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
...
+  private func showProcessingView() {
+    guard let window = UIApplication.shared.keyWindow else {
+      return
+    }
+
+    _ = (self.processingView, window)
+      |> ksr_addSubviewToParent()
+      |> ksr_constrainViewToEdgesInParent()
+
+    self.processingView.activityIndicator.startAnimating()
```
Scollaco, line 439: I think we could start animating the indicator in the ProcessingView's init method. This way we don't need to expose its UIActivityIndicatorView property and it's one less thing to worry about, since we don't need to start/stop the animation.

ifbarrera: Good idea!

Kickstarter-iOS/Views/ProcessingView.swift
```diff
@@ -0,0 +1,90 @@
...
+private let processingLabelStyle: LabelStyle = { label in
+  label
+    |> \.font .~ UIFont.ksr_callout()
+    |> \.textColor .~ UIColor.white
+    |> \.textAlignment .~ .center
+    |> \.lineBreakMode .~ .byWordWrapping
+    |> \.numberOfLines .~ 0
+    |> \.text %~ { _ -> String in
+      var processingString = Strings.project_checkout_finalizing_title()
+      processingString.append("...")
```
Scollaco, line 78: Considering the localizations, I wonder if it's ok to append the `...` to the string.

PR #1152, opened 2020-04-09
[NT-867] (1/2) Sign in with apple mutation

Scollaco:
# 📲 What
- Adds the `signInWithApple` mutation to the codebase.
** Note: ** This PR only contains the mutation. It doesn't fetch the user from v1, thus doesn't login into our Environment. This will be done in the second part of this work.

# 🤔 Why

- This work is part of the Sign-In With Apple epic. The mutation will allow users to login using the `Sign in with Apple` functionality.

[JIRA ticket](https://kickstarter.atlassian.net/browse/NT-867)

# 🛠 How

- Added the Sign in with Apple capability
- Created the `SignInWithAppleInput` containing `appId`, `firstName`, `lastName`, and `authToken` (provided by Apple)
- Created the `SignInWithAppleEnvelope` to be decoded once graphQL returns a value
- A new `didSignInWithApple` output was created.
- Added tests

# ✅ Acceptance criteria
To test this, change the environment to `Development`

- [ ] Being logged out, go to login/signup screen and tap the `Continue with Apple` button. Complete the CWA flow. The `didSignInWithApple` output should emit a signal with a `apiAccessToken` and a `user` (from graphQL).

### DIFF
--- a/KsApi/models/SignInWithAppleEnvelope.swift
+++ b/KsApi/models/SignInWithAppleEnvelope.swift
@@ -1,0 +1,5 @@
+import Foundation
+
+public struct SignInWithAppleEnvelope: Decodable {
+  public var apiAccessToken: String
+}
--- a/KsApi/models/SignInWithAppleEnvelope.swift
+++ b/KsApi/models/SignInWithAppleEnvelope.swift
@@ -9,9 +9,6 @@ public struct SignInWithAppleEnvelope: Decodable {
   }
 
   public struct User: Decodable {
-    public let id: String
-    public var intID: Int? {
-      return decompose(id: self.id)
-    }
+    public let uid: String
   }
 }
--- a/KsApi/models/SignInWithAppleEnvelope.swift
+++ b/KsApi/models/SignInWithAppleEnvelope.swift
@@ -1,5 +1,18 @@
 import Foundation
 
 public struct SignInWithAppleEnvelope: Decodable {
-  public var apiAccessToken: String
+
+  public let signInWithApple: SignInWithApple
+
+  public struct SignInWithApple: Decodable {
+    public let apiAccessToken: String
+    public let user: SignInWithAppleEnvelope.User
+  }
+
+  public struct User: Decodable {
+    public let id: String
+    public var intID: Int? {
+      return decompose(id: self.id)
+    }
+  }
 }
--- a/KsApi/models/SignInWithAppleEnvelope.swift
+++ b/KsApi/models/SignInWithAppleEnvelope.swift
@@ -1,7 +1,6 @@
 import Foundation
 
 public struct SignInWithAppleEnvelope: Decodable {
-
   public let signInWithApple: SignInWithApple
 
   public struct SignInWithApple: Decodable {
--- a/KsApi/mutations/inputs/SignInWithAppleInput.swift
+++ b/KsApi/mutations/inputs/SignInWithAppleInput.swift
@@ -1,11 +1,13 @@
 import Foundation
 
 public struct SignInWithAppleInput: GraphMutationInput {
+  public let appId: String
   public let authCode: String
-  public let firstName: String
-  public let lastName: String
+  public let firstName: String?
+  public let lastName: String?
 
-  public init(authCode: String, firstName: String, lastName: String) {
+  public init(appId: String, authCode: String, firstName: String?, lastName: String?) {
+    self.appId = appId
     self.authCode = authCode
     self.firstName = firstName
     self.lastName = lastName
@@ -13,9 +15,10 @@ public struct SignInWithAppleInput: GraphMutationInput {
 
   public func toInputDictionary() -> [String: Any] {
     return [
+      "iosAppId": appId,
       "authCode": authCode,
-      "firstName": firstName,
-      "lastName": lastName
+      "firstName": firstName ?? "",
+      "lastName": lastName ?? ""
     ]
   }
 }
--- a/KsApi/mutations/inputs/SignInWithAppleInput.swift
+++ b/KsApi/mutations/inputs/SignInWithAppleInput.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+public struct SignInWithAppleInput: GraphMutationInput {
+  public let authCode: String
+  public let firstName: String
+  public let lastName: String
+
+  public init(authCode: String, firstName: String, lastName: String) {
+    self.authCode = authCode
+    self.firstName = firstName
+    self.lastName = lastName
+  }
+
+  public func toInputDictionary() -> [String: Any] {
+    return [
+      "authCode": authCode,
+      "firstName": firstName,
+      "lastName": lastName
+    ]
+  }
+}
--- a/KsApi/mutations/inputs/SignInWithAppleInput.swift
+++ b/KsApi/mutations/inputs/SignInWithAppleInput.swift
@@ -17,8 +17,8 @@ public struct SignInWithAppleInput: GraphMutationInput {
     return [
       "iosAppId": appId,
       "authCode": authCode,
-      "firstName": firstName ?? "",
-      "lastName": lastName ?? ""
-    ]
+      "firstName": firstName,
+      "lastName": lastName
+    ].compact()
   }
 }
--- a/Library/ViewModels/LoginToutViewModel.swift
+++ b/Library/ViewModels/LoginToutViewModel.swift
@@ -15,7 +15,7 @@ public protocol LoginToutViewModelInputs {
 
   /// Call when Apple completes authorization
   @available(iOS 13.0, *)
-  func appleAuthorizationDidComplete(with authorization: ASAuthorization)
+  func appleAuthorizationDidComplete(with data: SignInWithAppleData?)
 
   /// Call when Apple completes authorization with error
   @available(iOS 13.0, *)
@@ -198,8 +198,7 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
     // MARK: - Sign-in with Apple
 
     if #available(iOS 13.0, *) {
-      let appleSignInInput = self.appleAuthorizationDidCompleteWithAuthorizatonProperty.signal
-        .map(continueWithAppleData(from:))
+      let appleSignInInput = self.appleAuthorizationDidCompleteWithDataProperty.signal
         .skipNil()
         .map { data in
           SignInWithAppleInput(appId: data.appId,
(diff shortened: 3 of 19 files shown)


Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -274,13 +343,20 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
     self.viewIsPresentedProperty.value = isPresented
   }
 
+  fileprivate var viewWillAppearProperty = MutableProperty(())
+  public func viewWillAppear() {
+    self.viewWillAppearProperty.value = ()
+  }
+
   public let attemptFacebookLogin: Signal<(), Never>
+  public private(set) var didSignInWithApple: Signal<SignInWithAppleEnvelope, Never> = .empty
```
Scollaco, line 352: Had to make this property a `var` with an empty initialization because it's only used inside the `if #available` and it's unavailable on iOS 12, and Xcode doesn't allow stored properties to be "potentially  unavailable".

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/models/SignInWithAppleEnvelope.swift
```diff
@@ -0,0 +1,17 @@
...
+public struct SignInWithAppleEnvelope: Decodable {
+  public let signInWithApple: SignInWithApple
+
+  public struct SignInWithApple: Decodable {
+    public let apiAccessToken: String
+    public let user: SignInWithAppleEnvelope.User
+  }
+
+  public struct User: Decodable {
+    public let id: String
```
ifbarrera, line 12: I think the `User` schema exposes `uid`, which I think is what you need.

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -1,12 +1,29 @@
...
   /// Call to set the reason the user is attempting to log in
   func configureWith(_ intent: LoginIntent, project: Project?, reward: Reward?)
 
+  /// Call when Apple completes authorization
+  @available(iOS 13.0, *)
+  func appleAuthorizationDidComplete(with data: SignInWithAppleData?)
+
+  /// Call when Apple completes authorization with error
+  @available(iOS 13.0, *)
+  func appleAuthorizationDidComplete(with error: Error)
```
ifbarrera, line 22: Maybe a better name for this would be `appleAuthorizationDidFail(with error: Error)`? That way you can differentiate between a success and a failure. Above, you could also renamed `appleAuthorizationDidComplete` to `appleAuthorizationDidSucceed`.

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -1,12 +1,29 @@
...
+
+  /// Call when the SignInWithAppleEnvelope is received from the server
+  func didReceiveSignInWithAppleEnvelope(_ envelope: SignInWithAppleEnvelope)
```
ifbarrera, line 25: I think usually we try to name our outputs in a way that describes the action that will take place by calling that output, for example `postNotification`, `dismissKeyboard`, etc. From what I understand, what will happen when the Apple envelope is received is we will attempt to fetch the v1 `User` object to log the user into our `AppEnvironment`. I wonder if then, this output should be `loginToEnvironment`?

ifbarrera: Whoops this is not an output. I meant to write it next to the `didSignInWithApple` output.

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -220,9 +269,24 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
...
-  fileprivate var viewWillAppearProperty = MutableProperty(())
-  public func viewWillAppear() {
-    self.viewWillAppearProperty.value = ()
+  fileprivate let appleLoginButtonPressedProperty = MutableProperty(())
+  public func appleLoginButtonPressed() {
+    self.appleLoginButtonPressedProperty.value = ()
+  }
+
+  // This property is being set as lazy because we can't use @available in computed properties.
+  @available(iOS 13.0, *)
```
ifbarrera, line 278: Do you think it's necessary to wrap these outputs with the `@available` check? I would imagine that as long as you can make sure the inputs are not being triggered on the VC, that it would be safe not to do `@available` checks in the VM, since you're not actually using any unavailable APIs here 🤔. 

Thoughts @justinswart ?

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -1,12 +1,29 @@
+import AuthenticationServices
```
ifbarrera, line 1: Are you actually using this API in this file? I don't see it anywhere but might be missing something.

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -1,3 +1,4 @@
+import AuthenticationServices
```
ifbarrera, line 1: Same here, is this needed?

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -585,4 +590,26 @@ final class LoginToutViewModelTests: TestCase {
 
     self.dismissViewController.assertValueCount(1)
   }
+
+  @available(iOS 13, *)
+  func testDidSignInWithApple() {
```
ifbarrera, line 595: Can you add a test for the `prepareSignInWithAppleRequest` output?

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -58,6 +79,9 @@ public protocol LoginToutViewModelOutputs {
   /// Emits when a login success notification should be posted.
   var postNotification: Signal<(Notification, Notification), Never> { get }
 
+  /// Emits when Continue with Apple button is tapped.
+  var prepareSignInWithAppleRequest: Signal<Void, Never> { get }
```
ifbarrera, line 83: I noticed that for Facebook login, the corresponding output is called `attemptFacebookLogin`. I wonder if we should try to keep the naming consistent for various 3rd party logins? So maybe `attemptAppleLogin` would be better here?

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -585,4 +590,26 @@ final class LoginToutViewModelTests: TestCase {
...
+    let response = SignInWithAppleEnvelope.template
+
+    withEnvironment(apiService: MockService(signInWithAppleResponse: response)) {
```
ifbarrera, line 598: Can you also add a test for the case where the `signInWithApple` request fails?

Library/ViewModels/LoginToutViewModel.swift
Scollaco: Agreed!

Library/ViewModels/LoginToutViewModel.swift
Scollaco: You're right. But I think that I added the availability because previously I was testing the whole flow (fetching v1 user, etc) and the compiler was complaining, I think due the involvement of other outputs and signals. But, because this is not happening on this PR, I will remove it.

Library/ViewModels/LoginToutViewModel.swift
Scollaco: We will, but only on the next PR. Will remove it for now (forgot to clean this up)

Library/ViewModels/LoginToutViewModelTests.swift
Scollaco: Same here

Library/ViewModels/LoginToutViewModelTests.swift
Scollaco: Goot call

Library/ViewModels/LoginToutViewModelTests.swift
Scollaco: The test for when the `signInWithApple` fails was added on the second part of this PR, where we also handle errors.

Library/ViewModels/LoginToutViewModel.swift
Scollaco: There's a difference here though. The `didSignInWithApple` emits when AuthorizationServices grants us an authorization, but we will only login to environment once we fetch the user from v1. We also already have an output `logIntoEnvironment` that's used after the facebook login flow is completed. We will use this same output to login with Apple.

ifbarrera: Yeah, I guess I'm wondering why we need an additional output then? Can't we just fetch the user immediately in the VM when the `signInWithApple` call succeeds, and then call `logIntoEnvironment` when the `fetchUser` call succeeds?

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -116,7 +139,7 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
           .materialize()
       }
 
-    self.logIntoEnvironment = facebookLogin.values()
+    self.attemptAppleLogin = self.appleLoginButtonPressedProperty.signal.ignoreValues()
```
oliverswitzer, line 142: is it ok to get rid of `self.logIntoEnvironment = facebookLogin.values()` here? Assuming maybe we moved this somewhere else?

Scollaco: We just moved it down within the `init` function, but still need it to login using Facebook. On the second PR (to be opened today), it will be used also to handle Apple login. i.e: 
```
self.logIntoEnvironment = Signal.merge(facebookLogin, appleLogin)
```

Library/ViewModels/LoginToutViewModel.swift
Scollaco: The main motivation was to test the values of the `appleSignInEvent`. Once we implement the  logic to fetch user, maybe this won't be necessary anymore, because we can simply test if `logIntoEnvironment` emits after signing in with Apple

Scollaco: Ended up deleting it 👍

KsApi/mutations/inputs/SignInWithAppleInput.swift
```diff
@@ -0,0 +1,24 @@
...
+    self.authCode = authCode
+    self.firstName = firstName
+    self.lastName = lastName
+  }
+
+  public func toInputDictionary() -> [String: Any] {
+    return [
+      "iosAppId": appId,
+      "authCode": authCode,
+      "firstName": firstName ?? "",
```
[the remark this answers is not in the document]
Scollaco: Oh nice! Will change it to use `compact`!

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| KsApi/MockService.swift | indent |


</p>

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -11,7 +11,9 @@ import XCTest
 final class LoginToutViewModelTests: TestCase {
   fileprivate let vm: LoginToutViewModelType = LoginToutViewModel()
 
+  fileprivate let attemptAppleLogin = TestObserver<(), Never>()
   fileprivate let attemptFacebookLogin = TestObserver<(), Never>()
+  fileprivate let didSignInWithApple = TestObserver<SignInWithAppleEnvelope, Never>()
```
ifbarrera, line 16: This can be removed right?

Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -168,6 +182,34 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
       genericFacebookErrorAlert
     )
 
+    self.logIntoEnvironment = facebookLogin.values()
+
+    // MARK: - Sign-in with Apple
+
+    if #available(iOS 13.0, *) {
```
ifbarrera, line 189: This can be removed too 😅

PR #1153, opened 2020-04-09
[NT-1016] Update app strings

ifbarrera:
# 📲 What

A big update of a bunch of translation strings. Also adds screenshot tests for `LandingPageViewController`, which were missing.

# 🤔 Why

Keeping our translations up to date ✨ 

# 🛠 How

- ran `make strings`
- updated and verified any breaking screenshots
- replaced any instances of `localizedString()` with the correct `Strings` functions
- added `LandingPageViewControllerTests` for screenshots of the Landing Page screen

# ✅ Acceptance criteria

- [x] skim all the screenshots 😬


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ifbarrera:
> Seems fine, the background looks weird on this device but I'm guessing it's just aspect ratio. Wonder if we can't just let it fill and see how it looks? Might be better for the 4.7 inch device. Also, is the button meant to be edge-to-edge like that?
> 
> image

Good eye, the `contentMode` was previously set to `.scaleToFill`. I've updated it to `.scaleAspectFill` and re-recorded the snapshots. It's not meant to be edge-to-edge, but I think because the scroll view is pinned to the margins, and there aren't any margins on the parent view controller, it's showing up as edge-to-edge. I've updated the parent view's margins and re-recorded.

PR #1161, opened 2020-04-14
[NT- 1142] Routing disclaimer/terms copy

cdolm92:
# 📲 What

Tapping `Terms of Use`, `Privacy Policy` or `Cookie Policy` takes user to the appropriate view.

# 🤔 Why

So that users can read our policies.

# 🛠 How

The `PledgeViewController` conforms to a `PledgeViewCTAContainerView` delegate to present the appropriate help screen. 

# ✅ Acceptance criteria

- [ ] Tapping on Terms of Use`, `Privacy Policy` and `Cookie Policy` takes the user to those screens.


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -385,6 +385,13 @@ final class PledgeViewController: UIViewController, MessageBannerViewControllerP
         }
       }
 
+    self.viewModel.outputs.showWebHelp
+      .observeForUI()
```
justinswart, line 389: Being the closest output to presenting the controller this one should be `observeForControllerAction()`

Kickstarter-iOS/Views/PledgeViewCTAContainerView.swift
```diff
@@ -109,6 +110,13 @@ final class PledgeViewCTAContainerView: UIView {
         guard let self = self else { return }
         self.delegate?.applePayButtonTapped()
       }
+
+    self.viewModel.outputs.notifyDelegateOpenHelpType
+      .observeForControllerAction()
```
justinswart, line 115: This can be `observeForUI()` but doesn't matter too much, just that this doesn't directly present the controller whereas the other one would specifically place the presentation on that queue. The fact that we even have these two different `observeFor` functions was to workaround a bug that might no longer exist so we should dig into that a bit more.

PR #1164, opened 2020-04-17
[NT-1176] Update SwiftFormat

justinswart:
# 📲 What

- Updates SwiftFormat to `0.44.7`.
- Removes `swiftlint`.

# 🤔 Why

For some time we've used these two formatting tools alongside each other but we only kept `swiftlint` around in order to warn when we exceed the maximum line length of `110`. SwiftFormat now supports wrapping long lines so we are able to use one tool for all of our formatting.

# 🛠 How

- Removed `swiftlint`.
- Updated SwiftFormat.
- Ran `bin/format.sh .` on the codebase and disabled the `wrap` rule where it seemed necessary.
- Removed all `// swiftlint` comments which are no longer needed.

# ✅ Acceptance criteria

- [ ] Examine the code changes that the formatter made, flag any that seem egregious.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS.playground/Pages/VideoPlayer.xcplaygroundpage/Contents.swift
```diff
@@ -9,7 +9,8 @@ import UIKit
 let project = .cosmicSurgery
   |> Project.lens.video .~ (
     .template
-      |> Project.Video.lens.high .~ "https://d2pq0u4uni88oo.cloudfront.net/projects/1846844/video-562464-h264_high.mp4"
+      |> Project.Video.lens
+      .high .~ "https://d2pq0u4uni88oo.cloudfront.net/projects/1846844/video-562464-h264_high.mp4"
```
ifbarrera, line 13: This line break is a little strange, are we ok with swiftformat making changes like this? 🤔

justinswart: I didn't really check the playgrounds too closely - I can tweak this.

justinswart:
> One thing I noticed is we need to remove the Swiftlint build phase from the app target. I wonder if instead we want to replace it bin/format.sh, or is that too aggressive?

Good point! Will remove the build phase.

I think we could consider SwiftFormat as a pre-commit hook possibly but the unfortunate side-effect of running it during development is that you use your undo history which isn't ideal. So running it as a build phase would be problematic.

PR #1168, opened 2020-04-21
[NT-1136] Add "Processing" loader to Login/Signup Screen

ifbarrera:
# 📲 What

Adds the "Processing" full-screen loader view to the Login/Signup screen, and presents the loader view while the sign-in-with-apple and user requests are in flight. We will also show the loader when the login with Facebook request is in flight.

# 🤔 Why

So that we can give some indication to the user that something is happening after they authenticate with Apple and Facebook.

# 🛠 How

- made a protocol and protocol extension `ProcessingViewPresenting` where the showing/hiding functions will live
- conformed both `PledgeViewController` and `LoginToutViewController` to `ProcessingViewPresenting`
- the `isLoading` output already existed in `LoginToutViewModel`, but wasn't being used. This PR couples the `isLoading` output to the `signInWithApple` mutation, as well as the `fetchUser` request
- added VM tests
- added Voice Over support
- added accessibility labels to the Kickstarter logos which were missing

# 👀 See

GyjckC8Cm1

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Run the app on your device, with iOS 13 and above. Navigate to the login/signup screen, and tap "Continue with Apple". Authenticate with Apple, and then you should see the "Processing" view momentarily until the request completes.

### DIFF
--- a/Library/ViewModels/LoginToutViewModel.swift
+++ b/Library/ViewModels/LoginToutViewModel.swift
@@ -127,6 +127,7 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
     let facebookLogin = tokenString
       .switchMap { token in
         AppEnvironment.current.apiService.login(facebookAccessToken: token, code: nil)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
           .on(
             starting: {
               isLoading.value = true
@@ -135,7 +136,6 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
               isLoading.value = false
             }
           )
-          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
           .materialize()
       }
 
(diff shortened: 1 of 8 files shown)


Library/ViewModels/LoginToutViewModel.swift
```diff
@@ -208,6 +207,15 @@ public final class LoginToutViewModel: LoginToutViewModelType, LoginToutViewMode
     let appleSignInEvent = appleSignInInput
       .switchMap { input in
         AppEnvironment.current.apiService.signInWithApple(input: input)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .on(
+            starting: {
+              isLoading.value = true
+            },
+            failed: { _ in
```
ifbarrera, line 215: Notice that here we set the `isLoading` value to `false` only if the request fails, because we don't want to hide the processing view until the next request (`fetchUser`) terminates.

PR #1169, opened 2020-04-21
[NT-1179] Optimizely configuration error logging

ifbarrera:
# 📲 What

We want to keep on top of how often Optimizely fails to configure, and why. So, we plan to report Optimizely configuration [errors to Crashlytics](https://docs.fabric.io/apple/crashlytics/logged-errors.html), using their `recordError(error:)` function.

# 🤔 Why

We believe that sometimes, the `OptimizelyClient` may fail to configure, leading to unexpected behavior down the road. We want to quantify how often this is happening in the App Store builds, which we can accomplish by logging errors to Fabric/Crashlytics/Firebase.

# 🛠 How

- first, we needed to update the `optimizelyConfigured` output to send back the `Error` returned from the Optimizely SDK
- then, we send this error to `Crashlytics.shared.recordError(error:)`
- updated tests

# ✅ Acceptance criteria

We don't configure Fabric/Crashlytics on debug builds, so we'll have to test reported errors on a `Release` build.

- [x] verify tests are passing


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -357,14 +357,22 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
         print("🔮 Optimizely SDK Successfully Configured")
         AppEnvironment.updateOptimizelyClient(optimizelyClient)
 
-        self?.viewModel.inputs.didUpdateOptimizelyClient(optimizelyClient)
+        self.viewModel.inputs.didUpdateOptimizelyClient(optimizelyClient)
+
+        return
       }
+
+      print("🔴 Optimizely SDK Configuration Failed with Error: \(optimizelyError.localizedDescription)")
```
eoji, line 373: I love it, I don't even log the reason 😆

PR #1172, opened 2020-04-23
[NT-1162] Add OptimizelyProperties

ifbarrera:
# 📲 What

Adds a set of "Optimizely Properties" that will be sent with certain events to the data lake.

# 🤔 Why

We want to be able to analyze experiment results in Amplitude. As part of this work, we will send some Optimizely data along with certain events to the data lake, to be passed forward to Amplitude from there. This PR adds the `optimizelyProperties()` function that will generate these properties to be sent with desired events.

# 🛠 How

- adds `optimizelyProperties` function, which calls `getVariation` on all known experiments. If `getVariation` throws an error, we report the variant as "unknown" for that experiment
- adds tests for `optimizelyProperties`
- adds tests for our helper `getVariation` function that didn't have tests previously

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

The `optimizelyProperties` function isn't being used anywhere yet, so all that's needed is code review and test review.

### DIFF
--- a/Library/OptimizelyClientType.swift
+++ b/Library/OptimizelyClientType.swift
@@ -6,7 +6,7 @@ public protocol OptimizelyClientType: AnyObject {
   func activate(experimentKey: String, userId: String, attributes: [String: Any?]?) throws -> String
   func getVariationKey(experimentKey: String, userId: String, attributes: [String: Any?]?) throws -> String
   func track(eventKey: String, userId: String, attributes: [String: Any?]?, eventTags: [String: Any]?) throws
-  func allExperiments() -> [OptimizelyExperiment.Key]
+  func allExperiments() -> [String]
 }
 
 extension OptimizelyClientType {
@@ -47,11 +47,11 @@ extension OptimizelyClientType {
    `variant(for experiment)`, which calls `activate` on the Optimizely SDK
    */
 
-  public func getVariation(for experiment: OptimizelyExperiment.Key) -> OptimizelyExperiment.Variant {
+  public func getVariation(for experimentKey: String) -> OptimizelyExperiment.Variant {
     let userId = deviceIdentifier(uuid: UUID())
     let attributes = optimizelyUserAttributes()
     let variationString = try? self.getVariationKey(
-      experimentKey: experiment.rawValue, userId: userId, attributes: attributes
+      experimentKey: experimentKey, userId: userId, attributes: attributes
     )
 
     guard
@@ -66,8 +66,8 @@ extension OptimizelyClientType {
 
   /* Returns all experiments the app knows about */
 
-  public func allExperiments() -> [OptimizelyExperiment.Key] {
-    return OptimizelyExperiment.Key.allCases
+  public func allExperiments() -> [String] {
+    return OptimizelyExperiment.Key.allCases.map { $0.rawValue }
   }
 }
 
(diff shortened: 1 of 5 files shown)


Library/OptimizelyClientType.swift
```diff
@@ -62,10 +63,63 @@ extension OptimizelyClientType {
 
     return variant
   }
+
+  /* Returns all experiments the app knows about */
+
+  public func allExperiments() -> [OptimizelyExperiment.Key] {
```
eoji, line 69: thinking 🆒

Library/OptimizelyClientTypeTests.swift
```diff
@@ -76,4 +76,100 @@ final class OptimizelyClientTypeTests: TestCase {
...
     }
   }
+
+  func testGetVariation() {
+    let mockOptimizelyClient = MockOptimizelyClient()
+      |> \.experiments .~
+      [OptimizelyExperiment.Key.pledgeCTACopy.rawValue: OptimizelyExperiment.Variant.variant1.rawValue]
+
+    withEnvironment(currentUser: User.template, optimizelyClient: mockOptimizelyClient) {
+      let variation = mockOptimizelyClient.getVariation(for: .pledgeCTACopy)
```
eoji, line 86: I used a fake experiment here because I didn't want to have to update the tests when we remove the experiment

ifbarrera: Good idea! Will update.

Library/OptimizelyClientType.swift
```diff
@@ -62,10 +63,63 @@ extension OptimizelyClientType {
...
 // MARK: - Tracking Properties
 
+public func optimizelyProperties(environment: Environment? = AppEnvironment.current) -> [String: Any]? {
+  guard let env = environment, let optimizelyClient = env.optimizelyClient else {
+    return nil
+  }
+
+  let environmentType = env.environmentType
+  let userId = deviceIdentifier(uuid: UUID())
+  let attributes = optimizelyUserAttributes()
```
eoji, line 83: I wrote a whole question but Github ate it?

I was asking if we always call this with no project and ref tags, those properties will always be `nil`. Since  `optimizelyUserAttributes ` calls `sessionRefTagProperties` on line 159. Do those get set somewhere else?

ifbarrera: I thought about this and decided that it doesn't make any sense that we have `ref_tag` and `referrer_credit` as user properties 😅 I don't see a world where we would use those properties to modify the audience of an experiment, and for that reason I decided it was simpler to always leave them out. Will check with product whether this is ok though!

Library/OptimizelyClientType.swift
```diff
@@ -62,10 +63,63 @@ extension OptimizelyClientType {
...
+  case .production:
+    sdkKey = Secrets.OptimizelySDKKey.production
+  case .staging:
+    sdkKey = Secrets.OptimizelySDKKey.staging
+  case .development, .local:
+    sdkKey = Secrets.OptimizelySDKKey.development
+  }
+
+  let allExperiments = optimizelyClient.allExperiments().map { experimentKey -> [String: String] in
+    do {
```
justinswart, line 97: Given that we aren't handling the exception, I think we only care about the value being `nil` so we could get away without repeating the dictionary keys:
```swift
let allExperiments = optimizelyClient.allExperiments().map { experimentKey -> [String: String] in
  let variation = try? optimizelyClient.getVariationKey(
    experimentKey: experimentKey,
    userId: userId,
    attributes: attributes
  )

  return [
    "optimizely_experiment_slug": experimentKey,
    "optimizely_variant_id": variation ?? "unknown"
  ]
}
```

PR #1173, opened 2020-04-23
[NT-1175] Coalesce experimental variants to control

justinswart:
# 📲 What

Coalesces optional `OptimizelyExperiment.Variant?` values to `.control`.

# 🤔 Why

We've received a handful of bug reports in which the _read more about the campaign_ button is missing on the project page. We suspect this might occur if the Optimizely client fails to initialize and the `skipNil()` on the signal that configures the button causes it to never emit.

# 🛠 How

Coalesced `nil` values to `.control` in functions where we return an optional value for the variant.

# ✅ Acceptance criteria

- [ ] Prevent the Optimizely client from being configured, navigate to the project page. The read more button should be visible.

### DIFF
--- a/Library/OptimizelyExperiment.swift
+++ b/Library/OptimizelyExperiment.swift
@@ -22,7 +22,7 @@ extension OptimizelyExperiment {
   static func projectCampaignExperiment(
     project: Project,
     refTag: RefTag?
-  ) -> OptimizelyExperiment.Variant? {
+  ) -> OptimizelyExperiment.Variant {
     return AppEnvironment.current.optimizelyClient?
       .variant(
         for: OptimizelyExperiment.Key.nativeProjectPageCampaignDetails,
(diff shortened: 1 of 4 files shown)


Library/OptimizelyExperiment.swift
```diff
@@ -27,6 +27,6 @@ extension OptimizelyExperiment {
       .variant(
         for: OptimizelyExperiment.Key.nativeProjectPageCampaignDetails,
         userAttributes: optimizelyUserAttributes(with: project, refTag: refTag)
-      )
```
ifbarrera, line 30: I think the function return value can also be changed to `OptimizelyExperiment.Variant` if we're coalescing?

justinswart: Ah thanks - I had it like that but then reverted to right a failing test and forgot to switch back 😅

PR #1174, opened 2020-04-24
[NT-1182] Fetch Backing for ManagePledgeViewController from GraphQL

justinswart:
# 📲 What

Fetches `ManagePledgeViewBackingEnvelope` from GraphQL for `ManagePledgeViewController`.

**Note:** Subsequent PRs will update the nested VCs to be configured with these new types as well as add a loading animation, etc. This PR is only to make the request and parse the data.

# 🤔 Why

We need additional data for this view and this is a step in moving away from the v1 REST API to get this data from GraphQL.

# 🛠 How

- Added `ManagePledgeViewBackingEnvelope`.
- Modified `GraphSchema`.
- Updated `ServiceType`.
- Added the request to `ManagePledgeViewModel` and used `ManagePledgeViewBackingEnvelope.Project` to set the navigation bar title.

# ✅ Acceptance criteria

- [ ] Schema changes.
- [ ] `enum` types that were extracted to be shared.
- [ ] When you navigate to `ManagePledgeViewController` observe that the navigation bar title is set by the GraphQL query response.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/models/ManagePledgeViewBackingEnvelopeTests.swift
```diff
@@ -0,0 +1,258 @@
...
+      XCTFail("Should have data")
+      return
+    }
+
+    let value = try? JSONDecoder().decode(ProjectSummaryEnvelope.self, from: data)
+
+    XCTAssertNotNil(value, "Should deserialize with only some values.")
+  }
+
+  func testJSONParsing_WithUnknownQuestion() {
```
ifbarrera, line 147: Unknown question? Que?

KsApi/models/ManagePledgeViewBackingEnvelopeTests.swift
```diff
@@ -0,0 +1,258 @@
...
+      XCTAssertEqual(value.backing?.reward?.amount, Money(amount: "129.0", currency: .usd, symbol: "$"))
+
+      XCTAssertEqual(value.backing?.backer?.id, "VXNlci0xMTA4OTI0NjQw")
+      XCTAssertEqual(value.backing?.backer?.name, "Backer McGee")
+    } catch {
+      XCTFail((error as NSError).description)
+    }
+  }
+
+  func testJSONParsing_WithPartialData() {
```
ifbarrera, line 121: I think this test is accidentally hanging out here 😄

KsApi/models/ProjectState.swift
```diff
@@ -0,0 +1,12 @@
+import Argo
```
ifbarrera, line 1: Do you need this Argo import?

KsApi/MockService.swift
```diff
@@ -76,16 +76,15 @@
     fileprivate let fetchGraphUserAccountFieldsResponse: UserEnvelope<GraphUser>?
     fileprivate let fetchGraphUserAccountFieldsError: GraphError?
 
-    fileprivate let fetchGraphUserBackingsResponse: UserEnvelope<GraphBackingEnvelope>?
```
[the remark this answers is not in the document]
justinswart: Oh, I couldn't find anything making use of the call. I'll confirm, thanks.

KsApi/models/ManagePledgeViewBackingEnvelope.swift
```diff
@@ -0,0 +1,96 @@
+import Foundation
+
+public struct ManagePledgeViewBackingEnvelope: Swift.Decodable {
+  public var project: Project
+  public var backing: Backing?
+
+  public struct Project: Swift.Decodable {
+    public var id: String
```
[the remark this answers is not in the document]
justinswart: Aren't we moving away from `Int` IDs?

KsApi/models/ManagePledgeViewBackingEnvelope.swift
```diff
@@ -0,0 +1,96 @@
...
+  private enum CodingKeys: CodingKey {
+    case backing
+    case project
+  }
+
+  init(from decoder: Decoder) throws {
+    let values = try decoder.container(keyedBy: CodingKeys.self)
+
+    self.project = try values.decode(Project.self, forKey: .project)
+    self.backing = try values.nestedContainer(keyedBy: CodingKeys.self, forKey: .project)
```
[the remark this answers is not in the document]
justinswart: This expects a nested container to be there, in this case `project`, and then decodes an optional `Backing?` so it should be ok if `Backing` is missing, will write a test to confirm.

Basically I'm flattening the envelope a little to put the `backing` alongside `project`.

KsApi/models/ManagePledgeViewBackingEnvelopeTests.swift
justinswart: lol copy paste fail, thanks.

KsApi/models/ManagePledgeViewBackingEnvelopeTests.swift
justinswart: This is supposed to be `testJSONParsing_NoErrorReason` I think, another copy-paste error.

KsApi/models/ProjectState.swift
justinswart: I think I needed it to disambiguate Decodable, will check.

KsApi/models/ManagePledgeViewBackingEnvelope.swift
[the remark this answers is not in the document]
ifbarrera: I think it depends what we end up using `id` for - I know we've pretty much been using `id` all over the place to mean either `relayId` or just "kickstarter" `id`, but I wonder if we want to be more explicit with this naming moving forward so there's less confusion around what this value actually is 🤔 we could consider naming these properties `graphId` or `relayId` but it would require a refactor of some of our other models. Maybe just a comment for now is fine?

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -260,3 +275,52 @@ private func managePledgeMenuCTAType(for managePledgeAlertAction: ManagePledgeAl
   case .viewRewards: return .viewRewards
   }
 }
+
+private func projectBackingQuery(withSlug slug: String) -> NonEmptySet<Query> {
+  return Query.project(
```
ifbarrera, line 280: Look at that beautiful query.

PR #1178, opened 2020-04-28
[NT-633] Errored Pledge Deeplink

justinswart:
# 📲 What

Adds deep-linking to `ManagePledgeViewController` for errored pledges via push notification and email link.

# 🤔 Why

This will allow users to easily navigate to fix their errored pledges when receiving a push notification or email.

# 🛠 How

- Updated `PushEnvelope` to deserialize the `errored_pledge` payload.
- Updated our `Navigation` `enum` to include a `.pledge(.manage)` route.
  - This looks the same as the root pledge URL and depends on the `ref` property in the email link being `ksr_email_backer_failed_transaction`.

# ✅ Acceptance criteria

#### Testing push notification deep-linking:

[Xcode 11.4](https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes) supports testing push notifications in the simulator:

1. Create a file called `test_push.apns`.
2. Paste the JSON snippet below. Replace the `project_id` value with the ID of a project that has an errored pledge.
3. In terminal type `xcrun simctl push booted com.kickstarter.kickstarter.debug test_push.apns`
```{
  "aps": {
    "alert": "Errored pledge"
  },
  "errored_pledge": {
    "project_id": 913348030 
  }
}
```
#### Testing email deep-linking:

Tap on a link for an errored project with the format:

https://www.kickstarter.com/projects/creator_param/project-slug/pledge?ref=ksr_email_backer_failed_transaction

- [ ] Tapping errored pledge push notification navigates to `ManagePledgeViewController`.
- [ ] Tapping on the link to fix your pledge in an email navigates to `ManagePledgeViewController`.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -538,6 +538,14 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
           }
       }
 
+    let fixErroredPledgeLink = projectLink
+      .filter { _, subpage, _, _ in subpage == .pledge(.manage) }
+      .map { project, _, vcs, _ -> [UIViewController] in
+        let vc = ManagePledgeViewController.instantiate()
+        vc.configureWith(project: project)
+        return vcs + [vc]
+      }
+
     let updateRootLink = updateLink
       .filter { _, _, subpage, _ in subpage == .root }
       .map { _, _, _, vcs in vcs }
@@ -558,9 +566,11 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
         updatesLink,
         updateRootLink,
         updateCommentsLink,
-        campaignFaqLink
+        campaignFaqLink,
+        fixErroredPledgeLink
       )
       .map { UINavigationController() |> UINavigationController.lens.viewControllers .~ $0 }
+      .logEvents(identifier: "***")
 
     self.configureFabric = self.applicationLaunchOptionsProperty.signal.ignoreValues()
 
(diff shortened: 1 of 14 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -538,6 +538,19 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
           }
       }
 
+    let fixErroredPledgeLink = projectLink
+      .filter { _, subpage, _, _ in subpage == .pledge(.manage) }
+      .map { project, _, vcs, _ -> [UIViewController] in
+        let vc = ManagePledgeViewController.instantiate()
+        vc.configureWith(project: project)
```
ifbarrera, line 545: Should we do some sort of check in here that the `Project` indeed contains a `Backing` for this user? Maybe if the user is logged out of the app at the time they hit the deeplink, for example.

justinswart: Ah yeah good point, I can add that.  This would have the effect of nothing happening after tapping a push/email link when logged out though, wondering if a better way to handle this is either to prompt login (probably tricky to handle the navigation after login), or have some error messaging on `ManagePledgeViewController` 🤔

justinswart: Apologies for the delay - got swept up in that layout bug. I'm going to send the user to the login flow in this case.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -436,33 +479,9 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
-
-    let projectLink = projectLinkValues
-      .filter { project, _, _, _ in project.prelaunchActivated != true }
-
-    let projectPreviewLink = projectLinkValues
-      .filter { project, _, _, _ in project.prelaunchActivated == true }
-
     let resolvedRedirectUrl = deepLinkUrl
       .filter { Navigation.deepLinkMatch($0) == nil }
+      .logEvents(identifier: "***")
```
ifbarrera, line 484: Remove?

justinswart: 😬

PR #1181, opened 2020-04-30
Bump version to 4.5.1 and skip_app_version_update true

ifbarrera:
# 📲 What

Bumps the version, and sets `skip_app_version_update` to `true`, so that fastlane doesn't automatically update the app version number for the next release.

# 🤔 Why

We had an issue where uploading a build to Testflight while another build was in review automatically updated the version number on the build in review 🙅‍♀️ 

We want to prevent this from happening in the future.


ifbarrera:
Seems that we don't want to turn this off after all!

PR #1184, opened 2020-05-05
[NT-1089] "native_project_cards" experiment setup

ifbarrera:
# 📲 What

Sets up the `native_project_cards` experiment and prepares the necessary outputs for implementation of this experiment.

# 🤔 Why

So that we can experiment on the design of the native project cards.

# 🛠 How

In order to properly show this experiment, we need to guarantee that we know the experimental variant before any project cards are loaded. In order to accomplish this, while also ensuring that we're not over-calling Optimizely's `activate` function on every instance of `DiscoveryPageViewController`, the logic for calling `activate` on the `native_project_cards` experiment was moved to `DiscoveryViewModel`, which is the higher level VM that manages the configuration and updating of the parent `pageViewController`'s VCs. By moving the Optimizely configuration and activation logic to this level, we can be guaranteed that the experiment variation has been assigned by the time `getVariation` is called in `DiscoveryPageViewModel`.

This PR:
- adds logic that waits for Optimizely configuration to be finished before loading the child `DiscoveryPageViewController`s into the `DiscoveryPagesDataSource`
- activates the `native_project_cards` experiment in `DiscoveryViewModel`. In order to get the variation in `DiscoveryPageViewModel` to be passed down to the VC, we call `getVariation`
- adds a `Optimizely.Variant` parameter to the `projectsLoaded` output on `DiscoveryPageViewModel`. This will be used to determine which class of project cards to load into the data source
- adds a `backgroundColor` output in `DiscoveryPageViewModel` that is updated based on the variant
- adds a `optimizelyConfigurationFailed` input in `AppDelegateViewModel` that is used to notify the VM when Optimizely configuration fails. This then fires a notification that can be observed so logic can proceed when there is a configuration failure

# 👀 See

A device white-listed into the experimental variant (grey background):

R3UvdbWK8J

# ♿️ Accessibility 

N/A

# 🏎 Performance

N/A

# ✅ Acceptance criteria

- [x] all Discovery navigation should be working as expected: changing filters, changing sorts, pagination
- [x] whitelist your device into the `native_project_cards` experiment. You should see the background of discovery pages is grey


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/DiscoveryViewController.swift
```diff
@@ -119,7 +141,7 @@ internal final class DiscoveryViewController: UIViewController {
     self.pageViewController.dataSource = self.dataSource
 
     self.pageViewController.ksr_setViewControllers(
-      [self.dataSource.controllerFor(index: 0)].compact(),
+      [self.dataSource!.controllerFor(index: 0)].compact(),
```
justinswart, line 144: Shouldn't need the force-unwrap here given that we're compacting the array.

Kickstarter-iOS/Views/Controllers/EditorialProjectsViewController.swift
```diff
@@ -76,6 +76,9 @@ public final class EditorialProjectsViewController: UIViewController {
       self.discoveryPageViewController.tableView.contentInset
 
     self.headerTopLayoutGuideHeightConstraint?.constant = self.view.safeAreaInsets.top
+
+    self.view.bringSubviewToFront(self.headerView)
```
justinswart, line 80: 🤔 Couldn't we just add them on top?

ifbarrera: Oh yeah that works!

PR #1185, opened 2020-05-05
[NT-1221] Lights On Editorial Card

cdolm92:
# 📲 What

Lights On view for the Lights On Editorial Collection
# 🤔 Why

So that backers can support backers that have been affected by COVID-19.

# 🛠 How

Created a new cell `DiscoveryLightsOnEditorialCell` to display in the Discovery screen.

# 👀 See

Simulator Screen Shot - iPhone 11 - 2020-05-05 at 19 22 40

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# ✅ Acceptance criteria

Note: Designs should NOT include the "x" close button

- [ ] Lights on editorial card is shown in the home screen for all users
- [ ] The card appears below Log In/Sign up Prompt and above activity (if present) and the onboarding experiment card (if present)


Kickstarter-iOS/DataSources/DiscoveryProjectsDataSource.swift
```diff
@@ -11,6 +11,7 @@ struct DiscoveryProjectCellRowValue {
 internal final class DiscoveryProjectsDataSource: ValueCellDataSource {
   internal enum Section: Int {
     case onboarding
+    case lightsOnEditiorial
```
justinswart, line 14: Typo here - `lightsOnEditiorial` -> `lightsOnEditorial`.

Kickstarter-iOS/Views/Cells/DiscoveryLightsOnEditorialCell.swift
```diff
@@ -0,0 +1,134 @@
+import Foundation
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+final class DiscoveryLightsOnEditorialCell: UITableViewCell, ValueCell {
```
justinswart, line 7: Given that this is so similar to `DiscoveryEditorialCell`, which is named in such a way that it's not specific to a particular editorial, could we not reuse that cell for this?

Library/ViewModels/DiscoveryLightsOnEditorialViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+  var subtitleText: Signal<String, Never> { get }
+  var titleText: Signal<String, Never> { get }
+}
+
+public protocol DiscoveryLightsOnEditorialViewModelType {
+  var inputs: DiscoveryLightsOnEditorialViewModelInputs { get }
+  var outputs: DiscoveryLightsOnEditorialViewModelOutputs { get }
+}
+
+public final class DiscoveryLightsOnEditorialViewModel: DiscoveryLightsOnEditorialViewModelType,
```
justinswart, line 26: This is also the same as `DiscoveryEditorialViewModel` (should be `DiscoveryEditorialCellViewModel`), I think there is opportunity to reuse this code.

Library/ViewModels/EditorialProjectsViewModel.swift
```diff
@@ -112,11 +112,17 @@ private func editorialImageName(for tagId: DiscoveryParams.TagID) -> String {
...
   }
+
+  return "go-rewardless-modal"
 }
 
 private func editorialTitleLabelText(for tagId: DiscoveryParams.TagID) -> String {
   switch tagId {
   case .goRewardless: return Strings.This_holiday_season_support_a_project_for_no_reward()
+  case .lightsOn:
+    return Strings.This_holiday_season_support_a_project_for_no_reward()
```
cdolm92, line 126: To address this in upcoming PR

Kickstarter-iOS/Views/Cells/DiscoveryEditorialCell.swift
```diff
@@ -109,15 +117,17 @@ final class DiscoveryEditorialCell: UITableViewCell, ValueCell {
   // MARK: - Configuration
 
   private func configureViews() {
-    _ = (self.containerView, self.contentView)
+    _ = (self.editorialImageView, self.contentView)
       |> ksr_addSubviewToParent()
       |> ksr_constrainViewToMarginsInParent()
 
-    _ = (self.rootStackView, self.containerView)
+    _ = (self.containerView, self.editorialImageView)
```
justinswart, line 124: We shouldn't make regular `UIView`s subviews of `UIImageView`s. In this case you can just add `editorialImageView` to `contentView` and the `containerView` above it both constrained to the edges of `contentView`.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -219,6 +219,14 @@ internal final class DiscoveryPageViewController: UITableViewController {
         self?.tableView.reloadData()
       }
 
+    self.viewModel.outputs.showLightsOnEditorialHeader
```
justinswart, line 222: We can just reuse the `showEditorialHeader` output for this.

Library/ViewModels/EditorialProjectsViewModel.swift
```diff
@@ -112,11 +112,17 @@ private func editorialImageName(for tagId: DiscoveryParams.TagID) -> String {
       return "go-rewardless-modal-large"
     }
     return "go-rewardless-modal"
+  case .lightsOn:
+    return "go-rewardless-modal"
```
justinswart, line 116: Seeing as we're removing go rewardless we probably don't need this function (unless we need a different image for iPad), this seems to return the image name for go rewardless though?

cdolm92: I'll address that in the next PR for the editorial list. I'm not including the asset with this PR, it's in the next.

PR #1187, opened 2020-05-07
[NT-1240] PledgeViewController layout bugfix

justinswart:
# 📲 What

Fixes an inconsistent layout bug in `PledgeViewController`.

# 🤔 Why

We discovered an issue that would cause `PledgeDisclaimerViewController` and `PledgeDescriptionViewControllers` to be arbitrarily taller than necessary.

# 🛠 How

This was a difficult one to track down but in the end @ifbarrera and I were able to prevent it from happening by changing the `PledgeDisclaimerViewController` and `PledgeDescriptionViewController` to be regular `UIView`s instead of nested `UIViewController`s. It's still not entirely clear to us why this change fixes the problem but it'll do for now.

In trying to track this down I was also able to simplify the layout and prevent an ambiguous content size in the `rootScrollView`.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

Navigate to the pledge view.

- [ ] All views should have the correct height.
- [ ] The heights should be correct for rewards that have shipping.
- [ ] The heights should be correct for rewards that don't have shipping.
- [ ] Dragging/interacting with the view while payment methods and shipping loads should not affect the layout.


Kickstarter-iOS/Views/Controllers/PledgeAmountViewController.swift
```diff
@@ -144,6 +144,10 @@ final class PledgeAmountViewController: UIViewController {
       }
   }
 
+  override func didMove(toParent parent: UIViewController?) {
+    self.verticalSpacer.isHidden = true
```
justinswart, line 148: This spacer is not needed when this view is contained and was causing the scrollview to have an ambiguous content size.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -142,6 +155,17 @@ final class PledgeViewController: UIViewController,
     self.viewModel.inputs.viewDidLoad()
   }
 
+  override func viewDidLayoutSubviews() {
+    super.viewDidLayoutSubviews()
+
+    self.rootScrollView.contentInset = UIEdgeInsets(
```
justinswart, line 161: Adjusting content insets instead of pinning the scrollview to the top of the button container.

justinswart:
> I was only able to reproduce this on a 11 pro max, however when you tap "new card payment" and then dismiss the view, the "kickstarter is not a store" view is stuck below the pledge buttons.

@jgsamudio thanks for testing. I wasn't able to reproduce - could you perhaps attach a screenshot?

PR #1196, opened 2020-05-14
[NT-1224] Add Editorial Lights On Optimizely feature flag

justinswart:
# 📲 What

Adds the `ios_lights_on` Optimizely feature flag and removes all references to Go Rewardless.

# 🤔 Why

This allows us to activate the Lights On editorial feature and future features using Optimizely.

# 🛠 How

- Updated our `OptimizelyClientType` to support feature flags and added tests to the mock type.
- Searched for and removed all references to Go Rewardless as that campaign is no longer running.

# ✅ Acceptance criteria

- [x] Activate the `ios_lights_on` feature in Optimizely - you should see the feature as active when starting the app.
- [x] Deactivate the `ios_lights_on` feature in Optimizely - you should not see the feature when starting the app.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/OptimizelyClientType.swift
```diff
@@ -62,6 +63,14 @@ extension OptimizelyClientType {
 
     return variant
   }
+
+  public func isFeatureEnabled(featureKey: String) -> Bool {
```
ifbarrera, line 67: Does this trigger an impression event similar to `activate` for experiments? If so, can we just add a comment above the function?

Library/OptimizelyClientTypeTests.swift
```diff
@@ -76,4 +76,20 @@ final class OptimizelyClientTypeTests: TestCase {
       XCTAssertTrue(mockClient.getVariantPathCalled)
     }
   }
+
+  func testIsFeatureEnabled() {
+    let mockClient = MockOptimizelyClient()
+      |> \.features .~ [
+        "my_enabled_feature": true,
+        "my_disbled_feature": false
```
ifbarrera, line 84: Typo here

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -331,7 +331,7 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
     )
 
     self.configureEditorialTableViewHeader = paramsChanged
-      .filter { $0.tagId == .goRewardless }
```
ifbarrera, line 334: Actually I noticed while testing that the light's on projects view has the "These projects could use your support" header -- it looked a bit odd, I'm not sure if we want to keep it?

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -349,18 +349,18 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
       }
 
     let cachedFeatureFlagValue = self.sortProperty.signal.skipNil()
-      .map { _ in featureGoRewardlessIsEnabled() }
-    let updatedFeatureFlagValue = self.configUpdatedProperty.signal.skipNil()
-      .map { _ in featureGoRewardlessIsEnabled() }
+      .map { _ in editorialLightsOnFeatureIsEnabled() }
+    let updatedFeatureFlagValue = self.optimizelyClientConfiguredProperty.signal
```
ifbarrera, line 353: So I think this logic needs to change a little bit now that we're using Optimizely for feature flags instead of the config. I think you can just use `optimizelyReady` and remove `latestFeatureFlag` stuff, which was in place before because it was reading the feature flag from the `Config`.

Library/OptimizelyClientType.swift
justinswart: The docs don't say anything explicitly about triggering an impression:

> This method traverses the client's datafile and checks the feature flag for the feature key that you specify. [...] The method then evaluates the feature rollout for a user. The method checks whether the rollout is enabled, whether the user qualifies for the audience targeting, and then randomly assigns either on or off based on the appropriate traffic allocation. If the feature rollout is on for a qualified user, the method returns true.

Library/OptimizelyClientTypeTests.swift
justinswart: Thanks 😅

Library/ViewModels/DiscoveryPageViewModel.swift
justinswart: Ah, yes forgot to remove that, thanks.

Library/ViewModels/DiscoveryPageViewModel.swift
justinswart: I was thinking it might help with race conditions if Optimizely takes time to configure and the data file changes resulting in the feature being enabled/disabled?

PR #1198, opened 2020-05-18
[NT-1256] iOS 12 pledge screen scroll view insets bug

ifbarrera:
# 📲 What

Fixes a bug in iOS 12 where presenting a VC from the pledge screen caused the scroll view insets to be reset to zero, which caused the pledge button container to overlap some content near the bottom of the screen.

# 🤔 Why/How

This was a strange one. For some reason on iOS 12.4, presenting a VC on top of a scroll view with its `contentInsetAdjustmentBehavior` set to `automatic` (the default), caused the content insets to abruptly be reset back to zero when the presented VC is dismissed. Strangely, examining the `contentInsets` after the dismissal showed that the `adjustedContentInsets` were still correctly set, but visually the scroll view appeared to have an `adjustedContentInset` of `.zero`. Setting the `contentInsetAdjustmentBehavior` on the `rootScrollView` to `.zero` seemed to fix the issue, but it caused a strange side effect where the scroll view no longer bounced properly.

As a workaround, this PR pins the `rootScrollView`'s `bottom` to the top of the `pledgeCTAContainerView` so we don't rely on the `contentInsets` being correct.

# 👀 See

| After 🦋 | Before 🐛 |
| --- | --- |
| xMB3tAfJzs | Fo4wht5WaJ |

# ✅ Acceptance criteria

- [x] on an iOS 12 device, navigate to the pledge screen and check that the view scrolls up fully without obscuring content. Then, tap the "New Payment Method" button and close the view. Check that the view scrolls up fully without obscuring content.
- [x] on an iOS 13 device, navigate to the pledge screen and check that the view scrolls up fully without obscuring content. Then, tap the "New Payment Method" button and close the view. Check that the view scrolls up fully without obscuring content.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
@@ -155,17 +155,6 @@ final class PledgeViewController: UIViewController,
     self.viewModel.inputs.viewDidLoad()
   }
 
-  override func viewDidLayoutSubviews() {
-    super.viewDidLayoutSubviews()
-
-    self.rootScrollView.contentInset = UIEdgeInsets(
-      top: self.rootScrollView.contentInset.top,
-      left: self.rootScrollView.contentInset.left,
-      bottom: self.pledgeCTAContainerView.frame.size.height - self.view.safeAreaInsets.bottom,
-      right: self.rootScrollView.contentInset.right
-    )
-  }
-
   deinit {
     self.sessionStartedObserver.doIfSome(NotificationCenter.default.removeObserver)
   }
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -214,8 +203,7 @@ final class PledgeViewController: UIViewController,
   }
 
   private func setupConstraints() {
-    _ = (self.rootScrollView, self.view)
-      |> ksr_constrainViewToEdgesInParent()
+    self.rootScrollView.translatesAutoresizingMaskIntoConstraints = false
```
justinswart, line 206: This is already set in the `lazy var` property.

PR #1201, opened 2020-05-20
[NT-1236] Fix share URLs in messaging apps

justinswart:
# 📲 What

Fixes broken shared project links in some messaging apps.

# 🤔 Why

We noticed that some URLs weren't being shared correctly in some apps, for example Telegram:

This appears to be due to a space character being incorrectly prepended to the beginning of the URL.

# 🛠 How

Updated the code that constructs this string to instead use the `joined(separator:)` function rather that multi-line string interpolation.

# ✅ Acceptance criteria

- [ ] Share a project link using messaging, observe that there are no spaces at the beginning of the project title or URL.

### DIFF
--- a/Library/ViewModels/ShareViewModel.swift
+++ b/Library/ViewModels/ShareViewModel.swift
@@ -195,20 +195,3 @@ private func activityController(forShareContext shareContext: ShareContext) -> U
 
   return controller
 }
-
-private func twitterInitialText(forShareContext shareContext: ShareContext) -> String {
-  switch shareContext {
-  case let .creatorDashboard(project):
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ShareViewModel.swift
```diff
@@ -195,20 +195,3 @@ private func activityController(forShareContext shareContext: ShareContext) -> U
 
   return controller
 }
-
-private func twitterInitialText(forShareContext shareContext: ShareContext) -> String {
```
justinswart, line 199: This function was unused.

PR #1202, opened 2020-05-20
[NT-1015] Project Card Clicked Event

ifbarrera:
# 📲 What

Adds the "Project Card Clicked" event when a project is selected from discovery or the thanks page.

# 🤔 Why

So we can track what projects users are navigating to from various locations.

# 🛠 How

- add the `Project Card Clicked` event to the data lake whitelist
- fire the event when a project is clicked from discovery
- fire the event when a project is clicked from the Thanks page

# 👀 See

image

# ✅ Acceptance criteria

- [x] turn the `KOALA_TRACKING` environment flag to `true`, then run the app and a tap a project from discovery. You should see the `Project Card Clicked` event
- [x] Run the app and back a project. On the Thanks page, tap any project. You should see the `Project Card Clicked` event

### DIFF
--- a/Library/ViewModels/ThanksViewModel.swift
+++ b/Library/ViewModels/ThanksViewModel.swift
@@ -179,16 +179,25 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs, 
       }
 
     project
-      .takeWhen(self.goToProject)
-      .observeValues { project in
-        AppEnvironment.current.koala.trackCheckoutFinishJumpToProject(project: project)
-      }
-
-    project
       .takeWhen(self.showRatingAlert)
       .observeValues { project in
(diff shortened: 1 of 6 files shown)


Library/ViewModels/ThanksViewModel.swift
```diff
@@ -178,18 +178,27 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
-      }
-
     project
       .takeWhen(self.showRatingAlert)
       .observeValues { project in
         AppEnvironment.current.koala.trackTriggeredAppStoreRatingDialog(project: project)
       }
 
+    self.projectTappedProperty.signal.skipNil().map { project in
+      let recommendedParams = DiscoveryParams.defaults
```
justinswart, line 188: Is this to match it to `recommendedParams` below on line 296?

Perhaps we should extract that to a global variable to be sure they are they same?

PR #1203, opened 2020-05-21
[NT-1247] ManagePledgeViewController loading state

justinswart:
# 📲 What

- Displays an error label on errored payment method when fixing an errored pledge.
- Adds a loading state to `ManagePledgeViewController`.

# 🤔 Why

These are just some outstanding finishes to the work to allow users to fix their errored pledges in the app. The loading state for `ManagePledgeViewController` is necessary now that we are fetching the backing for that view from GraphQL.

# 🛠 How

- Displaying text on errored payment methods in `PledgePaymentMethodCell`.
- Hiding the contents of `ManagePledgeViewController` while the project and backing for that view is fetched.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Navigate to manage pledge, observe that there is an activity indicator shown until the backing info is displayed.
- [ ] If the request fails, an error banner should be shown and the view (for now) remains blank.
- [ ] Navigate to change payment method for an errored pledge, you should see a label in the payment method cell indicating that you can retry that payment method.
- [ ] When selecting the payment method to retry, the submit button should read _Retry_.

# ⏰ TODO

- [ ] Might need to display something on the view when it fails to load.


Library/UIRefreshControl+StartRefreshing.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+import UIKit
+
+public extension UIRefreshControl {
+  func ksr_beginRefreshing() {
```
justinswart, line 5: Fixes an issue that prevented the refresh control from animating when the view appears. I think we have this same issue on `DiscoveryPageViewController`, this might be a simpler way of solving that 🤔

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -167,14 +180,25 @@ public final class ManagePledgeViewModel:
...
         // swiftformat:disable wrap
         Strings.We_dont_allow_cancelations_that_will_cause_a_project_to_fall_short_of_its_goal_within_the_last_24_hours()
         // swiftformat:enable wrap
       }
 
+    let networkErrorMessage = Signal.merge(
+      fetchProjectEvent.errors().ignoreValues(),
+      graphBackingEvent.errors().ignoreValues()
+    )
+    .map { _ in Strings.Something_went_wrong_please_try_again() }
```
justinswart, line 195: Do we have better messaging? We may also need to show something on the view when this happens. Currently it remains blank.

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -28,13 +28,18 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
   private let showActionSheetMenuWithOptions = TestObserver<[ManagePledgeAlertAction], Never>()
   private let showErrorBannerWithMessage = TestObserver<String, Never>()
   private let showSuccessBannerWithMessage = TestObserver<String, Never>()
+  private let startRefreshing = TestObserver<(), Never>()
   private let title = TestObserver<String, Never>()
 
   override func setUp() {
     super.setUp()
+
+    self.vm = ManagePledgeViewModel()
```
justinswart, line 41: Necessary for the test scheduler to play nice.

Library/ViewModels/PledgeViewCTAContainerViewModelTests.swift
```diff
@@ -98,10 +122,10 @@ internal final class PledgeViewCTAContainerViewModelTests: TestCase {
     self.vm.inputs.configureWith(value: pledgeData)
 
     self.submitButtonIsHidden.assertValues([false])
-    self.applePayButtonIsHidden.assertValues([false])
+    self.applePayButtonIsHidden.assertValues([true])
```
ifbarrera, line 125: I was wondering about this - didn't we use to support Apple Pay in Change Payment Method? 🤔

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -114,6 +116,17 @@ public final class ManagePledgeViewModel:
       .map { project, env in managePledgeSummaryViewData(with: project, envelope: env) }
       .skipNil()
 
+    self.rootStackViewHidden = Signal.merge(
+      projectOrParam.mapConst(true),
+      Signal.zip(backing, self.endRefreshing).mapConst(false)
```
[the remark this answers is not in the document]
justinswart: Yeah will continue to hide the contents, I am going to add a label and icon to suggest pull to refresh.

Library/ViewModels/PledgeViewCTAContainerViewModelTests.swift
justinswart: Ah, good catch - yes, will update.

PR #1204, opened 2020-05-21
[NT-1251] “You're a backer” label

ifbarrera:
# 📲 What

Adds the "You're a backer" label to the project cards when the current user has backed that project.

# 🤔 Why

To signal to users which projects they have backed while browsing through projects in discovery.

# 🛠 How

- added the `youreABackerView` and `youreABackerLabel`
- added the `youreABackerViewHidden` output to `DiscoveryProjectCardViewModel`
- added screenshot and VC tests
- notice that I updated the layout margins on the main description stack view, so I had to recreate all the snapshots

# 👀 See

Simulator Screen Shot - iPhone 8 - 2020-05-21 at 15 14 51

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [ ] back a project from discovery and then find that project again in the discovery list. You should see the "You're a backer" label

# ⏰ TODO

- [ ] get final approval from Colleen on the position of the tag, I had to make a small change compared to the designs


vickisiolosksr:
Can we remove the exclamation point from this string?

ifbarrera:
> Looks good! I know there was this bug at one point that discovery doesn't refresh to show this after backing. That's what I saw now, is that still expected? 🤔

Yeah that's still a bug, we'll have to fix it separately.

PR #1206, opened 2020-05-22
[NT-1274] Pull to refresh on failure to fetch Project/Backing

justinswart:
# 📲 What

Displays error messaging suggesting to the user to pull to refresh if we fail to fetch the project/backing when navigating to `ManagePledgeViewController`.

# 🤔 Why

We need to handle possible errors when fetching this data.

# 🛠 How

Updated `ManagePledgeViewModel` to handle the scenarios described in the ACs below.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] Navigate to `ManagePledgeViewController` from backed projects -> GraphQL Backing fails to fetch, display error, retry on pull to refresh. View is reloaded.
- [x] Navigate to `ManagePledgeViewController` from deeplink -> GraphQL Backing fails to fetch, display error, retry on pull to refresh. View is reloaded.
- [ ] Navigate to `ManagePledgeViewController` from an errored backing in the activity feed, -> Project fails to fetch, display error, retry on pull to refresh. View is reloaded.
- [ ] Navigate to `ManagePledgeViewController` from an errored backing in the activity feed, -> Project succeeds, GraphLQ backing fails to fetch, display error, retry on pull to refresh. View is reloaded. (Project is not fetched a second time if it succeeded before).

# ⏰ TODO

- [ ] Confirm copy.


Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -64,13 +65,25 @@ public final class ManagePledgeViewModel:
...
+    let projectLoaded = MutableProperty<Bool>(false)
+
+    let shouldFetchProjectWithProjectOrParam = Signal.merge(
+      projectOrParam,
+      projectOrParam.takeWhen(shouldBeginRefresh)
+    )
+
+    let fetchProjectEvent = shouldFetchProjectWithProjectOrParam
+      // Only fetch the project if it hasn't yet succeeded, to avoid this call occurring with each refresh.
+      .filter { [projectLoaded] _ in projectLoaded.value == false }
```
ifbarrera, line 78: 🤔 I thought the `.switchMap` took care of the scenario where a request is triggered while another is in-flight?

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -93,11 +106,22 @@ public final class ManagePledgeViewModel:
     let backing = graphBackingEnvelope
       .map { $0.backing }
 
-    self.endRefreshing = graphBackingEvent
+    let endRefreshingWhenProjectFailed = fetchProjectEvent
```
ifbarrera, line 109: Why not `fetchProjectEvent.errors().ignoreValues()`?

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -116,14 +140,29 @@ public final class ManagePledgeViewModel:
       .map { project, env in managePledgeSummaryViewData(with: project, envelope: env) }
       .skipNil()
 
+    let projectOrBackingFailedToLoad = Signal.merge(
+      fetchProjectEvent.map { $0.error as Error? },
```
ifbarrera, line 144: Also here, `fetchProjectEvent.errors()`? You shouldn't have to `.filter(isNotNil)` because `.errors()` only sends an event if there's an error.

Library/ViewModels/ManagePledgeViewModel.swift
justinswart: We chatted IRL but just to confirm - this value tracks that the project has been successfully fetched so that if the backing fails to fetch and we pull to refresh, it doesn't try to fetch the project again unnecessarily.

Library/ViewModels/ManagePledgeViewModel.swift
justinswart: Sure yeah that should work 👍

Library/ViewModels/ManagePledgeViewModel.swift
justinswart: In this case it was to erase the different types of errors so that they could be merged.

PR #1209, opened 2020-05-26
[NT-1249] Adding facepile to new project cards

ifbarrera:
# 📲 What

Added the social facepile to the new native project cards, to show when friends have backed that project.

# 🤔 Why

We want to show users when their friends have backed certain projects.

# 🛠 How

- added the subviews needed, added VM & screenshot tests

# 👀 See

Simulator Screen Shot - iPhone 11 Pro - 2020-05-26 at 14 59 18

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

Replace the following code in `DiscoverPageViewMode.swift` to guarantee experimental treatment on production:
```
private func nativeProjectCardsExperimentVariant() -> OptimizelyExperiment.Variant {
//  guard let optimizelyClient = AppEnvironment.current.optimizelyClient else {
//    return .control
//  }
//
//  let variant = optimizelyClient.getVariation(for: .nativeProjectCards)
//
//  return variant

  return .variant1
}
```
- [x] in the production environment, log into `therealnativesquad` account, and navigate to "Backed by people you follow". Verify that the facepile view appears above the category tags
- [x] When 1 friends has backed, the copy should read "[Friend_name] is a backer"
- [x] When 2 friends have backed the copy should read "[Friend_name] and 1 other"
- [x] When > 2 friends have backed the copy should read "[Friend_name] and [x-1] others"

# ⏰ TODO

- [x] copy will be updated in a subsequent PR addressing all copy needs


Kickstarter-iOS/Views/Cells/DiscoveryProjectCardCell.swift
```diff
@@ -354,7 +386,13 @@ final class DiscoveryProjectCardCell: UITableViewCell, ValueCell {
     _ = ([self.backersCountIconImageView, self.backersCountLabel], self.backersCountStackView)
       |> ksr_addArrangedSubviewsToStackView()
 
-    _ = ([self.goalPercentFundedStackView, self.backersCountStackView], self.projectInfoStackView)
+    let spacer = UIView(frame: .zero)
+      |> UIView.lens.contentHuggingPriority(for: .horizontal) .~ UILayoutPriority(rawValue: 249)
```
justinswart, line 390: `.defaultLow` didn't work here?

Library/ViewModels/DiscoveryProjectCardViewModel.swift
```diff
@@ -160,3 +173,34 @@ private func projectTags(project: Project, shouldShowPWLTag: Bool, shouldShowCat
...
 }
+
+private func facepileData(for friendsList: [User]) -> FacepileViewData? {
+  guard !friendsList.isEmpty else { return nil }
+
+  let avatars = friendsList.prefix(3).map(\.avatar.small).compactMap(URL.init(string:))
+
+  var description: String = ""
+
+  if friendsList.count == 1 {
```
justinswart, line 184: A `switch` on the count might read a bit better here. `description` can also be a `let` if the compiler is sure it'll be instantiated.

Kickstarter-iOS/Views/Cells/DiscoveryProjectCardCell.swift
ifbarrera: The default setting for `contentHuggingPriority` on views is `defaultLow`, so this needed something that was just below that value.

PR #1211, opened 2020-05-29
[NT-1299] Apple Sign In Processing View dismissal bugfix

justinswart:
# 📲 What

Fixes an bug which caused the processing view to never dismiss after logging in via Apple Sign In from the profile tab.

# 🤔 Why

A number of things are happening all at once here:

- Apple Pay Sign In event completes.
- User object is refreshed.
- User is logged into the environment.
- Notifications are posted for the app to update its state.
- Processing view is dismissed.

Once the notifications are posted about the user login state having changed our `RootViewModel` recreates all of the VCs and sets them again on the tab bar. This was causing the `LoginToutViewController` to be deallocated before the observing closure could execute and hide the processing view.

One way of solving this would be to have another input on the view model to inform of the processing view having been dismissed. Another fairly simple solution is to place the posting of the notifications on the next run loop with a zero nanosecond delay. This PR implements the latter and adds a test.

# 🛠 How

Imposed a zero nanosecond delay on `LoginToutViewModel`'s `postNotification` output to ensure that it will emit last.

# ✅ Acceptance criteria

- [ ] In a logged out state, tap on the profile tab, sign in with Apple - you should _not_ find yourself stuck with the processing view never dismissing.

### DIFF
--- a/Library/ViewModels/LoginToutViewModelTests.swift
+++ b/Library/ViewModels/LoginToutViewModelTests.swift
@@ -9,7 +9,7 @@ import ReactiveSwift
 import XCTest
 
 final class LoginToutViewModelTests: TestCase {
-  fileprivate let vm: LoginToutViewModelType = LoginToutViewModel()
+  fileprivate var vm: LoginToutViewModelType!
 
   fileprivate let appleButtonHidden = TestObserver<Bool, Never>()
   fileprivate let attemptAppleLogin = TestObserver<(), Never>()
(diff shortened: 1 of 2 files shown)


Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -31,6 +31,8 @@ final class LoginToutViewModelTests: TestCase {
   override func setUp() {
     super.setUp()
 
+    self.vm = LoginToutViewModel()
```
justinswart, line 34: In order for the test to pass we needed to create the VM after the scheduler is set in the environment. We've had to do this in a number of places.

PR #1216, opened 2020-06-08
[NT-1223, NT-1296] Reward Add-Ons GraphQL integration

justinswart:
# 📲 What

- Adds reward add-ons GraphQL integration to `ManagePledgeViewModel`.
- Displays the selected quantity in the title of add-ons where more than 1 was selected of that add-on.

**Note:** The section title above the reward cell will be brought back in a subsequent PR.

# 🤔 Why

This is required for us to be able to display add-on rewards in `ManagePledgeViewController`.

# 🛠 How

- Added required fields to `GraphSchema`.
- Added properties to `ManagePledgeViewBackingEnvelope`.
- Added some data-adapter code to transform `ManagePledgeViewBackingEnvelope.Backing.Reward`s into regular v1 `Reward`s in order to reduce the amount of refactoring required for views depending on the v1 model.

# 👀 See

From Figma:

# ✅ Acceptance criteria

Navigate to manage/view a backing that has reward add-on's selected.

- [ ] The selected add-ons should be displayed below the primary reward.
- [ ] If multiple were selected for a particular add-on this should be indicated in the title of the add-on.


KsApi/models/Reward+ManagePledgeView.swift
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+private func calculateConvertedMinimum(
+  from backingReward: ManagePledgeViewBackingEnvelope.Backing.Reward,
+  with project: Project
+) -> Double {
+  let rate = project.stats.currentCurrencyRate ?? project.stats.staticUsdRate
+
+  return Double(Int(ceil(Float(backingReward.amount.amount) * rate)))
```
ifbarrera, line 56: Any chance they could send you this value from the graphQL API? 😬 

I know we did some work on the `v1` `Reward` model to have the backend send converted minimum amounts (`converted_minimum`), would be nice to keep it consistent.

justinswart: Yeah we're going to schedule back-end work to address this, this is just for the interim 😅

KsApi/models/Reward+ManagePledgeView.swift
```diff
@@ -0,0 +1,91 @@
...
+    return RewardsItem(
+      id: 0, // not returned
+      item: Item(
+        description: nil, // not returned
+        id: id,
+        name: item.name,
+        projectId: project.id
+      ),
+      quantity: 0, // not needed
+      rewardId: Int(backingReward.id) ?? -1
```
justinswart, line 75: Note to self - I think this needs to use `decompose(id:)` and we can move it to the `guard let` above.

PR #1228, opened 2020-06-17
[NT-1359] Allow custom HQ environment in debug tools

justinswart:
# 📲 What

Allows one to enter a custom HQ environment in debug tools.

**Note:** I was unable to test this as the HQ environments appeared to be down at the time.

# 🤔 Why

This has been a requirement from time-to-time so no better time to add it than the present!

# 🛠 How

- Added a `custom` `EnvironmentType` with an associated value which we use to construct the HQ url.
- Added a text prompt when selecting this `EnvironmentType` to capture the value.

# 👀 See

| Prompt | Configured |
| --- | --- |
| image |  |

# ✅ Acceptance criteria

- [ ] Navigate to debug tools, change to the custom environment, enter the name of a valid HQ environment. Requests should be made to that environment.
- [ ] Restarting the app should retain the configured environment.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
@@ -215,6 +215,7 @@ internal final class BetaToolsViewController: UITableViewController {
       self?.viewModel.inputs.setEnvironment(.custom(url))
     }
     customAlertController.addAction(submitAction)
+    customAlertController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
 
     EnvironmentType.allCases.forEach { environment in
       alert.addAction(UIAlertAction(title: environment.description, style: .default) { [weak self] _ in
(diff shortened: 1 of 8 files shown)


Library/EnvironmentType.swift
```diff
@@ -1,17 +1,52 @@
...
+
+  public static var allCases: [EnvironmentType] {
+    return [.custom(nil), .production, .staging, .development, .local]
+  }
+
+  public var description: String {
+    switch self {
+    case let .custom(url):
+      guard let url = url else { return "Custom" }
+      return url
```
jgsamudio, line 41: url ?? "Custom"

Kickstarter-iOS/Views/Controllers/BetaToolsViewController.swift
```diff
@@ -204,9 +204,26 @@ internal final class BetaToolsViewController: UITableViewController {
       sourceView: sourceView
     )
 
+    let customAlertController = UIAlertController(
+      title: "Custom HQ",
+      message: "Enter the name of the HQ environment, e.g. for \"native.dev\" enter \"native\"",
+      preferredStyle: .alert
+    )
+    customAlertController.addTextField()
+    let submitAction = UIAlertAction(title: "Update", style: .default) { [weak self] _ in
```
jgsamudio, line 213: It might be worth adding a cancel button if you want to back out of the custom selection

PR #1230, opened 2020-06-18
[NT-1341 ] Rewards header expand/collapse animation

justinswart:
# 📲 What

Adds an expand/collapse animation to the new _expandable_ rewards header on `PledgeViewController`.

# 🤔 Why

When users first land on this view the header will be collapsed to conserve screen real estate, tapping the green chevron button will expand it (see below).

# 🛠 How

- Restructured the view so that the table view is not pinned to the bottom and won't try to push the views down causing the stack view to expand.
- Instead we "cover" the table view with a masking/container view and maintain a height constraint on that view to have it appear to expand/collapse.

# 👀 See

# ♿️ Accessibility 

# ✅ Acceptance criteria

- [ ] Navigate to the Pledge View, tapping the green chevron button should expand/collapse the header with an animation.

# ⏰ TODO

- [x] Investigate constraint warning.
- [x] Ensure the tappable area is 44x44
- [ ] Consider a11y for this view.


Kickstarter-iOS/Views/Controllers/PledgeExpandableRewardsHeaderViewController.swift
```diff
@@ -94,13 +129,69 @@ final class PledgeExpandableRewardsHeaderViewController: UIViewController {
...
+      }
+
+      if expanded {
+        self.tableViewContainerHeightConstraint?.constant = self.expandedHeight()
+      } else {
+        self.tableViewContainerHeightConstraint?.constant = self.collapsedHeight()
+      }
+
+      UIView.animate(
+        withDuration: 0.33,
```
jgsamudio, line 160: Do we have standardized animations? Unifying our types of animations with consistent durations might be worth it in the long run

justinswart: Not really, I just tweaked the timings here to what felt right in terms of duration and springiness. I'll review this with @colleenmacd to be sure we got it right 👍

PR #1233, opened 2020-06-22
[NT-1329, NT-1330] Bonus Support UI

justinswart:
# 📲 What

Updates `PledgeViewController` to introduce the _Bonus support_ amount.

# 🤔 Why

We are changing the messaging around the amount that backers pledge over and above the reward's minimum. This amount will now be surfaced as _bonus support_.

# 🛠 How

- Updated `PledgeAmountViewController` to behave as is when pledging without a reward and to display the new _Bonus support_ UI when selecting a regular reward.
- Shipping is now positioned above the amount input.
- Updated all of the tests that we had for validation of the amount.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ♿️ Accessibility 

- [X] Supports Dynamic Type

# ✅ Acceptance criteria

- [ ] Pledging without a reward is unchanged from before, minimum amount is `1`.
- [ ] Pledging with a regular reward sets the amount input field to `0` and the total is the reward's minimum + shipping + bonus support.

# ⏰ TODO

- [ ] Add strings for translation


Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -6,12 +6,18 @@ import ReactiveSwift
 
 public typealias PledgeAmountData = (amount: Double, min: Double, max: Double, isValid: Bool)
 
+public typealias PledgeAmountViewConfigData = (
```
jgsamudio, line 9: Curious why not make this a struct instead of a tuple?

Library/Backing+PledgeAmount.swift
```diff
@@ -9,4 +9,13 @@ extension Backing {
...
     )
   }
+
+  /// Returns the bonus support amount subtracting the reward minimum from the total pledge amount
+  public var bonusSupportAmount: Double {
+    guard let reward = self.reward else { return 0 }
+
+    let bonusSupportAmount = Decimal(self.pledgeAmount) - Decimal(reward.minimum)
+
+    return (bonusSupportAmount as NSDecimalNumber).doubleValue
```
jgsamudio, line 19: Don't know if there is historical context to this, but why not return:
`self.pledgeAmount - reward.minimum `

Library/ViewModels/PledgeAmountViewModel.swift
justinswart: We've just usually used tuples unless these get larger than 6 elements. If they're 7+ then one needs a struct for synthesized `Equatable` conformance. This seems somewhat supported by the [docs](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html):

image

Library/Backing+PledgeAmount.swift
justinswart: We just wrap these in `Decimal` when doing addition/subtraction for accuracy given the trouble with floating-point arithmetic. However, in searching for some supporting docs to confirm this, it seems we might not be initializing `Decimal` quite right here. I'll update this!

justinswart: Re slack convo - will update in separate PR.

PR #1234, opened 2020-06-23
[NT-1373] Show Location on DiscoveryPostcardCell for LightsOn Tag

justinswart:
# 📲 What

Shows the location on `DiscoveryPostcardCell` when `DiscoveryParam`'s `tagId` is `lightsOn`, defaults the editorial collection to use `magic` sort.

# 🤔 Why

Previously we were hiding the location when the sort was not `distance`, but we would like to show the location for all sorts and only hide it when the `tagId` is something other than `lightsOn`.

# 🛠 How

Updated the behaviour in `DiscoveryPostcardViewModel` and related tests.

# ✅ Acceptance criteria

- [ ] Location is visible on the the updated `DiscoveryPostcardCell`s in the Lights On editorial collection. 
- [ ] Location is not visible on `DiscoveryPostcardCell`s in other feeds.


justinswart:
> Looks good. When I tried to run the project however, I see the updated project cards. I might be bucketed in that experiment so maybe thats why I didn't see the location.

Yeah looks like location has not been added to the new project cards yet.

PR #1236, opened 2020-06-26
[NT-1330] Fix bonus support behaviour and validation

justinswart:
# 📲 What

Fixes a regression introduced in #1233 that prevented us from backing projects.

# 🤔 Why

#1233 was implemented with the expectation that we would soon use updated versions of the `CreateBacking` and `UpdateBacking` mutations and so adding bonus support was not fully implemented which left it in an incomplete state where, instead of the total pledge amount being passed to the mutation, _only_ the bonus support was.

# 🛠 How

- Made changes to the way in which we handle the total in `PledgeViewModel`.
- Updated a bunch of tests.
- Updated validation rules.

# ✅ Acceptance criteria

- [ ] You should be able to pledge with a bonus support amount.
- [ ] Totals on the Pledge View should add up correctly.
- [ ] No regressions should be introduced in any of the update flows and totals should all add up correctly.


Kickstarter-iOS/Views/Controllers/ManagePledgeViewControllerTests.swift
```diff
@@ -44,11 +44,17 @@ final class ManagePledgeViewControllerTests: TestCase {
       )
       |> \.backing.sequence .~ 10
       |> \.backing.location .~ ManagePledgeViewBackingEnvelope.Backing.Location(name: "United States")
+
+    envelope = envelope
```
justinswart, line 48: Had to split these out in a few places, was getting to complex for the compiler to parse 😓

KsApi/queries/ManagePledgeViewQueries.swift
```diff
@@ -17,6 +17,18 @@ public func managePledgeViewProjectBackingQuery(withBackingId backingId: String)
           .symbol
         ]
       ),
+      .bonusAmount(
```
justinswart, line 20: Added `bonusAmount` and `shippingAmount` to our GraphQL query.

Library/Backing+PledgeAmount.swift
```diff
@@ -9,13 +9,4 @@ extension Backing {
       subtractingShippingAmount: self.shippingAmount.flatMap(Double.init)
     )
   }
-
-  /// Returns the bonus support amount subtracting the reward minimum from the total pledge amount
-  public var bonusSupportAmount: Double {
```
justinswart, line 14: Moved to a shared function.

Library/SharedFunctions.swift
```diff
@@ -225,15 +225,9 @@ internal func sanitizedPledgeParameters(
   pledgeAmount: Double,
   shippingRule: ShippingRule?
 ) -> SanitizedPledgeParams {
-  var pledgeTotal = pledgeAmount
-  var shippingLocationId: String?
+  let shippingLocationId = (shippingRule?.location.id).flatMap(String.init)
 
-  if let shippingRule = shippingRule {
-    pledgeTotal = pledgeAmount.addingCurrency(shippingRule.cost)
```
justinswart, line 232: Previously we added the shipping cost to the total in this function, this has been extracted to happen outside of this function to make it easier to deal with the same total everywhere.

PR #1241, opened 2020-06-30
[NT-1296, NT-1297] Manage Pledge view info bugfixes

justinswart:
# 📲 What

Bugfixes from UAT testing on Manage Pledge View:

- Some summary data was not correctly displayed when viewing the pledge as a creator.
- Add-on rewards were shown twice in the list.

# 🤔 Why

Due to the two different back-ends that we are retrieving data from for this view the structure differs and has to be handled differently depending on the creator/backer context.

# 🛠 How

 I've made some in-line code comments to explain.

# ✅ Acceptance criteria

Create a project with add-ons and back it as a different user.

- [ ] Pledge info is shown correctly as a backer.
- [ ] Pledge info is shown correctly as a creator.
- [ ] There are no duplicate add-ons in the list.

### DIFF
--- a/Library/ViewModels/ManagePledgeViewModel.swift
+++ b/Library/ViewModels/ManagePledgeViewModel.swift
@@ -166,8 +166,9 @@ public final class ManagePledgeViewModel:
 
     self.configurePaymentMethodView = backing.map(managePledgePaymentMethodViewData)
 
-    self.configurePledgeSummaryView = Signal.combineLatest(project, graphBackingEnvelope)
-      .filterMap { project, env in managePledgeSummaryViewData(with: project, envelope: env) }
+    self.configurePledgeSummaryView = Signal.combineLatest(projectAndReward, graphBackingEnvelope)
+      .map(unpack)
+      .filterMap(managePledgeSummaryViewData)
 
     let projectOrBackingFailedToLoad = Signal.merge(
       fetchProjectEvent.map { $0.error as Error? },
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -468,12 +469,9 @@ private func managePledgePaymentMethodViewData(
 
 private func managePledgeSummaryViewData(
   with project: Project,
+  backedReward: Reward,
```
justinswart, line 472: We're now passing in the backed reward instead of trying to retrieve it from the `Project`'s `personalization` property as the latter is unavailable in the creator context.

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -511,8 +509,13 @@ private func rewardsData(
   let dateFormatter = DateFormatter()
   dateFormatter.dateFormat = "yyyy-MM-DD"
 
+  var existingAddOnIds = Set<String>()
```
justinswart, line 512: Maintaining a `Set` of add-on IDs here to avoid duplicates in the list.

Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -173,7 +173,7 @@ private func localizedDescription(project: Project, reward: Reward) -> String {
 }
 
 private func rewardTitle(project: Project, reward: Reward) -> NSAttributedString {
-  guard project.personalization.isBacking == true else {
+  guard project.personalization.isBacking == true || currentUserIsCreator(of: project) else {
```
justinswart, line 176: This `guard` was preventing the reward title from being formatted correctly in the creator context.

PR #1243, opened 2020-06-30
[NT-1366] Add-ons Selection Navigation

justinswart:
# 📲 What

Navigates to `RewardAddOnSelectionViewController` when a `Reward`'s `hasAddOns` flag is `true`.

# 🤔 Why

When a `Reward` has add-ons we will navigate to this VC for the backer to select add-ons.

# 🛠 How

Added the `hasAddOns` property to our v1 `Reward` model which we can use to determine whether or not a `Reward` has add-ons and should navigate to the selection view.

# 👀 See

Related: https://github.com/kickstarter/kickstarter/pull/20224

# ✅ Acceptance criteria

- [ ] Tapping a reward that has add-ons navigates to the empty `RewardAddOnSelectionViewController`.


Kickstarter-iOS/Views/Controllers/RewardsCollectionViewController.swift
```diff
@@ -227,6 +238,16 @@ final class RewardsCollectionViewController: UICollectionViewController {
       ?|> \.isActive .~ !isHidden
   }
 
+  private func goToAddOnSelection(
```
justinswart, line 241: Left these args the same as `goToRewards` for now, they will likely be a bit different but we'll figure that out in the next PR.

PR #1246, opened 2020-07-09
[NT-1367] Add-Ons List

justinswart:
# 📲 What

Retrieves Add-Ons for a project from GraphQL an displays them in `RewardAddOnSelectionViewController`.

**Note:** The add-on cell UI will be in a subsequent PR, for now these cells are rendered as regular `Reward` cells.

# 🤔 Why

Retrieving the add-ons for this interstitial view controller during our checkout flow is required for selecting add-ons for rewards that support them.

# 🛠 How

- Updated `GraphSchema`.
- Added shipping location selector to `RewardAddOnSelectionViewController`.
- Added `RewardAddOnSelectionViewEnvelope` to de-serialize the GraphQL response.
- Added `rewardAddOnSelectionViewAddOnsQuery` to retrieve add-ons for a project slug.
- Added the necessary extension to `Reward` to adapt the data for that model.
- Added tests.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Navigate to a reward that has add-ons, observe that the add-ons are displayed in a list as regular reward cells.

# ⏰ TODO

- [ ] Request error-handling and pull-to-refresh (will be a separate PR).


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -587,6 +587,8 @@ extension PledgeViewController: PledgeShippingLocationViewControllerDelegate {
   ) {
     self.viewModel.inputs.shippingRuleSelected(shippingRule)
   }
+
+  func pledgeShippingLocationViewControllerLayoutDidUpdate(_: PledgeShippingLocationViewController) {}
```
jgsamudio, line 591: Is this part of a protocol? Do we need this function?

KsApi/models/Reward+RewardAddOnSelectionView.swift
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    return Reward(
+      addOnData: addOnData,
+      backersCount: reward.backersCount,
+      convertedMinimum: reward.convertedAmount.amount,
+      description: reward.description,
+      endsAt: reward.endsAt,
+      estimatedDeliveryOn: estimatedDeliveryOn,
+      hasAddOns: false, // This value is only sent via the v1 API to indicate that a base reward has add-ons
```
jgsamudio, line 38: As a reminder that this will need to updated when the V1 API is updated

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
justinswart: Yeah - we conform to this in `PledgeViewController` and now in `RewardAddOnSelectionViewController`. The problem is that in the latter case the shipping location VC is contained in a table view header and so it needs to explicitly recalculate its size when it changes. In the case of `PledgeViewController` we don't need to do this but we're conforming to the same protocol 🤷‍♂️

KsApi/models/Reward+RewardAddOnSelectionView.swift
justinswart: This function is used to 'adapt' the GraphQL Reward data to our v1 `Reward` model which we use throughout the app, so this comment is more to indicate that this is v1 data that we shouldn't expect here. If it is returned from GraphQL then we can remove this.

PR #1251, opened 2020-07-15
[NT-1370] Filter Add-Ons by Location

justinswart:
# 📲 What

Filters the add-ons available for a given base reward by the selected shipping location.

# 🤔 Why

Add-On rewards are just `Reward`s related by their shipping location. Currently we need to handle this filtering at the client.

# 🛠 How

Implemented filtering according to the following rules:

- If the base reward is digital, only return digital add-ons.
- If the base reward has unrestricted shipping only return add-ons that also have unrestricted shipping.
- If the base reward has restricted shipping, return add-ons that also ship to the selected shipping rule's location.

# ✅ Acceptance criteria

- [ ] Navigate to a project that has rewards supporting the above scenarios. Observe that the correct rewards are returned in each of these scenarios.


Kickstarter-iOS/Views/Controllers/RewardAddOnSelectionViewController.swift
```diff
@@ -116,7 +116,6 @@ extension RewardAddOnSelectionViewController: PledgeShippingLocationViewControll
     _: PledgeShippingLocationViewController,
     didSelect shippingRule: ShippingRule
   ) {
-    self.tableView.ksr_sizeHeaderFooterViewsToFit()
```
justinswart, line 119: This wasn't needed here.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -113,3 +121,33 @@ private func rewardsData(
...
+      return addOn.shippingPreference != .noShipping
+    }
+
+    guard let selectedLocationId = shippingRule?.location.id else { return false }
+
+    /**
+     For a base reward that has restricted shipping, only return the add-on
+     if it can ship to the selected shipping location
+     */
+    let addOnShippingLocationIds: Set<Int> = Set(
```
justinswart, line 147: Using a `Set<Int>` just because its `contains` is `O(1)` 🤷‍♂️

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -113,3 +121,38 @@ private func rewardsData(
...
+  filteredBy shippingRule: ShippingRule?,
+  baseReward: Reward
+) -> [RewardAddOnSelectionViewEnvelope.Project.Reward] {
+  return addOns.filter { addOn in
+    // For digital-only base rewards only return add-ons that are also digital-only.
+    if baseReward.shipping.enabled == false {
+      return addOn.shippingPreference == .noShipping
+    }
+
+    // For restricted shipping base rewards, unrestricted shipping or digital-only add-ons are available.
```
iambchan, line 136:
```suggestion
    // For restricted or unrestricted shipping base rewards, unrestricted shipping or digital-only add-ons are available.
```

PR #1257, opened 2020-07-28
[NT-1450] Allow Dissimilar Shipping Cost for Add-Ons

justinswart:
# 📲 What

Allows the selected base reward and its add-ons to have dissimilar shipping costs.

# 🤔 Why

Previously the assumption was made that the shipping cost would be the same across the base reward and all of its add-ons. This is not the case and this PR addresses that.

# 🛠 How

- Update logic in `PledgeViewModel` and `RewardAddOnSelectionViewModel`.
- Display individual add-on shipping cost in `RewardAddOnCardView`.
- Add an optional `shippingRules` array to our `Reward` model.
  - This data is only deserialized from GraphQL.
- Included some fixes to `PledgeAmountViewModel` to ensure pledges are within the max amount.

# ✅ Acceptance criteria

- [ ] When selecting add-ons with shipping, note that the shipping amount displayed is specific to that add-on.
- [ ] Entering a bonus support amount greater than the country maximum should display a message in the pledge view. The allowed amount should be the country maximum minus the total of the rewards + shipping. In the case of US pledged this is $10,000.

### DIFF
--- a/KsApi/models/Reward+ManagePledgeView.swift
+++ b/KsApi/models/Reward+ManagePledgeView.swift
@@ -43,6 +43,7 @@ public extension Reward {
       remaining: backingReward.remainingQuantity,
       rewardsItems: rewardItemsData(from: backingReward, with: project),
       shipping: shippingData(from: backingReward),
+      shippingRules: nil,
       startsAt: backingReward.startsAt,
       title: backingReward.name
     )
--- a/KsApi/models/Reward.swift
+++ b/KsApi/models/Reward.swift
@@ -34,7 +34,8 @@ public struct Reward {
 
   /**
    Returns the closest matching `ShippingRule` for this `Reward` to `otherShippingRule`.
-   If no match is found `otherShippingRule` is returned.
+   If no match is found `otherShippingRule` is returned, this is to be backward-compatible
+   with v1 Rewards that do not include the `shippingRules` array.
    */
   public func shippingRule(matching otherShippingRule: ShippingRule?) -> ShippingRule? {
     return self.shippingRules?
--- a/KsApi/models/Reward.swift
+++ b/KsApi/models/Reward.swift
@@ -10,7 +10,7 @@ public struct AddOnData {
 }
 
 public struct Reward {
-  public var addOnData: AddOnData?
+  public var addOnData: AddOnData? // FIXME: to be removed in future
   public let backersCount: Int?
   public let convertedMinimum: Double
   public let description: String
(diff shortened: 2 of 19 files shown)


KsApi/models/Reward+ManagePledgeView.swift
```diff
@@ -43,6 +43,7 @@ public extension Reward {
       remaining: backingReward.remainingQuantity,
       rewardsItems: rewardItemsData(from: backingReward, with: project),
       shipping: shippingData(from: backingReward),
+      shippingRules: nil,
```
justinswart, line 46: At the moment we don't need this data in the `ManagePledgeViewModel`.

KsApi/models/Reward.swift
```diff
@@ -31,9 +32,19 @@ public struct Reward {
...
   }
 
+  /**
+   Returns the closest matching `ShippingRule` for this `Reward` to `otherShippingRule`.
+   If no match is found `otherShippingRule` is returned.
+   */
+  public func shippingRule(matching otherShippingRule: ShippingRule?) -> ShippingRule? {
+    return self.shippingRules?
+      .first { shippingRule in shippingRule.location.id == otherShippingRule?.location.id }
+      ?? otherShippingRule
```
justinswart, line 42: Does it make sense to coalesce this to `otherShippingRule`? I'm just thinking if there's no match for some reason that it should use the base reward's shipping cost, but I'm not sure if that would ever happen 🤔

KsApi/models/Reward.swift
```diff
@@ -10,7 +10,7 @@ public struct AddOnData {
 }
 
 public struct Reward {
-  public var addOnData: AddOnData?
+  public var addOnData: AddOnData? // FIXME: to be removed in future
```
justinswart, line 13: This will be removed when we refactor this code.

KsApi/models/Reward.swift
jgsamudio: I would leave it as nil in the edge case that a shipping rule is not provided. I'm also wondering how we handle creators that specify free shipping? Will that still be a shipping rule?

justinswart: Yeah, let me update to not coalesce this to the matching rule.

Yes, I confirmed it will be returned as a shipping rule with `0` cost.

justinswart: Update comment to explain that this is to work with legacy v1 `Reward`s.

PR #1260, opened 2020-07-29
[NT-1443] Choose another Reward -> Edit Reward

justinswart:
# 📲 What

Updates the string `Choose another reward` to `Edit reward`. This is seen in the Manage Pledge context menu as well as the rewards carousel title when navigating from the manage pledge context.

**Note:** This string has been added for translation and will be update along with other strings once translations have come back.

**Also note:** Editing the reward for a backing that has add-ons will be supported in an upcoming PR.

# 🤔 Why

We're improving our nomenclature around this with the introduction of add-ons.

# 🛠 How

Updated the strings and added them for translation.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] The updated string is observed in the context menu and as the title for the rewards carousel when editing a reward.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/ManagePledgeViewController.swift
@@ -471,7 +471,7 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
       case .changePaymentMethod:
         title = Strings.Change_payment_method()
       case .chooseAnotherReward:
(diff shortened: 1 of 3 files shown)


justinswart:
> Do we have any screenshot tests for the sheet?

Snapshots use the `UIView` function `snapshotView(afterScreenUpdates:)` I'm not entirely sure if we'd be able to take a snapshot of the activity sheet given that it's a UIKit control outside of the view 🤔

PR #1263, opened 2020-08-06
[NT-1455, NT-1458] Add-Ons GraphQL Refactor

justinswart:
# 📲 What

Performs a fairly large refactor of how we fetch data from GraphQL for add-ons and our manage pledge flows.

# 🤔 Why

We deal with two back-end APIs in the app, one is RESTful and the other GraphQL. Our approach until now has been to construct query-specific data types in order to de-serialize GraphQL responses and to then use those data types in the app. Previously when we only used the RESTful API we had standard object representations (as is the nature of REST) which we could use all over the app. The approach with GraphQL started to become unwieldy in that we would have our regular `Reward` type, then we would have the `Reward` type that was specific to the `MangePledgeViewBackingEnvelope` and then even another `Reward` within the context of `RewardAddOnSelectionViewEnvelope`, etc.

Instead this PR moves all of this to our `KsApi` framework so that the app is unaware of whether the request is made to our v1 API or GraphQL and instead always receives the same object representation. Data adaptation happens before the object is returned to the calling code in the view models. This simplifies our unit tests and makes things somewhat easier to reason about in the app.

# 🛠 How

- Made all `Graph` types `internal` to `KsApi` to ensure that we are not making use of them from the app code.
- Added _adapter-like_ constructors for the GraphQL data structures. For example `Project+GraphProject.swift` contains code that adapts the `GraphProject` data structure to a regular `Project` structure.
- Using our `GraphSchema` DSL I also added `baseQueryProperties` to most of our GraphQL types, these return query properties that are the minimum required in order to construct a given object. For example `Project` requires that certain values not be `nil`, so to simplify always fetching enough data to construct a `Project` we refer to `GraphProject.baseQueryProperties`. These are composable to make it possible to construct larger queries of varying types with their minimum required properties.

# ✅ Acceptance criteria

This a large change and will require regression testing, for the review of this PR I would advise skimming through the changes for anything that looks out of order or doesn't make sense.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/models/graphql/adapters/User+GraphUserTests.swift
```diff
@@ -0,0 +1,11 @@
+
```
jgsamudio, line 1: Remove extra heading space

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -254,82 +262,63 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
-      context: datum.context,
-      shippingRule: datum.shippingRule
-    )
-  }
-  .map(RewardAddOnSelectionDataSourceItem.rewardAddOn)
+private func unpack(
+  rewardsData: ([Reward], Project, Reward, PledgeViewContext, ShippingRule?),
+  selectedQuantities: SelectedRewardQuantities
+) -> ([Reward], Project, Reward, PledgeViewContext, ShippingRule?, SelectedRewardQuantities) {
+  return (rewardsData.0, rewardsData.1, rewardsData.2, rewardsData.3, rewardsData.4, selectedQuantities)
```
jgsamudio, line 269: Should we consider a model if there are 4 params in this tuple?

justinswart: In this case this function `unpack` just serves the purpose of taking a bunch of combines signals and returning it in the structure that we need it, so the call-site is neater:

```swift
self.loadAddOnRewardsIntoDataSource = rewardAddOnCardsViewData
  .takePairWhen(latestSelectedQuantities)
  .map(unpack) // this function performs the transformation
  .map(rewardsData) // this one returns `[RewardAddOnSelectionDataSourceItem]` for the data source
```

PR #1264, opened 2020-08-20
[NT-1455] Fix Add-Ons Filtering

justinswart:
# 📲 What

Fixes a small regression that @Arkariang noticed earlier where digital add-ons weren't being displayed alongside shipping add-ons.

# 🤔 Why

This was a regression introduced from the refactor in #1263.

# 🛠 How

Made sure to carry the shipping preference through from the GraphQL model.

# ✅ Acceptance criteria

- [ ] When changing the selected shipping location during add-ons selection digital-only add-ons should be available alongside add-ons that are have unrestricted shipping or match on location.


justinswart:
Going to merge this as it's a small change and don't want it to trip up UAT unnecessarily 😄

PR #1268, opened 2020-08-25
[NT-1372, NT-1459] Update Backing with Add-ons

justinswart:
# 📲 What

- Fixes a memory leak causing `PledgeViewController` to never be released.
- Carries the selected reward quantities from the user's `Backing` through the Edit Reward flow.
- Allows backers to complete the flow updating their pledge with different add-ons or a different reward.

# 🤔 Why

We're allowing backers the flexibility to update rewards and change their add-on selection.

# 🛠 How

- We had a signal observer that was strongly referencing `self` causing a retain-cycle in `PledgeViewController`, updated this to weakly reference `self`.
- Updated `RewardAddOnSelectionViewModel` to carry the current selected quantities through. Also moved the loading indicator to the CTA button for this view as it is less jarring this way.
- Updated `PledgeViewModel` to correctly update the backing with add-ons.
- Made various changes to code that determines reward availability, consolidated this into shared functions and added tests.
  - This now takes into account the current backed quantity and includes that in the remaining quantity that's displayed.

# ✅ Acceptance criteria

All of the edit reward flows should work:

- [ ] Back a reward with add-ons, change the payment source.
- [ ] Back a reward with add-ons, update the bonus support amount.
- [ ] Back a reward with add-ons, edit your reward and change the shipping location.
- [ ] Back a reward with add-ons, edit your reward and change the add-on selection.
- [ ] Back a reward with add-ons, edit your reward and change the add-on selection and bonus support amount.


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -396,8 +396,8 @@ final class PledgeViewController: UIViewController,
 
     self.viewModel.outputs.rootStackViewLayoutMargins
       .observeForUI()
-      .observeValues { margins in
-        self.rootStackView.layoutMargins = margins
+      .observeValues { [weak self] margins in
```
justinswart, line 399: Source of the retain cycle.

Kickstarter-iOS/Views/PledgeViewCTAContainerView.swift
```diff
@@ -230,17 +230,6 @@ private let rootStackViewStyle: StackViewStyle = { stackView in
     |> \.spacing .~ Styles.grid(1)
 }
 
-private let disclaimerLabelStyle: LabelStyle = { label in
```
justinswart, line 233: Unused.

Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
jgsamudio: Ahh good old weak self

PR #1269, opened 2020-08-25
[NT-1416] Edit Add-Ons Reward Alert

justinswart:
# 📲 What

Displays a prompt when editing a reward that previously had add-ons selected.

# 🤔 Why

Selecting a new reward might affect the add-ons that a backer selected before and we would like to inform them of this during the edit flow.

# 🛠 How

Updated `RewardsCollectionViewModel` to display a prompt when selecting a reward under these circumstances:
- Reward has add-ons, project is not backed, navigates to add-on selection without prompt.
- Reward has add-ons, project is backed with add-ons, triggers prompt before add-on selection.
- Reward does not have add-ons, project is not backed, navigates to pledge without prompt.
- Reward does not have add-ons, project is backed with add-ons, triggers prompt before pledge.

# 👀 See

image

# ✅ Acceptance criteria

- [ ] Back a reward with add-ons, edit the reward, select the same reward - no prompt is shown.
- [ ] Back a reward with add-ons, edit the reward, select a new reward - prompt is shown.
- [ ] Back a reward without add-ons, edit the reward, select a new reward - no prompt is shown.
- [ ] Back a reward without add-ons, edit the reward, should not be able to select the same reward.

### DIFF
--- a/Library/ViewModels/RewardsCollectionViewModel.swift
+++ b/Library/ViewModels/RewardsCollectionViewModel.swift
@@ -285,7 +285,7 @@ private func shouldTriggerEditRewardPrompt(_ data: PledgeViewData) -> Bool {
 
   let rewardChanged = data.rewards.first?.id != backing.reward?.id
 
-  // We show the prompt if they have previously backed with add-ons and they selecting a new reward.
+  // We show the prompt if they have previously backed with add-ons and they are selecting a new reward.
   return backing.addOns?.isEmpty == false && rewardChanged
 }
 
--- a/Library/ViewModels/RewardsCollectionViewModel.swift
+++ b/Library/ViewModels/RewardsCollectionViewModel.swift
@@ -10,6 +10,7 @@ public enum RewardsCollectionViewContext {
 
 public protocol RewardsCollectionViewModelInputs {
   func configure(with project: Project, refTag: RefTag?, context: RewardsCollectionViewContext)
+  func confirmedEditReward()
   func rewardCellShouldShowDividerLine(_ show: Bool)
   func rewardSelected(with rewardId: Int)
   func traitCollectionDidChange(_ traitCollection: UITraitCollection)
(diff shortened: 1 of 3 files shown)


Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -215,6 +276,19 @@ private func shouldNavigateToReward(project: Project, reward: Reward, refTag _:
...
 }
 
+private func shouldTriggerEditRewardPrompt(_ data: PledgeViewData) -> Bool {
+  // If the user is not backing the project then there is no need to show the prompt.
+  guard
+    userIsBackingProject(data.project),
+    let backing = data.project.personalization.backing
+  else { return false }
+
+  let rewardUnchanged = data.rewards.first?.id == backing.reward?.id
```
jgsamudio, line 286: Will the first reward id always be equal to the backing.reward.id?

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -215,6 +276,19 @@ private func shouldNavigateToReward(project: Project, reward: Reward, refTag _:
...
+
+  // We show the prompt if they have previously backed with add-ons and they selecting a new reward.
+  return backing.addOns?.isEmpty == false && !rewardUnchanged
```
jgsamudio, line 289: It took me a second to see what this was doing, consider:
`let rewardChanged = data.rewards.first?.id != backing.reward?.id`

and then
```suggestion
  return !backing.addOns?.isEmpty && rewardChanged
```

Library/ViewModels/RewardsCollectionViewModel.swift
justinswart: Yeah for now I know that it's the first one but this could probably be improved to be more explicit about what the base reward is in the `PledgeViewData` type. Let's aim to improve this once we get this release out 👍

Library/ViewModels/RewardsCollectionViewModel.swift
justinswart: Ah thanks that's clearer, I still need to do the `== false` because the boolean is optional.

PR #1273, opened 2020-08-28
[NT-1481] Remove Amount Shipping Locations List

justinswart:
# 📲 What

Removes the amount from the list of shipping locations.

# 🤔 Why

This might be confusing during add-on selection and the amount doesn't serve much purpose as it stands because a backer is unlikely to select a different shipping location based on the cost to ship there.

# 🛠 How

Removed the amount from the label title.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Back a reward that has shipping, select a new shipping location and observe that no amount is shown.


Library/ViewModels/ShippingRuleCellViewModel.swift
```diff
@@ -28,8 +28,7 @@ public final class ShippingRuleCellViewModel: ShippingRuleCellViewModelType,
       .map { $0.selectedShippingRule == $0.shippingRule }
 
     self.textLabelText = data
-      .map { ($0.project, $0.shippingRule) }
-      .map(formattedValue(_:shippingRule:))
+      .map(\.shippingRule.location.localizedName)
```
jgsamudio, line 31: Just to make sure, we don't want to show the shipping cost for any reward / add-on?

justinswart: Yeah that's the decision that we reached. We figured backers are unlikely to ship to a different location based on cost, they would just ship to where they live.

PR #1274, opened 2020-08-28
[NT-1416] Navigation bugfix

justinswart:
# 📲 What

Fixes a bug which would sometimes caused navigation to not work correctly when a backer edits their reward and selected a new reward.

# 🤔 Why

This bug was caused by the behaviour of `Signal.zip`. `zip` emits when all of its signals have emitted at least once and then emits each time all of the signals have emitted again. In the situation where the user dismissed the alert by cancelling the last selected reward ID would remain in the buffer and the following time that they tapped _yes_ to continue it would incorrectly navigate to the wrong reward.

# 🛠 How

Rearranged the signals to make use of `takeWhen` instead of `zip`.

# 👀 See

# ✅ Acceptance criteria

The ACs from #1269:

- [ ] Back a reward with add-ons, edit the reward, select the same reward - no prompt is shown.
- [ ] Back a reward with add-ons, edit the reward, select a new reward - prompt is shown.
- [ ] Back a reward without add-ons, edit the reward, select a new reward - no prompt is shown.
- [ ] Back a reward without add-ons, edit the reward, should not be able to select the same reward.

Also:

- [ ] When shown the alert, select _no_ to cancel and then select a different reward. You should not be shown the previously selected reward.

### DIFF
--- a/Library/ViewModels/RewardsCollectionViewModel.swift
+++ b/Library/ViewModels/RewardsCollectionViewModel.swift
@@ -145,17 +145,15 @@ public final class RewardsCollectionViewModel: RewardsCollectionViewModelType,
       )
     }
 
-    let goToAddOnSelectionBackedConfirmed = Signal.zip(
-      goToAddOnSelectionBackedWithAddOns,
-      self.confirmedEditRewardProperty.signal
-    )
-    .map(first)
+    let goToAddOnSelectionBackedConfirmed = goToPledge
+      .takeWhen(self.confirmedEditRewardProperty.signal)
+      .filter(second >>> isTrue)
+      .map(first)
 
-    let goToPledgeBackedConfirmed = Signal.zip(
-      goToPledgeBackedWithAddOns,
-      self.confirmedEditRewardProperty.signal
-    )
-    .map(first)
+    let goToPledgeBackedConfirmed = goToPledge
+      .takeWhen(self.confirmedEditRewardProperty.signal)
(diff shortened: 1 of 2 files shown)


justinswart:
@jgsamudio I added the navigation fix for when the creator of the project is viewing the rewards carousel (that it should not navigate to the pledge view or add-ons selection).

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -273,7 +271,9 @@ private func titleForContext(_ context: RewardsCollectionViewContext, project: P
 }
 
 private func shouldNavigateToReward(project: Project, reward: Reward, refTag _: RefTag?) -> Bool {
-  project.state == .live && (!userIsBacking(reward: reward, inProject: project) || reward.hasAddOns)
+  guard !currentUserIsCreator(of: project) else { return false }
```
jgsamudio, line 274: Do we have a test for this guard? Might be worth adding another test to check for this?

justinswart: Yeah that's the one below? `testGoToPledge_DoesNotEmitIfCreatorOfProject` or was there another case you wanted to test for?

jgsamudio: Ah, I missed the test for the creator, thank you 👍

PR #1294, opened 2020-09-17
[NT-1533] Filter Add-Ons From List When Unavailable

justinswart:
# 📲 What

- Filters add-ons from the list for selection when they are unavailable either by way of a timebased or quantity limit. Add-ons are always shown during editing when previously backed.
- Adds a small fix for haptics on the stepper.

# 🤔 Why

- During UAT we found that we were able to select add-ons for backing even though they might be unavailable. This resulted in unexpected behaviour.
- Haptics were being triggered on cell configuration but should only emit when adjusting the stepper.

# 🛠 How

- Added availability checking and tests.
- Updated the logic for haptics and updated the test.

# ✅ Acceptance criteria

- [ ] When backing a project that had unavailable add-ons, those add-ons should not appear in the list.
- [ ] When editing a backing that contains add-ons that have since become unavailable, those add-ons should appear in the list for editing.
- [ ] Haptics should only emit when adjusting the stepper, not when the cell is reconfigured while scrolling up and down the list.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/PledgeShippingLocationViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PledgeShippingLocationViewController.swift
@@ -119,7 +119,7 @@ final class PledgeShippingLocationViewController: UIViewController {
       .observeForUI()
       .observeValues { [weak self] isHidden in
         self?.shimmerLoadingView.amountPlaceholder.alpha = isHidden ? 0 : 1
-    }
+      }
 
     /**
      When any layout updates occur we need to notify the delegate. This is only necessary when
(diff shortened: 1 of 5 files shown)


justinswart:
> I tested with this [project](https://staging.kickstarter.com/projects/microcosmpublishing/how-to-be-accountable-take-responsibility-change-behavior?ref=section-homepage-view-more-recommendations-p1) and I still see the add-on that should've been sold out. Maybe the backend didn't update the values if it is sold out?

I'll take a look!

justinswart:
As far as I can tell it was filtering out the add-on with `remainingQuantity` of `0` correctly 👍

PR #1295, opened 2020-09-18
Fix User Settings Regression

justinswart:
# 📲 What

Fixes a regression in our user account settings which caused changing user settings to error.

# 🤔 Why

Why indeed. The refactor in #1263 introduced some changes that added a non-nil property `uid` to `GraphUser`. The various queries we have that fetch data for our user settings views make use of `GraphUser` but those queries unfortunately hadn't been updated to include the `uid` property resulting in `GraphUser` in those contexts failing to deserialize.

# 🛠 How

Added `GraphUser.baseQueryProperties` to `UserQueries`.

# ✅ Acceptance criteria

- [ ] All user settings views should be available and functioning.

### DIFF
--- a/KsApi/queries/UserQueries.swift
+++ b/KsApi/queries/UserQueries.swift
@@ -21,66 +21,69 @@ public enum UserQueries: Queryable {
 }
 
 public func accountQueryFields() -> NonEmptySet<Query.User> {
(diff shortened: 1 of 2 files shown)


jgsamudio (approved):
Thanks for fixing this!

PR #1296, opened 2020-09-18
Reduce CircleCI config verbosity

justinswart:
# 📲 What

Reduces our CircleCI config by ~200 lines by grouping jobs that are similar with a few variables that change.

# 🤔 Why

@yatriks and I started improving the verbosity of this config in #1143 and figured it would be good to reduce it even further, it makes the file easier to understand and maintain.

# 🛠 How

Identified jobs that were similar except for a couple of variable parts which we turned into environment variables.

# 👀 See

# ✅ Acceptance criteria

- [ ] Config is valid.
- [ ] Jobs pass on Circle.
- [ ] We're able to successfully run on of the distribute jobs like alpha/beta/itunes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,36 +124,6 @@ jobs:
       - TARGET=Library
       - SCHEME=Library-iOS
 
-  # Kickstarter UI Tests
-  kickstarter-ui-tests:
-    <<: *base_job
-    environment:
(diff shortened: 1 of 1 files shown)


yatriks (approved):
👍

singhhari (approved):
lgtm

PR #1297, opened 2020-09-21
[NT-1535] Add vendor/bundle and .bundle/config to .gitignore

singhhari:
# 📲 What
Updating .gitignore to include vendor/bundle and .bundle/config. 

# 🤔 Why
While setting up my local environment I required some additional set up for my gems and bundle installation.

# 🛠 How
Updated .gitignore directly.


justinswart (approved):
lgtm! 🎉

PR #1298, opened 2020-09-21
[NT-1529] Add PLN Currency Selection

justinswart:
# 📲 What

Adds the Polish Złoty (PLN) to our currency selection.

# 🤔 Why

Poland is a country that we will soon be rolling out currency support to.

# 🛠 How

- Added translations for the currency string.
- Updated our `Currency` type which makes the currency available for selection by way of its `CaseIterable` conformance.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Navigate to your user settings, update your currency to PLN, this change should be persisted. **Note:** This work is still pending on the back-end and this PR is blocked for now.

PR #1299, opened 2020-09-21
[NT-1524] Update backed reward copy. Add tests for RewardsCollectionViewControllerTests

singhhari:
# 📲 What

Updated the copy on the Edit Rewards flow to replace "You backed" with "Your selection".

# 🤔 Why

Android and Web are displaying "Your selection".

# 🛠 How

@justinswart updated the string for internationalization prior to the replacement. In addition, new snapshots were made for the view controller containing the Reward Cell.

# 👀 See

Simulator Screen Shot - iPhone 11 Pro - 2020-09-21 at 18 05 25

# ✅ Acceptance criteria

- [ ] When a user go through the edit pledge flow, the copy is updated from "You Backed" to "Your selection"


justinswart (approved):
Look good! I think we may still want to tweak codecov a little more, it looks like it is including our view controller tests in its coverage which it shouldn't 🤔 But we can deal with that separately. In #1298 I added a configuration for it to run in _informational_ mode so that it doesn't block PRs for now, until we have it configured properly and know how we really want to use it.

vickisiolos (commented):
Looks good to me!

PR #1300, opened 2020-09-21
[NT-1527] Add New Country Support

justinswart:
# 📲 What

Adds support for Greece, Slovenia and Poland.

# 🤔 Why

We'll soon be supporting projects in these countries.

# 🛠 How

- Updated our `Project.Country` struct to include these countries with mappings to `EUR` and `PLN` currencies respectively.
- Updated formatting tests.

# ✅ Acceptance criteria

- [ ] With these countries enabled on staging and test projects created, we should be able to back those projects and see the correct currency for each new country.


KsApi/models/Project.Country.swift
```diff
@@ -1,3 +1,6 @@
+private let eurMaxPledge: Int = 8_500
```
justinswart, line 1: Moved these to local vars seeing as EUR is used for several countries in the list.

KsApi/models/Project.Country.swift
```diff
@@ -9,41 +12,43 @@ extension Project {
 
     /*
      FIXME: The amount for maximum pledge can be found here:
-     https://github.com/kickstarter/kickstarter/blob/master/config/countries.yml
+     https://github.com/kickstarter/kickstarter/blob/master/config/currencies.yml
```
justinswart, line 15: This URL changed for the currency min/max.

singhhari (approved):
lgtm

PR #1301, opened 2020-09-22
[NT-542] Replace UIButton on CancelPledgeViewController with LoadingButton

singhhari:
# 📲 What

We are displaying a loading spinner within the Cancel pledge button while the API call is processing. 

# 🤔 Why

This behavior is currently implemented on the `Pledge` and `Confirm` buttons.

# 🛠 How

We replace the `UIButton` with a `LoadingButton` within the `CancelPledgeViewController`, provide functionality within the view model to observe the cancellation event and appropriately render either the buttons title or loading spinner.

# ✅ Acceptance criteria

- [x] A loading spinner is shown on Cancel pledge buttons while the API call is processing


justinswart (approved):
Nice!

PR #1303, opened 2020-09-23
[NT-1499] Xcode 12/iOS 14.0 Compatibility

justinswart:
# 📲 What

Updates our codebase for compatibility with Xcode 12.0 and builds the app against the iOS 14.0 SDK.

# 🤔 Why

Always good to keep things up to date, Xcode 12.0, Swift 5.3 and iOS 14.0 all bring improvements that we would like to have access to in our workflow.

# 🛠 How

- Addressed some broken code as as result of changes in Swift 5.3, in particular we were affected by the Swift Evolution proposal regarding forward-scan trailing closures.
  - The proposal: [SE-0286](https://github.com/apple/swift-evolution/blob/master/proposals/0286-forward-scan-trailing-closures.md).
  - Related [issue](https://github.com/apple/swift-evolution/blob/master/proposals/0286-forward-scan-trailing-closures.md) opened on ReactiveSwift.
  - Bug reported on Swift Forum: [SR-13453](https://bugs.swift.org/browse/SR-13453).
- Implemented a Carthage workaround until a fix is available, described [here](https://github.com/Carthage/Carthage/issues/3019#issuecomment-693381253). This is required for our dependencies to build currently.
- Re-recorded all snapshots. Unfortunately this is usually necessary when Xcode and iOS versions change due to tiny pixel changes in rendering.

# ✅ Acceptance criteria

- [ ] App passes full regression testing.


Library/ViewModels/ChangeEmailViewModel.swift
```diff
@@ -101,7 +101,10 @@ public final class ChangeEmailViewModel: ChangeEmailViewModelType, ChangeEmailVi
     let isEmailVerified = userEmailEvent.values().map { $0.me.isEmailVerified }.skipNil()
     let isEmailDeliverable = userEmailEvent.values().map { $0.me.isDeliverable }.skipNil()
     let emailVerifiedAndDeliverable = Signal.combineLatest(isEmailVerified, isEmailDeliverable)
-      .map { $0 && $1 }
+      .map { isEmailVerified, isEmailDeliverable -> Bool in
+        let r = isEmailVerified && isEmailDeliverable
+        return r
```
justinswart, line 106: These changes are kind've unfortunate and seem unnecessary but I wasn't able to get this code to compile without introducing some of these local scope vars.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -95,7 +95,7 @@ public final class CommentsViewModel: CommentsViewModelType, CommentsViewModelIn
       self.viewDidLoadProperty.signal,
       self.userSessionStartedProperty.signal
     )
-    .map { AppEnvironment.current.currentUser }
+    .map { _ in AppEnvironment.current.currentUser }
```
justinswart, line 98: This works around the Swift function overload bug.

Library/ViewModels/RootViewModel.swift
```diff
@@ -159,21 +159,27 @@ public final class RootViewModel: RootViewModelType, RootViewModelInputs, RootVi
...
       .map { ($0 != nil, ($0?.stats.memberProjectsCount ?? 0) > 0) }
       .skipRepeats(==)
 
-    let standardViewControllers = self.viewDidLoadProperty.signal.map { generateStandardViewControllers() }
-    let personalizedViewControllers = userState.map { generatePersonalizedViewControllers(userState: $0) }
+    let standardViewControllers = self.viewDidLoadProperty.signal.map { _ -> [RootViewControllerData] in
+      generateStandardViewControllers()
+    }
+    let personalizedViewControllers = userState.map { userState -> [RootViewControllerData] in
+      generatePersonalizedViewControllers(userState: (userState.isMember, userState.isLoggedIn))
```
justinswart, line 172: This was necessary to work around something that was deprecated in Swift called tuple shuffling: https://noahgilmore.com/blog/tuple-shuffling/

singhhari (approved):
lgtm!

PR #1304, opened 2020-09-24
[NT-1241] Added update link to third-party app sharing payload

andrewKstrt:
# 📲 What

Added update url to **UpdateActivityItemProvider** item payload for all types but **.copyToPasteboard** and **.postToFacebook** 

So now app shares  by default

- Update.sequence
- Update.title 
- Update.urls.web.update 

For **.message** and **.postToTwitter** title and update link is shared

For Facebook and pasteboard only update link is shared

# 🤔 Why

Both update title and update link should be shared with IM apps

# 🛠 How

**UpdateActivityItemProvider** data object string payload is constructed with both sequence, title, update link.

# ✅ Acceptance criteria

- [ ] Share payload for Whatsapp, Twitter, iMessage, Slack should have title and update link.


justinswart (approved):
Nice, thanks!

PR #1306, opened 2020-09-24
[NT-1543][NT-1544] Update Pledge and Change Payment Method stackview behavior updates

singhhari:
# 📲 What

On the "Update Pledge" screen, when the user has pledged for no reward, the "Pledge" line item will be hidden. Similarly, on the "Change Payment Method" screen, when the user has pledged for no reward, the "Pledge" line item will be hidden.

# 🤔 Why

When a user pledges for no reward, there was previously redundancy in the way we present the pledge amount.

# 🛠 How

Added some logic to a couple of our view models that determine whether there is a reward before determining the set up of their respective view controllers.


justinswart (approved):
lgtm!

PR #1308, opened 2020-09-28
[NT-1495] Update logic for quantity attributed text in RewardCardView

singhhari:
# 📲 What

When specifying a quantity of add-ons, we use `# x add-on name` on both the Pledge screen and the View/Manage Pledge screen. On the Pledge screen, we hide `# x` when # is > 1, but we don’t do that on the View/Manage screen

# 🤔 Why

Consistency within view controllers that display similar information.

# 🛠 How

All of the work is done in the `RewardCardViewModel`. We're doing a check on the quantity of the add-on which was updated.


justinswart (changes requested):
Ok, looks good - can we add a test to be sure? 🙂

justinswart (approved):

PR #1309, opened 2020-09-28
[NT-1559] Reward Availability Bugfixes

justinswart:
# 📲 What

Fixes two bugs with regards to the availability of rewards:

1. If a reward is both timebased and limited we were allowing users to back rewards if either of these conditions passed, for example someone could back a timebased reward, that was also limited, that had expired but had a remaining positive quantity.
2. We were allowing users to tap through on reward cards even when unavailable under certain conditions.

# 🤔 Why

Users should not be able to back rewards in these situations and we also did not have sufficient back-end fail-safes to prevent this.

# 🛠 How

1. Fixed the logic that determines the availability of rewards.
2. Shared the rewards carousel navigation filtering behaviour.
3. Updated tests.

# ✅ Acceptance criteria

- [ ] A reward that is both timebased and limited should not be backable if either the timelimit is reached or the remaining quantity drops to zero.

### DIFF
--- a/Library/ViewModels/RewardCardContainerViewModelTests.swift
+++ b/Library/ViewModels/RewardCardContainerViewModelTests.swift
@@ -20,6 +20,7 @@ final class RewardCardContainerViewModelTests: TestCase {
   let availableLimitedReward = Reward.postcards
     |> Reward.lens.limit .~ 100
     |> Reward.lens.remaining .~ 25
+    |> Reward.lens.endsAt .~ nil
   let availableTimebasedReward = Reward.postcards
     |> Reward.lens.limit .~ nil
     |> Reward.lens.remaining .~ nil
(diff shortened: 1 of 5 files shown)


Library/ViewModels/RewardCardContainerViewModelTests.swift
```diff
@@ -168,6 +182,7 @@ final class RewardCardContainerViewModelTests: TestCase {
       "No longer available",
       "No longer available",
       "No longer available",
+      "No longer available",
```
justinswart, line 185: These tests are a little hard to read - I should probably refactor them sometime, the intention was to try to cover a bunch of different project states and reward configurations but it would probably be easier to understand if we just broke them up.

Nevertheless, this is the test that asserts the change we made - we added the `availableLimitedUnavailableTimebasedReward` and in the arrays of true/false and strings in this test we see that the second last value that we added has the expected result.

singhhari (approved):
lgtm

PR #1310, opened 2020-09-29
Default UITableViewCell labels are used to show estimated delivery date

andrewKstrt:
# 📲 What

**PledgeExpandableHeaderRewardHeaderCell** cell style set to **subtitle** to show title and estimated delivery date with default **UITableViewCell** title and subtitle labels to avoid nested **UIStackView** layout issues

**amountLabel** is added as a subview to contentView

# 🤔 Why

UIKit default behavior for **UIStackView**

    - Apple automatically creates 1000 priority constraints for you when you add managed subviews to a stack view.
    - Apple automatically creates a 0-height constraint for you when you hide a subview of a stack view.

This creates layout issue for the cell content. Like clipping of the estimated delivery label.
Or plasing the hole content of the stack view much lover relative to parent than it should be.

[here](https://stackoverflow.com/questions/32428210/uistackview-unable-to-simultaneously-satisfy-constraints-on-squished-hidden/38064687#38064687) you can find more derails.

With wrapper **UIView** approach i was able to resolve issue with label clipping. But issue with Y axis  misplacement on view scroll of the entire **rootStackView** persisted.

So i've decided to use default cell properties. It fixed the issue with layout. And reduce amount of code in **PledgeExpandableHeaderRewardHeaderCell**

# ✅ Acceptance criteria

- [ ] Navigate to back this project screen
- [ ] Scroll up and down

### DIFF
--- a/Kickstarter-iOS/Views/Cells/PledgeExpandableHeaderRewardHeaderCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeExpandableHeaderRewardHeaderCell.swift
@@ -20,6 +20,7 @@ final class PledgeExpandableHeaderRewardHeaderCell: UITableViewCell, ValueCell {
     super.init(style: style, reuseIdentifier: reuseIdentifier)
 
     self.configureViews()
+    self.setupConstraints()
     self.bindViewModel()
   }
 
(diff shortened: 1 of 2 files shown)


justinswart:
@andrewKstrt interesting 🤔I wonder what is causing this clipping to begin with. I see that these changes seem to fix it most of the time, although I think I was still able to reproduce it. However, adding this constraint does also appear to introduce a constraint warning when the view loads which hopefully we can solve.

I wonder also whether we could simply reload the table view at the point where the layout is problematic?

justinswart:
Closing in favour of #1333.

PR #1311, opened 2020-09-30
[NT-1566] Fix Supported Countries List

justinswart:
# 📲 What

Adds a change that was meant to be in #1300 in which we added new country support.

# 🤔 Why

The array updated in this change was meant to be updated before but we weren't aware of it. This is necessary for project country min and max pledge values to used when pledging.

# 🛠 How

For some reason this array was maintained separately from `Project.Country.all`, for simplicity I've set them to be the same list, that is - any country that we add to `Project.Country` is supported. If at a later stage it is necessary for us to have countries in `Project.Country` that aren't supported then we can simply filter those out in the launched countries list.

# ✅ Acceptance criteria

- [ ] Navigate to a project in Poland, the minimum pledge value for _No Reward_ should be zl 5.
- [ ] Navigate to a project in Mexico, the minimum pledge value for _No Reward_ should be MX$10.


singhhari (approved):
lgtm

PR #1313, opened 2020-10-01
[NT-1565] Backers Can Edit an Expired Reward with Add-Ons

justinswart:
# 📲 What

Allows backers to edit an expired reward with add-ons.

# 🤔 Why

Fixes a bug from #1309. Rewards should be editable if they've expired and have add-ons.

# 🛠 How

Fixed the logic for this and added tests.

# 👀 See

# ✅ Acceptance criteria

- [ ] Back a reward with add-ons that will expire soon. Once it has expired you should be able to edit it.
- [ ] There should be no regressions to other reward editing behaviour (see tests).

### DIFF
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -350,3 +350,17 @@ public func selectedRewardQuantities(in backing: Backing) -> SelectedRewardQuant
 
   return quantities
 }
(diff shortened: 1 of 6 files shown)


singhhari (approved):
lgtm

PR #1315, opened 2020-10-02
[NT-1557] Updating pill view colors in RewardCardView

singhhari:
# 📲 What

Reward cards and add-ons cards share a pill component that is used to denote a variety of stats —number of backers, shipping locations, remaining quantity, and time left. During the add-ons work, we changed the color of the 'remaining quantity' and 'time left' pills that are shown on the add-ons card to better align with the web. We should follow up and make those changes on the reward card pills for consistency.

# 🤔 Why

There should be parity between the views of the app that display the same data within the pills.

# 🛠 How

Added a new model to the RewardCardViewModel and replaced the value being emitted by the appropriate signal to assist in setting up the text and colors. On the reward cards, pills that denote limited quantity (X left of Y) or limited time (Ends in X days) have background color KDS-celebrate-100 with text color dark-grey-500

# 👀 See

Simulator Screen Shot - Clone 1 of iPhone 8 - 2020-10-02 at 11 16 29


justinswart (approved):
Looks good! 👏

PR #1316, opened 2020-10-06
[NT-1575] Argo Decodable Function

andrewKstrt:
# 📲 What

First step to remove Argo dependency

# 🤔 Why

With **Swift.Codable** introduction there is no need to carry around **Argo** lib for JSON parsing 

[RFC](https://github.com/kickstarter/rfcs/blob/master/rfcs/0069-native-deprecate-argo.md)

# 🛠 How

**tryDecodable** function

Extended test coverage for tryDecodable function

- Added tests for primitive data types
- Added tests for missing key and key type mismatch

**Update.UrlsEnvelope** and **Update.UrlsEnvelope.WebEnvelope** converted to **Swift.Codable** 

- added test cases for missing key and key type mismatch

# ✅ Acceptance criteria

- [ ] Pass **UpdateTests**
- [ ] Pass **TryDecodableTests**

# ⏰ TODO

- [ ] Get an idea of how regression testing is going to take place

### DIFF
--- a/KsApi/models/Update.swift
+++ b/KsApi/models/Update.swift
@@ -49,7 +49,7 @@ extension Update: Argo.Decodable {
       <*> json <| "sequence"
       <*> (json <| "title" <|> .success(""))
     return tmp3
-      <*> json <| "urls"
+      <*> ((json <| "urls" >>- tryDecodable) as Decoded<Update.UrlsEnvelope>)
       <*> json <|? "user"
       <*> json <|? "visible"
   }
@@ -68,3 +68,7 @@ extension Update.UrlsEnvelope.WebEnvelope: Argo.Decodable {
       <^> json <| "update"
   }
 }
+
+extension Update.UrlsEnvelope.WebEnvelope: Swift.Codable {}
+
+extension Update.UrlsEnvelope: Swift.Codable {}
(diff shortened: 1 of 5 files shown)


KsApi/models/Update.swift
```diff
@@ -49,7 +49,7 @@ extension Update: Argo.Decodable {
       <*> json <| "sequence"
       <*> (json <| "title" <|> .success(""))
     return tmp3
-      <*> json <| "urls"
+      <*> ((json <| "urls" >>- tryDecodable) as Decoded<Update.UrlsEnvelope>)
```
singhhari, line 52: Should this logic in this line go in here or directly in the decode method of the `Update.UrlsEnvelope` extension? Not quite sure.

KsApi/models/UpdateTests.swift
```diff
@@ -70,4 +70,40 @@ internal final class UpdateTests: XCTestCase {
       update.value?.urls.web.update
     )
   }
+
+  func testJSONDecoding_WithBadUrlsWebData() {
```
singhhari, line 74: I think since you added two new methods here you can use a name like `testJSONDecoding_WithBadUrls_WebData` and the one below can be `testJSONDecoding_WithBadUrls_WrongKeyData`

singhhari (approved):
I left a couple of small comments but looks good so far.

KsApi/models/Update.swift
andrewKstrt: I would say it should stay here since eventually Argo will be removed and i don't see reasons for `Update.UrlsEnvelope` to know about Swift - to - Argo conversion process

PR #1317, opened 2020-10-07
[NT-1584][NT-1588] Handling rewards and add ons with future start dates

singhhari:
# 📲 What

Currently, backers are able to see rewards and add-ons with future start dates and back them. We want to conditionally render these options for selection only when the appropriate time has been reached.

# 🤔 Why

Currently, rewards and add-ons with start dates in the future are being shown as available for backing. These should be hidden from view until the start date set by the creator.

# 🛠 How

We're filtering in the `RewardAddOnSelectionViewController` and `RewardCollectionViewController` based on a date check.

### DIFF
--- a/Library/ViewModels/RewardAddOnSelectionViewModel.swift
+++ b/Library/ViewModels/RewardAddOnSelectionViewModel.swift
@@ -378,14 +378,14 @@ private func addOnIsAvailable(_ addOn: Reward, in project: Project) -> Bool {
   }
 
   let isUnlimitedOrAvailable = addOn.limit == nil || addOn.remaining ?? 0 > 0
-  let hasNoTimeLimitOrHasNotEnded = (
-    addOn.endsAt == nil ||
-      (addOn.endsAt ?? 0) >= AppEnvironment.current.dateType.init().timeIntervalSince1970
-  )
 
+  // Assuming the user has not backed the addOn, we only display if it's within range of the start and end date
+  let hasNoTimeLimitOrIsWithinRange = isStartDateBeforeToday(for: addOn) &&
+    (addOn.endsAt == nil || (addOn.endsAt ?? 0) >= AppEnvironment.current.dateType.init()
+      .timeIntervalSince1970)
   return [
     project.state == .live,
-    hasNoTimeLimitOrHasNotEnded,
+    hasNoTimeLimitOrIsWithinRange,
     isUnlimitedOrAvailable
   ]
   .allSatisfy(isTrue)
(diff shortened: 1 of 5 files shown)


Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -378,14 +378,14 @@ private func addOnIsAvailable(_ addOn: Reward, in project: Project) -> Bool {
...
 
   let isUnlimitedOrAvailable = addOn.limit == nil || addOn.remaining ?? 0 > 0
-  let hasNoTimeLimitOrHasNotEnded = (
-    addOn.endsAt == nil ||
-      (addOn.endsAt ?? 0) >= AppEnvironment.current.dateType.init().timeIntervalSince1970
-  )
 
+  // Assuming the user has not backed the addOn, we only display if it's within range of the start and end date
+  let hasNoTimeLimitOrIsWithinRange = isStartDateBeforeToday(for: addOn) &&
+    (addOn.endsAt == nil || (addOn.endsAt ?? 0) >= AppEnvironment.current.dateType.init()
```
andrewKstrt, line 384: Should we put this line into separate function?

andrewKstrt, line 384: Should we decouple nil checks and date range check? To make it easier to read.
Also we could create a range 

```
let startDate = Date(timeIntervalSince1970: addOn.startsAt)
let endDate = Date(timeIntervalSince1970: addOn.endsAt)
let range = startDate...endDate

return range.contains(AppEnvironment.current.dateType.init().date)
```

andrewKstrt: and seems we don't even need to convert to `Date`, we should be able to use range from `TimeInterval`

singhhari: I created a new method in `SharedFunctions` and wrote some tests for it.

singhhari: I like the idea and syntax of using the range but `RewardsCollectionViewModel` and the `RewardAddOnSelectionViewModel` have slightly different product definitions which is why I created two methods for checking the start and end dates. 

In this case, the `RewardsCollectionViewController` is supposed to hide rewards with a `startsAt` prior to the current date but should show rewards with `endsAt` before today. The `RewardAddOnSelectionViewController`, on the other hand, should be hiding AddOns in both cases (so I use both functions). Let me know if you have another suggestion or optimization.

andrewKstrt (approved):

PR #1321, opened 2020-10-22
[NT-1586] Fix shipping costs on Add On selection

singhhari:
# 📲 What

After selecting a reward and being directed to the add on selection screen, incorrect shipping costs were being shown. We determined the issue was caused by the wrong field being fetched and used in our GraphQL query. This PR updates the query to utilize `shippingRulesExpanded` and updates existing schema to decode `shippingRulesExpanded` for use.

# 🤔 Why

Some creators noted the shipping costs were being calculated incorrectly on the pledge screen which suggested an issue with the way we were using shipping rules. After taking a look it appeared that were using the wrong set of shipping rules being returned from our query.

# 🛠 How

Made some updates to the relevant query by adding a parameter for location id, This was necessary, otherwise we would be pulling hundreds of shipping locations per add on. Moreover, we're decoding our new field and using it for filtering add ons.

# ✅ Acceptance criteria

Steps to test this feature:
- Visit project with slug: `dungeon-drop-dropped-too-deep-and-tavern-tales` (or any project that has international shipping on production)
- Add a Reward with an AddOn like `Game Bundle` in the case of the example above.
- Add an add-on like `Limited Promo Pack (2020) mini-expansion` ($0 shipping) and pay attention to the price the creator has set for shipping.
- Change shipping selector to Canada (or any country other than US)
- See add-ons shipping cost on `Limited Promo Pack (2020) mini-expansion`. 

Expected: shipping cost should be $0 (based on what the creator set)


andrewKstrt (commented):
I wonder what exactly the difference between `ShippingRules` and `ShippingRulesExpanded`. What do we miss in `ShippingRules`?

andrewKstrt:
Also can you please add some Acceptance Criteria and steps to reproduce the issue. 
Code change looks good

andrewKstrt (approved):

singhhari:
@andrewKstrt  `shippingRules` unfortunately had an incorrect calculation for shipping costs which was corrected in `shippingRulesExpanded`. The other major difference is that `shippingRulesExpanded` has all possible shipping locations in the response and we filter by location to specifically get the one the user requests.

PR #1322, opened 2020-10-22
NT-1631: Blank Screen on Discover

jgsamudio:
# 📲 What

There have been reports that user's are seeing blank screens on discover. We've tracked down this issue to an issue with Optimizely. The fix was to reduce the amount of time the app will wait for Optimizely to initialize. After the fix we were still noticing issues with discover.

# 🤔 Why

User's are unable to access the discover page to view projects. Reproducing this was difficult until a user mentioned that their NextDNS VPN was causing it to not work. Testing on a physical device indicated that it was indeed causing blank screens on iOS. 

# 🛠 How

Another way to test for this is by setting the optimizely key to an empty string. When this happens the sdk is unable to download the datafile and thus cannot configure the sdk. The fix is to call the failed to configure sdk signal when this happens and also to enable nil sdk values in the discover view model. 

# Test
- Run branch on device
- Visit nextdns.io on device
- Download the NextDNS app
- Follow the instructions to setup a vpn
- Ensure that the discover screen does not show a blank screen

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -356,6 +356,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       print("🔴 Optimizely SDK Configuration Failed with Error: \(optimizelyError.localizedDescription)")
 
       Crashlytics.crashlytics().record(error: optimizelyError)
+      self.viewModel.inputs.optimizelyClientConfigurationFailed()
     }
   }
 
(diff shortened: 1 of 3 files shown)


singhhari (approved):
lgtm

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/DiscoveryViewModelTests.swift | indent |


</p>

PR #1326, opened 2020-11-03
[NT-1644][NT-1645] Adding email verification feature flags

singhhari:
# 📲 What

Before we begin the heavy lifting for the email verification work, we'll need to put the work behind feature flags to prevent incomplete work or bugs from being exposed in any future releases. We've added two feature flags to the project after setting them up on Flipper.

# 🤔 Why

Feature flags are a commonly used paradigm in software development that allow you to modify system behavior without changing the code explicitly. The email verification feature work is split into smaller pieces that will be incrementally developed and released as they are completed. The feature flags we've added here will ensure that any new releases prior to the completion of the feature will not introduce unexpected behavior for our current users.

# 🛠 How

The following two flags were added to the project:
- `ios_email_verification_flow`
- `ios_email_verification_skip`

# 👀 See

Simulator Screen Shot - iPhone 11 Pro - 2020-11-03 at 11 46 28

### DIFF
--- a/Library/Feature.swift
+++ b/Library/Feature.swift
@@ -2,12 +2,16 @@ import Foundation
 
 public enum Feature: String {
   case qualtrics = "ios_qualtrics"
+  case emailVerificationFlow = "ios_email_verification_flow"
+  case emailVerificationSkip = "ios_email_verification_skip"
 }
 
 extension Feature: CustomStringConvertible {
   public var description: String {
     switch self {
     case .qualtrics: return "Qualtrics"
+    case .emailVerificationFlow: return "Email Verification Flow"
+    case .emailVerificationSkip: return "Email Verification Skip"
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Library/Feature.swift
```diff
@@ -2,12 +2,16 @@ import Foundation
...
+  case emailVerificationFlow = "ios_email_verification_flow"
+  case emailVerificationSkip = "ios_email_verification_skip"
 }
 
 extension Feature: CustomStringConvertible {
   public var description: String {
     switch self {
     case .qualtrics: return "Qualtrics"
+    case .emailVerificationFlow: return "Email Verification Flow"
+    case .emailVerificationSkip: return "Email Verification Skip"
```
andrewKstrt, line 14: we don't need to localize this text, right?

andrewKstrt (approved):

Library/Feature.swift
singhhari: I don't see the other existing feature flag in the project localized so I think this is fine.

PR #1330, opened 2020-11-13
[NT-1663] Detect if users email is verified

singhhari:
# 📲 What

The User serializer on the backend is now providing a property `is_email_verified` that we need to use for our e-mail verification work. This PR is decoding the property into the `User` struct and subsequent work will require use of the property for conditionally rendering the associated UI.

# 🤔 Why

This is the first step needed for verifying a user. If the property we are now decoding returns true we won't need to present the interstitial screen.

# 🛠 How

We are leveraging Swifts `Decodable` protocol to create our objects now so a majority of the work is setting up our `User` struct with a new property we can plug into our logic later.


singhhari:
Coverage is slightly down due to `UserLenses` file. Not quite sure if I can improve coverage in there?

justinswart:
> Coverage is slightly down due to `UserLenses` file. Not quite sure if I can improve coverage in there?

Yeah that's ok we don't really have coverage of the lenses directly.

justinswart (approved):
Nice!

PR #1327, opened 2020-11-04
Feature argo removal

andrewKstrt:
# 📲 What

Argo removed from project

# 🤔 Why

With Swift.Codable we don't need Argo anymore

# 🛠 How

Swift.Codable protocol adoption for the models was implemented

# ⏰ TODO

- [ ] Push Notification Testing
- [ ] DeepLink Testing

PR #1331, opened 2020-11-16
[NT-1676] Argo Removal Clean-up

ncreated:
# 📲 What

Hello 👋. Following the "Getting Started" lead me to having build issues (`Xcode 12.0 (12A7209)`). I fixed it on my end by doing following, which you may consider integrating on your end:
* listing `thoughtbot/Runes` dependency in `Cartfile`,
* cleaning up references to `Argo` dependency removed in #1327

# 🤔 Why

`thoughtbot/Runes` is a required dependency, but `make bootstrap` didn't lead to installing it because it was not listed in `Cartfile`. As `make bootstrap` calls `carthage update --platform iOS`, it should be there IMHO.

`Argo` framework was still referenced in many places, so it was breaking the build (e.g. when processing `xcfilelist`).


justinswart:
hey @ncreated thanks for flagging. It does seem like there were a couple things missed from the removal of the Argo dependency in #1323. I'll push a couple more commits to this branch and we can get this merged 👍

Cartfile.resolved
```diff
@@ -1,6 +1,6 @@
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" "6.32.0"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseCrashlyticsBinary.json" "6.32.0"
-github "Alamofire/Alamofire" "5.2.2"
+github "Alamofire/Alamofire" "5.4.0"
```
ncreated, line 3: @justinswart indeed, I've been considering pushing this, but I wasn't sure if this bump doesn't break anything 👌.

justinswart: I think this is fine, we only use AlamofireImage for some simple image caching - so this is included by way of that dependency.

singhhari (approved):
lgtm!

justinswart (approved):
Will merge once tests pass 👍

PR #1332, opened 2020-11-16
[NT-1661][NT-1662] Connect email verification UI to login/signup flows

singhhari:
# 📲 What

Before dismissing the login or sign up view controllers we need to push an interstitial UI for email verification onto the navigation stack. The work is behind a feature flag and given the user successfully logs in (within the context of login or account creation) we should show the email verification UI before the tab bar is rendered.

# 🤔 Why

This PR continues the work for our email verification flow. It represents the connection between our `User` and UI/UX interactions.

# 🛠 How

Login and account creation rely on `NotificationCenter` to dismiss the view controllers after the user has successfully logged in. In between login and view controller dismissal, the email verification work is plugged in. Specifically in the `LoginViewController` and `SignUpViewController`.

# ⏰ TODO

- [ ] Will add tests


Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -132,8 +132,16 @@ internal final class LoginViewController: UIViewController {
       }
 
     self.viewModel.outputs.logIntoEnvironment
+      .observeForControllerAction()
```
singhhari, line 135: I added this bit because the compiler was throwing a fatal error about the main thread. I wasn't sure if `observeForUI()` was the right method to plug into here and chose `observeForControllerAction()` after looking at the same paradigm elsewhere in the view controller.

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -166,6 +174,14 @@ internal final class LoginViewController: UIViewController {
     self.navigationController?.pushViewController(vc, animated: true)
   }
 
+  fileprivate func startEmailVerificationViewController() {
+    /**
+     FIXME: `UIViewController` needs to be replaced with the EmailVerification UI when development is complete.
```
singhhari, line 179: Is leaving this comment in until the view controller UI is completed OK or is there another preferred way to handle it?

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -166,6 +174,14 @@ internal final class LoginViewController: UIViewController {
     self.navigationController?.pushViewController(vc, animated: true)
   }
 
+  fileprivate func startEmailVerificationViewController() {
```
singhhari, line 177: I declared this method in both view controllers and want to find a way to only declare this in a single place. I wasn't sure if there was a shared file I could put this? I thought about `SharedFunctions`?

Kickstarter-iOS/Views/Controllers/SignupViewController.swift
```diff
@@ -120,8 +120,16 @@ internal final class SignupViewController: UIViewController, MFMailComposeViewCo
 
     self.viewModel.outputs.logIntoEnvironment
       .observeValues { [weak self] in
```
singhhari, line 122: This didn't throw an error about the main thread despite not having `observeForControllerAction()` or `observeForUI()` so I left it as it is. Should I add one of those in for good measure?

Kickstarter-iOS/Library/SharedFunctions.swift
```diff
@@ -47,3 +47,15 @@ public func logoutAndDismiss(
 
   viewController.dismiss(animated: true, completion: nil)
 }
+
+// MARK: - Email Verification workflow
+
+public func startEmailVerificationViewController(
```
justinswart, line 53: I'm guessing this is more likely to be presented modally so we might name this
```swift
public func presentEmailVerificationViewController(on viewController: UIViewController) { ... }
```
Also a little confusing but there are two `SharedFunctions.swift` one in the `Library` framework and one in `Kickstarter_Framework`. I think this is better suited to the latter because it'll be called from VC's and not view models.

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
justinswart: Yeah whenever we're presenting or pushing we use `observeForControllerAction()` to be sure. I think if you click through on those functions there should be a comment about why they're there.

Kickstarter-iOS/Views/Controllers/SignupViewController.swift
justinswart: We would use it here too, it's possible that this signal emits on the main thread but we want to always be sure that it does so we should just use `observeForControllerAction()` to not take any chances.

Kickstarter-iOS/Views/Controllers/LoginViewController.swift
```diff
@@ -132,9 +132,19 @@ internal final class LoginViewController: UIViewController {
       }
 
     self.viewModel.outputs.logIntoEnvironment
+      .observeForControllerAction()
       .observeValues { [weak self] env in
+        guard let self = self else { return }
         AppEnvironment.login(env)
-        self?.viewModel.inputs.environmentLoggedIn()
+
+        if featureEmailVerificationFlowIsEnabled() {
```
justinswart, line 140: Any control flow logic like this should as far as possible live in the view model so that we can test it.

Kickstarter-iOS/Views/Controllers/SignupViewController.swift
```diff
@@ -120,8 +120,16 @@ internal final class SignupViewController: UIViewController, MFMailComposeViewCo
...
+        guard let self = self else { return }
         AppEnvironment.login($0)
-        self?.viewModel.inputs.environmentLoggedIn()
+        if featureEmailVerificationFlowIsEnabled() {
```
justinswart, line 125: 👉 view model for test coverage.

Kickstarter-iOS/Library/SharedFunctions.swift
singhhari: According to the designs, it looks like a push (back arrow on top left of view controller). The only thing I find weird is the interaction with going "back" after logging in or signing up...

justinswart: oh, interesting - i think we should just confirm that behaviour (when it will be pushed onto the stack and when it will be modal)

singhhari: brought this up in the email-verification channel for confirmation/discussion

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/LoginViewModelTests.swift | indent |
| Library/ViewModels/SignupViewModelTests.swift | indent |


</p>

PR #1333, opened 2020-11-16
[NT-1494] Fix for Pledge View Reward Title Being Cut Off

justinswart:
# 📲 What

Fixes an issue that was causing the reward title on the pledge view to sometimes be clipped.

# 🤔 Why

This is not the expected behaviour and hampers the experience.

# 🛠 How

Set the property `insetsLayoutMarginsFromSafeArea` to `false` for both of the cell's root stackviews in this table view. Sometimes when the cell is scrolled beyond the safe area of the view and performs a layout pass it does so incorrectly insetting its layout margins from the safe area. Setting this property to `false` prevents this from happening. I was not able to reproduce the issue after setting this to `false` but could consistently do so with it being `true` (default).

# 👀 See

https://developer.apple.com/documentation/uikit/uiview/2891101-insetslayoutmarginsfromsafearea

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] Navigate to the pledge view for a reward that has shipping. While the shipping locations are loading try to scroll the reward title out of view. This should _not_ cause the text to be cut off.

### DIFF
--- a/Kickstarter-iOS/Views/Cells/PledgeExpandableHeaderRewardCell.swift
+++ b/Kickstarter-iOS/Views/Cells/PledgeExpandableHeaderRewardCell.swift
@@ -95,6 +95,7 @@ private func rootStackViewStyle(_ isAccessibilityCategory: Bool) -> (StackViewSt
 
   return { (stackView: UIStackView) in
     stackView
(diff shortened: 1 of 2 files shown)


singhhari (approved):
lgtm

PR #1329, opened 2020-11-13
[NT-1657] Email verification interstitial UI

singhhari:
# 📲 What

Incomplete UI for the email verification work

PR #1334, opened 2020-11-18
[NT-1657] Add EmailVerificationViewController UI

justinswart:
# 📲 What

Adds the UI for `EmailVerificationViewController`.

# 🤔 Why

We will be displaying this prompt to users that have logged in but not yet verified the email address that they use on Kickstarter.

# 🛠 How

\* This work was started in #1329 but in that PR storyboards were used, we prefer programmatic UI. This PR replaces that one.

- Added the UI and empty VM.
- Added snapshot tests.

# 👀 See

image

# ✅ Acceptance criteria

- [ ] UI matches designs in Figma.

# ⏰ TODO

- [ ] Add strings for translation.


Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
```diff
@@ -0,0 +1,182 @@
...
+// MARK: - Styles
+
+private let skipButtonStyle: ButtonStyle = { (button: UIButton) in
+  button
+    |> UIButton.lens.titleLabel.textAlignment .~ .center
+    |> UIButton.lens.titleLabel.font .~ UIFont.ksr_subhead().bolded
+    |> UIButton.lens.titleColor(for: .normal) .~ .ksr_green_500
+    |> UIButton.lens.title(for: .normal) %~ { _ in
+      localizedString(
+        key: "Ill_do_this_later",
```
singhhari, line 102: Should we be pulling this and the other new strings from the server? I'm not sure what the process is to set that up.

singhhari (approved):
lgtm

Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
justinswart: Yeah I added them here for you to review: https://github.com/kickstarter/kickstarter/pull/20912

Once that merges they'll be translated and then we will re-generate our strings with `make strings`.

I've just put them in manually for now so that the UI renders correctly. I'm using the `localizedString` function so that localization still works correctly once they come in and these strings are easy to find to update them to their `Strings.*` counterpart.

PR #1337, opened 2020-11-20
[NT-1658] Re-send Email Verification

justinswart:
# 📲 What

Allows users to re-send the verification email from `EmailVerificationViewController`

# 🤔 Why

Users may have lost their verification email or its link might have expired. This allows them to recover from such a scenario.

# 🛠 How

Wires up the resend button and performs the necessary GraphQL mutation.

# ✅ Acceptance criteria

- [ ] From the `EmailVerificationViewController` tap _resend email_. A success banner should be shown.
- [ ] Repeat the above causing a network error using Charles proxy or otherwise, an error banner should be down.


Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
```diff
@@ -117,6 +127,13 @@ final class EmailVerificationViewController: UIViewController {
   internal override func bindViewModel() {
     super.bindViewModel()
 
+    self.viewModel.outputs.activityIndicatorIsHidden
+      .observeForUI()
+      .observeValues { [weak self] hidden in
```
emrcftci, line 132: `isHidden` may be better name than `hidden` because it is a Boolean.

Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
```diff
@@ -117,6 +127,13 @@ final class EmailVerificationViewController: UIViewController {
...
+        // set alpha instead of isHidden to avoid stackview bouncing when this is shown.
+        self?.activityIndicatorView.alpha = hidden ? 0 : 1
```
emrcftci, line 134: 🤔 
```swift
public extension Bool {

  var integer: Int {
    return self ? 1 : 0
  }
}
```
then
```swift
self?.activityIndicatorView.alpha = isHidden.integer
```

singhhari (approved):
lgtm!

PR #1339, opened 2020-11-26
[NT-1674] Email Verification Deeplink

justinswart:
# 📲 What

Adds a deeplink for `/profile/verify_email` which will open the app and subsequently be accessed in order to either verify the email address or cause a verification email to be re-sent.

**Note:** Draft PR as there are some outstanding questions to resolve, see TODOs below.

# 🤔 Why

This is a continuation of our work to add email verification to the app. One might call this the meat an potatoes of it all.

# 🛠 How

- Added the deeplink to `Navigation`.
- Added a request and response handling flow to `AppDelegateViewModel`. This is separated from our existing API networking code because it is an ad-hoc response for this purpose.
- Displaying a banner on `DiscoverViewController` with the success/failure message.

# ✅ Acceptance criteria

- [ ] Tap on an email verification link in an email, it should open the app.
- [ ] If the token has expired, a new verification email should be sent.
- [ ] If successful the banner should be displayed and indicated that.
- [ ] If unsuccessful the banner should be displayed and indicated that.

# ⏰ TODO

- [x] Dismiss `EmailVerificationViewController` if currently displayed.
- [x] Handle 403 specifically?
- [x] Add more copy for translation or will back-end send success and re-sent messages.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -234,6 +234,24 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
       }
 
+    self.viewModel.outputs.verifyEmailWithURLRequest
+      .observeForUI()
+      .observeValues { [weak self] in self?.verifyEmail($0) }
+
+    self.viewModel.outputs.emailVerificationCompleted
+      .observeForUI()
+      .observeValues { [weak self] message, success in
+        // FIXME: We may want to revisit this shortcut to find the VC to present on.
```
justinswart, line 244: This isn't ideal but it was becoming quite unwieldy to do this through a whole lot of message-passing. I think we need to consider when and how this might be displayed. It's possible a notification may serve us better here 🤔

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -234,6 +234,24 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        guard
+          let firstVC = self?.rootTabBarController?.viewControllers?.first as? UINavigationController,
+          let topVC = firstVC.topViewController,
+          let vc = topVC as? MessageBannerViewControllerPresenting
+        else { return }
+
+        // FIXME: Consider emitting the success/error type from the output.
+        vc.messageBannerViewController?.showBanner(with: success ? .success : .error, message: message)
```
justinswart, line 252: We could emit `success` or `error` here too but the `Bool` seemed right to be at the time 🤷‍♂️

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1126,3 +1153,27 @@ private func shouldGoToLandingPage() -> Bool {
...
+  guard let statusCode = urlResponse?.statusCode else { return nil }
+
+  switch statusCode {
+  case 200..<300:
+    let response = localizedString(
+      key: "Thanks_youve_successfully_verified_your_email_address",
+      defaultValue: "Thanks—you’ve successfully verified your email address."
+    )
+    return (response, true)
+  case 300..<Int.max:
```
justinswart, line 1168: Might need to explicitly handle `403` here if we're displaying a message for an email being re-sent.

Kickstarter-iOS/AppDelegate.swift
singhhari: Could we use `self.window` here? Not sure if that is better or worse that the existing implementation though.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
singhhari: Wouldn't the message being returned by the server handle all errors whether it's a `500` or specifically a `403` since it is a failure and also captured within the `300..<Int.max`?

Kickstarter-iOS/Views/Controllers/DiscoveryViewController.swift
```diff
@@ -3,12 +3,13 @@ import Library
 import Prelude
 import UIKit
 
-internal final class DiscoveryViewController: UIViewController {
+internal final class DiscoveryViewController: UIViewController, MessageBannerViewControllerPresenting {
```
singhhari, line 6: Wondering if we need to this for all the view controllers in the tab bar? Consider the situation where the user is in `SearchViewController` and then decides to go into their email and verify. They would be navigated back into the app but the banner won't show. Definitely a bit edge-casey.

singhhari (changes requested):
A couple of broad questions about implementation.

Kickstarter-iOS/AppDelegate.swift
singhhari: A similar approach to what I was originally thinking which was to use `self.window`. I think your approach is effectively the same.

Kickstarter-iOS/AppDelegate.swift
singhhari: Seems fine to me. Not sure what the trade off is for using a view model output?

justinswart: Nah this should be fine.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
justinswart: Yeah that seems right, so shouldn't need to handle it here.

Kickstarter-iOS/Views/Controllers/DiscoveryViewController.swift
justinswart: Going to look into this a bit more 👍

Kickstarter-iOS/AppDelegate.swift
justinswart: We might want to look into that for dismissing the presented VC.

justinswart: It actually works fine to display the banner on the tab bar controller so i've updated to do that.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

singhhari (approved):

PR #1340, opened 2020-11-30
[NT-1679] Re-send Verification Email on View Loading

justinswart:
# 📲 What

Requests that the verification email is re-sent upon `EmailVerificationViewController`'s view loading.

# 🤔 Why

When users are shown this view we want to ensure that they have a fresh verification email in their inbox.

# 🛠 How

Wired up the `viewDidLoad` signal to the same code as the re-send button and added in a boolean to decide whether or not to show the success message.

# ✅ Acceptance criteria

- [ ] Log-in as a user that is unverified and is shown the `EmailVerificationViewController`. Check your inbox, a verification email should have been re-sent.
- [ ] No success banner or activity indicator is shown.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
@@ -143,10 +143,10 @@ final class EmailVerificationViewController: UIViewController, MessageBannerView
         self.delegate?.emailVerificationViewControllerDidComplete(self)
       }
 
-    self.viewModel.outputs.showSuccessBannerWithMessageAndShouldShow
+    self.viewModel.outputs.showSuccessBannerWithMessageAndShowBanner
       .observeForUI()
-      .observeValues { [weak self] message, shouldShow in
-        guard shouldShow else { return }
+      .observeValues { [weak self] message, showBanner in
+        guard showBanner else { return }
         self?.messageBannerViewController?.showBanner(with: .success, message: message)
       }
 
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
```diff
@@ -143,9 +143,10 @@ final class EmailVerificationViewController: UIViewController, MessageBannerView
         self.delegate?.emailVerificationViewControllerDidComplete(self)
       }
 
-    self.viewModel.outputs.showSuccessBannerWithMessage
+    self.viewModel.outputs.showSuccessBannerWithMessageAndShouldShow
       .observeForUI()
-      .observeValues { [weak self] message in
+      .observeValues { [weak self] message, shouldShow in
```
singhhari, line 148: Not sure how strict we are about Swift syntax and the bool variable prefix `is` but this is readable and I don't mind keeping it as is.

singhhari (approved):
lgtm

Kickstarter-iOS/Views/Controllers/EmailVerificationViewController.swift
justinswart: Hmm yeah good point - _should_ is probably not great, although _is_ makes more sense for properties like `isEnabled`. Maybe this should just be `showBanner`?

singhhari: Yep good call. I think `showBanner` is a more descriptive.

PR #1345, opened 2020-12-03
[NT-1692] "Mark as received" causes View Pledge to error

singhhari:
# 📲 What

There is a decoding error within our `Backing` model object that is causing an error in our pledge view. We are currently expecting a `Bool` to be returned for our `backer_completed_at` key but the server is actually providing an `Int`. This PR includes a bit of logic to map these different types.

# 🤔 Why

This bug is preventing backers from viewing their pledge/reward. The footer will show an error and a Retry button because of the decoding failure. 
The expected behavior is a button that says `View Your Pledge` and once the user selects it they can view the details of their pledge/reward.

# 🛠 How

In our `Backing` object, we included a line to map the value of an existing timestamp to a boolean.

### DIFF
--- a/KsApi/models/Backing.swift
+++ b/KsApi/models/Backing.swift
@@ -75,8 +75,7 @@ extension Backing: Decodable {
     self.amount = try values.decode(Double.self, forKey: .amount)
     self.backer = try values.decodeIfPresent(User.self, forKey: .backer)
     self.backerId = try values.decode(Int.self, forKey: .backerId)
-    self.backerCompleted = try values
-      .decodeIfPresent(Int.self, forKey: .backerCompleted) != nil ? true : false
+    self.backerCompleted = try values.decodeIfPresent(Int.self, forKey: .backerCompleted) != nil
     self.bonusAmount = try values.decodeIfPresent(Double.self, forKey: .bonusAmount) ?? 0.0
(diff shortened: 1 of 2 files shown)


KsApi/models/Backing.swift
```diff
@@ -75,7 +75,8 @@ extension Backing: Decodable {
     self.amount = try values.decode(Double.self, forKey: .amount)
     self.backer = try values.decodeIfPresent(User.self, forKey: .backer)
     self.backerId = try values.decode(Int.self, forKey: .backerId)
-    self.backerCompleted = try values.decodeIfPresent(Bool.self, forKey: .backerCompleted)
+    self.backerCompleted = try values
+      .decodeIfPresent(Int.self, forKey: .backerCompleted) != nil ? true : false
```
jgsamudio, line 79: Should we check if it equals 1 instead of nil?

justinswart: I think the existence of a value means `true` in this case but there's no need for the ternary this can just be:

```swift
self.backerCompleted = try values.decodeIfPresent(Int.self, forKey: .backerCompleted) != nil
```

justinswart (changes requested):
Cool this should fix this - can we add a test to `BackingTests`?

justinswart (approved):

PR #1348, opened 2020-12-07
[NT-1683] Updating app icons

singhhari:
# 📲 What

Updating icons for KickBeta and KickAlpha.

# 🤔 Why

This is part of a larger initiative to update the color schemes we are using on the Kickstarter app. Prior to this we were using the same icon for KickBeta and KickAlpha; this PR will update both.

# 🛠 How

- Added new asset for the icons on KickAlpha.
- Updated existing icon set for KickBeta.
- Update settings on Kickstarter-iOS target.

# 👀 See

icon_76pt@1x
icon_76pt@1x


justinswart (approved):
Cool, looks good!

PR #1354, opened 2020-12-11
[NT-1696] Deleting email verification presentation logic/unused helper methods.

singhhari:
# 📲 What

Some new product specifications have come up for our e-mail verification work. Specifically, we're no longer presenting the interstitial screen on login or signup. This PR removes the logic in place for observing when a user is not verified and restores the previous flow from login/signup -> tab bar controller. The `EmailVerificationViewController/ViewModel` is still in the project and will be reused later.

# 🤔 Why

After Optimizely provided some additional insight about the existing email verification flow on the web, we're adjusting Native to account for the conclusive takeaways of the experiment.

# 🛠 How

- Deleted some components of the view model and updated the respective tests.
- No longer observing said view model components in the view controllers for Login or Signup.


justinswart (commented):
Cool, I think this ended up being easier to remove because we used the separate outputs. I'm thinking we should also remove the conformances to `EmailVerificationViewControllerDelegate` and those inputs on the VMs where we had implemented that. We'll reuse parts of that UI but I don't think in those places.

justinswart (approved):
Cool I think this is gtg 👍

PR #1355, opened 2020-12-14
[NT-1705] Updating SDKs for AppCenter, Facebook and Optimizely.

singhhari:
# 📲 What

We're taking some time to update some of our Carthage dependencies. This PR includes version updates in our `Cartfile` and `Cartfile.resolved` for the following SDKs:
- AppCenter 4.0.0
- Facebook 8.2.0
- Optimizely 3.6.1

# 🤔 Why

In general, we should be using updated versions for these SDKs as they are typically under continuous development. The version numbers we've chosen are stable releases and only represent a few of the dependencies we use on Kickstarter. 

# 🛠 How

We have a script: `bin/carthage.sh` that we use to handle updates or bootstrapping our dependencies. In the case of this PR, the `Cartfiles` are semantically updated and the script does the heavy lifting.


justinswart (approved):
Good to merge if everything passes on CI 👍

PR #1356, opened 2020-12-15
[SD-345] MessageThread and ProjectStatsEnvelope Decoding Bugfixes

justinswart:
# 📲 What

Fixes two decoding bugs that were found in `MessageThread` and `ProjectStatsEnvelope`.

# 🤔 Why

These are regressions that were introduced in #1323.

# 🛠 How

- `MessageThread` was expecting an array type for the incorrect key. Updated the key and added a test.
- `ProjectStatsEnvelope` needed a `CodingKeys` `enum` in order to decode the snake_cased `more_activities` key.

# ✅ Acceptance criteria

- [ ] Navigate to messages and tap into a message - the view should load.
- [ ] As a creator, navigate to the dashboard and tap into activities. The view should load.


singhhari (approved):
lgtm

PR #1360, opened 2021-01-09
Add Xcode 12.3 Fixes

DmitryPR:
# 📲 What

Update the project configuration for the latest Xcode.

# 🤔 Why

Without this change the tests do not run.

# 🛠 How

Use the VALIDATE_WORKSPACE build setting.

# 👀 See

https://stackoverflow.com/questions/63267897/building-for-ios-simulator-but-the-linked-framework-framework-was-built


justinswart:
Hi @DmitryPR, I'm not sure what this PR addresses?

DmitryPR:
@justinswart without this fix I was not able to launch the tests in the project after pulling it from master. I tired both from make file by running `make test-all` and from Xcode. After searching the stack overflow I encountered with the solution that it had to do with Xcode workspace validation.

justinswart:
Ok, I'll take a look at `make test-all` - it appears though that none of the tests are running on our CI for this branch since this change so while it may have fixed the issue locally for you it has introduced a regression in our CI builds.

DmitryPR:
@justinswart Thank you, I understand that something is wrong with the tests running now on the CI, let me see if I can find another solution also.

justinswart:
Thanks for raising this PR - I've addressed this in #1361. You were on the right track but it seems that the property only needs to be configured, it doesn't need to be set to `YES`.

PR #1361, opened 2021-01-13
[NT-1725] Fix Tests And Only Cache Frameworks on CI

justinswart:
# 📲 What

- Updates our `bin/carthage.sh` script so that when run locally arguments will be passed through and our `Carthage` directory won't be cached.
- Fixes the issue that the PR #1360 raised.

# 🤔 Why

- When running Carthage builds locally we typically need this behavior and it's inconvenient to have to edit the script each time.
- It turns out that Xcode 12.3 requires that our project includes the `VALIDATE_WORKSPACE` value in our project configuration, but this value does not need to be `YES` as it was set in #1360. See: https://stackoverflow.com/a/65354762/2271759

# 🛠 How

Made the changes described above.

# ✅ Acceptance criteria

- [ ] Builds pass on CI.
- [ ] Tests run locally.

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -6477,6 +6477,7 @@
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 				VALIDATE_PRODUCT = YES;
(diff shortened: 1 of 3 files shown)


singhhari (approved):
lgtm

DmitryPR:
Nice, glad that it got solved!

PR #1364, opened 2021-01-19
[NT-1723] Adding frameworks for Firebase Performance Monitor and Crashlytics fix.

singhhari:
# 📲 What

We're adding a couple of new Firebase dependencies into the iOS project. `Firebase Performance Monitor` (and its sub dependencies) will be an addition to our existing use of Firebase for crash reporting and mobile analytics.

While adding the new dependencies, we realized our Crashlytics data was no longer being collected. We determined the root cause was a number of sub-dependencies not being linked appropriately in our project. We included the fix for that in this PR, as well.

# 🤔 Why

- Improve the quality of the analytics we're receiving on mobile
- Parity with our Android counterpart

# 🛠 How

Our Cartfile has been updated to include the binaries for `FirebasePerformanceBinary` and `FirebaseProtobufBinary`. We've also set all our Firebase binaries to the latest release (7.4.0). It appears the developers at Firebase are considering [discontinuing support for Carthage](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md) so we should keep an eye on potentially changing our dependency manager down the line.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -3,9 +3,7 @@ import AppCenterAnalytics
 import AppCenterCrashes
```
justinswart, line 3: Let's remove `AppCenterAnalytics` and `AppCenterCrashes` as well as their linked frameworks and protocol conformance as we're not using either.

We can probably remove `crashManagerDidFinishSendingCrashReport()` as a VM input too unless we want to implement a similar callback from Crashlytics.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -6497,6 +6525,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
+				OTHER_LDFLAGS = "-ObjC";
```
justinswart, line 6528: Not sure if we need this in the linker flags? At least I didn't during testing.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -6985,6 +7016,7 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 12.4;
 				MARKETING_VERSION = 4.10.0;
 				MTL_ENABLE_DEBUG_INFO = NO;
+				ONLY_ACTIVE_ARCH = NO;
```
justinswart, line 7019: We can revert these `ONLY_ACTIVE_ARCH` values being set explicitly to `NO`. Just select wherever they've been set in Xcode and hit delete I think.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -2895,10 +2910,16 @@
 				A73924001D27230B004524C3 /* Kickstarter_Framework.framework in Frameworks */,
 				8A86D7EE24FDC84400037A7B /* GoogleAppMeasurement.framework in Frameworks */,
 				774D98DA23A9795100FC81C2 /* Optimizely.framework in Frameworks */,
+				4748C19125B7A2F80098E89E /* FirebaseCore.framework in Frameworks */,
 				D0B45B6B1EF858C00020A8DA /* KsApi.framework in Frameworks */,
 				D0D58D862257FAE000532AC1 /* Prelude.framework in Frameworks */,
 				D0D58D8E2257FAE000532AC1 /* Prelude_UIKit.framework in Frameworks */,
+				4748C14525B1021D0098E89E /* FirebaseABTesting.framework in Frameworks */,
```
justinswart, line 2917: How did you determine the list of deps that was needed in the end?

justinswart (commented):
Cool! Left some comments.

Let's also update `bin/upload-dysms-firebase.sh` with the latest from https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/run and `bin/upload-symbols` with the latest from https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/upload-symbols.

Kickstarter.xcodeproj/project.pbxproj
singhhari: Ah I was reading the directions here: https://firebase.google.com/docs/ios/setup#integrate-without-cocoapods and assumed those directions were relevant. Reading it again, I think that is only for situations where we are manually downloading the framework and dropping it into our project.

singhhari: Ah actually, I just ran a quick test and removed the linker flag and a Facebook framework throws an `unrecognized selector` error that is a result of the deletion of the linker flag. Going to leave this in.

Kickstarter.xcodeproj/project.pbxproj
singhhari: While setting it up, I was initially running into some compilation errors.  This [thread](https://github.com/firebase/firebase-ios-sdk/issues/9#issuecomment-387947163) from Firebase suggested I should take a look at the missing frameworks being referenced in the errors and manually link them.

justinswart (approved):
Cool!

PR #1365, opened 2021-01-20
[NT-1729] Landscape constraint fix for AddNewCardViewController

singhhari:
# 📲 What

While using the app, in landscape mode, and adding a new payment method, the `UIScrollView` containing the subviews was allowing for left-to-right scrolling despite an abundance of space between the "Name" `UILabel` and `UITextField`. This PR is a very small update to the width constraint in the relevant `UIStoryboard` which removes the unpleasant scrolling behavior on landscape mode.

# 🤔 Why

The need for scrolling, on landscape, was quite unnecessary given the space between the label and textfield. Adjusting the width constraint to conform to the superview is better for a form-like element.

# 🛠 How

Minor update to the constraint set in the Storyboard.

# 👀 See

Screen Shot 2021-01-20 at 6 11 08 PM


justinswart (approved):
👍

PR #1366, opened 2021-01-22
[NT-1734] Remove unused REST API references

singhhari:
# 📲 What

While documenting [a list of the API requests](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/908525614/Native+API+Documentation) currently being made in the project, we determined a few of our REST API methods were no longer in use. It's presumable that during the shift from REST to GraphQL they may have been left behind. 

# 🤔 Why

This is the first step in an eventual update of how we're handling our API interactions on iOS. Ideally, sometime later in 2021 we can add on to this work by investigating further re-factoring of some REST endpoints to GraphQL.

# 🛠 How

Updated the following files:

- `Route.swift`
- `Service.swift`
- `ServiceType.swift`
- `MockService.swift`


justinswart (approved):
Great, thanks for tidying that up!

PR #1368, opened 2021-02-05
Update README.md

singhhari:
# 📲 What

We're no longer using `Argo` as a dependency, so we've removed it from our README.

# 🤔 Why

We're now relying on Swift's `Codable`protocol to do this work for us.


jgsamudio (approved):
Now its officially removed!

PR #1371, opened 2021-02-11
[EP-157] Fix Properties (Phase One)

singhhari:
# 📲 What

This PR continues the ongoing work for phase one of our analytics overhaul. The changes, included in here, are primarily renaming and moving properties and their values to their newly defined areas.

# 🤔 Why

The properties provided context for each of our events and the Insights team has a defined set of criteria for them.


Library/OptimizelyClientType.swift
```diff
@@ -115,16 +113,8 @@ public func optimizelyProperties(environment: Environment? = AppEnvironment.curr
   }
 
   let allExperiments = optimizelyClient.allExperiments().map { experimentKey -> [String: String] in
-    let variation = try? optimizelyClient.getVariationKey(
```
jgsamudio, line 118: Question: Why did this change?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -1152,10 +1152,8 @@ private func checkoutPropertiesData(
 
   let staticUsdRate = Double(createBackingData.project.stats.staticUsdRate)
 
-  let pledgeTotalUsdCents = pledgeTotal
-    .multiplyingCurrency(staticUsdRate)
-    .multiplyingCurrency(100.0)
-    .rounded()
+  // Two decimal places to represent cent values
+  let pledgeTotalUsd = rounded(pledgeTotal.multiplyingCurrency(staticUsdRate), places: 2)
```
jgsamudio, line 1156: This variable is used in a core flow for checkout. Can we add tests for the way we are using it if we don't have them already

Library/OptimizelyClientType.swift
singhhari: A couple of relevant pieces of context:
- in phase 8 we are essentially deprecating the entirety of `optimizelyProperties`
- insights wants the format to be `session_variants_optimizely` : `[experimentKey1: variation, experimentKey2: variation ...]`

I decided to delete some of these properties here because after making the change to the format of `session_variants_optimizely` i realized those other values were not being used for anything.

Library/ViewModels/PledgeViewModel.swift
singhhari: The fix here entails changing the original representation of cents into a dollar + cents value (e.g. 1015 cents -> 10.15). It turns out that we already have a handful of tests written in the `PledgeViewModelTests` file. A lot of them broke after the fix which made sense because our templates had cent values that all needed to be updated. I want to note, however, that the change here is specifically occurring in the `checkoutPropertiesData` function which is only used for events/tracking and not rendering components of the UI, from my observation. 

After I fixed the templates and tests, it appears our coverage for both this method and the entire `PledgeViewModel` are good.

jgsamudio (approved):

Library/ViewModels/PledgeViewModel.swift
jgsamudio: Amazing! Thank you!

PR #1373, opened 2021-02-17
Update Makefile & Add Firebase In-App Message

jgsamudio:
# 📲 What

Updates the makefile to fix the `carthage-bootstrap` command. Also adds the Firebase dependency for the In-App messaging.

# 🤔 Why

We need the ability to use the Makefile to setup carthage dependencies


Cartfile.resolved
```diff
@@ -1,8 +1,9 @@
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" "7.4.0"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseCrashlyticsBinary.json" "7.4.0"
+binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseInAppMessagingBinary.json" "7.4.0"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebasePerformanceBinary.json" "7.4.0"
 binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseProtobufBinary.json" "7.4.0"
-binary "https://raw.githubusercontent.com/PerimeterX/px-iOS-Framework/master/PerimeterX.json" "1.13.8"
+binary "https://raw.githubusercontent.com/PerimeterX/px-iOS-Framework/master/PerimeterX.json" "1.13.9"
```
singhhari, line 6: I'm assuming this updated after you called `carthage update` ?

singhhari (approved):

Cartfile.resolved
jgsamudio: Yes! I just updated the cartfile to pin the perimeter x version

Makefile
```diff
@@ -42,7 +42,7 @@ dependencies: carthage-bootstrap configs secrets
 bootstrap: hooks dependencies
 
 carthage-bootstrap:
-	bin/carthage.sh || exit 1
+	bin/carthage.sh bootstrap --platform iOS || exit 1
```
justinswart, line 45: @jgsamudio catching up here - strange, this should have been handled within the script itself based on whether it's running on CI or not 🤔

jgsamudio: Yeah I was running into issues with this script where I needed to specify bootstrap and the platform. I agree that its odd that it didn't run the command for local builds

PR #1374, opened 2021-02-19
[NT-1774 ] Managing denied requests

singhhari:
# 📲 What

Re-factoring our implementation of the Perimeter X SDK with some extensions to the initial functionality. We're now enforcing a CAPTCHA screen on endpoints that are deemed suspicious.

# 🤔 Why

We're bolstering our security against bot attacks with Perimeter Xs SDK. This should add another layer of protection and mitigate future attacks.


KsApi/PerimeterXClientType.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+public protocol PerimeterXClientType: PerimeterXErrorHandler {
+  /**
+   Returns a dictionary of `[String: String]`, representing httpHeaders from Perimeter X.
+   */
+  func headers() -> [String: String]
+}
+
+public protocol PerimeterXErrorHandler {
```
jgsamudio, line 10: What do you think about creating another file for this and removing the Perimeter X prefix? I don't think this protocol is specific to the perimeter x, but open to go either way

KsApi/PerimeterXClientType.swift
```diff
@@ -0,0 +1,18 @@
...
+  /**
+   Verifies a URLs response and handles any errors with the Perimeter X SDK.
+
+   - parameter blockResponse: An `HTTPURLResponse` object with response data from a request.
+   - parameter data:           `Data` associated with the request.
```
jgsamudio, line 15: Extra comment space

Kickstarter-iOS/Library/PerimeterXClient+PerimeterXClientType.swift
```diff
@@ -0,0 +1,43 @@
...
+  public static var cookie = HTTPCookie(properties: [
+    .domain: "www.perimeterx.com", // Change according to the domain the webview will use
+    .path: "/",
+    .name: "_pxmvid",
+    .value: PXManager.sharedInstance()?.getVid() ?? "",
+    .secure: "FALSE",
+    .expires: NSDate(timeIntervalSinceNow: 3_600)
+  ])
+
+  public static func startPerimeterX(with pxManagerDelegate: PXManagerDelegate) {
```
jgsamudio, line 14: Can you document the functions that are no in the `PerimeterXClientType`?

Kickstarter-iOS/Library/PerimeterXClient+PerimeterXClientType.swift
```diff
@@ -0,0 +1,43 @@
...
+          PXManager.sharedInstance().handle(blockResponse, with: window.rootViewController) {
+            print("❌ Perimeter X CAPTCHA was successful.")
+          }
+        }
+      }
+    }
+  }
+
+  public func headers() -> [String: String] {
+    return (PXManager.sharedInstance().httpHeaders() as! [String: String])
```
jgsamudio, line 41: We should try to avoid !. What do you think about `PXManager.sharedInstance().httpHeaders() ?? [:]`

Kickstarter-iOS/Library/WebViewController.swift
```diff
@@ -9,6 +9,8 @@ internal class WebViewController: UIViewController {
   internal let webView = WKWebView(frame: .zero, configuration: WKWebViewConfiguration())
   internal var bottomAnchorConstraint: NSLayoutConstraint?
 
+  var webKitCookieStore: WKHTTPCookieStore?
```
jgsamudio, line 12: This variable can be private

Kickstarter-iOS/Library/WebViewController.swift
```diff
@@ -42,6 +46,19 @@ internal class WebViewController: UIViewController {
...
+
+  func webView(
+    _: WKWebView,
+    decidePolicyFor _: WKNavigationResponse,
+    decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void
+  ) {
+    if let newCookie = PerimeterXClient.cookie {
+      self.webKitCookieStore?.setCookie(newCookie, completionHandler: {
+        print("Perimeter X mobile VID cookie set.")
+        })
```
jgsamudio, line 58: Too much indentation

jgsamudio:
There seems to be a larger reduction in test coverage. What do you think about adding tests to the comments by codecov?

KsApi/PerimeterXClientType.swift
singhhari: Makes sense. Moved this into a separate file and removed the prefix in the latest commit.

Kickstarter-iOS/Library/PerimeterXClient+PerimeterXClientType.swift
singhhari: Ah good catch. This came right off of the documentation and I didn't notice the force casting.

Kickstarter-iOS/Library/WebViewController.swift
singhhari: Added a comment above the variable to explain what it's doing as well.

Kickstarter-iOS/Library/WebViewController.swift
singhhari: Called the formatter to fix this.

jgsamudio (approved):

PR #1376, opened 2021-02-22
[NT-1760] Adding feature flag for segment

singhhari:
# 📲 What

Adding a feature flag for Segment and implementing it within out Analytics client.

# 🤔 Why

This is simply an additional layer of control on the clients behavior in our environments.

# 🛠 How

- Added the flag on Flipper
- Decoding it on iOS
- Implementing it into the `Segment` client


jgsamudio (approved):
LGTM! Thanks Hari!

PR #1377, opened 2021-02-23
[NT-1779] Fix for physical/digital add-ons

singhhari:
# 📲 What

In our `RewardAddOnSelectionViewModel` we are currently getting the value of a reference to `shippingRuleExpanded` from the first add on returned from the rewards data. This resulted in physical add ons not being shown for `Reward`s with a digital add on as the first node returned in the response. This PR updates the logic for determining the correct shipping rule to use.

# 🤔 Why

A creator flagged this a couple of weeks ago and, after identifying the issue, only required a fairly minor tweak.


singhhari:
Snapshot testing failing due to different versions of Xcode...

jgsamudio (approved):
Looks good! Waiting on the snapshots to match the original

PR #1381, opened 2021-02-26
[Ep 224] CTA Clicked (Add Ons Continue)

tobitech:
# 📲 What

Modify how "Add-Ons Continue Button Clicked" event is tracked by adding a CTA context and checkout properties

# 🤔 Why

Convert event for segment


Library/Tracking/KSRAnalytics.swift
```diff
@@ -703,11 +703,13 @@ public final class KSRAnalytics {
...
     refTag: RefTag?
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(pledgeProperties(from: reward))
-      .withAllValuesFrom(contextProperties())
+      .withAllValuesFrom(contextProperties(ctaContext: .addOnsContinue))
+      .withAllValuesFrom(checkoutProperties(from: checkoutData, and: reward))
 
     self.track(
       event: ApprovedEvent.addOnsContinueButtonClicked.rawValue,
```
singhhari, line 715: The ticket is requesting the name of the event to be changed to "CTA Clicked". I think the best practice moving forward will be to delete the event being renamed from `ApprovedEvent`. In Afees' current PR he will be creating a new enum with the generic event names and once you rebase/pull his changes you can add `CTA Clicked` into there and replace the event on line 715.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -1215,7 +1215,7 @@ private func checkoutPropertiesData(
 
 /* A helper that assists in rounding the dollar converted amount to a given number of decimal places
  */
-private func rounded(_ value: Double, places: Int) -> Double {
+internal func rounded(_ value: Double, places: Int) -> Double {
```
singhhari, line 1218: Instead of making this internal to this view model you should move it to `SharedFunctions` in the Library and ensure it's covered by tests.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -269,13 +269,87 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
         )
       }
 
-    Signal.zip(project, baseReward, refTag)
+    let calculatedShippingTotal = Signal.combineLatest(
```
singhhari, line 272: I noticed this exact block of code you've added exists in the `PledgeViewModel`. You should re-factor so that both of these view models are not doing the same thing.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -455,3 +529,73 @@ private func isValid(
   return latestSelectedQuantities != selectedRewardQuantities(in: backing)
     || backing.locationId != selectedShippingRule?.location.id
 }
+
+private func checkoutPropertiesData(
```
singhhari, line 533: Same here. It's bad practice to copy-paste entire methods/blocks of code.

singhhari (changes requested):

Library/ViewModels/PledgeViewModel.swift
tobitech: Okay, noted.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
tobitech: They aren't entirely the same, one combines 3 signal and the other combines 2.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
tobitech: The signatures are the same but the parameters aren't.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
singhhari: I understand, it still looks like it can be cleaned up. I think you can do the work in this view model and pass the shared data to the PledgeViewModel.

tobitech: Yeah, I actually thought about that

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/SharedFunctionsTests.swift | indent |


</p>

Library/Tracking/KSRAnalytics.swift
```diff
@@ -706,14 +705,16 @@ public final class KSRAnalytics {
   public func trackAddOnsContinueButtonClicked(
     project: Project,
     reward: Reward,
+    checkoutData: CheckoutPropertiesData,
     refTag: RefTag?
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(pledgeProperties(from: reward))
```
singhhari, line 712: I don't see pledge properties listed in the `Property Groups` on the ticket.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -640,7 +640,13 @@ final class KSRAnalyticsTests: TestCase {
     let project = Project.cosmicSurgery
     let reward = Reward.template
 
-    ksrAnalytics.trackAddOnsContinueButtonClicked(project: project, reward: reward, refTag: .recommended)
+    ksrAnalytics
```
singhhari, line 643: After you remove the pledgeProperties, this method will fail. I would replace this tracking method with another one that is still using pledge properties so we can keep the coverage up.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -667,7 +673,13 @@ final class KSRAnalyticsTests: TestCase {
     let reward = Reward.noReward
       |> Reward.lens.minimum .~ 5.0
 
-    ksrAnalytics.trackAddOnsContinueButtonClicked(project: project, reward: reward, refTag: nil)
+    ksrAnalytics
```
singhhari, line 676: Same here

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1314,6 +1326,31 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertNil(self.segmentTrackingClient.traits)
   }
 
+  // MARK: CTAs Tracking
+
+  func testTrack_AddOnsContinue_CTA() {
```
singhhari, line 1331: `testTrackAddOnsContinueButtonClicked()`

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1314,6 +1326,31 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertNil(self.segmentTrackingClient.traits)
   }
 
+  // MARK: CTAs Tracking
```
singhhari, line 1329: I think the MARK is a good idea if we are putting all `CTA Clicked` events under here but right now I think they are all scattered. Two options imo:
1. Put all the existing CTA Clicked tests in here
2. Remove the MARK

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1314,6 +1326,31 @@ final class KSRAnalyticsTests: TestCase {
...
+        reward: reward,
+        checkoutData: .template,
+        refTag: nil
+      )
+
+    let dataLakeClientProps = dataLakeClient.properties.last
+    let segmentClientProps = segmentClient.properties.last
+
+    XCTAssertEqual("add_ons_continue", dataLakeClientProps?["context_cta"] as? String)
+    XCTAssertEqual("add_ons_continue", segmentClientProps?["context_cta"] as? String)
```
singhhari, line 1351: Can you add tests for the `CTA Clicked` event, `location` and assertProjectProperties and assertCheckoutProperties in here since this method is dedicated to specifically this event.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1107,8 +1107,8 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
     self.vm.inputs.configure(with: data)
     self.vm.inputs.viewDidLoad()
 
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.dataLakeTrackingClient.events)
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.segmentTrackingClient.events)
+    XCTAssertEqual(["CTA Clicked"], self.dataLakeTrackingClient.events)
```
singhhari, line 1110: A Page Viewed event should not be translating to a CTA Clicked. There is a good chance some of the logic you introduced to the view model is creating this issue, given it's occurring in multiple places throughout the file. Double check this.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -269,13 +269,89 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
         )
       }
 
-    Signal.zip(project, baseReward, refTag)
+    let calculatedShippingTotal = Signal.combineLatest(
```
singhhari, line 272: I understand that there are similarities between this and the `PledgeViewModel` but instead of copy/pasting entire blocks of code you should look to create a method that can encapsulate some of this logic and plug it in to the relevant portions of both view models/

singhhari (changes requested):

Library/ViewModels/RewardAddOnSelectionViewModel.swift
singhhari: I would also clarify w Laurie if the scope of this ticket is just for new pledges or for both new pledges and editing an existing pledge.

 If it's just new pledges then a lot of this logic becomes much simpler.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -269,13 +269,89 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+        guard reward.shipping.enabled else { return total }
+
+        let shippingCostForReward = reward.shippingRule(matching: shippingRule)?.cost ?? 0
+
+        let totalShippingForReward = shippingCostForReward
+          .multiplyingCurrency(Double(selectedQuantities[reward.id] ?? 0))
+
+        return total.addingCurrency(totalShippingForReward)
+      }
+    }
```
singhhari, line 288: A good idea here would be to re-factor these variables one block at a time. So this first one is essentially identical here and in the `PledgeViewModel`. Encapsulate this logic into one method and apply it to both view models.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -269,13 +269,89 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+
+    let defaultShippingTotal = Signal.zip(project, baseReward).map { project, baseReward -> Double in
```
singhhari, line 290: Another identical block that is a candidate for the same comment as above.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -269,13 +269,89 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+      guard baseReward.shipping.enabled, let backing = project.personalization.backing else {
+        return 0.0
+      }
+
+      return backing.shippingAmount.flatMap(Double.init) ?? 0.0
+    }
+
+    let allRewardsShippingTotal = Signal.merge(defaultShippingTotal, calculatedShippingTotal)
+
+    let allRewardsTotal = Signal.combineLatest(
```
singhhari, line 300: Same. Identify all of the snippets like these and re-factor plz.

singhhari (changes requested):

Library/Tracking/KSRAnalyticsTests.swift
tobitech: Do you mean another test method or another tracking method inside `KSRAnalytics`? If I add pledge properties wouldn't that still be sending that surplus data for that event?

Library/Tracking/KSRAnalytics.swift
```diff
@@ -706,14 +705,15 @@ public final class KSRAnalytics {
   public func trackAddOnsContinueButtonClicked(
     project: Project,
     reward: Reward,
+    checkoutData: CheckoutPropertiesData,
     refTag: RefTag?
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
-      .withAllValuesFrom(pledgeProperties(from: reward))
-      .withAllValuesFrom(contextProperties())
+      .withAllValuesFrom(contextProperties(ctaContext: .addOnsContinue))
```
singhhari, line 712: This line is not needed because of `717`.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -330,13 +330,47 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
       )
     }
 
-    Signal.zip(project, baseReward, refTag)
+    let calculatedPledgeTotal = Signal.combineLatest(
+      allRewardsShippingTotal,
+      allRewardsTotal
+    )
+    .map { shippingCost, rewardBaseAmount -> Double in
```
singhhari, line 337: You should be using the methods Afees refactored, specifically for this reason, in the last PR:
`.map(calculatePledgeTotal)`

singhhari: Also I don't think this should be in an `observeValues` block.

singhhari (changes requested):
Noticing when I track the event on Segment the `checkout_amount_count_total` and `checkout_amount_count_unique` are showing as 0 even after I add an add-on so there is a bug somewhere.

Library/Tracking/KSRAnalytics.swift
tobitech: `712` adds `context_cta` but `717` adds `context_page` so they're both needed

singhhari: good point but I also dont see a `context_page` defined for the event so  line 717 can be removed.

tobitech: Yea, that's right.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -330,13 +330,34 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
       )
     }
 
-    Signal.zip(project, baseReward, refTag)
-      .takeWhen(self.continueButtonTappedProperty.signal)
-      .observeForUI()
-      .observeValues { project, baseReward, refTag in
+    // Send updated checkout data with add-ons continue event
+    Signal.combineLatest(
```
singhhari, line 334: Looks like the add ons total count and unique count are showing up but I'm noticing another discrepancy you should look into.

I'm using the project `The Tarot Restless` and adding the following:
The $30 reward (2nd to 3rd Edition Booster Set)
2 x $15 add on (11" x 17" poster)
1 x $35 add on (18" x 24" poster)

The total in the pledge view after adding shipping is $105 but we are sending:
checkout_amount : "100.00",
checkout_amount_total_usd : 100,


My hunch is that shipping for the reward is not being accounted for.

singhhari (changes requested):

Library/ViewModels/RewardAddOnSelectionViewModel.swift
tobitech: Okay, I'll check that out

singhhari (approved):
lgtm

PR #1382, opened 2021-03-01
[EP-226] CTA Clicked (pledge_submit)

singhhari:
# 📲 What

When a user clicks the "Pledge" button we're currently firing an event labeled `Pledge Submit Button Clicked`. In this PR we are updating both the event name and the properties we send along with it.

# 🤔 Why

This is an addition to our ongoing initiative to move forward from our legacy tracking and over to Segment.


Library/Tracking/KSRAnalytics.swift
```diff
@@ -67,6 +66,7 @@ public final class KSRAnalytics {
 
   private enum NewApprovedEvent: String, CaseIterable {
     case pageViewed = "Page Viewed"
+    case ctaClicked = "CTA Clicked"
```
jgsamudio, line 69: Are we planning to rename this Enum? `NewApprovedEvent` is a little ambiguous

jgsamudio (approved):

PR #1383, opened 2021-03-01
[EP-225] CTA Clicked (pledge_initiate)

tobitech:
# 📲 What

"Project Page Pledge Button Clicked" event is updated to CTA Clicked with pledge_initiate context

# 🤔 Why

Convert DataLake event to Segment event

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -762,6 +762,7 @@ public final class KSRAnalytics {
       )
     case .pledge:
       let allProps = props
+        .withAllValuesFrom(optimizelyProperties() ?? [:])
         .withAllValuesFrom(contextProperties(ctaContext: .pledgeInitiate))
 
       self.track(
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -50,7 +50,6 @@ public final class KSRAnalytics {
     case onboardingSkipButtonClicked = "Onboarding Skip Button Clicked"
     case pledgeSubmitButtonClicked = "Pledge Submit Button Clicked"
     case projectCardClicked = "Project Card Clicked"
-    case projectPagePledgeButtonClicked = "Project Page Pledge Button Clicked"
     case projectPageViewed = "Project Page Viewed"
     case projectSwiped = "Project Swiped"
     case searchPageViewed = "Search Page Viewed"
@@ -66,6 +65,7 @@ public final class KSRAnalytics {
   }
 
   private enum NewApprovedEvent: String, CaseIterable {
+    case ctaClicked = "CTA Clicked"
     case pageViewed = "Page Viewed"
   }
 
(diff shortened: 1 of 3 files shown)


Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -899,8 +899,29 @@ final class KSRAnalyticsTests: TestCase {
...
+
+    let ksrAnalytics = KSRAnalytics(
+      dataLakeClient: dataLakeClient,
+      loggedInUser: loggedInUser,
+      segmentClient: segmentClient
+    )
+
+    ksrAnalytics.trackPledgeCTAButtonClicked(stateType: .pledge, project: project)
+
+    let dataLakeClientProps = dataLakeClient.properties.last
```
singhhari, line 920: I think you can move line 920-924 into the method above this. The tests are doing the same thing with the exception of the properties check so it's ok to consolidate.

Library/ViewModels/PledgeCTAContainerViewViewModelTests.swift
```diff
@@ -276,8 +276,8 @@ internal final class PledgeCTAContainerViewViewModelTests: TestCase {
 
     self.notifyDelegateCTATapped.assertValueCount(1)
 
-    XCTAssertEqual(["Project Page Pledge Button Clicked"], self.dataLakeTrackingClient.events)
-    XCTAssertEqual(["Project Page Pledge Button Clicked"], self.segmentTrackingClient.events)
+    XCTAssertEqual(["CTA Clicked"], self.dataLakeTrackingClient.events)
+    XCTAssertEqual(["CTA Clicked"], self.segmentTrackingClient.events)
```
singhhari, line 280: Can you check the `context_cta` property in here as well since we are doing it for the `session_` and `optimizely_` ones as well?

singhhari (changes requested):
Looks good, just a couple of small changes.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Tracking/KSRAnalyticsTests.swift | indent |


</p>

Library/ViewModels/PledgeCTAContainerViewViewModelTests.swift
```diff
@@ -294,6 +294,10 @@ internal final class PledgeCTAContainerViewViewModelTests: TestCase {
       dataLakeTrackingClientProperties?["session_variants_optimizely"],
       "Event includes Optimizely properties"
     )
+    XCTAssertNotNil(
```
singhhari, line 297: Instead of asserting `NotNil` you should check for the exact value we are expecting.

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
```diff
@@ -764,9 +763,10 @@ public final class KSRAnalytics {
     case .pledge:
       let allProps = props
         .withAllValuesFrom(optimizelyProperties)
```
singhhari, line 765: Ah, sorry, I missed this earlier but can you get rid of the `optimizelyProperties` param. It doesn't look like the new event requires this info.

singhhari (approved):
left a small comment but lgtm

Library/Tracking/KSRAnalytics.swift
tobitech: When I got rid of it, this test `testTrackingEvents_Pledge()` in `PledgeCTAContainerViewViewModelTests` fails, because it checks for those `optimizelyProperties`

PR #1384, opened 2021-03-02
[EP 228] iOS: Page Viewed (add_ons)

moderateepheezy:
# 📲 What

1. Update `Add-Ons Page Viewed` event name to `Page Viewed`
2. Add CheckoutProperties to `trackAddOnsPageViewed`

# 🤔 Why

Segment Integration

### DIFF
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -388,3 +388,94 @@ public func isEndDateAfterToday(for reward: Reward) -> Bool {
   return (reward.endsAt == nil || (reward.endsAt ?? 0) >= AppEnvironment.current.dateType.init()
     .timeIntervalSince1970)
 }
+
+/*
+ A helper that assists in rounding an amount to a given number of decimal places
+ */
+public func rounded(_ value: Double, places: Int) -> Double {
+  let divisor = pow(10.0, Double(places))
+  return (value * divisor).rounded() / divisor
+}
+
+public func checkoutProperties(
+  from project: Project,
+  baseReward: Reward,
+  rewards: [Reward],
+  selectedQuantities: SelectedRewardQuantities?,
+  additionalPledgeAmount: Double,
+  pledgeTotal: Double,
+  shippingTotal: Double,
+  checkoutId: Int? = nil,
+  isApplePay: Bool?
+) -> KSRAnalytics.CheckoutPropertiesData {
+  let staticUsdRate = Double(project.stats.staticUsdRate)
+
+  // Two decimal places to represent cent values
+  let pledgeTotalUsd = rounded(pledgeTotal.multiplyingCurrency(staticUsdRate), places: 2)
+
+  let bonusAmountUsd = additionalPledgeAmount
+    .multiplyingCurrency(staticUsdRate)
+
+  let addOnRewards = rewards
+    .filter { reward in reward.id != baseReward.id }
+    .map { reward -> [Reward] in
+      guard let selectedRewardQuantity = selectedQuantities?[reward.id] else { return [] }
+      return Array(0..<selectedRewardQuantity).map { _ in reward }
+    }
+    .flatMap { $0 }
+
+  let addOnsCountTotal = addOnRewards.map(\.id).count
+  let addOnsCountUnique = Set(addOnRewards.map(\.id)).count
+  let addOnsMinimumUsd = Format.decimalCurrency(
+    for: addOnRewards
+      .reduce(0.0) { accum, addOn in accum.addingCurrency(addOn.minimum) }
+      .multiplyingCurrency(staticUsdRate)
+  )
+
+  let shippingAmount: Double? = baseReward.shipping.enabled ? shippingTotal : nil
+
+  let amount = Format.decimalCurrency(for: pledgeTotal)
+
+  let bonusAmount = Format.decimalCurrency(for: additionalPledgeAmount)
+  let bonusAmountInUsd = Format.decimalCurrency(for: bonusAmountUsd)
+
+  let rewardId = baseReward.id
+  let estimatedDelivery = baseReward.estimatedDeliveryOn
+
+  var paymentType: String?
+  if let isApplePay = isApplePay {
+    paymentType = isApplePay
+      ? PaymentType.applePay.rawValue
+      : PaymentType.creditCard.rawValue
+  }
+
+  let shippingEnabled = baseReward.shipping.enabled
+  let shippingAmountUsd = (shippingAmount?.multiplyingCurrency(staticUsdRate))
+    .flatMap(Format.decimalCurrency)
+  let rewardTitle = baseReward.title
+  let rewardMinimumUsd = Format.decimalCurrency(for: baseReward.minimum.multiplyingCurrency(staticUsdRate))
+
+  let userHasEligibleStoredApplePayCard = AppEnvironment.current
+    .applePayCapabilities
+    .applePayCapable(for: project)
+
+  return KSRAnalytics.CheckoutPropertiesData(
+    addOnsCountTotal: addOnsCountTotal,
+    addOnsCountUnique: addOnsCountUnique,
+    addOnsMinimumUsd: addOnsMinimumUsd,
+    amount: amount,
+    bonusAmount: bonusAmount,
+    bonusAmountInUsd: bonusAmountInUsd,
+    checkoutId: checkoutId,
+    estimatedDelivery: estimatedDelivery,
+    paymentType: paymentType,
+    revenueInUsd: pledgeTotalUsd,
+    rewardId: rewardId,
+    rewardMinimumUsd: rewardMinimumUsd,
+    rewardTitle: rewardTitle,
+    shippingEnabled: shippingEnabled,
+    shippingAmount: shippingAmount,
+    shippingAmountUsd: shippingAmountUsd,
+    userHasStoredApplePayCard: userHasEligibleStoredApplePayCard
+  )
+}
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -398,17 +398,77 @@ public func rounded(_ value: Double, places: Int) -> Double {
 }
 
 /**
- An helper func that calculates default shipping
+ An helper func that calculates  shipping total for base reward
 
  - parameter project: The `Project` associated with a group of Rewards.
  - parameter baseReward: The reward being evaluated
 
- - returns: A `Double` value which is the initial shipping value. If `project` backing object is nil,
-            and `baseReward` shipping is not enabled return value would be `0.0`
+ - returns: A `Double` of the shipping value. If the `Project` `Backing` object is nil,
+            and `baseReward` shipping is not enabled, the value is `0.0`
  */
-public func getDefaultShipping(project: Project, baseReward: Reward) -> Double {
+public func getBaseRewardShippingTotal(project: Project, baseReward: Reward) -> Double {
   guard baseReward.shipping.enabled, let backing = project.personalization.backing else { return 0.0 }
   return backing.shippingAmount.flatMap(Double.init) ?? 0.0
+}
+
+/**
+ An helper func that calculates  shipping total for base reward
+
+ - parameter shippingRule: The `Project` associated with a group of Rewards.
+ - parameter addOnRewards: An array of `Reward` objects representing the available add-ons.
+ - parameter quantities: The reward being evaluated.
+
+ - returns: A `Double` of the shipping value.
+ */
+func calculateShippingTotal(
+  shippingRule: ShippingRule,
+  addOnRewards: [Reward],
+  quantities: SelectedRewardQuantities
+) -> Double {
+  let calculatedShippingTotal = addOnRewards.reduce(0.0) { total, reward in
+    guard reward.shipping.enabled else { return total }
+
+    let shippingCostForReward = reward.shippingRule(matching: shippingRule)?.cost ?? 0
+
+    let totalShippingForReward = shippingCostForReward
+      .multiplyingCurrency(Double(quantities[reward.id] ?? 0))
+
+    return total.addingCurrency(totalShippingForReward)
+  }
+
+  return calculatedShippingTotal
+}
+
+/**
+ An helper func that calculates  pledge total for all rewards
+
+ - parameter pledgeAmount: The `Project` associated with a group of Rewards.
+ - parameter shippingCost: The shipping cost for the pledge.
+ - parameter addOnRewardsTotal: The total amount of all addOn rewards.
+
+ - returns: A `Double` of the pledge value.
+ */
+func calculatePledgeTotal(
+  pledgeAmount: Double,
+  shippingCost: Double,
+  addOnRewardsTotal: Double
+) -> Double {
+  let r = [pledgeAmount, shippingCost, addOnRewardsTotal].reduce(0) { accum, amount in
+    accum.addingCurrency(amount)
+  }
+
+  return r
+}
+
+func calculateAllRewardsTotal(addOnRewards: [Reward],
+                              selectedQuantities: SelectedRewardQuantities) -> Double {
+  addOnRewards.filter { !$0.isNoReward }
+    .reduce(0.0) { total, reward -> Double in
+      let totalForReward = reward.minimum
+        .multiplyingCurrency(Double(selectedQuantities[reward.id] ?? 0))
+
+      return total.addingCurrency(totalForReward)
+    }
 }
 
 /**
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -397,7 +397,6 @@ public func rounded(_ value: Double, places: Int) -> Double {
   return (value * divisor).rounded() / divisor
 }
 
-
 /*
  * An helper func that calculates default shipping
  */
@@ -406,11 +405,27 @@ public func getDefaultShipping(project: Project, baseReward: Reward) -> Double {
   return backing.shippingAmount.flatMap(Double.init) ?? 0.0
 }
 
+/**
+ Creates `CheckoutPropertiesData` to send with our event properties.
+
+ - parameter from: The `Project` associated with the checkout.
+ - parameter baseReward: The reward being evaluated
+ - parameter addOnRewards: An array of `Reward` objects representing the available add-ons.
+ - parameter selectedQuantities: A dictionary of reward id to quantitiy.
+ - parameter additionalPledgeAmount: The bonus amount included in the pledge.
+ - parameter pledgeTotal: The total amount of the pledge.
+ - parameter shippingTotal: The shipping cost for the pledge.
+ - parameter checkoutId: The unique ID associated with the checkout.
+ - parameter isApplePay: A `Bool` indicating if the pledge was done with Apple pay.
+
+ - returns: A `CheckoutPropertiesData` object required for checkoutProperties.
+ */
+
 public func checkoutProperties(
   from project: Project,
   baseReward: Reward,
-  rewards: [Reward],
-  selectedQuantities: SelectedRewardQuantities?,
+  addOnRewards: [Reward],
+  selectedQuantities: SelectedRewardQuantities,
   additionalPledgeAmount: Double,
   pledgeTotal: Double,
   shippingTotal: Double,
@@ -425,10 +440,10 @@ public func checkoutProperties(
   let bonusAmountUsd = additionalPledgeAmount
     .multiplyingCurrency(staticUsdRate)
 
-  let addOnRewards = rewards
+  let addOnRewards = addOnRewards
     .filter { reward in reward.id != baseReward.id }
     .map { reward -> [Reward] in
-      guard let selectedRewardQuantity = selectedQuantities?[reward.id] else { return [] }
+      guard let selectedRewardQuantity = selectedQuantities[reward.id] else { return [] }
       return Array(0..<selectedRewardQuantity).map { _ in reward }
     }
     .flatMap { $0 }
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -475,7 +475,7 @@ func calculatePledgeTotal(
  An helper func that calculates  pledge total for all rewards
 
  - parameter addOnRewards: The `Project` associated with a group of Rewards.
- - parameter selectedQuantities: The shipping cost for the pledge.
+ - parameter selectedQuantities: A dictionary that aggregates the quantity of selected add-ons.
 
  - returns: A `Double` of all rewards add-ons total.
  */
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -397,6 +397,15 @@ public func rounded(_ value: Double, places: Int) -> Double {
   return (value * divisor).rounded() / divisor
 }
 
+
+/*
+ * An helper func that calculates default shipping
+ */
+public func getDefaultShipping(project: Project, baseReward: Reward) -> Double {
+  guard baseReward.shipping.enabled, let backing = project.personalization.backing else { return 0.0 }
+  return backing.shippingAmount.flatMap(Double.init) ?? 0.0
+}
+
 public func checkoutProperties(
   from project: Project,
   baseReward: Reward,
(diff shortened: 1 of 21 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -733,14 +732,16 @@ public final class KSRAnalytics {
   public func trackAddOnsPageViewed(
     project: Project,
     reward: Reward,
-    refTag: RefTag?
+    refTag: RefTag?,
+    checkoutData: CheckoutPropertiesData
```
singhhari, line 736: I've noticed on all the other tracking methods, the `refTag` is the last parameter. Can you update this one as well?

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1468,6 +1468,13 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("new_pledge", dataLakeClient.properties.last?["context_type"] as? String)
     XCTAssertEqual("new_pledge", segmentClient.properties.last?["context_type"] as? String)
 
+    ksrAnalytics
+      .trackAddOnsPageViewed(project: .template, reward: .template, refTag: nil, checkoutData: .template)
+    XCTAssertEqual("add_ons", dataLakeClient.properties.last?["context_page"] as? String)
+    XCTAssertEqual("add_ons", segmentClient.properties.last?["context_page"] as? String)
+    XCTAssertEqual(nil, dataLakeClient.properties.last?["context_type"] as? String)
```
singhhari, line 1475: No need to check for nil here

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -455,3 +460,36 @@ private func isValid(
...
+  let addOnsMinimumUsd = Format.decimalCurrency(
+    for: rewards
+      .reduce(0.0) { accum, addOn in accum.addingCurrency(addOn.minimum) }
+      .multiplyingCurrency(staticUsdRate)
+  )
+  return KSRAnalytics.CheckoutPropertiesData(
+    addOnsCountTotal: nil,
+    addOnsCountUnique: nil,
+    addOnsMinimumUsd: addOnsMinimumUsd,
+    amount: "",
```
singhhari, line 480: Is there a way we can avoid passing empty strings? Some of the other information available in this view model should make this possible.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1107,8 +1107,8 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
     self.vm.inputs.configure(with: data)
     self.vm.inputs.viewDidLoad()
 
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.dataLakeTrackingClient.events)
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.segmentTrackingClient.events)
+    XCTAssertEqual([], self.dataLakeTrackingClient.events)
```
singhhari, line 1110: Based on the older line this should contain a `Page Viewed` event?

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1199,8 +1199,8 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
...
+    XCTAssertEqual([], self.segmentTrackingClient.events)
```
singhhari, line 1203: Same here.

singhhari (changes requested):

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/RewardAddOnSelectionViewModel.swift | indent |
| Library/ViewModels/RewardAddOnSelectionViewModelTests.swift | indent |


</p>

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,15 +256,55 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
         AppEnvironment.current.ksrAnalytics.trackAddOnsPageViewed(
           project: project,
           reward: baseReward,
+          checkoutData: checkoutProperties(
+            from: project,
+            baseReward: baseReward,
+            rewards: selectedRewards,
+            selectedQuantities: configData.selectedQuantities,
+            additionalPledgeAmount: 0,
+            pledgeTotal: 0,
```
singhhari, line 304: Definitely not 0. We should be using the value of the reward for the total here.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,15 +256,55 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+      .observeValues { project, baseReward, selectedRewards, refTag, configData, rewardsShippingTotal in
         AppEnvironment.current.ksrAnalytics.trackAddOnsPageViewed(
           project: project,
           reward: baseReward,
+          checkoutData: checkoutProperties(
+            from: project,
+            baseReward: baseReward,
+            rewards: selectedRewards,
+            selectedQuantities: configData.selectedQuantities,
+            additionalPledgeAmount: 0,
```
singhhari, line 303: Depending on if Laurie wants this event to handle an edited pledge this may require you to use the existing `Backing` (project.personalization.backing) instead of hard-coding 0.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1107,25 +1107,25 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
...
     self.vm.inputs.viewDidLoad()
 
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.dataLakeTrackingClient.events)
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.segmentTrackingClient.events)
-
     self.scheduler.advance()
     self.vm.inputs.shippingRuleSelected(shippingRule)
     self.scheduler.advance()
 
+    XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)
```
singhhari, line 1114: We shouldn't be moving the line down. We expect the event to be fired after viewDidLoad.

singhhari: Ah and can you include a check for the properties after we check the event?

singhhari (changes requested):

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,15 +256,55 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
     .map(PledgeViewData.init)
     .takeWhen(self.continueButtonTappedProperty.signal)
 
+    let defaultShippingTotal = Signal.zip(project, baseReward)
```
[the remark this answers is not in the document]
moderateepheezy: The refactor would result to duplicate we are trying to avoid, because `PledgeViewModel` uses the result of it's calculation else where in the code.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -725,14 +726,16 @@ public final class KSRAnalytics {
   public func trackAddOnsPageViewed(
     project: Project,
     reward: Reward,
+    checkoutData: CheckoutPropertiesData,
     refTag: RefTag?
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(pledgeProperties(from: reward))
```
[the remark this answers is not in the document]
moderateepheezy: Will correct that now.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -26,6 +26,7 @@ public final class KSRAnalytics {
     case activityFeedViewed = "Activity Feed Viewed"
     case addNewCardButtonClicked = "Add New Card Button Clicked"
     case addOnsContinueButtonClicked = "Add-Ons Continue Button Clicked"
+    case campaignDetailsButtonClicked = "Campaign Details Button Clicked"
```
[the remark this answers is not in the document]
moderateepheezy: I didn't add this, this is probably from a merge.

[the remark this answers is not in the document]
singhhari: Cool, after you pull the changes it should be removed automatically.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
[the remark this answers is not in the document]
singhhari: Not sure I understand what you mean. We don't need to replicate the logic of the PledgeViewModel everywhere. My understanding is that the shipping, at this point, should only be the shipping cost associated with the base reward since the add on has not been selected yet. But again, clarify with Laurie about whether this event should handle only new pledges or both new pledges + edits.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
moderateepheezy: So what fires the "Page Viewed" event is inside an asynchronous task, `XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)` is being called immediately after viewDidLoad, so there's a possibility that `XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)` will be called before the "Page Viewed" event is fired. 

Moving `XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)` to after `self.scheduler.advance()` would ensure that all actions that have been scheduled are executed  as soon as possible.

 I think the fact that it passes previously is a false positive.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
moderateepheezy: This has updated as well.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
singhhari: Not sure I agree with this. If the test was working before then it is fair to assume it was correct. The issue is very likely related to the changes you applied to the view model. Keep in mind many more parameters were introduced to the tracking logic (`RewardAddOnSelectionViewModel line 284`) and that is likely the source of the issue.

singhhari (changes requested):

Library/SharedFunctions.swift
```diff
@@ -397,6 +397,88 @@ public func rounded(_ value: Double, places: Int) -> Double {
...
+ An helper func that calculates  shipping total for base reward
+
+ - parameter project: The `Project` associated with a group of Rewards.
+ - parameter baseReward: The reward being evaluated
+
+ - returns: A `Double` of the shipping value. If the `Project` `Backing` object is nil,
+            and `baseReward` shipping is not enabled, the value is `0.0`
+ */
+public func getBaseRewardShippingTotal(project: Project, baseReward: Reward) -> Double {
+  guard baseReward.shipping.enabled, let backing = project.personalization.backing else { return 0.0 }
```
singhhari, line 410: So I just realized something while I was testing on your branch. On a new pledge this always return 0 because there is no `Backing`.  The desired behavior is to return the shipping amount designated on the Reward for a new pledge.

I tested this while doing a new pledge for the project `The Tarot Restless: Third Edition`. The shipping amount is $5 to the destination I chose but we are sending `checkout_shipping_amount`: 0 to Segment.

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,116 @@ final class SharedFunctionsTests: TestCase {
       checkoutPropertiesData.userHasStoredApplePayCard
     )
   }
+
+  func testGetShipping_ShippingEnabled() {
```
singhhari, line 408: This test appears to be actually testing a situation where you are testing the shipping amount based on an existing Backing so the name is not accurate imo. 

I would consider renaming this something like `testGetShipping_Backing` and writing another test for the situation when the Backing doesn't exist and the Reward has a non-zero shipping amount.

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,116 @@ final class SharedFunctionsTests: TestCase {
...
+      )
+
+    let shippingTotal = getBaseRewardShippingTotal(project: project, baseReward: reward)
+
+    XCTAssertEqual(2.0, shippingTotal)
+  }
+
+  func testGetShipping_ShippingDisabled() {
+    let reward = Reward.template
+      |> Reward.lens.limit .~ 5
```
singhhari, line 432: The test does not require 432-434

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,116 @@ final class SharedFunctionsTests: TestCase {
...
+    let shippingTotal = calculateShippingTotal(
+      shippingRule: shippingRule,
+      addOnRewards: [reward],
+      quantities: [:]
+    )
+
+    XCTAssertEqual(0.0, shippingTotal)
+  }
+
+  func testCalculated_Pledge_Total() {
```
singhhari, line 489: Can we try to test this by combining the outputs of the other functions you defined instead of hardcoding the arguments?

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1107,25 +1107,27 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
     self.vm.inputs.configure(with: data)
     self.vm.inputs.viewDidLoad()
 
-    XCTAssertEqual(["Add-Ons Page Viewed"], self.dataLakeTrackingClient.events)
```
singhhari, line 1110: I remember we spoke about this last week. This should not have been deleted and moved down. Please look into this as I'm fairly certain it is the designed functionality. It is probably directly related to the changes applied to the `RewardAddOnSelectionViewModel`.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -1199,13 +1201,15 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
...
```
singhhari, line 1202: Same here.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,18 +256,79 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
     .takeWhen(self.continueButtonTappedProperty.signal)
 
+    let calculatedShippingTotal = Signal.combineLatest(
+      shippingRule.skipNil(),
+      allRewards,
+      selectedQuantities
+    )
+    .map(calculateShippingTotal)
+
+    let shippingTotal = Signal.zip(project, baseReward)
```
singhhari, line 266: This var should probably be called `baseRewardShippingTotal` based on the name of the method being used.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,18 +256,79 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
     .map(PledgeViewData.init)
     .takeWhen(self.continueButtonTappedProperty.signal)
 
+    let calculatedShippingTotal = Signal.combineLatest(
```
singhhari, line 259: This should be `shippingTotal` based on the method name

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,18 +256,79 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+    )
+    .take(first: 1)
+
+    let allRewardsTotal = Signal.combineLatest(
+      selectedRewards,
+      selectedQuantities
+    )
+    .map(calculateAllRewardsTotal)
+
+    let calculatedPledgeTotal = Signal.combineLatest(
```
singhhari, line 287: This is an opinion but the variable name here is also not very descriptive. It should probably be something like `combinedPledgeTotal` or something along those lines.

singhhari (changes requested):

Library/SharedFunctions.swift
```diff
@@ -397,6 +397,99 @@ public func rounded(_ value: Double, places: Int) -> Double {
...
+  }
+
+  return r
+}
+
+/**
+ An helper func that calculates  pledge total for all rewards
+
+ - parameter addOnRewards: The `Project` associated with a group of Rewards.
+ - parameter selectedQuantities: The shipping cost for the pledge.
```
singhhari, line 478: Update description of the parameter

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,145 @@ final class SharedFunctionsTests: TestCase {
...
+      baseReward: reward,
+      shippingRule: shippingRule
+    )
+
+    XCTAssertEqual(5.0, shippingTotal)
+  }
+
+  func testGetShipping_ShippingDisabled() {
+    let reward = Reward.template
+      |> Reward.lens.limit .~ 5
```
singhhari, line 455: 455-457 not needed here

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,145 @@ final class SharedFunctionsTests: TestCase {
...
+
+    let quantities: SelectedRewardQuantities = [
+      reward.id: 1,
+      addOn1.id: 2,
+      addOn2.id: 1
+    ]
+
+    let shippingTotal = calculateShippingTotal(
+      shippingRule: shippingRule,
+      addOnRewards: [reward],
```
singhhari, line 495: Shouldn't this contain the addOn1 and addOn2 defined above?

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,145 @@ final class SharedFunctionsTests: TestCase {
...
+      quantities: [:]
+    )
+
+    XCTAssertEqual(0.0, shippingTotal)
+  }
+
+  func testCalculated_Pledge_Total() {
```
singhhari, line 518: It would ideal if you tested this by defining the rewards/addons and using their amounts to calculate the total

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,145 @@ final class SharedFunctionsTests: TestCase {
...
+    let addOn2 = Reward.template
+      |> Reward.lens.id .~ 10
+
+    let quantities: SelectedRewardQuantities = [
+      reward.id: 1,
+      addOn1.id: 2,
+      addOn2.id: 1
+    ]
+
+    let rewardsTotal = calculateAllRewardsTotal(
```
singhhari, line 541: Shouldnt selectedQuantities and addOnRewards be related here? I'm seeing the quantities with the reward and addons but the addOnRewards argument is just an array with one reward in it.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -156,40 +156,17 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
-    }
-
-    let defaultShippingTotal = Signal.zip(project, baseReward)
-      .map { project, baseReward -> Double in
-        guard baseReward.shipping.enabled, let backing = project.personalization.backing else { return 0.0 }
+    .map(calculateShippingTotal)
 
-        return backing.shippingAmount.flatMap(Double.init) ?? 0.0
-      }
+    let defaultShippingTotal = Signal.zip(project, baseReward, selectedShippingRule)
```
singhhari, line 168: a better name would be related to the method being applied. something like `baseRewardShippingTotal`

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -256,18 +256,80 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
...
+      baseRewardShippingTotal,
+      shippingTotal
+    )
+
+    // Addtional pledge amount is zero if not backed.
+    let additionalPledgeAmount = Signal.merge(
+      configData.filter { $0.project.personalization.backing == nil }.mapConst(0.0),
+      project.map { $0.personalization.backing }.skipNil().map(\.bonusAmount)
+    )
+    .take(first: 1)
```
singhhari, line 279: Curious why take first here?

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
singhhari: We determined that because the shipping rule is required before the event is fired, it makes sense that this assertion is moved lower.

singhhari (changes requested):

Library/SharedFunctionsTests.swift
```diff
@@ -404,4 +404,148 @@ final class SharedFunctionsTests: TestCase {
...
+      |> Reward.lens.id .~ 5
+    let addOn2 = Reward.template
+      |> Reward.lens.id .~ 10
+
+    let shippingRule = ShippingRule.template
+
+    let shippingTotal = calculateShippingTotal(
+      shippingRule: shippingRule,
+      addOnRewards: [addOn1, addOn2],
+      quantities: [:]
```
singhhari, line 515: supply the quantities based on the addons and reward defined

singhhari (changes requested):
Whatever is unresolved is the only bit left and then we can merge this,.

Library/SharedFunctionsTests.swift
```diff
@@ -509,21 +509,113 @@ final class SharedFunctionsTests: TestCase {
...
+    )
+
+    let allRewardsTotal = calculateAllRewardsTotal(
+      addOnRewards: addOnRewards,
+      selectedQuantities: quantities
+    )
+
     let pledgeTotal = calculatePledgeTotal(
-      pledgeAmount: 20.0, shippingCost: 5.0, addOnRewardsTotal: 50
+      pledgeAmount: 0.0,
```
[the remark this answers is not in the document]
moderateepheezy: Done.

Library/SharedFunctionsTests.swift
```diff
@@ -509,21 +509,113 @@ final class SharedFunctionsTests: TestCase {
...
+      |> Reward.lens.hasAddOns .~ true
+      |> Reward.lens.shipping.enabled .~ true
+    let addOn1 = Reward.template
+      |> Reward.lens.id .~ 5
+      |> Reward.lens.shipping .~ (.template |> Reward.Shipping.lens.enabled .~ true)
+    let addOn2 = Reward.template
+      |> Reward.lens.id .~ 10
+      |> Reward.lens.shipping .~ (.template |> Reward.Shipping.lens.enabled .~ true)
+
+    let shippingRule = ShippingRule.template
```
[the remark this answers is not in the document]
moderateepheezy: Done

Library/SharedFunctionsTests.swift
```diff
@@ -509,21 +509,113 @@ final class SharedFunctionsTests: TestCase {
...
+      reward.id: 1,
+      addOn1.id: 2,
+      addOn2.id: 1
+    ]
+
     let shippingTotal = calculateShippingTotal(
       shippingRule: shippingRule,
-      addOnRewards: [addOn1, addOn2],
-      quantities: [:]
+      addOnRewards: [reward, addOn1, addOn2],
```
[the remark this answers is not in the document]
moderateepheezy: Done

Library/SharedFunctionsTests.swift
```diff
@@ -509,21 +509,113 @@ final class SharedFunctionsTests: TestCase {
...
+      |> Project.lens.personalization.backing .~ (
+        .template
+          |> Backing.lens.reward .~ reward
+          |> Backing.lens.rewardId .~ reward.id
+          |> Backing.lens.shippingAmount .~ 10
+          |> Backing.lens.amount .~ 700.0
+          |> Backing.lens.bonusAmount .~ 700.0
+      )
+
+    let shippingRule = ShippingRule.template
```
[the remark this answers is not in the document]
moderateepheezy: Done

singhhari (approved):
lgtm

PR #1385, opened 2021-03-02
[EP-230] Page Viewed (project)

singhhari:
# 📲 What

Updating the event that is called when a `Project` is viewed on native iOS. In addition, we are including functionality that will trigger an event when the Campaign, Updates or Comments associated with a given project, are viewed.

# 🤔 Why

Continuing to extend our Segment implementation and flesh out our new analytics client.

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -1138,8 +1138,9 @@ public final class KSRAnalytics {
   /**
    Call when a project page is viewed.
 
-   - parameter project:      The project being viewed.
-   - parameter refTag:       The ref tag used when opening the project.
+   - parameter project: The project being viewed.
+   - parameter refTag: The ref tag used when opening the project.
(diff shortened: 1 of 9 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -330,26 +328,20 @@ public final class KSRAnalytics {
...
-    case faq
     case overview
-    case risks
     case updates
 
     var trackingString: String {
       switch self {
       case .campaign: return "campaign"
       case .comments: return "comments"
-      case .faq: return "faq"
```
jgsamudio, line 350: I guess we aren't using this in the app

jgsamudio (approved):

Library/Tracking/KSRAnalytics.swift
singhhari: Correct. These are tabs in the webview and will likely be handled. by web.

PR #1386, opened 2021-03-02
Checkout Properties Refactor

tobitech:
# 📲 What

Refactor checkoutProperties into SharedFunctions

# 🤔 Why

We want to be able to reuse a function that creates checkout properties for different view models in the pledge flow

### DIFF
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -388,3 +388,94 @@ public func isEndDateAfterToday(for reward: Reward) -> Bool {
   return (reward.endsAt == nil || (reward.endsAt ?? 0) >= AppEnvironment.current.dateType.init()
     .timeIntervalSince1970)
 }
+
+/*
+ A helper that assists in rounding an amount to a given number of decimal places
+ */
+public func rounded(_ value: Double, places: Int) -> Double {
+  let divisor = pow(10.0, Double(places))
+  return (value * divisor).rounded() / divisor
+}
+
+public func checkoutProperties(
+  from project: Project,
+  baseReward: Reward,
+  rewards: [Reward],
+  selectedQuantities: SelectedRewardQuantities,
+  additionalPledgeAmount: Double,
+  pledgeTotal: Double,
+  shippingTotal: Double,
+  checkoutId: Int? = nil,
+  isApplePay: Bool?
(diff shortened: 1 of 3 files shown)


moderateepheezy (approved):

Library/SharedFunctions.swift
```diff
@@ -388,3 +388,94 @@ public func isEndDateAfterToday(for reward: Reward) -> Bool {
...
+}
+
+public func checkoutProperties(
+  from project: Project,
+  baseReward: Reward,
+  rewards: [Reward],
+  selectedQuantities: SelectedRewardQuantities,
+  additionalPledgeAmount: Double,
+  pledgeTotal: Double,
+  shippingTotal: Double,
```
singhhari, line 407: Same here? If possible, we should try to make the function signature as simple as functionally possible. That way, whenever we need to use it we only need to pass what's absolutely necessary and all the logic required to create `CheckoutPropertiesData` is encapsulated in the method body.

Library/SharedFunctions.swift
```diff
@@ -388,3 +388,94 @@ public func isEndDateAfterToday(for reward: Reward) -> Bool {
...
+ */
+public func rounded(_ value: Double, places: Int) -> Double {
+  let divisor = pow(10.0, Double(places))
+  return (value * divisor).rounded() / divisor
+}
+
+public func checkoutProperties(
+  from project: Project,
+  baseReward: Reward,
+  rewards: [Reward],
```
singhhari, line 403: I think rewards are a property of `project` or `project.rewardData` so I would double check if this parameter is necessary.

singhhari (changes requested):
Added some comments that will require you to test this/ play around with it a bit. Try and simplify the method even more so we can use this more flexibly.

Library/SharedFunctions.swift
tobitech: I think it represents selected rewards not all the project rewards, but I'm not so sure about that as well. I've seen cases in the project where a single selected reward was turned into an array of one item.

singhhari (approved):
lgtm. 

@moderateepheezy @tobitech let's try and be very observant of re-factoring further as we start using the method for more of the event work.

PR #1387, opened 2021-03-04
[EP-227] CTA Clicked (reward_continue)

singhhari:
# 📲 What

This PR includes an update to tracking we're currently using in the Rewards carousel. When a reward is selected, we are tracking a handful of properties associated with the context.

# 🤔 Why

Continuing phase one of our Segment implementation.

### DIFF
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -397,11 +397,26 @@ public func rounded(_ value: Double, places: Int) -> Double {
   return (value * divisor).rounded() / divisor
 }
 
+/**
+ Creates `CheckoutPropertiesData` to send with our event properties.
+
+ - parameter from: The `Project` associated with the checkout.
+ - parameter baseReward: The reward being evaluated
+ - parameter addOnRewards: An array of `Reward` objects representing the available add-ons.
+ - parameter selectedQuantities: A dictionary of reward id to quantitiy.
+ - parameter additionalPledgeAmount: The bonus amount included in the pledge.
+ - parameter pledgeTotal: The total amount of the pledge.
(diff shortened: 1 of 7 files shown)


Library/SharedFunctions.swift
```diff
@@ -397,11 +397,26 @@ public func rounded(_ value: Double, places: Int) -> Double {
...
+ - parameter isApplePay: A `Bool` indicating if the pledge was done with Apple pay.
+
+ - returns: A `CheckoutPropertiesData` object required for checkoutProperties.
+ */
 public func checkoutProperties(
   from project: Project,
   baseReward: Reward,
-  rewards: [Reward],
-  selectedQuantities: SelectedRewardQuantities?,
+  addOnRewards: [Reward],
```
jgsamudio, line 418: Nice rename 👍

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -177,9 +177,28 @@ public final class RewardsCollectionViewModel: RewardsCollectionViewModelType,
     // Tracking
     Signal.combineLatest(project, selectedRewardFromId, refTag)
       .observeValues { project, reward, refTag in
+
+        // The `Backing` is nil for a new pledge.
+        let backing = project.personalization.backing
```
jgsamudio, line 182: Do we have tests for this logic and the checkoutPropertiesData? It seems to be critical logic for these properties. Do we also use this anywhere else?

singhhari: Good call. Added some tests to handle whether there is a Backing or not

jgsamudio (approved):

PR #1392, opened 2021-03-11
[EP-242] CTA Clicked (search)

moderateepheezy:
# 📲 What

CTA Clicked (search)

# 🤔 Why

Segment Integration

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -25,7 +25,6 @@ public final class KSRAnalytics {
   private enum ApprovedEvent: String, CaseIterable {
     case activityFeedViewed = "Activity Feed Viewed"
     case addNewCardButtonClicked = "Add New Card Button Clicked"
-    case addOnsContinueButtonClicked = "Add-Ons Continue Button Clicked"
     case addOnsPageViewed = "Add-Ons Page Viewed"
     case collectionViewed = "Collection Viewed"
     case continueWithAppleButtonClicked = "Continue With Apple Button Clicked"
@@ -728,15 +727,14 @@ public final class KSRAnalytics {
   public func trackAddOnsContinueButtonClicked(
     project: Project,
     reward: Reward,
+    checkoutData: CheckoutPropertiesData,
     refTag: RefTag?
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
-      .withAllValuesFrom(pledgeProperties(from: reward))
-      .withAllValuesFrom(contextProperties())
-
+      .withAllValuesFrom(contextProperties(ctaContext: .addOnsContinue))
+      .withAllValuesFrom(checkoutProperties(from: checkoutData, and: reward))
     self.track(
-      event: ApprovedEvent.addOnsContinueButtonClicked.rawValue,
-      page: .addOnsSelection,
+      event: NewApprovedEvent.ctaClicked.rawValue,
       properties: props,
       refTag: refTag?.stringTag
     )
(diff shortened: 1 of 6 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -413,6 +413,19 @@ public final class KSRAnalytics {
     }
   }
 
+  /**
+   An optional context that provides an addtional context  about the location of the event.
```
singhhari, line 417: Just to clarify, the context itself is not optional but when you use it as a parameter it is. Replace with:
`A context providing additional details about the location the event occurs.`

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1126,6 +1143,20 @@ public final class KSRAnalytics {
...
+  /// Called when the search tab bar is clicked.
+  public func trackProjectSearchButtonClicked() {
+    let props = contextProperties(
+      ctaContext: .search,
+      locationContext: .globalNav
+    )
+
+    self.track(
+      event: NewApprovedEvent.ctaClicked.rawValue,
+      location: .search,
```
singhhari, line 1155: Thinking we should rename this parameter, globally, to `page` to avoid confusion with the new `LocationContext` you added.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -553,12 +566,16 @@ public final class KSRAnalytics {
...
   public func trackTabBarClicked(_ tabBarItemLabel: TabBarItemLabel) {
-    let properties = contextProperties(tabBarLabel: tabBarItemLabel)
-
-    self.track(
-      event: ApprovedEvent.tabBarClicked.rawValue,
-      properties: properties
-    )
+    switch tabBarItemLabel {
+    case .search:
+      self.trackProjectSearchButtonClicked()
```
singhhari, line 571: I think it's ok to drop the logic directly in here since we are only using the method in one place

Library/Tracking/KSRAnalytics.swift
singhhari: Look at line 761 for reference

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1541,6 +1573,7 @@ private func contextProperties(
   result["timestamp"] = AppEnvironment.current.dateType.init().timeIntervalSince1970
   result["tab_bar_label"] = tabBarLabel?.trackingString
   result["type"] = typeContext?.trackingString
+  result["location"] = locationContext?.trackingString
```
singhhari, line 1576: Looks like the values are in alphabetical order so move this up a few lines.

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
moderateepheezy: I understand, I usually have a preference of moving any logic more than 4 lines in a switch case to a new function, just for readability. I have updated this regardless.

Library/Tracking/KSRAnalytics.swift
moderateepheezy: Oh yeah, I thought about it too. I think I mentioned something similar last week or before, when I was working on `trackThanksPageViewed`, you said something about leaving it for now. I should have asked again.

I guess the new`LocationContext` I added requires that we change it.

moderateepheezy: Updated.

singhhari (approved):
lgtm

PR #1395, opened 2021-03-15
[EP-240] CTA clicked discover sort

moderateepheezy:
# 📲 What

A tracking event for when Discovery Param sorts are clicked, this is specific for Discovery Advanced sort (`magic`, `popular`, `newest`, `coming soon`).

# 🤔 Why

Segment Integration Phase 2

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -676,21 +676,26 @@ public final class KSRAnalytics {
   /**
    Call when the user swipes between sorts or selects a sort.
 
+   if a user is on Discover Advanced and has results sorted by magic, but then clicks the button to sort by popularity,
+   that would be discover_sort = magic, context_cta = discover_sort, context_type = popular, context_location = discover_advanced, context_page = discover.
+
    - parameter sort: The new sort that was selected.
+   - parameter prevSort: The last sort selected before the new sort.
+   - parameter params: addtional parameters associated with the current selected sort.
    */
   public func trackDiscoverySelectedSort(
-    prevSort sort: DiscoveryParams.Sort,
-    nextSort: DiscoveryParams.Sort,
+    sort: DiscoveryParams.Sort,
+    prevSort: DiscoveryParams.Sort,
     params: DiscoveryParams
   ) {
     let props = discoveryProperties(from: params)
       .withAllValuesFrom([
-        "discover_sort": sort.trackingString
+        "discover_sort": prevSort.trackingString
       ])
       .withAllValuesFrom(
         contextProperties(
           ctaContext: .discoverSort,
-          typeContext: .discovery(TrackingHelpers.discoveryContext(for: nextSort)),
+          typeContext: .discovery(TrackingHelpers.discoveryContext(for: sort)),
           locationContext: .discoverAdvanced
         )
       )
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -30,7 +30,6 @@ public final class KSRAnalytics {
     case continueWithAppleButtonClicked = "Continue With Apple Button Clicked"
     case editorialCardClicked = "Editorial Card Clicked"
     case explorePageViewed = "Explore Page Viewed"
-    case exploreSortClicked = "Explore Sort Clicked"
     case fbLoginOrSignupButtonClicked = "Facebook Log In or Signup Button Clicked"
     case filterClicked = "Filter Clicked"
     case fixPledgeButtonClicked = "Fix Pledge Button Clicked"
@@ -354,6 +353,7 @@ public final class KSRAnalytics {
     case backed
     case categoryName
     case creditCard
+    case discovery(DiscoveryContext)
     case facebook
     case location
     case percentRaised
@@ -369,6 +369,22 @@ public final class KSRAnalytics {
     case tag
     case unwatch
     case watch
+
+    public enum DiscoveryContext {
+      case endingSoon
+      case magic
+      case newest
+      case popular
+
+      var trackingString: String {
+        switch self {
+        case .endingSoon: return "ending_soon"
+        case .magic: return "magic"
+        case .newest: return "newest"
+        case .popular: return "popular"
+        }
+      }
+    }
 
     public enum PledgeContext {
       case fixErroredPledge
@@ -393,6 +409,7 @@ public final class KSRAnalytics {
       case .backed: return "backed"
       case .categoryName: return "category_name"
       case .creditCard: return "credit_card"
+      case let .discovery(discoveryContext): return discoveryContext.trackingString
       case .facebook: return "facebook"
       case .location: return "location"
       case .percentRaised: return "percent_raised"
@@ -416,10 +433,14 @@ public final class KSRAnalytics {
    A context providing additional details about the location the event occurs.
    */
   public enum LocationContext {
+    case discoverAdvanced
+    case discoverOverlay
     case globalNav
 
     var trackingString: String {
       switch self {
+      case .discoverAdvanced: return "discover_advanced"
+      case .discoverOverlay: return "discover_overlay"
       case .globalNav: return "global_nav"
       }
     }
@@ -662,9 +683,16 @@ public final class KSRAnalytics {
       .withAllValuesFrom([
         "discover_sort": sort.rawValue
       ])
+      .withAllValuesFrom(
+        contextProperties(
+          ctaContext: .discoverSort,
+          typeContext: .discovery(TrackingHelpers.discoveryContext(for: sort)),
+          locationContext: .discoverAdvanced
+        )
+      )
 
     self.track(
-      event: ApprovedEvent.exploreSortClicked.rawValue,
+      event: NewApprovedEvent.ctaClicked.rawValue,
       page: .discovery,
       properties: props
     )
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -678,15 +678,19 @@ public final class KSRAnalytics {
 
    - parameter sort: The new sort that was selected.
    */
-  public func trackDiscoverySelectedSort(nextSort sort: DiscoveryParams.Sort, params: DiscoveryParams) {
+  public func trackDiscoverySelectedSort(
+    prevSort sort: DiscoveryParams.Sort,
+    nextSort: DiscoveryParams.Sort,
+    params: DiscoveryParams
+  ) {
     let props = discoveryProperties(from: params)
       .withAllValuesFrom([
-        "discover_sort": sort.rawValue
+        "discover_sort": sort.trackString
       ])
       .withAllValuesFrom(
         contextProperties(
           ctaContext: .discoverSort,
-          typeContext: .discovery(TrackingHelpers.discoveryContext(for: sort)),
+          typeContext: .discovery(TrackingHelpers.discoveryContext(for: nextSort)),
           locationContext: .discoverAdvanced
         )
       )
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -583,6 +583,15 @@ public final class KSRAnalytics {
     )
   }
 
+  /// Call when the user is logged out, on the `Activity` tab and taps the `Explore Projects`button.
+  public func trackExploreButtonClicked() {
+    let properties = contextProperties(ctaContext: .discover, locationContext: .globalNav)
+    self.track(
+      event: NewApprovedEvent.ctaClicked.rawValue,
+      properties: properties
+    )
+  }
+
   // MARK: - Application Lifecycle
 
   public func trackTabBarClicked(_ tabBarItemLabel: TabBarItemLabel) {
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -353,7 +353,7 @@ public final class KSRAnalytics {
     case backed
     case categoryName
     case creditCard
-    case discovery(DiscoveryContext)
+    case discovery(DiscoverySortContext)
     case facebook
     case location
     case percentRaised
@@ -370,7 +370,7 @@ public final class KSRAnalytics {
     case unwatch
     case watch
 
-    public enum DiscoveryContext {
+    public enum DiscoverySortContext {
       case endingSoon
       case magic
       case newest
@@ -685,7 +685,7 @@ public final class KSRAnalytics {
   ) {
     let props = discoveryProperties(from: params)
       .withAllValuesFrom([
-        "discover_sort": sort.trackString
+        "discover_sort": sort.trackingString
       ])
       .withAllValuesFrom(
         contextProperties(
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -1545,11 +1545,11 @@ private func discoveryProperties(
   var result: [String: Any] = [:]
 
   // NB: All filters should be added here since `result["everything"]` is derived from this.
+  result["category_name"] = params.category?.name
   result["recommended"] = params.recommended
   result["social"] = params.social
   result["pwl"] = params.staffPicks
   result["watched"] = params.starred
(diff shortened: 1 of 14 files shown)


Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -194,7 +217,11 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
   }
 
   fileprivate let sortPagerSelectedSortProperty = MutableProperty<DiscoveryParams.Sort?>(nil)
+
+  // Stores last page before new page transition, default page is .magic, because it's the first page on Application Opened
+  fileprivate let previouslySortPagerSelectedSortProperty = MutableProperty<DiscoveryParams.Sort>(.magic)
```
singhhari, line 222: I see that you've used the property name above to name this but this does not read well at all. This is a very long, confusing name. I also don't think you should be defaulting given the other properties.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -194,7 +217,11 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
...
   public func sortPagerSelected(sort: DiscoveryParams.Sort) {
+    self.previouslySortPagerSelectedSortProperty.value = self.sortPagerSelectedSortProperty.value ?? .magic
```
singhhari, line 224: I've noticed within this project, a given view model input does not interact with multiple properties so I don't think this line belongs here. I think you're trying to use the value of `sortPageSelectedSortProperty` before a new value is assigned but I think there should be a cleaner way to do that.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -122,6 +122,11 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
     self.configureNavigationHeader = dataSourceConfiguredAndCurrentParams
     self.loadFilterIntoDataSource = dataSourceConfiguredAndCurrentParams
 
+    let prevSortedPage = Signal.merge(
+      self.previouslySortPagerSelectedSortProperty.signal,
+      self.previouslyTransitionedToPageProperty.signal.map { sorts[$0] }
+    ).take(first: 1)
```
singhhari, line 128: Could you clarify why you are only taking the first here?

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -155,17 +160,35 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
...
+      currentParams,
+      self.sortPagerSelectedSortProperty.signal.skipNil().skipRepeats(==),
+      prevSortedPage
+    ).observeValues { currentParams, currentSortedPage, prevSortedPage in
+      AppEnvironment
+        .current
+        .ksrAnalytics
+        .trackDiscoverySelectedSort(
+          prevSort: prevSortedPage,
+          nextSort: currentSortedPage,
```
singhhari, line 173: The parameter name and argument name represent two different things which suggests to me that one of them should be renamed.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -204,7 +231,10 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
   }
 
   fileprivate let willTransitionToPageProperty = MutableProperty<Int>(0)
+  fileprivate let previouslyTransitionedToPageProperty =
+    MutableProperty<Int>(1) // Store last page before new page transition
   public func willTransition(toPage nextPage: Int) {
+    self.previouslyTransitionedToPageProperty.value = self.willTransitionToPageProperty.value
```
singhhari, line 237: Same idea as above. I think you you should try to create a new input/output to handle the logic you are introducing.
https://github.com/kickstarter/native-docs/blob/master/inputs-outputs.md

singhhari (changes requested):

Library/ViewModels/DiscoveryViewModelTests.swift
```diff
@@ -394,96 +394,96 @@ internal final class DiscoveryViewModelTests: TestCase {
...
     )
 
     XCTAssertEqual(
-      ["popular", "popular"], self.dataLakeTrackingClient.properties(forKey: "context_type"),
+      ["popular"], self.dataLakeTrackingClient.properties(forKey: "context_type"),
       "Correct sort is tracked."
     )
     XCTAssertEqual(
-      ["popular", "popular"], self.segmentTrackingClient.properties(forKey: "context_type"),
+      ["popular"], self.segmentTrackingClient.properties(forKey: "context_type"),
```
singhhari, line 410: I think you will have to re-visit this when you re-factor but i suspect the behavior i master *should* be the same expectation since all we are doing is re-factoring existing event logic.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -657,14 +678,25 @@ public final class KSRAnalytics {
...
-  public func trackDiscoverySelectedSort(nextSort sort: DiscoveryParams.Sort, params: DiscoveryParams) {
+  public func trackDiscoverySelectedSort(
+    prevSort sort: DiscoveryParams.Sort,
+    nextSort: DiscoveryParams.Sort,
+    params: DiscoveryParams
+  ) {
     let props = discoveryProperties(from: params)
       .withAllValuesFrom([
-        "discover_sort": sort.rawValue
+        "discover_sort": sort.trackString
```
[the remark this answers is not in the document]
moderateepheezy: Yes it is, the requirement, from the analytics team is different from what exist in `discoveryProperties `

Library/Tracking/KSRAnalytics.swift
```diff
@@ -657,14 +678,25 @@ public final class KSRAnalytics {
...
       ])
+      .withAllValuesFrom(
+        contextProperties(
+          ctaContext: .discoverSort,
+          typeContext: .discovery(TrackingHelpers.discoveryContext(for: nextSort)),
+          locationContext: .discoverAdvanced
```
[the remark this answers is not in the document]
moderateepheezy: We don't have a The Sort items  in `discovery_overlay`

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: Because, there are two ways to update `prevSortedPage`, I only need the first one that emit at on page transition.

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: I'll revisit this, but the default seem to be need, when the app loaded the default sorted is magic, at the point where the function is called, `previouslySortPagerSelectedSortProperty ` can't be nil.

Library/Tracking/KSRAnalytics.swift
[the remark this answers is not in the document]
singhhari: i know. i'm saying that i see `sort` being assigned in `discoveryProperties` and I would like you to verify that this line and assignment here is necessary instead of within the method.

Library/Tracking/KSRAnalytics.swift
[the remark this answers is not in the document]
singhhari: i dont understand what you mean, can you please clarify?

Library/Tracking/TrackingHelpers.swift
```diff
@@ -11,4 +12,13 @@ final class TrackingHelpers {
       return .fixErroredPledge
     }
   }
+
+  static func discoveryContext(for sort: DiscoveryParams.Sort) -> KSRAnalytics.TypeContext.DiscoveryContext {
```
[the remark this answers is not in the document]
moderateepheezy: The `rawValues`are not identical. At least for `popular` -> `polularity`, `end_soon` -> `end_date`

Library/Tracking/KSRAnalytics.swift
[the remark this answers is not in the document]
moderateepheezy: From what I got from Leigh and Laurie, the task expects `magic`, `popular`, `newest`, and `ending soon` to be in `discovery_overlay `, but it's not. `discovery_overlay ` Is the sort options when you see when youclicked on `Recommended For You` drop down.

Library/Tracking/KSRAnalytics.swift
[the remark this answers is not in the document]
moderateepheezy: I understand, the one defined in `discoveryProperties ` get's it's value from `DiscoveryParams`, but base on the ticket, the value is wrong, because the ticket expects `discover_sort` to be the previously selected item.

Library/ViewModels/DiscoveryViewModelTests.swift
moderateepheezy: So the problem here is that, `discover_sort` means something different from what's on master, the ticket says it means previous sorted.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -657,14 +678,25 @@ public final class KSRAnalytics {
 
    - parameter sort: The new sort that was selected.
    */
-  public func trackDiscoverySelectedSort(nextSort sort: DiscoveryParams.Sort, params: DiscoveryParams) {
+  public func trackDiscoverySelectedSort(
+    prevSort sort: DiscoveryParams.Sort,
```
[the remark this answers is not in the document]
moderateepheezy: I made some changes to the naming.

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: Made some changes to this.

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: Made some changes to naming.

Library/Tracking/KSRAnalytics.swift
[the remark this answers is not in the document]
singhhari: nvm spoke with her myself. this is fine.

Library/Tracking/TrackingHelpers.swift
```diff
@@ -11,4 +12,14 @@ final class TrackingHelpers {
       return .fixErroredPledge
     }
   }
+
+  static func discoveryContext(for sort: DiscoveryParams.Sort) -> KSRAnalytics.TypeContext
```
singhhari, line 16: Mentioned this in a previous comment - I would like you to try a different approach here. We shouldn't be including a new import for functionality like this. I also think there has to be a cleaner way to map these enum values if they are nearly identical.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -126,6 +126,13 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
       .takeWhen(self.pageTransitionCompletedProperty.signal.filter(isTrue))
       .map { sorts[$0] }
 
+    let prevSortedPage = Signal.merge(
+      swipeToSort,
+      self.sortPagerSelectedSortProperty.signal.skipNil()
+    )
+    .combinePrevious(.magic)
+    .map { $0.0 }
```
singhhari, line 134: I think `$0.0` is a bit hard to read. I would opt for something that is more verbose here so it's easier to understand for someone who did not write the code.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -126,6 +126,13 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
       .takeWhen(self.pageTransitionCompletedProperty.signal.filter(isTrue))
       .map { sorts[$0] }
 
+    let prevSortedPage = Signal.merge(
+      swipeToSort,
+      self.sortPagerSelectedSortProperty.signal.skipNil()
+    )
+    .combinePrevious(.magic)
```
singhhari, line 133: Could you explain the use of `.combinePrevious(.magic)`?

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -155,17 +162,39 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
...
-      .takePairWhen(swipeToSort)
-      .observeValues {
-        AppEnvironment.current.ksrAnalytics.trackDiscoverySelectedSort(nextSort: $1, params: $0)
-      }
+    Signal.combineLatest(
+      currentParams,
+      swipeToSort,
+      prevSortedPage
+    )
+    .filter { $0.1.rawValue != $0.2.rawValue }
```
singhhari, line 187: Since we are reusing this filter logic twice here, probably want to write a private method for it.

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -155,17 +162,39 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
...
+      currentParams,
+      self.sortPagerSelectedSortProperty.signal.skipNil().skipRepeats(==),
+      prevSortedPage
+    )
+    .filter { $0.1.rawValue != $0.2.rawValue }
+    .observeValues { currentParams, currentSortedPage, prevSortedPage in
+      AppEnvironment
+        .current
+        .ksrAnalytics
+        .trackDiscoverySelectedSort(
```
singhhari, line 175: Could you explain why we are calling the event in two separate blocks with slightly different input signals?

singhhari (changes requested):
Good re-factor of the DiscoveryViewModel, it definitely reads better now.

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: `.combinePrevious(.magic)` uses scan to to get value of previously emitted and current event. I used `.magic` as the initial value because, on `Application Opened` previous sorted item should be equals current sorted, as the first sorted item is `.magic`.

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: Yeah. I understand the concern, I opted for a more function programming approach, a more verbose option would be to open up the map function or use `map(first)`, I have opted for `map(first)`

Library/ViewModels/DiscoveryViewModel.swift
moderateepheezy: I can move them in one block using `merge`, but I kind of have a fear totally changing what currently exist. I'll do that now, and you can review.

Library/Tracking/TrackingHelpers.swift
moderateepheezy: Okay, I'll think of something.\

Library/ViewModels/DiscoveryViewModelTests.swift
```diff
@@ -458,32 +458,32 @@ internal final class DiscoveryViewModelTests: TestCase {
     self.vm.inputs.sortPagerSelected(sort: .newest)
 
     XCTAssertEqual(
-      ["CTA Clicked", "CTA Clicked", "CTA Clicked"],
+      ["CTA Clicked", "CTA Clicked"],
```
singhhari, line 461: I keep seeing the number of events change from commit to commit. Could you explain what's happening? Should the older unit tests be emitting a different number of events with your changes?

singhhari (changes requested):

Library/ViewModels/DiscoveryViewModel.swift
```diff
@@ -155,17 +162,39 @@ public final class DiscoveryViewModel: DiscoveryViewModelType, DiscoveryViewMode
...
+          prevSort: prevSortedPage,
+          params: currentParams,
+          discoverySortContext: self.getDiscoverySortTypeContext(from: currentSortedPage)
+        )
+    }
+  }
+
+  /// convert DiscoveryParams.Sort to TypeContext.DiscoverySortContext
+  fileprivate func getDiscoverySortTypeContext(
```
singhhari, line 189: `private`

singhhari (changes requested):

singhhari (approved):
lgtm

PR #1396, opened 2021-03-15
[Ep-248] Page Viewed (search)

tobitech:
# 📲 What

Replace Search Page Viewed with a new event. Modify its tracking methods and unit tests.

# 🤔 Why

Convert Data Lake event to Segment


Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1197,10 +1197,24 @@ final class KSRAnalyticsTests: TestCase {
     let segmentClient = MockTrackingClient()
     let ksrAnalytics = KSRAnalytics(dataLakeClient: dataLakeClient, segmentClient: segmentClient)
 
-    ksrAnalytics.trackProjectSearchView()
+    ksrAnalytics.trackProjectSearchView(
+      params: .defaults |> DiscoveryParams.lens.query .~ "mavericks", results: 2
```
singhhari, line 1201: can you align params and results for readability?

Library/ViewModels/SearchViewModel.swift
```diff
@@ -187,8 +188,17 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
 
     // KSRAnalytics
```
singhhari, line 189: Change this to // Tracking

Library/ViewModels/SearchViewModel.swift
```diff
@@ -187,8 +188,17 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
 
-    viewWillAppearNotAnimated
-      .observeValues { AppEnvironment.current.ksrAnalytics.trackProjectSearchView() }
+    // Create SignalProducers to emit initial values
```
singhhari, line 191: No need for this line

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1149,9 +1148,23 @@ public final class KSRAnalytics {
...
-    self.track(event: ApprovedEvent.searchPageViewed.rawValue, page: .search)
+  /// Call whenever the search view is shown.
+  public func trackProjectSearchView(params: DiscoveryParams, results: Int) {
+    var props = discoveryProperties(from: params)
+
+    if let query = params.query {
+      props = props.withAllValuesFrom(["discover_search_term": query])
+    }
+
+    if results > 0 {
```
singhhari, line 1159: I don't think this conditional is necessary? If it's 0 that is probably meaningful information for insights.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1149,9 +1148,23 @@ public final class KSRAnalytics {
...
+      props = props.withAllValuesFrom(["discover_search_results_count": results])
```
singhhari, line 1160: Add this property into the `discoveryProperties` method and handle the logic elsewhere.. It was an original requirement but I was quite busy and could not do it.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -187,8 +188,17 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
+    // for projects and discovery params
+    let projectsProducer = SignalProducer(self.projects).prefix(value: [])
+    let discoveryParamsProducer = SignalProducer(requestFirstPageWith).prefix(value: initialDiscoveryParams)
+
+    discoveryParamsProducer.combineLatest(with: projectsProducer)
```
singhhari, line 196: I know this works but I haven't seen this convention used with multiple SignalProducers elsewhere in the app. Can you try to achieve this functionality with signals, perhaps combining them?

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -743,4 +743,47 @@ internal final class SearchViewModelTests: TestCase {
...
+
+    XCTAssertNil(dataLakeClientProps?["discover_search_term"] as? String)
+    XCTAssertNil(segmentClientProps?["discover_search_term"] as? String)
+
+    XCTAssertNil(dataLakeClientProps?["discover_search_results_count"] as? Int)
+    XCTAssertNil(segmentClientProps?["discover_search_results_count"] as? Int)
+  }
+
+  func testSearchPageViewed_ReturningAfterSearching() {
+    let searchProjects = (1...10).map { idx in .template |> Project.lens.id .~ (idx + 42) }
```
singhhari, line 767: Curious why + 42?

singhhari (changes requested):

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1700,7 +1715,7 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("global_nav", dataLakeClient.properties.last?["context_location"] as? String)
     XCTAssertEqual("global_nav", segmentClient.properties.last?["context_location"] as? String)
 
-    ksrAnalytics.trackProjectSearchView()
+    ksrAnalytics.trackProjectSearchView(params: .defaults, results: 0)
```
singhhari, line 1718: I would omit the results parameter (it has a default argument of nil) since this method is only checking certain context properties.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +186,27 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
 
     self.scrollToProjectRow = self.transitionedToProjectRowAndTotalProperty.signal.skipNil().map(first)
 
-    // KSRAnalytics
+    // Tracking
 
-    viewWillAppearNotAnimated
-      .observeValues { AppEnvironment.current.ksrAnalytics.trackProjectSearchView() }
+    let projectsProperty = Property(
```
singhhari, line 191: The general convention of this project is that we do not define properties unless they are explicitly defined as an output. Let's aim for a cleaner solution.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +186,27 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
+      initial: [],
```
singhhari, line 192: You probably want to try something closer to `self.projects.signal.combineLatest`. Something like that. There should be a relevant operator.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +186,27 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
+      then: self.projects
+    )
+
+    let discoveryParamsProperty = Property(
```
singhhari, line 196: Same idea here. We should remain as close to the style/convention of the project as we can.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +186,27 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
+      initial: initialDiscoveryParams,
+      then: requestFirstPageWith
+    )
+
+    Property.combineLatest(
```
singhhari, line 201: Signal.combineLatest is something we use quite a bit and could possibly be applied here.

singhhari (changes requested):

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +185,21 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
 
-    viewWillAppearNotAnimated
-      .observeValues { AppEnvironment.current.ksrAnalytics.trackProjectSearchView() }
+    Signal.combineLatest(query, searchResults)
+      .takeWhen(viewWillAppearNotAnimated)
+      .observeValues { query, results in
+        let projects = query.isEmpty ? [] : results
+        let params = .defaults |> DiscoveryParams.lens.query .~ query
+        AppEnvironment.current.ksrAnalytics
+          .trackProjectSearchView(params: params, results: projects.count)
```
singhhari, line 201: Working fine except we are only returning the paginated results instead of the total. We spoke about the details offline. I think once that is resolved we are good to go here. Fixing that will potentially allow for a refactor to remove the need for searchResults as it's own constant.

singhhari (changes requested):

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/PaginateTests.swift | indent, redundantReturn |


</p>

Library/Paginate.swift
```diff
@@ -43,6 +43,7 @@ public func paginate<Cursor, Value: Equatable, Envelope, ErrorEnvelope, RequestP
   skipRepeats: Bool = true,
   valuesFromEnvelope: @escaping ((Envelope) -> [Value]),
   cursorFromEnvelope: @escaping ((Envelope) -> Cursor),
+  statsFromEnvelope: ((Envelope) -> Int)? = nil,
```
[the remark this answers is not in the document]
singhhari: Should this be marked `@escaping` like the rest of the parameters, given it's value relies on the network request?

Library/PaginateTests.swift
```diff
@@ -11,12 +11,13 @@ final class PaginateTests: TestCase {
   let requestFromCursor: (Int) -> SignalProducer<[Int], Never> = { c in .init(value: c <= 2 ? [c] : []) }
   let valuesFromEnvelope: ([Int]) -> [Int] = id
   let cursorFromEnvelope: ([Int]) -> Int = { ($0.last ?? 0) + 1 }
+  let statsFromEnvelope: ([Int]) -> Int = { _ in return 10 }
```
singhhari, line 14: Why are you returning 10? This should probably be `$0.count`?

Library/ViewModels/SearchViewModel.swift
```diff
@@ -185,10 +186,21 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
 
     self.scrollToProjectRow = self.transitionedToProjectRowAndTotalProperty.signal.skipNil().map(first)
 
-    // KSRAnalytics
+    // Tracking
+
+    let searchResults = Signal.merge(
+      stats,
+      viewWillAppearNotAnimated.mapConst(0).take(first: 1)
```
singhhari, line 193: Can you add a brief inline comment here explaining why this needed to be done

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -779,4 +779,47 @@ internal final class SearchViewModelTests: TestCase {
...
+
+    XCTAssertEqual("", dataLakeClientProps?["discover_search_term"] as? String)
+    XCTAssertEqual("", segmentClientProps?["discover_search_term"] as? String)
+
+    XCTAssertEqual(0, dataLakeClientProps?["discover_search_results_count"] as? Int)
+    XCTAssertEqual(0, segmentClientProps?["discover_search_results_count"] as? Int)
+  }
+
+  func testSearchPageViewed_ReturningAfterSearching() {
+    let searchProjects = (1...10).map { idx in .template |> Project.lens.id .~ idx }
```
singhhari, line 803: I think you can delete this line and on the next line do `let searchResponse = DisoveryEnvelope.template` without the rest of the line and the test will have the same desired effect.

singhhari (changes requested):

Library/Paginate.swift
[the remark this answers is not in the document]
tobitech: I learnt that optional closures are already escaping, the compiler shows a warning when an optional closure is marked with `@escaping`. [More info here](https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20160905/003185.html)

Library/PaginateTests.swift
tobitech: 10 is an arbitrary number I chose. Returning `$0.count` would return the count of results for a page during pagination, but stats is supposed to be the same regardless of the requested page.

singhhari (approved):
lgtm

PR #1397, opened 2021-03-17
[Ep-246] Page Viewed discover

tobitech:
# 📲 What

Modify tracking method that sends event when user loads any of the explore tabs

# 🤔 Why

Convert Data Lake event to segment


singhhari (changes requested):
Works as expected so no issues there but we need to bump the coverage.

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -415,14 +415,9 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
...
 
-    requestFirstPageWith
-      .observeValues { params in
-        let optimizelyProps = optimizelyProperties() ?? [:]
-
-        AppEnvironment.current.ksrAnalytics.trackDiscovery(
-          params: params,
-          optimizelyProperties: optimizelyProps
-        )
+    Signal.combineLatest(requestFirstPageWith, self.viewWillAppearProperty.signal)
```
singhhari, line 418: How do you feel about something like this:
`    requestFirstPageWith.signal
      .takeWhen(self.viewWillAppearProperty.signal)
      .observeValues { params in
        AppEnvironment.current.ksrAnalytics.trackDiscovery(params: params)
      }`

Removes the _ in the observeValues block. Test it and see if it gives the same desired effect.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1648,7 +1645,7 @@ private func discoveryProperties(
     .withAllValuesFrom(parentCategoryProps ?? [:])
 
   result["everything"] = result.isEmpty
-  result["sort"] = params.sort?.rawValue
+  result["sort"] = params.sort?.trackingString
```
singhhari, line 1648: Hm why did this change?

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
tobitech: I realized the sort values that we were sending to segment was using the rawValue which wasn't correct. The ticket requires the string values held by the trackingString property. It was until I tested for other enum cases that I found out

singhhari: good catch

singhhari (approved):
lgtm

PR #1401, opened 2021-03-19
[EP-238] cta clicked discovery filter

moderateepheezy:
# 📲 What

Track an event where one of the items in discovery overlay is clicked for filter, and also trigger same event when `Saved` is clicked on `Profile` tab.

# 🤔 Why

Segment Integration.


Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -157,10 +157,32 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
...
-        AppEnvironment.current.ksrAnalytics.trackDiscoveryModalSelectedFilter(params: $0.params)
+        AppEnvironment.current.ksrAnalytics.trackDiscoveryModalSelectedFilter(
+          params: $0.params,
+          discoveryFilterContext: self.getDiscoveryFilterContextFrom($0.params),
+          locationContext: .discoverOverlay
+        )
       }
   }
 
+  private func getDiscoveryFilterContextFrom(
```
tobitech, line 168: I think you can add the "from" as first parameter label of the function, `getDiscoveryFilterContext(from discoveryParams: DiscoveryParams ...)`

moderateepheezy: I have a different opinion about this, but I'll change to your opinion as it seem conventional.

Library/ViewModels/BackerDashboardViewModel.swift
```diff
@@ -174,6 +174,18 @@ public final class BackerDashboardViewModel: BackerDashboardViewModelType, Backe
       .map { headerTopConstant, scrollViewYOffset in
         min(headerTopConstant, -scrollViewYOffset)
       }
+
+    self.currentSelectedTabProperty
+      .signal
+      .filter { $0 == .saved }
+      .observeValues { _ in
+        let params = DiscoveryParams.defaults |> DiscoveryParams.lens.starred .~ true
+        AppEnvironment.current.ksrAnalytics.trackDiscoveryModalSelectedFilter(
```
singhhari, line 183: The name of this event does not map to the location. This is not the `Discovery` tab. I would create a new method that performs the same event and re-factor the event itself so it doesn't need anything passed into it but params.

singhhari (changes requested):
I think you should go over this ticket with Laurie and verify some of the redundancies in the enum cases I referenced. Something about it doesn't seem optimal and that could be in either a misinterpretation of the task or maybe a typo on the ticket.

Library/ViewModels/DiscoveryFiltersViewModelTests.swift
```diff
@@ -433,4 +466,44 @@ internal final class DiscoveryFiltersViewModelTests: TestCase {
       "Server did not advance, categories loaded from cache."
     )
   }
+
+  /*
+   Helper for testing discoverProperties All Project Filter
+   */
+
+  private func assertDiscoveryPropertiesForAllProject(_ props: [String: Any]?) {
```
singhhari, line 474: I think you have the right idea here but this is very specific to one case and for that reason I don't think it should be a specifically defined method.

Library/ViewModels/DiscoveryFiltersViewModelTests.swift
```diff
@@ -433,4 +466,44 @@ internal final class DiscoveryFiltersViewModelTests: TestCase {
...
+    XCTAssertNil(props?["discover_category_name"] as? String)
+    XCTAssertNil(props?["discover_sort"] as? String)
+    XCTAssertNil(props?["discover_search_term"] as? String)
+  }
+
+  /*
+   Helper for testing discoverProperties for Documentary Filter
+   */
+
+  private func assertDiscoveryPropertiesForDocumentary(_ props: [String: Any]?) {
```
singhhari, line 494: Same idea as above.

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -157,10 +157,31 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
...
+          params: $0.params,
+          discoveryFilterContext: self.getDiscoveryFilterContext(from: $0.params),
+          locationContext: .discoverOverlay
+        )
       }
   }
 
+  private func getDiscoveryFilterContext(from discoveryParams: DiscoveryParams
+  ) -> KSRAnalytics.TypeContext.DiscoveryFilterContext {
+    if discoveryParams.staffPicks == true {
```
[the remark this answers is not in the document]
singhhari: Ideally we don't need this method. I think the enum to enum mapping is not good practice and we typically should not be doing that.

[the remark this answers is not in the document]
moderateepheezy: Okay, so this is actually a code I'm reusing from the project. I understand your concern about too many else-ifs, but I don't think I know another way to  resolve the booleans in a an enum type. Plus `DiscoveryParams ` is not an enum, it's an class.

Library/ViewModels/BackerDashboardViewModelTests.swift
```diff
@@ -221,4 +251,14 @@ internal final class BackerDashboardViewModelTests: TestCase {
 
     XCTAssertEqual(-500, self.vm.outputs.initialTopConstant)
   }
+
+  /*
+   Helper for testing discoverProperties Watched Filter
+   */
+
+  private func assertDiscoveryProperties(_ props: [String: Any]?) {
```
singhhari, line 259: I see why you added this but also notice we don't do this outside of the KSRAnalyticsTests file. Probably not a big deal if we delete this here because it's being tested in another file?

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -337,6 +341,23 @@ private func cache(categories: [KsApi.Category]) {
   AppEnvironment.current.cache[KSCache.ksr_discoveryFiltersCategories] = categories
 }
 
+private func getTypeContext(from discoveryParams: DiscoveryParams
```
singhhari, line 344: rename to `typeContext(from discoveryParams:)`

singhhari (changes requested):

singhhari (approved):
lgtm

PR #1403, opened 2021-03-22
[EP-323] Page Viewed (update_pledge)

moderateepheezy:
# 📲 What

Added a track event for when update pledge screen is viewed.
There are two ways to get to `Update Pledge Screen`, either from `Manage Pledge` or `Update Reward`.

# 🤔 Why

Phase 3 of segment implementation

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -84,6 +84,7 @@ public final class KSRAnalytics {
     case signup = "sign_up" // SignupViewController
     case thanks // ThanksViewController
     case twoFactorAuth = "two_factor_auth" // TwoFactorViewController
+    case updatePledgeScreen = "update_pledge" // PledgeViewController
   }
 
   /// Determines the authentication type for login or signup events.
@@ -971,11 +972,12 @@ public final class KSRAnalytics {
     )
   }
 
-  /* Call when the pledge screen is shown
+  /* Call when the pledge screen is shown, and pageContext = pledge, update, or updateReward.
 
    parameters:
    - project: the project being pledged to
    - reward: the chosen reward
+   - pageContext: The screen that's been tracked.
    - checkoutData: the `CheckoutPropertiesData` associated with the given project and reward
    - refTag: the associated RefTag for the pledge
    - cookieRefTag: The ref tag pulled from cookie storage when this project was shown.
@@ -985,6 +987,7 @@ public final class KSRAnalytics {
   public func trackCheckoutPaymentPageViewed(
     project: Project,
     reward: Reward,
+    pageContext: PageContext,
     checkoutData: CheckoutPropertiesData,
     refTag: RefTag?,
     cookieRefTag: RefTag?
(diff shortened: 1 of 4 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -84,6 +84,7 @@ public final class KSRAnalytics {
     case signup = "sign_up" // SignupViewController
     case thanks // ThanksViewController
     case twoFactorAuth = "two_factor_auth" // TwoFactorViewController
+    case updatePledgeScreen = "update_pledge" // PledgeViewController
```
singhhari, line 87: just `updatePledge`

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1808,6 +1846,21 @@ final class KSRAnalyticsTests: TestCase {
     self.assertCheckoutProperties(dataLakeClientProps)
     self.assertCheckoutProperties(segmentClientProps)
 
+    ksrAnalytics.trackCheckoutPaymentPageViewed(
+      project: .template,
+      reward: reward,
+      pageContext: .updatePledgeScreen,
+      checkoutData: KSRAnalytics.CheckoutPropertiesData.template,
```
singhhari, line 1853: can you update this and above to to `.template`

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -827,7 +827,8 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
 
     trackCheckoutPageViewData
       .observeValues { project, baseReward, rewards, selectedQuantities, refTag, additionalPledgeAmount, pledgeTotal, shippingTotal, context in
-        guard context == .pledge else { return }
+        let allowedPageContexts: [PledgeViewContext] = [.pledge, .update, .updateReward]
+        guard allowedPageContexts.contains(context) else { return }
```
singhhari, line 831: pass the context into the analytics method and perform this logic in something like a switch where you can control which method is called.

singhhari (changes requested):

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: Can you please clarify on this bit, there's only one method that needs to be called.

The `trackCheckoutPaymentPageViewed` method send exactly the same events properties to segment where the  `PledgeViewModel` `context_page` =  `.updatePledge`,  or `.checkout`, i.e If the context of `PledgeViewModel` is`.update` or `.updateReward` the context_page = `.updatePledge`, otherwise it's `.checkout`.

So is there a need to move the logic to KSRAnalytics if we are only working with 2 properties, `.checkout`, and `.updatePledge`?

singhhari: I would move it because this is an unconventional way to perform this kind of logic. In particular, `allowedPageContexts` probably does not need to be here because you can just switch on the context and perform the even conditionally if it falls under the right case.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/PledgeViewModel.swift | indent |


</p>

singhhari (approved):
lgtm

singhhari (approved):

PR #1405, opened 2021-03-24
[EP-249] Page Viewed (manage pledge)

tobitech:
# 📲 What

Implement page viewed event tracking for manage pledge screen

# 🤔 Why

This is part of phase 3 iOS Native events to send tracking data to segment

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -917,6 +917,21 @@ public final class KSRAnalytics {
     )
   }
 
+  public func trackManagePledgePageViewed(
+    project: Project,
+    reward: Reward,
+    checkoutData: CheckoutPropertiesData
+  ) {
+    let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
+      .withAllValuesFrom(checkoutProperties(from: checkoutData, and: reward))
+
+    self.track(
+      event: NewApprovedEvent.pageViewed.rawValue,
+      page: .managePledgeScreen,
+      properties: props
+    )
+  }
+
   /* Call when a reward is selected
 
    parameters:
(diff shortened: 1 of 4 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -1374,6 +1374,23 @@ public final class KSRAnalytics {
     )
   }
 
+  // MARK: - Manage Pledge Events
```
singhhari, line 1377: Can you remove this MARK and move the method up into the section called `Pledge Events`

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -909,13 +909,13 @@ internal final class ManagePledgeViewModelTests: TestCase {
 
       self.scheduler.advance()
 
-      XCTAssertEqual([], self.dataLakeTrackingClient.events)
-      XCTAssertEqual([], self.segmentTrackingClient.events)
+      XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)
```
singhhari, line 912: Could you explain why this changed from [] to Page Viewed?

singhhari (changes requested):
I think you have the wrong ticket number for this branch? Left a cpl of small comments.

Library/ViewModels/ManagePledgeViewModelTests.swift
tobitech: Previously there was no event tracked on that view model up to the point where the fix button is tapped. But now that I added a page viewed event, the `segmentTrackingClient.events` isn't empty anymore at that point, so it has to be appended.

singhhari (approved):
lgtm!

PR #1406, opened 2021-03-24
[EP-319] Page Viewed (change_payment)

singhhari:
# 📲 What

Extending the event tracking functionality of our `PledgeViewModel` to handle a `Page Viewed` event specifically for when the payment method is changed.

# 🤔 Why

Phase 3 Segment implementation

### DIFF
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -734,11 +734,10 @@ final class KSRAnalyticsTests: TestCase {
     self.assertProjectProperties(segmentClient.properties.last)
   }
 
-  func testTrackCheckoutPaymentMethodViewed() {
+  func testTrackCheckoutPaymentMethodViewed_PledgeViewContext_Pledge() {
     let dataLakeClient = MockTrackingClient()
     let segmentClient = MockTrackingClient()
     let ksrAnalytics = KSRAnalytics(dataLakeClient: dataLakeClient, segmentClient: segmentClient)
-
     let reward = Reward.template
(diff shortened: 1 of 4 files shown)


Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -734,22 +734,21 @@ final class KSRAnalyticsTests: TestCase {
     self.assertProjectProperties(segmentClient.properties.last)
   }
 
-  func testTrackCheckoutPaymentMethodViewed() {
+  func testTrackCheckoutPaymentMethodViewed_PledgeViewContext_Pledge() {
```
moderateepheezy, line 737: So there's a similar test inside `testContextLocationProperties()`, maybe we delete them in favour on the methods you created?

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -770,22 +769,56 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("activity", segmentClientProps?["session_ref_tag"] as? String)
   }
 
-  func testTrackUpdatePledgeScreenViewed() {
+  func testTrackUpdatePledgeScreenViewed_PledgeViewContext_Update() {
```
moderateepheezy, line 772: Same as this.

Library/Tracking/KSRAnalyticsTests.swift
singhhari: Ah I see. Pushed up a commit with the updates.

tobitech (approved):

moderateepheezy (approved):

PR #1407, opened 2021-03-24
[EP-316] CTA Clicked(Creator Details)

moderateepheezy:
# 📲 What

Added a new event for when a creator's name is clicked

# 🤔 Why

Phase 4 Segment Integration

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -1391,7 +1391,8 @@ public final class KSRAnalytics {
 
   /**
    Call  when a user clicks the creator's name on a project.
-   - parameter project: The project the creator details is clicked from.
+
+   - parameter project: The project the creator's name is clicked from.
    */
 
   public func trackGotoCreatorDetailsClicked(project: Project) {
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -1391,7 +1391,7 @@ public final class KSRAnalytics {
 
   /**
    Call  when a user clicks the creator's name on a project.
-   - parameter project: The project being watched
+   - parameter project: The project the creator details is clicked from.
    */
 
   public func trackGotoCreatorDetailsClicked(project: Project) {
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -1390,7 +1390,7 @@ public final class KSRAnalytics {
   }
 
   /**
-   Call  when a user clicks the creator's name on a project.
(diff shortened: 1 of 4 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -1389,6 +1389,22 @@ public final class KSRAnalytics {
...
 
+  /**
+   Call when a user clicks creator's name on a project.
+
+   - parameter project: The project the creator's name is clicked from.
+   */
+
+  public func trackGotoCreatorDetailsClicked(project: Project) {
+    let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
+      .withAllValuesFrom(contextProperties(ctaContext: .creatorDetails, page: .projectPage))
```
singhhari, line 1400: I don't see a `context_page` defined in the ticket?

Library/ViewModels/ProjectPamphletMainCellViewModel.swift
```diff
@@ -255,6 +255,10 @@ public final class ProjectPamphletMainCellViewModel: ProjectPamphletMainCellView
 
         AppEnvironment.current.optimizelyClient?.track(eventName: "Campaign Details Button Clicked")
       }
+
+    self.notifyDelegateToGoToCreator.observeValues { project in
```
singhhari, line 259: Right idea but I think it would better to observeValues on the signal of `creatorButtonTappedProperty` instead.

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
moderateepheezy: Well, there was another ticket that doesn't have a context_page as well, I asked Laurie and she said context_page should always be page the event was triggered from.

Library/ViewModels/ProjectPamphletMainCellViewModel.swift
moderateepheezy: So the tickets needs project, creating a new Signal for `creatorButtonTappedProperty `, and `project` would not be different from `notifyDelegateToGoToCreator`. So `notifyDelegateToGoToCreator` is essentially what I need.

singhhari: I see what you mean.

Library/Tracking/KSRAnalytics.swift
singhhari: Ok can you leave a comment on the ticket indicating you are adding this context_page property so it is documented somewhere. Tag Laurie for visibility.

singhhari (approved):

PR #1411, opened 2021-03-29
[EP-400] QA Phase 1 Add properties

tobitech:
# 📲 What

Add some new properties to User, Session and Project property groups

# 🤔 Why

Event Project Phase 1 QA additions

### DIFF
--- a/KsApi/models/templates/ProjectTemplates.swift
+++ b/KsApi/models/templates/ProjectTemplates.swift
@@ -15,7 +15,7 @@ extension Project {
     blurb: "A fun project.",
     category: .template,
     country: .us,
-    creator: User.template |> \.stats.createdProjectsCount .~ 1,
+    creator: User.template |> \.stats.createdProjectsCount .~ 2,
     memberData: Project.MemberData(
       lastUpdatePublishedAt: nil,
       permissions: [],
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -67,7 +67,7 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("en", dataLakeClientProperties?["session_display_language"] as? String)
     XCTAssertEqual("GB", dataLakeClientProperties?["session_country"] as? String)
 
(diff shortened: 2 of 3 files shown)


Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -2376,6 +2379,19 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(true, props?["discover_recommended"] as? Bool)
     XCTAssertEqual("recs_home", props?["discover_ref_tag"] as? String)
   }
+
+  /*
+   Helper to test logged in User properties from a template User
+   */
+  func assertUserProperties(_ props: [String: Any]?) {
```
moderateepheezy, line 2386: So since it's just in one function that you need this properties, you can just add it directly to the function and duplicate for DataLake or Segment. I think it would be evident when we are removing DataLake

moderateepheezy (changes requested):

KsApi/models/templates/ProjectTemplates.swift
```diff
@@ -15,7 +15,7 @@ extension Project {
     blurb: "A fun project.",
     category: .template,
     country: .us,
-    creator: User.template |> \.stats.createdProjectsCount .~ 1,
+    creator: User.template |> \.stats.createdProjectsCount .~ 2,
```
singhhari, line 18: Could you explain why you updated the template?

Library/Tracking/KSRAnalyticsTests.swift
singhhari: Yeah I think we can avoid writing a method for this since we are only asserting user properties in one particular place.

singhhari (changes requested):
I'm not seeing `checkout_reward_has_items`?

KsApi/models/templates/ProjectTemplates.swift
tobitech: This was changed so that I could check `isRepeatCreator` to be true. That property requires that `createdProjectsCount` be greater than 1

singhhari (approved):
lgtm

PR #1412, opened 2021-03-29
[EP-402] QA phase 1 fix time properties

moderateepheezy:
# 📲 What

Add a new extension that converts TimeInterval to ISO 8601 Date and Time  Format.
Update `checkout_reward_estimated_delivery_on `, `project_deadline `, and `project_launched_at ` to send ISO 8601 Date Time format to Segment. 

# 🤔 Why

Based on the feedback from QA Phase 1 event tasks, `checkout_reward_estimated_delivery_on `, `project_deadline `, and `project_launched_at ` are sending their values as `TimeInterval` .

### DIFF
--- a/Library/TimeInterval+ISO8601DateTests.swift
+++ b/Library/TimeInterval+ISO8601DateTests.swift
@@ -1,0 +1,11 @@
+@testable import Library
+import XCTest
+
+class TimeInterval_ISO8601DateTests: XCTestCase {
+  func testConvertToISO8601DateTimeString_FromTimeInterval() {
+    let timeInterval = 1_506_897_315.0
+
+    let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
+    XCTAssertEqual(iSO8601DateTimeString, "2017-10-01T22:35:15Z")
+  }
+}
--- a/Library/TimeInterval+ISO8601DateTests.swift
+++ b/Library/TimeInterval+ISO8601DateTests.swift
@@ -8,11 +8,11 @@ class TimeInterval_ISO8601DateTests: XCTestCase {
     let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
     XCTAssertEqual(iSO8601DateTimeString, "2017-10-01T22:35:15Z")
   }
-  
+
   func testToISO8601DateTimeString_Future() {
-    let timeInterval = 1_680_048_000.0
+    let timeInterval = 1_680_078_655.0
 
     let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
-    XCTAssertEqual(iSO8601DateTimeString, "2023-03-29T00:00:00Z")
+    XCTAssertEqual(iSO8601DateTimeString, "2023-03-29T08:30:55Z")
   }
 }
--- a/Library/TimeInterval+ISO8601DateTests.swift
+++ b/Library/TimeInterval+ISO8601DateTests.swift
@@ -5,14 +5,12 @@ class TimeInterval_ISO8601DateTests: XCTestCase {
   func testToISO8601DateTimeString() {
     let timeInterval = 1_506_897_315.0
 
-    let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
-    XCTAssertEqual(iSO8601DateTimeString, "2017-10-01T22:35:15Z")
+    XCTAssertEqual(timeInterval.toISO8601DateTimeString(), "2017-10-01T22:35:15Z")
   }
 
   func testToISO8601DateTimeString_Future() {
     let timeInterval = 1_680_078_655.0
 
-    let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
-    XCTAssertEqual(iSO8601DateTimeString, "2023-03-29T08:30:55Z")
+    XCTAssertEqual(timeInterval.toISO8601DateTimeString(), "2023-03-29T08:30:55Z")
   }
 }
--- a/Library/TimeInterval+ISO8601DateTests.swift
+++ b/Library/TimeInterval+ISO8601DateTests.swift
@@ -2,10 +2,17 @@
 import XCTest
 
 class TimeInterval_ISO8601DateTests: XCTestCase {
-  func testConvertToISO8601DateTimeString_FromTimeInterval() {
+  func testToISO8601DateTimeString() {
     let timeInterval = 1_506_897_315.0
 
     let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
     XCTAssertEqual(iSO8601DateTimeString, "2017-10-01T22:35:15Z")
+  }
+  
+  func testToISO8601DateTimeString_Future() {
+    let timeInterval = 1_680_048_000.0
+
+    let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
+    XCTAssertEqual(iSO8601DateTimeString, "2023-03-29T00:00:00Z")
   }
 }
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -262,8 +262,14 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(project.creator.id, dataLakeClientProperties?["project_creator_uid"] as? Int)
     XCTAssertEqual(24 * 15, dataLakeClientProperties?["project_hours_remaining"] as? Int)
     XCTAssertEqual(30, dataLakeClientProperties?["project_duration"] as? Int)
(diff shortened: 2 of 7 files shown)


Library/TimeInterval+ISO8601DateTests.swift
```diff
@@ -0,0 +1,11 @@
+@testable import Library
+import XCTest
+
+class TimeInterval_ISO8601DateTests: XCTestCase {
+  func testConvertToISO8601DateTimeString_FromTimeInterval() {
```
singhhari, line 5: `testToISO8601DateTimeString()`

singhhari: Also can you add a test case here? We should test the past and the future.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -262,8 +262,14 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(project.creator.id, dataLakeClientProperties?["project_creator_uid"] as? Int)
     XCTAssertEqual(24 * 15, dataLakeClientProperties?["project_hours_remaining"] as? Int)
     XCTAssertEqual(30, dataLakeClientProperties?["project_duration"] as? Int)
-    XCTAssertEqual(1_476_657_315, dataLakeClientProperties?["project_deadline"] as? Double)
-    XCTAssertEqual(1_474_065_315, dataLakeClientProperties?["project_launched_at"] as? Double)
+    XCTAssertEqual(
+      1_476_657_315.toISO8601DateTimeString(),
```
singhhari, line 266: Instead of calling the function, I think the assertion should just be the literal of the expected value.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -301,8 +307,14 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(project.creator.id, segmentClientProperties?["project_creator_uid"] as? Int)
     XCTAssertEqual(24 * 15, segmentClientProperties?["project_hours_remaining"] as? Int)
     XCTAssertEqual(30, segmentClientProperties?["project_duration"] as? Int)
-    XCTAssertEqual(1_476_657_315, segmentClientProperties?["project_deadline"] as? Double)
-    XCTAssertEqual(1_474_065_315, segmentClientProperties?["project_launched_at"] as? Double)
+    XCTAssertEqual(
+      1_476_657_315.toISO8601DateTimeString(),
```
singhhari, line 311: Same here.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -2309,8 +2321,8 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(1, props?["project_creator_uid"] as? Int)
     XCTAssertEqual(24 * 15, props?["project_hours_remaining"] as? Int)
     XCTAssertEqual(30, props?["project_duration"] as? Int)
-    XCTAssertEqual(1_476_657_315, props?["project_deadline"] as? Double)
-    XCTAssertEqual(1_474_065_315, props?["project_launched_at"] as? Double)
+    XCTAssertEqual(1_476_657_315.toISO8601DateTimeString(), props?["project_deadline"] as? String)
```
singhhari, line 2324: Same.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -5862,8 +5862,8 @@ final class PledgeViewModelTests: TestCase {
       dataLakeTrackingClientProps?["checkout_user_has_eligible_stored_apple_pay_card"] as? Bool
     )
     XCTAssertEqual(
-      1_506_897_315.0,
-      dataLakeTrackingClientProps?["checkout_reward_estimated_delivery_on"] as? TimeInterval
+      1_506_897_315.0.toISO8601DateTimeString(),
```
singhhari, line 5865: Same.

Library/ViewModels/ThanksViewModelTests.swift
```diff
@@ -408,8 +408,8 @@ final class ThanksViewModelTests: TestCase {
     )
     XCTAssertEqual("10.00", dataLakeTrackingClientProps?["checkout_shipping_amount_usd"] as? String)
     XCTAssertEqual(
-      12_345_678,
-      dataLakeTrackingClientProps?["checkout_reward_estimated_delivery_on"] as? TimeInterval
+      12_345_678.toISO8601DateTimeString(),
```
singhhari, line 411: Same.

Library/ViewModels/ThanksViewModelTests.swift
```diff
@@ -425,7 +425,10 @@ final class ThanksViewModelTests: TestCase {
     XCTAssertEqual(true, segmentClientProps?["checkout_reward_shipping_enabled"] as? Bool)
     XCTAssertEqual(true, segmentClientProps?["checkout_user_has_eligible_stored_apple_pay_card"] as? Bool)
     XCTAssertEqual("10.00", segmentClientProps?["checkout_shipping_amount_usd"] as? String)
-    XCTAssertEqual(12_345_678, segmentClientProps?["checkout_reward_estimated_delivery_on"] as? TimeInterval)
+    XCTAssertEqual(
+      12_345_678.toISO8601DateTimeString(),
```
singhhari, line 429: Same.

singhhari (changes requested):
Looks like the Segment debugger is getting a value of NSDate(dateWithTimeIntervalSince1970:) as its value instead of the expected ISO8601.

moderateepheezy:
> Looks like the Segment debugger is getting a value of NSDate(dateWithTimeIntervalSince1970:) as its value instead of the expected ISO8601.

If you check the Raw value you'll see the actual value been sent. That's the way segment works from my observation. You can verify that from Android and Web as well.

Library/Tracking/KSRAnalyticsTests.swift
moderateepheezy: While I agree with you that this can be the literal value. `1_476_657_315.toISO8601DateTimeString()` , because I considered it as well. I'm of the opinion that, the true test of `toISO8601DateTimeString()` is to do the conversion this way instead of literal value. The test is undergoing the same definition for the property as it's done in KSRAnalytics. 

This is my opinion, let me know what you think.

Library/TimeInterval+ISO8601DateTests.swift
moderateepheezy: So the function essentially converts TimeInterval to Date String, either present, past or current as `Date(timeIntervalSince1970:)` will always give you date from `TimeInterval` irrespective. I can add more of those test regardless, just wanted you to know that.

singhhari: Yup sounds good. Just want to test those distinct conditions. I think it will bump test coverage as well, which is nice.

Library/Tracking/KSRAnalyticsTests.swift
singhhari: What you're saying definitely makes sense. My concern is mostly a readability one. This test shouldn't be verifying if the extension works (you wrote some specific tests that do that) but instead should have a clear, readable expectation in the form of the output string.

singhhari:
> > Looks like the Segment debugger is getting a value of NSDate(dateWithTimeIntervalSince1970:) as its value instead of the expected ISO8601.
> 
> Pretty gives you like and object log of how the properties are been sent.
> 
> If you check the Raw value you'll see the actual value been sent. That's the way segment works from my observation. You can verify that from Android and Web as well.

Hmm good point, now I notice that. Can you let Laurie know that is the case so there is no issue when she QAs later on?

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/TimeInterval+ISO8601DateTests.swift | indent |


</p>

Library/TimeInterval+ISO8601DateTests.swift
```diff
@@ -2,10 +2,17 @@
...
 
     let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
     XCTAssertEqual(iSO8601DateTimeString, "2017-10-01T22:35:15Z")
   }
+  
+  func testToISO8601DateTimeString_Future() {
+    let timeInterval = 1_680_048_000.0
+
+    let iSO8601DateTimeString = timeInterval.toISO8601DateTimeString()
+    XCTAssertEqual(iSO8601DateTimeString, "2023-03-29T00:00:00Z")
```
singhhari: just had a second thought about this. no need to declare the constant, just drop timeInterval.toISO8601...() directly into line 9 and 16.

singhhari (changes requested):

singhhari (approved):

PR #1413, opened 2021-03-30
[EP-403] Fix Conversion Properties

tobitech:
# 📲 What

Fix conversion for `goal_usd` and `current_amount_pledged_usd` properties by deriving their values from project properties that use `fx_rate` for conversion rather than `static_usd_rate`

# 🤔 Why

Part of phase_1 Segment events QA

### DIFF
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -253,8 +253,8 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(project.country.currencyCode, dataLakeClientProperties?["project_currency"] as? String)
     XCTAssertEqual(project.id, dataLakeClientProperties?["project_pid"] as? Int)
     XCTAssertEqual(
-      project.stats.fundingProgress,
-      dataLakeClientProperties?["project_percent_raised"] as? Float
+      project.stats.percentFunded,
+      dataLakeClientProperties?["project_percent_raised"] as? Int
     )
     XCTAssertEqual(project.category.name, dataLakeClientProperties?["project_subcategory"] as? String)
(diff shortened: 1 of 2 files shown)


Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -2340,8 +2340,8 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("2016-09-16T22:35:15Z", props?["project_launched_at"] as? String)
     XCTAssertEqual("live", props?["project_state"] as? String)
     XCTAssertEqual(1_000, props?["project_current_pledge_amount"] as? Int)
-    XCTAssertEqual(1_000, props?["project_current_amount_pledged_usd"] as? Float)
-    XCTAssertEqual(2_000, props?["project_goal_usd"] as? Float)
+    XCTAssertEqual(2_000, props?["project_current_amount_pledged_usd"] as? Int)
```
singhhari, line 2343: The changes in the values of these two properties is quite significant. Can you explain?

singhhari (changes requested):
Can we find a way to bump the coverage back to even?

Library/Tracking/KSRAnalyticsTests.swift
tobitech: If you check the stats property of the `Project.template` which `Project.Stats.template`, you will notice that `convertedPledgedAmount` is 2_000 which is the new value use for `project_current_amount_pledged_usd` property.
For `goal_usd` the new value is `goalCurrentCurrency` which is the product of `currentCurrencyRate` and `goal` amount. That will give you 3_000.

moderateepheezy (approved):

singhhari (approved):

PR #1416, opened 2021-04-01
[EP-439] QA Phase 2 - Properties to remove

moderateepheezy:
# 📲 What

Remove the following properties from segment.

- session_device_distinct_id
- session_device_manufacturer
- session_device_model
- session_os_version
- session_screen_width
- session_user_agent

# 🤔 Why

Based on feedback from QA on Phase 2 tickets, some Segment event properties aren't needed.

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -9,7 +9,6 @@ public final class KSRAnalytics {
   private let dataLakeClient: TrackingClientType
   internal private(set) var config: Config?
   private let device: UIDeviceType
-  private let distinctId: String
   internal private(set) var loggedInUser: User? {
     didSet {
       self.identify(self.loggedInUser)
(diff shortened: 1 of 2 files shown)


tobitech (approved):

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1503,10 +1503,7 @@ public final class KSRAnalytics {
     props["country"] = self.config?.countryCode
     props["display_language"] = AppEnvironment.current.language.rawValue
     props["device_type"] = self.device.deviceType
-    props["device_manufacturer"] = "Apple"
-    props["device_model"] = KSRAnalytics.deviceModel
     props["device_orientation"] = self.deviceOrientation
-    props["device_distinct_id"] = self.distinctId
```
singhhari, line 1509: Doesn't look like we are using distinctId for anything else in the file. I would delete this and double check everything is compiling fine.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1503,10 +1503,7 @@ public final class KSRAnalytics {
     props["country"] = self.config?.countryCode
     props["display_language"] = AppEnvironment.current.language.rawValue
     props["device_type"] = self.device.deviceType
-    props["device_manufacturer"] = "Apple"
-    props["device_model"] = KSRAnalytics.deviceModel
```
singhhari, line 1507: Doesn't look like we are using deviceModel for anything else in the file. I would delete this and double check everything is compiling fine.

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
moderateepheezy: Thanks, I missed that.

singhhari (approved):
lgtm

PR #1421, opened 2021-04-07
[EP-426] CTA Clicked (pledge_submit)

moderateepheezy:
# 📲 What

Add event track for when Apple Pay button is clicked to submit a pledge.

# 🤔 Why

Segment Integration

### DIFF
--- a/Library/ViewModels/PledgeViewModel.swift
+++ b/Library/ViewModels/PledgeViewModel.swift
@@ -860,38 +860,75 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         )
       }
 
-    Signal
-      .combineLatest(
-        createBackingData,
-        baseReward,
-        additionalPledgeAmount,
-        allRewardsShippingTotal
+    // Emmits value when Pledge Button is clicked.
+
+    Signal.combineLatest(
+      createBackingData,
+      baseReward,
+      additionalPledgeAmount,
+      allRewardsShippingTotal
+    )
+    .takeWhen(createButtonTapped)
+    .map { data, baseReward, additionalPledgeAmount, allRewardsShippingTotal in
+
+      let checkoutData = checkoutProperties(
+        from: data.project,
+        baseReward: baseReward,
+        addOnRewards: data.rewards,
+        selectedQuantities: data.selectedQuantities,
+        additionalPledgeAmount: additionalPledgeAmount,
+        pledgeTotal: data.pledgeTotal,
+        shippingTotal: allRewardsShippingTotal,
+        checkoutId: nil,
+        isApplePay: false
       )
-      .takeWhen(createButtonTapped)
-      .map { data, baseReward, additionalPledgeAmount, allRewardsShippingTotal in
 
-        let checkoutData = checkoutProperties(
-          from: data.project,
-          baseReward: baseReward,
-          addOnRewards: data.rewards,
-          selectedQuantities: data.selectedQuantities,
-          additionalPledgeAmount: additionalPledgeAmount,
-          pledgeTotal: data.pledgeTotal,
-          shippingTotal: allRewardsShippingTotal,
-          checkoutId: nil,
-          isApplePay: false
-        )
+      return (data.project, baseReward, data.refTag, checkoutData)
+    }
+    .observeValues { project, reward, refTag, checkoutData in
+      AppEnvironment.current.ksrAnalytics.trackPledgeSubmitButtonClicked(
+        project: project,
+        reward: reward,
+        typeContext: .creditCard,
+        checkoutData: checkoutData,
+        refTag: refTag
+      )
+    }
 
-        return (data.project, baseReward, data.refTag, checkoutData)
-      }
-      .observeValues { project, reward, refTag, checkoutData in
-        AppEnvironment.current.ksrAnalytics.trackPledgeSubmitButtonClicked(
-          project: project,
-          reward: reward,
-          checkoutData: checkoutData,
-          refTag: refTag
-        )
-      }
+    // Emmits value when Apple Pay Button is clicked.
+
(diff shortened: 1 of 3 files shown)


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,38 +860,75 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         )
       }
 
-    Signal
-      .combineLatest(
-        createBackingData,
-        baseReward,
-        additionalPledgeAmount,
-        allRewardsShippingTotal
+    // Emmits value when Pledge Button is clicked.
```
singhhari, line 863: Typo. `Emits`

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,38 +860,75 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+
+    Signal.combineLatest(
+      createBackingData,
+      baseReward,
+      additionalPledgeAmount,
+      allRewardsShippingTotal
+    )
```
singhhari, line 870: I think we can assign this to a constant called `eventSignal` or something similar because we are reusing it twice here.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,38 +860,75 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
-      }
-      .observeValues { project, reward, refTag, checkoutData in
-        AppEnvironment.current.ksrAnalytics.trackPledgeSubmitButtonClicked(
-          project: project,
-          reward: reward,
-          checkoutData: checkoutData,
-          refTag: refTag
-        )
-      }
+    // Emmits value when Apple Pay Button is clicked.
```
singhhari, line 898: Typo. `Emits`

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,38 +860,75 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+    // Emmits value when Pledge Button is clicked.
+
+    Signal.combineLatest(
+      createBackingData,
+      baseReward,
+      additionalPledgeAmount,
+      allRewardsShippingTotal
+    )
+    .takeWhen(createButtonTapped)
+    .map { data, baseReward, additionalPledgeAmount, allRewardsShippingTotal in
```
singhhari, line 872: Is there a way to re-factor the map so we aren't repeating it word-for-word twice here as well?

singhhari (changes requested):

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: Not really, because a property on the `isApplePay` field is different in both maps, if we had Lenses for CheckoutPropertiesData, maybe we can manipulate things a bit.

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: That works, I'll add that change. Thanks.

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: I did a bit of refactor to CheckoutPropertiesData in the map so it can be reused, but in my opinion it's not necessary since we only need it in those two places.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,75 +860,65 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         )
       }
 
-    // Emmits value when Pledge Button is clicked.
+    // Signals events needed for when Pledge Button or Apple Pay Button is clicked.
```
singhhari, line 863: Maybe just change this to `Tracking` like we do in all the other files

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -860,75 +860,65 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
-        baseReward: baseReward,
-        addOnRewards: data.rewards,
-        selectedQuantities: data.selectedQuantities,
-        additionalPledgeAmount: additionalPledgeAmount,
-        pledgeTotal: data.pledgeTotal,
-        shippingTotal: allRewardsShippingTotal,
-        checkoutId: nil,
-        isApplePay: false
-      )
+    // Emits value when Pledge Button is clicked.
```
singhhari, line 872: Can probably delete this if we add the `Tracking` comment at the top

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -1194,3 +1184,23 @@ private func pledgeAmountSummaryViewData(
     shippingAmountHidden: !shippingViewsHidden
   )
 }
+
+private func checkoutPropertiesData(
```
singhhari, line 1188: Let's delete this helper method. I feel like we have something similar in a few different files and it's not very intuitive.

singhhari (changes requested):

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: There's a `Tracking` at the top, I can just remove this comment entirely.

Library/ViewModels/PledgeViewModel.swift
moderateepheezy: I agree with you, that's why I said in my previous comment that it's not necessary, I thought we can give it a try following your comment.

singhhari (approved):
lgtm

PR #1422, opened 2021-04-08
[EP-410] QA Property Fixes type changes

singhhari:
# 📲 What

We had a few properties that needed some changes to their type, specifically to reflect a rounded two-digit decimal.

Properties to fix:

`checkout_add_ons_minimum_usd`
`checkout_shipping_amount_usd`
`project_goal_usd`
`project_current_amount_pledged_usd`

# 🤔 Why

This is part of phase one of our Segment implementation.

Screen Shot 2021-04-08 at 7 35 04 PM


KsApi/models/Project.swift
```diff
@@ -65,7 +65,7 @@ public struct Project {
   public struct Stats {
     public var backersCount: Int
     public var commentsCount: Int?
-    public var convertedPledgedAmount: Int?
+    public var convertedPledgedAmount: Double?
```
moderateepheezy, line 68: I think we can use `Float` here since we will still need to `round` the value. Float has lesser precision and can be "faster" to read.

KsApi/models/ProjectTests.swift
```diff
@@ -213,6 +213,14 @@ final class ProjectTests: XCTestCase {
     XCTAssertEqual(2_000, project.stats.pledgedUsd)
   }
 
+  func testGoalCurrentCurrency() {
```
moderateepheezy, line 216: I'm of the opinion that we can make this function test a value that has precision, so we Cana true test of the function and know it actually a `rounded` value.

KsApi/models/Project.swift
singhhari: Updated.

moderateepheezy (approved):

tobitech (approved):

PR #1423, opened 2021-04-09
[EP-467] user_created_projects_count login 

moderateepheezy:
# 📲 What

Guard against sending `User Properties` to segment when user is not logged in.

# 🤔 Why
user_created_projects_count still get sent to Segment when user is not logged in.

Screenshot 2021-04-09 at 16 16 53


tobitech (approved):

singhhari (approved):
lgtm

PR #1425, opened 2021-04-12
[EP-459] QA CTA Clicked (discover) phase two

singhhari:
# 📲 What

- Updating some properties related to the `Explore projects` button on the activity tab.
- Sending an event for when the `Home` tab is selected from the tab bar navigation.

## Expected Result:
Event: CTA Clicked
Property Groups: Context Properties, Session Properties, User Properties
context_cta = discover
context_page = activity_feed
context_location = [null]

## Expected Result:
Event: CTA Clicked
Property Groups: Context Properties, Session Properties, User Properties
context_cta = discover
context_location = global_nav

# 🤔 Why

This is a refinement based on QA feedback from our insights team. The point of these properties is to draw a very specific context to each event. This PR includes updates to the context properties, specifically.
Screen Shot 2021-04-12 at 5 10 28 PM
Screen Shot 2021-04-12 at 5 09 33 PM

### DIFF
--- a/Library/ViewModels/RootViewModelTests.swift
+++ b/Library/ViewModels/RootViewModelTests.swift
@@ -490,28 +490,28 @@ final class RootViewModelTests: TestCase {
     self.vm.inputs.didSelect(index: 0)
 
     self.selectedIndex.assertValues([0, 1, 0], "Selects index immediately.")
-    XCTAssertEqual(["Tab Bar Clicked", "Tab Bar Clicked"], self.dataLakeTrackingClient.events)
+    XCTAssertEqual(["Tab Bar Clicked", "CTA Clicked"], self.dataLakeTrackingClient.events)
     XCTAssertEqual(
-      ["activity", "discovery"],
+      ["activity", nil],
       self.dataLakeTrackingClient.properties(forKey: "context_tab_bar_label")
     )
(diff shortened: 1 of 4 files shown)


Library/ViewModels/RootViewModelTests.swift
```diff
@@ -490,28 +490,28 @@ final class RootViewModelTests: TestCase {
...
     )
 
     self.vm.inputs.didSelect(index: 10)
 
     self.selectedIndex.assertValues([0, 1, 0, 3], "Selects index immediately.")
-    XCTAssertEqual(["Tab Bar Clicked", "Tab Bar Clicked"], self.dataLakeTrackingClient.events)
+    XCTAssertEqual(["Tab Bar Clicked", "CTA Clicked"], self.dataLakeTrackingClient.events)
     XCTAssertEqual(
-      ["activity", "discovery"],
+      ["activity", nil],
```
moderateepheezy, line 509: I guess the reason why it's nil here is because we've removed `discovery` from `context_tab_bar_label`

moderateepheezy (approved):

Library/ViewModels/RootViewModelTests.swift
singhhari: Ah so the real reason is because prior to this PR, we were handling the selection of the `Explore/Discovery` tab in a default case which set the `context_tab_bar_label` value. It is not in the spec for the newly defined event though.

PR #1426, opened 2021-04-13
[EP-460] Page Viewed (search) Search Terms

tobitech:
# 📲 What

Send Page Viewed event for search when user types a search term and a response is returned from the API whether that response contains some results or not.

# 🤔 Why

Search events were previously not sent to Segment when user types a search term.

# 👀 See

[iOS: Page Viewed (search) Search Terms](https://kickstarter.atlassian.net/browse/EP-460?atlOrigin=eyJpIjoiY2Q1ZWIzMGZkNGIwNGM2MWE2ZDg5OTA5MjA5MzZhZjEiLCJwIjoiaiJ9)

Screenshot 2021-04-13 at 20 13 46


Library/ViewModels/SearchViewModel.swift
```diff
@@ -212,27 +220,16 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
-    self.goToProject = Signal.combineLatest(self.projects, query)
-      .takePairWhen(self.tappedProjectProperty.signal.skipNil())
-      .map { projectsAndQuery, tappedProject in
-        let (projects, query) = projectsAndQuery
-
-        return (tappedProject, projects, refTag(query: query, projects: projects, project: tappedProject))
-      }
-
-    // Tracking
+    let sendEventForSearchResults = Signal.merge(
```
singhhari, line 223: I would consider renaming this given the name of the constant does not reflect its actual value. Seems like this is the searchResultsCount, only emits when a new query is searched for but does not actually fire the event.

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -508,8 +508,8 @@ internal final class SearchViewModelTests: TestCase {
       )
       projects.assertLastValue(popularProjects, "Brings back popular projects immediately.")
 
-      XCTAssertEqual(["Page Viewed"], self.dataLakeTrackingClient.events)
-      XCTAssertEqual(["Page Viewed"], self.segmentTrackingClient.events)
+      XCTAssertEqual(["Page Viewed", "Page Viewed"], self.dataLakeTrackingClient.events)
```
singhhari, line 511: Why is there an additional Page Viewed event here now?

singhhari (changes requested):
Needs a little coverage bump. Maybe writing tests for each of the four situations described on the ticket will do the trick.

Library/ViewModels/SearchViewModel.swift
```diff
@@ -212,27 +220,18 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
-      .observeValues { query, params, hasResults in
-        AppEnvironment.current.ksrAnalytics.trackSearchResults(
-          query: query,
-          params: params,
-          refTag: .search,
-          hasResults: hasResults
-        )
+      .observeValues { _, params, stats in
+        AppEnvironment.current.ksrAnalytics
+          .trackProjectSearchView(params: params, results: stats)
```
singhhari, line 234: I'm a little confused here. We're tracking the same event here and on line 211 but the logic is different. Is 211 necessary? Also the `searchResultsCount` property is very similar to `searchResults` above. Is this something we can simplify/re-factor?

tobitech: Yes there are two different logic and we are meant to use one event for both scenarios. They were previously tracked with two separate events. Here are the current two scenarios `Page Viewed (search)` (when user visits the page) and `Page Viewed Search Term` (when user enters a search term and gets a response).

Library/ViewModels/SearchViewModelTests.swift
tobitech: This has been fixed

singhhari (approved):
lgtm

PR #1427, opened 2021-04-14
[NT-1834] PerimeterX Refactor/Test Coverage Improvements

justinswart:
# 📲 What

This PR refactors our implementation of PerimeterX with the aim to simplify and improve test coverage. The changes made are:

- Moved all PerimeterX client code into KsApi.
- Added mocks for `PXManager` and `PXBlockResponse`.
- Added more test coverage for the `handleError(response:and:)` function in `PerimeterXClient`.
- Moved debug logging and removed tests for this.
- Moved the presentation of the captcha to `AppDelegate` and `AppDelegateViewModel`.
- Added some test coverage for the web view cookie creation.

# 🤔 Why

- We noticed a bug where changing environment or logging out on the simulator caused network activity to subsequently fail. Upon investigation it seemed necessary to improve test coverage and to simplify the implementation a bit as it's only really the `KsApi` framework  and `AppDelegate` that need to know about PerimeterX. It was previously also linked to `Library` which was not needed.
- Adding additional mocks allowed us to test the actual code paths that would be followed when handling the error rather than missing all of that by mocking the client. In fact what we wanted to mock was `PXManager` which returns whether or not a captcha should be displayed.
- The logging in `AppDelegate` did not need tests as it's only for our debugging purposes, these were moved into the client in `KsApi`.
- The simplest way for KsApi to reach out to the `AppDelegate` in order to trigger the captcha display is by using a notification. While we can reference the app's window for presentation directly, this at least allowed us to add a meaningful test in `AppDelegateViewModel` for the display of the captcha.

# 🛠 How

- Updated target memberships of various files.
- Renamed and moved files into the KsApi folder structure.
- Added and removed tests.
- Added additional protocols and mocks.

# ✅ Acceptance criteria

- [ ] No regression in the behavior of PerimeterX.
- [ ] Changing environments does not cause networking calls to fail.
- [ ] Logging out does not cause networking calls to fail.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/lib/ApiDateProtocol.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+
+public protocol ApiDateProtocol {
```
singhhari, line 3: We have something similar already in the project but I'm assuming this needed to be made because of target membership? Also, is this specifically for the cookie expiration?

KsApi/lib/ApiDateProtocol.swift
```diff
@@ -0,0 +1,44 @@
...
+  public var date: Date {
+    return self
+  }
+}
+
+internal struct ApiMockDate: ApiDateProtocol {
+  private let time: TimeInterval
+
+  internal init() {
+    self.time = 1_475_361_315
```
singhhari, line 22: What does this number represent?

KsApi/px/MockPerimeterXTypes.swift
```diff
@@ -0,0 +1,35 @@
...
+struct MockPerimeterXBlockResponse: PerimeterXBlockResponseType {
+  var blockType: PXBlockType
+
+  var type: PXBlockType {
+    return self.blockType
+  }
+}
+
+final class MockPerimeterXManager: PerimeterXManagerType {
+  var headers: [AnyHashable: Any] = ["PX-AUTH-TEST": "foobar"]
```
singhhari, line 20: I think we can make these variables private given the accessor methods below?

KsApi/extensions/NSURLSession.swift
```diff
@@ -24,7 +24,8 @@ internal extension URLSession {
       .flatMap(.concat) { data, response -> SignalProducer<Data, GraphError> in
         guard let response = response as? HTTPURLResponse else { fatalError() }
 
-        error?.handleError(blockResponse: response, and: data)
+        /// `error` is `nil` or `handleError` returns `false`.
+        guard [nil, false].contains(error?.handleError(response: response, and: data)) else { return .empty }
```
singhhari, line 28: Ah so `.empty` will terminate right here if there is an error?

moderateepheezy: Since `.empty` calls `completed` with no value returned to the Observer/Signal. `rac_graphDataResponse` pretends like nothing happened if there's an error or `handleError` = true while delegating error handling to `error?.handleError`.

Are we showing the captcha response to the user somewhere, or what kind of action do we perform when we call `GoToPerimeterXCaptcha`?

KsApi/lib/ApiDateProtocol.swift
justinswart: Yeah ideally we'd at some point move things like this into a shared _lib_ or something but right now in order to write that test in KsApi this was necessary. If we moved `DateProtocol` to `KsApi` we'd have to import `KsApi` everywhere that we use it and for now this is just for the one test. Not ideal 🤷‍♂️

KsApi/lib/ApiDateProtocol.swift
justinswart: Just a static timestamp so that the tests aren't affected by the device time.

KsApi/px/MockPerimeterXTypes.swift
justinswart: These are just exposed if we want to manipulate in tests.

KsApi/extensions/NSURLSession.swift
justinswart: Yeah so just in case the request errors out and we display some sort of error dialogue, this will just return an empty `SignalProducer` and the captcha should be displayed.

singhhari (approved):
lgtm

PR #1430, opened 2021-04-16
[EP-432] QA Phase 1 feedback - analyticsName

singhhari:
# 📲 What

This PR addresses a current issue when device language is set to a language other than English. Specifically, category and subcategory values are sent in the device language. The backend was updated to always send an English String under the `analyticsName` field which we've added here and updated the following Segment properties with:

- `discover_category_name`
- `discover_subcategory_name`
- `project_category`
- `project_subcategory`

# 🤔 Why

We noticed while testing, the category and subcategory names were being sent to Segment/Amplitude in the language set on the iPhone. This makes using the category and subcategory more difficult for Insights given different words representing the same thing. We've updated the graphQL query to accept a new parameter: `analyticsName`, which will always be the English language string.

# 👀 See
Screen Shot 2021-04-16 at 3 34 46 PM

# ✅ QA Steps

- [ ] Change simulator/device language to Spanish.
- [ ] Trigger an event that sends `discoverProperties` and `projectProperties` to Segment.
- [ ] Verify the four properties referenced above are sending English language strings for the category and subcategory name.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/DiscoveryFiltersViewControllerTests.swift | indent, wrap, wrapArguments |


</p>

justinswart (approved):
lgtm!

PR #1431, opened 2021-04-16
[NT-1836, NT-1841] Add Braze SDK and Feature Flag

justinswart:
# 📲 What

Adds the Braze SDK in a side-by-side configuration with Segment and also adds a feature flag that we will use to put Braze functionality behind.

**Note:** Currently due to the way that Segment and Braze are initialized (during the initialization of the `AppEnvironment`) the feature flag does not work for flagging the initialization of these libraries. I will be opening a subsequent PR to move this initialization to after the `AppEnvironment` is constructed and instead replace `KSRAnalytics` in the environment with a new instance that has the `Segment` client.

# 🤔 Why

We plan to use Braze to improve our app messaging and engagement.

# 🛠 How

Added the SDK and the feature flag.

# 👀 See

# ✅ Acceptance criteria

- [ ] Running the app with the Segment and Braze feature flags turned on should cause your user to appear in our Braze staging console.


singhhari (approved):
seeing my user on Braze. lgtm.

PR #1434, opened 2021-04-16
[NT-1844] Register Push Tokens With Segment/Braze

justinswart:
# 📲 What

Registers push tokens with Braze via Segment.

# 🤔 Why

This allows us to send pushes using Braze.

# 🛠 How

Passing the device token upon registration to the Segment library which is in a side-by-side configuration with the Braze SDK.

# 👀 See

https://segment.com/docs/connections/destinations/catalog/braze/#ios
https://www.braze.com/docs/user_guide/message_building_by_channel/push/push_registration/

# ✅ Acceptance criteria

Run the app on device using the KickBeta scheme and register for pushes.

- [ ] You should be able to send a push to your user ID from the Braze console for iOS Staging.


singhhari (approved):
lgtm

PR #1435, opened 2021-04-19
[EP-427] CTA-Clicked (project)

moderateepheezy:
# 📲 What

1. Update event properties we send to segment when tracking  projects clicked from the result of a search query. The event should send the following properties specifically. 
`events` = "CTA Clicked"
`context_cta` = `project`
`context_page` = `search`
`context_location` = `search_results`
`context_type` = `results`

2. Change the event name of `trackProjectCardClicked` to `ctaClicked`  from `cardClicked`
3. Update event name for tests using `trackProjectCardClicked ` to use `ctaClicked`

# 🤔 Why

Projects clicked from the list of projects from a search query are sending generic events properties applicable to all project card clicked. Some of the events properties should be different when the projects are clicked from a search query.

# 👀 See
Screenshot 2021-04-19 at 19 27 52

# ✅ Acceptance criteria

1. Click on the search tab
2. Click on search bar and type in some keyword
3. From the list of results, click on a project card
4. verify the following events are being sent to segment only when user clicks on projects from search results

> `events` = "CTA Clicked"
> `context_cta` = `project`
> `context_page` = `search`
> `context_location` = `search_results`
> `context_type` = `results`


Library/Tracking/KSRAnalytics.swift
```diff
@@ -827,6 +833,7 @@ public final class KSRAnalytics {
    - parameter page: The `PageContext` representing the specific area the UI is interacted in
    - parameter checkoutData: The `CheckoutPropertiesData` associated with this specific checkout instance
    - parameter project: The `Project` corresponding to the card that was clicked
+   - parameter typeContext: The context of the  project clicked.
```
singhhari, line 836: Additional information about an event that was not captured in other context properties for the `Project` tapped.

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1047,19 +1047,22 @@ final class KSRAnalyticsTests: TestCase {
     ksrAnalytics.trackProjectCardClicked(
       page: .discovery,
       project: .template,
+      typeContext: .project,
```
singhhari, line 1050: Did you verify on airtable that the type is always .project for this event unless it is clicked after a search query?

Library/Tracking/KSRAnalytics.swift
```diff
@@ -836,14 +843,16 @@ public final class KSRAnalytics {
   public func trackProjectCardClicked(page: PageContext,
                                       project: Project,
                                       checkoutData: CheckoutPropertiesData? = nil,
+                                      typeContext: TypeContext? = nil,
```
singhhari, line 846: I wonder if the default makes more sense to be `project` since that is the argument we are using in every situation except results

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -134,20 +134,20 @@ internal final class SearchViewModelTests: TestCase {
...
 
         let dataLakeProperties = self.dataLakeTrackingClient.properties.last
         let segmentProperties = self.segmentTrackingClient.properties.last
 
         XCTAssertEqual("search", dataLakeProperties?["context_page"] as? String)
-        XCTAssertEqual("project", dataLakeProperties?["context_type"] as? String)
+        XCTAssertEqual("results", dataLakeProperties?["context_type"] as? String)
         XCTAssertEqual("search", segmentProperties?["context_page"] as? String)
-        XCTAssertEqual("project", segmentProperties?["context_type"] as? String)
+        XCTAssertEqual("results", segmentProperties?["context_type"] as? String)
```
singhhari, line 150: Can you add tests for all the other properties defined for this event as well?

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
moderateepheezy: I'm not sure about that because it's not specific to results, I have seen one other ticket that has a different `typeContext`  and there could be more. Most of the ones that has projects probably have a ticket that changes the `typeContext` to something else.

https://kickstarter.atlassian.net/jira/software/projects/EP/boards/53?label=ios%2Cphase_2&selectedIssue=EP-431

Library/ViewModels/SearchViewModelTests.swift
moderateepheezy: The tests for the other properties are already taken care of in KSRAnalyticsTest, but I can always add it here if that's ok.

Library/Tracking/KSRAnalytics.swift
singhhari: Can you wrap Project with `` so it is highlighted when it is option clicked.

Library/ViewModels/SearchViewModelTests.swift
singhhari: Yeah add them in here just because we are checking page and type anyway.

PR #1436, opened 2021-04-19
[NT-1850] Retain Debug Feature Flag Values For Non-Release Builds

justinswart:
# 📲 What

Retains the feature flag toggles that we set in our debug menu between app launches.

# 🤔 Why

This makes our development workflow easier so that we don't have to remember to enable feature flags each time the app launches while we're testing.

# 🛠 How

When the app starts up it restores its config from storage, I've added a function to give the values retrieved from storage precedence over the incoming values.

Note that this means the toggles that we have in the app will always override those that are coming in from the back-end. In order to reset the debug toggles and receive the values from the back-end we will need to delete the app and install it again.

# ✅ Acceptance criteria

- [ ] Launch the app in alpha, beta or debug, toggle feature flags in the debug menu. Re-launch the app and navigate back to this menu, your flags should be configured as they were before you re-launched.


singhhari (approved):
lgtm

PR #1437, opened 2021-04-20
[EP-435] CTA Clicked (project)

tobitech:
# 📲 What

- Convert Project Card Clicked event to a CTA Clicked event when a project card is tapped on discovery screen when `Recommended For You` filter is selected.
- Add an initializer to `TypeContext` enum to determine type context for an event based on `DiscoveryParams`

# 🤔 Why

This is part of implementation for changes to event sent when a project card is tapped on discovery screen when `Recommended For You` filter is selected

# 👀 See
Screenshot 2021-04-20 at 11 24 22

# ✅ Acceptance criteria

1. Launch the app 
2. Confirm that `Recommended For You` is selected on the discovery screen
3. Tap on any project
4. Verify that this properties appear on the Segment Debugger
```
events = "CTA Clicked"
context_cta = project
context_page = discover
context_location = discover_advanced
context_type = recommended
```
5. Also verify that other Discovery Properties appear on the Segment Debugger


Library/Tracking/KSRAnalytics.swift
```diff
@@ -379,6 +379,26 @@ public final class KSRAnalytics {
...
+
+     - parameter params: a `DiscoveryParams` object
+     */
+    init?(params: DiscoveryParams) {
+      if let recommended = params.recommended, recommended {
+        self = .recommended
+      } else if let starred = params.starred, starred {
+        self = .watched
+      } else if let staffPicks = params.staffPicks, staffPicks {
+        self = .pwl
```
singhhari, line 394: based on the ticket im working on this should be `self = .results`

Library/Tracking/KSRAnalytics.swift
```diff
@@ -379,6 +379,26 @@ public final class KSRAnalytics {
...
+      } else if let social = params.social, social {
+        self = .social
+      } else {
+        return nil
```
singhhari, line 398: this should also be `self = .results`. see if you can consolidate the logic for both cases that require a `self = .results`

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -1044,17 +1044,19 @@ final class KSRAnalyticsTests: TestCase {
     let segmentClient = MockTrackingClient()
     let ksrAnalytics = KSRAnalytics(dataLakeClient: dataLakeClient, segmentClient: segmentClient)
 
+    let typeContext = KSRAnalytics.TypeContext.init(params: DiscoveryParams.recommendedDefaults)
```
singhhari, line 1047: I don't think the `.init` part of the line is needed. Try to remove it in all the places you added it in and see if it compiles.

singhhari: Also I don't think you need this constant since you are only using it on line 1052.

Library/ViewModels/DiscoveryPageViewModel.swift
```diff
@@ -448,10 +448,13 @@ public final class DiscoveryPageViewModel: DiscoveryPageViewModelType, Discovery
     paramsChanged
       .takePairWhen(self.tappedProject.signal.skipNil())
       .observeValues { params, project in
+
+        let typeContext = KSRAnalytics.TypeContext.init(params: params)
```
singhhari, line 452: Same note as above about `.init` and not needing a constant.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -379,6 +379,26 @@ public final class KSRAnalytics {
     case watch
     case watched
 
+    /**
+     Initialize `TypeContext` with `DiscoveryParams`
+     to determine type context for discovery collection filters.
+
+     - parameter params: a `DiscoveryParams` object
+     */
+    init?(params: DiscoveryParams) {
```
singhhari, line 388: After making the changes below, you can remove the `?` here. It will no longer need to be a failable init.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -379,6 +379,26 @@ public final class KSRAnalytics {
     case watch
     case watched
 
+    /**
+     Initialize `TypeContext` with `DiscoveryParams`
```
singhhari, line 383: Rename to:
Initialize a `TypeContext` value with `DiscoveryParams` for use with discovery filters.

singhhari (changes requested):

singhhari (approved):
lgtm

PR #1440, opened 2021-04-20
[NT-1851] Enable/Disable Segment By Feature Flag

justinswart:
# 📲 What

- Moves our configuration of Segment to be after the `AppEnvironment` is set up.
- Enables and disables all Segment tracking by feature flag.
- Tidies up our protocols that we were using before to mock and feature flag tracking.

# 🤔 Why

- We had a race condition in that we were instantiating Segment's `Analytics` during the instantiation of the `AppEnvironment` and therefore could not inspect the environment at the same time. Moving this makes more tests possible.
- Previously we were feature-flagging the `track`, `identify`, etc methods but Segment would still track app lifecycle events. It turns out Segment provides `disable()` and `enable()` functions to completely disable/enable tracking which we are now using instead.
  - Note: Segment is _disabled_ by default and enabled by the feature flag configuration.
- We no longer need some of the protocol functions as we don't need to test the feature flagging at those particular call-sites.

# 🛠 How

- Moved Segment initialization to be driven by an `AppDelegateViewModel` output.
- Observing configuration changes and enabling/disabling Segment.
- Tidied up these protocols.

# ✅ Acceptance criteria

Open Segment's debugger and run the app.

- [ ] Toggle the Segment feature flag in the app's debug menu, observe that events are or are not sent to Segment based on the state of the flag.
- [ ] Trigger a reload of the config and intercept it using Charles Proxy. Edit the response to enable or disable Segment using the feature flag. Observe that events are or are not sent to Segment based on the state of the flag.

### DIFF
--- a/Configs/Secrets.swift.example
+++ b/Configs/Secrets.swift.example
@@ -40,8 +40,8 @@ public enum Secrets {
   }
 
   public enum Segment {
-    public static let staging = "write-key"
-    public static let production = "write-key"
+    public static let staging = "write-key-production"
+    public static let production = "write-key-staging"
   }
 
   public enum Qualtrics {
--- a/Configs/Secrets.swift.example
+++ b/Configs/Secrets.swift.example
@@ -40,8 +40,8 @@ public enum Secrets {
   }
 
   public enum Segment {
-    public static let staging = "write-key-production"
-    public static let production = "write-key-staging"
+    public static let staging = "write-key-staging"
+    public static let production = "write-key-production"
   }
 
   public enum Qualtrics {
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -271,11 +271,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     self.viewModel.outputs.segmentIsEnabled
       .observeValues { enabled in
-        if enabled {
-          Analytics.shared().enable()
-        } else {
-          Analytics.shared().disable()
-        }
+        enabled ? Analytics.shared().enable() : Analytics.shared().disable()
       }
 
     NotificationCenter.default
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -263,6 +263,30 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         self?.viewModel.inputs.userSessionEnded()
       }
 
+    self.viewModel.outputs.configureSegment
+      .observeValues { writeKey in
+        let client = Analytics.configuredClient(withWriteKey: writeKey)
+        AppEnvironment.current.ksrAnalytics.configureSegmentClient(client)
+      }
+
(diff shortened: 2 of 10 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -263,6 +263,30 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    self.viewModel.outputs.segmentIsEnabled
+      .observeValues { enabled in
+        if enabled {
+          Analytics.shared().enable()
+        } else {
+          Analytics.shared().disable()
+        }
+      }
+
+    NotificationCenter.default
```
justinswart, line 281: Ordinarily we would hang on to the token for these observers and release them but the `AppDelegate` is never released so I guess it's ok here 😄

Library/Tracking/Segment.swift
```diff
@@ -7,54 +7,27 @@ public extension Analytics {
...
  The `TrackingClientType` and `IdentifyingTrackingClient` protocols
- allow us to create mocks to test these code paths. The protocol exposes functions that are named similarly
- but differently so that we can perform the check to see that tracking is enabled before calling any of the
- functions on the library itself so as to not unintentionally contribute to tracking data during debugging.
+ allow us to create mocks to test these code paths.
  */
-
-extension Analytics: IdentifyingTrackingClient {
-  public func identify(userId: String?, traits: [String: Any]?) {
-    guard AppEnvironment.current.environmentVariables.isTrackingEnabled,
```
justinswart, line 39: Weirdly no tests failed removing this so it looks like we weren't testing it anyway 🤦‍♂️

Configs/Secrets.swift.example
```diff
@@ -40,8 +40,8 @@ public enum Secrets {
   }
 
   public enum Segment {
-    public static let staging = "write-key"
-    public static let production = "write-key"
+    public static let staging = "write-key-production"
+    public static let production = "write-key-staging"
```
singhhari, line 44: looks like you mixed these up

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -263,6 +263,30 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
+    self.viewModel.outputs.configureSegment
+      .observeValues { writeKey in
+        let client = Analytics.configuredClient(withWriteKey: writeKey)
+        AppEnvironment.current.ksrAnalytics.configureSegmentClient(client)
+      }
+
+    self.viewModel.outputs.segmentIsEnabled
+      .observeValues { enabled in
+        if enabled {
```
singhhari, line 274: would a ternary operator work here?

Configs/Secrets.swift.example
justinswart: oh lol yeah, they mostly just need to not be the same but best have them the right way around 👍

Kickstarter-iOS/AppDelegate.swift
justinswart: Don't see why not! 👍

PR #1442, opened 2021-04-21
[EP-501] Update conversion rates

moderateepheezy:
# 📲 What

- Added a new field to Project.Stats called `usdExchangeRate`
- event property `project_current_amount_pledged_usd` = `usdExchangeRate` * `pledged`
- event property `project_goal_usd` = `usdExchangeRate` * `goal`
- Add a new Number extension that allows you do `Float?.orZero` instead of `Float? ?? 0`, also applicable to `Int` and `Double`

> Additional Information

- The values sent to segment should match the web, and Android
- The value being sent for `project_current_amount_pledged_usd ` and `project_goal_usd` will be rounded to 2 decimal (eg. 4.35)
- The only place the values won’t be the same are conversion rates a user sees in the UI will differ slightly from that in Segment, as we’re using different methods to calculate the conversion

# 🤔 Why

We previously use `fx_rate` to get value for event property `project_current_amount_pledged_usd ` and `project_goal_usd `. This causes some inconsistencies with the value web sends for event property `project_current_amount_pledged_usd ` and `project_goal_usd `. 

# 👀 See

# Web
Screenshot 2021-04-22 at 17 24 24

# iOS
Screenshot 2021-04-22 at 17 23 58

# ✅ Acceptance criteria

1. Clicked on a project on the iOS app.
2. Go to the web app and click on the same project.
3. Very on segment that `project_current_amount_pledged_usd ` and `project_goal_usd  are the same for both web and iOS.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Extensions/Optional+Numeric.swift | fileHeader |
| Library/Extensions/OptionalNumericTests.swift | fileHeader |


</p>

KsApi/extensions/Optional+Numeric.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
```
singhhari, line 1: I think we should delete this file and its corresponding test file. I don't think it's necessary.

KsApi/models/Project.swift
```diff
@@ -96,6 +97,11 @@ public struct Project {
       return floor(Float(self.pledged) * self.staticUsdRate)
     }
 
+    /// Total amount currently pledged to the project converted to USD irrespective of the user's selected currency
```
singhhari, line 100: /// Total amount currently pledged to the project, converted to USD, irrespective of the users selected currency

KsApi/models/Project.swift
```diff
@@ -106,6 +112,11 @@ public struct Project {
       return self.currentCurrencyRate.map { floor(Float(self.goal) * $0) }
     }
 
+    /// Goal amount converted to USD currency irrespective of the user's selected currency.
```
singhhari, line 115: /// Goal amount, converted to USD, irrespective of the users selected currency

KsApi/models/templates/ProjectStatsTemplate.swift
```diff
@@ -11,6 +11,7 @@ extension Project.Stats {
     goal: 2_000,
     pledged: 1_000,
     staticUsdRate: 1.0,
-    updatesCount: 1
+    updatesCount: 1,
+    usdExchangeRate: 1.56
```
singhhari, line 15: Should we test this with a number similar to what we were getting from the API? More trailing decimals is probably a good testing use case.

singhhari:
Need to bump coverage here as well.

KsApi/extensions/Optional+Numeric.swift
moderateepheezy: I have an argument for this but it's still subjective. Let me give some example.

There are cases where we have expressions similar to 
`(someValueA ?? 0)` + `( someValueB ?? 0)`, with the new extension you can now have this instead
`someValueA.orZero` + `someValueB.orZero`

Another example is this
`let someNotNilValue: Int? = 3`
`let someNilValue: Int? = nil`

`let value = someNotNilValue + someNilValue` //  you can't do this without introducing a bracket.
so you'll have to do
`let value = someNotNilValue + (someNilValue ?? 0)`

but with the new extension you can do this instead.
`let value = someNotNilValue + someNilValue.orZero`

Makes the code a little easier to read, and also reduces using brackets.

singhhari: I think `orZero` makes the code less readable actually. The nil coalescing operator is a powerful part of the Swift language and I think when you see one it immediately suggests you are handling an optional with a default case. There will probably be a time in the future where we don't want to default to zero for a Float or Double so I think this extension has a very small, specific use case.

singhhari (approved):
Update description with some more details.
lgtm.

PR #1444, opened 2021-04-22
[NT-1864] Set All Codesigning To Manual

justinswart:
# 📲 What

Sets our Debug configuration to use manual code signing.

**Note:** Once this has merged and you've pulled the latest changes, run `bundle exec fastlane match_all` locally. If you see a warning in Xcode about the provisioning profile not including a signing certificate, open your Keychain Access, locate the certificate by its ID and remove it.

# 🤔 Why

We've always kept these set to manual to simplify code signing on our team. It's ok to set this to automatic when one needs to build the app on device but we prefer to not commit these changes to the codebase. The main reason is because our CI relies on these settings being manual but we've also found that it is less likely to lead to unexpected behavior between different folks working on the codebase.

# 🛠 How

Set our Debug configuration's code signing to manual.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [ ] No build issues on simulator or device (set to automatic again when building to device, don't commit change).


singhhari (approved):
lgtm

PR #1447, opened 2021-04-26
[EP-238] CTA Clicked (discover_filter)

moderateepheezy:
# 📲 What

Update `discovery_category_name` event property to be the name of the category if the `parent` field in  DiscoveryParam of a selected filter  is nil

# 🤔 Why

When selecting a category_name filter (e.g. Crafts as the category and then no subcategory selected, just "All Crafts Projects") from the discover overlay on the iOS app, this is what should be sent:

`context_cta = discover_filter` CORRECT

`context_location = discover_overlay` CORRECT

`context_type = category_name` CORRECT

`discover_category_name = Crafts` <<< MISSING

`discover_subcategory_name` = Crafts CORRECT

The issue is only when the user selects a category, but does not select a subcategory. In that situation, the property discover_category_name is missing.

# 👀 See

Screenshot 2021-04-26 at 16 25 28

# ✅ Acceptance criteria

1. Select a Category in drop down menu on Discovery Tab. e.g Craft
2. Select `All Craft` Category
3. Verify on segment that
> `discover_category_name ` =  "Craft"
> `discover_subcategory_name ` =  "Craft"


singhhari (approved):
lgtm

PR #1449, opened 2021-04-26
[EP-450] context_page = other

moderateepheezy:
# 📲 What

- Remove `page` parameter in `track` function inside KSRAnalytics.
- default `page` in `contextProperties()` function to `other` for some events that Product/Insight hasn't defined `page_context` for.

# 🤔 Why

We previously had `page` as a parameter in our `track` function as well as `contextProperties()` function. This is not necessary as we need all of our `Context Properties` to be in on function called `contextProperties()`

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -980,7 +980,8 @@ public final class KSRAnalytics {
     case .pledge:
       let allProps = props
         .withAllValuesFrom(optimizelyProperties() ?? [:])
-        .withAllValuesFrom(contextProperties(ctaContext: .pledgeInitiate, page: .projectPage))
+        .withAllValuesFrom(contextProperties(ctaContext: .pledgeInitiate))
+        .withAllValuesFrom(props)
 
       self.track(
         event: NewApprovedEvent.ctaClicked.rawValue,
@@ -1153,14 +1154,14 @@ public final class KSRAnalytics {
    */
 
   public func trackAddNewCardButtonClicked(
-    location: KSRAnalytics.PageContext? = nil,
+    page: KSRAnalytics.PageContext? = nil,
     project: Project,
     refTag: RefTag?,
     reward: Reward
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(pledgeProperties(from: reward))
-      .withAllValuesFrom(contextProperties(page: location))
+      .withAllValuesFrom(contextProperties(page: page))
 
     self.track(
       event: ApprovedEvent.addNewCardButtonClicked.rawValue,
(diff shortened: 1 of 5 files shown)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Tracking/KSRAnalytics.swift | indent |


</p>

Library/Tracking/KSRAnalytics.swift
```diff
@@ -961,29 +969,27 @@ public final class KSRAnalytics {
...
         event: ApprovedEvent.managePledgeButtonClicked.rawValue,
-        page: .projectPage,
-        properties: props.withAllValuesFrom(contextProperties()) // .fixErroredPledge
+        properties: props // .fixErroredPledge
       )
     case .pledge:
       let allProps = props
         .withAllValuesFrom(optimizelyProperties() ?? [:])
-        .withAllValuesFrom(contextProperties(ctaContext: .pledgeInitiate))
+        .withAllValuesFrom(contextProperties(ctaContext: .pledgeInitiate, page: .projectPage))
```
singhhari, line 983: You can delete `.projectPage` here because you are assigning it on line 972?

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1460,6 +1469,7 @@ public final class KSRAnalytics {
     var props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(contextProperties(
         ctaContext: .watchProject,
+        page: location,
```
singhhari, line 1472: can you rename the location parameter to `page`?

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1806,7 +1815,7 @@ private func contextProperties(
 
   result["cta"] = ctaContext?.trackingString
   result["location"] = locationContext?.trackingString
-  result["page"] = page?.rawValue
+  result["page"] = page?.rawValue ?? "other"
```
singhhari, line 1818: leave an inline comment here explaining why we're using "other" based on insights definition.

singhhari (changes requested):

Library/Tracking/KSRAnalytics.swift
moderateepheezy: If I remove it there 
`contextProperties(ctaContext: .pledgeInitiate)` would reset page to nil because the function is called after you've already set it.

Another option is to explicitly define `contextProperties(page: .projectPage)` in each of the switch case.

singhhari: Ah good point. Can you test this out and see if it works:

`let allProps = contextProperties(ctaContext: .pledgeInitiate).withAllValuesFrom(optimizelyProperties() ?? [:]).withAllValuesFrom(props)`

moderateepheezy: This worked!

singhhari (approved):
lgtm

PR #1452, opened 2021-04-29
[NT-1866] Reduce Analytics Identify Calls

justinswart:
# 📲 What

Reduces the amount of times that we call `identify` on our analytics client.

# 🤔 Why

We incur costs per API calls on Segment and data points on Braze so we want to reduce unnecessary `identify` calls and only send deltas of user traits to reduce the amount of data points that are updated. Traits are cumulative so sending the deltas is acceptable and by design.

# 🛠 How

I've added a `KSRAnalyticsIdentityData` type which we use to perform the necessary comparisons. This type is also encoded and decoded into user defaults so that it's persisted between app sessions.

# ✅ Acceptance criteria

TBD. All of the traits are always shown on the Segment debugger - we may have to debug this using Charles Proxy.

By placing this print statement in the code we can see that it is only sending through the delta that I am toggling:

image

### DIFF
--- a/Library/Tracking/KSRAnalyticsIdentityData.swift
+++ b/Library/Tracking/KSRAnalyticsIdentityData.swift
@@ -1,0 +1,81 @@
+import Foundation
+import KsApi
+
+private enum KSRAnalyticsIdentityTraitValue: Equatable {
+  case string(String)
+  case bool(Bool)
+
+  var value: Any {
+    switch self {
+    case let .string(value): return value
+    case let .bool(value): return value
+    }
+  }
+}
+
+public struct KSRAnalyticsIdentityData: Equatable {
+  public let userId: Int
+  private let name: String
+  private let notifications: User.Notifications
+
+  init(_ user: User) {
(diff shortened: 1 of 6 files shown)


Library/Tracking/KSRAnalyticsIdentityData.swift
```diff
@@ -0,0 +1,81 @@
...
+}
+
+extension KSRAnalyticsIdentityData: Codable {
+  private enum CodingKeys: String, CodingKey {
+    case userId
+    case name
+    case notifications
+  }
+
+  public func encode(to encoder: Encoder) throws {
```
Arkariang, line 63: what codification do you use?

justinswart: Sorry, not sure what you mean? This is using `Swift`'s `Codable` protocol conformance so that we can serialize this model into a `Data` object and store it in our key value store.

singhhari (approved):
lgtm

PR #1454, opened 2021-04-29
Changing Estimated Delivery to format like a section title.

singhhari:
# 📲 What

In the rewards carousel, “Estimated delivery” should resemble treatment on "Includes" (bolded, grey, so that it stands out from reward description).

# 🤔 Why

This is part of the iOS teams Design Debt epic. The change gives a, much needed, section title for the Estimated Delivery section of the card and aligns with Includes in design.

# 👀 See
Simulator Screen Shot - iPhone 12 Pro - 2021-04-29 at 16 09 07

# ✅ Acceptance criteria

- [ ] Navigate to a Project you can engage in the Backing flow with
- [ ] When getting to the Rewards carousel, verify the formatting of Includes and Estimated Delivery are the same


Kickstarter-iOS/Views/RewardCardView.swift
```diff
@@ -120,7 +140,7 @@ public final class RewardCardView: UIView {
     self.minimumPriceConversionLabel.rac.hidden = self.viewModel.outputs.conversionLabelHidden
     self.minimumPriceConversionLabel.rac.text = self.viewModel.outputs.conversionLabelText
     self.descriptionLabel.rac.text = self.viewModel.outputs.descriptionLabelText
-    self.estimatedDeliveryDateLabel.rac.hidden = self.viewModel.outputs.estimatedDeliveryDateLabelHidden
+    self.estimatedDeliveryStackView.rac.hidden = self.viewModel.outputs.estimatedDeliveryDateLabelHidden
```
justinswart, line 143: Not such a big deal in this case but typically we'd want to keep the output name matching the element so in this case we'd rename to `estimatedDeliveryStackViewHidden` but it's still pretty clear here so just mentioning.

justinswart (approved):
Looks good, just the one comment and would be good to add the JIRA ticket number to the PR title so that it all gets tied up correctly 👍

Kickstarter-iOS/Views/RewardCardView.swift
singhhari: Renamed and pushing it up. Will merge when that finishes on CI.

PR #1457, opened 2021-04-30
[NT-1873] Implement BrazeDebounceMiddleware

justinswart:
# 📲 What

Updates our analytics `identify` call to always send all traits to Segment while de-duping the traits send to Braze using the middleware class provided by Braze.

# 🤔 Why

In #1452 we did some work to reduce the amount of identify calls due to the nature of our billing with Segment being per API call. We initially thought that this would cause Segment to only sent the deltas on to Braze but it appears that it always sends all of the traits that it has in its cache. Adding the middleware is the intended solution for this and inserts a step between Segment's identify call and the data being sent on to Braze during which these traits are de-duped.

# 🛠 How

- Added the `BrazeDebounceMiddleware` class and configured the Segment client with it.
- Removed the de-duplication from the traits that we send to Segment's `identify` call and instead always send everything but retain the behavior of not calling `identify` at all if nothing has changed.

# 👀 See

Segment/Braze Middleware - https://github.com/segmentio/segment-braze-mobile-middleware

# ✅ Acceptance criteria

With a fresh installation of the app:

- [x] Login and observe in the Segment debugger that all traits are being sent. Also observe in the Braze developer console that all traits are being sent.
- [x] Toggle a notification option on/off. Observe in the Segment debugger that all traits are being sent. Also observe in the Braze developer console that only the delta is being sent (only the trait that you updated).

### DIFF
--- a/.slather.yml
+++ b/.slather.yml
@@ -1,2 +1,3 @@
 ignore:
   - Library/Strings.swift
+  - Library/Tracking/Vendor/BrazeDebounceMiddleware.swift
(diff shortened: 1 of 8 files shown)


justinswart:
The hit in code coverage here is because we've added this vendor class directly to the codebase. I suppose I could go and specifically exclude it from codecov but I think it's ok.

singhhari (approved):
lgtm

PR #1458, opened 2021-04-30
[NT-1869] Feature-Flag Braze In-App Messages

justinswart:
# 📲 What

Allows us to control the display of in-app messages from Braze with a feature flag.

# 🤔 Why

This is currently under development and may be something that we want to roll out gradually in production.

# 🛠 How

Wired up the `ABKInAppMessageControllerDelegate` method `before(inAppMessageDisplayed:)` to an input on our `AppDelegateViewModel` which returns a `ABKInAppMessageDisplayChoice` based on whether or not the feature flag is enabled.

# 👀 See

https://segment.com/docs/connections/destinations/catalog/braze/#ios

# ✅ Acceptance criteria

- [x] Send yourself a test in-app message from the Braze developer console with the feature flag enabled. The message should be displayed.
- [x] Send yourself a test in-app message from the Braze developer console with the feature flag disabled. The message should not be displayed.


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -486,17 +499,33 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
...
     didReceive response: UNNotificationResponse,
     withCompletionHandler completion: @escaping () -> Void
   ) {
     guard let rootTabBarController = self.rootTabBarController else {
       completion()
       return
     }
+
+    // Braze
+    if Appboy.sharedInstance() == nil {
```
justinswart, line 513: The Segment docs say to add these lines in these places. I'm open to chatting about testability here - say adding something like a `SegmentBrazeIntegrationFactoryType` to the `AppEnvironment` and conforming the various Braze types so that we can test that if we call particular view model inputs that the correlating functions on the Braze client are also called.

This is a fair bit of work so I'm just wondering how necessary we think it is. On the one hand we'll prove that if we call a particular view model input we're also calling a function on the Braze client but I just want to avoid writing tests for vendor code. That is to say that - calling these does not produce real side-effects for us to test. Sure, we would like to know that the Braze code is called at the right time but we're not responsible for what the Braze SDK is doing 🤔 

I'll give it some more thought and we can discuss, if we want to, how best to test this.

Arkariang: The approach I've followed on Android was to create my own client that internally will call the vendor dependencies. I do not need - at least at this point- to refer from any viewModel to this client, see -> https://github.com/kickstarter/android-oss/blob/9834dd8de1ab0545683e87618cfb2ac65ab19fe1/app/src/main/java/com/kickstarter/libs/braze/BrazeClient.kt#L148

On my end it was not just about testing, it was focused on not speaking directly to third party dependencies and facilitate possible change of libraries on the future.

We can probably bring this as point for discussion on the Native Engineering meeting

singhhari: Good point. Might be best to have some parity between iOS and Android in our approach.

singhhari (approved):
lgtm

Kickstarter-iOS/AppDelegate.swift
justinswart: Hmm yeah so I considered this wrapper approach and it's true that it may help us if we were to switch the Braze client out with an entirely different push client doing the same/similar thing, I guess I just see this interface being very specific to Braze and if the client were to change the interface would probably change with it. In this case I am interested in trying to test some of the logic here that in this case we're missing. So that is - I'm not so interested in mocking this type but rather making sure that the checks for the `nil` Appboy instance is happening as it should. I've got some ideas around this and will push up another commit.

justinswart: Ok, after looking at this a couple of different ways I decided it doesn't make sense to us to try and mock this type, the only thing worth testing is the little bit of logic that checks if the `Appboy` instance is `nil` so I've added a simple test for that.

singhhari: Posting for reference: Paired with Justin on this and mocking the type/creating a wrapper turned out to be quite a lot of complexity. Makes sense to avoid all the extra effort for one small use case.

singhhari (approved):
lgtm

PR #1461, opened 2021-05-05
[NT-1870] Enable Notification Setting From In-App Message

justinswart:
# 📲 What

Allows us to enable a push notification setting from an in-app message dialogue sent from Braze.

# 🤔 Why

We are making use of in-app messages from Braze to inform users of new notifications that they may wish to subscribe to.

# 🛠 How

- Implement the `ABKURLDelegate` `handleAppboyURL(_:from:withExtras)` function which receives the URL from the in-app message button tap.
- Parse the URL through a `Navigation` deep-link.
- Update the currently logged in user and send a `PUT` request to the `/v1/users/self` endpoint.

# 👀 See

https://www.braze.com/docs/developer_guide/platform_integration_guides/ios/in-app_messaging/customization

# ✅ Acceptance criteria

Go to your user's notification settings, disable the push notification for mobile marketing updates (top of the list). Enable the Braze feature flag in the debug menu.

- [x] Send yourself a test IAM from Braze with a button configured to deep link with the expected URL structure (see `NavigationTests.swift`) Be sure to use the staging base URL when testing on simulator/beta. Launch the app (IAMs were only shown to me upon app-launch). You should see the IAM, tapping the button that the URL was assigned to should dismiss the modal and the notification option should be enabled for your user.
- [ ] Using any other valid URL replacing the notification option should also work for that option.
- [x] Using an invalid URL should do nothing (or cause Braze to handle it using its default behaviour).

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -46,6 +46,9 @@ public protocol AppDelegateViewModelInputs {
   /// Call when the application receives a request to perform a shortcut action.
   func applicationPerformActionForShortcutItem(_ item: UIApplicationShortcutItem)
 
+  /// Call when a Braze in-app message's button is tapped with its URL and extras.
+  func brazeDidHandleURL(_ url: URL?, extras: [AnyHashable: Any]?) -> Bool
+
   /// Call when the Braze SDK will display an in-app message, return a display choice.
   func brazeWillDisplayInAppMessage(_ message: BrazeInAppMessageType) -> ABKInAppMessageDisplayChoice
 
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1188,3 +1210,38 @@ private func configRetainingDebugFeatureFlags(_ config: Config) -> Config {
...
+  else { return (false, nil) }
+
+  guard let navigation = Navigation.match(url) else {
+    return (false, nil)
+  }
+
+  return (true, navigation)
+}
+
+private func updateUserNotificationSetting(navigation: Navigation) -> SignalProducer<User, Never> {
```
justinswart, line 1228: Bit convoluted but this is in order to keep the enabling/disabling at least somewhat dynamic. This allows us to pass in say `notify_mobile_of_messages` or `notify_mobile_of_follower` and have it retrieve those settings from the current `User` in a dictionary form, update that dictionary, turn it back into `User.Notifications` and then lens over the `User` to update it.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -1188,3 +1210,38 @@ private func configRetainingDebugFeatureFlags(_ config: Config) -> Config {
...
+  guard
+    let data = try? JSONSerialization.data(withJSONObject: updatedNotifications as Any, options: []),
+    let userNotifications = try? JSONDecoder().decode(User.Notifications.self, from: data)
+  else { return .empty }
+
+  let updatedUser = currentUser |> User.lens.notifications .~ userNotifications
+
+  return AppEnvironment.current.apiService.updateUserSelf(updatedUser)
+    .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+    .demoteErrors()
```
justinswart, line 1246: We're not handling errors here so we just allow this to fail silently if it does.

msadoon (approved):

singhhari (approved):
lgtm

PR #1462, opened 2021-05-06
[NT-1884] Send User ID as String in User Properties

justinswart:
# 📲 What

Sends the user ID property as a `String` instead of an `Int`.

# 🤔 Why

@lauriestark found that there were data anomalies related to the user ID property in Amplitude and upon investigation it seems that this property is being flagged as a violation for not being a `String`.

Also, we _do_ send this property as a `String` when we call `identify` because that does not accept an `Int`. It seems reasonable to assume that these data types should be the same but I've opened this PR for us to test and discuss as I have limited context of this project thus far. @singhhari should advise.

# 🛠 How

Wrapped the user ID in a `String` and updated tests.

# 👀 See

Screenshot of violation in Segment debugger:

image

Source: https://app.segment.com/kickstarter/protocols/violations/ios/track

# ✅ Acceptance criteria

- [ ] Test an event through to Amplitude to confirm that the IDs shown there are correct.


singhhari (approved):
lgtm

PR #1463, opened 2021-05-07
[NT-1888] Improve Redundant Identify Handling

justinswart:
# 📲 What

Improves how we handle the redundant calls to Segment's `identify` API.

# 🤔 Why

During testing it was found that `identify` was not called when logging out and logging in on a device with a different user. I suspect this is because I was not checking for changes in the user ID but only in the rest of the traits. I've also ensured that we clear the persisted data at logout.

# 🛠 How

- Added a check to see if the user ID changes.
- Cleared persisted data at logout.

# ✅ Acceptance criteria

To be sure, reset your simulator completely.

- [ ] Log in and observe that identify is called in the Segment debugger.
- [ ] Log out and back in as the same user, observe that identify is called.
- [ ] Log out and log in as a different user, observe that identify is called.
- [ ] When logged in change any notification option and observe that identify is called.

### DIFF
--- a/Library/AppEnvironmentTests.swift
+++ b/Library/AppEnvironmentTests.swift
@@ -60,19 +60,23 @@ final class AppEnvironmentTests: XCTestCase {
     XCTAssertNil(AppEnvironment.current.currentUser)
 
     AppEnvironment.login(AccessTokenEnvelope(accessToken: "deadbeef", user: User.template))
+    AppEnvironment.current.userDefaults.analyticsIdentityData = KSRAnalyticsIdentityData(.template)
 
     XCTAssertEqual("deadbeef", AppEnvironment.current.apiService.oauthToken?.token)
     XCTAssertEqual(User.template, AppEnvironment.current.currentUser)
+    XCTAssertNotNil(AppEnvironment.current.userDefaults.analyticsIdentityData)
 
     AppEnvironment.updateCurrentUser(User.template)
 
     XCTAssertEqual("deadbeef", AppEnvironment.current.apiService.oauthToken?.token)
     XCTAssertEqual(User.template, AppEnvironment.current.currentUser)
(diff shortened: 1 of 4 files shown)


Library/AppEnvironmentTests.swift
```diff
@@ -60,19 +60,23 @@ final class AppEnvironmentTests: XCTestCase {
     XCTAssertNil(AppEnvironment.current.currentUser)
```
msadoon, line 60: Super minor thing. Do we want to check the `analyticsIdentityData` here to be `nil`?

singhhari (approved):
working as expected but I do want to note that after step 4 of the acceptance criteria the entire identify call is sent and not just the diff of the notification I changed. not sure if this is expected behavior or not?

Library/AppEnvironmentTests.swift
```diff
@@ -60,19 +60,23 @@ final class AppEnvironmentTests: XCTestCase {
...
 
     AppEnvironment.login(AccessTokenEnvelope(accessToken: "deadbeef", user: User.template))
+    AppEnvironment.current.userDefaults.analyticsIdentityData = KSRAnalyticsIdentityData(.template)
 
     XCTAssertEqual("deadbeef", AppEnvironment.current.apiService.oauthToken?.token)
     XCTAssertEqual(User.template, AppEnvironment.current.currentUser)
+    XCTAssertNotNil(AppEnvironment.current.userDefaults.analyticsIdentityData)
 
     AppEnvironment.updateCurrentUser(User.template)
```
msadoon, line 69: I may be missing something, but do we want to update the `User.template` here? I think it's the same some as the instantiated value. Just because I think we're testing that method for a change.

msadoon (approved):

Library/AppEnvironmentTests.swift
justinswart: oh sure, let me add that 👍

PR #1471, opened 2021-05-11
[NT-1889, NT-1900] Replace Braze URL Handling With Regular Deep-Linking

justinswart:
# 📲 What

Replaces our manual interception of Braze URLs with iOS's regular deep-linking callbacks. This also allows us to pass in any known deep link via a Braze IAM or push notification.

**Note:** Updates first need to be made to our staging association file: https://staging.kickstarter.com/.well-known/apple-app-site-association - we need to remove KickAlpha and KickDebug from the list as they are causing deep linking into KickBeta to not work correctly. We typically test these things on KickBeta so we should remove the other two.

I've blocked the PR until we address this.

# 🤔 Why

In #1461 we implemented the enabling/disabling of user notification settings via deep link by implementing a Braze SDK delegate method which allows us to intercept the URL that Braze is about to handle via IAM or push. This worked but it's not the right way to do it. Initially it seemed as if Braze wasn't calling the standard iOS deep link delegate methods but after further investigation it turns out that it was due to a problem in the way that we register for universal links via our association file on staging. We'll address this separately but for now this PR adds the correct way to implement the deep link behavior for Braze.

# 🛠 How

Removed our URL interception code and replaced it with the simpler code that follows a regular deep link.

# ✅ Acceptance criteria

- [ ] Send yourself a test push from Braze with a deep link URL (Project or otherwise) this should deep link correctly in the app.
- [ ] Send yourself an IAM with a deep link URL (Project or otherwise) this should deep link correctly in the app.
- [ ] Send yourself an IAM with a deep link URL to update user notification settings as before, there should be no regression to this behavior.


msadoon (approved):
Yea looks good to me! I just didn't test it on device, for just iOS 14.5 being on device and not having XCode 12.5 working on this particular branch. 
We can test this on beta though as discussed.

singhhari (approved):
lgtm

PR #1472, opened 2021-05-12
[NT-1919] Keyboard Composer View Working Example

msadoon:
# 📲 What

This is just a working example of an expandable accessory view on the keyboard. It's far from the end result. In order to be complete, will have to merge this with the actual [keyboard composer view](https://kickstarter.atlassian.net/browse/NT-1893).

# 🤔 Why

As part of the spec required to make the comments composer view usable.

# 🛠 How

A lot of this is based on this stack overflow [post](https://stackoverflow.com/questions/31822504/how-can-i-increase-the-height-of-an-inputaccessoryview). 

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |
n/a just a testable example

# ♿️ Accessibility 
# 🏎 Performance
# ✅ Acceptance criteria

- [ ] Requires actual composer view

# ⏰ TODO

- [ ] As you can see in the example, there is a small scrolling issue, when scrolling becomes enabled, the top text gets cut off and extra padding is added to the bottom.  So its a minor hiccup that I want to spend some time fixing.
- [ ] Adding the composer view, view model and tests.


Kickstarter-iOS/Views/CommentInputAccessoryView.swift
```diff
@@ -0,0 +1,62 @@
...
+                                                       options: [],
+                                                       metrics: nil,
+                                                       views: ["view": view]))
+  }
+}
+
+// MARK: UITextViewDelegate
+// TODO: Is there a better reactive way to do this? I know RxSwift had things like textField.rx.didChange properties that could be subscribed to...
+extension CommentInputAccessoryView: UITextViewDelegate {
+  func textViewDidChange(_ textView: UITextView) {
```
tobitech, line 57: I think this is better. We mostly use `ReactiveSwift` for `Data Binding` between `ViewModels` and `ViewControllers` in the project.

tobitech:
Great job 👍

tobitech:
We could probably take a constraint approach for the height of the TextView. We would try out other approaches that would give a more stable experience for the scrolling.

msadoon:
Talking to @tobitech , he integrated pieces of what I did here into his PR (forthcoming), and it's the complete composer view with keyboard accessory view. Which expands in height. So I will close this PR as all the relevant pieces were added to the NT-1893 PR.

PR #1473, opened 2021-05-12
[NT-1915] Adding feature flag for comment threading to Optimizely.

singhhari:
# 📲 What

The comment threading feature we are beginning to work on will be encapsulated behind [this feature flag](https://app.optimizely.com/v2/projects/17017800748/features/20247874407/rules) configured in Optimizely. As of right now, it's only enabled for Staging but we will eventually shift this to build number based, instead of environmental. 

# 🤔 Why

The work behind this feature will take us a few sprints to get from development to release but we don't want to remove the possibility of any new releases to the AppStore between now and then. Using a feature flag enables us to maintain our release cadence and also be certain we are not exposing unfinished work to production.

# 🛠 How

The flag is set up on Optimizely and our existing application client will handle the rest.

# ✅ Acceptance criteria

- [ ] Toggle the value of the flag on Optimizely and verify the correct version of the view controller is rendering.


Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
```diff
@@ -133,8 +138,9 @@ public final class ProjectPamphletContentViewController: UITableViewController {
   }
 
   fileprivate func goToComments(project: Project) {
-    let vc = DeprecatedCommentsViewController.configuredWith(project: project, update: nil)
-
+    let vc = self
+      .commentThreadingFeatureIsEnabled() ? CommentsViewController() : DeprecatedCommentsViewController
```
justinswart, line 142: We could just pull this out to an internal function in this file and add a test for the VC being returned based on the flag.

Library/OptimizelyFeature.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 
 public enum OptimizelyFeature {
   public enum Key: String {
+    case commentThreading = "ios_comment_threading"
     case lightsOn = "ios_lights_on"
```
justinswart, line 6: Pretty sure we can remove lights on, but should do that in a separate PR if we do.

msadoon (approved):

Library/OptimizelyFeature.swift
singhhari: I'll raise another PR with this, in a moment.

Kickstarter-iOS/Views/Controllers/ProjectPamphletContentViewController.swift
singhhari: Added it into the `ProjectPamphletContentViewControllerTests` file. I'm running an assertion on the `Type` and not any snapshots. Let me know if that's the right location for the test.

justinswart (approved):
too legit

PR #1469, opened 2021-05-11
[NT-1922] Set up view controller and view model skeleton for Root Comments

singhhari:
# 📲 What

We're beginning new feature development on supporting comment threading for projects. This PR introduces a skeleton for the view controller and view model of what will eventually be known as our `RootCommentViewController`/ `RootCommentViewModel`. The files are not doing anything specific at the moment but will be fleshed out over the course of the next few weeks and we build out the functionality and extend it.

# 🤔 Why

As outlined in our [engineering plan](https://kickstarter.atlassian.net/wiki/spaces/EN/pages/1336213588/Native+Threaded+Comments+Engineering+Plan), our web application has supported this functionality for a while now and our users have requested we build this out on mobile. We will eventually wrap this work behind a feature flag and release this into the wild when we feel development and testing deems it production-ready.

# ✅ Acceptance criteria

- [ ] Tests on `RootCommentsViewModelTests` and `RootCommentsViewControllerTests` should pass.
- [ ] Target membership for the aforementioned files are organized correctly.

PR #1474, opened 2021-05-12
Adding Deprecated prefix to older Comments files.

singhhari:
# 📲 What

This PR is a continuation of some work done in [PR 1469](https://github.com/kickstarter/ios-oss/pull/1469). We're adding a `Deprecated` prefix to a handful of old files related to our legacy Comments implementation.

# 🤔 Why

We noticed that some files were left behind in the original PR. This should address the rest of them.


justinswart (approved):
Nice, thanks!

PR #1476, opened 2021-05-14
[NT-1894] Addition of createComment mutation

singhhari:
# 📲 What

Our upcoming work for comment threading on Native relies on a shift from our legacy usage of a V1 REST API to a newer, GraphQL API. This PR includes the addition of a `mutation` for creating a comment on a project or replying to an existing comment. Once our UI is fleshed out, we can apply the mutation in a more practical manner but for now we are simply focusing on getting our networking logic prepared for the remainder of the work.

# 🤔 Why

This mutation is a necessity for the feature. We can't comment on a thread without it, so it's crucial that this work is taken care of earlier in the feature development. It will certainly be expanded a bit further once we can determine how we want to approach the UI.

# 🛠 How

- Added a mutation to our GraphQL API
- Expanded our `ServiceType` object to support the use of the network request
- Created a response envelope class that will return the `body` of the message and its `id` (which may be useful for optimistic posting)
- Wrote some tests

# ✅ Acceptance criteria

- [ ] On staging, use a project you either Backed or Created, replace the IDs in the `CommentsViewModel`s call of the mutation and verify it's appearing on web.
- [ ] Attempt to comment on the root level of the project by omitting the `parentId` argument.
- [ ] Attempt to reply to an existing comment by supplying the `parentId` argument.


KsApi/models/graphql/GraphMutationPostCommentResponseEnvelopeTests.swift
```diff
@@ -0,0 +1,29 @@
...
+    ]
+
+    guard let data = try? JSONSerialization.data(withJSONObject: dictionary, options: []) else {
+      XCTFail("Should have data")
+      return
+    }
+
+    do {
+      let envelope = try JSONDecoder().decode(GraphMutationPostCommentResponseEnvelope.self, from: data)
+      XCTAssertEqual(envelope.createComment.comment.body, "Hello World")
```
singhhari, line 22: Not a fan of calling `envelope.createComment.comment.body` but I don't think we can avoid it, given the structure of the response?

KsApi/models/templates/GraphMutationPostCommentResponseEnvelopeTemplate.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+extension GraphMutationPostCommentResponseEnvelope {
+  internal static let template =
+    GraphMutationPostCommentResponseEnvelope(createComment: .init(comment: .init(
+      body: "Hello World",
+      id: "Q29tbWVudC0zMjY2MjU0MQ=="
+    )))
+}
```
singhhari, line 9: Wrote this to help with mocking the request, in a test, but I don't think I can do any assertions until we have some outputs in the view model to observe?

KsApi/mutations/inputs/PostCommentInput.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+public struct PostCommentInput: GraphMutationInput {
+  let body: String
+
+  /// This is base64 encoded ID of the Project-{projectId}}
+  let commentableId: String
+
+  /// If the comment is on the project level, this is nil. For replies, this is the ID of the comment.
```
singhhari, line 9: Is it more useful to leave this comment here or at the call site of the mutation (in the view model) ?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -28,6 +31,26 @@ public final class CommentsViewModel: CommentsViewModelType,
         .materialize()
     }
     .observeValues { print($0) }
+
+    // FIXME: We need to dynamically supply the IDs when the UI is built.
```
singhhari, line 35: Flagging this for reviewer visibility.

KsApi/mutations/PostCommentMutation.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+public struct PostCommentMutation<T: GraphMutationInput>: GraphMutation {
```
singhhari, line 3: Noticed we aren't testing any of these GraphMutation types. Is that intentional?

KsApi/models/graphql/GraphMutationPostCommentResponseEnvelopeTests.swift
msadoon: Do you need the `CreateComment` struct? Maybe it can be `envelope.comment.body`?

msadoon: Oh wait, I guess you do because the GQL query has a `createComment` object being returned.

KsApi/mutations/PostCommentMutation.swift
```diff
@@ -0,0 +1,22 @@
...
+  var input: T
+
+  public init(input: T) {
+    self.input = input
+  }
+
+  public var description: String {
+    return """
+    mutation ($input: PostCommentInput!) {
```
msadoon, line 12: More just for my own knowledge, but I guess we don't use the DSL (GraphSchema) for mutations?

msadoon (approved):

KsApi/models/graphql/GraphMutationPostCommentResponseEnvelopeTests.swift
singhhari: Yep, I'm relying on Decodable for parsing the response here.

KsApi/mutations/PostCommentMutation.swift
singhhari: I think @justinswart has more historical context here but based on the code, we are not "building" the mutation but instead relying on that `description` property in our Service helper.

KsApi/mutations/inputs/PostCommentInput.swift
justinswart: Think this is good, Xcode should show it during one of those rare autosuggestion moments.

KsApi/ServiceType.swift
```diff
@@ -255,13 +255,17 @@ public protocol ServiceType {
...
     -> SignalProducer<DeprecatedComment, ErrorEnvelope>
 
   /// Posts a comment to an update.
-  func postComment(_ body: String, toUpdate update: Update)
+  func deprecatedPostComment(_ body: String, toUpdate update: Update)
     -> SignalProducer<DeprecatedComment, ErrorEnvelope>
 
+  /// Posts a comment to a project or replies in a thread
+  func postComment(input: PostCommentInput)
+    -> SignalProducer<GraphMutationPostCommentResponseEnvelope, GraphError>
```
justinswart, line 267: So I actually think we should return `Comment` here and do the data transformation in KsApi. In the same way that we've hidden the deserialization of the comments pagination from the app.

justinswart (commented):
Cool looking mostly good, I think the main thing we would want to change is abstract the GraphQL data structure from the VM callsite and from our VM tests and only return `Comment` to the app.

KsApi/MockService.swift
```diff
@@ -134,8 +134,11 @@
 
     fileprivate let incrementVideoStartError: ErrorEnvelope?
 
-    fileprivate let postCommentResponse: DeprecatedComment?
-    fileprivate let postCommentError: ErrorEnvelope?
+    fileprivate let deprecatedPostCommentResponse: DeprecatedComment?
+    fileprivate let deprecatedPostCommentError: ErrorEnvelope?
+
+    fileprivate let postCommentResponse: GraphMutationPostCommentResponseEnvelope?
```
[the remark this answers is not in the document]
justinswart: However since my later suggestion to only return `Comment` it would just be:

```swift
let postCommentResult: Result<Comment, ErrorEnvelope>?
```

KsApi/models/graphql/GraphMutationPostCommentResponseEnvelopeTests.swift
justinswart: You could also override the initializer with an `init(from decoder: Decoder) { ... }` and do some manual drilling down through the nested containers but could go either way.

KsApi/models/templates/GraphMutationPostCommentResponseEnvelopeTemplate.swift
justinswart: You could output a templatized `Comment`.

KsApi/ServiceType.swift
justinswart: Oh I've also moved away from `GraphError` in other places and just returned `ErrorEnvelope`. The idea is also that once we implement Apollo all of this implementation detail is hidden and the `ServiceType` interface won't have to change in the places that we've already done this.

KsApi/models/templates/GraphMutationPostCommentResponseEnvelopeTemplate.swift
singhhari: Added an output thats emitting the template on viewDidLoad and wrote a small test for it.

KsApi/models/graphql/GraphMutationPostCommentResponseEnvelopeTests.swift
singhhari: Updated this.

KsApi/mutations/PostCommentMutation.swift
singhhari: Added some basic assertions on the properties.

KsApi/MockService.swift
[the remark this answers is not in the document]
singhhari: Ah I see. Made these updates with one difference: `GraphError` instead of `ErrorEnvelope` because we calling `applyMutation` which uses the `GraphError` type.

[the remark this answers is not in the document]
justinswart: There's a workaround for that 😄 
https://github.com/kickstarter/ios-oss/blob/master/KsApi/Service.swift#L179

KsApi/MockService.swift
```diff
@@ -1175,17 +1184,28 @@
...
+      } else if let comment = self.deprecatedPostCommentResponse {
         return SignalProducer(value: comment)
       }
       return .empty
     }
 
-    func postComment(_: String, toUpdate _: Update) -> SignalProducer<DeprecatedComment, ErrorEnvelope> {
+    func postComment(input _: PostCommentInput)
+      -> SignalProducer<GraphMutationPostCommentResponseEnvelope, GraphError> {
       if let error = self.postCommentError {
```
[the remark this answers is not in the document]
singhhari: I updated this to behave similarly to some other mutations we have. We return the error if it exists; otherwise, we return the successful results value or a template of it.

KsApi/models/graphql/GraphMutationPostCommentResponseEnvelope.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public struct GraphMutationPostCommentResponseEnvelope: Decodable {
```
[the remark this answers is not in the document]
singhhari: Yeah, good idea. It's an awful lot of typing to find the file lol

KsApi/MockService.swift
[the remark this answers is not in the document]
singhhari: Thanks, updated.

KsApi/mutations/inputs/PostCommentInput.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+public struct PostCommentInput: GraphMutationInput {
+  let body: String
+
+  /// This is base64 encoded ID of the Project-{projectId}}
```
justinswart, line 6: Do we have confirmation on how to handle this?

KsApi/models/graphql/GraphMutationPostCommentEnvelope.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+public struct GraphMutationPostCommentEnvelope: Decodable {
```
justinswart, line 3: Let's make this type `internal` to be sure it's not exposed to the app.

KsApi/mutations/inputs/PostCommentInput.swift
singhhari: Yeah @leighdouglas verified that on Android they have a reference to the graphID and can pass that in as the commentableId. I think the same thing should work for us.

KsApi/models/graphql/GraphMutationPostCommentEnvelope.swift
singhhari: Updated

justinswart (approved):
Cool, think we're gtg here! I think it's fine if we need to tweak it during implementation but this should be good 👍

PR #1478, opened 2021-05-14
[NT-1893][NT-1919] UI for Comment Composer, Add Comment Composer to Accessory View

tobitech:
# 📲 What

Create a Comment Composer component that would be used to post comments to a specified project in the app.

# 🤔 Why

This is part of a new commenting experience we're adding to the application.

# 👀 See

Simulator Screen Shot - iPhone SE (2nd generation) - 2021-05-14 at 19 52 34
Simulator Screen Shot - iPhone SE (2nd generation) - 2021-05-14 at 19 53 18

# ✅ Acceptance criteria

1. Launch the app
2. Tap on any project card
3. Tap on comment section to show the new comments screen
4. Tap on the input area where you see "Leave a comment..."
5. As the keyboard comes up, the comment composer should slide up with the keyboard
6. Enter some texts in the new comment input area. The input area should grow as you type in more texts and should stop growing when the maximum height is reached and text view should be scrollable.
7. Pan down the keyboard and the composer should slide downwards along with the keyboard

# ⏰ TODO

- [x] Find a way around the corner radius issue for the input container view
- [x] Limit input characters to 9000
- [x] Use appropriate colours for the interface when they are ready from design team.
- [ ] Replace raw strings with type-safe versions when copy is ready
- [x] Add Unit tests and screenshots tests.


Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,129 @@
...
+    UIView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private let inputContainerView: CommentInputContainerView = {
+    CommentInputContainerView()
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private let onlyBackersLabel: UILabel = {
```
msadoon, line 35: Should we bind this to some view model logic around if they are a backer/not?

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,129 @@
...
+    UILabel(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+      |> \.isHidden .~ true
+  }()
+
+  private lazy var avatarImageView = {
```
msadoon, line 41: Same here, should we bind this to some view model logic on their actual avatar image?

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -8,16 +8,34 @@ import UIKit
 internal final class CommentsViewController: UITableViewController {
   // MARK: - Properties
 
+  private lazy var commentComposer: CommentComposerView = {
```
msadoon, line 11: Can this raw value of height be referenced internally through an `enum` like with the other views?

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -8,16 +8,34 @@ import UIKit
...
+  }()
+
   fileprivate let viewModel: CommentsViewModelType = CommentsViewModel()
 
   // MARK: - Lifecycle
 
   internal override func viewDidLoad() {
     super.viewDidLoad()
 
+    tableView.keyboardDismissMode = .interactive
```
msadoon, line 24: This is cool! I didn't know this existed to make the keyboard disappear on scroll!

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -8,16 +8,34 @@ import UIKit
...
+    tableView.estimatedRowHeight = UITableView.automaticDimension
+    tableView.register(UITableViewCell.self, forCellReuseIdentifier: "cellId")
```
msadoon, line 26: Will probably need to replace this after the [cell card](https://kickstarter.atlassian.net/browse/NT-1892) is merged.

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -28,3 +46,19 @@ internal final class CommentsViewController: UITableViewController {
 
   internal override func bindViewModel() {}
 }
+
+extension CommentsViewController {
```
msadoon, line 50: Cool! Just putting a note here to replace this with [cell card](https://kickstarter.atlassian.net/browse/NT-1892)

msadoon (commented):
This looks great!
I made some comments, mostly for my own understanding.

Here are a few things I noticed while running the app:
First thing is the comment box looks a little bigger than intended in the designs:


The second thing is that when we resize our accessibility text to large, we lose the Post button text:
Simulator Screen Shot - iPhone 8 - 2021-05-17 at 10 59 13

Looking forward to reviewing the rest!

justinswart (changes requested):
@tobitech looking pretty good so far! The only issue is that we use `UIStackView`s to construct all of our views because it makes it easier to change things later if we need to. Is there a reason you chose not to use stack views here? Otherwise I'm requesting that we rework this to use them please, thanks!

Kickstarter-iOS/Views/CommentInputContainerView.swift
```diff
@@ -0,0 +1,157 @@
...
+    UIButton(type: .system)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+      |> \.isHidden .~ true
+  }()
+
+  // MARK: - Lifecycle
+
+  override init(frame: CGRect) {
+    super.init(frame: frame)
+
```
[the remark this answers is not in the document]
tobitech: I've seen a pattern in the project where setting up view hierarchy and constraints are placed in separate methods. That's why I did it that way.

singhhari:
A couple of observations:
- When I select a project and enter the comments the avatar is occasionally not rounded appropriately (diamond oval shape) 
- When I select a project I've backed and enter the comments the "Only backers can leave comments" label is showing instead of the textview
- When I select a project, enter the comments, swipe the page view controller and return back to the comments, the avatar/textview screen is no longer showing/accessible
- The cells are using a grey selection style

singhhari (commented):
A couple of observations:

- When I select a project and enter the comments the avatar is occasionally not rounded appropriately (diamond oval shape)
- When I select a project I've backed and enter the comments the "Only backers can leave comments" label is showing instead of the textview (I think this is bc you are hard-coding the Bool, I just read the inline comment)
- When I select a project, enter the comments, swipe the page view controller and return back to the comments, the avatar/textview screen is no longer showing/accessible
- The cells are using a grey selection style

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,140 @@
...
+    CircleAvatarImageView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private let onlyBackersLabel: UILabel = {
+    UILabel(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  // This will be replaced with a ViewModel ouput binding
```
justinswart, line 41: Let's prefix all of these sorts of TODO's with `// TODO:` so that they're easier to find later on.

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,140 @@
...
+    |> \.layoutMargins .~ .init(all: Styles.grid(3))
+    |> \.isLayoutMarginsRelativeArrangement .~ true
+}
+
+private let onlyBackersLabelStyle: LabelStyle = { label in
+  label
+    |> \.textColor .~ .ksr_support_400
+    |> \.font .~ UIFont.ksr_body(size: 15.0)
+    // To be replaced with a type-safe string when copy is available.
+    |> \.text .~ "Only backers can leave comments"
```
justinswart, line 138: For strings that are not yet translated it's good practice to use `localizedString(key:defaultValue:)` directly (which is what all the `Strings.*` functions use internally. That way these strings are easy to find later on and even if we miss one we should still get the translations correctly - but let's not miss one 😄

```swift
localizedString(key: "Only_backers_can_leave_comments", value: "Only backers can leave comments")
```

Kickstarter-iOS/Views/CommentInputTextView.swift
```diff
@@ -0,0 +1,51 @@
+import Library
+import Prelude
+import UIKit
+
+final class CommentInputTextView: UITextView {
+  // MARK: - Properties
+
+  let maxHeight: CGFloat = 200
```
justinswart, line 8: How is this value determined?

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,140 @@
...
+import UIKit
+
+private enum Layout {
+  enum Border {
+    static let height: CGFloat = 1.0
+  }
+
+  enum Avatar {
+    static let diameter: CGFloat = 44.0
+    static let margin: CGFloat = 18.0
```
justinswart, line 12: In the past we have had a concept of a 6pt grid for our designs, this allowed us to specify sizes using `Styles.grid(3)` for example to get value like this of `18`. I'm just noting this here because I'm not sure if these current designs are also using this grid, we might want to try to get the closest grid sizes to the design 🤔 My concern is that most of our current UI uses the `Styles.grid()` function and using the pt sizes directly like this would be a deviation.

cc @singhhari what do you think?

justinswart (commented):
Thanks for reworking to use stack views, I think we're getting closer!

Is there a separate ticket to add VMs and tests? I do see a lot of opportunities for testing here.

Kickstarter-iOS/Views/CommentInputTextView.swift
tobitech: It is part of the acceptance criteria for the ticket

Kickstarter-iOS/Views/CommentComposerView.swift
tobitech: Yeah, I also contemplated what approach to use because the `Styles.grid()` system has its units in multiples of 6 but the Figma design doesn't follow that convention.

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -119,6 +119,17 @@ final class CommentComposerView: UIView {
   }
 }
 
+// MARK: - Styles
+
+private let rootStackViewStyle: StackViewStyle = { stackView in
```
msadoon, line 124: May be duplicated from [CommentCellStyles.swift](https://github.com/kickstarter/ios-oss/pull/1480/commits/1b3df30485d07ec2aaad345fa68d0eb1f958ee95#diff-d5a4fea3497c0722b485362398bc4c619a47f187f5a9c47850c237c37ed5638cR5-R12)

Kickstarter-iOS/Views/CommentInputContainerView.swift
```diff
@@ -158,7 +125,26 @@ extension CommentInputContainerView: UITextViewDelegate {
...
+private let rootStackViewStyle: StackViewStyle = { stackView in
+  stackView
+    |> \.axis .~ .horizontal
+    |> \.distribution .~ .fill
+    |> \.alignment .~ .bottom
+    |> \.spacing .~ Styles.grid(1)
+    |> \.layoutMargins .~ .init(topBottom: Styles.grid(2), leftRight: Styles.grid(3))
+    |> \.isLayoutMarginsRelativeArrangement .~ true
+}
```
msadoon, line 147: Yea just double check with `CommentCellStyles` [here](https://github.com/kickstarter/ios-oss/pull/1480/commits/1b3df30485d07ec2aaad345fa68d0eb1f958ee95#diff-d5a4fea3497c0722b485362398bc4c619a47f187f5a9c47850c237c37ed5638cR5-R30).
Which ever PR gets merged first I suppose. haha.

Kickstarter-iOS/Views/CommentComposerView.swift
tobitech: That's why I made it `private`.

justinswart: Yep all the styles should be private to their views and VC's, if they're shared then we should move them to a shared location 👍

Kickstarter-iOS/Views/CommentComposerView.swift
justinswart: Yeah I'm just wondering if it wouldn't be sufficient to get it _close enough_ 🤔

justinswart:
> Is there a separate ticket to add VMs and tests? I do see a lot of opportunities for testing here.

@tobitech you might have missed this question ☝️

tobitech:
> > Is there a separate ticket to add VMs and tests? I do see a lot of opportunities for testing here.
> 
> @tobitech you might have missed this question ☝️

Oh yeah, I missed it. There is no separate ticket for that it will be part of this PR. What I have is a ViewModel specifically for the composer component and tests for it. I will be pushing that shortly

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,212 @@
+import Library
+import Prelude
+import UIKit
+
+private enum Layout {
+  enum Border {
+    static let height: CGFloat = 1.0
+  }
+
+  enum Avatar {
```
singhhari, line 10: small note by alphabetize?

Kickstarter-iOS/Views/CommentInputContainerView.swift
```diff
@@ -0,0 +1,128 @@
+import Library
+import Prelude
+import UIKit
+
+private enum Layout {
+  enum Container {
+    static let radius: CGFloat = 22
+  }
+
+  enum Button {
```
singhhari, line 10: alphabetize

Kickstarter-iOS/Views/CommentInputContainerView.swift
```diff
@@ -0,0 +1,128 @@
...
+  override var intrinsicContentSize: CGSize {
+    return .zero
+  }
+
+  // MARK: - Views
+
+  private func configureViews() {
+    _ = self
+      |> \.autoresizingMask .~ .flexibleHeight
+      |> \.layoutMargins .~ .init(left: 18)
```
singhhari, line 54: I thought we used the `Grid` syntax for layouts?

Kickstarter-iOS/Views/CommentInputTextView.swift
```diff
@@ -0,0 +1,51 @@
...
+
+  required init?(coder _: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  override var intrinsicContentSize: CGSize {
+    var size = super.intrinsicContentSize
+    if size.height > self.maxHeight {
+      size.height = self.maxHeight
+      isScrollEnabled = true
```
singhhari, line 24: self.isScrollEnabled

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -5,19 +5,49 @@ import Prelude
...
+  enum Composer {
+    static let originalHeight: CGFloat = 80.0
+  }
+}
+
 internal final class CommentsViewController: UITableViewController {
   // MARK: - Properties
 
+  private lazy var commentComposer: CommentComposerView = {
+    let frame = CGRect(x: 0, y: 0, width: view.frame.width, height: Layout.Composer.originalHeight)
```
singhhari, line 18: self.view.frame.width

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -0,0 +1,113 @@
...
+public enum CommentComposerConstant {
+  public static let characterLimit: Int = 9_000
+}
+
+public protocol CommentComposerViewModelInputs {
+  /// Call to configure composer avatar and input area visibility
+  func configure(with data: CommentComposerViewData)
+
+  /// Call when the comment text changes.
+  func bodyTextDidChange(_ text: String)
```
singhhari, line 16: alphabetize

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -0,0 +1,113 @@
+import Foundation
+import KsApi
+import ReactiveSwift
+
+public typealias CommentComposerViewData = (avatarURL: URL?, isBackingProject: Bool)
+
+public enum CommentComposerConstant {
+  public static let characterLimit: Int = 9_000
```
singhhari, line 8: This is fine but should probably be accompanied by an inline comment about where the number is coming from.

tobitech: How about `// From feature requirements`?

singhhari: Connected on Slack. Going to resolve this.

singhhari (approved):

justinswart (approved):

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -0,0 +1,212 @@
...
+
+  public func configure(with data: CommentComposerViewData) {
+    self.viewModel.inputs.configure(with: data)
+  }
+
+  /*
+   Call this to clear body text.
+   */
+  public func clearOnSuccess() {
+    self.viewModel.inputs.bodyTextDidChange("")
```
justinswart, line 74: This would be a separate VM input, `clearOnSuccess()` we can address then when it's implemented but then let's add a `TODO` here.

justinswart (approved):
Cool just one more comment and then I think this is gtg 👍

Kickstarter-iOS/Views/CommentComposerView.swift
tobitech: Okay

PR #1479, opened 2021-05-18
[NT-1889] Add Domains For Universal Links From Emails

justinswart:
# 📲 What

Adds new domains that we will use for universal links in emails.

**Note:** Blocked until the domains are correctly configured.

# 🤔 Why

We're using these new domains when sending emails from Braze.

# 🛠 How

Added the 3 new domains to our entitlements and `Navigation.swift` and updated tests.

# ✅ Acceptance criteria

Once these domains have been configured with the `apple-app-site-association` file then we should be able to tap links in emails that have any of the following domains.

- [ ] `ea.kickstarter.com`
- [ ] `me.kickstarter.com`
- [ ] `clicks.kickstarter.com`

### DIFF
--- a/Library/NavigationTests.swift
+++ b/Library/NavigationTests.swift
@@ -230,6 +230,11 @@ public final class NavigationTests: XCTestCase {
 
     XCTAssertEqual(
       .emailClick,
+      Navigation.match(URL(string: "https://click.e.kickstarter.com/wf/click?upn=deadbeef")!)
+    )
+
+    XCTAssertEqual(
+      .emailClick,
       Navigation.match(URL(string: "https://clicks.kickstarter.com/wf/click?upn=deadbeef")!)
     )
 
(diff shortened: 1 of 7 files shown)


Library/NavigationTests.swift
```diff
@@ -218,6 +218,21 @@ public final class NavigationTests: XCTestCase {
...
+      .emailClick,
+      Navigation.match(URL(string: "https://ea.kickstarter.com/wf/click?upn=deadbeef")!)
+    )
+
+    XCTAssertEqual(
+      .emailClick,
+      Navigation.match(URL(string: "https://clicks.kickstarter.com/wf/click?upn=deadbeef")!)
+    )
+
     XCTAssertEqual(
```
msadoon, line 236: Not technically part of this PR, but would be nice to also test `click.e` subdomain.

msadoon (approved):
Good to go!

Library/NavigationTests.swift
justinswart: Good eye! Added it 👍

PR #1483, opened 2021-05-26
[NT-1976] UX - Not logged in / not backer for Comment Card

singhhari:
# 📲 What

There is a small piece of conditional logic being applied to the reply button functionality on the Comment Card interface. In particular:
- If a user is logged out, the reply button should be hidden
- If a user is logged in and not backing, the reply button should be hidden
- If a user is logged in and a backer, the reply button should be visible

This PR introduces that logic and also lays the groundwork for eventually doing the same for the flag button (not included here because it will be put behind a feature flag in a later PR).

# 🤔 Why

Kickstarter allows creators, backers and admins the ability to leave comments on a project. On that note, when logged out, there's no way to determine if a user is a backer. The highlights in the `What` section describe how we, both, introduced this logic and tested it.

# 👀 See

Simulator Screen Shot - iPhone 12 Pro - 2021-05-26 at 17 40 06

# ✅ Acceptance criteria

Manual testing through a simulator will be a bit incomplete, given we are not rendering live data from staging. The button hiding condition of either being logged out, or being logged in and not a backer, can, however, be tested with unit and snapshot tests. Reviewing the snapshots and running tests should be sufficient to determine if the PR is good to go.


Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -51,30 +54,47 @@ public final class CommentCellViewModel:
...
       }
 
     self.authorBadge = Signal.merge(
       badge,
       badge.takeWhen(self.bindStylesProperty.signal)
     )
+
+    let isLoggedOut = self.commentUserAndProject.signal
+      .ignoreValues()
+      .map { _ in AppEnvironment.current.currentUser }
```
singhhari, line 69: @moderateepheezy when you originally wrote this you passed the user in to the `configureWith` method but I'm not sure I should be using that or the `currentUser` environmental var?

cc: @justinswart

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -51,32 +54,49 @@ public final class CommentCellViewModel:
...
+
+    let isBacker = self.commentUserAndProject.signal
+      .skipNil()
+      .map { _, _, project in project }
+      .map(userIsBackingProject)
+
+    self.replyButtonIsHidden = Signal.combineLatest(isLoggedOut, isBacker)
+      .map { isLoggedOut, isBacker in
+        // If the user is logged out OR logged in AND not a backer, hide replyButton
+        isLoggedOut || (!isLoggedOut && isBacker == false)
```
justinswart, line 80: what about:
```swift
guard isLoggedOut else {
  return !isBacker
}

return true
```
or:
```swift
let isNotABacker = self.commentUserAndProject.signal
  .skipNil()
  .map { _, _, project in project }
  .map(userIsBackingProject)
  .negate()

self.replyButtonIsHidden = Signal.combineLatest(isLoggedOut, isNotABacker)
  .map { isLoggedOut, isNotABacker in isLoggedOut || isNotABacker }
```

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -41,28 +46,65 @@ internal final class CommentCellViewModelTests: TestCase {
...
+
+    withEnvironment(currentUser: user) {
+      self.vm.inputs.configureWith(comment: .template, user: user, project: project)
+
+      self.replyButtonIsHidden
+        .assertValue(false, "The replyButton is not hidden because the user is a backer AND logged in.")
+    }
+  }
+
+  func testOutputs_replyButtonIsHidden_IsBacker_False_IsLoggedIn_False() {
```
justinswart, line 63: Think you want `IsLoggedOut` in the name of this test.

justinswart (commented):
Think it's looking good some small comments

Kickstarter-iOS/Views/CommentCellHeaderStackView.swift
```diff
@@ -71,8 +71,8 @@ internal final class CommentCellHeaderStackView: UIStackView {
 
   // MARK: - Configuration
 
-  internal func configureWith(comment: Comment, user: User?) {
-    self.viewModel.inputs.configureWith(comment: comment, user: user)
+  internal func configureWith(comment: Comment, user: User?, project: Project) {
```
msadoon, line 74: Does `CommentCellHeaderStackView`need the project?

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
```diff
@@ -16,10 +16,47 @@ internal final class CommentsViewControllerTests: TestCase {
...
-        let controller = CommentsViewController()
+      withEnvironment(currentUser: nil, language: language) {
+        let controller = CommentsViewController.configuredWith(project: .template)
+        let (parent, _) = traitControllers(device: .phone4_7inch, orientation: .portrait, child: controller)
+        parent.view.frame.size.height = 1_100
+
+        self.scheduler.run()
+
+        controller.tableView.layoutIfNeeded()
+        controller.tableView.reloadData()
```
msadoon, line 29: Just for my own curiousity, is `layoutIfNeeeded` added as a extra layer of precaution to ensure the table view gets refreshed accurately. Just because I assume `reloadData()` also invalidates the layout.

Library/ViewModels/CommentCellViewModel.swift
msadoon: Could be `loggedOut || !isBacker`

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -41,28 +46,65 @@ internal final class CommentCellViewModelTests: TestCase {
     self.postTime.assertValueCount(1, "The relative time of the comment is emitted.")
   }
 
+  func testOutputs_replyButtonIsHidden_IsBacker_True_IsLoggedIn_True() {
```
msadoon, line 49: `replyButtonIsNotHidden`

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
```diff
@@ -16,10 +16,47 @@ internal final class CommentsViewControllerTests: TestCase {
...
+
+        FBSnapshotVerifyView(parent.view, identifier: "Comments - lang_\(language)")
+      }
+    }
+  }
+
+  func testView_CurrentUser_LoggedIn_IsBacking_False() {
```
msadoon, line 56: great job on tests!

Library/ViewModels/CommentCellViewModel.swift
msadoon: Noticed this too, I guess its' good because it's dependency injected, making it more testable?
But then again we have `withEnvironment` to inject the `User` so maybe it is redundent.

singhhari: @moderateepheezy ^^

msadoon (approved):
Made some minor comments but also tested this on staging simulator and it works as outlined. Good to go! :)

Library/ViewModels/CommentCellViewModel.swift
singhhari: Nice, going with the latter. I think it reads well and I can delete the comment I left above it, as a result.

Kickstarter-iOS/Views/CommentCellHeaderStackView.swift
singhhari: Good point. I don't think any of them other than `CommentCell` needs it. Going to remove that parameter.

moderateepheezy: You either make `Project` optional or you create separate input for the other `Cell` and  `CommentCellHeaderStackViewneed `.

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
singhhari: I copied the original template to write the subsequent tests and now realize it passes without those calls. Removed the lines. Good catch!

Library/ViewModels/CommentCellViewModelTests.swift
singhhari: Ah so in this case I'm following the existing convention where we list the name of the output use an underscore delimiter to signify what properties are being tested with it.

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
singhhari: Always nice to get the coverage right on the first go hehe

Kickstarter-iOS/Views/CommentCellHeaderStackView.swift
singhhari: Yep, updated!

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
moderateepheezy: Yes. Basically, I think it's saying update immediately even when reloadData() is called. This is useful because `layoutIfNeeded()` is a synchronous task.

singhhari: I ended up removing those two lines on all the tests because it's passing without it. I think `scheduler` handles manually advancing synchronous tasks and we're using it here.

Library/ViewModels/CommentCellViewModel.swift
moderateepheezy: @msadoon is right about the injection part , but I think we can move it to the CommentCellViewModel so we can inject from there, I only added it to the main vm because that's what I have seen done in the code mostly.

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
justinswart: We've had to add these lines sometimes when snapshot tests keep failing, causing a layout pass and reloading has at times caused tests to work reliably that were otherwise flakey or kept failing on comparison. I think @moderateepheezy and I tried it here because he was seeing failures but if it's no longer needed that's good too 👍

Library/ViewModels/CommentCellViewModel.swift
justinswart: Yeah we can rely on our environment for all dependency injection so unless there is a reason I'm not aware of to want to configure and test this cell in isolation from the environment then it is probably not necessary here.

justinswart (approved):

Library/ViewModels/CommentCellViewModel.swift
singhhari: Pushed up a commit with the refactor.

PR #1484, opened 2021-05-27
[NT-1938] UI comment card with threads

moderateepheezy:
# 📲 What

`View replies` CTA is added to the Comment UI Card. The `View replies` CTA is only enabled for successful comments.

# 🤔 Why

Every root comments can be replied. User needs to be able to navigate to all replies to a comment with a CTA

# 👀 See

Simulator Screen Shot - Clone 1 of iPhone 12 Pro Max - 2021-05-27 at 20 15 42

# ✅ Acceptance criteria

1.  Navigate to the comment section of a project
2.  Validate that only successful comments have a `View Replies` CTA
3. Validate that `Reply` button is hidden if you are not a backer.
4.  The ticket mention that: 
      > Also, we need to support yet another variation of the Comment Card design, to display replies:
    
    This speaks to Comment Card  being reusable.

# Note
This PR does will not include
> Left indentation – should this be part of the card or the view it sits in?


moderateepheezy:
Will be closing this as it's already included here https://github.com/kickstarter/ios-oss/compare/NT-1939-thread-view-navigation

PR #1489, opened 2021-05-31
[NT-1935] Fetch Comment Replies Query

justinswart:
# 📲 What

Adds the query, types and deserialization for fetching a `Comment`'s reply `Comment`s.

# 🤔 Why

This is continuation of the work to bring threaded comments to iOS.

# 🛠 How

- Updated `GraphSchema`.
- Added `CommentRepliesEnvelope`.
- Added deserialization and model adaptation.
- Added tests.

# ✅ Acceptance criteria

- [x] Verify that I've used the correct cursor and paging behavior.
- [x] Use the query function to generate a query with real IDs and verify in GraphiQL that it works.


KsApi/models/CommentRepliesEnvelope.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+import ReactiveSwift
+
+public struct CommentRepliesEnvelope: Decodable {
```
msadoon, line 4: It's good that the `RepliesEnvelope` is separate from the `CommentsEnvelope` because it's a completely different screen and a brand new query even though the results are slightly different.

KsApi/models/graphql/GraphCommentRepliesEnvelope.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    let repliesContainer = try commentContainer
+      .nestedContainer(keyedBy: CodingKeys.self, forKey: .replies)
+
+    var edges = try repliesContainer
+      .nestedUnkeyedContainer(forKey: .edges)
+
+    var replies: [GraphComment] = []
+
+    while !edges.isAtEnd {
```
msadoon, line 40: neat, `isAtEnd`

KsApi/queries/CommentsQueries.swift
```diff
@@ -15,7 +25,11 @@ public func commentsQuery(
       .comments(
         args,
         .edges(
-          .node(GraphComment.baseQueryProperties) +| []
+          .node(
+            GraphComment.baseQueryProperties.op(
+              .replies([], Connection<Query.Comment>.totalCount +| []) +| []
```
msadoon, line 30: wow!

singhhari (commented):
code looks good but i can't seem to get this to work on graphiql when supplying a commentableID

KsApi/queries/CommentsQueries.swift
```diff
@@ -28,3 +42,38 @@ public func commentsQuery(
...
+public func commentRepliesQuery(
+  withCommentId id: String,
+  last: Int = Query.defaultPaginationCount,
+  before cursor: String? = nil
+) -> NonEmptySet<Query> {
+  let args = Set([cursor.flatMap(QueryArg<Never>.before), .last(last)].compact())
+
+  return Query.comment(
+    id: id,
+    GraphComment.baseQueryProperties.op(
```
msadoon, line 65: is `op` like a query in another query? I noticed it's used above as well to get the replies from within each `Comment` after the `baseProperties` get called.

msadoon (approved):
All looks good to me!

KsApi/queries/CommentsQueries.swift
justinswart: It allows us to concatenate the queries: https://github.com/kickstarter/Kickstarter-Prelude/blob/master/Prelude/NonEmpty.swift#L92

PR #1491, opened 2021-05-31
[NT-1896] UX - Post new comment

tobitech:
# 📲 What

When a user taps on the Post button on the composer after entering some text a new comment is created and added to the top of the list of comments for the project that was selected. If they aren't already at the top of the list the table view is scrolled to the top where they can see the new comment. A graph mutation request is also sent to post the comment to the server.

# 🤔 Why

Implementation of the interaction responsible for posting a new comment. We want users to  have an optimistic comment posting experience.

# 👀 See

https://user-images.githubusercontent.com/8595853/120249839-12108700-c274-11eb-9573-1dba5d172225.mov

# ✅ Acceptance criteria

1. Launch the app and ensure you are logged in
2. Select a project you have already backed. You should be able to enter text in the commend composer and the post button should be visible
3. Tap the post button. You should immediately see a new comment on the list with the text you typed.
4. If you had scrolled down, posting a new comment should scroll the list to the top for you to see the new comment you just posted.
5. The keyboard should also be dismissed.
6. If you navigate back to the project details and back to the comments screen, you should see your new comment at the top of the list.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -11,8 +11,8 @@ public protocol CommentsViewModelInputs {
   /// Call with a `Comment` when it is selected.
   func didSelectComment(_ comment: Comment)
 
-  /// Call when the User is posting a comment or reply.
-  func postCommentButtonTapped()
+  /// Call when the User is posting a comment or reply
+  func postCommentButtonTapped(_ text: String)
```
justinswart, line 15: Rename: `commentComposerDidSubmitText(_ text: String)`

singhhari (commented):
A couple of observations:
- ask Paulo if we should be supporting `return` to post a comment. As of right now the user would have to manually hit the `Post` button.
- There's a decoding error after the successful posting of a comment. for the `created_at` key.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -25,8 +25,11 @@ public protocol CommentsViewModelInputs {
 }
 
 public protocol CommentsViewModelOutputs {
-  /// Emits a CommentComposerViewData object that determines the avatarURL and whether the user is a backer.
-  var configureCommentComposerViewWithData: Signal<CommentComposerViewData, Never> { get }
+  /// Emits data to configure comment composer view.
+  var commentComposerData: Signal<CommentComposerViewData, Never> { get }
```
justinswart, line 29: Rename: `configureCommentComposerViewWithData`. This was its name before, not sure why the rename.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -39,6 +42,12 @@ public protocol CommentsViewModelOutputs {
 
   /// Emits a list of `Comment`s and the `Project` to load into the data source.
   var loadCommentsAndProjectIntoDataSource: Signal<([Comment], Project), Never> { get }
+
+  /// Emits when a comment was successfully posted.
+  var postCommentSuccessful: Signal<Comment, Never> { get }
```
justinswart, line 47: Combine this and the error output to be `var showBannerWithMessageAndType: Signal<(String, MessageBannerType), Never>`.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -73,6 +82,38 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+
+        guard let user = currentUser else {
+          return (nil, isBacker)
+        }
+
+        let url = URL(string: user.avatar.medium)
+        return (url, isBacker)
+      }
+
+    self.isCommentComposerHidden = currentUser.signal
```
justinswart, line 114: Rename this output to be `commentComposerViewHidden`.

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -111,7 +126,7 @@ internal final class CommentsViewController: UITableViewController {
         self?.tableView.reloadData()
       }
 
-    self.viewModel.outputs.configureCommentComposerViewWithData
+    self.viewModel.outputs.commentComposerData
```
singhhari, line 129: I don't think we should be observing the same output in two separate places

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -106,43 +145,31 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+    let optimisticPostComment = Signal.combineLatest(
+      initialProject,
+      currentUser.skipNil()
+    ).takePairWhen(self.postCommentButtonTappedProperty.signal.skipNil())
+      .map(unpack)
+      .map(Comment.optimisticComment)
+
+    self.postCommentSubmitted = optimisticPostComment.ignoreValues()
+
+    let commentsWithOptimisticPostComment = Signal
```
justinswart, line 157: Shorter version:

```swift
let commentsWithOptimisticPostComment = Signal.combineLatest(comments, optimisticPostComment)
  .map { loadedComments, optimisticComment in
    [optimisticComment] + loadedComments
  }
```

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
tobitech: Oh yeah, I must have missed that from merging from master. Thanks for spotting that.

KsApi/models/Comment.swift
```diff
@@ -66,6 +66,30 @@ extension Comment: Decodable {
   }
 }
 
+extension Comment {
+  public static func optimisticComment(
```
singhhari, line 70: Should probably rename this to be a more direct representation of what the method is doing

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -73,6 +82,38 @@ public final class CommentsViewModel: CommentsViewModelType,
         )
       }
 
+    let postCommentEvent = initialProject
+      .takePairWhen(self.postCommentButtonTappedProperty.signal.skipNil())
+      .switchMap { project, comment in
+        AppEnvironment.current.apiService
+          .postComment(input: .init(
+            body: comment,
+            commentableId: "Project-\(project.id)".toBase64()
```
justinswart, line 91: Please use `project.graphID` here.

justinswart (changes requested):
Cool, first set of comments - just bunch of renaming to be in line with our conventions, will review more closely on second pass.

Library/ViewModels/CommentsViewModel.swift
tobitech: Okay

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -138,34 +153,35 @@ internal final class CommentsViewController: UITableViewController {
...
 
-// MARK: - UITableViewDelegate
-
-extension CommentsViewController {
-  override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
-    guard let comment = self.dataSource.comment(at: indexPath) else { return }
-
-    self.viewModel.inputs.didSelectComment(comment)
-  }
+private let tableViewStyle: TableViewStyle = { tableView in
```
msadoon, line 181: `//MARK: Styles`

KsApi/models/Comment.swift
msadoon: Maybe something like `failableComment`

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -106,43 +145,31 @@ public final class CommentsViewModel: CommentsViewModelType,
       }
     )
 
-    let commentsAndProject = Signal.combineLatest(comments, initialProject)
+    let optimisticPostComment = Signal.combineLatest(
+      initialProject,
+      currentUser.skipNil()
+    ).takePairWhen(self.postCommentButtonTappedProperty.signal.skipNil())
```
msadoon, line 151: `takePairWhen` cool!

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -106,43 +145,31 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+
+    let commentsWithOptimisticPostComment = Signal
+      .combineLatest(comments, optimisticPostComment)
+      .map { loadedComments, optimisticComment -> [Comment] in
+        var comments = loadedComments
+        comments.insert(optimisticComment, at: 0)
+        return comments
+      }
+
+    let commentsAndProject = Signal.combineLatest(
```
msadoon, line 165: A few questions: 
1. Once the users' comment successfully gets posted, how do we remove the optimistic comment or do we not care about that? Which leads me to wonder...
2. After the comment has succeeded, refreshing the comments might cause an issue where `optimisticComment` is always added to the comments (even with the just posted comment in there already) because `optimisticComment`'s latest value will be emitted (ie. `combineLatest` will send its last value, which is comment already in `comments`).

msadoon: I haven't tested it, so let me know if I'm misunderstanding somehow :)

msadoon: Also, probably not part of this PR, but it would be helpful to add a `TODO: // What happens if comment posted fails`, which ties into a ticket regarding that comment state @moderateepheezy

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,404 @@ internal final class CommentsViewModelTests: TestCase {
...
-        self.loadCommentsAndProjectIntoDataSourceProject.assertDidNotEmitValue()
-        self.isCommentsLoading.assertDidNotEmitValue()
-      }
+    withEnvironment(
+      apiService: MockService(fetchCommentsEnvelopeResult: .success(envelope)),
+      currentUser: .template
+    ) {
+      self.vm.inputs.viewDidLoad()
+
+      self.scheduler.advance()
```
msadoon, line 489: Before `advancing` test that the optimistic comment is not returned by `loadCommentsAndProjectIntoDataSourceComments`

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,404 @@ internal final class CommentsViewModelTests: TestCase {
...
+      self.loadCommentsAndProjectIntoDataSourceComments.assertDidNotEmitValue()
+      self.loadCommentsAndProjectIntoDataSourceProject.assertDidNotEmitValue()
+      self.isCommentsLoading.assertDidNotEmitValue()
+    }
+  }
 
-        self.loadCommentsAndProjectIntoDataSourceComments.assertDidNotEmitValue()
-        self.loadCommentsAndProjectIntoDataSourceProject.assertDidNotEmitValue()
-        self.isCommentsLoading.assertDidNotEmitValue()
+  func testOptimisticComment() {
```
msadoon, line 478: `testFailableComment_WhenPosted_CommentsContainFailableComment`

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,404 @@ internal final class CommentsViewModelTests: TestCase {
...
+      let optimisticComment = Comment.optimisticComment(project: .template, user: .template, body: bodyText)
+
+      XCTAssertEqual(
+        self.loadCommentsAndProjectIntoDataSourceComments.values.last?.first?.body,
+        optimisticComment.body
+      )
+    }
+  }
+
+  func testPostComment() {
```
msadoon, line 504: `testFailableComment_AfterCommentPostButtonTapped_CommentPostedSucessfully`

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,404 @@ internal final class CommentsViewModelTests: TestCase {
...
+    self.vm.inputs.configureWith(project: .template, update: nil)
+    self.vm.inputs.viewDidLoad()
+
+    withEnvironment(currentUser: .template) {
+      let bodyText = "Posting another comment."
+
+      self.vm.inputs.postCommentButtonTapped(bodyText)
+
+      self.postCommentSuccessful.assertValueCount(1, "Comment posted successfully")
     }
```
msadoon, line 514: Could you also test the `postCommentSubmitted` output?

KsApi/models/CommentTests.swift
```diff
@@ -0,0 +1,20 @@
+@testable import KsApi
+import XCTest
+
+final class CommentTests: XCTestCase {
```
msadoon, line 4: is this necessary, because we test it as part of the `postCommentSuccessful` test in the `CommentsViewModel`. It has to deserialize the optimistic comment and compare it to what is in the `.template` no?

KsApi/models/PostCommentEnvelope.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public struct PostCommentEnvelope: Decodable {
```
msadoon, line 3: I guess this doesn't really require a test because it's pretty straightforward wrapper.

msadoon (changes requested):
Looks good!
I haven't run the code yet on the simulator, my biggest concern was [this comment](https://github.com/kickstarter/ios-oss/pull/1491/files#diff-169d5a4c7742d75c54a5ad5cfb46aa489dfb1c829177e5df148af5d00e5aaa4eR165).

I will run test the code and post here shortly (today), if there's anything I find.

Library/ViewModels/CommentsViewModel.swift
tobitech: I guess I can leave this to the ticket for handling showing success or failure feedback to the user.

Library/ViewModels/CommentsViewModel.swift
tobitech: This approach transforms the variable to a function `(Array<Comment>, Array<Comment>.ArrayLiteralElement) -> [Comment]` instead of the return type specified

Library/ViewModels/CommentsViewModel.swift
tobitech: This line of code `let commentsAndProject = Signal.combineLatest(
      Signal.merge(comments, commentsWithOptimisticPostComment),
      initialProject
    )`
ensures that when `comments` is emitted it would replace the datasource comments

tobitech: Yeah I added a `// TODO:` on `ln 96` in `CommentsViewModel.swift`

KsApi/models/CommentTests.swift
tobitech: Because we're manually creating a new comment object, this is to ensure the comment is created with the right values based on what is supplied i.e. `(project, user, body)`

Library/ViewModels/CommentComposerViewModelTests.swift
```diff
@@ -6,12 +6,13 @@ import XCTest
...
   private let vm: CommentComposerViewModelType = CommentComposerViewModel()
 
-  private let bodyText = TestObserver<String, Never>()
+  private let bodyText = TestObserver<String?, Never>()
   private let avatarURL = TestObserver<URL?, Never>()
   private let inputAreaHidden = TestObserver<Bool, Never>()
   private let notifyDelegateDidSubmitText = TestObserver<String, Never>()
   private let postButtonHidden = TestObserver<Bool, Never>()
   private let placeholderHidden = TestObserver<Bool, Never>()
+  private let inputTextViewResignFirstResponder = TestObserver<(), Never>()
```
singhhari, line 15: alphabetize

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -11,8 +11,8 @@ public protocol CommentsViewModelInputs {
   /// Call with a `Comment` when it is selected.
   func didSelectComment(_ comment: Comment)
 
-  /// Call when the User is posting a comment or reply.
-  func postCommentButtonTapped()
+  /// Call when the User is posting a comment or reply
+  func commentComposerDidSubmitText(_ text: String)
```
singhhari, line 15: alphabetize

singhhari (approved):
a few minor formatting issues but otherwise lgtm

Library/ViewModels/CommentsViewModelTests.swift
tobitech: An assertion has been added for this in the `testDataSourceContainsFailableComment` function

msadoon (approved):
Just one missing test and maybe a renaming, otherwise it's good to go!

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -72,12 +72,10 @@ final class CommentComposerView: UIView {
...
-  /*
-   Call this to clear body text.
-   */
-  public func clearOnSuccess() {
-    // TODO: Replace this with a viewModel input for clearing body text.
-    self.viewModel.inputs.bodyTextDidChange("")
+  public func resetInput() {
+    self.viewModel.inputs.bodyTextDidChange(nil)
+    self.inputContainerView.inputTextView.text = nil
+    self.inputContainerView.inputTextView.resignFirstResponder()
```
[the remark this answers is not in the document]
tobitech: This has been implemented as well. Thanks for pointing that out.

Library/ViewModels/CommentsViewModel.swift
justinswart: Oh weird, yeah I see one needs to do this to make it work:

```swift
let commentsWithOptimisticPostComment = Signal
  .combineLatest(comments, optimisticPostComment)
  .map { loadedComments, optimisticComment in
    ([optimisticComment] + loadedComments)
  }
  .map { $0 as Array }
```

Anyway, up to you, either is fine.

Library/ViewModels/CommentsViewModel.swift
justinswart: Be sure to mark it as `TODO` in that case and comment on the relevant ticket what needs to be done, please. I think these banners are also only shown when retrying after an error?

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -28,11 +31,14 @@ public protocol CommentComposerViewModelOutputs {
   var avatarURL: Signal<URL?, Never> { get }
 
   /// Emits a string that should be put into the body text view.
-  var bodyText: Signal<String, Never> { get }
+  var bodyText: Signal<String?, Never> { get }
```
justinswart, line 34: Remind me why we emit the body text here? What is the scenario in which we will need to set the text in the textfield from the view model?

Library/ViewModels/CommentsViewModel.swift
tobitech: Based on the design, I think banners are only shown when flagging a comment. The failed, posting and posted state are part of the comment cell design.

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -54,21 +60,21 @@ public final class CommentComposerViewModel:
...
 
-    self.notifyDelegateDidSubmitText = self.bodyText
+    self.notifyDelegateDidSubmitText = self.bodyText.skipNil()
       .takeWhen(self.postButtonPressedProperty.signal)
       .map { $0.trimmed() }
 
     self.placeholderHidden = Signal.merge(
       self.configDataProperty.signal.mapConst(false).take(first: 1),
-      self.bodyText.map { !$0.isEmpty }
+      self.bodyText.map { !($0?.isEmpty ?? true) }
```
justinswart, line 72: I think this is easier to read here:
```swift
self.bodyText.map { $0?.isEmpty == false }
```

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -54,21 +60,21 @@ public final class CommentComposerViewModel:
...
     )
 
     self.postButtonHidden = Signal.merge(
       self.configDataProperty.signal.mapConst(true).take(first: 1),
-      self.bodyText.map { $0.trimmed().isEmpty }
+      self.bodyText.map { ($0?.trimmed().isEmpty ?? true) }
```
justinswart, line 77:
```swift
self.bodyText.map { $0?.trimmed().isEmpty == true }
```

Library/ViewModels/CommentComposerViewModel.swift
```diff
@@ -95,6 +103,12 @@ public final class CommentComposerViewModel:
     self.postButtonPressedProperty.value = ()
   }
 
+  fileprivate let resetInputProperty = MutableProperty(())
+  public func resetInput() {
+    self.bodyTextDidChange(nil)
```
justinswart, line 108: Let's not call this other input from this input, instead in our `init` we should wire them up like:

```swift
self.bodyText = Signal.merge(
  self.bodyTextDidChangeProperty.signal,
  self.resetInputProperty.signal.mapConst(nil)
)
```

justinswart (commented):
Couple more questions

Library/ViewModels/CommentComposerViewModel.swift
tobitech: So that we can clear the text view after posting cause at that point I'm also calling `resetInput` that resets the `bodyText` in the view model via `bodyTextDidChange(nil)`

justinswart: Oh right - cool, can we change it so that it's not updating the body text while we're typing? it seems unnecessary to type -> call input -> set text in a cycle. It would make more sense that when `resetInput` is called or the comment is posted, we emit a `nil` bodyText for the textview.

tobitech: Just seeing this. I'll look into that

justinswart (approved):
gtg

PR #1492, opened 2021-06-01
[NT-1981] Post PX Notification On Main Thread

justinswart:
# 📲 What

Ensures that the notification that triggers the PerimeterX dialogue is always posted on the main thread.

# 🤔 Why

In #1427 we made some changes which triggers the PerimeterX Captcha using a `Notification` rather than reaching out directly to the app's `Window` for display purposes. It's possible that this wasn't implemented quite correctly and may be causing some intermittent crashes due to the response object being passed across threads.

# 🛠 How

Ensure that the notification is always posted and handled on the main thread.

# ✅ Acceptance criteria

- [ ] Navigate to PerimeterX's developer console and ensure that the captcha flow still works as before.

### DIFF
--- a/KsApi/px/PerimeterXClient.swift
+++ b/KsApi/px/PerimeterXClient.swift
@@ -44,7 +44,13 @@ public class PerimeterXClient: NSObject, PerimeterXClientType {
       [.Block, .Captcha].contains(response.type)
     else { return false }
 
-    NotificationCenter.default.post(name: Notification.Name.ksr_perimeterXCaptcha, object: response)
+    DispatchQueue.main.async {
+      NotificationCenter.default.post(
+        name: Notification.Name.ksr_perimeterXCaptcha,
+        object: nil,
+        userInfo: ["response": response]
+      )
+    }
 
     return true
   }
--- a/KsApi/px/PerimeterXClientTests.swift
+++ b/KsApi/px/PerimeterXClientTests.swift
@@ -50,20 +50,13 @@ final class PerimeterXClientTests: XCTestCase {
       return
     }
 
-    var notification: Notification?
+    XCTAssertTrue(client.handleError(response: response, and: data!))
 
-    let token = NotificationCenter.default.addObserver(
-      forName: Notification.Name.ksr_perimeterXCaptcha,
-      object: nil,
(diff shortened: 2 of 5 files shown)


KsApi/px/PerimeterXClient.swift
```diff
@@ -44,7 +44,13 @@ public class PerimeterXClient: NSObject, PerimeterXClientType {
       [.Block, .Captcha].contains(response.type)
     else { return false }
 
-    NotificationCenter.default.post(name: Notification.Name.ksr_perimeterXCaptcha, object: response)
+    DispatchQueue.main.async {
+      NotificationCenter.default.post(
+        name: Notification.Name.ksr_perimeterXCaptcha,
+        object: nil,
+        userInfo: ["response": response]
```
justinswart, line 51: This wasn't correct before, according to the [docs](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1410608-postnotificationname) the `object` param is meant to represent the _sending_ object and the `userInfo` param is meant to contain additional data to carry along with the notification which is what we're really doing here.

KsApi/px/PerimeterXClientTests.swift
```diff
@@ -50,20 +50,13 @@ final class PerimeterXClientTests: XCTestCase {
...
-    var notification: Notification?
+    XCTAssertTrue(client.handleError(response: response, and: data!))
 
-    let token = NotificationCenter.default.addObserver(
-      forName: Notification.Name.ksr_perimeterXCaptcha,
-      object: nil,
-      queue: nil
-    ) { note in
-      notification = note
+    expectation(forNotification: Notification.Name.ksr_perimeterXCaptcha, object: nil) { note in
```
justinswart, line 55: This is a better way of testing for the notification as it allows us to also inspect it and check that it's posted on the correct thread.

KsApi/px/PerimeterXClient.swift
singhhari: ah wow, this one definitely slipped through the cracks while I reviewed as well. good catch.

singhhari (approved):
CAPTCHA works as expected. lgtm.

PR #1493, opened 2021-06-01
[NT-1983] Add feature flag and implementation for hiding flag comment button

singhhari:
# 📲 What

While planning the implementation and roll out of the native comment threading feature, we decided that it would be best to isolate the comment flagging behavior from the rest of the feature. This PR includes the addition of a new feature flag and enables the hiding of our flag comment button.

# 🤔 Why

Using a feature flag to handle the comment flagging enables us to release this component independently of the greater threading feature. While our Product and Design teams work out the details behind flagging, we will rely on the flag to obfuscate the flag from our UI/UX.

# 🛠 How

We're using Optimizely to handle the feature flag, updating the `CommentCell`, `CommentCellViewModel` and various test files.

# 👀 See
|  |  |  |  |

# ✅ Acceptance criteria

- [x] While the feature flag is toggled true and the user is logged out, verify the entire bottomRowStackView is hidden.
- [x] While the feature flag is toggled true and the user is a backer, verify only the replyButton is visible.
- [x] While the feature flag is toggled false and the user is a backer, verify both buttons are visible.
- [x] While the feature flag is toggled false and the user is not a backer, verify only the flagButton is visible.
- [x] Verify screenshots tests are rendering appropriately, given the inputs


Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
```diff
@@ -17,16 +17,12 @@ internal final class CommentsViewControllerTests: TestCase {
   }
 
   func testView_WithFailedRemovedAndSuccessfulComments_ShouldDisplayAll() {
-    AppEnvironment.pushEnvironment(
```
singhhari, line 20: Moved this into `withEnvironment` via the `mockService` argument

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
```diff
@@ -45,8 +41,11 @@ internal final class CommentsViewControllerTests: TestCase {
   }
 
   func testView_CurrentUser_LoggedOut() {
+    let mockService =
```
singhhari, line 44: A lot of these tests were rendering blank because this component of the method was missing.

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -23,10 +23,16 @@ public protocol CommentCellViewModelOutputs {
   /// Emits text containing comment body.
   var body: Signal<String, Never> { get }
 
+  /// Emits a Bool determining if the bottomRowStackView is hidden.
+  var bottomRowStackViewIsHidden: Signal<Bool, Never> { get }
```
singhhari, line 27: Assuming the replyButton and flagButton should both be hidden we can just hide the entire stack view. Otherwise we'll see a larger-than-needed (read: unpleasant) padding at the bottom of the cell.

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -83,6 +96,14 @@ public final class CommentCellViewModel:
     self.replyButtonIsHidden = Signal.combineLatest(isLoggedOut, isNotABacker)
       .map { isLoggedOut, isNotABacker in isLoggedOut || isNotABacker }
 
+    // If both the replyButton and flagButton should be hidden, the entire stackview will be hidden too
+    self.bottomRowStackViewIsHidden = Signal.combineLatest(
+      self.flagButtonIsHidden.signal,
+      self.replyButtonIsHidden.signal
+    ).map { flagButtonIsHidden, replyButtonIsHidden in
+      flagButtonIsHidden && replyButtonIsHidden
+    }
```
singhhari, line 105: Assuming the replyButton and flagButton should both be hidden we can just hide the entire stack view. Otherwise we'll see a larger-than-needed (read: unpleasant) padding at the bottom of the cell.

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -68,6 +74,13 @@ public final class CommentCellViewModel:
       badge.takeWhen(self.bindStylesProperty.signal)
     )
 
+    self.flagButtonIsHidden = self.commentAndProject.signal
+      .ignoreValues()
```
justinswart, line 78: `.ignoreValues()` is redundant here is we're doing `.map { _ ... }` on the next line.

justinswart (approved):
Small comment but looks good 👍

Library/ViewModels/CommentCellViewModel.swift
singhhari: Removed it and renamed the feature flag.

PR #1494, opened 2021-06-02
[NT-1983] Implementing feature flag for hiding replies on comment threads

singhhari:
# 📲 What

Adding another feature flag to support our comment threading feature. This one is specifically for toggling visibility on the button and stackview that support replying/viewing replies. 

# 🤔 Why

This raises our grand total to three feature flags for the release of native comment threading. By isolating each of these discrete pieces of functionality into separate flags we can be flexible in our release and mitigate any risk along the way.

# 🛠 How

- Added a new feature flag on Optimizely.
- Updated the existing outputs on `CommentCellViewModel` to read the flag.
- Updated unit and snapshot tests.

# 👀 See
| feature flag true | feature flag false, not backing | feature flag false, backing | feature flag false, not logged in |
| --- | --- | --- | --- |
|  |  |  |  |

# ✅ Acceptance criteria

- [x] Toggle the feature flag to true, log in, go to a project you have not backed and verify you do not see the reply button or view replies stackview.
- [x] Toggle the feature flag to true, log in, go to a project you have backed and verify you do not see the reply button or view replies stackview.
- [x] Toggle the feature flag to false, log in, to to a project you have not backed and verify you do not see the reply button but do see the stackview (assuming there are replies).
- [x] Toggle the feature flag to false, log in, to to a project you have backed and verify you see the reply button and the stackview (assuming there are replies).
- [ ] Look through screenshots and verify they render appropriately given the value of a specific feature flag.


Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -92,19 +92,29 @@ public final class CommentCellViewModel:
...
       self.flagButtonIsHidden.signal,
       self.replyButtonIsHidden.signal
     ).map { flagButtonIsHidden, replyButtonIsHidden in
       flagButtonIsHidden && replyButtonIsHidden
     }
 
+    // If there are no replies or if the feature flag returns true, hide the stack view.
     self.viewRepliesStackViewIsHidden = comment.map(\.replyCount)
-      .map { $0 == 0 }
+      .map { replyCount in
```
justinswart, line 113: How about to improve readability here we break these out into private functions in this view model:

```swift
private func viewRepliesStackViewHidden(_ replyCount: Int) -> Bool {
  guard commentThreadingRepliesDisabled( == false else { return true }
  return replyCount == 0
}

private func commentThreadingRepliesDisabled() -> Bool {
  return AppEnvironment.current.optimizelyClient?.isFeatureEnabled(featureKey: OptimizelyFeature.Key.commentThreadingRepliesDisabled.rawValue) ?? false
}
```
And map over them like:

```swift
// If there are no replies or if the feature flag returns true, hide the stack view.
self.viewRepliesStackViewIsHidden = comment.map(\.replyCount)
  .map(viewRepliesStackViewHidden)
```

jgsamudio (approved):

Library/ViewModels/CommentCellViewModel.swift
singhhari: Updated this and the replyButtonIsHidden logic as well. Both are quite similar.

justinswart (approved):

PR #1495, opened 2021-06-02
[NT 1991] UX Spinner while loading comments

moderateepheezy:
# 📲 What

Add a Loading indicator when loading more comments. 

# 🤔 Why

When user scrolls to the bottom we automatically load more comments, at the point of fetching more comments, we don't want users with slow network to think there are no more comments when we are just loading more of them.

# 🛠 How

1. Add a Custom TableView Footer with an activity indicator
2. Add the Custom TableView Footer when we subscribe to an event that triggers user is 3 comments aways to the end of the list.

# 👀 See

Loadmore

# ✅ Acceptance criteria

- [ ] An activity indicator shows when loading more comments and stops when new comments are fetched.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -20,6 +24,12 @@ internal final class CommentsViewController: UITableViewController {
     return view
   }()
 
+  private lazy var footerView: CommentTableViewFooter = {
+    let frame = CGRect(x: 0, y: 1, width: UIScreen.main.bounds.width, height: Layout.Footer.height)
```
singhhari, line 28: Why is y 1?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -39,6 +39,9 @@ public protocol CommentsViewModelOutputs {
 
   /// Emits a list of `Comment`s and the `Project` to load into the data source.
   var loadCommentsAndProjectIntoDataSource: Signal<([Comment], Project), Never> { get }
+
+  /// Emits a boolean that determines if the loader more activity indicator should show
+  var shouldShowLoadMoreIndicator: Signal<Bool, Never> { get }
```
singhhari, line 44: I think `shouldShowIndicator` is sufficient.

singhhari: Also update the documentation to:
"Emits a Bool that determines if the activity indicator should render."

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -148,6 +151,12 @@ public final class CommentsViewModel: CommentsViewModelType,
         [comment.replyCount > 0, comment.status == .success].allSatisfy(isTrue)
       }
       .withLatestFrom(initialProject)
+
+    self.shouldShowLoadMoreIndicator = Signal
+      .combineLatest(isCloseToBottom.mapConst(true), self.isCommentsLoading)
```
singhhari, line 156: if you're mapping `isCloseToBottom` to true every time, is it needed to combine here?

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -17,6 +17,7 @@ internal final class CommentsViewModelTests: TestCase {
   private let isCommentsLoading = TestObserver<Bool, Never>()
   private let loadCommentsAndProjectIntoDataSourceComments = TestObserver<[Comment], Never>()
   private let loadCommentsAndProjectIntoDataSourceProject = TestObserver<Project, Never>()
+  private let shouldShowLoadMoreIndicator = TestObserver<Bool, Never>()
```
singhhari, line 20: Same renaming as above.

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,371 @@ internal final class CommentsViewModelTests: TestCase {
...
-      self.goToCommentRepliesComment.assertDidNotEmitValue()
-      self.goToCommentRepliesProject.assertDidNotEmitValue()
+  func testGoToCommentReplies_CommentHasReplies_GoToEmits() {
+    self.goToCommentRepliesComment.assertDidNotEmitValue()
+    self.goToCommentRepliesProject.assertDidNotEmitValue()
 
-      let project = Project.template
-      let comment = Comment.template
-        |> \.replyCount .~ 1
+    let project = Project.template
```
singhhari, line 122: If you are not changing the template, I would just delete the constant and supply an argument of `.template` wherever it's needed.

singhhari: Check all instances of this below as well.

singhhari (changes requested):

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: Yes, it's needed. `isCloseToBottom` ordinary returns void, so if it emits a value then I can map it as true. The condition is only satisfy if `isCloseToBottom` emits and `isCommentsLoading` emits value.

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: Isn't `shouldShowIndicator` a bit abstract since it doesn't say exactly what the indicator is about?

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
moderateepheezy: Haa, thanks I missed that.

Library/ViewModels/CommentsViewModelTests.swift
moderateepheezy: Oh I didn't noticed it, I only copied out all the nested test. I'll remove the instances.

Kickstarter-iOS/Views/CommentTableViewFooter.swift
```diff
@@ -0,0 +1,43 @@
...
+import Prelude
+import UIKit
+
+final class CommentTableViewFooter: UIView {
+  // MARK: - Properties
+
+  private lazy var activityIndicator: UIActivityIndicatorView = {
+    let indicator = UIActivityIndicatorView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+    indicator.hidesWhenStopped = true
```
msadoon, line 11: Can just do `|> \.hidesWhenStopped .~ true`

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -20,6 +24,12 @@ internal final class CommentsViewController: UITableViewController {
     return view
   }()
 
+  private lazy var footerView: CommentTableViewFooter = {
+    let frame = CGRect(x: 0, y: 0, width: UIScreen.main.bounds.width, height: Layout.Footer.height)
```
msadoon, line 28: Can you use the `tableView` width instead?

Kickstarter-iOS/Views/CommentTableViewFooter.swift
```diff
@@ -0,0 +1,43 @@
...
+    return indicator
+  }()
+
+  public var shouldShowActivityIndicator: Bool = false {
```
msadoon, line 15: Not sure if this should be updated via a view model? cc @justinswart

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -632,6 +632,7 @@
 		94114D8826547F620063E8F6 /* CommentCellViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 94114D8726547F620063E8F6 /* CommentCellViewModelTests.swift */; };
 		94114D90265488720063E8F6 /* CommentCellHeaderStackViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 94114D8F265488710063E8F6 /* CommentCellHeaderStackViewTests.swift */; };
 		941710B526557A6A00F5954E /* CommentCellStyles.swift in Sources */ = {isa = PBXBuildFile; fileRef = 941710B426557A6A00F5954E /* CommentCellStyles.swift */; };
+		94BA168E2667C37F0034CC3F /* CommentTableViewFooter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 94BA168D2667C37F0034CC3F /* CommentTableViewFooter.swift */; };
```
msadoon, line 635: Might want to rename `CommentsTableViewFooter`

Library/ViewModels/CommentsViewModel.swift
singhhari: The way I was looking at it was that there is no other "indicator" in the file but if you want to be slightly more descriptive you can do: `shouldShowActivityIndicator`

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,365 @@ internal final class CommentsViewModelTests: TestCase {
...
 
-      withEnvironment(apiService: MockService(fetchCommentsEnvelopeResult: .success(envelope))) {
-        self.vm.inputs.configureWith(project: project, update: nil)
-        self.vm.inputs.viewDidLoad()
+      self.loadCommentsAndProjectIntoDataSourceComments.assertValues(
+        [envelope.comments],
+        "A set of comments is emitted."
+      )
+      self.loadCommentsAndProjectIntoDataSourceProject.assertValues([.template])
+      self.isCommentsLoading.assertValues([true, false])
```
msadoon, line 342: May want to test `shouldShowIndicator` state here as well.

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,365 @@ internal final class CommentsViewModelTests: TestCase {
...
-            "No new comments are emitted."
-          )
-          self.loadCommentsAndProjectIntoDataSourceProject.assertValues([.template])
-          self.isCommentsLoading.assertValues([true, false, true])
+        self.loadCommentsAndProjectIntoDataSourceComments.assertValues(
+          [envelope.comments, envelope.comments + updatedEnvelope.comments],
+          "New comments are emitted."
+        )
+        self.loadCommentsAndProjectIntoDataSourceProject.assertValues([.template, .template])
+        self.isCommentsLoading.assertValues([true, false, true, false])
```
msadoon, line 366: test `shouldShowIndicator` here as well.

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -111,365 +113,365 @@ internal final class CommentsViewModelTests: TestCase {
...
+          .assertValues([.template, .template], "Same project is emitted again")
+      }
+    }
+  }
 
-          self.loadCommentsAndProjectIntoDataSourceComments
-            .assertValues([envelope.comments], "No new comments are emitted")
-          self.loadCommentsAndProjectIntoDataSourceProject
-            .assertValues([project], "No new projects are emitted")
+  func testProjectPagination_WhenLimitReached_CommentsAreUpdatedInDataSource() {
```
msadoon, line 320: If adding a test condition, would be good to update the name of the test.

msadoon (changes requested):
Few changes, mostly around testing, otherwise looks great!
There was an issue with the scrolling indicator looking like it didn't hit the bottom of the scrollview when the loading indicator is shown:

Maybe also make the indicator larger and more centered?

Library/ViewModels/CommentsViewModel.swift
singhhari: I guess what I'm suggesting is that you probably don't need to be calling mapConst(true) here and checking for it's value because it will always be true. Something like this will probably give you the same result:

    self.shouldShowIndicator = Signal
      .combineLatest(isCloseToBottom, self.isCommentsLoading)
      .map { _, isCommentsLoading in isCommentsLoading }
      
      maybe using `takeWhen(isCloseToBottom)` would give the same desired effect too but I'm not 100 percent certain.

Library/ViewModels/CommentsViewModelTests.swift
moderateepheezy: There's a test for it, some few lines below.

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: I think we can leave it at `shouldShowIndicator `, we might not be able to use `shouldShowRefreshControl ` because it's not for refresh, but a load more.

Library/ViewModels/CommentsViewModelTests.swift
moderateepheezy: As you've mentioned in one of the comments, rename would do well, testing `shouldShowIndicator` would return empty at those stages you mentioned.

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: @singhhari so I updated to this

```
self.shouldShowIndicator = Signal
  .combineLatest(isCloseToBottom, self.isCommentsLoading)
  .map { _, isCommentsLoading in isCommentsLoading }
```

But note that `isCloseToBottom.mapConst(true)` will only emit only when user scroll close to the bottom of the list. I have tested this with small and large comments.

Also `takeWhen(isCloseToBottom)` won't work in this case because you always need a `true` or `false` value.

Library/ViewModels/CommentsViewModelTests.swift
msadoon: I guess the idea is to test the "before" and "after" states of this output, just like the other ones.

msadoon (approved):
looks good to me!

Library/ViewModels/CommentsViewModel.swift
singhhari: Fair enough although you make a good point that it does not say what exactly the indicator is doing so something like `shouldShowLoadingIndicator` is half way between what you and I are suggesting.

Library/ViewModels/CommentsViewModel.swift
singhhari: Yup sounds good. I just wanted to simplify it without compromising functionality.

Kickstarter-iOS/Views/CommentTableViewFooter.swift
singhhari: We almost always use a view model output to update a component of the UI. e.g `.hidden`

singhhari (changes requested):
Coverage shouldn't be going down given the addition of new code. Should be good to go after code coverage.

moderateepheezy:
> Coverage shouldn't be going down given the addition of new code. Should be good to go after code coverage.

I think this was related to the merge from @msadoon, the codecov was up by about 0.32%, I think it will go backup after I fix the merge conflicts and push again.

msadoon:
> > Coverage shouldn't be going down given the addition of new code. Should be good to go after code coverage.
> 
> I think this was related to the merge from @msadoon last PR, the codecov was up by about 0.32%, I think it will go backup after I fix the merge conflicts and push again.

I think the last PR I merged had +0.01% code cov.

Kickstarter-iOS/Views/CommentTableViewFooterView.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    _ = (self.activityIndicator, self)
+      |> ksr_addSubviewToParent()
+      |> ksr_constrainViewToCenterInParent()
+  }
+
+  // MARK: - View model
+
+  override func bindViewModel() {
+    self.viewModel.outputs.shouldStartLoaderIndicator
```
singhhari, line 56: There is a disconnect with the name of this output and the name of `shouldShowActivityIndicator`. We should make them more similar.

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -140,6 +152,10 @@ internal final class CommentsViewController: UITableViewController {
         $0 ? self?.refreshControl?.beginRefreshing() : self?.refreshControl?.endRefreshing()
       }
 
+    self.viewModel.outputs.shouldShowLoadingIndicator
+      .observeForUI()
```
singhhari, line 156: there should be a capture list here.

Library/ViewModels/CommentTableViewFooterViewModelTests.swift
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  func testOutputs() {
+    self.shouldStartLoaderIndicator.assertDidNotEmitValue()
+
+    self.vm.inputs.configure(with: true)
+
+    self.shouldStartLoaderIndicator.assertValues([true], "loading indicator value is emitted")
+
+    self.vm.inputs.configure(with: false)
```
singhhari, line 24: I would separate the testing of true and false into two separate test cases for readability.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -46,6 +46,9 @@ public protocol CommentsViewModelOutputs {
   /// Emits a list of `Comment`s and the `Project` to load into the data source.
   var loadCommentsAndProjectIntoDataSource: Signal<([Comment], Project), Never> { get }
 
+  /// Emits a Bool that determines if the activity indicator should render.
+  var shouldShowLoadingIndicator: Signal<Bool, Never> { get }
```
singhhari, line 50: alphabetize

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -218,6 +227,7 @@ public final class CommentsViewModel: CommentsViewModelType,
   public let isCommentsLoading: Signal<Bool, Never>
   public let isCellSeparatorHidden: Signal<Bool, Never>
   public let loadCommentsAndProjectIntoDataSource: Signal<([Comment], Project), Never>
+  public let shouldShowLoadingIndicator: Signal<Bool, Never>
```
singhhari, line 230: same note as above.

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -18,6 +18,7 @@ internal final class CommentsViewModelTests: TestCase {
   private let isCommentsLoading = TestObserver<Bool, Never>()
   private let loadCommentsAndProjectIntoDataSourceComments = TestObserver<[Comment], Never>()
   private let loadCommentsAndProjectIntoDataSourceProject = TestObserver<Project, Never>()
+  private let shouldShowLoadingIndicator = TestObserver<Bool, Never>()
```
singhhari, line 21: alphabetize

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -314,27 +312,27 @@ internal final class CommentsViewModelTests: TestCase {
...
           .assertValues([envelope.comments, updatedEnvelope.comments], "New comments are emitted")
         self.loadCommentsAndProjectIntoDataSourceProject
-          .assertValues([project, project], "Same project is emitted again")
+          .assertValues([.template, .template], "Same project is emitted again")
       }
     }
   }
 
-  func testProjectPagination_WhenLimitReached_CommentsAreUpdatedInDataSource() {
+  func testProjectPagination_WhenLimitReached_CommentsAreUpdatedInDataSource_AndEmitsLoadMoreIndicatorValue() {
```
singhhari, line 327: the names of these functions are a bit much imo. We should simplify it something like:

`testLoadCommentsAndLoadingIndicator_WillDisplayRowAtEnd`

The exact name is open for discussion but we should either simplify the name or move some of the tests out into a separate method.

singhhari (changes requested):
Left some comments but also want to verify the following:

The screenshot for `Screenshots/_64/Kickstarter_Framework_iOSTests.CommentsViewControllerTests/testView_WithFailedRemovedAndSuccessfulComments_ShouldDisplayAll_Comments___lang_de@2x.png` changed. Is that expected?

moderateepheezy:
> WithFailedRemovedAndSuccess

This is probably due to a merge conflict I resolved, I think one of the screenshots was listed in the conflict, but can't remember which, but I recorded the tests again.

Library/ViewModels/CommentTableViewFooterViewModelTests.swift
```diff
@@ -6,23 +6,27 @@ import XCTest
...
   override func setUp() {
     super.setUp()
 
-    self.vm.outputs.shouldStartLoaderIndicator.observe(self.shouldStartLoaderIndicator.observer)
+    self.vm.outputs.shouldShowActivityIndicator.observe(self.shouldShowActivityIndicator.observer)
   }
 
-  func testOutputs() {
-    self.shouldStartLoaderIndicator.assertDidNotEmitValue()
+  func testShowActivityIndicator() {
```
singhhari, line 17: I know this is annoyingly specific but you should be following the conventions we already use in the project:

`test{OUTPUTNAME}_{VariableX}_{true/false}`

so in the case of these two tests it will be something like `testShouldShowActivityIndicator_Configure_True`

there's no set way to do it but we do follow this throughout the project

singhhari (approved):

Library/ViewModels/CommentTableViewFooterViewModelTests.swift
moderateepheezy: Oh, Alright then.

PR #1496, opened 2021-06-02
[NT-1942] Empty State Styling and Datasource Update

msadoon:
# 📲 What

This PR wraps up the remaining work on the empty state view if there are no comments. [ticket](https://kickstarter.atlassian.net/browse/NT-1942).

# 🤔 Why

Whenever a project has no comments its good to inform the user of this state instead of leaving them with a blank page.

# 🛠 How

The styling was using the existing `PaddingLabel` and the section update required clearing the `CommentsDataSource.Section.empty` before `appending` to the `CommentsDataSource.Section.comments`
, otherwise both the empty cell and comment cell would show when refreshing with a new comment.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 
# 🏎 Performance
# ✅ Acceptance criteria

- [x] Ensure [designs](https://www.figma.com/file/zyeDBH6vk2DMmt39S1yBEx/Native-Comment-Threading?node-id=491%3A8738) are as expected
- [x] Ensure composer view only shows for logged in backers
- [x] Ensure blocked composer view shows for logged in nonbackers
- [x] Ensure no composer view shows for not logged in users
- [x] Ensure posting a comment updates empty state optimistically ([dependency](https://github.com/kickstarter/ios-oss/pull/1491))

# ⏰ TODO

- [x] Check last requirement above once [Post Comment PR](https://github.com/kickstarter/ios-oss/pull/1491) is merged


Kickstarter-iOS/DataSources/CommentsDataSource.swift
```diff
@@ -23,6 +23,8 @@ internal final class CommentsDataSource: ValueCellDataSource {
       return
     }
 
+    self.clearValues(section: Section.empty.rawValue)
```
justinswart, line 26: I think if we just update the first `clearValues` to `clearValues()` without passing a section in then it'll clear all of the values and we won't need two calls.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -40,6 +40,9 @@ public protocol CommentsViewModelOutputs {
   /// Emits a boolean that determines if comments are currently loading.
   var isCommentsLoading: Signal<Bool, Never> { get }
 
+  /// Emits a boolean that determines if cell separator is to be hidden.
+  var isCellSeparatorHidden: Signal<Bool, Never> { get }
```
justinswart, line 44: `cellSeparatorHidden` is typically our convention for outputs related to a view's `hidden` property.

Although - in our case we're actually not hiding it but just setting its color, so here I would in fact emit the color and test that value.

Kickstarter-iOS/Views/Cells/EmptyCommentsCell.swift
```diff
@@ -34,6 +37,8 @@ final class EmptyCommentsCell: UITableViewCell, ValueCell {
       |> \.numberOfLines .~ 0
       |> \.adjustsFontForContentSizeCategory .~ true
       |> \.font .~ UIFont.ksr_callout()
+      |> \.textColor .~ UIColor.ksr_support_400
+      |> \.textAlignment .~ .center
```
singhhari, line 41: alphabetize

singhhari (approved):
Minor formatting stuff but otherwise lgtm

Library/ViewModels/CommentsViewModel.swift
justinswart: Hmm although, does `self.tableView.separatorStyle = .none` not work here?

msadoon: Actually yes, that is a better output. I'll change it to that.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift:10:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

PR #1497, opened 2021-06-03
[NT-1859] Xcode 12.5 And iOS 14.5 Compatibility

justinswart:
# 📲 What

Updates our codebase to build on Xcode 12.5 and iOS 14.5.

# 🤔 Why

This allows us to keep working on the latest version of Xcode and not fall behind in our iOS version support.

# 🛠 How

We encountered some peculiar runtime crashes when first running the app on iOS 14.5. These were each solved by changing the offending styles function from a regular composed function to a closure capturing the function and then composing it. It's still not entirely clear to me what might be causing this crash (some Swift internals likely changed) but making these changes does reliably fix them wherever we've encountered them.

# 👀 See

[Xcode 12.5 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-release-notes)

# ✅ Acceptance criteria

- [ ] Due to the broad change here and runtime crashes the app will require a full regression test when this merges.
- [x] Snapshots should not have changed in any noticeable way.


milreyes:
Cannot wait for threaded comments to be released. Thanks for your work on this.

singhhari (approved):
all tests passing locally.
ran the app on 12.5 + 14.5 simulator with no issues

lgtm!

msadoon (approved):
Codewise looks good to me. 
I'm just going to try running it on device and play around with it, but no need to wait on that because I have to set up provisioning profiles etc. If any issues are found I'll post on Slack.

PR #1499, opened 2021-06-04
[NT-1896] UX - Post New Comment

tobitech:
# 📲 What

A bug was noticed that after posting a comment the input area's height stays expanded. Realised that this happens only when the input area expands beyond the maximum height where scrolling becomes enabled. The height needed to be updated immediately a comment is posted as part of resetting of the composer.

# 🤔 Why

This is part of the new experience we're building when posting comment so that the input area resizes based on the text the user enters into it.

# 👀 See

https://user-images.githubusercontent.com/8595853/120828428-6c4f6780-c554-11eb-8e00-6b2ef883216f.mov

# ✅ Acceptance criteria

1. Launch the app
2. Select a project you're already backing
3. Enter some comments. The input area height should expand based on the text in it.
4. Keep entering texts until the input area scrolling is enabled
5. Tap the post button. The input area should be cleared and the height be reset.


msadoon (approved):
good to go (after fixed merge conflicts)

Kickstarter-iOS/Views/CommentComposerView.swift
```diff
@@ -144,6 +144,12 @@ final class CommentComposerView: UIView {
         self?.inputContainerView.inputTextView.resignFirstResponder()
       }
 
+    self.viewModel.outputs.updateTextViewHeight
+      .observeForUI()
+      .observeValues { [weak self] _ in
+        self?.inputContainerView.inputTextView.invalidateIntrinsicContentSize()
```
singhhari, line 150: Solution seems fine but I'm curious about one thing. Does resetInput() not achieve this for us?

tobitech: This line of code `self?.inputContainerView.inputTextView.invalidateIntrinsicContentSize()` was initially called in `textViewDidChange(_:)` delegate method but I wanted that to be controlled by a VM output and YES `self.viewModel.inputs.resetInput()` is one of the actions that triggers `self.viewModel.outputs.updateTextViewHeight`.

justinswart (approved):

PR #1502, opened 2021-06-04
[Unticketed] Update Readme With Testing Steps

msadoon:
# 📲 What

Small PR to update `README.md` on front page of project with testing steps that took me a few tries to figure out.

# 🤔 Why

Helps anyone that is starting work on the project to be able to test the project with less confusion.

# 🛠 How

# 👀 See
# ♿️ Accessibility 
# 🏎 Performance
# ✅ Acceptance criteria
# ⏰ TODO
n/a


justinswart:
Hey @msadoon thanks for adding this, I think it's useful info but it's really more for our internal purposes so may be better to put into a guru card? thoughts @singhhari @jgsamudio ?

msadoon:
> Hey @msadoon thanks for adding this, I think it's useful info but it's really more for our internal purposes so may be better to put into a guru card? thoughts @singhhari @jgsamudio ?

Ok, no problem. I just want to confirm @jgsamudio is okay with it? It was a goal we set during our 1:1.

jgsamudio:
@msadoon @justinswart The changes could also benefit anyone looking the project as an open source repo. Documentation like this should be in the readme, while more "how to" related to our process make more sense in a guru card. I'm open to either approach since I know we typically don't update the readme very often

justinswart:
> @msadoon @justinswart The changes could also benefit anyone looking the project as an open source repo. Documentation like this should be in the readme, while more "how to" related to our process make more sense in a guru card. I'm open to either approach since I know we typically don't update the readme very often

Sure, just wanting to avoid bloating the readme. Currently the readme mentions how to run tests using  `make test-all` command. That _should_ be sufficient for OSS purposes. Knowing that one should switch targets to only run that target's tests is more about how Xcode works than how our project works. Our readme also makes reference to outdated documentation there which we should also review and decide whether or not is still relevant.

Secondly, we don't really accept or, at least encourage, OSS contributions, so speaking of running each scheme before pushing a commit is more internal documentation.

My suggestion would be to just have something like:

```markdown
## Testing the project

- Run all tests from the command line by running `make test-all`. // move the existing text.
- Run an individual scheme's tests by selecting that scheme in Xcode and hitting CMD+U.
```

That's about all that I think is relevant for OSS. The more detailed stuff is useful but I would put it in our upcoming Guru card.

In terms of this one:

```markdown
Running `CMD+U` on the `Kickstarter-iOS` schema may not always correctly run all test schemas above even if it is meant to.
```

We should figure out why that is 😄

singhhari:
Yep I think putting the more detail stuff into a guru card would be useful for any future engineers on the team but a small update here is fine.

msadoon:
Thanks all for suggestions, updated the `README.md` and added a [Guru card ](https://app.getguru.com/card/cL749Gji/How-iOS-Test-Schemas-Work)with more internal focused instructions.

jgsamudio:
Thanks @msadoon! Verified your Guru card!

jgsamudio (approved):

justinswart (approved):

PR #1503, opened 2021-06-07
[NT-2010] Invert comment feature flag logic

singhhari:
# 📲 What

After corresponding with Android, we realized we had the logic for the feature flags inverted. In order to achieve functional parity within the team, we're going to invert the existing logic for the replies and flagging feature flags.

# 🤔 Why

It's easier for the rest of the team to test these features when the flags on Android and iOS are doing the same thing. The desired behavior is for a `true` feature flag on Optimizely resulting in the feature being enabled on both platforms.

# ✅ Acceptance criteria

- [x] While the reply flag is true on Optimizely, test the reply buttons are visible
- [x] While the reply flag is false on Optimizely, test the reply buttons are NOT visible
- [x] While the flagging flag is true on Optimizely, test the flag button is visible
- [x] While the flagging flag is false on Optimizely, test the flag button is NOT visible


Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -76,8 +76,12 @@ public final class CommentCellViewModel:
 
     self.flagButtonIsHidden = self.commentAndProject.signal
       .map { _ in
-        AppEnvironment.current.optimizelyClient?
-          .isFeatureEnabled(featureKey: OptimizelyFeature.Key.commentFlaggingDisabled.rawValue) ?? false
+        guard let isFeatureEnabled = AppEnvironment.current.optimizelyClient?
```
justinswart, line 79: I think having these split out into functions makes it easier on the eye:

```swift
self.flagButtonIsHidden = self.commentAndProject.signal
  .ignoreValues()
  .map(commentFlaggingEnabled)
  .map(isFalse) // If the feature is enabled, the flag button should be visible.
```

```swift
private func commentFlaggingEnabled() -> Bool {
  AppEnvironment.current.optimizelyClient?
    .isFeatureEnabled(featureKey: OptimizelyFeature.Key.commentFlaggingEnabled.rawValue) ?? true
}
```

justinswart (approved):

Library/ViewModels/CommentCellViewModel.swift
msadoon: Just a minor thing, should the nil-coelescing operator be default `false`, because that means the feature is enabled?

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -79,75 +85,63 @@ internal final class CommentCellViewModelTests: TestCase {
...
         .assertValue(
-          false,
-          "The feature flag is false and the user is a backer, therefore the stack view is not hidden."
+          true,
+          "The feature flags are false, therefore the stack view is hidden."
         )
     }
   }
 
   func testOutputs_bottomRowStackViewIsHidden_LoggedOut_FeatureFlag_True() {
```
msadoon, line 94: `isNotHidden`

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -79,75 +85,63 @@ internal final class CommentCellViewModelTests: TestCase {
...
         .assertValue(
-          true,
-          "The feature flag is true and the user is not logged in, therefore the stack view is hidden."
+          false,
+          "The comment flagging feature is enabled, therefore the stack view is not hidden."
         )
     }
   }
 
   func testOutputs_bottomRowStackViewIsHidden_IsNotBacking_FeatureFlag_True() {
```
msadoon, line 109: `isNotHidden`

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -173,7 +167,7 @@ internal final class CommentCellViewModelTests: TestCase {
 
   func testOutputs_replyButtonIsHidden_viewRepliesStackViewIsHidden_IsBacker_True_IsLoggedIn_FeatureFlag_True() {
```
msadoon, line 168: `isNotHidden`

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -184,47 +178,50 @@ internal final class CommentCellViewModelTests: TestCase {
...
       self.replyButtonIsHidden
-        .assertValue(true, "The replyButton is hidden because the feature flag is true.")
+        .assertValue(false, "The replyButton is not hidden because the feature flag is true.")
     }
     self.viewRepliesStackViewIsHidden
-      .assertValue(true, "The stack view is hidden because the feature flag is true.")
+      .assertValue(false, "The stack view is not hidden because the feature flag is true.")
   }
 
   func testOutputs_replyButtonIsHidden_viewRepliesStackViewIsHidden_IsBacker_False_IsLoggedIn_FeatureFlag_True() {
```
msadoon, line 187: `isNotHidden` for stack view

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -314,13 +311,19 @@ internal final class CommentCellViewModelTests: TestCase {
   }
 
   func testViewRepliesContainerHidden_IsNotHiddenWhenCommentHasReplies() {
```
msadoon, line 313: change name to reflect feature flag.

msadoon (approved):
Added some minor test name changes, otherwise looks good!

Library/ViewModels/CommentCellViewModelTests.swift
singhhari: `bottomRowStackViewIsHidden` is the name of the output I'm testing so I'm leaving the name as is so we can be aware of which output the test is focusing on.

singhhari: Resolving the ones below.

Library/ViewModels/CommentCellViewModel.swift
singhhari: Updated this although the nil-coalescing is going to default to false because `true` means the feature is enabled now.

PR #1504, opened 2021-06-08
[NT-2020] Update Paginate Function To Emit Errors

justinswart:
# 📲 What

Updates our pagination function to emit errors in anticipation of needing this for handling paging errors for comments.

**Note:** This just modifies the `paginate` function, we will make use of this in an upcoming PR for comments paging.

# 🤔 Why

We have historically not done any error handling for pagination and we've decided to bring this improvement to threaded comments.

# 🛠 How

Modified `paginate` to return an additional `Signal` that can emit an `Error` without causing the existing value signal to error.

# ✅ Acceptance criteria

- [ ] No regressions on existing behaviour (tests pass).
- [x] New functionality and tests that were added make sense.

### DIFF
--- a/Library/Paginate.swift
+++ b/Library/Paginate.swift
@@ -25,17 +25,17 @@ import ReactiveSwift
  - parameter skipRepeats:          A boolean that determines if results can be repeated.
  - parameter valuesFromEnvelope:   A function to get an array of values from the results envelope.
  - parameter cursorFromEnvelope:   A function to get the cursor for the next page from a results envelope.
- - parameter statsFromEnvelope:    A function to get the stats from the results envelope.
  - parameter requestFromParams:    A function to get a request for values from a params value.
  - parameter requestFromCursor:    A function to get a request for values from a cursor value.
  - parameter concater:             An optional function that concats a page of values to the current array of
                                    values. By default this simply concatenates the arrays, but you might want
                                    to do something more specific, such as concatenating only distinct values.
 
- - returns: A tuple of signals, (paginatedValues, isLoading, pageCount). The `paginatedValues` signal will
+ - returns: A tuple of signals, (paginatedValues, isLoading, pageCount, errors). The `paginatedValues` signal will
             emit a full set of values when a new page has loaded. The `isLoading` signal will emit `true`
             while a page of values is loading, and then `false` when it has terminated (either by completion
-            or error). Finally, `pageCount` emits the number of the page that loaded, starting at 1.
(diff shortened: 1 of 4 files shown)


Library/Paginate.swift
```diff
@@ -44,7 +44,6 @@ public func paginate<Cursor, Value: Equatable, Envelope, ErrorEnvelope, RequestP
   skipRepeats: Bool = true,
   valuesFromEnvelope: @escaping ((Envelope) -> [Value]),
   cursorFromEnvelope: @escaping ((Envelope) -> Cursor),
-  statsFromEnvelope: ((Envelope) -> Int)? = nil,
```
justinswart, line 47: Due to the very specific nature of needing stats here I've removed this and I am instead picking out this value where we need it.

Library/ViewModels/SearchViewModelTests.swift
```diff
@@ -684,7 +684,7 @@ internal final class SearchViewModelTests: TestCase {
 
         self.scrollToProjectRow.assertValues([5, 6, 7, 8])
 
-        self.hasAddedProjects.assertValues([true, false, true, true], "Paginated projects are loaded.")
```
justinswart, line 687: I know I said no regressions but I'm open to guesses as to why this test needed to change 😅. I can't see any reason that the change I made would cause this 🤔

Library/ViewModels/SearchViewModel.swift
```diff
@@ -3,6 +3,11 @@ import KsApi
 import Prelude
 import ReactiveSwift
 
+private struct SearchPageData: Equatable {
```
justinswart, line 6: Just a simple type to allow us to carry the projects and stats together.

Library/ViewModels/SearchViewModelTests.swift
justinswart: Appears there is another failing snapshot test so will investigate.

Library/Paginate.swift
singhhari: I think we added a ticket into the backlog to do this piece so you may have killed two birds with one stone on this PR.

Library/ViewModels/SearchViewModelTests.swift
singhhari: Not sure if @tobitech may know what's going on here? I remember him working on the stats piece. Feel free to disregard if you don't know.

singhhari (approved):
Other than the snapshot failure and change in the SearchViewModelTests, lgtm!

PR #1506, opened 2021-06-08
[NT 2007] UI Show error posting comment

moderateepheezy:
# 📲 What

Added a retrying failed comment post UX to Comment Threading.

# 🤔 Why

When a user tries to post a new comment and it fails, we want to add a UX where the user can see a visual state of retrying a failed comment. The visual states goes from `Error` -> `Posting` -> `Posted`.

# 🛠 How

# 👀 See

RetryingComment

# ✅ Acceptance criteria

- [x] Make a post comment fail either using Charles Proxy or anything comfortable with you. Confirm that the comment cell state changes from success to a failed state.
- [x] When you tap to retry while still disabling your network, confirm that the state goes from `Failed` -> `Posting` -> `Failed`.
- [x] Test the happy path and confirm the state goes from `Failed` -> `Posting` -> `Posted` -> `Success`.
- [x] Confirm that there's a bit of delay in transitioning from `Failed` -> `Posting` -> `Posted` and also from `Failed` -> `Posting` -> `Failed`.
- [x] Confirm that the `You` badge doesn't shrink after retrying a failed comment, and also all badges are correctly styled when moving from theses error states.
- [ ] Also test with Pull to refresh.

# ⏰ TODO

- [ ] UI Tests for the Error states


moderateepheezy (commented):
Will need to merge https://github.com/kickstarter/ios-oss/pull/1505#pullrequestreview-678639877 before we can test the delay for the retrying error state.

KsApi/models/templates/CommentsEnvelopeTemplates.swift
```diff
@@ -37,4 +37,17 @@ extension CommentsEnvelope {
     slug: "slug",
     totalCount: 0
   )
+
+  internal static let successFailedRetryingRetrySuccessCommentsTemplate = CommentsEnvelope(
```
moderateepheezy, line 41: Used this name `successFailedRetryingRetrySuccessCommentsTemplate` for two reasons
1. The flow is the status is actually `success` -> `failed` -> `retrying` -> `retryingSuccess` -> `success`
2. If I named it starting from prefix failed, I might probably want to name it `failedRetryingRetrySuccessSuccessCommentsTemplate `

singhhari (changes requested):
Noticing a bug when I'm testing the acceptance criteria. I invoked it by doing the following:

- Add breakpoint to Charles and post a comment unsuccessfully
- Pull to refresh a cpl of times, unsuccessfully (nothing should change so far)
- Hit tap to retry and whether it is successful or not, the rest of the comments will disappear from the tableview

Kickstarter-iOS/DataSources/CommentsDataSourceTests.swift
```diff
@@ -5,14 +5,16 @@ import Prelude
 import XCTest
 
 class CommentsDataSourceTests: XCTestCase {
-  let dataSource = CommentsDataSource()
-  let tableView = UITableView()
-  let commentsSection = CommentsDataSource.Section.comments.rawValue
-  let emptySection = CommentsDataSource.Section.empty.rawValue
+  private let dataSource = CommentsDataSource()
```
singhhari, line 8: alphabetize

Kickstarter-iOS/DataSources/CommentsDataSourceTests.swift
```diff
@@ -5,14 +5,16 @@ import Prelude
...
-  let dataSource = CommentsDataSource()
-  let tableView = UITableView()
-  let commentsSection = CommentsDataSource.Section.comments.rawValue
-  let emptySection = CommentsDataSource.Section.empty.rawValue
+  private let commentsSection = CommentsDataSource.Section.comments.rawValue
+  private let dataSource = CommentsDataSource()
+  private let emptySection = CommentsDataSource.Section.empty.rawValue
+  private let tableView = UITableView()
+
+  private let templateComments = Comment.templates + [Comment.retryingTemplate, Comment.retrySuccessTemplate]
```
singhhari, line 13: Is there a cleaner syntax we can use for this? Maybe even create a template that has all these templates within it?

Kickstarter-iOS/Views/Cells/CommentCell.swift
```diff
@@ -109,6 +116,8 @@ final class CommentCell: UITableViewCell, ValueCell {
     self.flagButton.rac.hidden = self.viewModel.outputs.flagButtonIsHidden
     self.replyButton.rac.hidden = self.viewModel.outputs.replyButtonIsHidden
     self.viewRepliesStackView.rac.hidden = self.viewModel.outputs.viewRepliesStackViewIsHidden
+
+    self.postedButton.rac.hidden = self.viewModel.outputs.commentStatus.map { $0 != .retrySuccess }
```
justinswart, line 120: This comparison should be in the view model not in the cell.

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
```diff
@@ -41,6 +41,30 @@ internal final class CommentsViewControllerTests: TestCase {
     }
   }
 
+  func testView_WithSuccessFailedRetryingRetrySuccessComments_ShouldDisplayAll() {
```
msadoon, line 44: Just double check the screenshot is actually showing the different cell states. Because I had an issue where I was committing empty screenshots. Probably not happening here, but worth mentioning.

Kickstarter-iOS/Views/Cells/CommentCell.swift
singhhari, line 120: Did the formatter indent this? I would just alphabetize it again within this method for simplicity sake.

Kickstarter-iOS/Views/Cells/CommentPostFailedCell.swift
```diff
@@ -77,6 +78,13 @@ final class CommentPostFailedCell: UITableViewCell, ValueCell {
 
   internal override func bindViewModel() {
     self.bodyTextView.rac.text = self.viewModel.outputs.body
+
+    self.viewModel.outputs.commentStatus
+      .observeForUI()
+      .observeValues { status in
```
singhhari, line 84: missing a capture list [weak self]

Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -299,6 +302,21 @@ internal final class CommentCellViewModelTests: TestCase {
...
+    let comment = Comment.template
+      |> \.status .~ .retrying
+
+    self.vm.inputs.configureWith(comment: comment, project: .template)
+
+    self.commentStatus.assertValues([.retrying], "The comment status is emitted.")
+
+    self.vm.inputs.bindStyles()
+
+    self.commentStatus.assertValues([.retrying, .retrying], "The comment status is emitted.")
```
msadoon, line 317: Can you test a change of status like going from `.retrying` to `.retrySuccess`? Just makes the test more robust when we know we emit more than one type of value. Also we can update the test name to reflect this.

KsApi/models/templates/CommentsEnvelopeTemplates.swift
singhhari: Hmm I see. I wonder if naming it something simple like `allStatesTemplate` might be more helpful to the reader?

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -59,6 +62,13 @@ public final class CommentCellViewModel:
 
     self.authorName = comment.map(\.author.name)
 
+    let status = comment.map(\.status)
+
+    self.commentStatus = Signal.merge(
```
singhhari, line 67: Does `self.commentStatus = comment.map(\.status).signal.takeWhen(self.bindStylesProperty.signal)` achieve the same desired effect?

singhhari: I think if indented appropriately the readability is good too but I didn't test out whether the syntax actually results in the functionality you want here.

moderateepheezy: We want to always have an initial status before the `self.bindStylesProperty` is called, should there be a need for reset e.g screen orientation change.

msadoon (approved):
Yea aside from the odd pull to refresh issue, it's good to go from me. Left a few small comments as well.

Kickstarter-iOS/Views/Cells/CommentCell.swift
moderateepheezy: This was suppose to be in the VM, I actually forgot it there and I didn't catch it in my review.

Kickstarter-iOS/DataSources/CommentsDataSourceTests.swift
moderateepheezy: I added this because it's only useful in the tests.

Kickstarter-iOS/Views/Controllers/CommentsViewControllerTests.swift
moderateepheezy: Yeah, I checked and it was fine.

Library/ViewModels/CommentCellViewModelTests.swift
moderateepheezy: Good point, I added this and removed it again because I though we are only testing the output, and the logic that changes the state is in the `CommentsViewModel`, do you think there's still a need?

msadoon: No that's fine. 👍🏿

Kickstarter-iOS/Views/Cells/CommentCell.swift
moderateepheezy: Thanks. I missed it too.

KsApi/models/templates/CommentsEnvelopeTemplates.swift
moderateepheezy: The comment has other states that are not included, so `allStatesTemplate` might be.
Maybe we you can help with a better name lol couldn't think of any that captures it.

singhhari: should be fine given the extra info you provided

singhhari (approved):

PR #1508, opened 2021-06-08
[EP-524] Updating context_page values for trackTabBarClicked analytics event

singhhari:
# 📲 What

We're currently sending `other` as the `context_page` for the analytics event that is fired when we select a tab bar item. We should not be sending `other` but rather the value of the tab that we are navigating from.

# 🤔 Why

This is one of the final pieces of our Segment implementation. This is a blocker for production so we are wrapping up all the final bits before we can go live.

# ✅ Acceptance criteria

- [ ] Tap the `Explore` tab from the `Search` tab and verify `context_page` == `search`
- [ ] Tap the `Explore` tab from the `Profile` tab and verify `context_page` == `profile`
- [ ] Tap the `Explore` tab from the `Activity` tab and verify `context_page` == `activity_feed`


Library/Tracking/KSRAnalytics.swift
```diff
@@ -566,12 +566,22 @@ public final class KSRAnalytics {
...
 
-  /// Called when the user taps on a TabBarItem
-  public func trackTabBarClicked(_ tabBarItemLabel: TabBarItemLabel) {
+  /**
+   Called when the user selected a tab bar item.
+
+   - parameter tabBarItemLabel: The tab the user is navigating to.
+   - parameter prevTabBarItemLabel: The tab the user is navigating from.
+   */
+  public func trackTabBarClicked(
```
msadoon, line 575: Seems only returning a `pageContext` for `.discovery` case, maybe change the name here to reflect that?

Library/ViewModels/RootViewModel.swift
```diff
@@ -357,14 +357,20 @@ public final class RootViewModel: RootViewModelType, RootViewModelInputs, RootVi
       }
 
     self.tabBarItemsData
+      .combineLatest(with: prevSelectedTabBarItem.signal)
```
msadoon, line 360: Unrelated to this PR, but the name of this view model doesn't indicate any kind of analytics framework `RootViewModel`. Only reason is because I just created a `RootCommentCellViewModel` and it might get a bit confusing as these two are completely unrelated.

Library/ViewModels/RootViewModelTests.swift
```diff
@@ -492,7 +492,7 @@ final class RootViewModelTests: TestCase {
     self.selectedIndex.assertValues([0, 1, 0, 2], "Selects index immediately.")
 
     XCTAssertEqual(
-      ["other", "discover"],
+      ["activity_feed", "discover"],
```
msadoon, line 495: spelling on line `476` "selected"

msadoon (approved):
Looks good to me codewise, I couldn't log into segment to test the real values, but assuming they are relayed correctly this is good to go!

Library/Tracking/KSRAnalytics.swift
singhhari: Good point although technically the method is still called each time a tab bar is tapped and truthfully this could change in the near future.

Library/ViewModels/RootViewModel.swift
singhhari: The analytics client is part of our `AppEnvironment` so you'll see it referenced in a bunch of different view models like this:

`AppEnvironment.current.ksrAnalytics`

Library/ViewModels/RootViewModelTests.swift
singhhari: Updated.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -566,12 +566,22 @@ public final class KSRAnalytics {
 
   // MARK: - Application Lifecycle
 
-  /// Called when the user taps on a TabBarItem
-  public func trackTabBarClicked(_ tabBarItemLabel: TabBarItemLabel) {
+  /**
+   Called when the user selected a tab bar item.
+
+   - parameter tabBarItemLabel: The tab the user is navigating to.
+   - parameter prevTabBarItemLabel: The tab the user is navigating from.
```
moderateepheezy, line 573: Should be `previousTabBarItemLabel `

singhhari: Just updated.

moderateepheezy (approved):

PR #1510, opened 2021-06-09
[NT-2020] Retrying Paging On Error

justinswart:
# 📲 What

Allows us to retry a pagination failure if a request fails during paging.

# 🤔 Why

Our `paginate` function has, until now, suppressed errors during paging. For the new comments experience we are handling those errors and affording the user the opportunity to retry a failed pagination request.

# 🛠 How

- Modified the `paginate` function in #1504 to emit errors.
- Handling those errors in this PR.
- Configuring the table view's footer view with various states depending on whether there is paging `activity`, an `error` or whether it should be `hidden`.

# ✅ Acceptance criteria

- [x] Using Charles proxy intercept an initial request and abort it - the retry button should be shown in the footer view. Subsequent requests by tapping retry  should succeed and results shown.
- [x] Using Charles proxy intercept an "next page" request and abort it - the retry button should be shown in the footer view. Subsequent requests by tapping retry should succeed and results shown.
- [x] No regressions on previous activity indicator behaviour.


Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -142,10 +147,17 @@ internal final class CommentsViewController: UITableViewController {
...
 
-    self.viewModel.outputs.showLoadingIndicatorInFooterView
+    self.viewModel.outputs.configureFooterViewWithState
       .observeForUI()
-      .observeValues { [weak self] shouldShow in
-        self?.footerView.configureWith(value: shouldShow)
+      .observeValues { [weak self] state in
+        self?.footerView.configureWith(value: state)
+
+        // Force tableFooterView to layout
```
justinswart, line 155: Not pretty but wasn't really successful getting the footer to find the correct size any other way 🤷‍♂️ Open to suggestions.

justinswart: I think an alternative might be to use entirely different footer views and switch them out but if we go that route I'll do it in a separate PR as an iteration on this.

Library/ViewModels/CommentTableViewFooterViewModel.swift
```diff
@@ -17,15 +26,26 @@ public protocol CommentTableViewFooterViewModelType {
...
   CommentTableViewFooterViewModelInputs, CommentTableViewFooterViewModelOutputs {
   public init() {
-    self.shouldShowActivityIndicator = self.loadingIndicatorSignal
+    let state = self.stateProperty.signal.skipNil()
+
+    self.activityIndicatorHidden = state.map { $0 == .activity }.negate()
+    self.retryButtonHidden = state.map { $0 == .error }.negate()
+    self.rootStackViewHidden = state.map { $0 == .hidden }
+    self.bottomInsetHeight = state.map { state -> Int in
+      if case .error = state { return 4 }
```
singhhari, line 35: Why the explicit distinguishing of 2 and 4 in this context? Might be useful to have an inline comment somewhere here.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -221,9 +242,21 @@ public final class CommentsViewModel: CommentsViewModelType,
       .map { ($1, $0) }
       .flatMap(retryCommentProducer)
 
-    self.showLoadingIndicatorInFooterView = Signal
-      .combineLatest(isCloseToBottom, self.beginOrEndRefreshing)
-      .map(second >>> isTrue)
+    let footerViewActivityState = Signal
+      .combineLatest(isCloseToBottom, isLoading)
+      .filter(second >>> isTrue)
```
singhhari, line 247: Can you remind me what the >>> operator does again?

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -831,6 +835,114 @@ internal final class CommentsViewModelTests: TestCase {
     }
   }
 
+  func testConfigureFooterViewWithState_HiddenOnViewDidLoad() {
+    self.configureFooterViewWithState.assertDidNotEmitValue()
+
+    let envelope = CommentsEnvelope.singleCommentTemplate
+    let project = Project.template
```
singhhari, line 842: Do we need this constant for this and the other tests if we are only using it in one location (configureWith) ?

Library/ViewModels/CommentTableViewFooterViewModel.swift
moderateepheezy: This logic here could be a good cause to have different footer view for both state, as the logic seem to be managing the dynamic layout of both state by adjusting the bottom inset.

Kickstarter-iOS/Views/CommentTableViewFooterView.swift
```diff
@@ -29,34 +33,98 @@ final class CommentTableViewFooterView: UIView {
 
     self.bindViewModel()
     self.configureViews()
+
+    self.retryButton.addTarget(self, action: #selector(self.retryButtonTapped), for: .touchUpInside)
```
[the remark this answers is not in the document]
justinswart: init is ok, we've done this in a number of places. it's just important for it to be called once.

Library/ViewModels/CommentTableViewFooterViewModel.swift
justinswart: Oh, it's because with the button we're aligning the content to `top` and this doesn't seem to correctly increase the space below the button so we need to handle this manually here 🤷‍♂️  I can add a comment.

Library/ViewModels/CommentsViewModel.swift
justinswart: Evaluates the composition of these two functions, same as saying:

```swift
.filter { isTrue($0.1) }
```

`second` is a function to pick out the second part of a tuple.

moderateepheezy: in my understanding I think it says, if a function take input type A and return type B, and another function takes input of type B and returns different type C, you can compose them using the `>>>` operator.

```
func toString(_  value: Int) -> String { return String(value) } 
func split(_  value: String) -> [String] { return value.split } 
```

You can compose this by saying 

`toString >>> split`

moderateepheezy (approved):

singhhari (approved):
lgtm

PR #1511, opened 2021-06-10
[NT-2017] Updates – Show new comments experience

moderateepheezy:
# 📲 What

Each update to a project has its own comment section which would also benefit from the new comment threading. 

# 🤔 Why

We want to use the new comment threading UI for comment related part of the app.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Click on a project and navigate a project pamphlet, then click on update and click all way to the comment section. Validate that the update comment section behave the same as the project comment section.


moderateepheezy (commented):
Ignore the rough transition that happens on the composer, it's going to be fixed in this ticket https://kickstarter.atlassian.net/browse/NT-2009

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -86,8 +86,10 @@ public final class CommentsViewModel: CommentsViewModelType,
       }
 
     self.configureCommentComposerViewWithData = Signal
-      .combineLatest(initialProject.signal, currentUser.signal)
-      .takeWhen(self.viewDidLoadProperty.signal)
+      .combineLatest(initialProject, currentUser.signal, self.viewDidLoadProperty.signal.ignoreValues())
```
justinswart, line 89: This doesn't work because it's waiting for the project to load in order to configure the composer view and the view is temporarily shown in a broken state until configured:

image

We either need to update `CommentComposerView` to improve how it looks until configured or manage this a different way. Possibly hide the composer until configured? cc @tobitech.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -86,8 +86,10 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+      .map { project, currentUser, _ in
```
moderateepheezy, line 90: Couldn't remove the `_` completly. The compiler lost context when I tried `.map{($0, $1)}`

Library/ViewModels/CommentsViewModel.swift
justinswart: Ok, I saw your comment now:

> Ignore the rough transition that happens on the composer, it's going to be fixed in this ticket https://kickstarter.atlassian.net/browse/NT-2009

That's fine.

Library/ViewModels/CommentsViewModel.swift
justinswart: Try `.map { ($0.0, $0.1) }`

moderateepheezy: Oh yeah, it's not an array. Thanks!

Kickstarter-iOS/Library/OptimizelyFeature+HelpersTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    let project = self.cosmicSurgery
+      |> Project.lens.state .~ .live
+      |> Project.lens.stats.pledged .~ (self.cosmicSurgery.stats.goal * 3 / 4)
+
+    let mockService = MockService(fetchProjectResponse: project)
+
+    withEnvironment(
+      apiService: mockService, optimizelyClient: mockOptimizelyClient
+    ) {
+      XCTAssert(commentsViewController(for: project).isKind(of: CommentsViewController.self))
```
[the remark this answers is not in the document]
moderateepheezy: Makes more sense this way. Thanks.

Kickstarter-iOS/Library/OptimizelyFeature+Helpers.swift
```diff
@@ -0,0 +1,13 @@
+import KsApi
+import Library
+import UIKit
+
+public func commentsViewController(for project: Project? = nil,
```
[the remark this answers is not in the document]
moderateepheezy: I think we have couple of feature flags for the comment threading that needs to do the same.

justinswart (approved):

PR #1512, opened 2021-06-10
[NT-2033] Allow Concurrent Retrying/Posting Of Comments

justinswart:
# 📲 What

Allows us to retry and post comments in a non-blocking manner (up to a reasonable limit).

# 🤔 Why

We found that if we were attempting to retry multiple comments at once, newer retries wouldn't start until older ones had completed. Also, tapping cells repeatedly while blocked in this way would result in duplicate comments being created due to post being called multiple times once the thread was unblocked.

# 🛠 How

- Changed to a `.concurrent` `flatMap` strategy allowing us to specify an amount of concurrent operations to run on the queue. 
  - I've set this to `3` which seemed like a reasonable amount to me but we could consider a higher number if necessary.
- Also added some duplication filtering to prevent against retrying the same comment while its request is already in flight.

# 👀 See

https://user-images.githubusercontent.com/3735375/121602939-e53f3b00-c9fc-11eb-9247-af8c19ee18f2.mp4

# ✅ Acceptance criteria

- [ ] Create several errored comments, tap them in quick succession. The UI should not block requests up to the limit set and each comment should transition through the necessary states.
- [ ] Try to post multiple times as quickly as possible, the dialogue should not block and you should not find yourself tapping post multiple times to post duplicate comments.

### DIFF
--- a/Library/ViewModels/CommentsViewModel.swift
+++ b/Library/ViewModels/CommentsViewModel.swift
@@ -3,7 +3,7 @@ import Prelude
 import ReactiveExtensions
 import ReactiveSwift
 
-private let concurrentCommentLimit: UInt = 3
+private let concurrentCommentLimit: UInt = 5
 
 public protocol CommentsViewModelInputs {
   /// Call when the User is posting a comment or reply.
--- a/Library/ViewModels/CommentsViewModel.swift
+++ b/Library/ViewModels/CommentsViewModel.swift
@@ -3,6 +3,8 @@ import Prelude
 import ReactiveExtensions
 import ReactiveSwift
 
+private let concurrentCommentLimit: UInt = 3
+
 public protocol CommentsViewModelInputs {
   /// Call when the User is posting a comment or reply.
   func commentComposerDidSubmitText(_ text: String)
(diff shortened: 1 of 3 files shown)


Library/ViewModels/CommentsViewModel.swift
```diff
@@ -177,12 +179,7 @@ public final class CommentsViewModel: CommentsViewModelType,
       .map(unpack)
       .map(commentsReplacingCommentById)
 
-    self.resetCommentComposerAndScrollToTop = commentsWithFailableOrComment
-      // We only want to scroll to top for failable comments as they represent the initial
-      // comment that's inserted. We check here to see that the first comment's ID is a UUID
-      // to determine this. There may be better ways.
-      .map { UUID(uuidString: $0.first?.id ?? "") }
-      .filter(isNotNil)
+    self.resetCommentComposerAndScrollToTop = self.commentComposerDidSubmitTextProperty.signal.skipNil()
```
justinswart, line 182: I've changed this so that we don't actually wait for the failable comment to be created, simply knowing the the user posted text should be enough to know that it _will_ be created and placed at the top. This also reliably always clears the textfield.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -235,12 +232,26 @@ public final class CommentsViewModel: CommentsViewModelType,
     )
     .takePairWhen(commentComposerDidSubmitText)
     .map(unpack)
-    .flatMap(postCommentProducer)
+    .flatMap(.concurrent(limit: concurrentCommentLimit), postCommentProducer)
```
justinswart, line 235: The `.concurrent(limit:)` strategy here allows use to schedule multiple `SignalProducer`s on the queue simultaneously, I have the limit set to 3 because it seems like a good amount but it can likely be quite a bit higher. We need to bear in mind that it will spawn multiple threads and one would not want this to overload the system.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -3,6 +3,8 @@ import Prelude
 import ReactiveExtensions
 import ReactiveSwift
 
+private let concurrentCommentLimit: UInt = 3
```
singhhari, line 6: Ah I left four pending comments while I was testing this (before looking at the code) and noticed the last one was not budging when I tapped it. This explains it.

singhhari (approved):
lgtm!

PR #1515, opened 2021-06-11
[NT-2046] Allow multiple identify calls for logged in users

jgsamudio:
# 📲 What

We noticed that when opening the app when the user is logged in, the app does not identify the user. This is due to the app identifying the user while the segment feature is disabled. We store the identify data even if the identify calls fail due to the feature flag disabling the integration. When the feature flag is then enabled, the app does not identify users because it had values stored in user defaults when in reality it has never identified the user. 

After syncing with Paulo, we've decided to allow multiple identify calls if we can be sure that the user is identified when the app opens. 

# 🛠 How

The stored analytics identify data is removed in favor of comparing old and new user values. This will force the app to identify on app open since the old user value is nil and the new user value is the stored user. 

[NT-2046](https://kickstarter.atlassian.net/browse/NT-2046)

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -9,8 +9,8 @@ public final class KSRAnalytics {
   internal private(set) var config: Config?
   private let device: UIDeviceType
   internal private(set) var loggedInUser: User? {
-    didSet {
-      self.identify(self.loggedInUser)
+    willSet {
+      self.identify(oldUser: loggedInUser, newUser: newValue)
     }
   }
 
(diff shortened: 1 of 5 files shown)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:68:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalytics.swift:13:1 | warning: (redundantSelf) Insert/remove explicit self where applicable. |
| Library/Tracking/KSRAnalyticsTests.swift:1553:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

Library/Tracking/KSRAnalytics.swift
```diff
@@ -533,24 +533,19 @@ public final class KSRAnalytics {
...
       self.segmentClient?.reset()
       return
     }
 
-    let previousIdentityData = AppEnvironment.current.userDefaults.analyticsIdentityData
-
-    let newData = KSRAnalyticsIdentityData(user)
-
-    guard newData != previousIdentityData else { return }
+    guard newUser != oldUser else { return }
```
justinswart, line 542: Unfortunately due to our limited `Equatable` conformance in `User` (https://github.com/kickstarter/ios-oss/blob/master/KsApi/models/User.swift#L102), this will probably cause no `identify` calls to occur when notification options are updated.

I think we should just remove this de-duping altogether and optimize if it becomes an issue.

justinswart (changes requested):

Library/Tracking/KSRAnalytics.swift
jgsamudio: Updated logic to compare the analytics data instead of the user data

justinswart (approved):
Thanks! Added one test 👍

PR #1522, opened 2021-06-18

moderateepheezy:
# 📲 What

For every project `Update` we want to fetch all comments including pagination related to the project update only.

# 🤔 Why

Comments under updates should benefit from the same new experience as project comments. However, when we fetch comments for updates we were fetching  project comments, this PR changes the implementation to fetch comments related to project updates.  

# 🛠 How

In order to retrieve the comments from the updates we need to wrap the comments data in a `FreeformPost` fragment like this

```
{
  post(id: "RnJlZWZvcm1Qb3N0LTMxNTI4NjY=") {
    ... on FreeformPost {
      comments(first: 25) {
        edges {
          node {
            author {
              id
              imageUrl(width: 200)
              isCreator
              name
            }
            authorBadges
            body
            createdAt
            deleted
            id
            parentId
            replies {
              totalCount
            }
          }
        }
        pageInfo {
          endCursor
          hasNextPage
        }
        totalCount
      }
      id
    }
  }
}
```

# 👀 See

Update

# ✅ Acceptance criteria

- [x] Navigate to Project updates, confirm that each updates fetch it's own comments, and no regression on pagination.
- [x] Confirm that there's no regression on the Project comments and pagination.
- [ ]* Test project: https://staging.kickstarter.com/projects/tundralabs/tundra-tracker/posts/3143964


KsApi/models/graphql/GraphCommentsEnvelopeTests.swift
```diff
@@ -103,4 +104,107 @@ final class GraphCommentsEnvelopeTests: XCTestCase {
...
+    guard let data = try? JSONSerialization.data(withJSONObject: dictionary, options: []) else {
+      XCTFail("Should have data")
+      return
+    }
+
+    do {
+      let envelope = try JSONDecoder().decode(GraphCommentsEnvelope.self, from: data)
+
+      XCTAssertEqual(envelope.comments[0].id, "Q29tbWVudC0zMDQ5MDQ2NA==")
+      XCTAssertNil(envelope.comments[0].parentId)
```
msadoon, line 188: why do we need a parent id?

KsApi/models/graphql/adapters/CommentsEnvelope+GraphCommentsEnvelopeTests.swift
```diff
@@ -12,6 +13,22 @@ final class CommentsEnvelope_GraphCommentsEnvelopeTests: XCTestCase {
...
+
+    XCTAssertEqual(
+      envelope.comments,
+      [Comment.comment(from: .template), Comment.comment(from: .template), Comment.comment(from: .template)]
+    )
+    XCTAssertEqual(envelope.hasNextPage, true)
+    XCTAssertEqual(envelope.cursor, "WzMwNDkwNDY0XQ==")
+    XCTAssertEqual(envelope.totalCount, 100)
+    XCTAssertEqual(envelope.updateID, 1_234)
     XCTAssertEqual(envelope.slug, "jadelabo-j1-beautiful-powerful-and-smart-idex-3d-printer")
```
msadoon, line 32: I guess real data would not return both a `slug` and an `updateId` for project update comments, may want the test to reflect this like in `testProjectComments`

KsApi/models/CommentsEnvelope.swift
```diff
@@ -6,6 +6,7 @@ public struct CommentsEnvelope: Decodable {
   public var cursor: String?
   public var hasNextPage: Bool
   public var slug: String
+  public var updateID: Int?
```
singhhari, line 9: alphabetize

KsApi/models/graphql/GraphCommentsEnvelope.swift
```diff
@@ -6,14 +6,17 @@ struct GraphCommentsEnvelope: Decodable {
   var cursor: String?
   var hasNextPage: Bool
   var slug: String
+  var updateID: Int?
```
singhhari, line 9: alphabetize

KsApi/models/graphql/adapters/CommentsEnvelope+GraphCommentsEnvelopeTests.swift
msadoon: Also may want to create a new template just for project update comments in `GraphCommentsEnvelope` and use that instead of modifying the existing template meant for project comments.

KsApi/models/graphql/GraphCommentsEnvelope.swift
```diff
@@ -24,12 +27,19 @@ extension GraphCommentsEnvelope {
 
   init(from decoder: Decoder) throws {
     let values = try decoder.container(keyedBy: CodingKeys.self)
+    let container: KeyedDecodingContainer<GraphCommentsEnvelope.CodingKeys>
```
singhhari, line 30: can this implementation be done without declaring this constant here? I think you can avoid it?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -414,3 +403,38 @@ private func commentsReplacingCommentById(
...
   return mutableComments
 }
+
+private func commentsFirstPage(from projectOrUpdate: Either<Project, Update>)
+  -> SignalProducer<CommentsEnvelope, ErrorEnvelope> {
+  return projectOrUpdate.ifLeft {
+    AppEnvironment.current.apiService.fetchComments(
+      query: commentsQuery(
+        withProjectSlug:
+        $0.slug
```
msadoon, line 413: May want to make this clear as `project`

KsApi/models/graphql/GraphCommentsEnvelopeTests.swift
```diff
@@ -103,4 +104,107 @@ final class GraphCommentsEnvelopeTests: XCTestCase {
...
+
+    guard let data = try? JSONSerialization.data(withJSONObject: dictionary, options: []) else {
+      XCTFail("Should have data")
+      return
+    }
+
+    do {
+      let envelope = try JSONDecoder().decode(GraphCommentsEnvelope.self, from: data)
+
+      XCTAssertEqual(envelope.comments[0].id, "Q29tbWVudC0zMDQ5MDQ2NA==")
```
singhhari, line 187: Can you replace `[0]` with `first`?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -414,3 +403,38 @@ private func commentsReplacingCommentById(
...
+      )
+    )
+  } ifRight: {
+    AppEnvironment.current.apiService.fetchComments(
+      query: projectUpdateCommentsQuery(id: $0.id)
```
msadoon, line 418: May want to make this clear as `update`

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -139,9 +139,9 @@ public final class CommentsViewModel: CommentsViewModelType,
     let pullToRefresh = self.refreshProperty.signal
 
     let requestFirstPageWith = Signal.merge(
-      initialProject,
-      initialProject.takeWhen(retryFirstPage),
-      initialProject.takeWhen(pullToRefresh)
+      projectOrUpdate,
+      projectOrUpdate.takeWhen(retryFirstPage),
+      projectOrUpdate.takeWhen(pullToRefresh)
```
msadoon, line 144: Can we write a test for this when no project is sent in, but only an update and see that comments are returned/ paginated, can pull to refresh?
I believe the tests are already written, it's just a matter of updating the input to be an `Update` instead of a `Project`.

msadoon (changes requested):
Code looks good, just some changes mostly around testing.
Also, not sure if I can test this in app yet, I only see `goToUpdates` function in `ProjectPamphletContentViewController` using `ProjectUpdatesViewController` instead of `func commentsViewController`.

msadoon:
> Code looks good, just some changes mostly around testing.
> Also, not sure if I can test this in app yet, I only see `goToUpdates` function in `ProjectPamphletContentViewController` using `ProjectUpdatesViewController` instead of `func commentsViewController`.

Ok, please ignore my comment above about in-app testing, seems to work in the screenshot you provided above.

KsApi/models/graphql/GraphCommentsEnvelope.swift
moderateepheezy: I don't think so because you can't use `try?` with nil-coalescing.

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: There are already some tests check `func testUpdatePagination_WhenLimitReached_CommentsAreLoadedIntoDataSource()` and some others below it.

KsApi/models/graphql/GraphCommentsEnvelope.swift
```diff
@@ -6,14 +6,17 @@ struct GraphCommentsEnvelope: Decodable {
   var cursor: String?
   var hasNextPage: Bool
   var slug: String
```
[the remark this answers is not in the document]
msadoon: Can move this `self.slug = try container.decodeIfPresent(String.self, forKey: .slug)` inside the `if`

KsApi/models/graphql/GraphCommentsEnvelopeTests.swift
msadoon: I removed the `parentId` from `GraphComment` and `GraphCommentTemplates` and the app still compiled, so maybe we don't use it anywhere?

Library/ViewModels/CommentsViewModel.swift
msadoon: ok, good point, looks good to me!

KsApi/models/graphql/GraphCommentsEnvelopeTests.swift
moderateepheezy: parentId is used here`PostCommentInput`.

msadoon (approved):
good to go! 
For some reason Updates page doesn't work on staging (not related to this PR), but I get "HTTP Basic: Access denied"

moderateepheezy:
> good to go!
> For some reason Updates page doesn't work on staging (not related to this PR), but I get "HTTP Basic: Access denied"

It's working for me, maybe you are not logged in or there's an issue with that particular project

Simulator Screen Shot - iPhone 12 Pro Max - 2021-06-21 at 20 43 49

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -414,3 +403,39 @@ private func commentsReplacingCommentById(
...
+      )
+    )
+  } ifRight: {
+    AppEnvironment.current.apiService.fetchComments(
+      query: projectUpdateCommentsQuery(id: $0.id.description)
+    )
+  }
+}
+
+private func commentsNextPage(from projectSlug: String?, cursor: String?,
```
justinswart, line 423: For readability format this to be:

```swift
private func commentsNextPage(
  from projectSlug: String?,
  cursor: String?,
  updateID: String?
) -> SignalProducer<CommentsEnvelope, ErrorEnvelope> {
```

singhhari (approved):
lgtm

KsApi/models/graphql/GraphCommentsEnvelope.swift
justinswart: I was just curious if this was actually possible and looks like it can be done this way:

```swift
/// Try to decode a `Project` then `Update` container.
let tryContainer = (
  (try? values.nestedContainer(keyedBy: CodingKeys.self, forKey: .project)) ??
  (try? values.nestedContainer(keyedBy: CodingKeys.self, forKey: .post))
)

guard let container = tryContainer else {
  throw DecodingError.dataCorruptedError(
    forKey: .post,
    in: values,
    debugDescription: "Unable to decode a Project or Update."
  )
}

self.slug = try container.decodeIfPresent(String.self, forKey: .slug)
self.updateID = try container.decodeIfPresent(String.self, forKey: .id)
```

I had hoped you could coalesce from `try?` to `try` but Swift doesn't allow you to put a `try` on the right-hand side of the `??` operator.

justinswart (approved):

PR #1525, opened 2021-06-21
[NT-1947] UI/UX - Error Loading Root Comments

tobitech:
# 📲 What

Add a new view to present an error state whenever something goes wrong while loading root comments at the initial stage.

# 🤔 Why

Because every now and then things will fail, we need to handle problems loading comments.

# 👀 See

# ✅ Acceptance criteria

1. Launch the app and select a project.
2. Tap on Comments to view the projects comments and simulate an error scenario either by interrupting the initial comments loading request or by tampering with the comments query prior to launching the app from Xcode.
3. An error state view should appear when the request is done that shows that there was an error with loading the comments.


Library/ViewModels/CommentsViewModel.swift
```diff
@@ -252,10 +260,21 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+    )
+
+    self.configureFooterViewWithState = Signal.merge(
+      hideFooterView,
       footerViewActivityState.mapConst(.activity),
-      errors.mapConst(.error)
+      errors
+        .mapConst(.error)
+        .withLatestFrom(hasRequestedNextPage.filter(isTrue))
+        .map(first)
```
msadoon, line 277: I'm a bit confused by this logic, does it mean that only next page requests with an error hide the footer view?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -252,10 +260,21 @@ public final class CommentsViewModel: CommentsViewModelType,
       self.loadCommentsAndProjectIntoDataSource.ignoreValues()
     )
 
-    self.configureFooterViewWithState = Signal.merge(
+    let hideFooterView: Signal<CommentTableViewFooterViewState, Never> = Signal.merge(
       initialLoadOrReload.mapConst(.hidden),
+      errors
+        .mapConst(.hidden)
+        .withLatestFrom(hasRequestedNextPage.filter(isFalse))
+        .map(first)
```
msadoon, line 268: Does this mean we only hide the footerview if it's a first page request with an error?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -197,10 +197,20 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+        .map { comments, project in (comments, project, false) },
+      // Continue to paginate normally.
+      commentsAndProject
+        .filter { comments, _ in comments.isEmpty == false }
+        .map { comments, project in (comments, project, false) },
+      // If there are errors emit empty comments array, project and error boolean.
+      Signal.combineLatest(errors, hasRequestedNextPage)
+        .filter(second >>> isFalse)
+        .map(first)
+        .withLatestFrom(initialProject).map { ([], $1, true) }
```
msadoon, line 213: This is quite confusing for me, not sure if I follow it, but wouldn't be easier to just replace changes in this VM with:
```
    let showErrorLoadingForInitialComments = Signal.zip(initialLoadOrReload.signal,
                                                        errors.signal).mapConst(true)
```
and then create an output that just reloads the table view with an error state?

I'm not 100% of the side effects for these changes, that's all, because it touches to empty state, as well as the error states on pagination. Best to keep all of that separate imo.

msadoon (commented):
Looks like it's working on device. Didn't test every possible state change (ie. pagination errors)
I just can't really understand the full scope of the possible side effects in the view model. Suggested a simpler alternative because the ticket only asks for an error state on intial load of comments before the empty state, but happy to discuss.

Kickstarter-iOS/Views/Cells/CommentsErrorCell.swift
```diff
@@ -0,0 +1,88 @@
+import Library
+import Prelude
+import UIKit
+
+final class CommentsErrorCell: UITableViewCell, ValueCell {
+  private let rootStackView: UIStackView = {
```
singhhari, line 6: alphabetize

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -212,6 +214,18 @@ extension CommentsViewController: CommentTableViewFooterViewDelegate {
...
 
+// MARK: - UITableViewDelegate
+
+extension CommentsViewController {
+  override func tableView(_: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+    if self.dataSource.isErrorStateIndexPath(indexPath) {
+      return (self.view.safeAreaLayoutGuide.layoutFrame.height * 0.70)
+    }
+
+    return UITableView.automaticDimension
```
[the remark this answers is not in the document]
singhhari: I may be out of the loop here but where is 0.70 coming from?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -49,7 +49,7 @@ public protocol CommentsViewModelOutputs {
   var goToCommentReplies: Signal<Comment, Never> { get }
 
   /// Emits a list of `Comment`s and the `Project` to load into the data source.
```
singhhari, line 51: Update doc to reflect the new parameter added here.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -197,10 +197,20 @@ public final class CommentsViewModel: CommentsViewModelType,
...
 
     self.loadCommentsAndProjectIntoDataSource = Signal.merge(
-      // If we start off with no comments, show an empty state.
-      commentsAndProject.take(first: 1),
-      // Subsequent empty emissions (pull to refresh) should be ignored until replaced.
-      commentsAndProject.skip(first: 1).filter { comments, _ in comments.isEmpty == false }
+      // Allow empty arrays from the first emission.
+      Signal.zip(comments, initialProject)
+        .filter { $0.0.isEmpty }
+        .map { ($0.0, $0.1) }
```
singhhari, line 203: Lines 199-214 are doing a lot and it's a bit hard to follow along. I think it would be best to take the three separate signals you are merging here and put them into separate constants for readability.

Library/ViewModels/CommentsViewModel.swift
singhhari: Add some inline docs here.

Library/ViewModels/CommentsViewModel.swift
singhhari: Add some inline docs here too.

singhhari (changes requested):

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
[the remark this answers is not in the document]
tobitech: `0.70` is just 70% of the parent view's height. Full height was showing the scrolling indicator when error state cell is shown and I wanted to prevent that.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -3564,6 +3567,7 @@
 				94114D47264EEA2D0063E8F6 /* CommentCell.swift */,
 				94114D5A265305200063E8F6 /* CommentPostFailedCell.swift */,
 				94114D5F2653135E0063E8F6 /* CommentRemovedCell.swift */,
+				20455E6626712A6A00F55EDC /* CommentsErrorCell.swift */,
```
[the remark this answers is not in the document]
tobitech: It's alphabetized, `R` comes before `S`

Library/ViewModels/CommentsViewModel.swift
tobitech: Yes that is what it means @msadoon I will add some inline docs

Library/ViewModels/CommentsViewModel.swift
tobitech: The error state would show if there is an error when you pull to refresh.

msadoon: I think that's ok because pull to refresh is like loading the first page. cc @singhhari

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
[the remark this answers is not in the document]
singhhari: Ah ok is that a product requirement? I would leave a comment in place, if so, otherwise it seems like a magic number.

Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -63,6 +67,7 @@ internal final class CommentsViewController: UITableViewController {
     self.tableView.registerCellClass(CommentPostFailedCell.self)
     self.tableView.registerCellClass(CommentRemovedCell.self)
     self.tableView.registerCellClass(EmptyCommentsCell.self)
+    self.tableView.registerCellClass(CommentsErrorCell.self)
```
singhhari, line 70: alphabetize

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -196,12 +196,29 @@ public final class CommentsViewModel: CommentsViewModelType,
       // Thread hop so that we don't circularly buffer.
       .ksr_debounce(.nanoseconds(0), on: AppEnvironment.current.scheduler)
 
+    // Allow empty arrays from the first emission.
+    let emptyCommentsWithInitialProject = Signal.zip(comments, initialProject)
+      .filter { $0.0.isEmpty }
```
singhhari, line 201: I think we should use the variable names instead of $0.0 and $0.1 for readability in line 201 and 202.

singhhari (changes requested):
Working as expected and looks good. Left a couple of minor suggestions.

msadoon (approved):
Good to go!
Found this when pulling to refresh a next page failed state. It shows both the error of the first page load and the next page error load.
May not be a blocker, just thought I'd mention it.

singhhari (approved):
lgtm

PR #1520, opened 2021-06-16
[NT-2006][NT-2013] Update Dependencies Latest Version

msadoon:
# 📲 What

Updating some of our outdated frameworks on `ios-oss`. See _Why_ section for more detail.

# 🤔 Why

As part of regular maintanence keep these libraries up to date to not introduce regressions with new version of iOS and improve stability of dependencies.

Also moves the supported frameworks to using `XCFramework` which are supported in XCode 11+. Hopefully improving our build times on CI but also containing support for multiple frameworks (iOS, MacOS, WatchOS, etc...) in one.

Frameworks upgraded:
```
**xcframeworks**
AlamofireImage "4.1.0" -> "4.1.0" (Latest: "4.2.0") 
FirebaseAnalyticsBinary "7.4.0" -> "7.4.0" (Latest: "8.1.1") 
FirebaseInAppMessagingBinary "7.4.0" -> "7.4.0" (Latest: "8.1.0") 
ReactiveSwift "6.5.0" -> "6.5.0" (Latest: "6.6.0") 
FirebasePerformanceBinary "7.4.0" -> "7.4.0" (Latest: "8.1.0") 
FirebaseCrashlyticsBinary "7.4.0" -> "7.4.0" (Latest: "8.1.0") 
swift-sdk "v3.6.1" -> "v3.6.1" (Latest: "v3.8.0") 
FirebaseProtobufBinary "7.4.0" -> "7.4.0" (Latest: "7.6.0") 

**regular framework**
PerimeterX "1.13.9" -> "1.13.9" (Latest: "1.14.1")  
NOTE: This framework (above) did not need to be included in any target. App still builds and runs, all tests pass.
appcenter-sdk-apple "4.0.0" -> "4.0.0" (Latest: "4.1.1") 
appboy_ios_sdk_full "4.1.0" -> "4.1.0" (Latest: "4.3.0")
```

Frameworks to be upgraded (in a separate PR):
```
analytics-ios "4.1.2" -> "4.1.2" (Latest: "4.1.4")
appboy-segment-ios "4.0.0" -> "4.0.0" (Latest: "4.2.0")
facebook-objc-sdk "v6.5.2" -> "v6.5.2" (Latest: "v11.0.0") <==== MAJOR
stripe-ios "v18.0.0" -> "v18.0.0" (Latest: "21.6.0") <==== MAJOR
```

This PR does not resolve all of these outdated dependencies:

```
analytics-ios "4.1.2" -> "4.1.2" (Latest: "4.1.4")
[...]

PR #1526, opened 2021-06-22
[NT-2006][NT-2013] Update Carthage Bootstrap to use XCFrameworks

msadoon:
In order to avoid this error on [the xc-frameworks pr](https://github.com/kickstarter/ios-oss/pull/1520)
which is a CI failure, we need to add this flag to the ci before running the tests.


msadoon:
Unneeded if [base PR](https://github.com/kickstarter/ios-oss/pull/NT-2006/update-dependencies-latest-version) passes all tests on CI.

PR #1529, opened 2021-06-22

moderateepheezy:
# 📲 What

Fix issue where comments for some project update are not showing until pull to refresh.

# 🤔 Why

Fetched comments for some project update doesn't render in the TableView because `initialProject` and `paginatedComments` using `takePairWhen` signal emits value at the same time and prevents `takePairWhen` from emitting.

# 🛠 How

Replaced

```
let commentsAndProject = initialProject
      .takePairWhen(paginatedComments)
      .map { ($1, $0) }
```
with
```
let commentsAndProject = Signal.combineLatest(
      paginatedComments,
      initialProject
      )
```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|||

# ✅ Acceptance criteria

- [x] When you navigate to comments for each update in a project, they show their respective comments.
- [x] No regression on project comments and pagination.


singhhari (commented):
Working as expected. Coverage is down a bit though. I think there way be an existing unit test you can update to bump this.

moderateepheezy:
> Working as expected. Coverage is down a bit though. I think there way be an existing unit test you can update to bump this.

Yeah, I noticed now and already added it.

singhhari (approved):

PR #1530, opened 2021-06-23

moderateepheezy:
# 📲 What

Fix issue where comments posted for a project update are reflecting on project comment section.

# 🤔 Why

We want to separate comments for project and comments for a project update, there for every comment posted on a project update should not reflect on project comments, and vice-versa.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|||

# ✅ Acceptance criteria

- [x] Confirm that posting a comment on a project update doesn't reflect on the project comment section.
- [x] Confirm that posting a comment on a project doesn't reflect on any of the project update comment section.
- [x] Confirm that when you retry a failed comment post on a project update, it doesn't reflect on project comment, and vice-versa.
- [ ] Verify that there's no regression on project comments section.
- [x] Confirm that pagination works as expected for both project comments and project update comments.


Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -192,7 +192,7 @@ extension CommentsViewController {
   override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
     guard let comment = self.dataSource.comment(at: indexPath) else { return }
 
-    self.viewModel.inputs.didSelectComment(comment)
+    self.viewModel.inputs.viewReplies(for: comment)
```
singhhari, line 195: I think didSelectComment is the right name for the input. We usually follow the convention of the input being similarly named to the delegate method it is called in.

KsApi/GraphSchema.swift
```diff
@@ -45,6 +45,10 @@ public func decodeBase64(_ input: String) -> String? {
     .flatMap { String(data: $0, encoding: .utf8) }
 }
 
+public func encodeToBase64(_ input: String) -> String {
```
singhhari, line 48: alphabetize

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -216,13 +216,30 @@ public final class CommentsViewModel: CommentsViewModelType,
...
-    )
-    .takePairWhen(commentComposerDidSubmitText)
-    .map(unpack)
-    .flatMap(.concurrent(limit: concurrentCommentLimit), postCommentProducer)
+    let commentableId = projectOrUpdate
+      .flatMap { projectOrUpdate in
+        projectOrUpdate.ifLeft { project in
+          SignalProducer.init(value: project.graphID)
+        } ifRight: { update in
+          SignalProducer.init(value: encodeToBase64("FreeformPost-\(update.id)"))
```
singhhari, line 225: What is the relationship between `Update.swift` and `FreeformPost`. Can we conform to GraphIDBridging like we do in the other graph schema classes so we get the `.graphID` property for use here?

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -216,13 +216,30 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+        }
+      }
+
+    let postFailableCommentConfigData:
+      Signal<(project: Project, commentableId: String, user: User), Never> = Signal.combineLatest(
+        initialProject,
+        commentableId,
+        currentUser.skipNil()
+      ).map { configData in
```
singhhari, line 234: replace with:

`.map { project, commentableId, user in
(project: project, commentableId: commentableId, user: user)
}` for readability.

Library/ViewModels/CommentsViewModel.swift
```diff
@@ -216,13 +216,30 @@ public final class CommentsViewModel: CommentsViewModelType,
...
+        (project: configData.0, commentableId: configData.1, user: configData.2)
+      }
+
+    self.failableOrComment <~ postFailableCommentConfigData
+      .takePairWhen(commentComposerDidSubmitText)
+      .map { data in
+        (data.0.project, data.0.commentableId, data.0.user, data.1)
```
singhhari, line 241: I think you can decompose this a bit for readability as well. Kinda like:

      `.map { data in
        let ((project, commentableId, user), text) = data
        return (project, commentableId, user, text)
      }`

singhhari (changes requested):

Library/ViewModels/CommentsViewModel.swift
moderateepheezy: So we can use `GraphIDBridging` however the modelName would be `FreeformPost` , the only relationship between `Update` and `FreeformPost` is that the response to fetch comments for Update is wrapped inside `FreeformPost`.

msadoon (commented):
Codewise seems fine, may want to double check failable and retry functionality carefully between project and update comments to ensure no regression.

msadoon (approved):
Tested basic posting between project and updates in app, may want to double-check retryable comments (the multiple failed comments case) but otherwise looks good to me.

singhhari (approved):
lgtm

PR #1533, opened 2021-06-29
[NT-2034] UI – Reply button not showing for creators/collaborators

moderateepheezy:
# 📲 What

Reply button not showing on comments the user is the creator or a collaborator of the project.

# 🤔 Why

* Creators and Collaborators should be able to reply to comments on their projects even if they are not backing. Currently that's not happening as the reply button is hidden for creators and collaborators until they back.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

    1. Log in with paulomcg@icloud.com / KSR2021
    2.  Go to the dashboard tab
    3. View THE project
    4. Go to comments
    5. Reply button should show because the user is a collaborator.

- [ ] Projects where user is not logged in, not a backing or  not a creator/collaborator should not have reply button.

### DIFF
--- a/Library/ViewModels/CommentCellViewModel.swift
+++ b/Library/ViewModels/CommentCellViewModel.swift
@@ -117,15 +117,15 @@ public final class CommentCellViewModel:
       .map { _ in AppEnvironment.current.currentUser }
       .map(isNil)
 
-    let isNotABacker = self.commentAndProject.signal
+    let isNotABackerCreatorOrCollaborator = self.commentAndProject.signal
       .skipNil()
       .map { _, project in project }
       .skipNil()
-      .map(userIsBackingProject)
+      .map(userIsBackingCreatorOrCollaborator)
       .negate()
 
     // If the user is either logged out, not backing or the flag is disabled, hide replyButton.
-    self.replyButtonIsHidden = Signal.combineLatest(isLoggedOut, isNotABacker)
+    self.replyButtonIsHidden = Signal.combineLatest(isLoggedOut, isNotABackerCreatorOrCollaborator)
       .map(replyButtonHidden)
 
     // If both the replyButton and flagButton should be hidden, the entire stackview will be hidden too.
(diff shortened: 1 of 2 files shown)


Library/ViewModels/CommentCellViewModelTests.swift
```diff
@@ -213,6 +213,40 @@ internal final class CommentCellViewModelTests: TestCase {
     }
   }
 
+  func testOutputs_replyButtonIsHidden_FeatureFlag_True_IsCreatorOrCollaborator_False() {
```
msadoon, line 216: May want to change this name to `IsNotCreatorOrCollaborator` because `IsCreatorOrCollaborator_False` is the same as `IsNotCreatorOrCollaborator_True` below. Also may want to explicitly state the `project.personalization.backing` to `nil` and `project.personalization.isBacking` to `.some(false)` to make it clear you're testing the `.memberData.permissions` as the creator/collaborator.

msadoon (approved):
Good to go!

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -190,9 +190,14 @@ public final class CommentCellViewModel:
   public var outputs: CommentCellViewModelOutputs { self }
 }
 
-private func replyButtonHidden(isLoggedOut: Bool, isNotABacker: Bool) -> Bool {
+private func userIsBackingCreatorOrCollaborator(_ project: Project) -> Bool {
+  return (project.personalization.backing != nil || project.personalization.isBacking == .some(true)) ||
+    !project.memberData.permissions.isEmpty
```
singhhari, line 195: Looking at the enum for `Permission` I don't see anything specifically suggesting creator or collaborator. How does this work?

singhhari (commented):
Works as expected but left a question about the implementation.

Library/ViewModels/CommentCellViewModel.swift
moderateepheezy: When you fetch projects, each project comes with permissions relating to the current user.

singhhari: I understand but I am unclear what specifically about this code you've written represents a `creator` or `collaborator`.

moderateepheezy: So on a project MemberData is tied to current logged in user's token, i.e the data for MemberData is personalized to the user.

For example 
This is a snippet response when I logged in with my account to view project "The Project"

```
{
	"id": 1641211140,
	"photo": {
		"key": "assets/033/406/309/9c8d4de5de265da3db0b4fc4c949a557_original.jpg"
	},
	"friends": [],
	"is_starred": false,
	"is_backing": false,
	"permissions": [],
}
```

This is the snippet when I login with Paulo's details who is the creator of  "The Project"

```
{
	"id": 1641211140,
	"photo": {
		"key": "assets/033/406/309/9c8d4de5de265da3db0b4fc4c949a557_original.jpg",
	},
	"friends": [],
	"is_starred": false,
	"is_backing": false,
	"permissions": ["edit_project", "edit_faq", "post", "comment", "view_pledges", "fulfillment"],
}
```

singhhari (approved):
lgtm although I would like @justinswart to confirm the usage of the `permissions` enum is fine here.

Library/ViewModels/CommentCellViewModel.swift
justinswart: @singhhari yeah a valid observation but it seems that in the past all that was done to determine whether the user is a creator or collaborator was to check if this array is empty or not 🤷‍♂️ 

Example: https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/MessagesViewModel.swift#L146

PR #1538, opened 2021-07-01
[NT-1917, NT-2062] Apollo GraphQL Client Implementation

justinswart:
# 📲 What

Implements the Apollo GraphQL Client and updates our existing root comment queries to use it.

# 🤔 Why

We're moving all of our GraphQL queries and mutations to use Apollo as a more robust solution moving forward.

# 🛠 How

This branch started out as a spike implementation to inform our engineering plan for this work. Most of the detail is contained that plan and has been discussed in our team. The plan is referenced below.

During this spike I began updating our root comment queries to use Apollo and to act as a template for how we could update all of our other queries and mutations.

# 👀 See

https://kickstarter.atlassian.net/wiki/spaces/EN/pages/1382613004/iOS+Apollo+GraphQL+Migration

# ✅ Acceptance criteria

- [x] There should be no regression on the behaviour of our root comments for projects.
- [x] There should be no regression on the behaviour of our root comments for updates.
- [ ] Team to review the architectural changes and raise any potential red flags for moving forward with this.


KsApi/models/graphql/adapters/CommentsEnvelope+GraphCommentsEnvelopeTests.swift
```diff
@@ -1,10 +1,18 @@
+import ApolloAPI
 @testable import KsApi
 import Prelude
 import XCTest
 
 final class CommentsEnvelope_GraphCommentsEnvelopeTests: XCTestCase {
   func testProjectComments() {
-    let envelope = CommentsEnvelope.commentsEnvelope(from: .template)
+    // FIXME: Add templates for Apollo models and update these tests.
```
justinswart, line 8: Will solve this in a subsequent PR, I'm hoping there's a way that we can code-generate these templates.

.circleci/config.yml
```diff
@@ -7,6 +7,9 @@ xcode_version: &xcode_version 12.2.0
 iphone_name: &iphone_name iPhone 8
 preload_simulator: &preload_simulator xcrun instruments -w "iPhone 8 (14.2) [" || true
 
+# Works around an issue in xcodebuild where it requires GitHub to have been added to the known_hosts in order for it to clone dependencies.
+spm_ssh_workaround: &spm_ssh_workaround for ip in $(dig @1.1.1.1 github.com +short); do ssh-keyscan github.com,$ip; ssh-keyscan $ip; done 2>/dev/null >> ~/.ssh/known_hosts
```
justinswart, line 11: This is needed for one of Apollo's dependencies which downloads an SPM module.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.circleci/config.yml
```diff
@@ -27,6 +30,9 @@ base_job: &base_job
 test_job: &test_job
   steps:
     - checkout
+    - run:
+        name: SPM SSH Workaround
+        command: *spm_ssh_workaround
```
singhhari, line 35: what does the asterisk in front of the variable suggest here?

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-Framework-iOS.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1200"
+   LastUpgradeVersion = "1220"
```
singhhari, line 3: What changed this?

KsApi/GraphAPI.swift
```diff
@@ -0,0 +1,1039 @@
+// @generated
```
singhhari, line 1: Was this entirely auto generated? Did you manually edit a lot of this to tailor to our project? Trying to get a feel for what Apollo is doing for us.

KsApi/MockService.swift
```diff
@@ -200,7 +201,8 @@
         currency: currency,
         buildVersion: buildVersion,
         deviceIdentifier: deviceIdentifier,
-        perimeterXClient: perimeterXClient
+        perimeterXClient: perimeterXClient,
+        fetchProjectResponse: nil
```
singhhari, line 205: Do we have to do this for each of the GraphQL queries/mutations?

KsApi/extensions/ApolloInterceptors.swift
```diff
@@ -0,0 +1,43 @@
+import Apollo
+import Foundation
+
+// MARK: - NetworkInterceptorProvider
+
+class NetworkInterceptorProvider: LegacyInterceptorProvider {
```
singhhari, line 6: Ah so I'm unfamiliar with Interceptors. Are all requests the ApolloClient is performing going through the `async` method below?

singhhari (commented):
works as expected. left a few comments about the general implementation.

.circleci/config.yml
justinswart: It references the definition at the top, just yaml's way of making things a little more reusable.

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-Framework-iOS.xcscheme
justinswart: Ah yeah just the Xcode version I was using I guess, I can revert but this is pretty innocuous.

KsApi/MockService.swift
justinswart: This is still our `ServiceType` interface which will work the same way, each time we add a new query yes we'd add it to this protocol, add the query etc. Behind this protocol the Apollo client will be used to make the request.

KsApi/extensions/ApolloInterceptors.swift
justinswart: Interceptors are like hooks, just an API that Apollo provides to allow us to intercept requests and modify them if necessary. We need to do this in order to inject the PerimeterX headers because they could change between requests, or at least aren't known immediately at startup.

justinswart:
> also curious about the graph-ql-schema.json file. is that something that's automatically updated as we add new queries/mutations? is there anything in particular to be aware of with that file?

This represents the entire schema and is used by Apollo to code-generate the models and API. We will also in a subsequent PR be adding a step to download this during CI. This will act as a unit test during our CI builds in case the schema changes and breaks something.

KsApi/GraphAPI.swift
justinswart: Auto-generated by Apollo during compile time, not a user-editable file.

Library/ViewModels/CommentsViewModelTests.swift
```diff
@@ -503,8 +505,10 @@ internal final class CommentsViewModelTests: TestCase {
       self.configureFooterViewWithState.assertValues([.hidden])
 
       let updatedEnvelope = CommentsEnvelope.multipleCommentTemplate
+        |> \.slug .~ nil
+        |> \.updateID .~ "1"
```
msadoon, line 509: Just curious about why this test failed if this was omitted.

msadoon (approved):
Yea, this is great, ran the app, still fetching/paginating comments on projects and updates.
Just left some implementation questions more for my own understanding.

KsApi/extensions/ApolloClient+Singleton.swift
```diff
@@ -0,0 +1,69 @@
+import Apollo
+import Foundation
+
+class GraphQL {
+  // MARK: - Properties
+
+  private var apollo: ApolloClient?
+  private(set) lazy var client: ApolloClient = {
```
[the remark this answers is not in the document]
msadoon: or vice-versa, just think `client` is enough.

KsApi/MockService.swift
justinswart: > Yea, I don't think we need this line as it's covered below when set our `Result`'s to `nil`.

Actually, adding this just suppresses an Xcode warning that "all calls to this function will call itself", it just differentiates it enough for that warning to be silenced.

KsApi/extensions/ApolloInterceptors.swift
justinswart: > Yea why do we need this, can't we just pass in additional headers through `RequestChainNetworkTransport` in the singleton?

That would only be at instantiation and we don't know when these headers could change.

KsApi/Service.swift
```diff
@@ -47,6 +48,13 @@ public struct Service: ServiceType {
 
     // Initialize PerimeterX
     PXManager.sharedInstance().start(with: Secrets.PerimeterX.appId)
+
+    // Configure GraphQL Client
+    GraphQL.shared.configure(
+      with: serverConfig.graphQLEndpointUrl,
+      headers: self.defaultHeaders,
+      additionalHeaders: { perimeterXClient.headers() }
```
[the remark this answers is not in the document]
justinswart: You can't because `self` in this case is a struct, only reference types can be captured in this way which was necessary for the implementation.

KsApi/extensions/ApolloClient+Singleton.swift
```diff
@@ -0,0 +1,69 @@
+import Apollo
+import Foundation
+
+class GraphQL {
+  // MARK: - Properties
+
+  private var apollo: ApolloClient?
```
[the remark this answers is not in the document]
justinswart: That would only be necessary if we wanted the getter to be public. Both setter and getter are private in this case.

KsApi/fragments/CommentFragment.graphql
```diff
@@ -0,0 +1,17 @@
+fragment CommentFragment on Comment {
+  author {
+    id
+    imageUrl(width: 200)
```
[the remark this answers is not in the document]
justinswart: I couldn't find a simple way of doing this unfortunately, seems we can pass query args to fragments but then we can't have them fully detached in this way.

KsApi/extensions/ApolloClient+Singleton.swift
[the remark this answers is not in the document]
justinswart: `apollo` is instantiated when we call `configure` which we do after the singleton is created. This just allows us to be able to refer to `client` without having to deal with it being optional all over the codebase.

Library/ViewModels/CommentsViewModelTests.swift
justinswart: Because if this has a slug it's seen as being a response from a project, not an update.

KsApi/extensions/ApolloInterceptors.swift
singhhari: Gotcha, that makes sense.

singhhari (approved):
lgtm

KsApi/Service.swift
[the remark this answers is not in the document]
msadoon: I just meant because `self.defaultHeaders` is referenced as the computed property in the same way as `self.pxHeaders`. Also `self.pxHeaders` is not being used anywhere, I replaced it and it seems to compile.

PR #1534, opened 2021-06-30
[NT-1970] Show Last Page of Replies

tobitech:
# 📲 What

Fetch and display a list of the last 7 replies to a comment on the Replies screen when users tap View replies from the comment screen.

# 🤔 Why

We want users to read through the latest replies on a comment from the bottom to top.

# 👀 See

# ✅ Acceptance criteria

1. Launch the app and select a project that has comments with replies (Here is one - https://staging.kickstarter.com/projects/ivanrl/elementallis/comments)
2. Tap the comments section of the project details to show the root comments on the project.
3. Scroll to a comment with replies and tap View replies. You should see the root comment at the top and the last 7 replies to that comment.
4. The replies look like the root comments except they are indented.

PR #1539, opened 2021-07-01
[NT-2090] Refactor Replies DataSource

justinswart:
# 📲 What

Refactors the datasource for replies in an attempt to simplify the solution over our approach to root comments.

**Note:** This PR is based off #1534. If that PR merges first then this should be rebased on the main branch, although it is simpler to merge this one first and then #1534 in a cascading manner.

# 🤔 Why

When we built the first part of threaded comments (root comments), we did all of the heavy lifting for inserting, retrying, empty and errored states, etc in the view model using signals. This introduces quite a bit of complexity and for replies we're trying a different approach.

# 🛠 How

Moved most of the code to the datasource with the expectation that, instead of modifying the entire array in the view model and emitting it to be reloaded, we only emit the changes that are necessary and manipulate the data in the datasource where we can also write tests for it and where it may be easier to reason about.

# ✅ Acceptance criteria

- [ ] Team to review approach and tests.
- [x] Currently the first page of replies should still load normally.


Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -5,25 +5,63 @@ import UIKit
...
+
+    let allComments = newComments + current
+
+    allComments.forEach { comment, project in
+      self.loadValue(comment, project: project)
+    }
+  }
+
+  // TODO: Implement and write tests
+  internal func showEmptyState() {
```
singhhari: Deleting this as there is no empty state for replies. If there are no replies, the `View Replies` button is hidden.

moderateepheezy (approved):

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift:83:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift:121:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -36,4 +62,152 @@ internal final class CommentRepliesDataSource: ValueCellDataSource {
       assertionFailure("Unrecognized combo: \(cell), \(value).")
     }
   }
+
+  /// Determines if a `Comment` object is present in the data source at a given `IndexPath`.
```
moderateepheezy, line 66: Maybe should be 
`///get a comment at indexpath if present`.
Or something similar.

singhhari: Updated.

singhhari (approved):
Added documentation, tests and cleaned up some unused code.

PR #1540, opened 2021-07-01
[NT-1944] UI - User badges

moderateepheezy:
# 📲 What

Used the same author badge style for `Collaborator`s and `Creator`s.

# 🤔 Why

Every author of a comment should have a badge except if the author of the comment is a backer. However, we are currently not showing a badge for collaborators. `Creator`s and `Collaborator`s are more like, we want to also add a badge to comments from a collaborator, the style of the badge for a `Collaborator` should be the same as the one for a `Creator`.

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ]  Users can clearly identify posts by creators/collaborators, superbackers, and their own

For Collaborator, test with project https://staging.kickstarter.com/projects/paulomcg/the-project


Library/Styles/CommentCellStyles.swift
```diff
@@ -27,6 +27,18 @@ public let commentBodyTextViewStyle: TextViewStyle = { textView in
     |> \.dataDetectorTypes .~ .link
 }
 
+// Internationalize
+
+public let collaboratorAuthorBadgeStyle: PaddingLabelStyle = { label in
```
singhhari, line 32: Can we pass the text as an argument and eliminate the need for a separate style here?

singhhari (changes requested):
working as expected. left a cpl of small comments.

Library/Styles/CommentCellStyles.swift
```diff
@@ -27,6 +27,18 @@ public let commentBodyTextViewStyle: TextViewStyle = { textView in
...
+  label
+    |> \.text .~ localizedString(key: "Collaborator", defaultValue: "Collaborator")
```
msadoon, line 34: Do we not have internationalized `Strings.` for this? If not, please add a `TODO: Internationalize`

msadoon (approved):
good to go!
One small thing, in `CommentCellHeaderStackViewTests` we might in the future want to test all localized strings for these badges to ensure there are no cut-offs.

Library/Styles/CommentCellStyles.swift
moderateepheezy: No, I couldn't fine one.

singhhari (approved):
lgtm

PR #1541, opened 2021-07-05
[NT-2094] Fix Root Comments Regressions

justinswart:
# 📲 What

Fixes two regressions introduced in #1538:

- Pull-to-refresh was clearing posted comments.
- Comments weren't showing the correct author badge.

# 🤔 Why

- The Apollo client uses in-memory caching by default, when we pulled to refresh it was not making a new request to the back-end to fetch the first page again but instead reading from cache.
- In our initial implementation at the point of deserialization we were decomposing the author ID, this needed to also happen for the new `CommentFragment`.

# 🛠 How

- Disabled all caching on Apollo requests for now until we decide how we want to use this.
- Added decomposition for the author ID.

# ✅ Acceptance criteria

- [x] Post a root comment, pull to refresh, the comment should not be cleared.
- [x] All `You` author badges should be shown correctly.


msadoon (approved):
good to go!

PR #1542, opened 2021-07-05
[NT-2071] Add Backing Query Fragments

justinswart:
# 📲 What

Adds the fragments that we will need to fetch a `Backing` for the Manage Pledge View.

**Note:** Ignore changes to `GraphAPI.swift` as this is generated by Apollo.

# 🤔 Why

This is part of our migration to Apollo, we will add the implementation in a subsequent PR.

# 🛠 How

- Took the query created by the `managePledgeViewProjectBackingQuery(withBackingId:)` function and broke it up into several fragments.
- Ran the code generation on these fragments to update `GraphAPI`.
- Updated the code generation script to use `GraphAPI` as a namespace to disambiguate some type names.

# ✅ Acceptance criteria

- [ ] Code review only, no changes to functionality.

### DIFF
--- a/KsApi/fragments/BackingFragment.graphql
+++ b/KsApi/fragments/BackingFragment.graphql
@@ -1,0 +1,32 @@
+fragment BackingFragment on Backing {
+  amount {
+    ...MoneyFragment
+  }
+  backer {
+    ...UserFragment
+  }
+  backerCompleted
+  bonusAmount {
+    ...MoneyFragment
+  }
+  cancelable
+  creditCard: paymentSource {
+    ...CreditCardFragment
+  }
+  id
+  location {
+    ...LocationFragment
+  }
+  pledgedOn
+  project {
+    ...ProjectFragment
+  }
+  reward {
+    ...RewardFragment
+  }
+  sequence
+  shippingAmount {
+    ...MoneyFragment
+  }
+  status
+}
--- a/KsApi/fragments/CommentFragment.graphql
+++ b/KsApi/fragments/CommentFragment.graphql
@@ -1,9 +1,6 @@
 fragment CommentFragment on Comment {
   author {
-    id
-    imageUrl(width: 200)
-    isCreator
-    name
+    ...UserFragment
   }
   authorBadges
   body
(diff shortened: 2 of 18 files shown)


KsApi/fragments/UserFragment.graphql
```diff
@@ -0,0 +1,7 @@
+fragment UserFragment on User {
```
msadoon, line 1: Maybe we can make a small image size fragment called `UserCommentFragment` with `200` as the image width. Just because it'll be a smaller download size for our comments cells to use,  maybe even in `Backing`,  it could in theory help performance for repeatedly used ui components like table view cells.

KsApi/fragments/CommentFragment.graphql
```diff
@@ -1,9 +1,6 @@
 fragment CommentFragment on Comment {
   author {
-    id
-    imageUrl(width: 200)
-    isCreator
-    name
+    ...UserFragment
```
msadoon, line 3: Or maybe, instead of creating a new fragment we want to leave this as-is because it's using a different image size from `UserFragment`?

KsApi/fragments/BackingFragment.graphql
```diff
@@ -0,0 +1,32 @@
...
+  }
+  backer {
+    ...UserFragment
+  }
+  backerCompleted
+  bonusAmount {
+    ...MoneyFragment
+  }
+  cancelable
+  creditCard: paymentSource {
```
msadoon, line 13: is `paymentSource` something baked into the `GraphAPI` when the script runs?

msadoon (approved):
Good to go!

singhhari (approved):
lgtm

KsApi/fragments/UserFragment.graphql
justinswart: I'll look at doing this in a subsequent PR if it makes sense 👍

KsApi/fragments/BackingFragment.graphql
justinswart: The property is called `paymentSource` in the schema, we do this to alias it as `creditCard`.

PR #1543, opened 2021-07-05
[NT-1941] Post A Reply

msadoon:
# 📲 What

Adds the post reply to reply comments thread as per this [ticket](https://kickstarter.atlassian.net/browse/NT-1941).

# 🤔 Why

User should be able to post a reply to a comment of their project (assuming they are logged in backers, creators or collaborators)

# 🛠 How

- Reused the `postCommentProducer` in `CommentsViewModel`, made it `static` in an extension of `CommentsViewModel`
- Used the new `replace` function of the `CommentsRepliesDataSource` which is an easier interface than the data handling going in `CommentsViewModel`.
- Noticed that only `GraphQL.shared.client` were used on `fetchProjectComments` and `fetchUpdateComments`, not in `postComment`, so I didn't get into updating the request using the new `GraphAPI`.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [x] Users can post a reply to a comment
- [x] Users can view their reply after they post it
- [x] View scrolls to the bottom of the replies, showing the newly posted reply.


Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -98,6 +107,22 @@ final class CommentRepliesViewController: UITableViewController {
         self.tableView.reloadData()
       }
 
+    self.viewModel.outputs.loadFailableReplyIntoDataSource
+      .observeForUI()
+      .observeValues { failableComment, replaceableCommentId, project in
```
justinswart, line 112: By not using `[weak self]` here this will cause a retain cycle. I see we also missed this for `loadRepliesAndProjectIntoDataSource` so please add to both.

The purpose of the `IndexPath` being returned from the `replace` function is to allow us to only reload that row which makes for a better experience. I think we want this to look more like:

```swift
self.viewModel.outputs.loadFailableReplyIntoDataSource
  .observeForUI()
  .observeValues { [weak self] failableComment, replaceableCommentId, project in
    self.tableView.beginUpdates()

    let (indexPath, newComment) = self.dataSource.replace(
      comment: failableComment,
      and: project,
      byCommentId: replaceableCommentId
    )

    if let lastIndexPath = indexPath {
      if newComment {
        self.tableView.insertRows(at: [lastIndexPath], with: .fade)
        self.tableView.scrollToRow(at: lastIndexPath, at: .bottom, animated: true)
      } else {
        self.tableView.reloadRows(at: [lastIndexPath], with: .fade)
      }
    }

    self.tableView.endUpdates()
  }
```

justinswart (commented):
Bit difficult to review at the moment, something seems to be up with codecov, not sure why it's flagging our test files as not having coverage 🤔

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
msadoon: Got it! Yes I thought I was missing something here regarding insertions/reloading.

msadoon: Thanks for pointing this out!

msadoon: I used `performBatchUpdates` instead as the `UITableView` says `beginUpdates` `endUpdates` will be deprecated soon in favour of `performBatchUpdates`.  The reloading works as it should for both row insert and row reload. 

The scrolling to the bottom however needs a delay, because it cannot be done inside the `performBatchUpdates` or `beginUpdates` blocks. The error is an index out of bounds issue that states "Attempting to scroll to row 8 when only 7 rows in section". The `insertRow` is still continuing to finish when `scrollToRow` is called which causes the crash. It's not an animation issue either. Open to suggestions to avoid the delay, but it seems acceptable.

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -81,23 +84,60 @@ final class CommentRepliesViewController: UITableViewController {
...
+          byCommentId: replaceableCommentId
+        )
+
+        guard let lastIndexPath = indexPath else { return }
+
+        let rowUpdate: Void = newComment ?
+          self.tableView.insertRows(at: [lastIndexPath], with: .fade) :
+          self.tableView.reloadRows(at: [lastIndexPath], with: .fade)
+
+        self.tableView.performBatchUpdates {
```
msadoon, line 129: [Reference comment](https://github.com/kickstarter/ios-oss/pull/1543#discussion_r664775240)

singhhari (approved):
lgtm

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -81,23 +84,60 @@ final class CommentRepliesViewController: UITableViewController {
...
+          rowUpdate
+        } completion: { isComplete in
+          DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 0.5) {
```
justinswart, line 132: Seems strange to need to use an arbitrary delay here, does this not work?

This also checks the value of `isComplete` before trying to dispatch to the thread.

```swift
if isComplete, newComment {
  DispatchQueue.main.async {
    self.tableView.scrollToRow(at: lastIndexPath, at: .bottom, animated: true)
  }
}
```

msadoon: Yea I did try that, didn't work. My suspicion is that `isComplete` is updated as soon as the `rowUpdate` is started but whatever `rowUpdate` does is done in concurrently, so nothing is waiting for `rowUpdate` to complete...
The real issue is there is no completionHandler on `insertRows`...
I'll try it without the delay again though..

msadoon: Ok, tried again, it does work by not crashing...but it doesn't always scroll to the very bottom of the scrollview to show the new cell..sometimes we get this..

justinswart: Ah interesting....... seems like a run loop thing, perhaps just delaying it by `0.01` will even be enough, `0.5` just seems quite long?

msadoon: I _think_ I can get this to work without an arbitrary delay using the `UITableView`'s `didDisplayCell` delegate method that calls an input into the `viewModel`. Using `takePairWhen` when new output to `loadFailableComment` is triggered I can call the `scrollToRow` in the tableview. It shouldn't affect any other actions when cells appear because of `takePairWhen`.

justinswart: I don't think that will work if the cell is outside of the visible area.

msadoon: Ok yea, I just realized. To answer your original question, we can use `0.01` but it results in the situation above when the scrollview just scrolls sort of close to the bottom. I tried it at `0.3` and sometimes it scrolls close to the bottom, other times all the way.

justinswart (approved):

PR #1544, opened 2021-07-06
[NT-2054] Add pagination to comment replies

singhhari:
# 📲 What

Part of a smooth user experience for most table or collection views involves not attempting to render a massive sequence of data all at once. This could lead to issues when working with particularly large data sets; luckily, a commonly implemented solution for this exact situation is pagination. This PR will add pagination support to our reply threads, which are currently pulling a maximum of seven replies per request.

# 🤔 Why

Although this doesn't occur frequently there are still situations where a thread may have hundreds to thousands of replies. Rendering this all at the same time may lead to performance issues so we're implementing a fairly common solution in paginating the response. A new cell has been introduced that, when tapped, will fetch a maximum of seven more comments each time. 

# 🛠 How

Our GraphQL API supports cursor-based pagination. What this involves is passing an argument of a cursor (String) to inform the API that we are incrementally moving forward (or backward) in the sequence of data. We concatenate the data locally so that we are showing the older responses at the top and the newer closer to the bottom of the scroll view.

# 👀 See

https://user-images.githubusercontent.com/15615702/124636647-c777c100-de56-11eb-9de9-67ec864c964a.mp4

# ✅ Acceptance criteria

- [ ] Find a project with a particularly long thread of replies in the comment section. A good example for testing on staging is [Elemtentalis](https://staging.kickstarter.com/projects/ivanrl/elementallis?ref=nav_search&result=popular&term=Elementalis).
- [ ] Test selecting `View More Replies` to paging through the response. The latest reply should be at the bottom and paging should take us further back in time.

# ⏰ TODO

- [ ] Adding pagination support for `Updates` in a separate ticket.


Kickstarter-iOS/Views/Cells/ViewMoreRepliesCell.swift
```diff
@@ -0,0 +1,75 @@
...
+  // MARK: - View model
+
+  override func bindViewModel() {
+    super.bindViewModel()
+  }
+
+  // MARK: - Configuration
+
+  func configureWith(value _: Void) {
+    self.viewModel.inputs.configureWith()
```
singhhari, line 64: No real use for this here but the protocol requires it.

ViewMoreRepliesCellViewModelTests.swift
```diff
@@ -0,0 +1,13 @@
+@testable import KsApi
+@testable import Library
+import Prelude
+import ReactiveExtensions_TestHelpers
+import XCTest
+
+internal final class ViewMoreRepliesCellViewModelTests: TestCase {
```
singhhari, line 7: Is this empty view model fine given it doesn't require any real functionality other than being tapped?

Library/ViewModels/ViewMoreRepliesCellViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+  var inputs: ViewMoreRepliesCellViewModelInputs { get }
+  var outputs: ViewMoreRepliesCellViewModelOutputs { get }
+}
+
+public final class ViewMoreRepliesCellViewModel: ViewMoreRepliesCellViewModelType,
+  ViewMoreRepliesCellViewModelInputs, ViewMoreRepliesCellViewModelOutputs {
+  public init() {}
+
+  fileprivate let configureWithProperty = MutableProperty(())
+  public func configureWith() {
```
singhhari, line 23: Not doing anything with this here but added it in to the body of the `configureWith` method the `ValueCell` protocol requires for conformance.

Kickstarter-iOS/DataSources/CommentRepliesDataSourceTests.swift
```diff
@@ -248,15 +263,17 @@ class CommentRepliesDataSourceTests: XCTestCase {
...
-    self.dataSource.load(comments: firstPage, project: .template)
+    let totalCount = 10
+
+    self.dataSource.load(repliesAndTotalCount: (firstPage, totalCount), project: .template)
 
     let firstPageComments = Array(0..<firstPage.count).compactMap { index -> Comment? in
       self.dataSource.comment(at: IndexPath(row: index, section: self.repliesSection))
     }
 
     XCTAssertEqual(firstPageComments, firstPage)
```
msadoon, line 274: Can test the `datasource` has `Section.viewMoreReplies.rawValue` here.

msadoon: below this `assert` statement

Kickstarter-iOS/Views/Cells/ViewMoreRepliesCell.swift
```diff
@@ -0,0 +1,75 @@
...
+        top: Styles.grid(1),
+        left: Styles.grid(CommentCellStyles.Layout.leftIndentWidth),
+        bottom: Styles.grid(1),
+        right: Styles.grid(1)
+      )
+
+    _ = self.titleTextLabel
+      |> \.font .~ UIFont.ksr_callout()
+      |> UILabel.lens.textColor .~ .ksr_create_700
+      |> UILabel.lens.text .~ localizedString(key: "View more replies", defaultValue: "View more replies")
```
msadoon, line 52: Maybe add a  `TODO://` to later add internationalized `Strings` here.

Library/Styles/CommentCellStyles.swift
```diff
@@ -2,6 +2,12 @@ import Prelude
 import Prelude_UIKit
 import UIKit
 
+public enum CommentCellStyles {
```
msadoon, line 5: Glad we moved this here.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -14,7 +14,10 @@ public protocol CommentRepliesViewModelInputs {
    **/
   func configureWith(comment: Comment, project: Project, inputAreaBecomeFirstResponder: Bool)
 
-  /// Call when the view appears.
+  /// Call when a `ViewMoreRepliesCell` on the `CommentRepliesViewController` is selected.
+  func viewMoreRepliesCellWasTapped()
```
msadoon, line 18: Tiny thing, I don't think we need to specify that it's a `Cell`. Just because it's more like a user action, `viewMoreRepliesTapped`.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -94,18 +97,35 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
+      cursorFromEnvelope: { envelope in
+        (envelope.comment, envelope.cursor)
+      },
+      requestFromParams: { comment in
+        AppEnvironment.current.apiService
+          .fetchCommentReplies(query: commentRepliesQuery(withCommentId: comment.id))
+      },
+      requestFromCursor: {
+        AppEnvironment.current.apiService
+          .fetchCommentReplies(query: commentRepliesQuery(withCommentId: $0.0.id, before: $0.1))
```
msadoon, line 120: Maybe can make this a little clearer to read like in `requestFromParams`. `commentAndCursor.comment.id` and `commentAndCursor.cursor`

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -94,18 +97,35 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
+      },
+      // only return new pages, we'll concat them ourselves
+      concater: { _, value in value }
```
msadoon, line 123: So I guess `totalCountProperty.value = value` wouldn't work here?

Library/ViewModels/CommentRepliesViewModelTests.swift
```diff
@@ -28,8 +29,14 @@ internal final class CommentRepliesViewModelTests: TestCase {
...
       .observe(self.loadRepliesAndProjectIntoDataSourceProject.observer)
-    self.vm.outputs.loadRepliesAndProjectIntoDataSource.map(first)
+    self.vm.outputs.loadRepliesAndProjectIntoDataSource
+      .map(first)
+      .map { $0.0 }
       .observe(self.loadRepliesAndProjectIntoDataSourceReplies.observer)
+    self.vm.outputs.loadRepliesAndProjectIntoDataSource
+      .map(first)
+      .map { $0.1 }
+      .observe(self.loadRepliesAndProjectIntoDataSourceTotalCount.observer)
```
msadoon, line 39: Small thing, but could replace the `$0` with `repliesAndProject` to improve readability.

Library/ViewModels/CommentRepliesViewModelTests.swift
```diff
@@ -208,13 +215,17 @@ internal final class CommentRepliesViewModelTests: TestCase {
...
       self.loadRepliesAndProjectIntoDataSourceProject.assertValues([])
       self.loadRepliesAndProjectIntoDataSourceReplies.assertValues([])
+      self.loadRepliesAndProjectIntoDataSourceTotalCount.assertValues([])
 
-      self.vm.inputs.viewDidLoad()
+      self.scheduler.advance()
 
       self.loadRepliesAndProjectIntoDataSourceProject.assertValues([project])
       self.loadRepliesAndProjectIntoDataSourceReplies.assertValues([envelope.replies])
+      self.loadRepliesAndProjectIntoDataSourceTotalCount.assertValues([envelope.totalCount])
```
msadoon, line 228: Maybe update test name to reflect the `totalCount` being checked.

msadoon: If it's not too much trouble, maybe we could test pagination here as well?

There are two tests in `CommentsViewModel` that do this pretty thoroughly.
`testUpdatePagination_WhenLimitReached_CommentsAreLoadedIntoDataSource`
`testProjectPagination_WhenLimitReached_CommentsAreUpdatedInDataSource`

If you copy paste them in here, and just check the relevant stuff like the `replies`, `total` and `project` would give us further confidence in this piece of the feature.

Kickstarter-iOS/Views/Cells/ViewMoreRepliesCell.swift
```diff
@@ -0,0 +1,75 @@
...
+  // MARK: - Properties
+
+  private lazy var rootStackView = {
+    UIStackView(frame: .zero)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  private let titleTextLabel = UILabel(frame: .zero)
+
+  private let viewModel: ViewMoreRepliesCellViewModelType = ViewMoreRepliesCellViewModel()
```
msadoon, line 15: Can remove, unneeded.

Library/ViewModels/ViewMoreRepliesCellViewModel.swift
msadoon: Actually I don't think we need this entire file. It's never used.

ViewMoreRepliesCellViewModelTests.swift
msadoon: Can remove this entire test file, we're not testing anything.

msadoon (changes requested):
Great job, works like a charm!
I left some minor comments below mostly around unneeded files and 1/2 tests that can be added.

msadoon:
One thing I just thought of is loading the `CommentRepliesViewController` requires the `Comment`'s to be fetched, which requires the `Project`. Now even if you access the `CommentRepliesViewController` from the `Update`, you're still fetching the `Project` for free before you `CommentRepliesViewController` so it doesn't need to be refactored for `Update`.

ViewMoreRepliesCellViewModelTests.swift
singhhari: Yeah, good point. Deleting.

Library/ViewModels/ViewMoreRepliesCellViewModel.swift
singhhari: Deleting this too.

Kickstarter-iOS/Views/Cells/ViewMoreRepliesCell.swift
singhhari: Ah you might have missed it a few lines above. It's on line 30 above the method name,

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: Going to leave it as is because this input is called, if and only if, the `ViewMoreRepliesCell` is selected.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: Good catch.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: We decided to set it in `valuesFromEnvelope` because it only needs to be set once and we don't want the signal to emit any more times than necessary (the value of totalCount never changes).

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -30,14 +34,23 @@ internal final class CommentRepliesDataSource: ValueCellDataSource {
 
     self.clearValues(section: Section.replies.rawValue)
 
-    let newComments = comments.map { comment in
-      (comment, project)
+    let newReplies = replies.map { [project] reply in
```
justinswart, line 37: Not sure we need the capture list here for `project` because this isn't an async closure.

justinswart (approved):
Just the one comment, lgtm.

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
singhhari: Updated.

Library/ViewModels/CommentRepliesViewModelTests.swift
singhhari: We now handle the pagination in the data source as opposed to the view model so I don't think I can look at the previous view model tests to test this. It's proving to be a bit tricky but I'll take a look at it for a bit longer.

singhhari: Extended this test to include tests for pagination.

msadoon (approved):
Good to go!

PR #1545, opened 2021-07-07
[NT-1974] Show Failed Comment Reply

msadoon:
# 📲 What

When posting a reply fails, we need to indicate that state to the user, because the optimistic comment is displayed successfully. [Ticket](https://kickstarter.atlassian.net/browse/NT-1974).

# 🤔 Why

Just following the same product design for main thread comments now for replies.

# 🛠 How

- Added the view model output `shouldIndentContent` to the `CommentPostFailedCell` which allows the `CommentCellViewModel` to update it with `isReply` signal like it does for `CommentCell`

Why are there so many replaced screenshot tests?
- One small caveat with using `shouldIndentContent` on the failed cell, is that we need to update our `CommentsViewControllerTests` due to the fact that [this](https://github.com/kickstarter/ios-oss/compare/NT-1974/show-failed-reply?expand=1#diff-2de3edb02d32cd559393389d3f5a457523c431892cf31f58d9ca0da4c8d49410R96) adjustment is made to the `rootStackView` even when `isReply` is `false`. That adjustment to `top` being `1` instead of `3` in the `commentCellRootStackViewStyle` means tests need to be re-recorded. I also added iPad tests for what was previously. I manually checked the `en` versions of the screens before/after recording, no other changes stood out except for above mentioned.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| n/a |  |

**Note**
So you might notice after replacing the optimistic comment with the failed state, the retry bottom is cutoff because the scroll view doesn't scroll all the way down. This is because we want to scroll for the optimistic comment only. If the comment fails to post at some time in the future, we don't want to then again scroll the user to the bottom.

# ✅ Acceptance criteria

- [x] If a posted comment fails, the comment state is updated to failed. Can test with interrupting a post request on Charles.


Kickstarter-iOS/Views/Cells/CommentPostFailedCell.swift
```diff
@@ -86,6 +86,21 @@ final class CommentPostFailedCell: UITableViewCell, ValueCell {
         _ = self.retryButton
           |> status == .retrying ? postingButtonStyle : retryButtonStyle
       }
+
+    self.viewModel.outputs.shouldIndentContent
+      .observeForUI()
+      .observeValues { shouldIndent in
+        _ = self.rootStackView
+          |> \.axis .~ .vertical
```
singhhari, line 94: Should we save this as a style constant somewhere globally and use it both here and in the `CommentCell`/wherever else we use `shouldIntentContent`?

KsApi/models/templates/CommentTemplates.swift
```diff
@@ -46,10 +46,27 @@ extension Comment {
     id: "89DJa89jdSDJ89sd8==",
     isDeleted: false,
     parentId: "Q29tbWVudC0zMjY2NDAxMg==",
-    replyCount: 2,
+    replyCount: 0,
```
singhhari, line 49: Ah good catch. Replies always return a 0 for this count, right?

KsApi/models/templates/CommentRepliesEnvelopeTemplates.swift
```diff
@@ -17,11 +17,11 @@ extension CommentRepliesEnvelope {
...
   )
 
-  static let singleReplyTemplate = CommentRepliesEnvelope(
+  static let failedAndSuccessRepliesTemplate = CommentRepliesEnvelope(
     comment: .template,
     cursor: "cursor",
     hasPreviousPage: false,
-    replies: [.replyTemplate],
-    totalCount: 1
+    replies: [.replyFailedTemplate, .replyTemplate],
```
singhhari, line 24: Not sure if I'm overthinking this but looking at the screenshots I'm wondering if the order of these two should be swapped? Is there a situation where a failed reply is above a successful one in a given thread?

singhhari: Ah well it's certainly possible if a user simply does not retry the failed one and then manages to send a successful one.

singhhari (commented):
Working as expected! Left a couple of comments.

KsApi/models/templates/CommentTemplates.swift
msadoon: Yup, none of the tests broke either after the change, which is good because that would indicate we were checking replies on replies.

KsApi/models/templates/CommentRepliesEnvelopeTemplates.swift
msadoon: Yea, it's unlikely, because it's easier to a failed comment before retyping a new one. There could be a situation where optimistically the comment is posted, the user posts another because they thing the first one was successful, then the first comment shows failed.

Kickstarter-iOS/Views/Cells/CommentPostFailedCell.swift
msadoon: yea that would be ideal. Right now though I only see it being used for threaded replies, but if another place in the app is using this kind of indentation we can more it into `CommentCellStyles`.

singhhari (approved):
lgtm

KsApi/models/templates/CommentRepliesEnvelopeTemplates.swift
msadoon: Good call making it a common style, it fixes that `top` padding of `1` when the cell was not indented (above).

justinswart (approved):

PR #1549, opened 2021-07-08
[NT-1995] Handle pagination failure on comment reply threads

singhhari:
# 📲 What

Part of supporting pagination on our comment threads is handling the failure case. Specifically, when selecting the `View More Replies` cell, we want to show the user when the pagination fails and offer them a way to retry the request. In this PR, we add a new cell to handle the pagination failure and retry the request when it is selected in the tableview.

# 🤔 Why

Error cases need to be handled to support a smooth user experience. If anything goes wrong and an error is thrown, we should also enable the user to retry an errant request to continue where they left off.

# 🛠 How

- Add a new section and cell in the data source of our `CommentRepliesViewController`.
- Handling the error signal from the `paginate` method in the `CommentRepliesViewModel` to return an empty array.
- Update the `load` method of the `CommentRepliesDataSource` to handle the empty array case.

# 👀 See

https://user-images.githubusercontent.com/15615702/124970315-c630ca80-dff5-11eb-80d2-af3c242d64f5.mp4

# ✅ Acceptance criteria

- [ ] On staging, head to a project with a testable comment thread like [Elementalis](https://staging.kickstarter.com/projects/ivanrl/elementallis?ref=nav_search&result=popular&term=Elementalis)
- [ ] Once in the thread, selec `View More Replies` while you have a breakpoint in on Charles and abort the request.
- [ ] You should now see the red failure cell instead of the `View More Replies` cell.
- [ ] Tap the cell to retry and this time successfully execute the request.
- [ ] Verify the pagination works and there is no side effect.


Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -54,29 +62,20 @@ internal final class CommentRepliesDataSource: ValueCellDataSource {
...
-      cell.configureWith(value: value)
     case let (cell as CommentCell, value as (Comment, Project)):
       cell.configureWith(value: value)
-    case let (cell as ViewMoreRepliesCell, _):
-      cell.configureWith(value: ())
     case let (cell as CommentPostFailedCell, value as Comment):
       cell.configureWith(value: value)
     case let (cell as CommentRemovedCell, value as Comment):
       cell.configureWith(value: value)
+    case let (cell as CommentViewMoreRepliesFailedCell, _):
```
singhhari, line 73: Reordering alphabetically here.

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -234,7 +233,28 @@ internal final class CommentRepliesDataSource: ValueCellDataSource {
...
+  }
+
+  /**
+   Returns `true` when  the `IndexPath`provided is from `Section.viewMoreReplies` or `Section.viewMoreRepliesError`.
+
+   - parameter indexPath: `IndexPath` object that we need the `Section` value from.
+
+   - returns: A  `Bool` which is only true is the section is either `viewMoreReplies` or `viewMoreRepliesError`.
+   */
+  public func sectionFor(_ indexPath: IndexPath) -> Bool {
```
singhhari, line 252: Replaced the `isCellInViewMoreRepliesSection` method with this one because we are now supporting two different types of cells in `didSelectRowAt`.

Kickstarter-iOS/Views/Cells/CommentViewMoreRepliesFailedCell.swift
```diff
@@ -0,0 +1,102 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+final class CommentViewMoreRepliesFailedCell: UITableViewCell, ValueCell {
```
singhhari, line 6: There's no view model here because the cell is dumb and only needs to be rendered/tapped.

Kickstarter-iOS/Views/Cells/CommentViewMoreRepliesFailedCell.swift
```diff
@@ -0,0 +1,102 @@
...
+    _ = self.retryImageViewStackView
+      |> \.axis .~ .vertical
+
+    _ = self.rootStackView
+      |> rootStackViewStyle
+  }
+
+  // MARK: - Configuration
+
+  internal func configureWith(value _: Void) {
```
singhhari, line 58: Needed this for protocol conformance but it's not doing anything.

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -14,7 +14,7 @@ final class CommentRepliesViewController: UITableViewController {
   // MARK: Properties
 
   private let dataSource = CommentRepliesDataSource()
-  private let viewModel: CommentRepliesViewModelType = CommentRepliesViewModel()
+  internal let viewModel: CommentRepliesViewModelType = CommentRepliesViewModel()
```
singhhari, line 17: Made this internal (similar to the `CommentsViewController`) to enable testing the failed pagination in the view controller test file.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -25,6 +22,9 @@ public protocol CommentRepliesViewModelInputs {
 
   /// Call when the view loads.
   func viewDidLoad()
+
+  /// Call in `didSelectRow` when either a `ViewMoreRepliesCell` or `CommentViewMoreRepliesFailedCell` is tapped.
+  func viewMoreRepliesOrViewMoreRepliesFailedCellWasTapped()
```
singhhari, line 27: This is definitely a long function name but I wanted it to be obvious what was going on here.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -162,9 +162,21 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
       )
 
-    let repliesAndTotalCount = replies.withLatestFrom(totalCountProperty.signal)
+    let errorLoadingMoreReplies =
+      error
+        .withLatestFrom(totalCountProperty.signal)
+        .combineLatest(with: project)
+        .map { errorAndTotalCount, project -> (([Comment], Int), Project) in
+          let (_, totalCount) = errorAndTotalCount
+          return (([], totalCount), project)
```
singhhari, line 171: Empty array, similar to what we're doing in `CommentsViewModel` for the same situation.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -162,9 +162,21 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
+        }
 
-    self.loadRepliesAndProjectIntoDataSource = Signal.combineLatest(repliesAndTotalCount, project)
+    let repliesAndTotalCount = replies.withLatestFrom(totalCountProperty.signal).combineLatest(with: project)
+
+    self.loadRepliesAndProjectIntoDataSource = Signal.merge(
+      repliesAndTotalCount,
+      errorLoadingMoreReplies
```
msadoon, line 178: `repliesTotalCountAndProject`
`erroredRepliesTotalCountAndProject`

Library/ViewModels/CommentRepliesViewModelTests.swift
```diff
@@ -263,6 +263,70 @@ internal final class CommentRepliesViewModelTests: TestCase {
...
+
+          self.loadRepliesAndProjectIntoDataSourceProject.assertValues([project, project, project])
+          self.loadRepliesAndProjectIntoDataSourceReplies
+            .assertValues([envelope.replies, [], updatedEnvelope.replies])
+          self.loadRepliesAndProjectIntoDataSourceTotalCount
+            .assertValues([envelope.totalCount, envelope.totalCount, updatedEnvelope.totalCount])
+        }
+      }
+    }
+  }
```
msadoon, line 328: great test!

msadoon (changes requested):
Looks pretty good, just found an issue with the logic around showing the error state to prevent this issue:

Otherwise some minor comments. I foresee some conflicts with the retrying PR, but you can merge yours first, it's not going to be too much work to fix up the issues in mine. 👍🏿

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
singhhari: The Section is internal to the data source so I didn't want to expose any of that in the view controller. Let me know what you think, I can change if it makes it easier for subsequent PRs.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: Updated.

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
```diff
@@ -44,6 +46,12 @@ internal final class CommentRepliesDataSource: ValueCellDataSource {
       self.loadValue(reply, project: project)
     }
 
+    // If newReplies.count == 0 there was an issue loading more replies and we should show the error cell.
+    guard newReplies.count > 0 else {
```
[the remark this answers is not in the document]
singhhari: hmm wondering if I should just apply some optionality to the array of comments instead?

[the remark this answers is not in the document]
justinswart: I think just emit specifically from the datasource and call `showErrorState()` directly. The VM has more context of whether or not this is in fact errored.

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
justinswart: Best to do as much switching logic in the datasource as possible because we can cover that with tests.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -162,9 +162,21 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
         CommentsViewModel.postCommentProducer
       )
 
-    let repliesAndTotalCount = replies.withLatestFrom(totalCountProperty.signal)
+    let errorLoadingMoreReplies =
+      error
```
justinswart, line 166: Can move this to the previous line for readability.

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -162,9 +162,21 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
+        .withLatestFrom(totalCountProperty.signal)
+        .combineLatest(with: project)
+        .map { errorAndTotalCount, project -> (([Comment], Int), Project) in
```
justinswart, line 169: If we go with the other suggestion to call that function directly on the datasource then we shouldn't need this I think?

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
singhhari: Cool, going to keep this as is.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: This doesn't work. Going to change it.

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -149,9 +150,8 @@ final class CommentRepliesViewController: UITableViewController {
 
 extension CommentRepliesViewController {
   override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
-    if self.dataSource.isCellInViewMoreRepliesSection(indexPath) {
-      self.viewModel.inputs.viewMoreRepliesCellWasTapped()
-    }
+    guard self.dataSource.sectionFor(indexPath) else { return }
```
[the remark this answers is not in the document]
singhhari: Going to leave it in the data source since that's the cleaner paradigm for testability, as justin mentioned in another thread here.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: Ended up deleting both of these constants in the re-factor.

Library/ViewModels/CommentRepliesViewModel.swift
singhhari: Yup I re-factored to have a separate output handle the error. It's directly calling the showPaginationErrorState() method from the data source.

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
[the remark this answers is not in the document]
singhhari: Re-factored this in the last commit.

justinswart (approved):

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
msadoon: Could we name it `sectionForViewMoreReplies`?

msadoon (approved):
Good to go!

Kickstarter-iOS/DataSources/CommentRepliesDataSource.swift
singhhari: Thing is, it's checking viewMoreReplies and viewMoreRepliesError so I think that's a bit misleading.

PR #1552, opened 2021-07-13
[NT-1974] Interrupted Scrolling Replies Thread.

msadoon:
# 📲 What

There is an issue when posting a reply that fails, scrolling is interrupted when the post fails.

# 🤔 Why

So previously we did not actually delay the comments that were errorred using `demoteErrors`, the `ksr_delay` was being called on the network requests' return, but not on the error after it, so if the network request failed, the error posted immediately, which skipped the 1 second delay after the prefix. So likely the table view is scrolling at this point which was why it was interrupted.

# 🛠 How

Replaced `demoteErrors` with `flatMapError` and added the `ksr_delay` directly on the signal producer.

Also added a testable function to the `loadFailableCommentIntoDataSource` `output` to handle cases where we don't scroll if the comment is successful upon returned from the network. We only scroll when it's a new comment, or a posted comment fails on the network. All scrolling occurs with animation.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ⏰ TODO

- [ ] Ensure posting a reply scrolls the user to new comment without being interrupted after an error occurs or the post is successful.


Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -183,3 +179,8 @@ private let tableViewStyle: TableViewStyle = { tableView in
     |> \.rowHeight .~ UITableView.automaticDimension
     |> \.separatorStyle .~ .none
 }
+
+internal func commentRepliesRowBehaviour(for comment: Comment,
```
singhhari, line 183: I would add some documentation to this method. it's not obvious what's happening based on the method name.

singhhari (approved):
lgtm

justinswart (approved):

PR #1553, opened 2021-07-13
[NT-2073] Don't Fetch Expanded Shipping Rules For Digital-Only Rewards

justinswart:
# 📲 What

Excludes expanded shipping rules for digital-only rewards.

# 🤔 Why

When fetching add-ons, if we pass in a `nil` location ID, all possible locations are returned. This returns all possible shipping locations in the known universe and can cause the response to time out.

# 🛠 How

Added a [directive](https://graphql.org/learn/queries/#directives) to our query to make this property conditional.

# ✅ Acceptance criteria

- [x] No regression on reward add-on selection.
- [x] Digital-only add-ons load fine.

### DIFF
--- a/Library/ViewModels/RewardAddOnSelectionViewModel.swift
+++ b/Library/ViewModels/RewardAddOnSelectionViewModel.swift
@@ -97,6 +97,7 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
     let projectEvent = slugAndShippingRule.switchMap { slug, shippingRule in
       AppEnvironment.current.apiService.fetchRewardAddOnsSelectionViewRewards(
         slug: slug,
+        shippingEnabled: shippingRule?.location.graphID != nil,
         locationId: shippingRule?.location.graphID
       )
       .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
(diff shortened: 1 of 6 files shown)


Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -97,6 +97,7 @@ public final class RewardAddOnSelectionViewModel: RewardAddOnSelectionViewModelT
     let projectEvent = slugAndShippingRule.switchMap { slug, shippingRule in
       AppEnvironment.current.apiService.fetchRewardAddOnsSelectionViewRewards(
         slug: slug,
+        shippingEnabled: shippingRule?.location.graphID != nil,
```
justinswart, line 100: @singhhari are there ever legitimate situations where we would want to pass `nil` and fetch all locations?

singhhari: The only thing I can think of is the pledge view. We do need to render the entire list of available shipping locations there but I don't think there is a practical use case for ALL locations.

singhhari (approved):
Cool stuff. Didn't know about directives. 

lgtm.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
justinswart: Ah right, but that should rely on the regular shipping views that aren't fetched from the expanded list.

PR #1555, opened 2021-07-13
[NT-2104] Migrate CreateBacking To Apollo

justinswart:
# 📲 What

Migrates our existing `CreateBackingMutation` to use Apollo.

# 🤔 Why

Continued migration of queries and mutations to Apollo.

# 🛠 How

I tried to change as little as possible in `PledgeViewModel` so I've added adapters to go from our existing `CreateBackingInput` to the generated `GraphAPI.CreateBackingInput` that Apollo uses. There is also an adapter to return the expected `CreateBackingEnvelope` from Apollo's `GraphAPI.CreateBackingMutation.Data`.

# ✅ Acceptance criteria

There should be no regressions to pledging:

- [ ] No reward backing.
- [ ] Regular reward.
- [ ] Regular reward with shipping.
- [ ] Backing with add-ons.
- [ ] Variations of the above using Apple Pay.

PR #1556, opened 2021-07-14
[NT-2109] Migrate UpdateBacking to Apollo

justinswart:
# 📲 What

Migrates the `UpdateBacking` mutation to use Apollo.

**Note:** This PR is based off #1555 as that is currently blocked. Once unblocked it would be easiest to merge this first and then #1555.

# 🛠 How

Mostly a repeat of the work done in #1555, this also moves the shared `checkout` data to `CheckoutFragment` as it's used by both `CreateBacking` and `UpdateBacking`.

# ✅ Acceptance criteria

- [ ] No regressions to updating a backing in the manage pledge view.


KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -15,6 +16,9 @@ extension ApolloClient {
       self.fetch(query: query, cachePolicy: .fetchIgnoringCacheCompletely) { result in
         switch result {
         case let .success(response):
+          if let error = response.errors?.first?.errorDescription {
```
justinswart, line 19: Successful responses from GraphQL can contain validation errors, so we handle that here.

KsApi/models/graphql/adapters/CreateBackingEnvelope+CreateBackingMutation.DataTests.swift
```diff
@@ -0,0 +1,76 @@
...
+          "state": GraphAPI.CheckoutState.authorizing,
+          "backing": [
+            "clientSecret": "client-secret",
+            "requiresAction": true
+          ]
+        ]
+      ]
+    ]
+
+    let data = GraphAPI.CreateBackingMutation.Data(unsafeResultMap: dict)
```
justinswart, line 19: These tests are from #1555 but were renamed and updated to be constructed from a dictionary which made things easier for use with `CheckoutFragment`.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -482,8 +482,6 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
       .switchMap { [checkoutIdProperty] input in
         AppEnvironment.current.apiService.createBacking(input: input)
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
-          // FIXME: Remove once UpdateBacking mutation is aligned on error type.
```
justinswart, line 485: As promised, this could now be removed as both mutations share the same error type. This also means that against the main branch there are no changes to `PledgeViewModel`.

KsApi/models/graphql/adapters/CreateBackingEnvelope+CreateBackingMutation.DataTests.swift
singhhari: is the "client-secret" only returned on success?

singhhari (approved):
lgtm

KsApi/models/graphql/adapters/CreateBackingEnvelope+CreateBackingMutation.DataTests.swift
justinswart: I've seen it sometimes returned on failure too but it's only needed if `requiresAction` is `true`.

PR #1557, opened 2021-07-14
[NT-2131] Adding translated strings for native comment threading work

singhhari:
# 📲 What

We had a few strings in place that were not localized as we were continuing development on comment threading. Now that those translations are complete and deployed to production we can substitute our English-only placeholders for these.

# 🤔 Why

Kickstarters users speak a multitude of languages and it's important that we support them with the best user experience possible. Replacing the English-only placeholders will now enable our French, German, Japanese and Spanish language readers from engaging with a fully localized comment threading experience.

# 🛠 How

We use a translation service via `CrowdIn` that supports the translation of English strings into multiple others. After those are complete, our web repository deploys these updates to production which enabled us to use them on the native apps.

# ✅ Acceptance criteria

- [ ] Verify the `Collaborator` string is being localized within the context of comment threading.
- [ ] Verify the `Replies` string is being localized within the context of comment threading.
- [ ] Verify the `View more replies` string is being localized within the context of comment threading.


justinswart (approved):
lgtm!

PR #1559, opened 2021-07-15
[NT-2074] Download GraphQL Schema During CI Builds

justinswart:
# 📲 What

Adds a step to our builds on CircleCI so that the latest GraphQL schema on staging is downloaded.

# 🤔 Why

Downloading the schema ahead of the build will ensure that if there are breaking changes, the app will fail to compile and we'll be afforded the opportunity to catch those sorts of issues earlier.

# 🛠 How

Added a script that installs the Apollo CLI client and then runs it to download the schema from staging.

# ✅ Acceptance criteria

- [ ] Build step runs, app compiles.
- [ ] We're happy to download the schema from staging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ test_job: &test_job
         command: make bootstrap
     - run:
         name: Download GraphQL Schema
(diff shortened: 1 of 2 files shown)


singhhari (approved):
lgtm

PR #1561, opened 2021-07-15
[NT-2114] Adding collaboratorPermissions to ProjectFragment

singhhari:
# 📲 What

`MemberData` is currently not being fetched from GraphQL and we need the various properties of this object for the `ProjectFragment`. It doesn't seem like our GraphQL `Project` object currently supports any of the properties other than `collaboratorPermissions` so that is what we will add in this PR.

# 🤔 Why

Part of our native comment threading feature involves the use of permissions for determining whether a collaborator can comment or not. Moreover, with the migration toward Apollos GraphQL client we are updating our fragments and ensuring they are both reusable and bug free when used elsewhere.

# 🛠 How

Added the `collaboratorPermissions` field to our `ProjectFragment` and extended our adapter to take a given `CollaboratorPermission` from GraphQL and turn it into a `Project.MemberData.Permission` type.

# ✅ Acceptance criteria

- [ ] Run the app on staging and navigate to the various places we make use of the `ProjectFragment`. That includes the query for fetching add ons and the query for backing.
- [ ] Using Charles, verify the response is not erroring and there is no regression in the behavior of the queries.

# ⏰ TODO

- [ ] Inform @paulomcg that backend support will be needed to add the following properties to the GraphQL response for a `Project`:
 - `lastUpdatePublishedAt`
 - `unreadMessagesCount`
 - `unseenActivityCount`

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -2062,6 +2062,67 @@ public enum GraphAPI {
     }
   }
 
+  /// A permission for a collaborator on a project
+  public enum CollaboratorPermission: RawRepresentable, Equatable, Hashable, CaseIterable, Apollo.JSONDecodable, Apollo.JSONEncodable {
+    public typealias RawValue = String
+    case editProject
+    case editFaq
+    case post
+    case comment
+    case viewPledges
+    case fulfillment
+    /// Auto generated constant for unknown enum values
+    case __unknown(RawValue)
+
+    public init?(rawValue: RawValue) {
+      switch rawValue {
+        case "edit_project": self = .editProject
+        case "edit_faq": self = .editFaq
+        case "post": self = .post
+        case "comment": self = .comment
+        case "view_pledges": self = .viewPledges
+        case "fulfillment": self = .fulfillment
+        default: self = .__unknown(rawValue)
+      }
+    }
+
+    public var rawValue: RawValue {
+      switch self {
+        case .editProject: return "edit_project"
+        case .editFaq: return "edit_faq"
+        case .post: return "post"
+        case .comment: return "comment"
+        case .viewPledges: return "view_pledges"
+        case .fulfillment: return "fulfillment"
+        case .__unknown(let value): return value
+      }
+    }
+
+    public static func == (lhs: CollaboratorPermission, rhs: CollaboratorPermission) -> Bool {
+      switch (lhs, rhs) {
+        case (.editProject, .editProject): return true
+        case (.editFaq, .editFaq): return true
+        case (.post, .post): return true
+        case (.comment, .comment): return true
+        case (.viewPledges, .viewPledges): return true
+        case (.fulfillment, .fulfillment): return true
+        case (.__unknown(let lhsValue), .__unknown(let rhsValue)): return lhsValue == rhsValue
+        default: return false
+      }
+    }
+
+    public static var allCases: [CollaboratorPermission] {
+      return [
+        .editProject,
+        .editFaq,
+        .post,
+        .comment,
+        .viewPledges,
+        .fulfillment,
+      ]
+    }
+  }
+
   /// Various project states.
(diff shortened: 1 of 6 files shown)


KsApi/GraphAPI.swift
```diff
@@ -2062,6 +2062,67 @@ public enum GraphAPI {
     }
   }
```
singhhari, line 2064: As @justinswart has alluded to before, all of this was generated by Apollo.

KsApi/fragments/ProjectFragment.graphql
```diff
@@ -6,6 +6,7 @@ fragment ProjectFragment on Project {
   category {
     ...CategoryFragment
   }
+  collaboratorPermissions
```
singhhari, line 9: This is the only property of our `MemberData` object that is currently supported by Graph. I will raise this to Paulo and request backend support.

KsApi/models/graphql/adapters/Project+FetchAddOnsQueryDataTests.swift
```diff
@@ -9,6 +9,7 @@ final class Project_FetchAddOnsQueryDataTests: XCTestCase {
       XCTAssertNotNil(fragment)
 
       let project = Project.project(from: fragment)
+      XCTAssertEqual(project?.memberData.permissions.count, 6)
```
singhhari, line 12: Not sure if this makes sense including in here but think it's a good idea to add little tests like this any time we extend a Fragment. What do you think?

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
```diff
@@ -83,21 +84,14 @@ private func finalCollectionDateTimeInterval(
...
-    convertedPledgedAmount: nil,
-    currency: projectFragment.currency.rawValue,
-    currentCurrency: nil,
-    currentCurrencyRate: nil,
-    goal: projectFragment.goal?.fragments.moneyFragment.amount.flatMap(Int.init) ?? 0,
-    pledged: projectFragment.pledged.fragments.moneyFragment.amount.flatMap(Int.init) ?? 0,
-    staticUsdRate: projectFragment.usdExchangeRate.flatMap(Float.init) ?? 0,
-    updatesCount: nil
-  )
+private func projectMemberData(from projectFragment: GraphAPI.ProjectFragment) -> Project.MemberData? {
```
singhhari, line 89: Assuming we get some backend support to add the other three properties of `MemberData`, we would extend this method to support that.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
```diff
@@ -117,3 +111,21 @@ private func projectPhoto(from projectFragment: GraphAPI.ProjectFragment) -> Pro
 private func projectState(from projectState: GraphAPI.ProjectState) -> Project.State? {
   return Project.State(rawValue: projectState.rawValue.lowercased())
 }
+
+/**
+ Returns a minimal `Project.Stats` from a `ProjectFragment`
+ */
+private func projectStats(from projectFragment: GraphAPI.ProjectFragment) -> Project.Stats {
```
singhhari, line 118: Didn't do anything here, just alphabetizing.

justinswart (approved):
Nice all looks good.

KsApi/models/graphql/adapters/Project+FetchAddOnsQueryDataTests.swift
msadoon: Yea I did the same for `personalization`

msadoon (approved):
Good to go!
Didn't need to run the app on this one because tests confirm the addition of this parameter are supported by the schema. Otherwise the sample dictionary would fail the `do`/`catch`. Found this to be the case for `personalization`.

msadoon:
Just ran app for member permissions on Comments Threading, seems to work fine. ⚡

PR #1563, opened 2021-07-16
[NT-2115] Add Personalization To Project Fragment

msadoon:
# 📲 What

`Personalization` is a property on a `Project` that tailors UI components on project cards (ex. facebook friends' avatars , is user backing?, project being watched) and connects a `Backing` to the logged in user in several places of the app.

# 🤔 Why and How

`project.personalization` contains 4 optional properties:

`backing` - a `backing` of the project

`isStarred` - did the current user "heart"(iOS) or "like" or "save"(on web) the project, seen in project cards with "heart symbol"

`isBacking` - is the current user someone backing this project.

`friends` - if two friends that are connected via facebook back the same project, and either logs in via facebook this property will show that friend on the project card in a facepile.

**How the project fragment populates personalization**

`backing` - this is currently passed only in `ProjectAndBackingEnvelope`. The other callsite is `Project+FetchAddOnsQueryData` and it doesn't need to pass it in because whereever uses this data, is getting it's backing  via a separate fetch call not tied to the fragment.

`isStarred` - as per [this](https://kickstarter.slack.com/archives/C411RRLCQ/p1626302788322100) slack conversion with a backend dev, `isWatched` can be replaced with `isStarred` from `v1`.

`isBacking` - my assumption was that `backing` above would inform this property, because it makes sense to only store a `backing` in `personalization` if it was for the current user. Storing a `backing` that isn't the current user's would still reflect correctly here.

`friends` - this is untested on staging, as it I tried scrolling through hundreds of projects on the iOS looking for one with friends but came up with none and it is difficult to set up this scenario. My theory is you need 2 facebook friends who are signed up on staging kickstarter and back the same project. Once that's done, signing in with one of them will show the other's avatar as you find the project's card in the explore tab.

Reason it was difficult to set up this test: 

- the staging app is not setup to be tested using real facebook accounts. When trying to sign up with facebook: (ie. 

[Relevant conversion on slack with backend dev.](https://kickstarter.slack.com/archives/C411RRLCQ/p1626382700335700)

# 🛠 More Notes

Right now personalization in the project fragment is completely unused.

We rely on v1 calls for `isStarred` and `friends`, and gql (with the fragment?) for the `backing` object.

Most of the information needed to populate `personalization` comes from the fragment itself, and it is tested well enough to be confident that the schema provides the right data used by the app currently. 

Only thing that cannot be self contained is the `backing` which references `project` and might cause some kind of infinite loop when fetching `project`.

Here are all the places in the app where each personalization is currently being used:

_backing_
> AppDelegateViewModel
RewardCellProjectBackingStateType
SharedFunctions
CommentCellViewModel
ManagePledgeViewModel
PledgeCTAContainerViewViewModel
PledgePaymentMethodsViewModel
PledgeShippingLocationViewModel
PledgeViewModel
ProjectPamphletViewModel
RewardAddOnCardViewModel
RewardAddOnSelectionViewModel
RewardCardViewModel
RewardsCollectionViewModel

_isStarred_
> KSRAnalytics
WatchProjectViewModel
BackerDashboardProjectCellViewModel

_isBacking_
> PledgeTitleCell
KSRAnalytics
CommentCellViewModel
MessagesViewModel
PledgeCTAContainerViewViewModel
PledgePaymentMethodsViewModel
ProjectPamphletMainCellViewModel
ProjectPamphletViewModel
RewardCardViewModel

_friends_
> DiscoveryPostcardViewModel
DiscoveryProjectCardViewModel

All these places use a combo of `v1` and `schema` to populate `personalization`.

There might be a lot info here, but in case we start to refactor `personalization` at some point in could be useful.

# ✅ Acceptance criteria

- [x] App builds, tests pass


KsApi/fragments/ProjectFragment.graphql
```diff
@@ -17,6 +22,13 @@ fragment ProjectFragment on Project {
   description
   finalCollectionDate
   fxRate
+  friends {
+    edges {
```
justinswart, line 26: I typically only use edges if I need information from the edge, in this case we could probably just do:
```graphql
friends {
  nodes {
    ...UserFragment
  }
}
```
But same result really.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
```diff
@@ -7,7 +7,8 @@ extension Project {
   static func project(
     from projectFragment: GraphAPI.ProjectFragment,
     rewards: [Reward] = [],
-    addOns: [Reward]? = nil
+    addOns: [Reward]? = nil,
+    backing: Backing? = nil
```
justinswart, line 11: So in this case, you've made `backing` part of the `ProjectFragment`, in which case we wouldn't need to pass it into this function because it's already in the fragment. If we were to only fetch it in specific queries then we would add it as an argument here.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
```diff
@@ -48,6 +59,27 @@ extension Project {
...
+                                    backing: Backing?,
+                                    friends: [User],
+                                    projectBackingUserId: String?) -> Project.Personalization {
+  var currentUserIsBacker: Bool {
+    guard let userId = projectBackingUserId,
+      let userIdValue = Int(userId) else {
+      return false
+    }
+
+    return backing?.backerId == userIdValue
```
justinswart, line 72: Don't think we need this ID comparison, if the `Project` has a `Backing` then the current user is backing it.

justinswart (commented):
Looking good, thanks for the detail in the description, some initial comments.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
msadoon: I see, I will create a backing from the `ProjectFragment` instead of passing it in.

KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
```diff
@@ -411,6 +418,10 @@ private func backingDictionary() -> [String: Any] {
       "deadlineAt": 1620478771,
       "description": "Dark Fantasy Novel & Tarot Cards",
       "finalCollectionDate": null,
+      "friends": {
+        "__typename": "ProjectBackerFriendsConnection",
+        "edges": []
```
singhhari, line 423: I wonder if it would be a more useful test if we had friends returning in this array?  I know it's a tough situation to mock while using the app on staging though.

singhhari (approved):
nice write up. left one small comment. after addressing justins comments lgtm!

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
msadoon: My only question is if I use `...BackingFragment` it will self reference `...ProjectFragment` which will result in a infinite loop, so instead of `...BackingFragment`, I'll add all the properties inside the fragment to project fragment's `backing { ... }` except `project` if that makes sense.

KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
msadoon: Yea, fair point, I left this out only because `Backing+BackingFragmentTests` is only really testing the `Backing` object creation, `friends` is on `project.personalization` which is out of testing scope here. Also trying not to bloat up this `jsonDictionary`.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
msadoon: If you have a minute to pair on this, would appreciate it!

msadoon: Ok, so revisiting your original comment, it seems like not making `backing` part of the fragment is the way to go. This static func is being called by `static func envelopeProducer(from data: GraphAPI.FetchBackingQuery.Data)` which has made the backing query already, which contains the `project`, so we can just pass it in here and remove `backing` from the `project fragment`.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| KsApi/models/graphql/adapters/Project+ProjectFragment.swift:34:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

justinswart (approved):
lgtm!

PR #1564, opened 2021-07-20
[NT-2065] Migrating User account fields, CCs and email queries to UserFragment

singhhari:
# 📲 What

We're adding a new query to handle fetching the user with the Apollo graphQL client. Part of this work includes updating our `UserFragment` to consolidate a number of fields we were fetching separately before. That includes: `account`, `email` and `storedCards`.

# 🤔 Why

Our old implementation in `UserQueries.swift` contained separate queries to fetch the minimal data for their respective views. We are consolidating this information because it's not a tremendous amount of data but also because it allows our `UserFragment` to encapsulate all the information for us. 

# 🛠 How

- Updated `UserFragment` with all fields required for a given `GraphUser`.
- Updated `Service` to eliminate the need for referencing `UserQueries.swift` and rely on the graphQL client instead.
- Update all tests that reference the `UserFragment` (this was painful).
- Create adapters and producers for supporting our new Apollo networking paradigm.

# ✅ Acceptance criteria

- [x] Open up settings and verify the Account Settings are rendering with the correct information.
- [x] Open up payment methods from the settings and verify stored cards are displaying correctly.
- [x] Open up payment methods from a backing and verify stored cards are displaying correctly.
- [x] Test all queries referencing `UserFragment` and verify there are no regressions (a global search for `...UserFragment` will show where it's being used)


KsApi/GraphAPI.swift
```diff
@@ -3871,6 +3871,112 @@ public enum GraphAPI {
     }
   }
 
+  public final class FetchUserQuery: GraphQLQuery {
```
singhhari, line 3874: This entire diff is auto-generated by the Apollo client.

Kickstarter-iOS/ViewModels/SettingsAccountViewModel.swift
```diff
@@ -43,7 +43,7 @@ public final class SettingsAccountViewModel: SettingsAccountViewModelInputs,
     )
     .switchMap { _ in
       AppEnvironment.current.apiService
-        .fetchGraphUserAccountFields(query: UserQueries.account.query)
```
singhhari, line 46: `fetchGraphUser` is now the only method we need to call to get any of the various details associated with the Graphser. Including but not limited to `storedCards`, `account` or `email`.

Kickstarter-iOS/ViewModels/SettingsAccountViewModelTests.swift
```diff
@@ -33,7 +33,7 @@ internal final class SettingsAccountViewModelTests: TestCase {
 
   func testReloadData() {
     let userEnvelope = UserEnvelope(me: GraphUser.template)
-    let mockService = MockService(fetchGraphUserAccountFieldsResponse: userEnvelope)
+    let mockService = MockService(fetchGraphUserResponse: userEnvelope)
```
singhhari, line 36: I renamed the parameter to reflect the change in functionality so you'll see this in a handful of files.

KsApi/MockService.swift
```diff
@@ -76,11 +76,8 @@
 
     fileprivate let fetchGraphUserEmailFieldsResponse: UserEmailFields?
 
-    fileprivate let fetchGraphCreditCardsResponse: UserEnvelope<GraphUserCreditCard>?
-    fileprivate let fetchGraphCreditCardsError: GraphError?
-
-    fileprivate let fetchGraphUserAccountFieldsResponse: UserEnvelope<GraphUser>?
-    fileprivate let fetchGraphUserAccountFieldsError: GraphError?
+    fileprivate let fetchGraphUserResponse: UserEnvelope<GraphUser>?
```
singhhari, line 79: Consolidating the fetching of the user to a single response and error now.

KsApi/models/UserEnvelope.swift
```diff
@@ -1,3 +1,16 @@
+import Foundation
+import ReactiveSwift
+
 public struct UserEnvelope<T: Decodable>: Decodable {
   public let me: T
 }
+
+// MARK: - GraphQL Adapters
+
+extension UserEnvelope {
```
singhhari, line 10: Transforming the Apollo clients `Data` type, for the given query, to a `SignalProducer` that emits our envelope.

KsApi/models/graphql/GraphBackingTests.swift
```diff
@@ -22,7 +22,18 @@ final class GraphBackingTests: XCTestCase {
               "id": "QmFja2luZy0xMTMzMTC5ODE=",
               "uid": "565656",
               "name": "Backer McGee",
-              "imageUrl": "http://www.kickstarter.com/avatar.jpg"
+              "imageUrl": "http://www.kickstarter.com/avatar.jpg",
```
singhhari, line 25: Because the `UserFragment` was updated and it's referenced in other queries/fragment, other than the `FetchUser` query, lots of tests and stub data needed to be updated. This was a lot of tedious work.

KsApi/models/graphql/GraphUserCreditCard.swift
```diff
@@ -2,7 +2,7 @@
 
 // TODO: Remove public access
 public struct GraphUserCreditCard: Decodable {
-  public var storedCards: CreditCardConnection
+  public var nodes: [CreditCard]
```
singhhari, line 5: We don't need the `CreditCardConnection` class anymore based on the testing. By removing it we no longer need to access the cards from the envelope like this:

`$0.fragments.userFragment.storedCards.storedCards.nodes` and we can eliminate one of those `storedCards` properties.

I also did this because a handful of tests were failing because the decoding was expecting two `storedCards` keys and our response from the server is not structured that way. This is a more accurate representation of what graph is returning.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/PledgeViewControllerTests.swift
```diff
@@ -100,11 +108,15 @@ final class PledgeViewControllerTests: TestCase {
...
-    ]))
-    let mockService = MockService(fetchGraphCreditCardsResponse: userEnvelope)
+    let cardTemplate = GraphUserCreditCard(
+      nodes: [
+        GraphUserCreditCard.visa,
+        GraphUserCreditCard.masterCard
+      ]
+    )
+    let withCards = GraphUser.template |> \.storedCards .~ cardTemplate
+    let response = UserEnvelope<GraphUser>(me: withCards)
```
msadoon, line 118: Actually we seem to be using the same `response` a lot, can't you make it a global constant to the file?

Kickstarter-iOS/Views/Controllers/PledgeViewControllerTests.swift
```diff
@@ -402,11 +426,15 @@ final class PledgeViewControllerTests: TestCase {
...
+    let mockService = MockService(fetchGraphUserResponse: response)
```
msadoon, line 437: Can reduce a lot of code duplication and reference the same `response` in every change on this file.

KsApi/models/graphql/GraphBackingTests.swift
msadoon: glad you got it working!

msadoon: what ended up being the issue with the `missingValue`?

KsApi/models/graphql/GraphBackingTests.swift
```diff
@@ -63,6 +74,17 @@ final class GraphBackingTests: XCTestCase {
...
                 "id": "VXNlci0xMjA3OTk3NjQ5",
                 "name": "Creator McBaggins",
+                "storedCards": [
+                  "nodes": [
+                    [
+                      "expirationDate": "2023-01-01",
+                      "id": "6",
+                      "lastFour": "4242",
+                      "type": "VISA"
+                    ]
```
msadoon, line 84: It might be simpler and less json to return empty arrays of objects we don't directly test. In this case we technically don't even check the stored cards, just that the JSON builds with some (even an empty) array of stored cards.

msadoon: It's fine in this case, I just mean when we have like 8 objects in an array, we don't absolutely need to make every json file that reads in those objects the same.

KsApi/models/graphql/GraphUserCreditCardTests.swift
```diff
@@ -25,23 +23,21 @@ final class GraphUserCreditCardTests: XCTestCase {
...
-              "type": "VISA"
-            },
-            {
-              "expirationDate": "2020-02-01",
-              "lastFour": "1111",
-              "id": "2768",
-              "type": "VISA"
-            }
-        ],
-        "totalCount": 0
```
msadoon, line 43: weird that this was 0 before

KsApi/models/graphql/ManagePledgeViewBackingEnvelopeTests.swift
```diff
@@ -605,7 +696,20 @@ final class ManagePledgeViewBackingEnvelopeTests: XCTestCase {
             "imageUrl": "http://www.kickstarter.com/avatar.jpg",
             "id": "VXNlci0xMjA3OTk3NjQ5",
             "uid": "1100793144",
-            "name": "Creator McBaggins"
+            "name": "Creator McBaggins",
+            "storedCards": [
```
msadoon, line 700: It might be worth it (later) to store these as separate `templateFragments` (just strings) that we can compose these json files with in the future. They are quite big.

KsApi/models/graphql/adapters/GraphUserCreditCard+UserFragment.swift
```diff
@@ -0,0 +1,23 @@
...
+        expirationDate: node.expirationDate,
+        id: node.id,
+        lastFour: node.lastFour,
+        type: CreditCardType(rawValue: node.type.rawValue)
+      )
+    }
+
+    return GraphUserCreditCard(nodes: creditCards)
+  }
+}
```
msadoon, line 23: Missing tests

KsApi/models/graphql/adapters/UserEnvelope+GraphUserEnvelopeTests.swift
```diff
@@ -0,0 +1,63 @@
...
+    XCTAssertEqual(envelope.me.isEmailVerified, true)
+    XCTAssertEqual(envelope.me.name, "Hari Singh")
+    XCTAssertEqual(
+      envelope.me.storedCards,
+      GraphUserCreditCard(nodes: [
+        GraphUserCreditCard
+          .CreditCard(expirationDate: "2023-01-01", id: "6", lastFour: "4242", type: .visa)
+      ])
+    )
+    XCTAssertEqual(envelope.me.uid, "1470952545")
```
msadoon, line 61: great test!

msadoon (changes requested):
Overall looks good, I checked my account settings, payment methods in settings and in pledging and all of it looks normal.

Seems like we missed a couple places where the `fetchGraphUser` call can replace `fetchGraphUserEmailFields`
`ChangeEmailViewModel` `line 75`, `SettingRequestDataCellViewModel` `line 42`
Just so then we can remove `fetchGraphUserEmailFields`.

Other than that minor comments regarding testing. Not all comments need to be addressed, some are just observations.

KsApi/models/graphql/GraphBackingTests.swift
singhhari: turns out by keeping `CreditCardConnection` which had a property called storedCards too the decoder was expecting two `storedCards` keys instead of the one actually returning from the response. took me far longer than im willing to admit to catch that.

KsApi/ServiceType.swift
```diff
@@ -144,13 +144,9 @@ public protocol ServiceType {
...
 
-  /// Fetch User's stored cards.
-  func fetchGraphCreditCards(query: NonEmptySet<Query>)
-    -> SignalProducer<UserEnvelope<GraphUserCreditCard>, GraphError>
-
-  /// Fetch a User's account fields
-  func fetchGraphUserAccountFields(query: NonEmptySet<Query>)
-    -> SignalProducer<UserEnvelope<GraphUser>, GraphError>
+  /// Fetches various fields of a given User using graphQL.
+  func fetchGraphUser()
```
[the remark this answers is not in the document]
singhhari: Thought about that but it would be misleading because we have both a `User` and `GraphUser` object. This particular call is fetching the `GraphUser`.

KsApi/fragments/UserFragment.graphql
```diff
@@ -1,7 +1,22 @@
 fragment UserFragment on User {
```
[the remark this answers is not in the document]
singhhari: Ah I didn't notice that but it definitely will have to be addressed eventually. Going to flag this to Paulo to put in another ticket because this PR is way too large already.

Kickstarter-iOS/Views/Controllers/PaymentMethodsViewControllerTests.swift
```diff
@@ -17,26 +17,26 @@ class PaymentMethodsViewControllerTests: TestCase {
...
 
-  private func generateSnapshots(with response: UserEnvelope<GraphUserCreditCard>) {
+  private func generateSnapshots(with response: UserEnvelope<GraphUser>) {
     combos(Language.allLanguages, Device.allCases).forEach {
       arg in
 
       let (language, device) = arg
       withEnvironment(
-        apiService: MockService(fetchGraphCreditCardsResponse: response),
+        apiService: MockService(fetchGraphUserResponse: response),
```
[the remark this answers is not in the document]
singhhari: the last commit I just pushed (addressing the userEmailFields) broke a bunch of snapshots because the email address changed but that's not a big deal. again lots of tedious, time consuming small tasks but functionality is still fine.

Kickstarter-iOS/Views/Controllers/PledgeViewControllerTests.swift
singhhari: Created a private constant for userWithCards which uses the GraphUser.template and sets the storedCards but I can't store the `response` because it references self so that will have to be re-created for each test method.

KsApi/models/graphql/GraphUserCreditCardTests.swift
singhhari: A very old typo for sure.

KsApi/models/graphql/GraphBackingTests.swift
singhhari: Yeah good point. The reason you see a lot of copy pasted json is because the UserFragment is referenced in a lot of different places. When applying this migration it was breaking everywhere so I was individually debugging a ton of places. Will keep this in mind moving forward.

KsApi/models/graphql/ManagePledgeViewBackingEnvelopeTests.swift
singhhari: Justin left a few `FIXME` in his original few migrations that mention the need to add templates for these models and updating the tests. We should start doing that.

KsApi/models/graphql/adapters/GraphUserCreditCard+UserFragment.swift
singhhari: Updated in the last commit.

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -133,7 +133,7 @@ final class ChangeEmailViewModelTests: TestCase {
 
       // Disabled if new email is equal to the old one.
       self.vm.inputs.emailFieldTextDidChange(text: "ksr@kickstarter.com")
-      self.saveButtonIsEnabled.assertValues([true, false])
+      self.saveButtonIsEnabled.assertValues([true, true])
```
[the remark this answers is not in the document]
msadoon: I tried changing the template email to the "kst@kickstarter.com" but it still sends true back as its last value.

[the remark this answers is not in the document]
msadoon: So not sure.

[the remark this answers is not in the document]
msadoon: In app functionality only enables the "Save" button when the email is different..

msadoon (approved):
Good to go!

Library/ViewModels/ChangeEmailViewModelTests.swift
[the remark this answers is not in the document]
singhhari: Yeah I'm under the impression this was a bug that was never addressed in the test because in this particular unit test the email address is changing therefore the save button should be enabled.

PR #1566, opened 2021-07-26
[NT-2113] Migrate WatchProjectMutation to Apollo

singhhari:
# 📲 What

In this PR we're migrating our old implementation of the `WatchProjectMutation`, in the service layer, to our Apollo client. We're pushing this to our feature branch in order to separate concerns between our Apollo work and anything else that may need to go into `main` beforehand. The `WatchProjectMutation`, as the name suggests, handles the "watching" (tapping the heart button) of a project via GraphQL.

# 🤔 Why

This is just another PR in our large migration to GraphQL with Apollo. We are moving on from our original custom DSL based implementation and planning on relying on Apollo from this point forward.

# 🛠 How

- Added the .graphql mutation file
- Updated our `ServiceType` protocol to handle errors with the `ErrorEnvelope` class.
- Update the `Service` object to use the `GraphQL` client to perform the mutation.
- Created adapters for converting types between the `GraphAPI` and our local definitions.
- Added new tests and updated older ones.

# ✅ Acceptance criteria

- [ ] Log in with a test user and "watch" a project from the various areas we allow that. That includes the Discovery, Search and Profile tabs. 
- [ ] Verify with Charles proxy that the network requests succeed and the details of the request are accurate.


KsApi/MockService.swift
```diff
@@ -170,12 +170,15 @@
 
     fileprivate let unwatchProjectMutationResult: Result<
       GraphMutationWatchProjectResponseEnvelope,
-      GraphError
+      ErrorEnvelope
```
singhhari, line 173: Although the migration to Apollo, for this method, is not included here, I had to update the function signature because the view model was consolidating some of the logic between `watch` and `unwatch`.  This was needed for compilation.

KsApi/models/graphql/adapters/GraphMutationWatchProjectResponseEnvelope+WatchProjectMutation.Data.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+import ReactiveSwift
+
+extension GraphMutationWatchProjectResponseEnvelope {
```
msadoon, line 4: Can create a new struct called `WatchProjectResponseEnvelope` and extend it here, just like we did for `CreateBackingEnvelope` and `UpdateAccountEnvelope`, that way we can remove dependency on `GraphMutationWatchProjectResponseEnvelope` or in theory any other `Graph` response. Reason being they will be easier to clean up in the future.

KsApi/models/graphql/adapters/GraphMutationWatchProjectResponseEnvelope+WatchProjectMutation.Data.swift
```diff
@@ -0,0 +1,31 @@
...
+      isWatched: projectFromData.isWatched
+    )
+
+    return GraphMutationWatchProjectResponseEnvelope(watchProject: WatchProject(project: project))
+  }
+
+  /**
+   Return a signal producer containing `GraphMutationWatchProjectResponseEnvelope` or `ErrorEnvelope`
+   */
+  static func producer(from data: GraphAPI.WatchProjectMutation
```
msadoon, line 23: Going with the suggestion above, this producer would work for something like `WatchProjectResponseEnvelope`.

msadoon (commented):
Functionality works as expected.
I just wanted to know if a) we could convert the `GraphMutationWatchProjectResponseEnvelope` to a `WatchProjectResponseEnvelope` so as to reduce dependency on `Graph` objects being passed around. 
b) If we could convert over the `unwatchProject` mutation as well, doesn't seem to be too much overhead.

KsApi/Service.swift
```diff
@@ -572,7 +573,9 @@ public struct Service: ServiceType {
   }
 
   public func watchProject(input: WatchProjectInput) ->
-    SignalProducer<GraphMutationWatchProjectResponseEnvelope, GraphError> {
-    return applyMutation(mutation: WatchProjectMutation(input: input))
+    SignalProducer<GraphMutationWatchProjectResponseEnvelope, ErrorEnvelope> {
```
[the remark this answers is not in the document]
singhhari: Makes sense. Renamed it.

KsApi/models/graphql/adapters/GraphMutationWatchProjectResponseEnvelope+WatchProjectMutation.Data.swift
singhhari: Hmm am I missing something here. Wouldn't globally renaming the envelope to remove the `GraphMutation` prefix be sufficient here?

KsApi/models/graphql/adapters/GraphMutationWatchProjectResponseEnvelope+WatchProjectMutation.Data.swift
singhhari: Yup, globally renaming the namespace in the last commit.

singhhari:
> Functionality works as expected.
> I just wanted to know if a) we could convert the `GraphMutationWatchProjectResponseEnvelope` to a `WatchProjectResponseEnvelope` so as to reduce dependency on `Graph` objects being passed around.
> b) If we could convert over the `unwatchProject` mutation as well, doesn't seem to be too much overhead.

Addressed the renaming with global Refactor -> Rename but did not address the unwatchProject migration because that is represented in a separate ticket and we prioritized other tickets, yesterday, over it in the sprint. Let me know what you think.

msadoon (approved):
Yea great, I didn't realize it was just a renaming, good to go!

PR #1567, opened 2021-07-27
[NT-2147] Comments Not Displaying For Unauthenticated Users

msadoon:
# 📲 What

Two issues were found on staging:
1. Couldn't scroll comments due to each `CommentFragment` requiring a `UserFragment` to contain the `storedCards` property. The `storedCards` property is only available for authenticated sessions.
2. An issue related only to my staging accounts was found to mix up credit cards when logging in and out of the pledge screen. Upon further testing by @singhhari and @jgsamudio it was found to only be related to staging and my accounts in particular.

[ticket](https://kickstarter.atlassian.net/browse/NT-2147)

# 🤔 Why

We want to allow un-authenticated users to scroll comments and replies of a project. Also want to minimize the amount of data required on fragments selectively based on it's use case.

# 🛠 How

Made the `...UserStoredCardsFragment` that will call the stored cards portion of the `UserFragment` if the declarative (`showStoredCards`) is `true` in a query.

Effected queries `FetchUser.graphql`, `FetchUpdateComments.graphql`, `FetchProjectComments.graphql`, `FetchBacking.graphql` and `FetchAddOns.graphql`.

Alot of minor changes around testing were made because the `GraphUser` is used in many places and the `storedCards` property was changed on it. Also the changes to `MockService` where `Errors` and `Responses` were changed to `Result` that could contain both.

Finally the addition of a declarative requires `variables` to be passed into creating a `Fragment` separately of a `Query`.

# 👀 See

|  |  |

# ✅ Acceptance criteria

- [x] Can scroll comments and replies when authenticated/not.
- [x] Lots of calls were touched here, and while I'm confident nothing broke, might be good to smoke test things like add-ons, backings, rewards, comments, projects. Not thoroughly just ensuring that there isn't any weirdness.


KsApi/MockService.swift
```diff
@@ -856,7 +853,11 @@
       return producer(for: self.fetchManagePledgeViewBackingResult)
     }
 
-    func fetchManagePledgeViewBacking(id _: Int) -> SignalProducer<ProjectAndBackingEnvelope, ErrorEnvelope> {
+    func fetchManagePledgeViewBacking(id _: Int,
```
singhhari, line 856: Strange formatting here.

KsApi/MockService.swift
```diff
@@ -736,11 +733,11 @@
       return SignalProducer(value: CategoryEnvelope(node: .template |> Category.lens.id .~ "\(query.head)"))
     }
 
-    internal func fetchGraphUser()
+    internal func fetchGraphUser(withStoredCards _: Bool)
```
singhhari, line 736: IS the underscore after the `withStoredCards` parameter needed here and the other altered methods above?

KsApi/fragments/UserFragment.graphql
```diff
@@ -9,14 +9,8 @@ fragment UserFragment on User {
...
-    nodes {
-      expirationDate
-      id
-      lastFour
-      type
-    }
-    totalCount
-  }
   uid
+  storedCards @include(if: $withStoredCards) {
```
singhhari, line 13: nice use of the directive here.

KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
```diff
@@ -5,7 +5,8 @@ import XCTest
 final class Backing_BackingFragmentTests: XCTestCase {
   func test() {
     do {
-      let fragment = try GraphAPI.BackingFragment(jsonObject: backingDictionary())
+      let variables = ["withStoredCards": true]
```
singhhari, line 8: Think you can pass the dictionary directly in as an argument since we aren't referencing it anywhere else?

KsApi/models/graphql/adapters/GraphUserCreditCard+UserFragment.swift
```diff
@@ -4,20 +4,22 @@ extension GraphUserCreditCard {
   /**
    Returns a `GraphUserCreditCard` from a `GraphAPI.UserFragment`
    */
-  static func graphUserCreditCard(from userFragment: GraphAPI.UserFragment) -> GraphUserCreditCard? {
-    guard let storedCards = userFragment.storedCards,
-      let nodes = storedCards.nodes else { return nil }
+  static func graphUserCreditCard(from userFragment: GraphAPI.UserFragment) -> GraphUserCreditCard {
+    guard let storedCards = userFragment.storedCards?.fragments.userStoredCardsFragment.nodes else {
+      return GraphUserCreditCard(storedCards: [])
```
singhhari, line 9: Coverage is down a bit. This might be an easy place to write a test for to bump it up.

KsApi/models/graphql/adapters/GraphUserCreditCard+UserFragmentTests.swift
```diff
@@ -4,33 +4,42 @@ import XCTest
...
 
-    XCTAssertNotNil(GraphUserCreditCard.graphUserCreditCard(from: userFragment))
+    do {
+      let storedCardsFragment = try GraphAPI.UserFragment
+        .StoredCard(jsonObject: sampleCardDict, variables: variables)
+      let userFragment = GraphAPI.UserFragment(
+        chosenCurrency: "USD",
+        email: "foo@bar.com",
+        hasPassword: true,
+        id: GraphQLID(),
```
singhhari, line 28: Found out yesterday that a standard String of an id works in place of the GraphQLID object. Can you try replacing that here and see if it works?

KsApi/models/graphql/adapters/GraphUserCreditCard+UserFragmentTests.swift
```diff
@@ -4,33 +4,42 @@ import XCTest
...
+        isDeliverable: true,
+        isEmailVerified: true,
+        name: "Hari Singh",
+        uid: "12345",
+        storedCards: storedCardsFragment
+      )
+
+      XCTAssertTrue(GraphUserCreditCard.graphUserCreditCard(from: userFragment).storedCards.count == 1)
+    } catch {
+      print("*** \(error.localizedDescription)")
```
singhhari, line 41: Left a print here.

KsApi/queries/FetchAddOnsQuery.graphql
```diff
@@ -1,4 +1,4 @@
-query FetchAddOns($projectSlug: String!, $shippingEnabled: Boolean!, $locationId: ID) {
+query FetchAddOns($projectSlug: String!, $shippingEnabled: Boolean!, $locationId: ID, $withStoredCards: Boolean!) {
```
singhhari, line 1: I might be missing something here but I see the parameter declaration here but where is it being passed into the fragment for the declarative?

Library/ViewModels/ChangeEmailViewModelTests.swift
```diff
@@ -27,6 +28,21 @@ final class ChangeEmailViewModelTests: TestCase {
   private let warningMessageLabelHidden = TestObserver<Bool, Never>()
   private let verificationEmailButtonTitle = TestObserver<String, Never>()
 
+  // MARK: Computed Properties
+
+  private var userChangeEmailSuccessMockService: MockService {
+    guard let envelope = UserEnvelope<GraphUser>.userEnvelope(from: fetchUserQueryData) else {
+      return MockService()
```
singhhari, line 35: Should this be MockService() or should it pass a failure as the result?

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -303,36 +317,39 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
       return
     }
 
     let result1 = DeletePaymentMethodEnvelope(storedCards: [card, card])
-
-    let apiService1 = MockService(deletePaymentMethodResult: .success(result1))
+    let response = UserEnvelope<GraphUser>(me: userTemplate)
+    let apiService1 = MockService(
+      deletePaymentMethodResult: .success(result1),
+      fetchGraphUserResult: .success(response)
```
singhhari, line 336: Why was fetchGraphUserResult added to all these tests if they weren't here before?

singhhari (changes requested):
Works as expected so nice job on that. Left some comments and hoping to get the coverage back up to at least even.

KsApi/MockService.swift
msadoon: We don't need it, just following the conventions above for things like `fetchProjectComments` and `fetchUpdateComments`, but okay removed in this call.

KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
msadoon: Yea I can, no problem, but I try to keep variables declared separately because it's easier to read.

KsApi/queries/FetchAddOnsQuery.graphql
msadoon: yea I was curious about that too. Having done some testing with GraphiQL, I found queries like this just pass down the parameter into fragments of fragments that use it. I suppose when everything gets compiled it's just one long string that is like one big query and all the parameters get referenced somewhere.

Try this out in GraphiQL to get an idea of not needing to pass every fragment a copy of the parameter:
```
query FetchUser($withStoredCards: Boolean = true) {
  me {
    ...UserFragment
  }
}

fragment UserFragment on User {
  chosenCurrency
  email
  hasPassword
  id
  imageUrl: imageUrl(blur: false, width: 1024)
  isAppleConnected
  isCreator
  isDeliverable
  isEmailVerified
  name
  uid
  storedCards @include(if: $withStoredCards) {
    __typename
    ...UserStoredCardsFragment
  }
}

fragment UserStoredCardsFragment on UserCreditCardTypeConnection {
    nodes {
      __typename
      expirationDate
      id
      lastFour
      type
    }
    totalCount
}
```

Also taking this parameter out causes compile errors even for seemingly unrelated queries like `FetchAddOns`.

Library/ViewModels/ChangeEmailViewModelTests.swift
msadoon: Well, I'm not sure that we can then pass the failure to the `withEnvironment(apiService:` parameter, might be simpler to just pass in the basic `MockService()` with no results to fetch, the test should fail in this anyway.

Library/ViewModels/PaymentMethodsViewModelTests.swift
msadoon: Yea, I don't have a great answer here.
I felt comfortable doing this here because these were the view models where we added `fetchGraphUser(withStoredCards: true)` and changed the return type to `Result` instead of a `GraphResponse` in the `MockService`. Short answer is they were set up incorrectly to begin with, so adding these calls fixed pre-existing issues.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| KsApi/MockService.swift:736:1 | warning: (unusedArguments) Mark unused function arguments with _. |


</p>

KsApi/MockService.swift
msadoon: ok nvm, `bin/format.sh .` does this.

KsApi/queries/FetchAddOnsQuery.graphql
singhhari: Ah yeah looks like it's doing that for us. Cool stuff.

Library/ViewModels/PaymentMethodsViewModelTests.swift
singhhari: Gotcha, wasn't sure so thanks for clarifying.

singhhari (approved):
lgtm!

PR #1569, opened 2021-07-27
[NT-2101] Migrating CancelBacking mutation to Apollo.

singhhari:
# 📲 What

The mutation we use for canceling a backing was originally implemented with our custom domain specific language for handling GraphQL. This PR includes the changes that migrate our implementation of this mutation to Apollo.

# 🤔 Why

We are taking all of our GraphQL networking layer and using Apollos GraphQL client to handle this for us. This PR represents another piece, among many others, that we are incrementally updating.

# 🛠 How

- Updated the `ServiceType`, `Service` and `MockService` objects to handle errors with an `ErrorEnvelope` and forming the graph request with `GraphQL.shared.client`.
- Rename `GraphMutationEmptyResponseEnvelope` to `EmptyResponseEnvelope` to free up the namespace that we are eventually planning on deleting.
- Update tests.

# ✅ Acceptance criteria

- [ ] On staging, with your test user, navigate to the Profile and select a project you have successfully backed.
- [ ] Using the kebab menu, cancel your backing.
- [ ] Using Charles proxy, verify the request was successful.
- [ ] Verify the backing is removed from the user and no regressions were introduced.


KsApi/GraphMutation.swift
```diff
@@ -10,7 +10,7 @@ protocol GraphMutation: CustomStringConvertible {
   var input: Input { get }
 }
 
-public struct GraphMutationEmptyResponseEnvelope: Decodable {}
+public struct EmptyResponseEnvelope: Decodable {}
```
singhhari, line 13: Removing the `GraphMutation` prefix from this like we did in the previous PR for the WatchProject mutation.

KsApi/Service.swift
```diff
@@ -92,8 +92,13 @@ public struct Service: ServiceType {
...
 
   public func cancelBacking(input: CancelBackingInput)
-    -> SignalProducer<GraphMutationEmptyResponseEnvelope, GraphError> {
-    return applyMutation(mutation: CancelBackingMutation(input: input))
+    -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope> {
+    return GraphQL.shared.client
+      .perform(mutation: GraphAPI
+        .CancelBackingMutation(input: GraphAPI.CancelBackingInput(id: input.backingId)))
+      .flatMap { _ in
+        SignalProducer(value: EmptyResponseEnvelope())
```
singhhari, line 100: No need for an adapter here since we are simply returning the empty envelope, like before.

KsApi/Service.swift
```diff
@@ -92,8 +92,13 @@ public struct Service: ServiceType {
   }
 
   public func cancelBacking(input: CancelBackingInput)
-    -> SignalProducer<GraphMutationEmptyResponseEnvelope, GraphError> {
-    return applyMutation(mutation: CancelBackingMutation(input: input))
+    -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope> {
+    return GraphQL.shared.client
+      .perform(mutation: GraphAPI
+        .CancelBackingMutation(input: GraphAPI.CancelBackingInput(id: input.backingId)))
```
singhhari, line 98: We don't test these files so I can't really bump coverage on the diff of this PR any further.

KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -55,8 +55,8 @@ extension ApolloClient {
           }
           observer.send(value: data)
           observer.sendCompleted()
-        case let .failure(error):
-          observer.send(error: .couldNotDecodeJSON(error))
+        case .failure:
+          observer.send(error: .couldNotParseJSON)
```
msadoon, line 59: yea so I'm not sure why we wouldn't want to send the error through.

msadoon: Yea so I guess my only suggestion against using the `EmptyResponseEnvelope` in `cancelBacking` and in `changeCurrency` is that a failed mutation can still return from GQL as a successful parseable response, which means it will contain parent key but `nil` value . In the case of `cancelBacking`:
```
    let dict: [String: Any?] = [
      "cancelBacking": [
        "clientMutationId": nil
      ]
    ]
```
will still be successful even though clientMutationId is `nil`, but 
```
   let dict: [String: Any?] = [
      "cancelBacking": nil
    ]
```
will be a successful response but we can consider it a failure because top level dict didn't return anything.

Really what I'm trying to figure out here is if a mutation fails server side, we will still get a successful response. The json may still be parsable, but how do we know it really did succeed? The `clientMutationId`? and top level dict in the response?

msadoon (commented):
Yea so as discussed on the call, we just need to figure out what an error Envelope looks like so we can customize an error envelope for `cancelBacking`

KsApi/extensions/ApolloClient+RAC.swift
singhhari: Synced on a call about this and determined we are going to continue to use the `EmptyResponseEnvelope`.

msadoon (approved):
Not 100% sure how the test in `CancelPledheViewModelTests` is passing, but otherwise good to go!

PR #1576, opened 2021-08-04
[NT-2102] Migrating ClearUserUnseenActivityMutation to Apollo.

singhhari:
# 📲 What

When a user receives an update on their project or pledge, it appears in the Activity tab on the Kickstarter iOS application. We use the `ClearUserUnseenActivityMutation` to appropriately render the number badge that shows on the tab. This PR will handle migrating our original implementation of the mutation to the Apollo GraphQL client variant.

# 🤔 Why

This is another addition to our Apollo GraphQL client epic. We are merging all of these PRs into a feature branch first, planning on a larger regression test and then merging to main.

# 🛠 How

- Update `ServiceType` protocol to emit an `ErrorEnvelope` instead of `GraphError`.
- Update method in `Service` to use the GraphQL client instead of our older DSL.
- Create an adapter for the mutation, its input and a `SignalProducer`.
- Add unit tests.

# ✅ Acceptance criteria

- [ ] Produce a scenario on staging where the activity tab displays a red badge on the tab (e.g. errored payment)
- [ ] While Charles proxy is open, verify the mutation is occurring successfully and without regression.


msadoon (approved):
good to go!
Didn't test in-app functionality, but I trust that it's working as intended. I'm sure you've tested it and there is no regression.

PR #1577, opened 2021-08-04
[NT-2168] User Authentication Not Cleared After Logout and Login

msadoon:
# 📲 What

A bug was found with our current Apollo Client that was making requests on behalf of a previously logged in user _after_ logging out and logging back in with a different user.

# 🤔 Why

One scenario this was happening was with stored user cards. Log out current user and log in with new one, go to pledge check out and see the previous users' cards listed. Also effected the Account Settings (email, payment method).

# 🛠 How

The culprit in this case was a lazily instantiated `ApolloClient` singleton instance. Because the `client` in `ApolloClient+Singleton` was lazily instantiated, everytime `configure` was called through `Service`, this underlying instance of the `apollo`  var was changed, but returned value of `client` was the previous instance of `apollo`. That's how lazy vars work, they only instantiate once and only when called upon.

So to give more context, we replace the current environment with new auth tokens everytime we log out/in. Whenever this happens `GraphQL.shared.configure` is called with the new headers. Previously we expected our `client` to pass those headers on to the intercepters, which it was doing correctly. Because the `client` was never calling the new instance of `apollo` after the environment was reset, the old headers were being used in calls.

Also updated the passing on of perimeterX headers as a dictionary instead of a function, because even though the closure is escaping, it still references a class, and if the Service struct gets reset before the escaping closure gets called, there is memory leak and the app crashes. Passing the result of the headers function on to the Service struct then to the GraphQL class is safer because it is a struct itself.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 🏎 Performance

Note on this:
Looking through the [documentation](https://www.apollographql.com/docs/ios/initialization/) it does seem like we should be creating an interceptor for values that might change on a per call basis (ie. oauthtoken). However, we simply recreate a new Apollo Client everytime that variable updates, so didn't think it was necessary to add here. But a good way to do this is to have some sort of `UserManager` singleton that can be directly called by an interceptor to update the authtoken at time of in-flight request and that way authtoken is managed separately from networking layer.

# ✅ Acceptance criteria

- [x] Switch users, notice no payment options/emails/account data gets transferred over.


KsApi/extensions/ApolloClient+Singleton.swift
```diff
@@ -25,7 +25,7 @@ class GraphQL {
   func configure(
     with url: URL,
     headers: [String: String],
-    additionalHeaders: @escaping () -> [String: String]
+    additionalHeaders: [String: String]
```
singhhari, line 28: is there a reason you removed @escaping here? we are relying on a network request to get these headers so it's probably a good idea to keep it.

KsApi/Service.swift
```diff
@@ -53,7 +53,7 @@ public struct Service: ServiceType {
     GraphQL.shared.configure(
       with: serverConfig.graphQLEndpointUrl,
       headers: self.defaultHeaders,
-      additionalHeaders: { perimeterXClient.headers() }
+      additionalHeaders: self.perimeterXClient.headers()
```
singhhari, line 56: What's the reason this change had to be made? Seems like it's doing the same thing but still curious.

KsApi/extensions/ApolloClient+Singleton.swift
```diff
@@ -5,13 +5,13 @@ class GraphQL {
   // MARK: - Properties
 
   private var apollo: ApolloClient?
-  private(set) lazy var client: ApolloClient = {
-    guard let client = self.apollo else {
+  var client: ApolloClient {
+    guard let client = apollo else {
```
singhhari, line 9: `self.apollo` instead of just `apollo`

KsApi/extensions/ApolloClient+Singleton.swift
singhhari: Most of the changes below seem to be a result of this so I'm trying to be careful about the specifics here.

singhhari (commented):
Works as expected but have a few questions about the implementation.

singhhari (approved):

msadoon:
I won't merge this now, just want to get on a call and discuss these changes further, kind of explain my thinking a little and get your input first.

msadoon:
Maybe there's further refactoring to do. Because yes I see now the Apollo client documentation suggests a `lazy var`

PR #1579, opened 2021-08-06
Add Apollo Schema Download CI Step Headers

msadoon:
Quick PR correcting missing headers from our schema download step.
Somehow, the changeover from production to staging (data re-seed) on Aug 5/6 2021 caused the script to fail to download the schema on CI.
This fix seems to correct the issue.


bin/apollo-schema-download.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 npm install -D apollo
-npx apollo schema:download --endpoint=https://staging.kickstarter.com/graph KsApi/graphql-schema.json --header "Authorization: token ${GRAPHQL_API_AUTH_TOKEN}"
+npx apollo schema:download --header="Content-Type: application/json" --header="User-Agent: Kickstarter iOS" --endpoint=https://staging.kickstarter.com/graph KsApi/graphql-schema.json
```
singhhari, line 3: Just a heads up that if we make changes to the user agent in the future (as some discussions have been occurring around this for our Segment implementation), we will probably have to update this too.

singhhari (approved):
lgtm

bin/apollo-schema-download.sh
msadoon: Ah okay good to know.

PR #1581, opened 2021-08-10
[NT-2107] Migrate SignInWithAppleMutation to Apollo

singhhari:
# 📲 What

We're continuing to move forward with our migration from the legacy Domain Specific Language to Apollo for our GraphQL implementation. This PR introduces an update to our mutation for `SignInWithApple`. Included is the schema for the mutation, adapters and testing.

# 🤔 Why

We are aiming to handle 100 percent of our GraphQL networking with the Apollo client and this PR is one of the last few mutations remaining for migration. Apollo provides a more readable, understandable codebase for developers to work with and should enable future devs on the project to quickly move forward on introducing new queries or mutations.

# 🛠 How

- Added the schema to the mutation in a `.graphql` file.
- Created an adapter for the `SignInWithAppleInput`.
- Created an adapter for the `SignInWithAppleEnvelope`.
- Included a few small tests, including one that will need to be updated once we introduce a mock client for Apollo.

# ✅ Acceptance criteria

- [x] This is unfortunately not testable with the iOS simulator so we'll need a physical device. Build the debug app onto your physical device and proceed to select the `Continue With Apple` button in our login/signup interface.
- [x] Logging in with the relay id should work as expected and direct the user into Explore, upon success.
- [x] Verify there are no regressions.


KsApi/MockService.swift
```diff
@@ -116,8 +116,7 @@
     fileprivate let fetchUserProjectsBackedResponse: [Project]?
     fileprivate let fetchUserProjectsBackedError: ErrorEnvelope?
 
-    fileprivate let fetchUserResponse: User?
-    fileprivate let fetchUserError: ErrorEnvelope?
+    fileprivate let fetchUserResult: Result<User, ErrorEnvelope>?
```
singhhari, line 119: Updated this after some unit tests broke when I did a similar refactor for the `signInWithAppleResult` type.

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -570,7 +570,13 @@ final class LoginToutViewModelTests: TestCase {
   }
 
   func testShowAppleErrorAlert_FetchUserEventError() {
-    withEnvironment(apiService: MockService(fetchUserError: .couldNotParseJSON)) {
+    let envelope = SignInWithAppleEnvelope.template
+      |> \.signInWithApple.apiAccessToken .~ "some_token"
+
+    withEnvironment(apiService: MockService(
+      fetchUserResult: .failure(.couldNotParseJSON),
+      signInWithAppleResult: .success(envelope)
```
singhhari, line 578: Interesting observation:  

this unit test broke after I updated the MockService's signInWithApple method to return `producer(from:)`. I had to manually include the argument to fix it whereas previously the omission of the parameter was being handled as an error. This is related to some of the MockService methods defaulting to a response of a template (Justin mentioned this was troublesome).

msadoon: Oh right I remember now, ok that's why some tests suddenly fail when the `response` is changed to `result`. I wasn't sure exactly why it was happening, but as you mention its because those values were previously set to `nil` by default. I had the same issue when converting over mock service responses and errors into results for the `UserStoredCards` fragment.

msadoon (approved):
Good to go!

Wasn't able to test the case where the call fails, need to set up some additional Charles Proxy related on device things.

I'll do that separately, no need to block this PR from going through.

PR #1587, opened 2021-08-18
[NT-2160] Making error response human readable in the case of request failure.

singhhari:
# 📲 What

In the case of an outright request failure, our GraphQL client is sending an error prefixed with the string "JSONDecoder decoding error" which is useful for developers but not for a user of the application. This PR includes an update to remove that prefix and add a print statement above it in our `fetch` and `perform` method so we still keep the debugging utility of knowing what the error is specifically throwing.

# 🤔 Why + 🛠 How

GraphQL differs from a standard REST API in a number of ways including the way errors are thrown. A standard REST API will throw an http status code that suggests what the issue is, whereas GraphQL will return a `200` but include an error message, within its response. We are handling these legitimate errors in our `success` block but our `failure` block (which runs when there are network issues) was not as human readable as we liked. We removed anything other than the localized description of the error which is produced by the system and displayed when necessary.

# ✅ Acceptance criteria

- [ ] Run Charles proxy and introduce a breakpoint that aborts Graph requests. Verify the error message is human readable.
- [ ] Turn off your WiFi and attempt a Graph request on the app. Verify the error message is human readable.
- [ ] Compose a request on Charles that outputs a response structured as an error (200 http status code). e.g.:

`
{
  "errors": [
    {
      "message": "Field 'backing' doesn't exist on type 'User'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "fields": [
        "query FetchUserBackings",
        "me",
        "backing"
      ]
    }
}
`

Verify that error is displayed to the user.

### DIFF
--- a/KsApi/extensions/ApolloClient+RAC.swift
+++ b/KsApi/extensions/ApolloClient+RAC.swift
@@ -27,6 +27,7 @@ extension ApolloClient {
           observer.send(value: data)
           observer.sendCompleted()
         case let .failure(error):
+          print("🔴 [KsApi] ApolloClient query failure - error : \((error as NSError).description)")
           observer.send(error: .couldNotDecodeJSON(error))
         }
       }
@@ -56,6 +57,7 @@ extension ApolloClient {
           observer.send(value: data)
           observer.sendCompleted()
         case let .failure(error):
+          print("🔴 [KsApi] ApolloClient mutation failure - error : \((error as NSError).description)")
           observer.send(error: .couldNotDecodeJSON(error))
         }
       }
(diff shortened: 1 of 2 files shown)


KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -27,6 +27,7 @@ extension ApolloClient {
           observer.send(value: data)
           observer.sendCompleted()
         case let .failure(error):
+          print("🔴 [KsApi] ApolloClient query failure - error : \((error as NSError).description)")
```
singhhari, line 30: line 21 and 24 handle the error returned from the API. This failure case is run when there's a network failure or something similar.

msadoon (approved):
yea so good call removing the "JSON Decoding Error" prefix to the error displayed to user.
I think that was confusing to the user, but the rest of the error is relatively simple to understand. There is some internal iOS error lingo, but that's fine.
Here's the error I was getting on multiple failed Charles calls':


Also I was having the same issues regarding Charles not allowing network requests/breakpoints to get hit. I just [reset](https://www.detroitlabs.com/blog/2018/05/01/how-to-set-up-charles-proxy-for-an-ios-simulator/) the simulator/certs and Charles environment and things were working again.

PR #1588, opened 2021-08-18
[NT-2186] Upgrading dependency version for Braze.

singhhari:
# 📲 What

iOS 15 is introducing new breaking changes for Braze and push notifications. Braze has emailed us that we should upgrade to their latest SDK to make sure we are able to support the new iOS 15 features.

# 🛠 How

Updated `Cartfile` and `Cartfile.resolved`.

# ✅ Acceptance criteria

- [x] Run `carthage bootstrap --platform iOS` and run the app.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -19,7 +19,7 @@ github "appboy/appboy-segment-ios" == 4.0.0
 
 ### Binaries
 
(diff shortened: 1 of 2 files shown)


msadoon (approved):
good to go!

PR #1590, opened 2021-08-19
[NT-2160] Switching error envelope type in failure block of ApolloClient+RAC

singhhari:
# 📲 What

We're switching the type of `ErrorEnvelope` being send in the `failure` block of our ApolloClients `perform` and `fetch` method. The reason we're switching these is so the error message the user is seeing is human readable.

# 🤔 Why

The new error type defaults to a string of "Something went wrong." when we catch an obscure error. This is far more useful to the user than what our API or system may occasionally send back.

# ✅ Acceptance criteria

- [x] Create a new account and rapidly tap "re-send verification e-mail" in the Account settings. This will invoke a 429 http status code error which should trigger the `failure` block of our ApolloClient+RAC methods. The error message should return "Something went wrong."
- [x] Try this out with a few other queries or mutations.

### DIFF
--- a/KsApi/extensions/ApolloClient+RAC.swift
+++ b/KsApi/extensions/ApolloClient+RAC.swift
@@ -28,7 +28,7 @@ extension ApolloClient {
           observer.sendCompleted()
         case let .failure(error):
           print("🔴 [KsApi] ApolloClient query failure - error : \((error as NSError).description)")
-          observer.send(error: .couldNotDecodeJSON(error))
+          observer.send(error: .couldNotParseJSON)
         }
       }
     }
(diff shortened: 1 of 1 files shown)


KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -28,7 +28,7 @@ extension ApolloClient {
           observer.sendCompleted()
         case let .failure(error):
           print("🔴 [KsApi] ApolloClient query failure - error : \((error as NSError).description)")
-          observer.send(error: .couldNotDecodeJSON(error))
+          observer.send(error: .couldNotParseJSON)
```
singhhari, line 31: If you look at the implementation for this error envelope, the error message is an empty array. We have an extension on ErrorEnvelope on the localizedDescription property that will check if this is empty and if it is, returns 
 the localized string for "Something went wrong."

msadoon: and it's localized which is good.

msadoon (approved):
It's fine for now, but maybe as a more comprehensive fix for later, get a proper server side error returning for send verification email 429 error. (It may be a larger issue).
It's always good to let the user know what the actual issue is (in this case they are spamming the send verification email button, but it could also be their internet is not connected or a server side issue, problems with user-input fields, etc.)

PR #1591, opened 2021-08-20
[NT-1854] Remove Braze feature flag

singhhari:
# 📲 What

Removing the feature flag for Braze and all of its references.

# 🤔 Why

We are live with our Braze implementation and therefore no longer need to worry about toggling its state.

# ✅ Acceptance criteria

- [ ] Give the app a spin and verify we are building, passing tests and nothing was left behind.


msadoon (approved):
Ran tests locally, ran app, didn't see Braze feature flag under config, I did however see the red banner "We're doing things" on prod, not sure what that refers to.

singhhari:
> Ran tests locally, ran app, didn't see Braze feature flag under config, I did however see the red banner "We're doing things" on prod, not sure what that refers to.

Ah that is because @paulomcg has a test campaign set on staging!

PR #1592, opened 2021-08-23
[NT-2183] Segment identify call fix

singhhari:
# 📲 What

A previously introduced bug resulted in Segments identify method being called too many times. This was a result of our use of a `willSet` that fired each time `loggedInUser` was set. The fix wasn't as simple as checking if the older user == new user because of what appears to be a specific limitation of the SDK. We've re-factored where we call identify into specific locations in the application within this PR.

# 🤔 Why

The identify call is one of the core methods in the Segment SDK and we want to ensure we are not spamming the method and introducing noise into our analytics. By specifically calling identify in choice locations we are ensuring a more refined set of results being sent to Segment.

# 🛠 How

- Resetting the segmentClient in `AppEnvironment.logout`
- Calling `identify` in when the current user is updated in the `AppDelegate`
- Calling `identify` in when the current user is updated in the `SettingsNotificationsViewModel`

# ✅ Acceptance criteria

- [x] On staging, log in and verify `identify` is being sent to Segment.
- [x] On staging, while already logged in, launch the app and verify `identify` is being sent to Segment.
- [x] On staging, change your notifications within the settings and verify `identify` is being sent to Segment.
- [x] On staging, background the app, foreground it and verify `identify` is being sent to Segment.

### DIFF
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -82,6 +82,8 @@ public struct AppEnvironment: AppEnvironmentType {
     let storage = AppEnvironment.current.cookieStorage
     storage.cookies?.forEach(storage.deleteCookie)
 
+    // Resetting the segment client
+    AppEnvironment.current.ksrAnalytics.identify(newUser: nil)
     self.replaceCurrentEnvironment(
       apiService: AppEnvironment.current.apiService.logout(),
       cache: type(of: AppEnvironment.current.cache).init(),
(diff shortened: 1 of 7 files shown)


Library/AppEnvironment.swift
```diff
@@ -82,6 +82,8 @@ public struct AppEnvironment: AppEnvironmentType {
     let storage = AppEnvironment.current.cookieStorage
     storage.cookies?.forEach(storage.deleteCookie)
 
+    // Resetting the segment client
+    AppEnvironment.current.ksrAnalytics.identify(newUser: nil)
```
paulomcg, line 86: @singhhari have we done this previously? Curious about what's the outcome of this call 🤔

singhhari: Yeah, we were doing this before but it was baked into the behavior of the property observer on `loggedInUser`. If loggedInUser was not set we were passing `nil` into the `identify` method and that causes the client to reset the traits and user id.

So we're still doing this now but because I removed the property observer I now have to explicitly reset it by passing in nil on log out.

msadoon (approved):
Good to go!

PR #1596, opened 2021-08-27
NE-33 Deeplink, PN, In-App Message Categories and Subcategories by ID

msadoon:
# 📲 What

Allow deeplinking, pns and in-app messaging categories (and subcategories) by id.

# 🤔 Why

Marketing push to use relevant deeplinking in the app.

# 🛠 How

Refactored the parsing of categories to include names and ids. Essentially this came down to looking for matches against all categories and subcategories returned from the `fetchGraphCategories` endpoint. 

A match is found if the category id/name or subcategory id/name passed into the app and the results of the endpoint call are equal. If a match is found the user is routed.

Nothing changes in terms of braze pns/in-messaging, works out of the box.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Ensure [all test cases](https://github.com/kickstarter/ios-oss/pull/1593) in the category name only deeplinking/pn/in-messaging still pass.
- [x] Ensure Deeplinks/PNs/In-App messaging deeplink correctly for these urls:
[All Art Projects](https://www.kickstarter.com/discover/advanced?category_id=1&sort=magic&seed=2714369&page=1)
[Design-Civic Design](https://www.kickstarter.com/discover/advanced?category_id=259&sort=magic&seed=2715348&page=1)
   - [x] Ensure to test each of the above when the app is backgrounded and terminated.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
@@ -403,6 +403,27 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     self.findRedirectUrl = deepLinkUrl
       .filter { Navigation.match($0) == .emailClick }
 
+    let deepLinkCategories: ([String: String]) -> (Param?, Param?) = { rawParams in
+      let parentCategoryParams = rawParams["parent_category_id"]
+      let subCategoryParams = rawParams["category_id"]
+      var categoryParam: Param?
+      var subcategoryParam: Param?
+
+      switch (parentCategoryParams, subCategoryParams) {
+      case let (.some(rawCategoryParams), .some(rawSubcategoryParams)):
+        categoryParam = .some(Param.slug(rawCategoryParams))
+        subcategoryParam = .some(Param.slug(rawSubcategoryParams))
+      case (let .some(rawCategoryParams), nil):
+        categoryParam = .some(Param.slug(rawCategoryParams))
+      case (nil, let .some(rawSubcategoryParams)):
+        categoryParam = .some(Param.slug(rawSubcategoryParams))
+      case (nil, nil):
+        return (nil, nil)
+      }
+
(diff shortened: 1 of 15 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -403,27 +403,71 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     self.findRedirectUrl = deepLinkUrl
       .filter { Navigation.match($0) == .emailClick }
 
+    /// INFO: Handles the deeplink route with both an id and text based name for a deeplink for categories.
```
singhhari, line 406: Haven't seen us use this INFO prefix before so I think the /// with your inline doc is fine.

Kickstarter-iOS/ViewModels/AppDelegateViewModel.swift
```diff
@@ -403,27 +403,71 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
-      case let (.some(rawCategoryParams), .some(rawSubcategoryParams)):
-        categoryParam = .some(Param.slug(rawCategoryParams))
-        subcategoryParam = .some(Param.slug(rawSubcategoryParams))
-      case (let .some(rawCategoryParams), nil):
-        categoryParam = .some(Param.slug(rawCategoryParams))
-      case (nil, let .some(rawSubcategoryParams)):
-        categoryParam = .some(Param.slug(rawSubcategoryParams))
-      case (nil, nil):
-        return (nil, nil)
+      let rawId: (String?) -> Int? = { rawParam in
```
singhhari, line 413: This is quite a bit of logic for an inline closure. I still think this and the method below should be an explicit method declaration.

singhhari (approved):
lgtm left a small suggestion about the larger closures we have defined in the AppDelegateViewModel

PR #1597, opened 2021-08-31
[NTV-13] Supporting deep links to a specific reply.

singhhari:
# 📲 What

One of the final pieces to our deep link integration with comment threading is the ability to take a user to a specific reply. We're using the replyId as a query parameter and scrolling the table view to a particular index, if the replyId matches.

# 🤔 Why

This is part of our larger goal to drive engagement with deeplinks. This should be something we incorporate into all new feature releases. A good gameplan would probably be to release the feature, add analytics support and then finish it off with a supporting deeplink.

# 🛠 How

Most of the work involving deeplinks is done in two files: `AppDelegateViewModel.swift` and `Navigation.swift`. It's in these two files that we parse the route, determine the relevant query parameters and subsequently navigate the user to the correct point in the stack.

# ✅ Acceptance criteria

I used the following links to test:
Update: https://staging.kickstarter.com/projects/fjorden/fjorden-iphone-photography-reinvented/posts/3254626/comments?comment=Q29tbWVudC0zMzU2MTY4Ng%3D%3D&reply=Q29tbWVudC0zMzY1NTUxMQ%3D%3D
Comments: https://staging.kickstarter.com/projects/fjorden/fjorden-iphone-photography-reinvented/comments?comment=Q29tbWVudC0zMzYzMzAyOA%3D%3D&reply=Q29tbWVudC0zMzY0Nzg3Nw%3D%3D

- [x] Check while the app is backgrounded, terminated and in the foreground that a deeplink for both URLs responds appropriately.
- [x] Check while the app is backgrounded, terminated and in the foreground that a push notification for both URLs responds appropriately.

### DIFF
--- a/Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
+++ b/Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
@@ -739,6 +739,28 @@ final class AppDelegateViewModelTests: TestCase {
     }
   }
 
+  func testPresentViewController_ProjectCommentThread_Reply_Success() {
+    self.vm.inputs.applicationDidFinishLaunching(
+      application: UIApplication.shared,
+      launchOptions: [:]
+    )
+
+    withEnvironment(apiService: MockService(fetchCommentRepliesEnvelopeResult: .success(CommentRepliesEnvelope
+        .successfulRepliesTemplate), fetchProjectResponse: .template)) {
+      let url =
+        "https://\(AppEnvironment.current.apiService.serverConfig.webBaseUrl.host ?? "")/projects/fjorden/fjorden-iphone-photography-reinvented/comments?comment=Q29tbWVudC0zMzY0OTg0MQ%3D%3D&reply=deadbeef"
+
+      let result = self.vm.inputs.applicationOpenUrl(
+        application: UIApplication.shared,
+        url: URL(string: url)!,
+        options: [:]
+      )
+      XCTAssertTrue(result)
+
+      self.presentViewController.assertValues([3])
+    }
+  }
+
   func testPresentViewController_UpdateCommentThread_Success() {
     self.vm.inputs.applicationDidFinishLaunching(
       application: UIApplication.shared,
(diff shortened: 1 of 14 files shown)


Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
```diff
@@ -761,6 +783,28 @@ final class AppDelegateViewModelTests: TestCase {
...
+        "https://\(AppEnvironment.current.apiService.serverConfig.webBaseUrl.host ?? "")/projects/fjorden/fjorden-iphone-photography-reinvented/posts/3254626/comments?comment=Q29tbWVudC0zMzY0OTg0MQ%3D%3D&reply=deadbeef"
+
+      let result = self.vm.inputs.applicationOpenUrl(
+        application: UIApplication.shared,
+        url: URL(string: url)!,
+        options: [:]
+      )
+      XCTAssertTrue(result)
+
+      self.presentViewController.assertValues([4])
```
msadoon, line 804: might be nice to know what is in the stack too. (ie `CommentRepliesViewController`) here and above.

Kickstarter-iOS/Views/Controllers/CommentRepliesViewController.swift
```diff
@@ -114,6 +116,7 @@ final class CommentRepliesViewController: UITableViewController {
         guard let self = self else { return }
         self.dataSource.load(repliesAndTotalCount: repliesAndTotalCount, project: project)
         self.tableView.reloadData()
+        self.viewModel.inputs.loadRepliesAndProjectIntoDataSourceComplete()
```
msadoon, line 119: could also call this input `dataSourceLoaded`

msadoon (approved):
good to go!

Kickstarter-iOS/ViewModels/AppDelegateViewModelTests.swift
singhhari: Ah so the view model output is being mapped to the observer as a count of view controllers so I think this is the designed testing behavior.

PR #1599, opened 2021-09-02
[NT-2061] Apollo GraphQL Service and ViewModel integration

singhhari:
# 📲 What

This PR is handling the fairly large task of merging our feature branch into `main`. In particular, this feature branch contains 90 percent (there is a small amount that will be handled next cooldown) of the Apollo GraphQL integration. We are updating our `Service` objects and making any updates to ViewModels and their respective unit testing.

# 🤔 Why

This is a healthy step forward before we deprecate our old GraphSchema. That work is not done yet but will be handled in the near future. Apollos GraphQL client is much more readable + easier to use than our existing DSL.

# 🛠 How

We basically entirely recanvassed any use of GraphQL in our Service objects. We replaced almost all instances of our old GraphSchema approach to Apollo and slowly integrated them into the feature branch over the last two months.

# ✅ Acceptance criteria

- [ ] Ensure we're building and nothing unusual is occurring.


msadoon (approved):
We've been testing PR's merged into `NT-2061-feature-apollo-integration` every step of the way so there is no need to through this thoroughly. Tests passing on CI and build ready on alpha is good enough for me.
Great work! 🥳

PR #1602, opened 2021-09-08
[NTV-155] Optimizely experiment set up for checkout risk messaging

singhhari:
# 📲 What

This is the first piece of our new checkout risk messaging feature. We're adding an Optimizely experiment to wrap up the work, given how crucial checkout is to our application. For now, we're handling the `Pledge` and `Pay With Apple` with a check for `variant1`. Similarly, is `variant1` is coming through, we are not showing the existing risk messaging in checkout.

# 🤔 Why

The experiment enables us to control the rollout and potentially pull the feature if we see some strange conversions at checkout. We're going to incrementally merge this work into a feature branch and then release the Kraken when we are comfortable with starting the experiment in prod.

# 🛠 How

- Added a new Optimizely experiment -- `native_risk_messaging`.
- If `control` is coming back from Optimizely, there is no change from the existing behavior in prod.
- If `variant1` is coming back from Optimizely, we hide the risk messaging from checkout and tapping pledge/pay with apple will result in a modal presenting. This is in development and so for now we are just rendering a blank red view controller.

# 👀 See

| Variant1 | Control |
| --- | --- |
| https://user-images.githubusercontent.com/15615702/132556657-697c169a-63ed-445c-9a12-213241a3c929.mp4 | https://user-images.githubusercontent.com/15615702/132557013-c1dbad50-6e30-4dbe-b7d0-c2c2b06d456d.mp4 |

# ✅ Acceptance criteria

- [x] By default on staging, the `variant1` experiment is being broadcasted to everyone. Navigate to a project and verify the risk messaging is not appearing at checkout. Also verify that tapping pledge and pay with apple results in a red view controller being presented (placeholder).
- [x] Go to edit an existing pledge. Verify the risk messaging is not showing.
- [x] Now go to Optimizely and switch the experiment back to control (please revert this when done testing). Verify the behavior you're seeing aligns with prod in both of the situations above.


Kickstarter-iOS/Library/SharedFunctionsTests.swift
```diff
@@ -322,6 +322,30 @@ internal final class SharedFunctionsTests: TestCase {
...
+        OptimizelyExperiment.Key.nativeRiskMessaging.rawValue:
+          OptimizelyExperiment.Variant.control.rawValue
+      ]
+
+    withEnvironment(optimizelyClient: optimizelyClient) {
+      XCTAssertTrue(isNativeRiskMessagingControlEnabled())
+    }
+  }
+
+  func testIsNativeRiskMessagingControlEnabled_Variant1() {
```
msadoon, line 337: Title should be `Disabled`

Library/SharedFunctions.swift
```diff
@@ -257,6 +257,19 @@ public func discoveryPageBackgroundColor() -> UIColor {
   }
 }
 
+public func isNativeRiskMessagingControlEnabled() -> Bool {
+  guard let variant = AppEnvironment.current.optimizelyClient?
+    .variant(for: .nativeRiskMessaging)
+  else { return true }
+
+  switch variant {
+  case .control, .variant2:
```
msadoon, line 266: seems like we're only testing `control` and `variant1`, so can probably remove `variant2`?

msadoon (approved):
Looks good, few minor comments, otherwise working as intended.
I didn't check prod, just staging, but the behaviour should carry over.
Also one question: I guess the optimizely flags locally in the app is not a part of this epic, or is another story?

Library/SharedFunctions.swift
singhhari: Ah so if I remove that we have to include a default which is also just returning true so I didn't want to add those two extra lines. We follow a similar convention for the other experiment.

singhhari:
> Looks good, few minor comments, otherwise working as intended.
> I didn't check prod, just staging, but the behaviour should carry over.
> Also one question: I guess the optimizely flags locally in the app is not a part of this epic, or is another story?

So we technically aren't using a feature flag this time. The experiment looks and acts like a flag but is technically different. I'm adding the experiment into this ticket here and the rest of the stories of the epic will be building out design and actions.

PR #1604, opened 2021-09-13
[NTV-157] Modal design for checkout risk messaging

singhhari:
# 📲 What

This PR includes the new view controller that will be handling the checkout risk messaging. No logic has been added yet and all we've included are the designs, which include a few labels and a button. Over the course of the next week we will integrate the logic and iron out the UI/UX flow.

# 🤔 Why

The new designs are a necessary step in the right direction for improving the trust our backers have in us. By having a very obvious checkout flow that includes this risk messaging we are informing backers or the risks and potential for failure in fulfillment.

# 🛠 How

Added the `RiskMessagingViewController` and presenting it from the `PledgeViewController`. We are setting the height constraint to be around 35 percent of the screen unless the text set on the device is particularly large.

# 👀 See

Simulator Screen Shot - iPhone 8 - 2021-09-13 at 15 06 32

# ✅ Acceptance criteria

- [x] On staging, navigate to the checkout of a project and in both Pledge and Pay With Apple you should be presented the risk messaging view controller, once tapped.
- [x] Tap to dismiss should work.
- [x] Test out on a few different simulators
- [x] Open up the accessibility inspector and increase the font (Paulo and I agreed that once fonts get VERY large we cant support the entirety of the screen so we will probably work on that in the future or find a solution that is as optimal as possible)


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -7,6 +7,7 @@ import UIKit
 private enum Layout {
   enum Style {
     static let cornerRadius: CGFloat = Styles.grid(2)
+    static let modalHeightMultiplier: CGFloat = 0.65
```
singhhari, line 10: This really only applies to users with the default sized font. If you have larger fonts in use the view controller height will expand up to a certain point.

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
```diff
@@ -2,6 +2,10 @@ import Library
 import Prelude
 import UIKit
 
+private enum SheetOverlayViewControllerStyles {
+  static let topAnchorMargin: CGFloat = 65
```
singhhari, line 6: This is used to add a buffer at the top of the screen to enable tapping to dismiss even when the font is quite large.

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
```diff
@@ -48,6 +52,12 @@ final class SheetOverlayViewController: UIViewController {
       |> \.backgroundColor .~ UIColor.ksr_support_700.withAlphaComponent(0.8)
   }
 
+  override func touchesBegan(_ touches: Set<UITouch>, with _: UIEvent?) {
```
singhhari, line 55: This function was added in here to enable tapping to dismiss. I looked at the other places we use this VC and didn't see any issues.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Views/Controllers/RiskMessagingViewController.swift
```diff
@@ -0,0 +1,201 @@
...
+
+final class RiskMessagingViewController: UIViewController {
+  // MARK: - Properties
+
+  private lazy var bannerImageView: UIImageView = { UIImageView(frame: .zero) }()
+  private lazy var bannerImageStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var bannerLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var cartIconImageView: UIImageView = { UIImageView(frame: .zero) }()
+  private lazy var confirmButton: UIButton = { UIButton(frame: .zero) }()
+  private lazy var footnoteLabel: UILabel = { UILabel(frame: .zero) }()
```
msadoon, line 14: Minor note: the designs make this look like a link so we may want to change it to a button.

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
msadoon: I would add a comment here to say that `/// Allows tap to dismiss`, just because a few years from now it may not be clear as to intention, also helps our documentation.

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
```diff
@@ -80,11 +90,16 @@ final class SheetOverlayViewController: UIViewController {
     NSLayoutConstraint.activate([
       childView.centerXAnchor.constraint(equalTo: self.view.centerXAnchor),
       childView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor),
-      childView.widthAnchor.constraint(equalToConstant: portraitWidth)
+      childView.widthAnchor.constraint(equalToConstant: portraitWidth),
+      childView.topAnchor
+        .constraint(
+          greaterThanOrEqualTo: self.view.topAnchor,
```
msadoon, line 96: does this mean that the child view's top anchor can technically be higher than the overlay view it's inside?

msadoon (approved):
Great job on this, the ui code is clean and looks great on ipad and iphone.
Also tests cover landscape and portrait!
I found one thing on the largest iPad (12.9 inch iOS 14.5) I leave it to you if you want address it in this PR.

Kickstarter-iOS/Views/Controllers/SheetOverlayViewController.swift
singhhari: Greater than or equal to, in this case, implies that the childs top anchor will be "lower" on the screen than the parent. As the offset increases in number, the top anchor gets lower in the frame.

singhhari:
> Great job on this, the ui code is clean and looks great on ipad and iphone.
> Also tests cover landscape and portrait!
> I found one thing on the largest iPad (12.9 inch iOS 14.5) I leave it to you if you want address it in this PR.
> 

Went with a solution that doesn't entirely resolve the ipad issues but makes it a bit easier on the eyes. A form sheet is presented with a static height and so we will probably have empty space unless the font is quite large because the copy isn't very lengthy.

PR #1606, opened 2021-09-22
[NTV-158] Risk Messaging: Confirm Button Functionality and HelpWebViewController support

singhhari:
# 📲 What

In this PR we're ironing out the details with the `RiskMessagingViewController`. Specifically, we're ensuring that tapping the `I Understand` button leads to the correct series of events in the UI/UX for both a standard pledge and paying with Apple. For a standard pledge, we want the confirm button to sending a backing request, dismiss the `RiskMessagingViewController`, displaying our processing loading indicator and finally present the `ThanksViewController`. Similarly, when paying with Apple we want to present the Apply Pay authorization view controller after tapping `I Understand`. After confirming the payment on the Native form sheet we want to process the backing and show Thanks.

# 🤔 Why

It's crucial this experience is smooth because this is where checkouts occur. Any issues here could lead to a decline in backings which would not be healthy for Kickstarters mobile business. We're handling errors and also ensuring the UI/UX is a smooth, understandable experience.

# 🛠 How

A lot of work occurred in one of the more complex files of the app: `PledgeViewModel`. The `RiskMessagingViewController` uses this same view model and it's quite necessary given both the `control` and `variant` support for our Optimizely experiment.

# ✅ Acceptance criteria

- [x] On staging, with the `control` returning on Optimizely (alternatively return `false` from our helper method), ensure the behavior for both standard pledging and paying with Apple is the same as `main`.
- [x] On staging, with the `variant` returning on Optimizely (should be the default as of right now), ensure the standard pledge follows these steps:
- Tap pledge and `RiskMessagingViewController` should present
- Tapping `I Understand` should process a pledge successfully
- The `ThanksViewController` should be presented after processing the backing on /graph
- [x] On staging, with the `variant` returning on Optimizely (should be the default as of right now), ensure that Pay With Apple follows these steps:
- Tap pledge and `RiskMessagingViewController` should present
- Tapping `I Understand` should lead to the Apple Pay Authorization view controller being presented
- Tapping `Pay with Passcode` will lead to a backing request occurring
- The `ThanksViewController` should be presented after processing completes.
- [x] Test that the error banner is appearing when the request is interrupted (I used Breakpoint Settings on Charles)


Kickstarter-iOS/Views/Controllers/PledgeViewController.swift
```diff
@@ -479,8 +478,9 @@ final class PledgeViewController: UIViewController,
     self.present(paymentAuthorizationViewController, animated: true)
   }
 
-  private func goToRiskMessagingModal() {
-    let viewController = RiskMessagingViewController()
+  private func goToRiskMessagingModal(isApplePay: Bool) {
+    self.viewModel.inputs.configureRiskMessagingModal(isApplePay: isApplePay)
+    let viewController = RiskMessagingViewController(viewModel: self.viewModel)
```
msadoon, line 483: Yea not 100% about this approach. Usually, view models are contained within their own view controllers. You can pass _in_ data that can be used to create a new view model within `RiskMessagingViewController` (ie. `RiskMessagingViewModel`), but passing in `PledgeViewModel` to `RiskMessageViewController` directly seems a bit hacky to me.

I would suggest creating a new data type struct/enum similar to `ManagePledgeViewParamConfigData` that can be passed around, I see a lot of examples like that in the app.

A view model should be privately held by its view not accessible outside of it, separation of concerns.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -54,6 +55,10 @@ public protocol PledgeViewModelInputs {
   func paymentAuthorizationViewControllerDidFinish()
   func pledgeAmountViewControllerDidUpdate(with data: PledgeAmountData)
   func pledgeDisclaimerViewDidTapLearnMore()
+  func riskMessagingFootnoteLabelTapped()
+  func riskMessagingModalDismissedForApplePay()
+  func riskMessagingModalDismissedForStandardPledge()
+  func riskMessagingPledgeConfirmationButtonTapped()
```
msadoon, line 61: Yea so this is why we don't want to re-use the `PledgeViewModelType` for `RiskMessagingViewController`, because we're now extending the responsibilities of the `PledgeViewModel` to include actions specific to the `RiskMessagingViewController`, when previously all that `PledgeViewModel` cared about was `PledgeViewController`. Similarly `PledgeViewController` now has a bunch of inputs/outputs it does not need.

msadoon (changes requested):
I didn't review the functionality yet because there is something that's kind of a show-stopper in terms of architecture, but I'll keep reviewing once we've made some changes. Tell me if you agree/disagree, I may not be seeing the whole picture.
I can also definitely go deeper into it and we can figure it out together on a call.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -737,8 +778,30 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      .filter { _, isApplePay in
+        isApplePay && !isNativeRiskMessagingControlEnabled()
+      }
+      .ignoreValues()
+
+    let goToThanksTrigger = Signal.merge(
+      createBackingCompletionEvents.filter(isNativeRiskMessagingControlEnabled), // Control
+      self.riskMessagingModalDismissedForStandardPledgeProperty.signal, // Variant support for Pledge
+      createBackingCompleteAndIsApplePay // Variant support for Apple Pay
+    )
```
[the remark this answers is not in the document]
msadoon: So I have a commit that removes this logic and addresses this [comment](https://github.com/kickstarter/ios-oss/pull/1606/files#diff-8c79e81d24641ca4e7c3f635d05a4e4b8d63bf89ae31607f4a4ba2db1bf285a1R801).
Functionality and test cases pass exactly as before. Let me know if you want me to push it.

Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -1357,7 +1443,7 @@ final class PledgeViewModelTests: TestCase {
     }
   }
 
-  func testOutputGoToRiskMessagingModal_OptimizelyClientVariant1() {
+  func testOutputGoToRiskMessagingModal_OptimizelyClientVariant1_StandardPledge() {
```
msadoon, line 1446: Shouldn't this be `Control` because variant one means risk messaging would show.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/PledgeViewModel.swift:397:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

msadoon (changes requested):
So as per our discussion, we're gonna pass in the `PledgeViewModelType` into `RiskMessagingViewController` because this is an experiment.

Just basically correct the tests and address this one [comment](https://github.com/kickstarter/ios-oss/pull/1606/files#diff-8c79e81d24641ca4e7c3f635d05a4e4b8d63bf89ae31607f4a4ba2db1bf285a1R493) and its' good to go.
Even after my commit I tested in-app logic for control and variant1 and things we working as expected, but you might want to run your tests.

Good job on the tests!

Library/ViewModels/PledgeViewModelTests.swift
singhhari: The variant is set here. I think the assertion at the bottom may have confused you because its checking for `false` here:
`self.goToRiskMessagingModal.assertValue(false)`


I'll add some documentation to the code but here the `false` implies it is a standard pledge and `true` would mean it is Pay With Apple.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -448,11 +474,24 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
 
-    self.goToRiskMessagingModal = self.submitButtonTappedSignal
-      .filter { _ in !isNativeRiskMessagingControlEnabled() }
+    // Signal for nativeRiskMessaging experiment variant type
+    // The mapConst here is sending a Bool representing whether this is Apple Pay
+    let submitOrApplePayButtonTapped = Signal.merge(
+      self.submitButtonTappedSignal.mapConst(false),
+      self.applePayButtonTappedSignal.mapConst(true)
+    )
+    .filter { _ in !isNativeRiskMessagingControlEnabled() }
```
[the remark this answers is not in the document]
singhhari: Yeah I can see why this is confused, sorry about the lack of documentation here. The `Bool` going to the `goToRiskMessagingModel` is then used in the body of the `observeValues` closure to pass as an argument to an input. That input will then emit to a specific signal based on whether it is control/variant. and true/false.

singhhari:
> So as per our discussion, we're gonna pass in the `PledgeViewModelType` into `RiskMessagingViewController` because this is an experiment.
> 
> Just basically correct the tests and address this one [comment](https://github.com/kickstarter/ios-oss/pull/1606/files#diff-8c79e81d24641ca4e7c3f635d05a4e4b8d63bf89ae31607f4a4ba2db1bf285a1R493) and its' good to go.
> Even after my commit I tested in-app logic for control and variant1 and things we working as expected, but you might want to run your tests.
> 
> Good job on the tests!

Fixed the tests, addressed the comments and added some docs. Going to take some time tomorrow to determine if there is a viable alternative approach but g2g as is.

msadoon (approved):
good to go!

PR #1610, opened 2021-09-24
[NTV-156] Adding Segment analytics events for Risk Messaging confirm button

singhhari:
# 📲 What

Adding a `CTA Clicked` analytics event for whenever the "I Understand" button is tapped in the `RiskMessagingViewController`.

# 🤔 Why

It's important for us to monitor the effectiveness of our experiment. Given the proximity to the point of checkout, the analytics will give us an understanding if the risk messaging is leading to revenue drops or anything else unexpected.

# 🛠 How

- Adding a new event to `KSRAnalytics`
- Updating the `PledgeViewModel` to account for the `RiskMessagingViewController`s `confirmButton` being tapped.
- Updating some older events to respond specifically to the `submitButtonTappedSignal` and `applePayButtonTappedSignal`.

# ✅ Acceptance criteria

- [x] On staging, while you have the Segment debugger open, go through the checkout flow for a given project.
- [x] When the `RiskMessagingViewController` has been presented, before moving forward, verify that when you tap either `Pledge` or `Pay With Apple` that a `CTA Clicked` event with a `context_cta` of `pledge_submit` has emitted.
- [x] After tapping `I Understand`, verify a `CTA Clicked` event with a `context_cta` of `pledge_confirm` has emitted.

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -952,7 +952,6 @@ public final class KSRAnalytics {
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(checkoutProperties(from: checkoutData, and: reward))
-      // the context is always "newPledge" for this event
       .withAllValuesFrom(contextProperties(
         ctaContext: .pledgeConfirm,
         page: .checkout,
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -147,6 +147,7 @@ public final class KSRAnalytics {
     case logInInitiate
     case logInOrSignUp
     case logInSubmit
+    case pledgeConfirm
     case pledgeInitiate
     case pledgeSubmit
     case project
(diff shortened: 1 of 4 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -930,6 +932,40 @@ public final class KSRAnalytics {
...
+  public func trackPledgeConfirmButtonClicked(
+    project: Project,
+    reward: Reward,
+    typeContext: TypeContext,
+    checkoutData: CheckoutPropertiesData,
+    refTag: RefTag?
+  ) {
+    let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
+      .withAllValuesFrom(checkoutProperties(from: checkoutData, and: reward))
+      // the context is always "newPledge" for this event
```
msadoon, line 955: but the context passed in is `pledgeConfirm`

msadoon (approved):
good to go!

Library/Tracking/KSRAnalytics.swift
singhhari: Good catch. Updated.

PR #1611, opened 2021-09-24
[Unticketed] Put Back Add Ons Selection To Checkout

msadoon:
# 📲 What  and 🤔 Why

While testing, @paulomcg found that we are missing the add-ons selection step in the checkout process. This is because the `Reward` object was always passing in `false` to the `hasAddons` property of `Reward` model.

It wasn't completely within the scope of `Project` model, but due to relationships created previously through `v1/projects` endpoint, this flag is now important to both show the **Addons** pill and guide the user through the **Addons** step if the reward has addons.

# 🛠 How

Added this property to the `RewardFragment` and adjusted the `Rewards+RewardsFragment` to check if there are nodes and `hasAddons` will be `true` if its not empty.

```
allowedAddons {
  __typename
      nodes {
         __typename
         id
   }
}
```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |   |

# ✅ Acceptance criteria

- [x] Ensure a project with addons ([example](https://www.kickstarter.com/projects/fjorden/fjorden-iphone-photography-reinvented)) allows the user to checkout with addons on a reward.

### DIFF
--- a/KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
+++ b/KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
@@ -25,6 +25,7 @@ final class Backing_BackingFragmentTests: XCTestCase {
       XCTAssertEqual(backing?.projectCountry, "US")
       XCTAssertEqual(backing?.projectId, 1_596_594_463)
       XCTAssertNotNil(backing?.reward)
+      XCTAssertTrue(backing!.reward!.hasAddOns)
       XCTAssertEqual(backing?.rewardId, decompose(id: "UmV3YXJkLTgxNzM5MDE="))
       XCTAssertEqual(backing?.sequence, 148)
       XCTAssertEqual(backing?.shippingAmount, 10.0)
@@ -67,6 +68,10 @@ private func backingDictionary() -> [String: Any] {
             "amount": "30.0",
             "currency": "USD",
             "symbol": "$"
+          },
+          "allowedAddons": {
+             "__typename": "RewardConnection",
+             "nodes": []
           },
           "backersCount": 2,
           "convertedAmount": {
@@ -144,6 +149,10 @@ private func backingDictionary() -> [String: Any] {
             "currency": "USD",
             "symbol": "$"
           },
+          "allowedAddons": {
+            "__typename": "RewardConnection",
+            "nodes": []
+          },
           "backersCount": 23,
           "convertedAmount": {
             "__typename": "Money",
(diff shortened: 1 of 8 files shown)


KsApi/models/graphql/adapters/Backing+BackingFragmentTests.swift
```diff
@@ -25,6 +25,7 @@ final class Backing_BackingFragmentTests: XCTestCase {
       XCTAssertEqual(backing?.projectCountry, "US")
       XCTAssertEqual(backing?.projectId, 1_596_594_463)
       XCTAssertNotNil(backing?.reward)
+      XCTAssertTrue(backing!.reward!.hasAddOns)
```
singhhari, line 28: we should avoid implicit optionals whenever possible.

KsApi/models/graphql/adapters/Project+FetchProjectQueryDataTests.swift
```diff
@@ -253,6 +253,8 @@ final class Project_FetchProjectQueryDataTests: XCTestCase {
     )
     XCTAssertNil(lastReward.endsAt)
     XCTAssertFalse(lastReward.hasAddOns)
+    XCTAssertTrue(project.rewards.first!.hasAddOns)
```
singhhari, line 256: same here.

KsApi/models/graphql/adapters/Reward+RewardFragment.swift
```diff
@@ -26,14 +26,16 @@ extension Reward {
     let estimatedDeliveryOn = rewardFragment.estimatedDeliveryOn
       .flatMap(dateFormatter.date(from:))?.timeIntervalSince1970
 
+    let rewardHasNoAddons = rewardFragment.allowedAddons.nodes?.isEmpty ?? true
```
singhhari, line 29: I think this would read better as an argument below if it was `rewardHasAddOns` (invert logic) but that is just a personal style preference.

KsApi/models/graphql/adapters/Reward+RewardFragmentTests.swift
```diff
@@ -27,6 +27,7 @@ final class Reward_RewardFragmentTests: XCTestCase {
       XCTAssertEqual(v1Reward?.limit, nil)
       XCTAssertEqual(v1Reward?.limitPerBacker, 1)
       XCTAssertEqual(v1Reward?.minimum, 25.0)
+      XCTAssertTrue(v1Reward!.hasAddOns)
```
singhhari, line 30: implicit optional

singhhari (approved):
left a few small remarks but lgtm!

PR #1612, opened 2021-09-27
[NTV-156] Fixing pledge_confirm events apple_pay/credit_card property values

singhhari:
# 📲 What

This is a quick fix to correct an issue with our `pledge_confirm` event. We were previously sending a value of `credit_card` for all payment types whether is was `Pay With Apple` or not and this PR will correct that.

# 🤔 Why

Our insights team needs accurate data in order to correctly interpret and take action on our analytics data.

# 🛠 How

Combining an additional signal to the `pledgeSubmitEventsSignal` in the `PledgeViewModel`.

# ✅ Acceptance criteria

- [x] Successfully pledge a project, on Staging, using `Pay With Apple` and verify the `CTA Clicked` event that is emitted has a `context_cta` of `pledge_confirm` and a `context_type` of `apple_pay`.
- [x] - [ ] Successfully pledge a project, on Staging, using `Pledge` and verify the `CTA Clicked` event that is emitted has a `context_cta` of `pledge_confirm` and a `context_type` of `credit_card`.

### DIFF
--- a/Library/ViewModels/PledgeViewModel.swift
+++ b/Library/ViewModels/PledgeViewModel.swift
@@ -946,8 +946,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
 
     // Risk Messaging Modal pledge_confirm event
     pledgeSubmitEventsSignal
-      .combineLatest(with: self.riskMessagingViewControllerDismissedProperty.signal.skipNil())
-      .takeWhen(self.riskMessagingViewControllerDismissedProperty.signal)
+      .takePairWhen(self.riskMessagingViewControllerDismissedProperty.signal.skipNil())
       .map { pledgeSubmitEvent, isApplePay in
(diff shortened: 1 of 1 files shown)


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -946,8 +946,10 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
 
     // Risk Messaging Modal pledge_confirm event
     pledgeSubmitEventsSignal
+      .combineLatest(with: self.riskMessagingViewControllerDismissedProperty.signal.skipNil())
       .takeWhen(self.riskMessagingViewControllerDismissedProperty.signal)
```
msadoon, line 950: Instead of using both `combineLatest` and `takeWhen`, can you use replace both with `.takePairWhen(self.riskMessagingViewControllerDismissedProperty.signal.skipNil())`?

msadoon (changes requested):
One minor change please

msadoon (approved):
good to go!

PR #1615, opened 2021-09-28
[NTV-194][NTV 222] New Project Page Skeleton

msadoon:
# 📲 What

A skeleton of the new project page based heavily on the old project page minus now unsupported functionality.
Effectively trying to recreate the new project page product initiative to increase user engagement.

# 🤔 Why

Trying to increase user engagement with the project page, surface important information that was previously not clearly accessible. 
Move away from v1 project data to GQL project data.

Unblock creation of the navigation selection in the paging view controller.

# 🛠 How

The new `ProjectPageViewController` and `ProjectPageViewModel` move away from using the side-side pagination functionality supported by `ProjectPamphletViewController`.

The `ProjectPageViewController` also does not use any storyboard components, and the navigation is built inside the view controller itself, borrowing heavily from `ProjectNavBarController` (previously used in `ProjectPamphletViewController`).

Only one entry point at this phase from `DiscoveryPageViewController` (basically the main project list when the app is launched.)

**Unrelated Changes**
- Removed a duplicate datasource initializer in `CommentsViewController`
- `ProfileViewModel` and `ProfileViewModelTests` were not being accessed anywhere in the app (ran a search on `ProfileViewModel` and it wasn't being called anyway) -- deleted.
- `ShareViewModel` `firstPartyShareTypes` not being used -- deleted

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] One entry point for testing until the Optimizely flag updates all the entry points for ProjectPamphletViewController or new project page.
- [ ] Navigation bar in code constraints. Works as expected/before.
- [ ] Pledge CTA view/behavior remains unchanged.
- [ ] MessagingBannerViewController view/behavior remains unchanged.
- [ ] Landscape/portrait still supported.

# ⏰ TODO

- [ ] Write tests.


Kickstarter-iOS/Views/Controllers/CommentsViewController.swift
```diff
@@ -62,7 +62,6 @@ internal final class CommentsViewController: UITableViewController {
 
     self.commentComposer.delegate = self
 
-    self.tableView.dataSource = self.dataSource
```
singhhari, line 65: I don't think this line needs to be removed? Putting it back in did not break anything from my observation.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -362,17 +362,31 @@ internal final class DiscoveryPageViewController: UITableViewController {
...
   }
 
-  fileprivate func goTo(project: Project, initialPlaylist: [Project], refTag: RefTag) {
-    let vc = ProjectNavigatorViewController.configuredWith(
-      project: project,
-      refTag: refTag,
-      initialPlaylist: initialPlaylist,
-      navigatorDelegate: self
+  fileprivate func goTo(project: Project, initialPlaylist _: [Project], refTag: RefTag) {
+    /** FIXME: This is the only entry point to the new project page, as we build it out and get it fully functional we'll need to refactor entry points in `ProjectNavigatorPagesDataSource` and `ProjectNavigatorViewController` (basically search for places where `ProjectPamphletViewController` is used in the app.)
```
singhhari, line 366: Cool, assuming we need to insert the Optimizely feature flag logic into here?

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
```diff
@@ -362,17 +362,31 @@ internal final class DiscoveryPageViewController: UITableViewController {
...
+     )
+
+     if UIDevice.current.userInterfaceIdiom == .pad {
+        vc.modalPresentationStyle = .fullScreen
+     }
+
+     self.present(vc, animated: true, completion: nil)
+     */
+
+    let projectParam = Either<Project, Param>(left: project)
```
singhhari, line 381: Could you explain why you're using `Either` here? Looks like you are initializing it with the left and the right is not set up to do anything in particular? 

My understanding of this is that there is just one entry in and a `Project` is absolutely necessary to set up the views.

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+public final class ProjectPageViewController: UIViewController, MessageBannerViewControllerPresenting {
+  // MARK: Properties
+
+  private enum NavigationButtonSizes: CGFloat {
```
singhhari, line 9: It looks like the convention we use everywhere else in the app follows this:

`private enum Layout {
  enum Button {
    static let height: CGFloat = 34
  }
`

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+  }
+
+  private let viewModel: ProjectPageViewModelType = ProjectPageViewModel()
+  private let shareViewModel: ShareViewModelType = ShareViewModel()
+  private let watchProjectViewModel: WatchProjectViewModelType = WatchProjectViewModel()
+  /**
+   FIXME: This `contentController` can be renamed `contentViewController` and has to be embedded in a `PagingViewController` in https://kickstarter.atlassian.net/browse/NTV-195
+   Maybe check `BackerDashboardViewController`'s pageViewController for in-app examples on how to do this.
+   */
+  private var contentController: ProjectPamphletContentViewController?
```
singhhari, line 20: yeah let's name this `contentViewController` so it matches the object it's holding the value of.

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+
+  // MARK: User session state properties
+
+  private var sessionEndedObserver: Any?
```
singhhari, line 24: Not sure what this is used for yet but I'm assuming this was carried over from the original implementation?

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+
+  public var messageBannerViewController: MessageBannerViewController?
+
+  private lazy var navigationShareButton: UIBarButtonItem = {
+    let contentView = UIButton()
+      |> shareButtonStyle
+      |> UIButton.lens.imageEdgeInsets .~ UIEdgeInsets(left: -NavigationButtonSizes.spacing.rawValue)
+      |> UIButton.lens.accessibilityLabel %~ { _ in Strings.dashboard_accessibility_label_share_project() }
+
+    contentView.addTarget(
```
singhhari, line 35: Most of the time I see the styling done in a `bindStyles` method and the `addTarget` being done in `viewDidLoad`. I don't know if we also do it like this but I bet it's for simpler initialization and easier readability up top.

Kickstarter-iOS/Views/Controllers/DiscoveryPageViewController.swift
singhhari: Ah it looks like this is the original implementation based on some code below?

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+
+    return vc
+  }
+
+  public override func viewDidLoad() {
+    super.viewDidLoad()
+
+    self.configurePledgeCTAContainerView()
+
+    /** FIXME:  - https://kickstarter.atlassian.net/browse/NTV-195
```
singhhari, line 100: Assuming this is where we plug in our new page view controller and the first screen that loads by default?

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+
+    self.viewModel.outputs.popToRootViewController
+      .observeForControllerAction()
+      .observeValues { [weak self] in
+        self?.navigationController?.popToRootViewController(animated: false)
+      }
+  }
+
+  private func bindSharingViewModel() {
+    self.shareViewModel.outputs.showShareSheet
```
singhhari, line 245: Is the `ProjectNavBarViewController`and view model not reusable with delegation to this view controller? A little concerned about the bloat of three view models in one view controller.

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -0,0 +1,415 @@
...
+      .observeValues { project in
+        NotificationCenter.default.post(
+          name: Notification.Name.ksr_projectSaved,
+          object: nil,
+          userInfo: ["project": project]
+        )
+      }
+  }
+
+  fileprivate func showProjectStarredPrompt() {
```
singhhari, line 290: should we move this to `SharedFunctions` since we are re-declaring this in two separate files?

singhhari: Wondering the same for the methods below as well.

singhhari: We will probably be deleting the old one after this is shipped though so let me know your thoughts.

Library/ViewModels/ProfileViewModel.swift
```diff
@@ -1,146 +0,0 @@
-import Foundation
-import KsApi
-import Prelude
-import ReactiveExtensions
-import ReactiveSwift
-
-public protocol ProfileViewModelInputs {
```
singhhari, line 7: Not sure if we wanna delete any files just yet since we are going to be toggling between versions using the flag?

Library/ViewModels/ShareViewModel.swift
```diff
@@ -21,13 +21,6 @@ public struct ShareActivityCompletionData {
   }
 }
 
-/// These share types provide us access to knowing when the user successfully shares through that method,
-/// or when the user cancels.
-private let firstPartyShareTypes: [UIActivity.ActivityType] = [
```
singhhari, line 26: Why was this snippet deleted specifically?

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
msadoon: Yea good call, can probably remove these and use `NotificationCenter.default` just like `didBackProject`

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
msadoon: Yep!

Library/ViewModels/ShareViewModel.swift
msadoon: Wasn't being used, its private so it cannot be used anywhere but inside this file. Searched for references in this file and there were none.

Library/ViewModels/ProfileViewModel.swift
msadoon: This `ProfileViewModel` and its' associated tests is not used anywhere in the app. I did a search for `ProfileViewModel` and `ProfileViewModelTests` in the app and it did not show anywhere in the entire project except in its own declaration.
I found this because I was looking for how `BackDashboardViewController` was implementing its' navigation bar, which is similar to the one we want in this page.

I believe its safe to delete, if you think its still being used somewhere let's keep it in? What do you think? I can make a separate PR for this as well :)

PR #1616, opened 2021-09-29
[Unticketed] No Rewards Reward Added To Rewards Selection Screen

msadoon:
# 📲 What

Last minute bug fix for release `4.15.0`.

# 🤔 Why

With 3/4 % of pledges coming from this reward tier, its' important we support it in every release.

# 🛠 How

Pretty straightforward, thanks @Arkariang for the tip of making the project minimum pledge the minimum of the no-reward pledge. 
- Added `minPledge` to `ProjectFragment` under `extendedProjectProperties` because it wasn't in the v1 model.

Also updated the `Project+FetchProjectQueryData.swift` file to include the `minimum` and `convertedMinimum` fields of the no reward reward.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Adds no rewards reward to reward selection screen.

### DIFF
--- a/KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
+++ b/KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
@@ -55,27 +55,8 @@ extension Project {
         Reward.reward(from: fragment)
       } ?? []
 
-    /** FIXME: This is unfortunately a consequence of the no-reward reward being returned on v1 but not in GQL. Eventually we'll want to talk with backend about the possibility of returning a no-reward reward as part of the project query, just as they did with v1. The benefit of that is no reward reward doesn't have to be maintained locally. We want to show the rewards that the backend returns without modification to the raw data.
-     */
-
-    var projectMinimumPledgeAmount = 1.0
-    var currentUsersCurrencyFXRate = 1.0
-
-    if let fxRateValue = data.project?.fragments.projectFragment.fxRate {
-      currentUsersCurrencyFXRate = Double(fxRateValue)
-    }
-
-    if let projectMinPledgeSingleTierRawValue = data.project?.fragments.projectFragment.minPledge {
-      projectMinimumPledgeAmount = Double(projectMinPledgeSingleTierRawValue)
-    }
-
-    let convertedMinimumAmount = currentUsersCurrencyFXRate * projectMinimumPledgeAmount
-
-    let emptyReward = Reward.noReward
-      |> Reward.lens.minimum .~ projectMinimumPledgeAmount
-      |> Reward.lens.convertedMinimum .~ convertedMinimumAmount
-
-    let updatedRewardsWithNoReward = [emptyReward] + rewards
+    let emptyRewards = [noRewardReward(from: data.project?.fragments.projectFragment)]
+    let updatedRewardsWithNoReward = emptyRewards + rewards
 
     var projectBackingId: Int?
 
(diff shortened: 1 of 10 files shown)


KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
```diff
@@ -54,6 +55,9 @@ extension Project {
         Reward.reward(from: fragment)
       } ?? []
 
+    let emptyRewards = [noRewardReward(from: data.project?.fragments.projectFragment)]
+    let updatedRewardsWithNoReward = emptyRewards + rewards
```
singhhari, line 59: I wonder if a better name for this constant would be `allRewards` or something similar.

KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
```diff
@@ -114,4 +121,28 @@ extension Project {
 
     return (project, projectBackingId)
   }
+
+  /** FIXME: This is unfortunately a consequence of the no-reward reward being returned on v1 but not in GQL. Eventually we'll want to talk with backend about the possibility of returning a no-reward reward as part of the project query, just as they did with v1. The benefit of that is no reward reward doesn't have to be maintained locally. We want to show the rewards that the backend returns without modification to the raw data.
+   */
+
+  private static func noRewardReward(from fragment: GraphAPI.ProjectFragment?) -> Reward {
+    var projectMinimumPledgeAmount = 1.0
```
singhhari, line 129: Could 129-138 be simplified with a nil-coalescing operator on one line for each of these variables?

singhhari: Also curious if the fxRate ever returns as nil? Seems like something that we should never be hard-coding a value for.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
```diff
@@ -228,12 +228,14 @@ private func extendedProject(from projectFragment: GraphAPI.ProjectFragment) ->
   let risks = projectFragment.risks
   let environmentalCommitments = extendedProjectEnvironmentalCommitments(from: projectFragment)
   let faqs = extendedProjectFAQs(from: projectFragment)
+  let minimumSingleTierPledgeAmount = projectFragment.minPledge
```
singhhari, line 231: What does single tier mean here?

msadoon: single tier is just a reward basically, just a fancy name for single `reward` (got it from the GQL Schema Docs for `fxRate` - `The min pledge amount for a single reward tier.`)

msadoon: Want to change it to `minimumRewardPledgeAmount`?

KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
msadoon: Good call with the `nil-coelesing`, added that. 
The `fxRate` and `minPledge` always have a value from GQL, never `null`, but I guess because we pass the project fragment here as an optional, there has to be some unwrapping. I could pass in an non-optional project fragment and wrap this around an `if-let` at the call-site if that's preffered.

setting the `fxRate` to 1 just means we return whatever the minimum pledge amount is for the reward in the currency of the project.

KsApi/models/graphql/adapters/Project+ProjectFragment.swift
singhhari: Nope, as long as it references something I think it's fine. I wasn't aware there were tiers until just now.

singhhari (approved):
lgtm

PR #1617, opened 2021-09-30
[NTV-158] No longer presenting risk messaging modal when updating pledge

singhhari:
# 📲 What

After merging our feature branch for risk messaging into `main` we realized we were presenting the risk messaging copy to backers even in the update pledge context. This PR will make it so that risk messaging is only shown when creating a backing and not when updating.

# 🤔 Why

The risk messaging is particularly important before the backer has initiated their pledge but not necessary to show after that. Updating a pledge typically involves either changing a reward or bonus amount and risk messaging does not need to be shown again within this context.

# 🛠 How

Updated the `PledgeViewModel` to check the `context` signal for an `isCreating` status before presenting risk messaging.

# ✅ Acceptance criteria

- [x] On staging, make a pledge using the green Pledge button. After that processes, manage the pledge and attempt to update the amount of the pledge. Risk Messaging should not be presented and there should be no regression.
- [x] On staging, make a pledge using the Pay With Apple button. After that processes, manage the pledge and attempt to update the amount of the pledge. Risk Messaging should not be presented and there should be no regression.

### DIFF
--- a/Library/ViewModels/PledgeViewModelTests.swift
+++ b/Library/ViewModels/PledgeViewModelTests.swift
@@ -2602,6 +2602,110 @@ final class PledgeViewModelTests: TestCase {
     }
   }
 
+  func testUpdateBacking_Success_OptimizelyClientVariant1() {
+    let reward = Reward.postcards
+      |> Reward.lens.shipping.enabled .~ true
+
+    let project = Project.cosmicSurgery
+      |> Project.lens.state .~ .live
+      |> Project.lens.personalization.isBacking .~ true
+      |> Project.lens.personalization.backing .~ (
+        .template
+          |> Backing.lens.paymentSource .~ Backing.PaymentSource.template
+          |> Backing.lens.status .~ .pledged
+          |> Backing.lens.reward .~ reward
+          |> Backing.lens.rewardId .~ reward.id
+          |> Backing.lens.shippingAmount .~ 10
+          |> Backing.lens.amount .~ 700.0
+      )
+
+    let updateBackingEnvelope = UpdateBackingEnvelope(
+      updateBacking: .init(
+        checkout: .init(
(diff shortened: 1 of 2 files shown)


Library/ViewModels/PledgeViewModelTests.swift
```diff
@@ -2602,6 +2602,110 @@ final class PledgeViewModelTests: TestCase {
...
+      self.beginSCAFlowWithClientSecret.assertDidNotEmitValue()
+      self.notifyDelegateUpdatePledgeDidSucceedWithMessage.assertDidNotEmitValue()
+      self.popToRootViewController.assertDidNotEmitValue()
+      self.showErrorBannerWithMessage.assertDidNotEmitValue()
+      self.configurePledgeViewCTAContainerViewIsEnabled.assertValues([false, true])
+      self.processingViewIsHidden.assertDidNotEmitValue()
+
+      self.vm.inputs.submitButtonTapped()
+
+      self.goToRiskMessagingModal.assertDidNotEmitValue()
```
msadoon, line 2688: This is the only output we are testing, so to reduce bloat in this test, the entire test after `self.vm.inputs.viewDidLoad()` could be 
```
      self.vm.inputs.pledgeAmountViewControllerDidUpdate(
        with: (amount: 25.0, min: 25.0, max: 10_000.0, isValid: true)
      )
      self.vm.inputs.shippingRuleSelected(.template)
      self.vm.inputs.submitButtonTapped()
      self.goToRiskMessagingModal.assertDidNotEmitValue()
```

Because that is all we really care about in this particular patch. Also then the test could be more specific to that use `testRiskModalNotShowing_WhenUpdating_Success`

Unless you want to test everything again under these specific conditions, which is also valid. Just a thought.

msadoon (approved):
good to go!

PR #1621, opened 2021-10-14
[NTV-193] project_page_v2 Optimizely Feature Flag

msadoon:
# 📲 What

Added the optimizely feature (not experiment) `project_page_v2` to all entry points.

# 🤔 Why

Need to update all entry points for our new project page, which the option to toggle all those entry points off at once to return to the previous project page.

# 🛠 How

Created the feature (not experiment) to toggle on/off through Optimizely Settings in app.

# 👀 See
| Before | After|
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [x] Ensure toggling new flag works in Optimizely Settings (in-app)
- [x] Ensure all the entry-points in these [scenarios](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/1805123593/Project+Page+Navigation+Selector+Feature+Entry+Points) launch the new project page when the feature in Optimizely is ON.
- [x] Ensure the same entry points above launch the old project page when the feature in Optimizely is OFF.

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -10838,9 +10838,9 @@ public enum GraphAPI {
         }
         allowedAddons {
           __typename
-          pageInfo {
+          nodes {
             __typename
-            startCursor
+            id
           }
         }
         description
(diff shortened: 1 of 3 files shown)


msadoon:
Left tests as is as well. I mean its' difficult to test "presentation" of a view controller. If it was an output out a view model, then yes it is testable, because the signals have to ui component. But presentation out of uikit requires waiting for an unknown amount of time, and testing "the next page" so to speak which I haven't seen examples of in this codebase. We could build a test harness probably using FBSnapShot test case, but all that requires a lot of investigation and foundation code.

PR #1627, opened 2021-10-27
[NTV-196] Add Video Player to new Project Page

msadoon:
# 📲 What

So our designs require a video player to be shown only on the overview tab.

# 🤔 Why

Need to support existing functionality on `ProjectPamphletViewController` ported over to this page.

# 🛠 How

Studying what was being done on `ProjectPamphletMainCell`, the video player delegation and interactions with navigation bar and other intermediate views.

Most of those interactions were unnecessary because previously delegation was built to communicate with all the views in the heirarchy in this way:

videoplayer --> main cell --> content view controller --> pamphlet view controller --> nav bar controller

So looking at this, noticed most of the delegation calls were there for the nav bar controller to do some animation to project categories, which I've never seen supported in the app before and isn't currently supported in our non-storyboard project nav bar. The backer label is also effected by this delegation in main cell but its' confirmed removed in this Slack [convo](https://kickstarter.slack.com/archives/CG55BBBBL/p1635282157010500) for now. 

Bottom line: No need for delegation.

The most complicated part of this was getting the video player showing inside the view controller. The video player it self handles all the playback.

**Huge caveat**
Regarding this [conversation](https://kickstarter.slack.com/archives/CG55BBBBL/p1635372019046400) on Slack, we've decided to move away from the paging view controller and support scrolling on the `ProjectPageViewController` with one table view that will be refreshed with new data from existing `pagingDataSource` whenever a tab is selected by the user. 

Here's the problem we are facing:

Notice that in landscape the iPad is unscrollable. So imagine if users' open a project on iPad in landscape and they cannot scroll, same problem on iPhone. This will require the data source re-jig above.

Added as a todo, will probably make this a new ticket, leave this ticket unmerged until that is done.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Videoplayer shown on all tabs (for hide/show is implemented [here](https://kickstarter.atlassian.net/browse/NTV-198))
- [ ] Can play/stop video if its there.
- [ ] Shows a project image if not there.

# ⏰ TODO

- [ ] Figure out how to re-write the new data source to support landscape scrolling on iphone and ipad. Also solves the "not enough space on smaller screens" issue. (Probably another ticket)
- [ ] Handle situation where project has no image. Check if this is even possible, maybe need an additional constraint to handle the expansion/hiding of the video player. Just don't want it to completely blank in any scenario.
- [ ] Write tests


msadoon:
closing because the base branch `main` has changed a lot since this pr was opened, including changes to the navigation selector and the whole page being able to scroll.

PR #1629, opened 2021-11-03
[NTV-122] Replace Category GraphSchema Based Calls With Apollo

msadoon:
# 📲 What and  🤔 Why

As part of our ongoing maintenance/technical debt/effort to remove the previously written `GraphSchema` and replace it with Apollo, there are two calls that need to be updated:

`fetchGraphCategory`
`fetchGraphCategories`

in our `ServiceType` network layer.

# 🛠 How

Ensured the DSL created by the two queries above (as examined on data dog) was the same query created by `FetchCategory` and `FetchRootCategories`.

Replace instances of the calls in ""What And Why" section are replaced with these new Apollo supported calls.

Lots of testing ensures that these queries properly map to the preexisting model `Category`

# 🏎 Performance

Should not affect performance as the `CategoryFragment` that is used only within the `FetchProject` queries has not changed.

Also the new `FetchCategory` and `FetchRootCategories` are exactly the same GQL as before (created by the DSL), so performance shouldn't be affected.

# ✅ Acceptance criteria

- [x] Ensure after pledging, `ThankYouViewController` shows the other projects in the category (as it did before).
- [x] Ensure after picking the category filter, shows the other categories (as it did before).


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/MockService.swift
```diff
@@ -702,19 +695,26 @@
...
+    internal func fetchGraphCategories()
+      -> SignalProducer<RootCategoriesEnvelope, ErrorEnvelope> {
+      guard let client = self.apolloClient else {
+        return .empty
       }
-      return SignalProducer(value: RootCategoriesEnvelope.template)
+
+      let fetchGraphCategoriesQuery = GraphAPI.FetchRootCategoriesQuery(withParentCategoryAnalyticsName: true)
+
+      return client.fetchWithResult(query: fetchGraphCategoriesQuery, result: self.fetchGraphCategoriesResult)
```
singhhari, line 706: Noticed we're only using this syntax with the requests you've edited in this PR. Do we need to retrofit the other mock service methods with this implementation too? If so we should have a ticket in the backlog so we remember.

KsApi/fragments/CategoryFragment.graphql
```diff
@@ -5,5 +5,6 @@ fragment CategoryFragment on Category {
   parentCategory {
     id
     name
+    analyticsName @include(if: $withParentCategoryAnalyticsName)
```
singhhari, line 8: in what situation are we calling false here?

singhhari (approved):
lgtm just curious about the boolean we're passing into the query and the use cases between true/false

KsApi/fragments/CategoryFragment.graphql
msadoon: Yea so I noticed that the `parentCategory` `analyticsName` added about 0.5 seconds of lag to our fetch project call above, so I didn't want to add it blindly to a pre-existing fragment used by many queries (fetch project, fetch backing, fetch addons), so I included it optionally. The only queries that set it to true should be the new ones (`fetchCategories` and `fetchCategory`) every other call should have it set to `false`

KsApi/MockService.swift
msadoon: Yea good call, most of the methods are already using `client.fetchWithResult` already, ones missing are:
`facebookConnect(facebookAccessToken`
`fetchCommentReplies(query` (will be handled in upcoming ticket)
`fetchConfig()` (not sure if this is a network call necessarily, need to investigate)
`fetchFriends()`
`fetchFriendStats()`
`followAllFriends()`
`exportData()`
`followFriend(userId`
`fetchErroredUserBackings(status` (there is a note about this needing it's return envelope to be decodable)
`fetchGraph<A>` (use case)
`unfollowFriend(userId`
`login`
`logout`
`fetchActivities(count`
`fetchBacking(forProject`
`backingUpdate(forProject`
...there's a lot more...and I just realized a lot of these are v1 calls, so I think as we move those to GQL we'll do the refactoring here as well (have to, project will fail to build otherwise).
All of the GQL calls though should be using the `client.fetchWithResult(` method above, and that should be complete once we remove the DSL.

PR #1630, opened 2021-11-04
[NTV-275] Refactor ProjectPageViewController to a table view implementation

singhhari:
# 📲 What + 🤔 Why

While building our new project page we ran into an unforeseen issue related to the scrolling video feature. In particular, we realized a few different things didn't quite work out with our current implementation. First off, the amount of scrollable real estate on the screen was quite small. We resolved this by turned the header labels into cells and by eliminating our page view controller implementation. We're using a table view with its own data source to reload the data when the navigation selector changes values.

# 🛠 How

- Deleted individual section view controllers, view models and data sources
- Migrated specific outputs from aforementioned view models to `ProjectPageViewModel`
- Consolidated all of the aforementioned data sources into a single one that drives the tableview in `ProjectPageViewController`
- Converted header labels into specific table view cells

# ✅ Acceptance criteria

- [x] Navigation between sections should behave as before
- [x] Section headers (now cells) should be scrollable
- [x] Scrolling works consistently across the different tabs (only FAQs and Environmental Commitments are done so far)


Kickstarter-iOS/DataSources/ProjectPageViewControllerDataSourceTests.swift
```diff
@@ -0,0 +1,487 @@
+@testable import Kickstarter_Framework
```
singhhari, line 1: Interesting theme in this file is that the data source will evaluate any of the cases in the enum as an existing section but obviously the number of rows will be zero unless it is set. Made testing a bit verbose but does explain the gist of what's happening behind the hood.

Kickstarter-iOS/DataSources/ProjectPageViewControllerDataSource.swift
```diff
@@ -0,0 +1,114 @@
+import KsApi
```
singhhari, line 1: Renamed and consolidated all the deleted data source logic into here.

singhhari (commented):
The screenshot tests are a bit weird because we can't render the cells on the screen without view model interaction. They show the navigation selector and the pledge cta container view but nothing else. I imagine when the overview screen is built out we will need to re-record snapshots and that will be the default render.

singhhari:
Pretty sure coverage is down because I deleted a bunch of files that were testing the individual view controllers, data sources and view models.

Library/ViewModels/ProjectPageViewModelTests.swift
```diff
@@ -1140,6 +1148,194 @@ final class ProjectPageViewModelTests: TestCase {
...
+
+    self.vm.inputs.projectNavigationSelectorViewDidSelect(index: overviewSection)
+
+    // The view model skips the first emission
+    self.updateDataSourceProjectProperties.assertDidNotEmitValue()
+
+    self.vm.inputs.projectNavigationSelectorViewDidSelect(index: environmentalCommitmentsSection)
+
+    self.updateDataSourceProjectProperties.assertDidEmitValue()
+  }
```
msadoon, line 1232: I'm not sure how this test and the one above are different?

msadoon:
> The screenshot tests are a bit weird because we can't render the cells on the screen without view model interaction. They show the navigation selector and the pledge cta container view but nothing else. I imagine when the overview screen is built out we will need to re-record snapshots and that will be the default render.

Cool yea, so I noticed we wrote no new tests for the `ProjectPageViewController`. 
We could improve our coverage a bit here too by adding tests for FAQ tab (logged in and logged out), EC tab tab. (3 tests). In order to trigger the new tab can we call `projectNavigationSelectorViewDidSelect` in our instance of the vc and use `scheduler.advance()` to unsure the update happens?

msadoon (changes requested):
Great refactor!
Just some minor odds/ends to clean up and its' good to go.

Noticed these things in-app testing.

After logging in we should refresh the datasource to ensure the user doesn't need to switch between tabs to see any content. I believe there should be an existing observer in the view controller we can tie into, if not make it a separate ticket and we can tackle it later.


So noted this in the code too, would be nice to preserve state in open faqs between tab switches. If there are some weird edge cases/not super straightforward can make it a separate ticket.



Notice the cutoff on the pledge cta? I noted why in the comments below. Not a huge deal, I can fix it in my next Pr, just drop a note in my video player ticket.


So noted this in the code comments too, but the extra cell "lines" are due to their being no footer on the tableview. Either set it to `nil` in this ticket or to the pledge cta fix above which should fix this issue.

Pretty close to done, maybe a couple hours worth of changes here, mostly with respect to testing. Let me know if you want me to tackle anything here 👍🏿

Kickstarter-iOS/DataSources/ProjectPageViewControllerDataSourceTests.swift
singhhari: Ah I see. When we were working on comment replies there was an explicit call to `padValuesForSection` which was required because we ran into some out of bounds errors. Never actually looked into the implementation details for the `set` method but it makes sense that it's in there. Thanks for digging into that!

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -21,16 +24,22 @@ public protocol ProjectPageViewModelInputs {
...
+
+  /// Call when the view loads.
+  func viewDidLoad()
 }
 
 public protocol ProjectPageViewModelOutputs {
-  /// Emits a tuple of a `NavigationSection` and `Project` to configure the page view controller.
-  var configurePagesDataSource: Signal<(NavigationSection, Project), Never> { get }
+  /// Emits a tuple of a `NavigationSection` and `ExtendedProjectProperties` to configure the data source
+  var configureDataSource: Signal<(NavigationSection, ExtendedProjectProperties), Never> { get }
```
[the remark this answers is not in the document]
singhhari: Remembered your previous review while I did this haha

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -198,12 +217,42 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
+
     // We skip the first one here because on `viewDidLoad` we are setting .overview so we don't need a useless emission here
-    self.navigatePageViewController = self.projectNavigationSelectorViewDidSelectProperty.signal
+    self.updateDataSource = self.projectNavigationSelectorViewDidSelectProperty.signal
       .skipNil()
       .map { index in NavigationSection(rawValue: index) }
       .skipNil()
+      .combineLatest(with: project.map(\.extendedProjectProperties).skipNil())
+      .map { navSection, projectProperties in
+        let initialIsExpandedArray = Array(repeating: false, count: projectProperties.faqs.count)
```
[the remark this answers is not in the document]
singhhari: Yeah right now we are not preserving state as it wasn't explicitly listed in the acceptance criteria. I'm going to look into how much effort it will take to do that after I push the first set of changes up.

Library/ViewModels/ProjectPageViewModelTests.swift
singhhari: A single output which has a tuple in it was mapped into two separate observers (we have to do this because tuples are not equatable). The tests are probably identical in execution but are testing two different outputs. It looks like throughout our project we have a habit of testing these mapped observers in individual tests rather than collectively in one.

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -129,50 +135,32 @@ public final class ProjectPageViewController: UIViewController, MessageBannerVie
...
     _ = (self.projectNavigationSelectorView, self.view)
       |> ksr_addSubviewToParent()
 
     let constraints = [
-      self.projectNavigationSelectorView.leftAnchor.constraint(equalTo: self.view.leftAnchor),
-      self.projectNavigationSelectorView.rightAnchor.constraint(equalTo: self.view.rightAnchor),
-      self.projectNavigationSelectorView.topAnchor.constraint(equalTo: self.view.topAnchor),
+      self.tableView.topAnchor.constraint(equalTo: self.view.topAnchor),
+      self.tableView.widthAnchor.constraint(equalTo: self.view.widthAnchor),
+      self.tableView.bottomAnchor.constraint(equalTo: self.pledgeCTAContainerView.topAnchor),
```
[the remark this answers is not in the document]
msadoon: handling this in video player ticket.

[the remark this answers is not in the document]
singhhari: Messaged you about this on Slack but for posterity will summarize here. If I change the constraint to what is suggested the `pledgeCTAContainerView` will no longer be in the view. I'm going to set the `tableFooterView` property to `nil` and we can tackle these changes in the video player ticket.

singhhari:
> Great refactor! Just some minor odds/ends to clean up and its' good to go.
> 
> Noticed these things in-app testing.  After logging in we should refresh the datasource to ensure the user doesn't need to switch between tabs to see any content. I believe there should be an existing observer in the view controller we can tie into, if not make it a separate ticket and we can tackle it later.
> 
> 
> 
> So noted this in the code too, would be nice to preserve state in open faqs between tab switches. If there are some weird edge cases/not super straightforward can make it a separate ticket.
> 
> 
> Notice the cutoff on the pledge cta? I noted why in the comments below. Not a huge deal, I can fix it in my next Pr, just drop a note in my video player ticket.
> 
> 
> So noted this in the code comments too, but the extra cell "lines" are due to their being no footer on the tableview. Either set it to `nil` in this ticket or to the pledge cta fix above which should fix this issue.
> Pretty close to done, maybe a couple hours worth of changes here, mostly with respect to testing. Let me know if you want me to tackle anything here 👍🏿

Added an observer for `userSessionStarted` which will trigger the data source reloading via the `project` signal being updated. Seems to be working well and I added a test for it. Let me know what you think.

I'm going to create a new ticket to handle the preservation of the `isExpanded` states. It would be best to unblock the other tickets and treat that as a product optimization toward the end of the epic. 

Otherwise addressed the other comments and pushed those up.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/DataSources/ProjectPageViewControllerDataSourceTests.swift
```diff
@@ -0,0 +1,487 @@
...
+      )
+      XCTAssertEqual(
+        "ProjectFAQsEmptyStateCell",
+        self.dataSource.reusableId(item: 0, section: self.faqsEmptySection)
+      )
+    }
+  }
+
+  func testLoadEnvironmentalCommitments_LoggedIn() {
+    let environmentalCommitments = [
```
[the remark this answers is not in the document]
msadoon: Could do this for some of the faqs, but no big deal.

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
```diff
@@ -179,6 +186,14 @@ public final class ProjectPageViewController: UIViewController, MessageBannerVie
         name: .ksr_projectBacked,
         object: nil
       )
+
+    self.sessionStartedObserver = NotificationCenter.default
```
msadoon, line 190: Just curious why use an observer stored locally that you need to de-init? You probably don't need it (as done in `ProjectPageNavigationBarView`)

Library/ViewModels/ProjectPageViewModelTests.swift
```diff
@@ -1231,6 +1206,36 @@ final class ProjectPageViewModelTests: TestCase {
...
+      self.updateDataSourceProjectProperties.assertDidNotEmitValue()
+
+      self.vm.inputs.projectNavigationSelectorViewDidSelect(index: environmentalCommitmentsSection)
+
+      self.updateDataSourceProjectProperties.assertDidEmitValue()
+
+      withEnvironment(currentUser: .template) {
+        self.vm.inputs.userSessionStarted()
+
+        self.updateDataSourceProjectProperties.assertDidEmitValue()
```
msadoon, line 1234: Just curious, does this means the observer emits again, or if it emitted once, it's still considered `emittted` the second time the method is called?

msadoon (approved):
good to go!
one minor thing about the need for an observer in the code, also the screen tests probably should cover the other two tabs, but I can fix those in the video player ticket. 👍🏿

Library/ViewModels/ProjectPageViewModelTests.swift
singhhari: I believe this implies it is emitted again although I can see how this is a bit ambiguous and not entirely obvious. The gist of the test was to verify that after the user session starts that emits so I think it is on the right page?

Kickstarter-iOS/Views/Controllers/ProjectPageViewController.swift
singhhari: Updated.

Library/ViewModels/ProjectPageViewModelTests.swift
msadoon: yep, its' probably that it emits again, I'll double check today, not 100% on it, otherwise everything else is gtg.

PR #1635, opened 2021-11-15
[NTV-284] - Currency Display Differences Between Android and iOS

msadoon:
# 📲 What

Earlier last week it was brought to our attention that projects created in countries that set their project currency to something different than their country's currency.

# 🤔 Why

Obviously this is a big issue and effects pledging volume as evidenced in this large [project](https://www.kickstarter.com/projects/shortcutlabs/flic-twist-the-wireless-dial-for-your-smart-home?ref=bgtrbl&token=e1d08d69).

https://kickstarter.atlassian.net/browse/NTV-284 contains the exact problem and this [document](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/1840840709/Price+Display+Errors+on+iOS.) investigates the full scope of the issue.

# 🛠 How

**Part 1**
In the first part of this fix, noticed that we were simply getting the project's country code, looking it up in our `Project.Country` enum's `all` and mapping it to the `project`'s `country` object we returned to be displayed.
However, this breaks down when the creator specifies a project currency _different_ from the hard coded currencies returned from within the app. Like in the example project above, the project's country is Sweden, but its' built to accept pledges in "EUR". 

So updated the `Project.Country` adapter to take in the project fragment's currency and update the hardcoded country's currency if it's different.

This solved the currency symbol display issues on project, rewards selection, add-ons and checkout screens. (Both logged in and logged out)

**Part 2**
Thanks @Arkariang for showing us where the users' currency is [populated](https://kickstarter.slack.com/archives/C02LY6LM0CU/p1637087305018100).

Basically there is a `fetchConfig` endpoint called on `AppDelegate`, which is what Android uses' to populate the default currency for conversions if the user is not logged in. Previously iOS was just using `USD` if it wasn't there, which causes incorrect conversions to show up for example in the rewards (because each reward uses' the configs' currency to display its converted amount.)

**Part 3**
Another [issue](https://kickstarter.atlassian.net/browse/SD-1061?atlOrigin=eyJpIjoiNGUzMzRhMTQzNjQ4NGJjNmI1NGUzN2E5YzYwYjcwNmEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) noticed was on the manage pledge screen, no reward rewards on iOS were displaying an incorrect currency conersion (ie. "About $0"), Android was displaying correctly (ie. About $1). That was corrected by adding the `minPledge` and `fxRate` properties from the backing's project fragment. (`Backing+BackingFragment.swift`)

| Before 🦋 | After 🐛 |
| --- | --- |
|  |  |

# 👀 See

| Before 🦋 | After 🐛 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria
- [x] Investigate all screens in the app listed in the document. Check that there are no differences between the Android or iOS app for _any_ project, not just the one above.

# ⏰ TODO

**Follow up**
- [x] Write tests
- [x] Go through follow up scenarios in document and note any follow up issues.

### DIFF
--- a/KsApi/models/graphql/adapters/Backing+BackingFragment.swift
+++ b/KsApi/models/graphql/adapters/Backing+BackingFragment.swift
@@ -1,4 +1,4 @@
-import Foundation
+import Prelude
 import ReactiveSwift
 
 extension Backing {
@@ -56,7 +56,18 @@ private func backingStatus(from backingFragment: GraphAPI.BackingFragment) -> Ba
 }
 
 private func backingReward(from backingFragment: GraphAPI.BackingFragment) -> Reward? {
-  guard let reward = backingFragment.reward?.fragments.rewardFragment else { return .noReward }
+  guard let reward = backingFragment.reward?.fragments.rewardFragment else {
+    let projectMinimumPledgeAmount: Int = backingFragment.project?.fragments.projectFragment.minPledge ?? 1
+    let projectFXRate: Double = backingFragment.project?.fragments.projectFragment.fxRate ?? 1.0
+
+    let convertedMinimumAmount = projectFXRate * Double(projectMinimumPledgeAmount)
+
+    let emptyReward = Reward.noReward
+      |> Reward.lens.minimum .~ Double(projectMinimumPledgeAmount)
+      |> Reward.lens.convertedMinimum .~ convertedMinimumAmount
+
+    return emptyReward
+  }
 
   return Reward.reward(from: reward)
 }
--- a/KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
+++ b/KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
@@ -4,37 +4,53 @@ import Prelude
 import ReactiveSwift
 
 extension Project {
-  public typealias ProjectPamphletData = (project: Project, backingId: Int?)
+  public typealias ProjectPamphletData = (
+    project: Project,
+    backingId: Int?
+  )
 
   static func projectProducer(
-    from data: GraphAPI.FetchProjectByIdQuery.Data
+    from data: GraphAPI.FetchProjectByIdQuery.Data,
+    configCurrency: String?
   ) -> SignalProducer<ProjectPamphletData, ErrorEnvelope> {
-    let projectAndBackingId = Project.project(from: data)
+    let projectAndBackingId = Project.project(
+      from: data,
+      configCurrency: configCurrency
+    )
 
     guard let project = projectAndBackingId.0 else {
       return SignalProducer(error: ErrorEnvelope.couldNotParseJSON)
     }
 
-    let data = ProjectPamphletData(project: project, backingId: projectAndBackingId.1)
+    let data = ProjectPamphletData(
+      project: project,
+      backingId: projectAndBackingId.1
+    )
 
     return SignalProducer(value: data)
   }
 
-  static func projectProducer(
-    from data: GraphAPI.FetchProjectBySlugQuery.Data
-  ) -> SignalProducer<ProjectPamphletData, ErrorEnvelope> {
-    let projectAndBackingId = Project.project(from: data)
+  static func projectProducer(from data: GraphAPI.FetchProjectBySlugQuery.Data,
+                              configCurrency: String?) -> SignalProducer<ProjectPamphletData, ErrorEnvelope> {
+    let projectAndBackingId = Project.project(
+      from: data,
+      configCurrency: configCurrency
+    )
 
     guard let project = projectAndBackingId.0 else {
       return SignalProducer(error: ErrorEnvelope.couldNotParseJSON)
     }
 
-    let data = ProjectPamphletData(project: project, backingId: projectAndBackingId.1)
+    let data = ProjectPamphletData(
+      project: project,
+      backingId: projectAndBackingId.1
+    )
 
     return SignalProducer(value: data)
   }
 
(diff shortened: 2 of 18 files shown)


KsApi/models/graphql/adapters/Backing+BackingFragment.swift
```diff
@@ -56,7 +56,18 @@ private func backingStatus(from backingFragment: GraphAPI.BackingFragment) -> Ba
 }
 
 private func backingReward(from backingFragment: GraphAPI.BackingFragment) -> Reward? {
-  guard let reward = backingFragment.reward?.fragments.rewardFragment else { return .noReward }
+  guard let reward = backingFragment.reward?.fragments.rewardFragment else {
+    let projectMinimumPledgeAmount: Int = backingFragment.project?.fragments.projectFragment.minPledge ?? 1
```
singhhari, line 60: Pretty interesting that we have to do this ourselves in GraphQL but didn't have to with V1? I wonder what happened there and why this bug is only happening on iOS (and not on Android)?

KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
```diff
@@ -48,14 +64,15 @@ extension Project {
         rewards: [noRewardReward(from: fragment)],
         addOns: nil,
         backing: nil,
-        currentUserChosenCurrency: data.me?.chosenCurrency
+        currentUserChosenCurrency: data.me?.chosenCurrency ?? configCurrency
```
singhhari, line 67: I imagine this is one of the more important bits to the PR?

KsApi/models/graphql/adapters/Project.Country+CountryFragment.swift
```diff
@@ -1,10 +1,26 @@
...
 extension Project.Country {
-  static func country(from countryFragment: GraphAPI.CountryFragment) -> Project.Country? {
-    guard let country = Project.Country.all
-      .first(where: { $0.countryCode == countryFragment.code.rawValue }) else { return nil }
+  static func country(from countryFragment: GraphAPI.CountryFragment,
+                      minPledge: Int,
+                      maxPledge: Int,
+                      currency: GraphAPI.CurrencyCode) -> Project.Country? {
+    guard let countryWithCurrencyDefault = Project.Country.all
+      .first(where: { $0.countryCode == countryFragment.code.rawValue }) else {
```
singhhari, line 9: Could use trailing closure syntax here and below but definitely not a big deal.

singhhari (approved):
Looks good! I ran through some European projects that were both unpledged and pledged on my prod account. I didn't run into any issues or observe any bugs. Nice job with getting to the bottom of this and testing it. One quick note on an Australian project:

Simulator Screen Shot - iPhone 8 - 2021-11-17 at 17 22 07

I'm not sure if we should be showing "about $0" here? Just as an FYI, 1 USD == .73 cents in Australian Dollars.

KsApi/models/graphql/adapters/Backing+BackingFragment.swift
msadoon: well v1/ returned the no reward reward right? in graph the rewards we just the ones the user created, so we manually have to create the no reward reward in instances (like this one) where its' displayed. Just graph things lol

msadoon: I guess to elaborate Android did this accurately with GQL the first time on these pages.

KsApi/models/graphql/adapters/Project+FetchProjectQueryData.swift
msadoon: Yep, so this is something learned from how Android defaults their currency conversion. Basically before this we were using "USD" if the `data.me?.chosenCurency` was nil, which threw off all the conversions, but now because we're using an endpoint `v1/config` that gives us the users' default "global state" values, we can use its' currency instead.

Why did we miss this in the first place? Had no idea `v1/config` until Isa told me about it :)

KsApi/models/graphql/adapters/Project.Country+CountryFragment.swift
msadoon: So oddly this works:
```
    let state = Project.Country.all.first { $0.countryCode == countryFragment.code.rawValue }
    
    guard let countryWithCurrencyDefault = state else {
      return nil
    }
```
but this does not:
```
    guard let countryWithCurrencyDefault = Project.Country.all.first { $0.countryCode == countryFragment.code.rawValue } else {
      return nil
    }
```

Getting a bunch of errors here, so I think I'll leave it for now, as the former doesn't really save us much space/is necessarily cleaner.

msadoon:
> Looks good! I ran through some European projects that were both unpledged and pledged on my prod account. I didn't run into any issues or observe any bugs. Nice job with getting to the bottom of this and testing it. One quick note on an Australian project:
> 
> Simulator Screen Shot - iPhone 8 - 2021-11-17 at 17 22 07
> 
> I'm not sure if we should be showing "about $0" here? Just as an FYI, 1 USD == .73 cents in Australian Dollars.

Ok, yea good point, I know its' weird to show like "About $0" anywhere in the app (not just this page, it exists on manage pledge, project page, backing confirm, I think at least one more..) Its' not really do much harm being there as most of the time it provides a valuable conversion for none zero amounts, but my reluctance to hide it when its' zero it is mostly due to the same change required on multiple screens. Really also should be a product discussion first. Just my two cents, but happy to change it if we decide as a team to remove it.

PR #1640, opened 2021-11-19
[NTV-262] Fetch and apply translations for new project page

singhhari:
# 📲 What + 🤔 Why

Fetching and applying new translations that are required for the new project page. Kickstarter natively translates in several different languages and its important that we apply all translations before shipping a feature. 

# ✅ Acceptance criteria

- [x] Switch the language on your device and open up a project.
- [x] Navigate through the new project page and verify all strings that can be translated are being translated.
- [ ] Do this for each of the languages we support and check that there isn't anything unusual going on with the constraints, truncation of labels etc.

# ⏰ TODO

- [x] Wait for the final String to be translated and deployed from [this PR](https://github.com/kickstarter/kickstarter/pull/22450) (already merged)
- [x] Once that String is available on iOS, replace line 114 of the `ProjectEnvironmentalCommitmentDisclaimerCell` with the new string. Feel free to remove the TODO and/or re-factor the implementation as you see fit.
- [x] Additionally, check if the "Something else" string is available and apply it to line 63 of `ProjectEnvironmentalCommitmentCellViewModel`


Library/ViewModels/ProjectEnvironmentalCommitmentCellViewModel.swift
```diff
@@ -42,3 +45,20 @@ public final class ProjectEnvironmentalCommitmentCellViewModel:
...
+  case .sustainableMaterials:
+    return Strings.Sustainable_materials()
+  case .environmentallyFriendlyFactories:
+    return Strings.Environmentally_friendly_factories()
+  case .sustainableDistribution:
+    return Strings.Sustainable_distribution()
+  case .reusabilityAndRecyclability:
+    return Strings.Reusability_and_recyclability()
+  case .somethingElse:
+    return ""
```
paulomcg, line 62: Are we still missing this string?

singhhari: Did not see this come back from the script when I ran it. Will include it into the TODO section above.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/ProjectEnvironmentalCommitmentCellViewModel.swift:49:1 | warning: (spaceInsideComments) Add leading and/or trailing space inside comments. |


</p>

Library/ViewModels/ProjectEnvironmentalCommitmentCellViewModel.swift
singhhari: No deploy as of right now so this will have to wait until next week. Everything in the TODO is accurate but will probably depend on the web deploy + translations being complete beforehand.

msadoon (approved):
The string translations look good enough.
We support accessibility so I'll note in this ui odds/ends [ticket](https://kickstarter.atlassian.net/browse/NTV-297) that the navigation selector scrolling could be improved for larger accessibility sizes (basically when large text sizes is on, and the selector is above half in Settings).



Would like to create a separate ticket to fix the scrolling behaviour of the selector before we ship the feature, because it does affect new iPhones too (iPhone X and up).

PR #1642, opened 2021-11-24
[NTV-292] Pin Navigation Bar To Top Of Project Page Under Navigation Bar

msadoon:
# 📲 What

Small UI update to pin nav bar while scrolling on all tabs

# 🤔 Why

Design/product simplification until video player is removed from the main cell of the overview tab and put above the navigation bar.

# 🛠 How

Set the `tableview.nil` to `nil` and updated `projectSelectionViewNavigationBar` `height` based on presentation style, accounting for differences in height between iPad and iPhone due to the way the view controller is presented.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [ ] Ensure no regressions from previously built navigation selector behaviour.
- [ ] Ensure navigation selector is always seen on every tab while scrolling.


Arkariang (approved):
lgtm!

PR #1644, opened 2021-12-01
[NTV-295] Place Campaign Tab Under Optimizely Feature Flag

msadoon:
# 📲 What

The flag is `project_page_v2_story` on Optimizely.

There's also a few more (unrelated?) updates here in this pr.

- dynamic loading of extended tabs as content becomes available.
- fixed up tests for `ProjectPageViewController` not rendering tab contents
- updated the segment event tab to `campaign` from `story`
- fixed up a white header gap under the `bottomBorderView` on `ProjectNavigationSelectorView`
- removed centering of tabs on iPad, was adding unncessary complexity to the middle tab also not a design req. (cc. @singhhari if you still want this we can add it back in, just requires a little more customization for iPad specifically.)

# 🤔 Why

The core of this pr addresses the Optimizely flag mentioned above and is integrated like any other feature listed under `OptimizelyFeature.swift`

**Dynamic tab loading**
So as you can see in `ProjectNavigationSelectorViewModel` we are starting with `overview` as the only required tab and then adding tabs as we get the `extendedProperties` filled out by the network request. That way we don't have to deal with all complexity related to the pledge cta as documented in this [ticket](https://kickstarter.atlassian.net/browse/NTV-258). The pledge cta will act as it always did, to inform the overview tab which is _exactly_ the same as the the `ProjectContentViewController` of `ProjectPamphletViewController`.

**Write tests for `ProjectPageViewController`**
Previously we had ignored these tests as this page was being built out. This was happening because the the table view wasn't generating any content. After some investigation found that we were incorrectly using `observeOnControllerAction` to load the data source, as opposed to `observeOnUI` which should be correct one, because we're pushing/presenting anything in those outputs, simply loading the datasource.

...which leads to

**Refactoring the `ProjectNavigationSelectorView` to fill its own height with its' own content**
This one was tricky and leads to the final fix which is that pesky white underline documented by @paulomcg  [here](https://kickstarter.atlassian.net/browse/NTV-292?focusedCommentId=31989)

Boiled down to setting the scrollview height equal to the buttons stack view height. Otherwise the scrollview doesn't know its height after the view appears.

This is beneficial from our previous approach because the content can change depending on accessibility fonts and there is no unnecessary padding that is static leading to the white space problem above.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [x] Ensure optimizely flag turns off/on "Campaign" tab
- [x] Ensure no white underline under dropshadow
- [x] Ensure that tabs load as content becomes available, so a good test is load the page with no network, should be one tab "Overview", then hit retry on pledge cta and should see the other tabs load as the network request completes.
- [x] Ensure that segment event for "Campaign" tab sends "campaign" as the event name, not "story".

# ⏰ TODO

- [x] Re-run and add new tests to `ProjectPageViewController` for all tabs.
- [x] Correct tests after the segment event name change.


singhhari:
When the feature flag is turned off, the campaign tab is hidden but we do have a strange alignment issue where there is some white space due to not centering. I think as long as design/product are ok with that we are g2g.

Library/OptimizelyFeature.swift
```diff
@@ -3,13 +3,15 @@ import Foundation
 public enum OptimizelyFeature: String, CaseIterable {
   case commentFlaggingEnabled = "ios_comment_threading_comment_flagging"
   case navigationSelectorProjectPageEnabled = "project_page_v2"
+  case projectPageStoryTabEnabled = "project_page_v2_story"
```
singhhari, line 6: wonder if we should prefix this with ios or native? is android using the same flag?

Library/ViewModels/ProjectNavigationSelectorViewModel.swift
```diff
@@ -52,16 +52,27 @@ public final class ProjectNavigationSelectorViewModel: ProjectNavigationSelector
   public init() {
     self.animateButtonBottomBorderViewConstraints = self.buttonTappedProperty.signal
 
+    let displayableTabs: (Project) -> [NavigationSection] = { project in
```
singhhari, line 55: Ah so this is handling the conditional loading of the overview tab at first. I must have missed some context while I was out but this experience is cool with product/design?

singhhari (approved):
Nice job on this! Left a few small comments but otherwise lgtm.

Library/OptimizelyFeature.swift
msadoon: android hasn't added the flag on optimizely yet. I see they are prefixing theres' with `android`, but the ticket said `project_page_v2_story` which is what I went with, happy to change it later if that's cool.

Library/ViewModels/ProjectNavigationSelectorViewModel.swift
msadoon: Yea it should be, because it really handles all our failing cases for the pledge cta pretty nicely. Which solves the issues raised in this [ticket](https://kickstarter.atlassian.net/browse/NTV-258) as well, which is a nice win.

Essentially the pledge cta is tied only to the overview tab which is always loaded when the vc viewDidLoad gets called.
So any failure/inability for the network or project to have `extendedProjectProperites` (which currently we only get through these new network requests) means we don't have to worry about confusion when the tabs are visible, they just simply are not created until the data is there. As a test can use Charles to fail the network request on this page, then "retry" on the pledge cta and see the tabs appear.
 
I believe product is good with this because they approved this [ticket](https://kickstarter.atlassian.net/browse/NTV-294) for release which is related to the same issue above.

msadoon:
> When the feature flag is turned off, the campaign tab is hidden but we do have a strange alignment issue (left alignment essentially) where there is some white space due to not centering. I think as long as design/product are ok with that we are g2g.

I can't see a noticeable white space issue. But it might have been fixed in the next pr for NTV-297, if you notice it there, please let me know, I'll fix it there as its' a ui consistency fix which is what that ticket was created for.

PR #1652, opened 2021-12-14
[NTV-94] Removal (most) of GraphSchema

msadoon:
# 📲 What

As the final step in our Apollo integration, here is a clean up of the model objects and query enums/structs that allowed us to use GQL without Apollo.

With this final PR, we no longer have any support for the GraphQL Schema and conclude the app's dependence on the schema. We are using Apollo in 100% of non-v1 calls in the app.

# 🤔 Why

Background in the JIRA [Epic](https://kickstarter.atlassian.net/browse/NTV-88).

# 🛠 How

Started by removing all the `Query` related enums and structs from `GraphSchema`.
Built project found dependencies to the schema, removed those with `Graph` prefixes, including their tests and template files.

A really good sign that none of these changes effect the existing functionality of the app is that no application logic has changed. Only models, tests, and templates (which are almost exclusively used in tests). No view model, view controller, view, networking layer has been changed to support these changes.

So that gives us confidence these are are deprecated files.

`GraphComment` and `GraphUser` were left in only in name, as they are referenced in envelopes that get created by `GraphAPI` `Data` (`PostCommentEnvelope` and `<UserEnvelope<GraphUser>` respectively). These objects have no connections to `GraphSchema` properties.

# ✅ Acceptance criteria

- [x] Run the app, back a project, leave a comment, manage a backing, select a new reward, see that it all works as expected.


KsApi/GraphSchema.swift
```diff
@@ -57,53 +35,6 @@ public func encodeToBase64(_ input: String) -> String {
...
-public enum Edges<Q: QueryType> {
-  case cursor
-  case node(NonEmptySet<Q>)
-}
-
-public enum Nodes<Q: QueryType> {
-  case nodes(NonEmptySet<Q>)
-}
-
 public struct GraphResponseError: Decodable {
```
singhhari, line 38: You can delete this, the `GraphResponseErrorEnvelope` and all the methods referencing them since these were specifically created for the old approach.

KsApi/MockService.swift
```diff
@@ -750,12 +750,6 @@
       return producer(for: self.fetchErroredUserBackingsResult)
     }
 
-    internal func fetchGraph<A>(
-      query _: NonEmptySet<Query>
-    ) -> SignalProducer<A, GraphError> where A: Decodable {
```
singhhari, line 755: The last reference to `GraphError` in this file is the `fetchCategory` method on line 1235. I don't see it being called anywhere so I think we can delete that too.

KsApi/Service+RequestHelpers.swift
```diff
@@ -6,19 +6,6 @@ import ReactiveSwift
...
-      forURL: self.serverConfig.graphQLEndpointUrl,
-      queryString: queryString
-    )
-
-    print("⚪️ [KsApi] Starting query:\n \(queryString)")
-    return Service.session.rac_graphDataResponse(request, and: self.perimeterXClient)
-      .flatMap(self.decodeGraphModel)
-  }
-
   func applyMutation<A: Decodable, B: GraphMutation>(mutation: B) -> SignalProducer<A, GraphError> {
```
singhhari, line 9: We can delete this method too. I don't see any callers.

singhhari (changes requested):
GraphCategory, GraphCountry, GraphComment, GraphMutationPostCommentEnvelope and GraphUser should all be removed as well but they will be a bit gnarly to approach.

singhhari:
Everything in the acceptance criteria is g2g!

KsApi/GraphSchema.swift
msadoon: Cannot remove this yet because its' used by `GraphError` which is used by `ErrorEnvelope` which is used by every endpoint in `Service` (v1 and apollo).

KsApi/MockService.swift
msadoon: Deleted `fetchCategory`, good call was not being used. But `GraphError` needs to still be in `GraphSchema` for the reason above.

msadoon:
> GraphCategory, GraphCountry, GraphComment, GraphMutationPostCommentEnvelope and GraphUser should all be removed as well but they will be a bit gnarly to approach.

Good call, I know `GraphCountry` was not being used anymore, so I removed it, however the others:

KsApi/Service+RequestHelpers.swift
msadoon: good call, also removing some `rac_decodingGraph` `Service+RequestHelpers` method.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

singhhari (approved):
Ran thru all the acceptance criteria again, lgtm!

PR #1663, opened 2022-01-28
[NTV-341] Image UI Element

msadoon:
# 📲 What

We want to display images in the campaign tab alongside our text components. Captions under images can be clicked and link externally to a Safari browser.

# 🤔 Why

Images are in the campaign tab of our web and android campaign tabs.

# 🛠 How

This was a very time consuming issue.

Some of the pitfalls of this feature is that images on the web are big (ie. have a max size of 200mb _each_). And a creator can upload any number of them of varying sizes.

`Kingfisher` was a good library (suggestion props @hadia) to use because it had a more robust cache manager than `AlamofireImage`. AlamofireImage is currently used in the app for on-demand image loading in cells using `af.setImage`, but in this use case wasn't ideal because projects (for example) with 60+ high-resolution images were not being cached at a high enough resolution to be zoomable/pannable (which we'll implement later).

However, the Kingfisher manages this problem by caching the high resolution image and retrieving it from cache pretty quickly (not sure exactly why, maybe its' just a bigger cache).

The other pitfall was the dynamic height of the image. The cell itself needed a fixed height before setting an image through `kf.setImage` which calls out to the network if the image isn't cached. So we can't set a fixed height on this cell because the images would be cut off. 

**Enter the prefetcher**

Kingfisher has a `ImagePrefetcher` that downloads images to the cache. Combined with the `ImageCache.default` we get a retrievable `UIImage` before the cell is displayed (or after, but reloaded). The basic idea is the project page view model will get the `ImageViewElements` right after the campaign tab is switched to. In those elements are `src` properties that are image urls. The view model sends those urls to the view controller for download. The VC uses the `ImagePrefetcher` and `ImageCache.default.retrieveImage` right after the image is either downloaded or cached by the prefetcher. The completion for the prefetcher sends back the `Data` of the retrieved image to the view controller, which then updates the datasource.

Use iOS 10's, `prefetchRowsAt` to preload the datasource with retrieved images as the system prepares rows to be displayed based on scrolling. (Uses technique above).

This works nicely because the `Data` ends up sent to the cell, and without a height, setting the data directly to an `UIImage` allows the image view's content size to reflect the size of the actual image. Doing this with `UIImageView.kf.retrieveImage` did not properly set the content size of the `UIImageView` leading to images loading but with 1 pixel height on the view port, unless reloaded.

One thing that could be improved is the repetitive calls on `prefetchRowsAt` delegate method when the view first loads. Noted in the code.

# 👀 See

 **Before 🐛** 

https://user-images.githubusercontent.com/4282741/153052869-7e7eb89e-b5ad-459a-a53a-6586052d6272.mp4

**After 🦋** 

 https://user-images.githubusercontent.com/4282741/153052144-939fe16f-6b91-43be-8849-94961797f525.mp4

# ✅ Acceptance criteria

- [x] Images loading as per [designs](https://www.figma.com/file/Vk262JvQlfddgdkfXN0rx0/Native%3A-Project-Page-C5-2021?node-id=166%3A5281).
- [x] Image captions clickable.

# ⏰ TODO

- [x] Fix issue with loaded image not refreshing table view cell.
- [x] Fix issue with image view port is larger in height than the image.
- [x] Fix padding in text view that has no text to zero.
- [x] Write tests.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #1691, opened 2022-05-16
NTV-1435 - Migrate Shipping V1 to GQL

msadoon:
# 📲 What

Fix the missing shipping badges. [Slack Thread](https://kickstarter.slack.com/archives/C3LS41US0/p1652715231204919).

"Shipping Worldwide"
"Limited Shipping"
"Only {country}"

These badges have been missing since we migrated our project model to gql. Also is connected to the data/UX for updating local receipt information.

# 🤔 Why

This [investigation](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/2003763209/iOS+Rewards+Shipping+Selector+Upgrade+to+GraphQL) was needed to understand the logic between the shipping data in the reward card and the badge related to shipping. [ui/ux](https://www.figma.com/file/QCpTglZe3PcprhI5TyCKmm/Local-Pickup-Reward?node-id=1338%3A18649)

Also all the places where shipping address selector, shipping label got its data. Will hopefully help with QA all possible scenarios when feature complete.

# 🛠 How

[gql logic](https://github.com/kickstarter/kickstarter/blob/d7f547e567c43bd40b7fb9eb69360141877f508e/app/decorators/backer_reward_decorator.rb#L120-L140) for returning shipping summary information.

If the reward isn't local, above logic states shipping summary can be used to display the badge data for shipping labels. This replaces the v1 `shipping_type` previously used.

Left the `shipping_type` in the existing v1 model because it causes crashes when v1's `discover` endpoint is called (app launch). Didn't do a deep dive but there are still use cases for `Reward` model `shipping` `type` field.

Anyway, the logic above also clarifies that if the reward has `localReceiptLocation` then the badge for shipping data should be hidden, and instead `shipping` `summary` is used for the `Reward Location` field.

# 👀 See

Before 🐛 

Simulator Screen Recording - iPhone 8 - 2022-05-16 at 18 43 33

After 🦋

Simulator Screen Recording - iPhone 8 - 2022-05-16 at 18 46 02

# ✅ Acceptance criteria

- [x] Shipping badges "Anywhere in the world", "Only ships to certain countries" and "Only ships to { country }" get displayed in app.

# ⏰ TODO
- [x] Update tests.

### DIFF
--- a/Library/ViewModels/RewardCardViewModel.swift
+++ b/Library/ViewModels/RewardCardViewModel.swift
@@ -294,32 +294,19 @@ private func backerCountOrRemainingString(project: Project, reward: Reward) -> R
 }
 
 private func shippingSummaryString(project: Project, reward: Reward) -> RewardCardPillData? {
-  if project.state == .live, reward.shipping.enabled, let type = reward.shipping.type {
-    switch type {
-    case .anywhere:
-      return RewardCardPillData(
-        backgroundColor: UIColor.ksr_create_700.withAlphaComponent(0.06),
-        text: Strings.Ships_worldwide(),
-        textColor: UIColor.ksr_create_700
-      )
-    case .multipleLocations:
-      return RewardCardPillData(
-        backgroundColor: UIColor.ksr_create_700.withAlphaComponent(0.06),
-        text: Strings.Limited_shipping(),
-        textColor: UIColor.ksr_create_700
-      )
-    case .noShipping: return nil
-    case .singleLocation:
-      if let name = reward.shipping.location?.localizedName {
-        return RewardCardPillData(
-          backgroundColor: UIColor.ksr_create_700.withAlphaComponent(0.06),
-          text: Strings.location_name_only(location_name: name),
-          textColor: UIColor.ksr_create_700
-        )
-      }
-
-      return nil
-    }
+  if project.state == .live,
+    reward.shipping.enabled,
+    let shippingSummaryText = reward.shipping.summary {
+    return RewardCardPillData(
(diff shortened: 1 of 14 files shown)


Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -294,32 +294,19 @@ private func backerCountOrRemainingString(project: Project, reward: Reward) -> R
...
-        return RewardCardPillData(
-          backgroundColor: UIColor.ksr_create_700.withAlphaComponent(0.06),
-          text: Strings.location_name_only(location_name: name),
-          textColor: UIColor.ksr_create_700
-        )
-      }
-
-      return nil
-    }
+  if project.state == .live,
```
iambchan, line 297: does this mean this info is hidden when the project is not live?

msadoon: Hey Bessie, yes, the project needs to be in `live` state to show the shipping labels. 
For all the base reward cards, (add-ons are only shown if project is live) we show 2 strings when the project isn't live (backer count and has add-ons) and two more if the project is live (shipping and time left).
Confirmed its the same on Android for non-line projects.
Screen Shot 2022-05-17 at 1 44 04 PM

msadoon:
Hope its cool if I merged this before you finished the review @iambchan . If you have further questions I can reopen this pr.

Library/ViewModels/RewardCardViewModel.swift
iambchan: thanks for clarifying!

iambchan (commented):
🥳

PR #1698, opened 2022-06-09
[PAY-1525] Local Pickup Optimizely Feature Flag

msadoon:
# 📲 What

This PR enables/disables the appearance of local rewards both in the rewards selection page and the add on selections page.

Note: Because we've done sort of a 1% on web, some people might have already backed local rewards. So to prevent those people from losing their local rewards on for example the manage pledge, they will still be shown as backed rewards if this ff is on. The only thing this flag prevents is new rewards and add-ons from being displayed to the user (if ff is off).

# 🤔 Why

To ensure we can turn on the feature at our convenience via optimizely after release qa is complete.

# 🛠 How

Boilerplate code that adds a new case to the `OptimizelyFeature` enum and then a helper function that checks user defaults before returning the value from the `Optimizely` SDK. Finally a case for the new `rewardLocalPickupEnabled` feature case in `OptimizelyFeatureFlagTools` 

# 👀 See

n/a feature flag on optimizely.

# ✅ Acceptance criteria

- [x] Test when flag is on/off, if rewards are shown to be local pickup/not.

# ⏰ TODO

- [x] Add filtering to rewards collection view (add ons collection view should use the reward from the collection to fetch add-ons, so if the reward was local and not allowed to be displayed to reward collection view, add on view will not list any of its local add-ons)


Kickstarter-iOS/Views/Controllers/RewardsCollectionViewControllerTests.swift
```diff
@@ -118,9 +128,126 @@ final class RewardsCollectionViewControllerTests: TestCase {
...
+        optimizelyClient: optimizelyClient
+      ) {
+        let vc = RewardsCollectionViewController.instantiate(
+          with: project,
+          refTag: nil,
+          context: .createPledge
+        )
+        _ = traitControllers(device: device, orientation: .portrait, child: vc)
+
+        FBSnapshotVerifyView(vc.view, identifier: "lang_\(language)_device_\(device)")
```
lukert33, line 185: Is the function `FBSnapshotVerifyView` the verification portion of these tests? Does it either return true or raise an error?

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -339,3 +339,38 @@ private func backingAndShippingTotal(for project: Project, and reward: Reward) -
...
+  }
+
+  let allowedNonbackedRewardsFromFeatureFlagRules = rewards.filter { reward in
+    allowedDisplay = true
+
+    if !featureRewardLocalPickupEnabled(),
+      isRewardLocalPickup(reward),
+      !isRewardBeingBacked(reward) {
+      allowedDisplay = false
+    }
```
lukert33, line 370: Am I reading this right? Here's my impression:
- If the RewardLocalPickupFeature is not enabled, do not display
- If the reward is meant for local pickup, do not display
- If the reward is not the reward presently being backed, do not display
-> otherwise, do display

If so, is that right?^

 ^ This comment is meant merely as a sanity check, and as a test of my swift reading skillz.

Library/ViewModels/RewardsCollectionViewModelTests.swift
```diff
@@ -64,6 +64,234 @@ final class RewardsCollectionViewModelTests: TestCase {
...
+
+      self.reloadDataWithValues.assertDidNotEmitValue()
+
+      self.vm.inputs.viewDidLoad()
+
+      self.reloadDataWithValues.assertDidEmitValue()
+
+      let displayableRewards = self.reloadDataWithValues.lastValue?.compactMap { $0.reward }
+
+      XCTAssertEqual(displayableRewards, project.rewards)
```
lukert33, line 87: nice! This is easy to follow.

lukert33 (commented):
👋  First time swift reviewer here! Based on your tests in `Library/ViewModels/RewardsCollectionViewModelTests.swift` this looks g2g to me. 

That said, I left some comments just to confirm that I was seeing what I think I was seeing. If you don't mind replying to those, I'll take a second look, hopefully to approve.

I'm looking forward to gaining some competence in reviewing iOS code!

Kickstarter-iOS/Views/Controllers/ProjectPageViewControllerTests.swift
```diff
@@ -607,7 +607,7 @@ internal final class ProjectPageViewControllerTests: TestCase {
           parent.view.frame.size.height = 2_300
         }
 
-        self.scheduler.run()
+        self.scheduler.advance(by: .seconds(5))
```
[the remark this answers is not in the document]
msadoon: Hey Luke, yes so this is not related to the changes for the feature but was a flakey test that I'm _hoping_ gets fixed on CI with a longer delay. I suspect that the `run()` doesn't advance the clock by enough for the signals that update the UI in the snapshot.
I also think I've tried this before, so it'll probably fail on CI. It's something I hope to take a longer look at in a cooldown/tech-debt party.

Kickstarter-iOS/Views/Controllers/RewardsCollectionViewControllerTests.swift
msadoon: It's part of our snapshot testing framework `github "uber/ios-snapshot-test-case" == 6.2.0` which is a dependency in our `Cartfile`. (Carthage is what we use to manage third party dependency alongside Swift Package Manager).

msadoon: Its' just a function that creates snapshots and stores them in the `Screenshots` folder with the `identifier` listed.

Library/ViewModels/RewardsCollectionViewModel.swift
msadoon: Appreciate the sanity check, so the logic here is meant to be inclusive for all rewards of a project. 

We only exclude a reward from the reward selection screen if _all_ these conditions are **true**.
- the feature flag is _off_ AND
- the reward is local AND
  - (`isRewardLocalPickup()` defines this as `Reward.shippingPreference` == `.local` && `Reward.localPickup != nil` && `Reward.shipping.enabled` == `false`) 
- the reward is _not_ part of `Project.personalization.backing`

if any of the above is false, the reward is shown. For example the ff might be off and the reward might be local, but it is backed, so we show it. Normally, if the reward was not backed, we wouldn't show it. But it makes sense here to show because user has already backed a reward and should still be able to manage it even if we decide not show local rewards for selection on all projects.

msadoon:
@lukert33 gonna merge this guy but happy to answer any more questions here. Just because I want to fire off a beta build.

PR #1700, opened 2022-06-23
[NTV-564] Fix Flakey Test on Kickstarter-Framework-iOSTests

msadoon:
# 📲 What and 🤔 Why

This test `Kickstarter-Framework-iOSTests/ProjectPageViewControllerTests/testErrorFetchingProject` has been failing on multiple occasions and caused multiple of hours of waiting for CI builds to pass.

As this is a cooldown, fixing this test so it reliably passes on CI is going to save us many painful hours of waiting.

Useful because when we run CI, it takes ~ 20 min for `Kickstarter-Framework-iOSTests` to go through all its steps and run all the tests. So when one fails (ie this one), the entire suit has to be run again.

# 🛠 How

Turns out we were relying on an Alamo Networking extension to download a creator image url and set it to the creator image/name in the project overview tab. We shouldn't be relying on real network requests when running tests, as the timing is asynchronous and will always be different everytime the request is made. So assuming that's the problem, setting the creator image url string in the project to an empty string should display the `nil` image in the project page every time.

# ✅ Acceptance criteria

- [ ] Run CI on this PR multiple times (maybe up to 10 times), even if all jobs pass. Its' flakey so sometimes it'll pass, sometimes fail.


msadoon:
@Arkariang no need to review, you can if you want to, but because its merging to `main` I have to put in one reviewer. Mostly likely will run through Acceptance Criteria and merge with administrator privileges.

msadoon:
@lukert33 just in case you're curious as to how this was fixed, because I know we paired on it a little.

Kickstarter-iOS/Views/Controllers/ProjectPageViewControllerTests.swift
```diff
@@ -590,14 +590,23 @@ internal final class ProjectPageViewControllerTests: TestCase {
...
 
+    // This test was previously flakey on CI because it relied on Alamofire to download an image url to populate the user image, which may/may not be ready in time.
+    let projectWithNoUserImageURL = self.project
+      |> Project.lens.creator .~ (
+        User.template
+          |> \.avatar.large .~ ""
+          |> \.avatar.medium .~ ""
+          |> \.avatar.small .~ ""
+      )
+
```
lukert33, line 601: So the solution here is to blank out the image, right? Did you have to regenerate the existing snapshot that's used as a comparison?

Kickstarter-iOS/Views/Controllers/ProjectPageViewControllerTests.swift
```diff
@@ -590,6 +590,15 @@ internal final class ProjectPageViewControllerTests: TestCase {
       fetchProjectPamphletResult: .failure(.couldNotParseJSON)
     )
 
+    // This test was previously flakey on CI because it relied on Alamofire to download an image url to populate the user image, which may/may not be ready in time.
```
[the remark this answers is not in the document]
msadoon: Yea that's actually a bug, Alamofire does have the functionality to add a placeholder that isn't being used right now. I'm hesitant to change that because it's an application change that will affect that page rendering. Very small and probably an improvement, but I just want to quickly fix the tests here to suit the existing application logic, reason being is because it needs further testing and some tests may need re-recording.

Kickstarter-iOS/Views/Controllers/ProjectPageViewControllerTests.swift
msadoon: Yes the latest [commit]([9b629f3](https://github.com/kickstarter/ios-oss/pull/1700/commits/9b629f35db244a6696ad17e411fa95ceee83880d) has the newly recorded snapshots. 

When this pr was first made ready to review, the reference snapshot had the error which was a creator url, blanked out by the code above. So that is why it failed. I assumed it was recorded correctly and that was a mistake. Turns out more than one reference snapshot had the creator avatar recorded. So technically those were flakey too although for some reason this one test `testErrorFetchingProject` was the one failing the most.

The current application logic as mentioned in response to Isabel's comment is to display the creator url only if it is a valid url. We don't at the moment support an in-app image placeholder before this url is downloaded (or if its not there). That being the case, without updating in-app logic, its safest to send in a blank url string and not have the image show anything. Not perfect, but the point of these tests is to update that Pledge Container CTA not to test the creator/project images.

Anyway, tldr: yes snapshot tests needed re-recorded after nulling out images on all offending tests (3)

PR #1704, opened 2022-07-18
[PAY-1757] Upgrade the Stripe SDK from 18 to 22

msadoon:
# 📲 What

Having done an [investigation](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/2056519681/Stripe+Payment+Sheet+Preliminary+Investigation+iOS) on how the existing Stripe/Apple Pay/Payment flows work inside the app, we have a lot of context describing the existing feature. This doesn't yet include any upgrades to Stripe/integration of the payment sheet/updates to apple pay/updates to the `createBacking`/`updateBacking` calls.

# 🤔 Why

We need to do this as not only technical debt/maintenance of the iOS application but to use the latest version of the Payment Sheet provided by Stripe.

# 🛠 How

Firstly we remove stripe-sdk 18 from carthage. Then we add it through SPM.

`KsAPI` and `Library-iOS` both compile fine and tests pass with new Stripe SPM version 22.

`Kickstarter-iOS` is the only remaining framework that fails to build. Here are some details about how the Stripes' changes effect our existing properties.

`STP` is the prefix Stripe uses, not us.

**Case 1**

Affected files:
`AddNewCardViewController`
`AddNewCardViewModel`

`STPCardBrand` has updated its enum case name from `masterCard` to `mastercard`: 
  - Just a name change here.

`STPAPIClient` `shared()` to `shared`.  Objective-C singleton to Swift singleton on Stripes' side. 
  - Just a name change here.

`STPPaymentCardTextField` has changed its return type of `.expirationMonth` and `.expirationYear` from `NSUInteger` to `Int`:  
  - (Obj-C) to `Int` (Swift).
  - Existing `CardDetails` using `Month` and `Year` typealias, which was `UInt`, but `Int` is not convertible to `UInt`. So wrapping the `Int` value in `NSNumber` does not lose us any precision. That allows us to use the existing `Month` and `Year` properties of our in-app `CardDetails` `PaymentDetails` if they are changed from `UInt` to `NSNumber`.
    - With `CardNumber` and `PaymentDetails` updated for `NSNumber` now (because they use `Month` and `Year`) we a conversion of `NSNumber` to `UInt` inside where `PaymentDetails` is implement (`createStripeToken` uses `STPCardParams` which still uses `Unit`). That conversion is loss-less. `UInt` is larger than `Int`, and the `NSNumber` acts as a bridge between `Int` and `UInt`. Apple Docs [info](https://developer.apple.com/documentation/foundation/nsnumber).

**Case 2**

Affected files:
`PledgeViewController`

Missing `PK` delegate methods, view controllers and objects.
  - The new Stripe framework doesn't include `PassKit`, (might've been moved to their `StripeApplePay` SDK).
  - Solution was simple: `import PassKit`

`appleMerchantIdentifier` and `publishableKey`
  - These have been moved from `STPPaymentConfiguration.shared` to `STPAPIClient.shared` so setting these values on `STPAPIClient.shared` instead.

`STPPaymentHandler.shared().confirmSetupIntent`
  - Update to parameter names in this function
 
`STPAPIClient.shared()` changed to Swift singleton `STPAPIClient.shared`

**Case 3**

Affected files:
`STPPaymentHandler+StripePaymentHandlerActionStatus`

  - `STPPaymentHandler.ActionStatus` is now a ObjC compatible Swift enum with the name `STPPaymentHandlerActionStatus`

**Case 4**

Affected fiels:
`AddNewCardViewControllerTests` 

We found an issue included `STPPaymentCardTextField` within our snapshot test. It _somehow_ isn't properly linked through XCode. Currently an open [issue](AddNewCardViewControllerTests) with Stripe because their depdendencies need to be updated so library resource can be accessed through a test framework. Unlikely an issue with our project setup because we didn't change anything in our tests, and the app builds and runs and uses the `STPPaymentCardTextField` just as before.

Here are some investigation notes:

We're gonna see a drop in testing, because we need to exclude this file for now while Stripe fixes (or tells us what to fix) on our end.

** Case 5**

Affected files:
`AddNewCardViewController`

Current:

Updated:
Screen Shot 2022-07-20 at 4 29 50 PM

As you can see, the `zip` field is duplicated, so better correct this to not show duplicate UI.

Fix for now was to set `self.creditCardTextField.postalCodeEntryEnabled = false`. Stripe should handle zip codes for us, but we will be deprecating this view controller in the near future so this setting is a shortcut to use with our existing zip code view.

# ✅ Acceptance criteria

- [ ] Existing pledge flows/Apple Pay/anything that touches a component of the Stripe library still works.

# ⏰ TODO

- [x] Fix all broken/uncompiling code, new conventions around the Stripe library.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Views/Controllers/AddNewCardViewController.swift:363:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

msadoon:
merging this one before fully doing the acceptance testing so the alpha build can be tested on a real device.

PR #1706, opened 2022-07-26
[PAY-1763] Present Stripe Payment Sheet On Pledge View

msadoon:
# 📲 What

As part of our further integration with Stripe we are adding the new payment sheet to allow payment method creation that offloads the work to Stripe (to support more payment methods) and removes the need for the client apps to call `CreatePaymentSourceMutation` whenever we want to add a new payment source.

The only thing the client apps own is the setup intent created for the Stripe payment sheet. Which is a replacement for the mutation above, but since it allows us to use Stripe's built in payment sheet and maybe later address element, it makes sense to hand responsibility of this work to them.

The setup intent the client side apps provide to ensure the payment sheet created payment method exists on Stripe's backend before it creates/updates the backing on our backend.

# 🤔 Why

Reasons outlined above for payment sheet integration. 

# 🛠 How

We're following [these](https://stripe.com/docs/payments/accept-a-payment?platform=ios&ui=payment-sheet#ios-flowcontroller) instructions to present the payment sheet. But after the sheet is presented, we are not handling any operations after the continue button is pressed. Simply just presenting using `PaymentSheet.FlowController.create`

We're handling the completion of the process here which involves: `paymentSheetFlowController.confirm` and modifying the existing payment cards to include the new `paymentOption` passed back from the `PaymentSheet.FlowController`.

# 👀 See

Before 🐛 

https://user-images.githubusercontent.com/4282741/181631646-794fca4f-96ba-4954-8b75-46a8cb1a306a.MP4

After 🦋

Simulator Screen Recording - iPhone 8 - 2022-07-28 at 16 25 10

# ✅ Acceptance criteria

- [x] Payment sheet appears after user hits "Add New Card"

All we're doing in this PR is presenting the Stripe payment sheet. 

# ⏰ TODO

- [x] Display payment sheet.
- [x] Handle edge where there might be an error in displaying it.

### DIFF
--- a/Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
@@ -124,13 +124,8 @@ final class PledgePaymentMethodsViewController: UIViewController {
       |> \.delegate .~ self
     addNewCardViewController.configure(with: intent, project: project)
     let navigationController = UINavigationController.init(rootViewController: addNewCardViewController)
-    let offset = navigationController.navigationBar.bounds.height
 
-    if #available(iOS 13.0, *) {
-      self.present(navigationController, animated: true)
-    } else {
-      self.presentViewControllerWithSheetOverlay(navigationController, offset: offset)
-    }
+    self.present(navigationController, animated: true)
   }
 }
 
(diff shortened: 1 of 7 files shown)


Kickstarter-iOS/Views/Controllers/PledgePaymentMethodsViewController.swift
```diff
@@ -124,13 +124,8 @@ final class PledgePaymentMethodsViewController: UIViewController {
       |> \.delegate .~ self
     addNewCardViewController.configure(with: intent, project: project)
     let navigationController = UINavigationController.init(rootViewController: addNewCardViewController)
-    let offset = navigationController.navigationBar.bounds.height
 
-    if #available(iOS 13.0, *) {
-      self.present(navigationController, animated: true)
-    } else {
-      self.presentViewControllerWithSheetOverlay(navigationController, offset: offset)
```
msadoon, line 132: our minimum version is iOS 13, so this code path will never be hit.

PR #1710, opened 2022-08-09
[Unticketed] Fix With Device Framework Link

msadoon:
Turns out the latest beta build 5.4.0 (16600083) crashes because there is a `Stripe` library in `Link Binary with Libraries` in `Build Settings` of `Kickstarter-Framework-iOS` that doesn't need to be there because `Kickstarter-Framework-iOS` includes `Library-iOS`, which includes `Stripe`.


msadoon:
merging to generate beta build.

PR #1715, opened 2022-08-30
[WEB-697] Adds new Facebook Login Deprecation feature flag

scottkicks:
# 📲 What

This PR makes the necessary code changes to add a new 'Facebook Login Deprecation' feature flag. 

# 🤔 Why

This will allow us to slowly rollout a new flow that allows our facebook users to make sure there is a password associated with their KS account. This way when we remove Facebook Login entirely, everyone will still be able to access their accounts.

# 🛠 How

Boilerplate code that adds a new case to the `OptimizelyFeature` enum and then a helper function that checks user defaults before returning the value from the `Optimizely` SDK. Finally a case for the new `facebooLoginDeprecationEnabled` feature case in `OptimizelyFeatureFlagTools` 

# 👀 See

Optimizely Feature Flag menu in our Beta Tools

| After 🦋 |

|  |

# ✅ Acceptance criteria

- [ ] Make sure the new 'Facebook Login Deprecation' flag shows in "Optimizely Feature Flags" in our Beta tools screen

# ⏰ TODO
- [ ] Setup feature flag in Optimizely


msadoon (changes requested):
To me these changes look good!
I will run the app and see the option to turn off/on the flag once we get that setup in Optimizely later today and then approve!
Only change required is to fix the broken test suite.

msadoon (approved):
Good to go!

PR #1717, opened 2022-09-01
[NTV-605] Alamofire/AlamofireImage Carthage -> SPM

scottkicks:
# 📲 What

As we migrate over to M1 machines for local development, we'll want to move over our Carthage dependencies into Xcode dependency manager where we can.

# 🤔 Why

There are many benefits of using SPM over Carthage, but the main reason is to avoid handling manual compilation scripts to support multiple architecture types like this one found inside `carthage.sh`
```
# For Xcode 13 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
# the build will fail on lipo due to duplicate architectures.

CURRENT_XCODE_VERSION=$(xcodebuild -version | grep "Build version" | cut -d' ' -f3)
echo "EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1300__BUILD_$CURRENT_XCODE_VERSION = arm64 arm64e armv7 armv7s armv6 armv8" >> $xcconfig

echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1300 = $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64__XCODE_1300__BUILD_$(XCODE_PRODUCT_BUILD_VERSION))' >> $xcconfig
echo '
```
As you can see it's verbose, very cryptic, and we're not entirely sure if we can add M1 support to strings like `EXCLUDED_ARCHS`

At least for now, we're going to move all dependencies that we can over to SPM that are listed inside the `Cartfile`.
For now we'll do the 3rd party dependencies. There's also binaries which aren't built as schemas (the step failing on `make bootstrap` with M1 because of the code snippet above). Finally there's internal libraries that do get converted to schemas, which we might need to create custom SPM packages out of as we did with `Kickstarter-Prelude`.

# 🛠 How

- Removed `Alamofire` and `AlamofireImage` from `Cartfile`.
- Removed `Alamofire` and `AlamofireImage` from `Cartfile.resolved`
- Remove references inside `Kickstarter/Frameworks` folder and within each target.
- Remove references inside `Carthage-xcfilelist`
- Add `https://github.com/Alamofire/AlamofireImage.git` to Package Dependencies version 4.2
- Add package once downloaded to `Library` target
   - SPM downloads and caches the library, but it's still needed wherever Alamofire was included before, in this case the only place was "Library" target.
- Ensure app builds, tests pass.
- Set breakpoints for where import statements are `import AlamofireImage`, set breakpoints for where framework references are, run app and hit breakpoints. Continue running app and smoke tests where framework call sites occur.

# ✅ Acceptance criteria

- [x] Breakpoint most instances of code used in files that `import AlamofireImage` and run the app on device to check the breakpoints are hit and the app continues to run as expected.


msadoon:
Hey hope you don't mind, I did the first one from start to finish, just to see what the process was like. It's a bit hard reading the project files and understanding the full context of the changes.

Turns out the step missing here (that generated that `x86_64` architecture not found issue) was not including the `AlamofireImage` package inside `Library`. We also should generally be careful about not including this package in frameworks that don't need it (ie. `Kickstarter-FrameworkiOS` has references to `AlamofireImage` but also includes `Library` which can have its' own dependency for `AlamofireImage` package)

I'm going to try to run through as many of these as possible to get the project in a running state on M1.
It's not hard, just a lot of manual changes. The thing is I don't know what we are going to do with the binaries or internal libraries yet. My hunch is the `make bootstrap` step is not going through those 
`Building scheme "FBSnapshotTestCase iOS" in FBSnapshotTestCase.xcworkspace` steps for the binaries, so we can leave them in Carthage for now.
The only internal library we have is `Kickstarter-ReactiveExtensions` which can be added to SPM by giving it a Package.resolved file. (There might be issues building it in Xcode though...hopefully not.)

Lots of unknowns.

scottkicks:
@msadoon Nice find! Thanks for the extra info too. I'll make sure to include it in Library and smoke test accordingly.

msadoon (changes requested):
Yea so when I did this, you only needed to bring in `AlamofireImage` not `Alamofire`.
So just make that change in the Package Dependencies and I'll run the app on device, do some testing and approve.

msadoon (approved):
good to go!

PR #1726, opened 2022-09-12
[NTV-615] Facebook-ObjC-SDK Carthage -> SPM

scottkicks:
# 📲 What

As part of  our Carthage -> SPM migration we need to migrate Optimizely

# 🤔 Why

# 🛠 How

- Remove the dependencies from Cartfile and Cartfile.resolved and add them to Package.resolved
- Use minimum version would have been 6.5.2 to match Carthage, but there were some missing symbols in that version of the new `facebook-ios-sdk` that we need to use now. Those issues have been resolved, but required setting the minimum version to 9.0.0
- Resolve any merge conflicts

# ✅ Acceptance criteria

- [x] Breakpoint most instances of code used in files that `import FBSDKCoreKit` & `import FBSDKLoginKit` and run the app on device to check the breakpoints are hit and the app continues to run as expected. Specifically in the Feature Flag viewer


msadoon (approved):
Good to go!

I didn't do any on device testing on this, you might have, but there seems to be lots of places to breakpoint and check the app for regressions (esp with the verison bump).

Screen Shot 2022-09-13 at 8 15 12 AM

I'll leave this to you though, you can own the facebook stuff as you are working on deprecating it 💯

scottkicks:
> ... Also not 100% sure all those imports are being used.

Nice catch. I  was able to remove a couple of unused imports. I'm still working on testing the Find Friends FB flow. Once that looks good I will merge

PR #1728, opened 2022-09-13
[NTV-616]  Moves FBSnapshotTestCase to iOSSnapshotTestCase Swift Package

scottkicks:
# 📲 What

As part of  our Carthage -> SPM migration we need to migrate FBSnapshotTestCase

# 🤔 Why

# 🛠 How

- Remove the dependencies from Cartfile and Cartfile.resolved and add them to Package.resolved
- Minimum version would have been 6.2.0 to match Carthage, but there were some missing symbols in that version of the new `iOSSnapshotTestCase` that we need to use now. Those issues have been resolved, but required setting the minimum version to 8.0.0
- Resolve any merge conflicts

# ✅ Acceptance criteria

- [x] Breakpoint most instances of code used in files that `import iOSSnapshotTestCase` and run the app on device to check the breakpoints are hit and the app continues to run as expected. This package is basically just used in test cases FYI.


.gitignore
```diff
@@ -41,6 +41,7 @@ playground.xcworkspace
 # Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
 # Packages/
 .build/
+Package.resolved
```
scottkicks, line 44: Since a new `Package.resolved` is created with each build, I'm adding this to  .gitignore

msadoon (approved):
Good work!

I made one commit as I noticed the `iOSSnapshotTestCase` was being included in `Library`, `Library-Tests` and `Kickstarter-Framework-Tests` targets. It only gets used on `TestCase.swift`, which is included in both the test frameworks but not in `Library`, so took it out of there.

It's pretty easy to test this one as it only affects tests.

PR #1734, opened 2022-09-20
[WEB-742] Migrates PerimeterX to SPM

scottkicks:
# 📲 What

As part of our Carthage -> SPM migration we need to migrate PerimeterX

# 🤔 Why

# 🛠 How

- Remove the dependencies from Cartfile and Cartfile.resolved and add them to KsApi target
- Minimum version to 1.13.9
- Resolve any merge conflicts

# ✅ Acceptance criteria

- [x] Breakpoint most instances of code used in files that `import PerimeterX` (KsAPI Target) and run the app on device to check the breakpoints are hit and the app continues to run as expected.. 

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -9,4 +9,3 @@ github "ReactiveCocoa/ReactiveSwift" == 6.5.0
 ### Binaries
 
 binary "https://raw.githubusercontent.com/Appboy/appboy-ios-sdk/master/appboy_ios_sdk_full.json" == 4.3.2
-binary "https://raw.githubusercontent.com/PerimeterX/px-iOS-Framework/master/PerimeterX.json" == 1.13.9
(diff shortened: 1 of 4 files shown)


msadoon (approved):
Looks good, good job, just remove it from the main `Kickstarter-iOS` app target. It only needs to be in `KsAPI`. 

(Oddly though, it was never included in any of the targets when it was a binary but was still referenced somewhere)

Anyway, including it it only `KsApi` target is totally fine for now.

scottkicks:
> (Oddly though, it was never included in any of the targets when it was a binary but was still referenced somewhere)

@msadoon Yea I noticed the same thing with the Appboy-iOS-SDK binary. Thanks i'll update the main target. I forgot that it gets cached there.

PR #1737, opened 2022-09-21
[WEB-741] Moves Appboy-iOS-SDK and its dependency SDWebImage to SPM

scottkicks:
# 📲 What

As part of our Carthage -> SPM migration we need to migrate Appboy-iOS-SDK (AppboyKit) and its dependency SDWebImage

# 🤔 Why

# 🛠 How

- Remove the dependencies from Cartfile and Cartfile.resolved and add them to main and Library targets
- Minimum version to 4.3.2
```
AppDelegate.swift
import Appboy_iOS_SDK -> import AppboyKit

AppDelegateViewModel.swift
import Appboy_iOS_SDK -> import AppboyKit

BrazeTypes.swift
import Appboy_iOS_SDK -> import AppboyKit

SharedFunctions.swift
import Appboy_iOS_SDK -> import AppboyKit
```

# ✅ Acceptance criteria

- [x] Breakpoint most instances of code used in files that `import AppboyKit` (KsAPI Target) and run the app on device to check the breakpoints are hit and the app continues to run as expected.. 

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -6,6 +6,3 @@ github "kickstarter/Kickstarter-ReactiveExtensions" "e3f7786b5bcc7b99c14b9fd3133
 
 github "ReactiveCocoa/ReactiveSwift" == 6.5.0
 
-### Binaries
-
-binary "https://raw.githubusercontent.com/Appboy/appboy-ios-sdk/master/appboy_ios_sdk_full.json" == 4.3.2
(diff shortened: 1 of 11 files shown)


msadoon (changes requested):
Hey @scottkicks I did the work on a separate branch, and used a feature-proof SDK instead [here](https://github.com/braze-inc/braze-swift-sdk). As that page mentions the `appboy-ios-sdk` is going to be deprecated so we can use the Braze Swift SDK instead (which includes the `AppboyKit` and `AppyboyUI` libraries) Also no need to specifically include `SDWebImage` as I think it gets included as a dependency.

I would say switch the branch on this pr to `fix/appboy-sdwebimage-to-spm` and do the on-device testing on it to make sure we didn't break anything. I can help with this too, when you get to this stage let's pair on it.

PR #1738, opened 2022-09-21
[PAY-1954] Loading User Settings And Cancelling Payment Sheet

msadoon:
# 📲 What

Due to the time required to get a setup intent token and for show the payment sheet. Also after the payment sheet is dismissed there is a lag between showing the newly created payment method. That lag is caused by the network to create the new payment source. We might want to let the user know that their actions are being process via a "loading" state.

There's also the necessity to cancel the payment sheet, if its about to be presented but another user action occurs to cancel its appearance (ie. click on "Add new card", click on "Edit" while waiting for the sheet to be presented.)

# 🤔 Why

User should know that any action taken while loading the payment sheet, cancels its appearance.
User should also know any action taken to change the state of the page will cancel the sheet's appearance.

# 🛠 How

New output on `PaymentMethodsViewModel` `cancelAddNewCardLoadingState` to inform the `PaymentMethodsFooterView` through a new delegate that only cancels the loading state on that view.
The view itself has a button that whenever clicked and the feature flag is on, will show the loading state.

In the view model, all actions that cancel the loading state on the view go through `shouldCancelPaymentSheetAppearance` to update `cancelAddNewCardLoadingState` if they are `true`. And they are `true` only when user clicks, Edit, deletes an existing card or the payment methods get updated.

# 👀 See

Before 🐛

https://user-images.githubusercontent.com/4282741/191841558-37f784dc-4d0c-48aa-ac30-1be5d800e04c.mp4

After 🦋

https://user-images.githubusercontent.com/4282741/191840654-7c149c21-f433-47cc-a0e8-b46956214093.mp4

|  |  |

# ✅ Acceptance criteria

- [x] Add new card shows a loading state before and after the payment sheet appears.
- [x] Before payment sheet appears the loading state and appearance of the payment sheet can be cancelled by another user action (ie. Edit, delete of a card).

# ⏰ TODO

- [x] Add loading state to "Add new card"
- [x] Write tests.


Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -183,10 +185,20 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
...
 
-    /** TODO: https://kickstarter.atlassian.net/browse/PAY-1954
-     * Add cancellation signal similiar to `shouldCancelPaymentSheetAppearance` in `PledgePaymentMethodsViewModel`
-     */
+    self.shouldCancelPaymentSheetAppearance <~ self.didTapAddCardButtonProperty.signal.ignoreValues()
+      .mapConst(false)
+
+    self.cancelAddNewCardLoadingState = self.shouldCancelPaymentSheetAppearance.signal.filter(isTrue)
+      .ignoreValues()
+    self.shouldCancelPaymentSheetAppearance <~ self.addNewCardSucceededProperty.signal.ignoreValues()
```
scottkicks, line 193: newbie Reactive question probably, but I see that we're setting `self.shouldCancelPaymentSheetAppearance` twice here. Is there a Reactive way to combine them into a single line? Or is there a benefit to having them split separated like this?

Library/ViewModels/PaymentMethodsViewModel.swift
```diff
@@ -195,6 +207,9 @@ public final class PaymentMethodsViewModel: PaymentMethodsViewModelType,
     )
     .map { $0.localizedDescription }
 
+    self.shouldCancelPaymentSheetAppearance <~ self.errorLoadingPaymentMethodsOrSetupIntent.signal
```
scottkicks, line 210: MI wonder if there is a way to assign all of these signals to `shouldCancelPaymentSheetAppearance` in one line 🤔 

Or maybe we can group these lines together so that it's a little easier to see all of the states that `shouldCancelPaymentSheetAppearance` reacts to. Might make it easier to debug in the future. Not a big deal

Library/ViewModels/PaymentMethodsViewModelTests.swift
```diff
@@ -174,6 +176,31 @@ internal final class PaymentMethodsViewModelTests: TestCase {
...
+      self.cancelLoadingState.assertDidNotEmitValue()
+
+      self.vm.inputs
+        .shouldCancelPaymentSheetAppearance(state: true)
+
+      self.scheduler.advance()
+
+      self.cancelLoadingState.assertDidEmitValue()
+    }
+  }
```
scottkicks, line 202: Would it be useful to add tests to ensure that the `addCardButton` is is/isn't hidden or interact-able at the right times? i.e. when`featureSettingsPaymentSheetEnabled()`

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodsViewController.swift
```diff
@@ -45,6 +50,7 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
     ]
 
     self.dataSource.deletionHandler = { [weak self] creditCard in
+      self?.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: true)
```
scottkicks, line 53: I noticed that we're not passing `false` to `shouldCancelPaymentSheetAppearance(state: Bool)` other than in the test. Can we remove this param and assume that if whenever this input is called, we should cancel the payment sheet?

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodsViewController.swift
```diff
@@ -145,6 +158,7 @@ internal final class PaymentMethodsViewController: UIViewController, MessageBann
 
   @objc private func edit() {
     self.viewModel.inputs.editButtonTapped()
+    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: true)
```
scottkicks, line 161: Should we also call this if a user taps the `< Account` button in the nav bar?

Library/ViewModels/PaymentMethodsViewModelTests.swift
msadoon: The only time the Add New Card button isn't interactable is when the loading wheel is showing, but its' hard to get at the actual `UIButton` itself, as it is the `tableView'`'s `footerView` which is `private` to the view controller.

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodsViewController.swift
msadoon: We don't send in `false` via the view controller, you're right but we do send it in through the view model `PaymentMethodsViewModel`. In fact it's default state is `false`. Also those signal's we merged in the above comment, one of them does map `false` to the `shouldCancelPaymentSheetAppearance` which is convenient. 

So just stepping back, see this as a signal that acts as an `input` _outside_ the view model and and `input` _inside_ the view model`.

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodsViewController.swift
msadoon: So good catch, if you find this to be the case in testing then yes we should do so.
However in my testing I didn't find that popping the page off the stack caused the in-flight payment sheet to come up. I'm assuming it's because the delegates are released and there are no more references to the payment sheet.

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodsViewController.swift
scottkicks: Ah I see. Cool thanks

scottkicks (approved):
LGTM 👍

PR #1741, opened 2022-10-03
[PAY-1990] Fix Crash On Apple Pay Pledge Page

msadoon:
# 📲 What

A crash was introduced with the new loading states for the payment sheet. An untested flow allowed the pre-mature update of a section before it had any visible data. This causes a crash related to the table view being out-of-sync with its' data source when reloading.

# 🤔 Why

We don't want to break this page in any way, even if we're introducing a feature hidden behind a flag. All pre-existing flows must work as they always did or conversion rates will go down.

# 🛠 How

Before we simply took the last section in the table view as "Add New Card" section. This was incorrect because the last section in the table view is not the "Add New Card" section (even though that is the last one we saw visually). It was the "Loading/Shimmer" state. 

The key to understanding this is to know how `ValueCellDataSource` creates the sections. When it calls its' `set` method, it also calls its' `padValuesForSection` method. Which uses the section integer value (an enum raw value in this case `PaymentMethodsTableViewSection`) to fill out the empty sections between 0 and the section integer value.

That means, the section has been created for `Add New Card`, but during loading it isn't visible because its datasource values are empty but _do exist_ within the UI for the tableview at the time of reload.

We just need to check that the reload of that section is updating an actual non-empty value. Again not necessary, but why update something that doesn't exist twice?

# 👀 See

Before 🐛 

https://user-images.githubusercontent.com/4282741/193699506-bba5deb2-cb34-40d2-8baa-5e97f514b32f.MP4

After 🦋

https://user-images.githubusercontent.com/4282741/193699691-67c369d0-cff9-4a0b-a8d9-05543eab762b.MP4

# ✅ Acceptance criteria

- [x] Ensure hitting the Apple Pay button doesn't crash the app while the payment methods are loading.


sclampet (approved):
Nice catch 👍🏻

msadoon:
I think @sclampet you approved this, but I didn't assign that id to the PR initially and re-assigning lost the approval. Hopefully its' not a huge deal to merge this as admin.

PR #1746, opened 2022-10-12
[PAY-1980] Update Backing for Errored Pledge Flow

msadoon:
# 📲 What

Continue development of using Stripe's payment sheet on the pledge view page.

# 🤔 Why

We need to cover the final context `.fixPaymentMethod` for using the payment sheet.

# 🛠 How

So removed the context logic and the only remaining logic is to use the feature flag to expose the payment sheet vs the add new card page.

# 👀 See

Before 🐛 

...coming...

After 🦋

...coming...

# ✅ Acceptance criteria

- [ ] Fixed errored pledge flow shows up and successfully updates an errored backing. Preferrably do this on device.
- [ ] As a final check, ensure all contexts show payment sheet when feature flag is on. Off otherwise.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,7 +1,5 @@
 ### Internal
 
-github "kickstarter/Kickstarter-ReactiveExtensions" "e3f7786b5bcc7b99c14b9fd313302bb59d9c3fe9"
-
 ### 3rd Party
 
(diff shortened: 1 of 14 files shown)


msadoon:
Based branch off incorrect base branch, should have been `main`. Will redo with only necessary changes.

PR #1747, opened 2022-10-12
[PAY-1980] Update Backing Errored Pledge Flow

msadoon:
# 📲 What

Continue development of using Stripe's payment sheet on the pledge view page.

# 🤔 Why

We need to cover the final context `.fixPaymentMethod` for using the payment sheet.

# 🛠 How

So removed the context logic and the only remaining logic is to use the feature flag to expose the payment sheet vs the add new card page.

# 👀 See

Before 🐛 

Didn't record this as the old flow is a little difficult to co-ordinate. Requires backend support to close project, but backing into errored state.

After 🦋

Success flow:

https://user-images.githubusercontent.com/4282741/195427591-462177b8-74c8-4154-b1dc-3380cc3df107.MP4

Errored flow:

https://user-images.githubusercontent.com/4282741/195429889-cd544a9b-313b-4802-8846-94c9858f10f8.MP4

# ✅ Acceptance criteria

- [x] Fixed errored pledge flow shows up and successfully updates an errored backing. Preferrably do this on device.
- [x] As a final check, ensure all contexts show payment sheet when feature flag is on. Off otherwise.

### DIFF
--- a/Library/ViewModels/PledgePaymentMethodsViewModel.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModel.swift
@@ -291,7 +291,7 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
     let didTapToAddNewCard = self.didSelectRowAtIndexPathProperty.signal.skipNil()
       .filter { $0.section == PaymentMethodsTableViewSection.addNewCard.rawValue }
 
-    let paymentSheetOnPledgeContext = context.ignoreValues().map(paymentSheetEnabled)
+    let paymentSheetOnPledgeContext = context.map { _ in paymentSheetEnabled }
 
     self.goToAddCardScreen = Signal.combineLatest(
       project,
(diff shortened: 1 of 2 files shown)


scottkicks:
@msadoon nice! Do we a handle any other error states on our side besides "Card was declined"? Or is everything else pretty much handled server side or through that `STPCardValidator` class?

msadoon:
> @msadoon nice! Do we a handle any other error states on our side besides "card declined"? Or is everything else pretty much handled server side or through that `STPCardValidator` class?

So we don't use `STPCardValidator` anymore on this new Payment Sheet, which is all handled on Stripes' end. There was a view where we rolled our own card validation called `AddNewCardViewController`that used `STPPaymentCardTextField` and `STPCardValidator`, but that whole page is replaced with Stripes' `PaymentSheet.FlowController` which you'll see in `PledgePaymentMethodsViewController` and I think `PaymentMethodsViewController`. if that flow controller returns an error we display in using our existing snack bar like popup - `MessageBannerViewController`. If its a serverside error we'll return it through our signal `.errors` you see is mapped to that same message banner.

scottkicks:
Oh got it. Yea I see that flow now. Thanks 🙏🏼

scottkicks (approved):

PR #1752, opened 2022-10-26
[PAY-2040] Payment Sheet Displayed On Unrelated Context

msadoon:
# 📲 What and 🤔 Why

We want to prevent any janky behaviour with the payment sheet because it does take a moment to show and the user has the chance to back to the previous page in that time.

# 🛠 How

Check the view type before presenting but after creating the `PaymentSheet.FlowController`.

# 👀 See

Before 🐛

https://user-images.githubusercontent.com/4282741/198148557-31c03abe-2550-46ae-8cc9-252c731ac2ec.MP4
 
After 🦋

https://user-images.githubusercontent.com/4282741/198148680-6e6ff0e9-e65d-4375-8b33-41c1b8c7bee2.MP4

# ✅ Acceptance criteria

- [x] Neither `PledgeViewController` or `PaymentMethodsViewController` show the payment sheet.

# ⏰ TODO

- [x] Test both contexts > 10 times (tricky due to timing)


Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -172,11 +172,16 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
             .pledgeViewController(strongSelf, didErrorWith: error.localizedDescription)
         case let .success(paymentSheetFlowController):
-          strongSelf.paymentSheetFlowController = paymentSheetFlowController
-          strongSelf.paymentSheetFlowController?.presentPaymentOptions(from: strongSelf) { [weak self] in
-            guard let strongSelf = self else { return }
+          let topViewController = strongSelf.navigationController?.topViewController
+          let paymentSheetShownWithinPledgeContext = topViewController is PledgeViewController
 
-            strongSelf.confirmPaymentResult(with: data.clientSecret)
+          if paymentSheetShownWithinPledgeContext {
```
scottkicks, line 178: Tested and this is working as expected. I'm confused why this logic was being hit though.

I would think that `PaymentSheet.FlowController.create` wouldn't finish executing if the backer hits the Back button before the sheet is finished creating/loading. 
I've set some breakpoints in the `PaymentSheet.FlowController.create` flow as well as around your changes in this completion handler and can't seem to hit them when backing out prematurely.

scottkicks (approved):
Tested and this is working as expected. Nice catch!

The instances of these ViewControllers are actually deallocated when going back to the previous screens most of the time. It's only in those very rare cases where the sheet's create completion makes it past those `guard` statements. 

Took me a minute to actually time that right 😆  to see the bug!

msadoon:
> Tested and this is working as expected. Nice catch!
> 
> The instances of these ViewControllers are actually deallocated when going back to the previous screens most of the time. It's only in those very rare cases where the sheet's create completion makes it past those `guard` statements.
> 
> Took me a minute to actually time that right 😆 to see the bug! Just need to update this branch with `main`

Cool, thanks for the review, it is a timing issue, I didn't think it would matter given how rarely it happens but probably best to tighten up the logic a little. I updated the branch, there's an update branch with main button on the bottom right.

PR #1759, opened 2022-11-29
[PAY-2053] Add New Payment Button Padding FIx

msadoon:
# 📲 What

An issue rarely occurs where the padding of the "New Payment Method" is skewed and appears broken.

# 🤔 Why

Considering the traffic through this page, we should ensure UI is properly spaced and "pretty" for the user.

# 🛠 How

Added constraints to `addButton` and `activityIndicator` that would place them in the center and take up all the space of their container view (which is a stackview). Because only one of them are shown at a time, it should correct the skewing issue shown below in "before".

It's hard to replicate this issue on portrait iPhone, but it was reliably happening in landscape mode for a new iPhone (13 Pro Max). So with this fix its' no longer occurring on landscape iPhone 13 Pro Max, which I'm thinking would mean its' permanently fixed for portrait too.

# 👀 See

This issue should be fixed by this PR, but its' hard to replicate:

Before:

After:

# ✅ Acceptance criteria

- [x] Landscape on iPhone 13 Pro Max doesn't show skewed "New Payment Method" button


scottkicks (approved):
Nice! Tested on a few devices both in landscape and portrait. UI looks good and tests pass 👍 

I'll go ahead and merge this and notify the payments team since @msadoon is OOO.

PR #1772, opened 2023-01-18
[WEB-862] AppTrackingTransparency Authorization

scottkicks:
# 📲 What

Presents Apple's AppTransparencyFramework [dialog](https://developer.apple.com/app-store/user-privacy-and-data-use/#:~:text=Learn%20more-,Asking%20Permission%20to%20Track,-With) on app didFinishLaunching.

# 🤔 Why

With iOS 14.5+ you need to receive the user’s permission through the AppTrackingTransparency framework in order to track them or access their device’s advertising identifier. 

We need to meet this requirement in order to ethically track facebook ads related data. 

context around our [consent management initiative](https://kickstarter.atlassian.net/wiki/spaces/NEW/pages/2225766401/iOS+Consent+Management)
context around our [facebook conversions api initiative ](https://kickstarter.atlassian.net/wiki/spaces/NEW/pages/2231762945/iOS+Facebook+Conversion+API)that will be gates by this authorization status

# 🛠 How

- Add `NSUserTrackingUsageDescription` in the info.plist
- when application `didFinishLaunchingWithOptions` use the [AppTrackingTransparency Framework](https://developer.apple.com/documentation/apptrackingtransparency) to request authorization
- The framework takes care of saving the authorization status for us. I have the completion handler here returning an internal `ATTrackingAuthorizationStatus` enum to be used by whomever consumes the Signal.

Apple only allows us to request this dialog once per fresh install until someone decides to allow or deny permission.

# 👀 See

| After 🦋 |
| ---- |
|  Simulator Screen Recording - iPhone 13 Pro - 2023-01-18 at 10 27 06 |

# ⏰ TODO
- [ ] verify that the dialog still shows up when the app is opened from a deep-link

### DIFF
--- a/Kickstarter-iOS/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/AppDelegateViewModel.swift
@@ -798,7 +798,10 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
 
     self.requestATTrackingAuthorizationStatus = self.applicationLaunchOptionsProperty.signal
       .skipNil()
-      .map { _ in atTrackingAuthorizationStatus() }
+      .map { _ -> ATTrackingAuthorizationStatus in
+        guard featureConsentManagementDialogEnabled() else { return .notDetermined }
+        return atTrackingAuthorizationStatus()
+      }
   }
 
   public var inputs: AppDelegateViewModelInputs { return self }
@@ -997,7 +1000,7 @@ private func atTrackingAuthorizationStatus() -> ATTrackingAuthorizationStatus {
       }
     })
   }
-  
+
   return authorizationStatus
 }
 
--- a/Kickstarter-iOS/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/AppDelegateViewModel.swift
@@ -1,4 +1,5 @@
 import AppboyKit
+import AppTrackingTransparency
 import KsApi
 import Library
 import Prelude
@@ -16,6 +17,12 @@ public enum NotificationAuthorizationStatus {
   case denied
   case notDetermined
   case provisional
+}
+
+public enum ATTrackingAuthorizationStatus {
+  case authorized
+  case denied
+  case notDetermined
 }
 
 public protocol AppDelegateViewModelInputs {
@@ -179,6 +186,9 @@ public protocol AppDelegateViewModelOutputs {
 
   /// Emits when we should register the device push token in Segment Analytics.
   var registerPushTokenInSegment: Signal<Data, Never> { get }
+  
+  /// Emits when  application didFinishLaunchingWithOptions.
+  var requestATTrackingAuthorizationStatus: Signal<ATTrackingAuthorizationStatus, Never> { get }
 
   /// Emits when our config updates with the enabled state for Semgent Analytics.
   var segmentIsEnabled: Signal<Bool, Never> { get }
@@ -785,6 +795,10 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     self.brazeWillDisplayInAppMessageReturnProperty <~ self.brazeWillDisplayInAppMessageProperty.signal
       .skipNil()
       .map { _ in .displayInAppMessageNow }
+    
+    self.requestATTrackingAuthorizationStatus = self.applicationLaunchOptionsProperty.signal
+      .skipNil()
+      .map { _ in atTrackingAuthorizationStatus() }
   }
 
   public var inputs: AppDelegateViewModelInputs { return self }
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -18,6 +19,12 @@ public enum NotificationAuthorizationStatus {
   case provisional
 }
 
+public enum ATTrackingAuthorizationStatus {
```
msadoon, line 22: Can you make this it's own file and put it in the "Library/Tracking" folder?

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -965,6 +983,27 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
   public let updateConfigInEnvironment: Signal<Config, Never>
 }
 
+private func atTrackingAuthorizationStatus() -> ATTrackingAuthorizationStatus {
+  var authorizationStatus: ATTrackingAuthorizationStatus = .notDetermined
+
+  DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
scottkicks, line 989: The app has to be completely active before the dialog can be shown. Since AppDelegate doesn't have a `viewDidAppear` method, this seems to be the standard way to approach this on app launch.

Kickstarter-iOS/Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSUserTrackingUsageDescription</key>
+	<string>We use personal data to provide a good experience on Kickstarter, and to help connect you with projects you&apos;ll love.</string>
```
scottkicks, line 6: I'm not sure if descriptions in the plist are translated for us. Do you know?

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -965,6 +983,27 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
+    ATTrackingManager.requestTrackingAuthorization(completionHandler: { status in
+      switch status {
+      case .notDetermined:
+        authorizationStatus = .notDetermined
+      case .authorized:
+        authorizationStatus = .authorized
+      case .denied:
+        authorizationStatus = .denied
+      default:
```
msadoon, line 998: Pretty sure you don't need a `default` case if all the other cases are covered.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -965,6 +983,27 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
   public let updateConfigInEnvironment: Signal<Config, Never>
 }
 
+private func atTrackingAuthorizationStatus() -> ATTrackingAuthorizationStatus {
```
msadoon, line 986: Maybe update the name of this function to `attTracckingAuthorizationStatus`?

Kickstarter-iOS/AppDelegateViewModelTests.swift
```diff
@@ -3004,6 +3007,12 @@ final class AppDelegateViewModelTests: TestCase {
       self.updateCurrentUserInEnvironment.assertValues([user, updatedUser])
     }
   }
+
+  func testRequestATTrackingAuthorizationStatus_CalledOnceOnDidFinishLaunching() {
+    self.vm.inputs.applicationDidFinishLaunching(application: UIApplication.shared, launchOptions: nil)
+
+    self.requestATTrackingAuthorizationStatus.assertValueCount(1)
```
msadoon, line 3014: test the assert before the input to ensure it doesn't emit.

Kickstarter-iOS/Info.plist
msadoon, line 6: Cool! I guess this is where marketing will add their flavor text :)

msadoon (changes requested):
Hey mostly good, made some comments. 
We can discuss any questions in our meeting too!

Kickstarter-iOS/AppDelegateViewModel.swift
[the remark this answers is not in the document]
scottkicks: The app has to be completely active before the dialog can be shown. Since AppDelegate doesn't have a viewDidAppear method, this seems to be the standard way to approach this on app launch.

I'm just not used to the ReactiveSwift way of adding a delay. I'll update the SIgnal directly with your suggestion 👍

Kickstarter-iOS/Info.plist
msadoon: good question! Seems like the system itself would handle that, can you check? I know we have some other popups on first launch too..wonder if they have translations.

Kickstarter-iOS/AppDelegateViewModel.swift
scottkicks: yep - I replied and will address your other comment. I think we are commenting at the same time lol

Kickstarter-iOS/AppDelegateViewModel.swift
msadoon: it's just weirdly camel cased, but ok.

msadoon (approved):
Looks good! Just one thing I just realized `ksr_debounce` might work but not be completely accurate as it calculates the time between signals, which might work in this case. `ksr_delay` is another option that seems like a better fit here.

sclampet:
@msadoon good call out. thanks for the insight. I'll go with the more consistent approach of `ksr_delay`

PR #1773, opened 2023-01-19
Revert "attempt to fix Fastlane S3 access"

msadoon:
Reverts kickstarter/ios-oss#1771 to update branch name so it propagates to `ios-private`


msadoon:
merging this before tests pass, but only to add these changes in a seperate branch that triggers ci for `ios-private`

PR #1779, opened 2023-01-24
[PAY-1527] Local Reward Feature Flag Removed

msadoon:
# 📲 What

So Optimizely feature flag `ios_local_pickup` is no longer needed to be a check for displaying rewards in the app.

# 🤔 Why

The feature has been in production for over 6 months and therefore considered safe for permanent use.

# 🛠 How

Removed all Optimizely related code for storing and mapping the key from the the optimizely client.
Removed the logic for filtering the rewards based on this flag. So double the logic in `RewardsCollectionViewModel` but since we're not required to filter out any local rewards, we no longer need the feature flag or the backing reward.

Previously a reward was hidden only when ALL these conditions were true 1) feature flag OFF 2) reward is local 3) reward is not backed. So now that we don't need the flag check, the rest of the conditions (2, 3) should no longer gate any rewards.

# ✅ Acceptance criteria

- [x] - [ ] Run the app, check out a project with local rewards, see if they are displayed the same as web.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/RewardsCollectionViewModelTests.swift:163:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

PR #1780, opened 2023-01-25
[WEB-935] Creator Dashboard Analytics

scottkicks:
# 📲 What

Adding the following segment events in the creator dashboard screen.
- Page Viewed, creator dashboard screen
- CTA Clicked, when a creator switches between projects via the dropdown at the top of the screen.
- CTA Clicked when a creator taps “Post Update”

# 🤔 Why

The creator dashboard doesn't have any analytic events. We want to start collecting data around whether members are finding it valuable enough to keep. The current assumption is that they don't and that we can possibly remove this feature in an effort to focus more on the backing experience.

# 🛠 How

I've added 3 new SegmentEvents in `KSRAnalytics`, as well as tests in `KSRAnalyticsTests` and `DashboardViewModelTests`

1. `trackCreatorDasboardPageViewed`
2. `trackCreatorDasboardSwitchProjectClicked`
3. `trackCreatorDasboardPostUpdateClicked`

# 👀 See

No UI updates

# 🧪 TESTING

- [ ] Check in [Segment](https://app.segment.com/kickstarter/sources) that these events are being received at the appropriate times.

### DIFF
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -528,6 +528,18 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
   }
 
+  func testCreatorDasboardPostUpdateClicked() {
+    let segmentClient = MockTrackingClient()
+    let ksrAnalytics = KSRAnalytics(segmentClient: segmentClient)
+
+    ksrAnalytics.trackCreatorDasboardPostUpdateClicked(project: .template, refTag: RefTag.dashboard)
+
+    XCTAssertEqual(["CTA Clicked"], segmentClient.events)
+    XCTAssertEqual(["project_update_create_draft"], segmentClient.properties(forKey: "context_cta"))
+    XCTAssertEqual(["creator_dashboard"], segmentClient.properties(forKey: "context_page"))
+    XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
+  }
+
   // MARK: - Discovery Properties Tests
 
   func testDiscoveryProperties() {
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -499,6 +499,23 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual("Art", segmentClientProperties?["project_category"] as? String)
   }
 
+  // MARK: - Creator Dashboard Properties Tests
+
+  func testCreatorDashboardPageViewed() {
+    let segmentClient = MockTrackingClient()
+    let loggedInUser = User.template |> \.id .~ 23
+    let ksrAnalytics = KSRAnalytics(
+      loggedInUser: loggedInUser,
+      segmentClient: segmentClient
+    )
+
+    ksrAnalytics.trackCreatorDasboardPageViewed()
+
+    XCTAssertEqual(["Page Viewed"], segmentClient.events)
+    XCTAssertEqual(["creator_dashboard"], segmentClient.properties(forKey: "context_page"))
+    XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
+  }
+
   // MARK: - Discovery Properties Tests
 
   func testDiscoveryProperties() {
--- a/Library/Tracking/KSRAnalyticsTests.swift
+++ b/Library/Tracking/KSRAnalyticsTests.swift
@@ -516,6 +516,18 @@ final class KSRAnalyticsTests: TestCase {
     XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
   }
 
+  func testCreatorDasboardSwitchProjectClicked() {
+    let segmentClient = MockTrackingClient()
+    let ksrAnalytics = KSRAnalytics(segmentClient: segmentClient)
+
+    ksrAnalytics.trackCreatorDasboardSwitchProjectClicked(project: .template, refTag: RefTag.dashboard)
+
+    XCTAssertEqual(["CTA Clicked"], segmentClient.events)
+    XCTAssertEqual(["project_select"], segmentClient.properties(forKey: "context_cta"))
+    XCTAssertEqual(["creator_dashboard"], segmentClient.properties(forKey: "context_page"))
+    XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
+  }
+
   // MARK: - Discovery Properties Tests
 
   func testDiscoveryProperties() {
--- a/Library/ViewModels/DashboardViewModel.swift
+++ b/Library/ViewModels/DashboardViewModel.swift
@@ -46,6 +46,9 @@ public protocol DashboardViewModelInputs {
   /// Call when to show or hide the projects drawer.
   func showHideProjectsDrawer()
 
+  /// Call when Post Update is clicked
+  func trackPostUpdateClicked()
+
   /// Call when the view loads.
   func viewDidLoad()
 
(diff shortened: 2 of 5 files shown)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Tracking/KSRAnalytics.swift:782:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalytics.swift:804:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

scottkicks:
good to review. checks are just running after updating this branch with main

Library/Tracking/KSRAnalyticsTests.swift
```diff
@@ -499,6 +499,47 @@ final class KSRAnalyticsTests: TestCase {
...
+    )
+
+    ksrAnalytics.trackCreatorDasboardPageViewed()
+
+    XCTAssertEqual(["Page Viewed"], segmentClient.events)
+    XCTAssertEqual(["creator_dashboard"], segmentClient.properties(forKey: "context_page"))
+    XCTAssertEqual(["dashboard"], segmentClient.properties(forKey: "context_section"))
+  }
+
+  func testCreatorDasboardSwitchProjectClicked() {
```
msadoon, line 519: So yea just check spelling here and for the test below. If you could also test maybe a specific project property like `project_id` or `name` for the `.template` project that would validate that the right project is being track alongside the event info.

Library/ViewModels/DashboardViewModel.swift
```diff
@@ -290,6 +293,28 @@ public final class DashboardViewModel: DashboardViewModelInputs, DashboardViewMo
...
+
+    // MARK: - Tracking
+
+    self.viewWillAppearAnimatedProperty.signal.observeValues { _ in
+      AppEnvironment.current.ksrAnalytics.trackCreatorDasboardPageViewed()
+    }
+
+    _ = self.project
+      .takeWhen(self.switchToProjectProperty.signal)
+      .observeValues { project in
```
msadoon, line 305: We probably don't want to send the existing project with this tracking, but the project that is being switched _to_. try `takePairWhen`. Same thing for below signal for post update clicked.

msadoon: Also just be sure that these events actually get tracked in segment, because sometimes `_ = ` doesn't register an event at least in the past I've noticed sometimes' they can be flakey.

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -335,4 +335,68 @@ internal final class DashboardViewModelTests: TestCase {
       self.vm.inputs.showHideProjectsDrawer()
     }
   }
+
+  func testTrackingEvents_CreatorDashboardViewed() {
+    let projects = (0...4).map { .template |> Project.lens.id .~ $0 }
```
msadoon, line 340: ohh nice use of inclusive range with map ❇️

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -335,4 +335,68 @@ internal final class DashboardViewModelTests: TestCase {
...
+    let project1 = Project.template
+    let project2 = .template |> Project.lens.id .~ 4
+    let projects = [project1, project2]
+
+    withEnvironment(apiService: MockService(fetchProjectsResponse: projects)) {
+      XCTAssertEqual([], self.segmentTrackingClient.events)
+
+      self.vm.inputs.viewWillAppear(animated: false)
+
+      self.scheduler.advance()
```
msadoon, line 369: Do you need this `advance` noticed above test wasn't using it.

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -335,4 +335,68 @@ internal final class DashboardViewModelTests: TestCase {
...
+
+      XCTAssertEqual(["Page Viewed"], self.segmentTrackingClient.events)
+
+      self.vm.inputs.switch(toProject: .id(project2.id))
+
+      XCTAssertEqual(["Page Viewed", "CTA Clicked"], self.segmentTrackingClient.events)
+
+      self.vm.inputs.switch(toProject: .id(project2.id))
```
msadoon, line 377: So you're probably not gonna switch from the same project to itself. So I guess if you start with project 1 you should switch project 2. Also along with the change to track the project being switched _to_, you can include a check to say the right project is being tracked when switching back and forth.

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -335,4 +335,68 @@ internal final class DashboardViewModelTests: TestCase {
...
+    }
+  }
+
+  func testTrackingEvents_CreatorDashboardPostUpdateClicked() {
+    withEnvironment(apiService: MockService(fetchProjectsResponse: [Project.template])) {
+      XCTAssertEqual([], self.segmentTrackingClient.events)
+
+      self.vm.inputs.viewWillAppear(animated: false)
+
+      self.scheduler.advance()
```
msadoon, line 389: Again no need advance here I think

Library/ViewModels/DashboardViewModelTests.swift
```diff
@@ -335,4 +335,68 @@ internal final class DashboardViewModelTests: TestCase {
...
+
+      XCTAssertEqual(["Page Viewed"], self.segmentTrackingClient.events)
+
+      self.vm.inputs.trackPostUpdateClicked()
+
+      XCTAssertEqual(["Page Viewed", "CTA Clicked"], self.segmentTrackingClient.events)
+
+      self.vm.inputs.trackPostUpdateClicked()
```
msadoon, line 397: No need to test again if it's identical event.

msadoon (changes requested):
Some changes here, most of them minor.
Biggest thing is to ensure we track the project clicked, not the project we're currently on.
👍🏿 Exciting to see this tracking being implemented.

Library/ViewModels/DashboardViewModelTests.swift
scottkicks: yes actually. it needs time for the `fetchProjectsResponse` to complete and set the given projects I believe. these tests fail without it.

Library/ViewModels/DashboardViewModelTests.swift
scottkicks: this test also needs this. left my thoughts in the previous comment.

Library/ViewModels/DashboardViewModelTests.swift
msadoon: So instead of using `[.template |>]` to check the assert values, you can pass in `project1` and `project2` directly. Makes the code less cumbersome, and easier to follow. Should pass tests, if it doesn't we'll revisit.

Library/ViewModels/DashboardViewModel.swift
msadoon: hey just wanted to call out a change here when using `takePairWhen`, we use `_, param in` because we want the second parameter, not the first one like before. However, the catch here is that the second argument is not a project, so to find the `project` from the `param`, we see that in this view model all the projects were found through `fetchProjects` call and some filtering is done around line `155` to find the existing project from a param. Just copying and modifying slightly that logic applies to this tracking call too. Tested in Segment and it works as expected.

msadoon (approved):
Tested all the calls and they are working as expected.
Made slight modifications to the track switch signals. Encourage you to kind of study it for more exposure to reactive swift and we get the project from param (in this case). Ask any and all questions you have, I'm happy to assist :)

Good work!

Library/ViewModels/DashboardViewModel.swift
msadoon: Good question, I had to look at this carefully, I think you can't map directly on `projectsAndSelected`, because it goes to other signals, but `selectProjectPropertyOrFirst` is an intermediate signal that does tie in to `switchToProjectProperty`, which we use in our own `.map` function. It also includes the projects navigated to and from activities and message threads. If you want to track those other events as well, you can use `selectProjectPropertyOrFirst`, but you'd still need to re-write the map operation again, because the existing one lives inside the `projectsAndSelected`

PR #1791, opened 2023-02-10
[NTV-647] Braze In-App Notification Fix

msadoon:
# 📲 What

We noticed in-app notifications were broken (or not appearing) in release 5.6.1.
Needed to fix those. 

# 🤔 Why

Strangely two different instances of `SEGAppboyIntegrationFactory` were being created when calling `instance()` on that class. Somehow this caused in-app notifications not to show up.

# 🛠 How

Created one instance of `SEGAppboyIntegrationFactory` inside the `Analytics` `configuredClient` function and pass it back out to be configure in the `AppDelegate`. (Previously passing it into the function).

Also it seems like segment-appboy-ios is pinning Appboy_iOS_SDK to `4.5.1` now, so let's use it's dependency instead of doing the pinning on our own.

# ✅ Acceptance criteria

- [x] Test that in-app notifications show on staging.
- [x] Test push notifications deeplink on staging.
- [x] Test in-app notification deeplinks on staging.
 
# ⏰ TODO

- [ ] Validate that Apply Pay crash scenario doesn't occur without manually pinning Appboy_iOS_SDK.
- [x] Check to see if 5.6.0 had in-app notifications deeplink correctly and re-test with this code.

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -269,6 +269,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         appBoyInstance?.saveLaunchOptions(launchOptions)
         appBoyInstance?.appboyOptions = [
           ABKInAppMessageControllerDelegateKey: strongSelf,
+          ABKURLDelegateKey: strongSelf,
           ABKMinimumTriggerTimeIntervalKey: 5
         ]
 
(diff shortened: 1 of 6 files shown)


scottkicks (approved):
code LGTM. Odd that the appBoyInstance is changing.

I still can't test app notifications on my physical device so I'll defer to you to make sure it is still working.

msadoon:
> code LGTM. Odd that the appBoyInstance is changing.
> 
> I still can't test app notifications on my physical device so I'll defer to you to make sure it is working.

They do work I tested them yesterday.

PR #1798, opened 2023-03-02
[WEB-970] Add CAPI Events

scottkicks:
# 📲 What

Triggers our new [triggerCapiEvent](https://kickstarter.atlassian.net/browse/WEB-671) mutation in the following cases

#### Event Cases:
- Project page visit
- Initiate checkout (selects a reward)
- Add new payment method
- Purchase (have reached the thank you page)

#### Events should only trigger if the following is true:
1. The member has given their consent for KS to track them via [AppTrackingTransparencey.trackingAuthorizationRequest](https://developer.apple.com/documentation/apptrackingtransparency)
2. The `sendMetaCapiEvents` property on the project is true.

# 🤔 Why

This mutation helps us collect data on those members that have entered the app via a FB ad. Since iOS members can opt out of 3rd party app tracking, our FB pixel could be rendered useless. Facebook's CAPI (Conversions API) allows us to collect tracking data server to server, essentially getting around this issue. 

It's important to note that Facebook does some magic on their end to associate the member emails we send to them with ad clicks so it's important that we send this value back whenever possible. There is of course the case where someone has enters from an ad and doesn't not log in/create an account.

# 🛠 How
Created a new service: 
- `AppTrackingTransparency`: for getting authorizationStatus and advertisingIdentifier

Then I just made those calls in the correct places according to the cases listed above

**Note:** We've been using this [project](https://www.kickstarter.com/projects/ringobottle/ringo-the-water-bottle-that-magnetically-holds-your-phone?ref=discovery&term=ringo) for testing. It has its `sendMetaCapiEvents` property set to true.


Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -341,6 +341,42 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
       }
 
+    // FB CAPI
+    let graphUser = self.viewDidAppearAnimatedProperty.signal.ignoreValues()
+      .switchMap { _ in
+        AppEnvironment.current.apiService
+          .fetchGraphUser(withStoredCards: false)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .materialize()
+      }
```
scottkicks, line 351: fetching the graph user this way in a couple of places. I was thinking it could be cleaner to put this in `FacebookCAPIEventService` as well, but it doesn't look like we typically subscribe to Signals from anywhere other than View Models. What do you think?

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -199,6 +199,32 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
+        AppEnvironment.current.apiService
+          .fetchGraphUser(withStoredCards: false)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .materialize()
+      }
+
+    _ = Signal.combineLatest(project, self.viewDidLoadProperty.signal.ignoreValues())
+      .combineLatest(with: graphUser)
+      .observeValues { projectSignal, graphUser in
+        let (project, _) = projectSignal
```
msadoon, line 215: Okay so getting the user data is a bit tricky. But we have conventions kind of in place to pass existing data back up the ladder of views in place of making a completely new network call that gets ALL the user data when really all we need is one field "email". 

So the way I would do this is start with `PledgePaymentMethodsViewModel`, because it already does the `fetchGraphUser` and has outputs with the names `notify` prefix to them. Create a new output `notifyFacebookCAPIUserData` and send the email from the existing `fetchGraphUser` call. The signal can be the same as the one you use to get the email for tracking the payment sheet appearance.

next step is the output in `PledgePaymentMethodsViewController`, which can call a delegate kind of like  
```
extension PledgeViewController: PledgePaymentMethodsViewControllerDelegate {
  func pledgePaymentMethodsViewController(
    _: PledgePaymentMethodsViewController,
    didSelectCreditCard paymentSource: PaymentSourceSelected
  ) {
    self.viewModel.inputs.creditCardSelected(with: paymentSource)
  }
```
that stores the user data (just the email) within the `PledgeViewModel` as maybe a mutable property?. Now we see that `ThanksPageData` is created in the `PledgeViewModel` before triggering the thanks page. So pass along the stored value alongside existing checkout data.

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -199,6 +199,32 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
+
+        guard featureFacebookConversionsAPIEnabled(), project.sendMetaCapiEvents == true else { return }
+
+        let userEmail = graphUser.value?.me.email
+
+        FacebookCAPIEventService
```
msadoon, line 221: Same comments above apply for this call.

Library/ViewModels/ProjectPageViewModel.swift
msadoon: @scottkicks great question! Answered this in the other spots. Let me know if anything doesn't quite makes sense, happy to jump on a call too.

msadoon: Actually because user email is optional for `TriggerCapiEventInput`, let's not pass it here. It's complicated with `User` as our `AppEnvironment.currentUser` shared user (the one updated when they log in for the global app to see) and our `GraphUser` objects which are called in context. `User` is v1 and `GraphUser` is GQL. We need to consolidate those two models, which is a lot of work.

msadoon (changes requested):
Ok, good first shot at this.
There's a lot of comments of here, but mostly just following existing conventions and reducing the network requests required.

Let's pair on any of this if you have questions.

Library/FacebookCAPIEventService.swift
```diff
@@ -0,0 +1,53 @@
...
+
+public enum FacebookCAPIEventName: String {
+  /// viewing a project page
+  case ViewContent
+  /// viewing the reward selection screen
+  case InitiateCheckout
+  /// viewing the checkout page where you enter credit card details
+  case AddPaymentInfo
+  /// completion of backing
+  case Purchase
```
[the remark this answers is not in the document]
scottkicks: Yes I actually pointed this out to the web team, but got push back on that. I can definitely name these better for our end of though. i.e. `case ProjectPageViewed = "ViewContent"...`

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -378,6 +378,33 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+          .fetchGraphUser(withStoredCards: false)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .materialize()
+      }
+
+    _ = Signal.combineLatest(project, self.viewDidLoadProperty.signal.ignoreValues())
+      .takeWhen(didTapToAddNewCard)
+      .combineLatest(with: graphUser)
+      .observeValues { projectSignal, graphUser in
+        let (project, _) = projectSignal
```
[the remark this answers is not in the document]
scottkicks: can you clarify what you're suggesting to remove here?

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -487,11 +490,6 @@
...
 		606F216B299E8D8500BA5CDF /* TriggerCapiEventInputTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 606F2169299E8C9F00BA5CDF /* TriggerCapiEventInputTests.swift */; };
-		606F214C2994158500BA5CDF /* TriggerCapiEventInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 606F214B2994158500BA5CDF /* TriggerCapiEventInput.swift */; };
-		606F215C299D3FA900BA5CDF /* GraphAPI.TriggerCapiEventInput+TriggerCapiEventInputTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 606F215A299D3EF800BA5CDF /* GraphAPI.TriggerCapiEventInput+TriggerCapiEventInputTests.swift */; };
-		606F215E299D414800BA5CDF /* GraphAPI.TriggerCapiEventInput+TriggerCapiEventInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 606F215D299D414800BA5CDF /* GraphAPI.TriggerCapiEventInput+TriggerCapiEventInput.swift */; };
-		606F2166299D456D00BA5CDF /* TriggerCapiEvent.graphql in Sources */ = {isa = PBXBuildFile; fileRef = 606F2165299D456D00BA5CDF /* TriggerCapiEvent.graphql */; };
-		606F2168299D45F900BA5CDF /* TriggerCapiEventInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 606F2167299D45F900BA5CDF /* TriggerCapiEventInput.swift */; };
```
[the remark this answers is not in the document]
scottkicks: This is interesting, because all of these files exist locally for me, and the ci tests would have failed if they weren't actually there..

Library/ViewModels/ProjectPageViewModel.swift
scottkicks: oh wait seeing your other comment now....

Library/ViewModels/ThanksViewModel.swift
scottkicks: Ok I was think I was able to follow this pattern. IMO though this pattern is pretty confusing to follow, especially as someone still pretty new to ReactiveSwift. It takes more effort just to traced down where that signal is coming from and I'd prefer to have a single source of truth for pulling this information. However, I see how making a new network call for every time we want a user's email isn't great.

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -199,6 +199,39 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
     ) }
+
+    // FB CAPI
+    let facebookCAPIUserEmail = self.configureWithDataProperty.signal
+      .skipNil()
+      .map { $0.checkoutData?.facebookCAPIUserEmail }
+
+    let checkoutTotal = self.configureWithDataProperty.signal
+      .skipNil()
+      .map { ($0.checkoutData?.revenueInUsd) }
```
msadoon, line 210: So you can combine these into a tuple and have one signal instead of two, I would suggest a call like this:

```
    _ = Signal.combineLatest(project,
                         self.configureWithDataProperty.signal.skipNil())
      .observeValues { project, configData in
        var userEmail: String?
        var checkoutUSDAmount = ""
        
        if let checkoutDataValues = configData.checkoutData {
          userEmail = checkoutDataValues.facebookCAPIUserEmail
          checkoutUSDAmount = "\(checkoutDataValues.revenueInUsd)" + "\(checkoutDataValues.addOnsMinimumUsd)"
          checkoutUSDAmount += String(describing: checkoutDataValues.shippingAmountUsd)
          checkoutUSDAmount += String(describing: checkoutDataValues.bonusAmountInUsd)
        }

        guard featureFacebookConversionsAPIEnabled(), project.sendMetaCapiEvents,
          let externalId = AppTrackingTransparency.advertisingIdentifier() else { return }

        _ = AppEnvironment
          .current
          .apiService
          .triggerCapiEventInput(
            input: .init(
              projectId: "\(project.id)",
              eventName: FacebookCAPIEventName.BackingComplete.rawValue,
              externalId: externalId,
              userEmail: userEmail,
              appData: .init(extinfo: ["i2"]),
              customData: .init(currency: Currency.USD.rawValue, value: checkoutUSDAmount)
            )
          )
      }
```

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -199,6 +199,39 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
+          .apiService
+          .triggerCapiEventInput(
+            input: .init(
+              projectId: "\(project.id)",
+              eventName: FacebookCAPIEventName.BackingComplete.rawValue,
+              externalId: externalId,
+              userEmail: facebookCAPIUserEmail ?? "",
+              appData: GraphAPI.AppDataInput(extinfo: ["i2"]),
+              customData: GraphAPI
+                .CustomDataInput(currency: project.stats.currency, value: "\(checkoutTotal)")
```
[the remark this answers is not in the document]
msadoon: I wrote this call above, it cleans things up a bit.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -378,6 +379,39 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+    let userEmail = storedCardsEvent.values()
+      .filter(second >>> isFalse)
+      .map(first)
+      .skipNil()
+      .map { $0.me.email }
+
+    self.notifyFacebookCAPIUserEmail = userEmail.signal
+
+    _ = Signal.combineLatest(project, self.viewDidLoadProperty.signal.ignoreValues())
+      .takeWhen(didTapToAddNewCard)
```
msadoon, line 393: So I know I suggested the `goToAddCardViaStripeScreen` as the signal to use here. But rethinking this, if we follow that signal out to the view and see `goToPaymentSheet` as the signal the _eventually_ shows the Stripe payment sheet. The network request might fail here `private func goToPaymentSheet`. So use a new signal input from the `.success` case of that closure (just like `shouldCancelPaymentSheetAppearance` in the failure case). And upon receipt of the signal do this: 
```
    _ = Signal.combineLatest(project, self.viewDidLoadProperty.signal.ignoreValues())
      .takeWhen(stripePaymentSheetDidAppear)
```

That way you guarantee you send the tracking event only if the user actually sees the payment sheet. (For example, with no network, it could fail to appear)

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -47,6 +47,7 @@ public protocol PledgePaymentMethodsViewModelInputs {
 public protocol PledgePaymentMethodsViewModelOutputs {
   var goToAddCardScreen: Signal<(AddNewCardIntent, Project), Never> { get }
   var goToAddCardViaStripeScreen: Signal<PaymentSheetSetupData, Never> { get }
+  var notifyFacebookCAPIUserEmail: Signal<String?, Never> { get }
```
msadoon, line 50: can test this.

msadoon (changes requested):
We're getting there!

A few more suggestions and its' good to go!

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -341,6 +341,66 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
```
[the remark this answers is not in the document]
scottkicks: @msadoon  The reason this property is optional is because a non logged in user might view a project. However, if a logged in user does, we need to send the email. I'm not sure how I'd get the current user's email here without making the network request

Library/ViewModels/ThanksViewModel.swift
scottkicks: got it. thanks for the example!

Library/ViewModels/PledgePaymentMethodsViewModel.swift
scottkicks: definitely makes sense. thanks for the helpful context here. not super familiar with the payments flow just yet

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -139,6 +140,14 @@ final class PledgePaymentMethodsViewController: UIViewController {
 
         strongSelf.updateAddNewPaymentMethodButtonLoading(state: showLoadingIndicator)
       }
+
+    self.viewModel.outputs.notifyFacebookCAPIUserEmail
+      .observeForUI()
```
msadoon, line 145: Oops missed this one, no need for observe for UI, can go straight to `observeValues`

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -666,6 +668,31 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
     }
   }
 
+  func testNotifyFacebookCAPIUserEmail_LoggedIn() {
```
msadoon, line 671: Good tests!

Library/ViewModels/ProjectPageViewModel.swift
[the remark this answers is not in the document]
msadoon: @scottkicks okay so as discussed yesterday we'll make this a separate ticket and an email we receive from a small one field query in `AppDelegate` should be accessible app-wide. `AppEnvironment.currency.loggedInUserEmail` For the purposes of this PR though let's remove `graphUser` signal and pass in `nil` as the `userEmail`

Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -228,6 +228,40 @@ public final class RewardsCollectionViewModel: RewardsCollectionViewModelType,
           refTag: refTag
         )
       }
+
+    // FB CAPI
+    let graphUser = self.viewDidAppearProperty.signal.ignoreValues()
```
msadoon, line 233: Similar to above, let's remove this heavy call here and add it back in later with `AppEnvironment.current.loggedInUserEmail`.

msadoon (changes requested):
Ok very close, just a few more comments and next pass should be mergeable.

msadoon (approved):
Let tests pass and good to go!

PR #1799, opened 2023-03-06
[WEB-863] Guard Segment Events Behind AppTrackingTransparency

scottkicks:
# 📲 What

We should only send segment analytic events if the member has given consent via the AppTrackingTransparency dialog

# 🤔 Why

If the member doesn't want KS to track their data, then we shouldn't.

# 🛠 How

- Using AppTrackingTransparency library, we can access the current authorization status and use that to gate our `KSRAnalytics.track()` method. 
- This should also be behind the Consent Dialog feature flag

# ✅ Acceptance criteria

- [x] If a member's consent permission status has been denied, or has yet to be determined, then no Segment calls should be made
- [x] If permission has been authorized (granted) Segment calls should continue being triggered as usual


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:134:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/AppTrackingTransparency.swift:12:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Environment.swift:23:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/MockAppTrackingTrasparency.swift:6:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/TestHelpers/XCTestCase+AppEnvironment.swift:22:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalytics.swift:1243:1 | warning: (trailingSpace) Remove trailing space at end of a line. |
| Library/Tracking/KSRAnalytics.swift:1243:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/Tracking/KSRAnalytics.swift:1244:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalytics.swift:1245:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalyticsTests.swift:1612:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/Tracking/KSRAnalyticsTests.swift:1624:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/Tracking/KSRAnalyticsTests.swift:1624:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1625:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1645:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/Tracking/KSRAnalyticsTests.swift:1645:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1646:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1669:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/Tracking/KSRAnalyticsTests.swift:1669:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1670:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1693:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/Tracking/KSRAnalyticsTests.swift:1693:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |
| Library/Tracking/KSRAnalyticsTests.swift:1694:1 | warning: (wrapArguments) Align wrapped function arguments or collection elements. |


</p>

scottkicks:
based off of the [Add CAPI events PR](https://github.com/kickstarter/ios-oss/pull/1798). 
I'll merge this one in once that one is done.

Library/AppEnvironment.swift
```diff
@@ -162,6 +164,7 @@ public struct AppEnvironment: AppEnvironmentType {
         apiDelayInterval: apiDelayInterval,
         applePayCapabilities: applePayCapabilities,
         application: application,
+        appTrackingTransparency: appTrackingTransparency,
```
msadoon, line 167: Nice one, adding this to environment.

Library/MockAppTrackingTrasparency.swift
```diff
@@ -0,0 +1,14 @@
+import Library
+
+struct MockAppTrackingTransparencyService: AppTrackingTransparencyType {
```
msadoon, line 3: Nice, good work adding this.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1239,6 +1239,11 @@ public final class KSRAnalytics {
     properties: [String: Any] = [:],
     refTag: String? = nil
   ) {
+    if featureConsentManagementDialogEnabled() {
+      guard AppEnvironment.current.appTrackingTransparency.authorizationStatus() == .authorized
+      else { return }
+    }
+
```
msadoon, line 1246: can be cleaner with 
```
guard featureConsentManagementDialogEnabled(), AppEnvironment.current.appTrackingTransparency.authorizationStatus() == .authorized else { return }
```

msadoon: So given we're requiring IDFA here, would KS want to get IDFA on every call? If that's the case then we can add it to `userProperties` which is sent up on every call.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -398,7 +398,7 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
         let (project, _) = projectSignal
 
         guard featureFacebookConversionsAPIEnabled(), project.sendMetaCapiEvents == true,
-          let externalId = AppTrackingTransparency.advertisingIdentifier() else { return }
+          let externalId = AppTrackingTransparency().advertisingIdentifier() else { return }
```
msadoon, line 401: Since we have it in the app environment, we can use `AppEnvironment.current.appTrackingTransparency` instead of creating a new instance of the struct. The only place we'd need to create that struct is after user is prompted and the dialog returns their answer.

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -348,7 +348,7 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
         let (project, _) = projectAndRefTag
 
         guard featureFacebookConversionsAPIEnabled(), project.sendMetaCapiEvents,
-          let externalId = AppTrackingTransparency.advertisingIdentifier() else { return }
+          let externalId = AppTrackingTransparency().advertisingIdentifier() else { return }
```
msadoon, line 351: Same as above.

msadoon (changes requested):
Almost there, a few things here.
1. Should we send the IDFA with every call now that's required for every call?
2. Save the advertisingIdentifier to AppEnvironment so we can use AppEnvironment, instead of recreating the struct everywhere (defeats the purpose of adding it to AppEnvironment)

Library/Tracking/KSRAnalytics.swift
scottkicks: I'm not sure I understand. We aren't requiring IDFA here

Library/Tracking/AppTrackingTransparency.swift
```diff
@@ -2,8 +2,15 @@ import AdSupport
...
 
-public struct AppTrackingTransparency {
-  public static func authorizationStatus() -> ATTrackingAuthorizationStatus {
+public protocol AppTrackingTransparencyType {
+  func authorizationStatus() -> ATTrackingAuthorizationStatus
+  func advertisingIdentifier() -> String?
+}
+
+public struct AppTrackingTransparency: AppTrackingTransparencyType {
+  public init() {}
```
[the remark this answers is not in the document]
scottkicks: In order to instantiate it as the default param in Environment's initializer, I need this init() to be publicly accessible

Library/Tracking/KSRAnalytics.swift
scottkicks: Do you think this as simple as adding 
`props["advertising_identifer"] = AppEnvironment.current.appTrackingTransparency.advertisingIdentifier`
to the userProperties method in KSRAnalytics? Or would it require updating the user model with a new property and potentially add too much to this current PR?

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Tracking/AppTrackingTransparency.swift:12:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

msadoon (approved):
Ok all good, great work!
One minor thing - if you could quickly add `appTrackingTransparency` to `replaceEnvironment` static func on `AppEnvironment` that would make it consistent there as well.

msadoon:
Oh yea and get tests to pass ;)

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -477,6 +477,7 @@
 		59D1E6581D1866F800896A4C /* DashboardVideoCellViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59D1E6571D1866F800896A4C /* DashboardVideoCellViewModel.swift */; };
 		59E877381DC9419700BCD1F7 /* Newsletter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 59E877371DC9419700BCD1F7 /* Newsletter.swift */; };
 		6008633829BA51CD00B87B39 /* MockAppTrackingTrasparency.swift in Sources */ = {isa = PBXBuildFile; fileRef = 609C7DF029B67DC4008D66F2 /* MockAppTrackingTrasparency.swift */; };
+		6008633929BA61CE00B87B39 /* MockAppTrackingTrasparency.swift in Sources */ = {isa = PBXBuildFile; fileRef = 609C7DF029B67DC4008D66F2 /* MockAppTrackingTrasparency.swift */; };
```
msadoon, line 480: Please remove this reference, I think its' the old file misspelled fileXthat accidently got included.

Library/TestHelpers/TestCase.swift
```diff
@@ -33,6 +33,9 @@ internal class TestCase: XCTestCase {
     var calendar = Calendar(identifier: .gregorian)
     calendar.timeZone = TimeZone(identifier: "GMT")!
 
+    self.optimizelyClient = MockOptimizelyClient()
+      |> \.features .~ [OptimizelyFeature.consentManagementDialogEnabled.rawValue: true]
+
```
msadoon, line 38: Can you remove this from the base `TestCase` and put it back into the `override setUp` for the individual test class? I was thinking about this [method](https://developer.apple.com/documentation/xctest/xctestcase/set_up_and_tear_down_state_in_your_tests#3909114).

msadoon: Create a class level variable `MockOptimizelyClient` in the `override class setup` func in `PledgeViewModelTests` for example, and then update it to `false` for whatever flags you need to be setup, then update the individual cases that need to be true as you've already done.

msadoon (changes requested):
Hey just asking for a few changes on the base `TestCase`

Library/TestHelpers/TestCase.swift
scottkicks: right. I was trying to avoid that because there are a few test files and it was easier to add it here. And then remove it once the feature is fully released. Removing it from one place would also be easier. What do you think about that?

scottkicks: Its a good point. I'll go ahead and add a class level variable to all of the test files that are failing

scottkicks: Its a good point. We are making an effort to audit and clean up our flags as we think about moving to firebase as a manager though. There are a few different test files that have multiple tests that are failing which feels like a lot of code changes for this. 

I can make a ticket to remove as well so I don't forget, but I can definitely go through and add class level variables too so all of our testcases aren't impacted

msadoon: If it's not too much work, let's use the class level variables. I now we're trying to figure out Optimizely replacements and then yes we'll go ahead and rip out Optimizely everywhere for the replacement but that work isn't quite finalized _yet_.

I think the original creators intent of `TestCase` is not to add customizations of `AppEnvironment`.
We'll see that `withEnvironment(` in `XCTestCase+AppEnvironment` is intended to be customized for every test case.
Likely when we add the replacement for `Optimizely`, we'll follow this same approach and add it to the `withEnvironment(` function and customize it in every test case that needs it.

It shouldn't be too much boilerplate right? Because we're updating the _class_ level `mockOptimizelyClient` not in every test. The only tests that need to change are the the ones that require a `true` flag. I understand its probably four different files, but it still feels like the tradeoff is worth it and sticks to existing conventions. Also when we have this new replacement for Optimizely, then those files are easy to change with the new dependency because we just go through AppEnvironment.

My feel for the app is that it's very consistent in its' patterns and conventions, which ends up being super super nice later when you go to refactor something like `AppEnvironment` out, and then you don't need to worry that `TestCase` specifically needs to also get an update for one dependency....because `AppEnvironment` contains all the dependencies.

msadoon: Ok, looking into this now, it's fine, just do it the way you have in `TestCase`, I'll approve this. The class level update doesn't really seem to work either.

msadoon (approved):
Good to go!

PR #1801, opened 2023-03-08
[WEB-999] Prelaunch Dates Blockers

msadoon:
# 📲 What

As outlined in this [ticket](https://kickstarter.atlassian.net/browse/WEB-999) we needed to update our existing project model `deadline` and `launchedAt` properties to optional instead of non-optional.

Reason is because the prelaunch pages do not have those fields populated in GQL and the existing project page works off of GQL not v1.

Obviously this isn't the final version of the prelaunch, but necessary to get the latest data for the eventual page UI.

# 🤔 Why

As stated above, we need the `deadline` and `launchedAt` dates to be non-optional for those fields to be populated in order for the guard check to pass in `Project+ProjectFragment` `dates`

# 🛠 How

Did a project search for both properties and found instances of their use. In cases where optionals where being returned already, just needed to add the property to the `if let` or `guard let`. Otherwise returned a sensible value (ie. for String was "", or Int was 1).

There are lot of places where these fields occur and going through them all with breakpoints helped validate that nothing was broken or out of place.

Mostly though relied on optional checks to safely unwrap properties that were once non-optional.

Also fixed up tests to pass, not too much to do there.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| non-optional_v1 | optional_gql |

# ✅ Acceptance criteria

- [x] Run through the app with breakpoints for whereever `deadline` or `launchedAt` or computed properties of those two were being used, hit breakpoint, ensure optional check passes.

### DIFF
--- a/Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewControllerTests.swift
+++ b/Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewControllerTests.swift
@@ -130,5 +130,5 @@ private let fundingStats = stats.enumerated().map { idx, pledged in
   .template
     |> ProjectStatsEnvelope.FundingDateStats.lens.cumulativePledged .~ pledged
     |> ProjectStatsEnvelope.FundingDateStats.lens.date
-    .~ (cosmicSurgery.dates.launchedAt + TimeInterval(idx * 86_400))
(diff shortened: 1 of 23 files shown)


scottkicks (approved):
ran through breakpoints and all optional checks are passing

PR #1805, opened 2023-03-27
[MBL-670] Saved Prelaunch Projects on Backer Dashboard -- Merge After MBL-671

msadoon:
# 📲 What

Trying to get the saved/unsaved prelaunch projects to update the backer dashboard's `Saved` count and project list.

# 🤔 Why

There was a bug on v1 when fetching the user to update the saved projects, a prelaunch project is not counted. GQL fetch user calls doesn't have this bug, so we swapped the call in this instance to GQL. 

# 🛠 How

We didn't create a new adapter or anything we just used the existing `User.user(from: UserFragment) -> User?` static function to send back an `UserEnvelope<User>` instead of the `UserEnvelope<GraphUser> from a new call.

It's kind of surface level change because its' reusing the old adapter and query to map to a `User` instead of a `GraphUser`.

# 👀 See

Before 🐛 

https://user-images.githubusercontent.com/4282741/227980869-46de08ac-6b37-4538-a8db-ffc1297a637c.mp4

After 🦋

https://user-images.githubusercontent.com/4282741/227979844-94ab671f-f8ee-4d02-9e52-9f55a55a9edf.mp4

# ✅ Acceptance criteria

- [x] Ensure projects and prelaunch projects are saved accurately and reflected in the backer dashboard.


msadoon:
@scottkicks Again just making you aware this is being merged without review for release timeline purposes.

PR #1806, opened 2023-03-27
Release 5.7.0 to Main

msadoon:
This is the release from 5.7.0 to Main.
The reason we are doing this so early and not the first week of April is because Apple will not let us submit apps that are not built with Xcode 14 after April. [Source](https://developer.apple.com/ios/submit/).


msadoon:
Again just trying to get a release submitted before April due to Apple's guidelines. (No review merge)

PR #1807, opened 2023-04-03
[Unticketed] UI on the Creator Dashboard Breaking

msadoon:
# 📲 What

On the latest build noticed a weird UI break with inconsistent number of tabs in the dashboard leading to a black background for creators and collaborators.

# 🤔 Why

We are going to remove the creator dashboard in the future (for something like 3000 users), but for now we support that side of things so we shouldn't introduce regressions there.

# 🛠 How

There was no `isCollaborator` field in GQL (unlike `isCreator`) UserFragment we already have, so had to create a new query, fragment, GraphUser adapter and test them all.

The new call is chained within `BackerDashboardViewModel` you'll see even if it fails the GraphUser data is propogated without `GraphUserMemberStatus`

# 👀 See

Before 🐛 

https://user-images.githubusercontent.com/4282741/229403649-e439c69c-cc66-426c-b457-6fd67c68e5f7.mp4

After 🦋 

https://user-images.githubusercontent.com/4282741/229403060-003a85d3-425b-4cbc-80c8-eac84808319e.mp4

# ✅ Acceptance criteria

- [x] Ensure logging in as a collaborator (user invited by a creator as a collaborator on the created project) still has saving/unsaving prelaunch and other projects working. (realnativesquad@proton.me in 1Pass)
- [x] Ensure logging in as a creator still has saving/unsaving prelaunch and other projects working. (therealnativesquad@gmail.com in 1Pass)
- [x] Ensure regular backer can save/unsave prelaunch or other projects working. (your login)

Note: There are minor hiccups in the UI, _sometimes_ the saved projects doesn't always show the updated "saved" project in the list. Try a few times and it works. Seems to be an issue only on creator/collaborator users.

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -247,6 +247,7 @@
 		1937A71F28C94FFC00DD732D /* SettingsFormFieldView.xib in Resources */ = {isa = PBXBuildFile; fileRef = 7790DF882200D3BD005DBB11 /* SettingsFormFieldView.xib */; };
 		1937A72328C9570A00DD732D /* ErroredBackingView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1937A72228C9570900DD732D /* ErroredBackingView.swift */; };
 		1937A72628C959DD00DD732D /* FundingGraphViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A7ED205B1E83240D00BFFA01 /* FundingGraphViewTests.swift */; };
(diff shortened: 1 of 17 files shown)


scottkicks (approved):
tested and working as expected.

what introduced this issue? I'm surprised I haven't seen it until now

msadoon:
> tested and working as expected.
> 
> what introduced this issue? I'm surprised I haven't seen it until now

It's a mismatch between the v1/fetchUser call and the fetchGraphUser call. We have this notification sent out `ksr_updatedUser` whenever we do the v1 call and that notification is observed in several places, once of which is `RootViewModel`. That's the vm that sets up the tabs for creators/backers. So the GQL call in `DashboardViewModel` gets the gql user but not a property that sets the tabs, `memberProjects`, which is why I added the new fragment.

PR #1811, opened 2023-04-09
[Unticketed] Rollback Dashboard Changes for Prelaunch Pages

msadoon:
# 📲 What

Our QA party revealed to us that Dashboard tab was being shown on mobile when it previously wasn't there for users that were "draft" creators. [Context](https://kickstarter.atlassian.net/browse/MBL-702).

These PR's are being reversed: 
https://github.com/kickstarter/ios-oss/pull/1805
https://github.com/kickstarter/ios-oss/pull/1807

This is due to a difference in the way v1 and GQL handle the idea of a "creator" when a project is not live. More detail below.

# 🤔 Why

This is an issue because a lot of users still have the creator side on the web and we shouldn't introduce a broken experience on mobile (ie. dashboard tab when previously there was none). Wrong approach anyway because we just need to handle the saved tab project list and count _not_ update the user.

# 🛠 How

https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/2294120449/Prelaunch+Pages+on+iOS#Resources

Goes into more detail about the differences between v1 and GQL properties ie. how a member of a project is defined.
One of the biggest issues is that GQL "draft" project members are considered "true" if they are invited before the project is launched and "false" in v1.

This is also the wrong approach because we don't need to update the `User`. All we need to do is update the list of "Saved Projects". This isolates any new GQL query to the tab itself. That will be addressed in the ticket above.

# ❌ Breaking Change

This will introduce one regression instead of the UI break. That is - dashboards saved tab will NOT properly reflect the saved states of prelaunch projects for _any_ type of user (creator with published projects, creator with draft projects, backer (no created projects) or collaborator (accepted invited member to a project)).

All users should still be able to save launched projects and see their dashboards updated accurately.

Hoping we can launch this build with this incomplete functionality and patch quickly after just so we can hit our Optimizely removal deadline in a few days.

# ✅ Acceptance criteria

- [x] Ensure saving projects are working as before. Saved tabs updating count and list for saving from profile tab, saving from search, saving from explore tab and saving within the project page.
- [x] If project is prelaunch - user can save in all cases above, but the saved tab will not be updated accurately.

# ⏰ Developer QA
- [x] Log in as a collaborator of a project (nativeappssquad@proton.me), ensure dashboard tab is not showing. **Note: Saving/unsaving prelaunch projects will not update the profile saved tab correctly.**
- [x] Log in as a creator of a project (therealnativesquad@gmail.com), ensure dashboard tab is showing and correctly working. **Note: Saving/unsaving prelaunch projects will not update the profile saved tab correctly.**
- [x] Log in as a drafter of a project (just create a draft project in our normal backer account), ensure dashboard tab is not showing. **Note: Saving/unsaving prelaunch projects will not update the profile saved tab correctly.**
- [x] Log in as a non-creator/regular backer (delete created draft in your normal backer account), ensure dashboard tab is not showing. **Note: Saving/unsaving prelaunch projects will not update the profile saved tab correctly.**


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

scottkicks (approved):
Working as expected according to the AC

PR #1814, opened 2023-04-19
[MBL-702] Update Project Watch Count and List on Dashboard for Prelaunch Projects

msadoon:
# 📲 What

Previous build 5.7.0 didn't account for the number of saved prelaunch projects in the "Saved" tab of the dashboard.

# 🤔 Why

Would be nice for the saved tab to accurately present the number of and list of saved projects accurately regardless of type.

# 🛠 How

This boiled down to getting the GQL users' saved projects count instead of relying on v1 numbers.

Full write up [here](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/2294120449/Prelaunch+Pages+on+iOS#QA-Bug---Dashboard-Saved-Tab-Not-Updating-Count-or-List.). 

`BackerDashboardViewModel`
To compensate for the new GQL call - introduced a 1 second debounce to the `projectSavedProperty` signal. This makes sense because `WatchProjectViewModel` emits a notification for the project everytime the save button is tapped _and_ everytime the network request returns, which happens right after. 

So we've always been making back-to-back `fetchGraphUser`  and `fetchErroredUserBackings` calls for the globally updated user (ie. 2 of each = 4 total on every save/unsave). 

Debounce introduces a time window where the last emission is the one forwarded in that second. 

So if the button is tapped and the network request returns in that same second - the network request is the signal received. 

If the network request returns after one second and there are no more button taps - it is the one sent in that _second_ second.

We go from 4 calls total to 2 calls total on each save/unsave and we add 1 call to each save/unsave to get the updated saved user project numbers so we're still reducing our total network request by 1.

That's a gain because network requests can be expensive.

# 👀  See

**Before**

https://user-images.githubusercontent.com/4282741/233404090-dcadf85d-3fab-44fd-9c0d-188649b43abc.mp4

**After**

https://user-images.githubusercontent.com/4282741/233404142-86b1229b-0517-4195-82d3-b77441ef525b.mp4

# ✅ Acceptance criteria / Dev QA

- [ ] Ensure all saved projects (prelaunch or otherwise) appear on dashboard "Saved" tab
- [x] If you have time, run through Airtable script [here](https://airtable.com/apphgMmDaoUJM0Kqp/tblVmSzyftcUiAbU7/viwx4z9dixFnlOLK1?blocks=hide) - but I'll have this done on my end too just to be 100% sure there aren't regressions.


Library/ViewModels/BackerDashboardViewModel.swift
```diff
@@ -118,9 +119,33 @@ public final class BackerDashboardViewModel: BackerDashboardViewModelType, Backe
...
+    let updatedUserWithSavedProjectsCount = graphUserEvent.values()
+      .map { $0.me }
+
     self.updateCurrentUserInEnvironment = user.skip(first: 1)
+      .takePairWhen(updatedUserWithSavedProjectsCount)
+      .map { user, graphEventUser in
+        var updatedUserWithStarredProjectsCount = user
+
+        updatedUserWithStarredProjectsCount.stats.starredProjectsCount = graphEventUser.stats
+          .starredProjectsCount
```
scottkicks, line 145: it might be nice, at some point, to update the naming here since we call them saved projects and use a heart icon.

scottkicks (approved):
LGTM! Nice `.debounce` addition  👍

Library/ViewModels/BackerDashboardViewModel.swift
msadoon: Yea definitely. We have this big gap in our models for `User` and `GraphUser`. A lot of pages use the v1 endpoints that map to `User`. We'll at some point convert `User` to gql endpoints app-wide...so good opportunity to do it then.

PR #1810, opened 2023-04-09
[MBL-658] Replace Real Optimizely Client With Mock

msadoon:
# 📲 What

To address our deadline of shipping the removal of Optimizely by April 17, here is a tiddy solution that should hold us over until we implement our next feature flagging tool. Probably Firebase Remote Config.

This [document](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2337996801/Optimizely+Removal) goes into more detail about the hows and whys.

# 🤔 Why

Removing Optimizely saves us an expensive subscription that can be replaced for free by more useful tools like Firebase Remote Config.

# 🛠 How

`MockOptimizelyClient` is actually a 1-1 replacement of the core functionality of the real `Optimizely` client.

In fact we pass around a `OptimizelyClientType` which is a protocol that is mapped to the functions used in `Optimizely`. we pass it around via our `AppEnvironment.current.optimizelyClient` singleton variable.

So in the sake of saving the time of removing each individual flag in each use case throughout our app, why not hardcode those values as they are now on `Optimizely` dashboard?

We don't need to worry about older versions either, because as that document goes into more detail, even if the configuration of the real Optimizely SDK fails, we handle all experiments as a `control` case and all features as a `false` case. So at worst features that were enabled for versions 5.4.0 (project story tab), 5.5.0 (stripe payment settings) and 5.6.0 (stripe payment sheet) stop working _until_ users update to this version 5.7.0.

Updating those users is probably going to be a campaign with in-app notifications down the line.

The thing to note is this `MockOptimizelyClient` is good enough that it covers of the entire scope of work for this [ticket](https://kickstarter.atlassian.net/browse/MBL-658).

To get a sense of our active _daily_ users for each version, check our app analytics in the document above.

# ✅ Acceptance criteria

- [ ] Ensure existing features and experiments behave exactly as they do in the production app version.

# ⏰ Developer QA

**Step 1. Record the optimizely variable values after launching this pr in a build.**

Experiments
`DiscoveryViewModel` `activateNativeProjectCardsExperiment` —> .control
`OptimizelyExperiment` `nativeProjectCardsExperimentVariant` —> .control
`AppDelegateViewModel` `shouldSeeCategoryPersonalization` —> .control
`AppDelegateViewModel` `shouldGoToLandingPage` —> .control
`DiscoveryPageViewModel` `shouldShowPersonalization` —> .control

Features

`AppDelegateViewModel` `featureConsentManagementDialogEnabled` —> false
[...]

PR #1815, opened 2023-04-25
[MBL-720] Configure Remote Config

scottkicks:
# 📲 What

Configures a new `RemoteConfigClient` responsible for loading RC default values, fetching/activating values from the cloud, and providing helpers for determining specific key values.

# 🤔 Why

Remote Config is the feature flagging tool that we are replacing Optimizely with.

# 🛠 How

1. Once FirebaseApp.configure() is called in AppDelegate, we set a `RemoteConfigClient` instance in AppEnvironment to be used throughout the app.
2. On init the new client:
   - loads default values that we may want to change in the future
   - fetches any new values from the cloud. You can actually find these stored in a cached holding pattern on your device
   - "activates" those fetched values. When this happens, it applies those fetched values on top of our existing default values
3. RemoteConfigClient also helps us query our RC values. The way it works is that it gives us either a value from the cloud, if it found one, or a default value based on the provided key.

I've set up 2 keys in our [Kickstarter-Alpha firebase project](https://console.firebase.google.com/u/0/project/kickstarter-ios-alpha-94cf4/config) for testing. Changes to those will reflect in our Feature Flag screen in real time once I get to that bit.

### **Important note regarding the configuration crashes I was seeing during this setup**

In order for RemoteConfig to be initialized the following has to be true:
1. The firebase instance _has to be configured first_
2. Only one copy of firebase can exist in an app. Since we're configuring firebase in AppDelegate we import Firebase into the main target. I tried to keep our new RemoteConfigClient in the Library target, but that requires importing Firebase there as well. However, since the, main target was already importing Library, and thus firebase again, a dependency conflict occurred showing the following error in the console `Class FIRApp is implemented in both...` causing a crash. I needed to keep RemoteConfigClient in the main target to get around this.

    Google intentionally does this to keep multiple copies of Firebase from existing in one app. 
    See [this doc on the Firebase iOS SDK GitHub](https://github.com/firebase/firebase-ios-sdk/blob/master/docs/firebase_in_libraries.md) for more details.

# ✅ Acceptance criteria

- [ ] The should get the correct values from our [firebase alpha project](https://console.firebase.google.com/u/0/project/kickstarter-ios-alpha-94cf4/config) when RemoteConfigClient.fetchCloudValues is called.
- [ ] Consent Management Dialog and Facebook Interstitial behave accordingly based on the RC value

Use the `RemoteConfigFeature+Helpers` to verify

# ⏰ TODO

- Updating the feature flag debug menu and user defaults will be done in the [next ticket](https://kickstarter.atlassian.net/browse/MBL-721)


GoogleService-Info.plist
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
scottkicks, line 1: I deleted and redownloaded/added this file as I was debugging. It should be the same file and its just for the Alpha build.

Library/AppEnvironment.swift
```diff
@@ -241,6 +247,7 @@ public struct AppEnvironment: AppEnvironmentType {
     optimizelyClient: OptimizelyClientType? = AppEnvironment.current.optimizelyClient,
     pushRegistrationType: PushRegistrationType.Type = AppEnvironment.current.pushRegistrationType,
     reachability: SignalProducer<Reachability, Never> = AppEnvironment.current.reachability,
+    remoteConfigClient: RemoteConfigClientType? = AppEnvironment.current.remoteConfigClient,
```
msadoon, line 250: remember to update `pushEnvironment` too (only way I knew was looking at the old `optimizelyClient`) in this file on an older branch.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -386,7 +386,6 @@
 		473DE01A273C757D0033331D /* ProjectRisksDisclaimerCellViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 473DE019273C757D0033331D /* ProjectRisksDisclaimerCellViewModelTests.swift */; };
 		4746FFF3272C564C00EC3429 /* ProjectEnvironmentalCommitmentCellViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4746FFF2272C564C00EC3429 /* ProjectEnvironmentalCommitmentCellViewModelTests.swift */; };
 		4746FFF5272C588900EC3429 /* ProjectEnvironmentalCommitmentDisclaimerCellViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4746FFF4272C588900EC3429 /* ProjectEnvironmentalCommitmentDisclaimerCellViewModelTests.swift */; };
-		4748C17C25B79DA20098E89E /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 4748C17B25B79DA20098E89E /* GoogleService-Info.plist */; };
```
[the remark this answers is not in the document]
msadoon: If you need to start on a new branch, try reverting these changes locally (`git reset --soft commtHash`) and applying the stash to a new branch without the file adds/deletes.

Kickstarter-iOS/Library/RemoteConfigClient.swift
```diff
@@ -0,0 +1,62 @@
+import Firebase
```
msadoon, line 1: So I think we don't need this file.

Here's my thinking: 

The pattern we followed for `OptimizelyClient` is to create it in `AppDelegate` and then conform a protocol for functions we need from it. It was `OptimizelyClientType`.

Then all we do is pass around that `OptimizelyClientType` instance via `AppEnvironment`. 

This is why the real `OptimizelyClient` could easily be replaced with the mock version when we removed the library.

With the new `RemoteConfigClientType` we already conform it to whatever methods we get access to in `RemoteConfig.remoteConfig()`and pass around the `RemoteConfigClientType`, so there is no need for this wrapper class around `RemoteConfig.remoteConfig()` singleton. Just like there was no need for a wrapper around `OptimizelyClient` from `Optimizely` framework.

It's also easier to mock - for that matter add a `MockRemoteConfigClient` in the same way we had `MockOptimizelyClientType`.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -188,6 +188,8 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
           AppEnvironment.current.ksrAnalytics.logEventCallback = { event, _ in
             Crashlytics.crashlytics().log(format: "%@", arguments: getVaList([event]))
           }
+
+          AppEnvironment.updateRemoteConfigClient(RemoteConfigClient())
```
msadoon, line 192: So my top level pr review comment will address the changes here.

msadoon (changes requested):
Okay lots of writing here, hehe, please bear with:

So generally pretty good start to setting this up. It took me a while to get a grasp of the different ways in which we could setup `RemoteConfig` and I found that there's basically three ways:

1. [Fetch and activate on load](https://firebase.google.com/docs/remote-config/loading#strategy_1_fetch_and_activate_on_load)
2. [Activate behing a loading screen](https://firebase.google.com/docs/remote-config/loading#strategy_2_activate_behind_loading_screen)
3. [Load new values for startup](https://firebase.google.com/docs/remote-config/loading#strategy_3_load_new_values_for_next_startup)

The tutorial was doing no. 2, but because we don't have a loading screen I suggest no. 3.

1. `AppDelegate` `configureFirebase` output should call a new function `configureRemoteConfig` instead of directly calling `AppEnvironment.updateRemoteConfigClient(RemoteConfigClient())`...because there are other steps here for sending configuration signals into the view model.

2. In that function configure `RemoteConfig` in this way:

```
func configureRemoteConfig() {
	let remoteConfig = RemoteConfigClient()

	AppEnvironment.updateRemoteConfigClient(remoteConfig)

    let appDefaults: [String: Any?] = [
      RemoteConfigFeature.consentManagementDialogEnabled.rawValue: false,
      RemoteConfigFeature.facebookLoginInterstitialEnabled.rawValue: false
    ]
    RemoteConfig.remoteConfig().setDefaults(appDefaults as? [String: NSObject])

	RemoteConfig.remoteConfig().activate { _, error in
		guard let remoteConfigActivationError = error else { 
			print("🔮 Remote Config SDK Successfully Activated")

			// TODO: Eventually replace self.viewModel.inputs.didUpdateOptimizelyClient(optimizelyClient), when we remove `MockOptimizelyClient`
			return 
		}

		print("🔴 Remote Config SDK Activation Failed with Error: \(remoteConfigActivationError.localizedDescription)")

		Crashlytics.crashlytics().record(error: remoteConfigActivationError)

		self.viewModel.inputs.optimizelyClientConfigurationFailed()
		// we'll rename this later but keep it this way for now - because notifications are attached to it.
	}

	RemoteConfig.remoteConfig().fetch { }

	self.remoteConfigUpdateListener = RemoteConfig.remoteConfig().addOnConfigUpdateListener { }
	// This will be a global private var in `AppDelegate`, remember to deallocate when `applicationWillTerminate` delegate is called - https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623111-applicationwillterminate. (New method + signal to and out of view model)
}
```

So how did I know to set it up this way? I looked at the previous Optimizely [setup](https://github.com/kickstarter/ios-oss/pull/1810/files#diff-2b5cc9e2481b28948ced1e33f97217f053db9fd9c2de5070819b45ff03463369L394).

So just doing the reading for no. 3. I understood this setup to be what they meant.
It might not work this way..but I thought I'd suggest because it seemed to be the fastest, least complicated and most similar to the way Optimizely loaded the FFs.

Two cents on testing:
Load the app and see that the features keys on Firebase are accurately reflected. 
(ie. campaign tab is showing, stripe payment sheets are showing) - those are easiest to test because they are large UI things.

Then go ahead and turn those features off on the dashboard. Continue to test the active app and switch between features. Those features should not be hidden at this point. Even backgrounding/foregrounding shouldn't hide them.

Third, kill and relaunch the app. Our dashboard flags should be reflectedand those features should be hidden this gives us confidence our`addOnConfigUpdateListener` cached the changes last time the app was active.

Let me know your thoughts!

Library/RemoteConfigClientType.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public protocol RemoteConfigClientType: AnyObject {
```
msadoon, line 3: Good job using this pattern!

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +422,48 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+      // we'll rename this later but keep it this way for now - because notifications are attached to it.
+    }
+
+    AppEnvironment.current.remoteConfigClient?.fetch { _, error in
+      if let remoteConfigFetchError = error {
+        print("🔴 Remote Config SDK Fetch Failed with Error: \(remoteConfigFetchError.localizedDescription)")
+      }
+    }
+
+//    self.remoteConfigUpdateListener = AppEnvironment.current.remoteConfigClient?.addOnConfigUpdateListener { }
```
scottkicks, line 464: @msadoon this was throwing a compile error saying that `addOnConfigUpdateListener` doesn't exist on the RemoteConfig.remoteConfig() instance. I see it in the docs, but I commented out for now.

scottkicks:
@msadoon this latest commit makes the suggested changes. However, this doesn't compile yet, I just wanted to get it up so you can take a look.

The error I'm getting is that  `RemoteConfig does not conform to expected type RemoteConfigClientType`. I tried to add all of the methods that I think are "required" into this type, but I'm not sure what I'm missing yet.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +426,48 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+      Crashlytics.crashlytics().record(error: remoteConfigActivationError)
+
+      self.viewModel.inputs.optimizelyClientConfigurationFailed()
+      // we'll rename this later but keep it this way for now - because notifications are attached to it.
+    }
+
+    AppEnvironment.current.remoteConfigClient?.fetch { _, error in
+      if let remoteConfigFetchError = error {
+        print("🔴 Remote Config SDK Fetch Failed with Error: \(remoteConfigFetchError.localizedDescription)")
```
msadoon, line 464: Can remove these lines - because we just need to call `fetch`. no need to error handle or anything..from my reading the firebase instructions it seems like a trigger for the `addOnConfiguUpdateListener`

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +426,48 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
     task.resume()
   }
+
+  private func configureRemoteConfig() {
+    let remoteConfig = RemoteConfig.remoteConfig()
+
+    // WARNING: Don't actually do this in production!
+    let settings = RemoteConfigSettings()
+    settings.minimumFetchInterval = 0
+    remoteConfig.configSettings = settings
```
msadoon, line 436: Remove these lines - those Firebase instructions indicate on step 2:

```
 Real-time updates bypass the minimum fetch interval setting.
```

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +426,48 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        // TODO: Eventually replace self.viewModel.inputs.didUpdateOptimizelyClient(optimizelyClient), when we remove `MockOptimizelyClient`
+        return
+      }
+
+      print("🔴 Remote Config SDK Activation Failed with Error: \(remoteConfigActivationError.localizedDescription)")
+
+      Crashlytics.crashlytics().record(error: remoteConfigActivationError)
+
+      self.viewModel.inputs.optimizelyClientConfigurationFailed()
+      // we'll rename this later but keep it this way for now - because notifications are attached to it.
```
msadoon, line 459: Add a `TODO`

GoogleService-Info.plist
msadoon: Ok so let's go ahead and remove this file in this pr.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -10611,6 +10634,11 @@
 			package = 606754B728CF8A190033CD5E /* XCRemoteSwiftPackageReference "facebook-ios-sdk" */;
 			productName = FacebookLogin;
 		};
+		606C45FB29FAD091001BA067 /* FirebaseRemoteConfig */ = {
```
msadoon, line 10637: Shouldn't see this...we're using the existing Firebase dependency for now.

Library/RemoteConfigClientType.swift
```diff
@@ -0,0 +1,35 @@
...
+
+  var localizedDescription: String {
+    switch self {
+    case let .activate(error):
+      return "Error activating Remote Config \(error?.localizedDescription ?? "")"
+    case let .fetch(error):
+      return "Error fetching Remote Config \(error?.localizedDescription ?? "")"
+    }
+  }
+}
```
msadoon, line 16: Can remove because we don't use/need the error.
We can add this back in later if required.

Library/MockRemoteConfigClient.swift
```diff
@@ -0,0 +1,36 @@
+import FirebaseRemoteConfig
+import Foundation
+
+public enum MockRemoteConfigError: Error {
+  case generic
+
+  var localizedDescription: String {
+    return "Mock Remote Config Error"
+  }
+}
```
msadoon, line 10: Again we don't need this error - there's no place to test it. So remove it.

Library/RemoteConfigClientType.swift
```diff
@@ -0,0 +1,35 @@
...
+
+extension RemoteConfigClientType {
+  /* Returns all features the app knows about */
+
+  public func allFeatures() -> [RemoteConfigFeature] {
+    return RemoteConfigFeature.allCases
+  }
+
+  public func bool(forKey key: RemoteConfigFeature) -> Bool {
```
msadoon, line 32: Because we aren't tied to the `bool(forKey:` naming structure - let's go ahead and use our own naming structure `isFeatureEnabled` made more sense imo.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +426,48 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    }
+
+    AppEnvironment.current.remoteConfigClient?.fetch { _, error in
+      if let remoteConfigFetchError = error {
+        print("🔴 Remote Config SDK Fetch Failed with Error: \(remoteConfigFetchError.localizedDescription)")
+      }
+    }
+
+    // TODO: self.remoteConfigUpdateListener = AppEnvironment.current.remoteConfigClient?.addOnConfigUpdateListener { }
+    // This will be a global private var in `AppDelegate`, remember to deallocate when `applicationWillTerminate` delegate is called - https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623111-applicationwillterminate. (New method + signal to and out of view model)
```
msadoon, line 469: So two things - 
1. We need to test that this function in-app for those "next launch" steps.
2. We need to find a way to call it through `RemoteConfigClientType`protocol.

Library/MockRemoteConfigClient.swift
```diff
@@ -0,0 +1,36 @@
...
+    value.bool = self.features[key ?? ""] == true
+
+    return value
+  }
+
+  public func activate(completion _: ((Bool, Error?) -> Void)?) {}
+
+  public func fetch(completionHandler _: ((RemoteConfigFetchStatus, Error?) -> Void)?) {}
+
+  public func setDefaults(_: [String: NSObject]?) {}
```
msadoon, line 35: These are just stubbed functions, but just noting here that the next step is to test them in the view controllers' that use them - probably why tests are failing at the moment.

msadoon:
@scottkicks made some additional comments - were making some progress here.
Let's try to see if this strategy (update on next launch) will work by running build as-is on this branch and testing the scenario described above.

Kickstarter.xcodeproj/project.pbxproj
scottkicks: So the firebase package is only added to the main target. In order for these new files that have been added to the Library target to access remote config (` RemoteConfig.remoteConfig()`) we need to add FirebaseRemoteConfig to the same target

Kickstarter-iOS/AppDelegate.swift
msadoon: Let's discuss in our 1-1

scottkicks:
@msadoon i've made the changes we worked through today. A couple of issues still coming up...

1. RemoteConfig is still not conforming to RemoteConfigClientType even after the package upgrade
2. I started to update `didUpdateOptimizelyClient`, but it started leading me down a path that quickly became complicated. So I've left that commented out for now. It looks for an argument of OptimizelyClientType so it's not as simple as we first thought

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -422,6 +426,47 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        // TODO: Rename/rework self.viewModel.inputs.didUpdateOptimizelyClient(optimizelyClient)
+        return
+      }
+
+      print("🔴 Remote Config SDK Activation Failed with Error: \(remoteConfigActivationError.localizedDescription)")
+
+      Crashlytics.crashlytics().record(error: remoteConfigActivationError)
+
+      self.viewModel.inputs.optimizelyClientConfigurationFailed()
+      // TODO: Rename/rework input signal with Optimizely gone.
```
msadoon, line 455: We'll discuss how to change these in a 1-1 Monday (`line 446` and `line 455`)
Should be pretty easy, we can pair on it.

Kickstarter.xcodeproj/project.pbxproj
msadoon, line 1: Yea so I went ahead and made some of those "Link Binary With Library" changes from the xcode upgrade here, so we can merge this in first and minimize conflicts in that branch.

Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
msadoon, line 1: Again just changes related to updating ReactiveExtensions and Prelude in SPM

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/RemoteConfig/RemoteConfigClientType.swift
```diff
@@ -0,0 +1,25 @@
+import FirebaseRemoteConfig
+import Foundation
+
+public protocol RemoteConfigClientType: AnyObject {
```
msadoon, line 4: This was your protocol - the reason it wasn't mapping to the `RemoteConfig.remoteConfig()` object was because that was ObjC. And - I had to look this up - there's quite a bit of annoying bridging that needs to be done when conforming ObjC to Swift code even if its' a 1-1 protocol. Which leads me to...

Library/RemoteConfig/RemoteConfligClient.swift
```diff
@@ -0,0 +1,32 @@
+import FirebaseRemoteConfig
+
+public class RemoteConfigClient: RemoteConfigClientType {
```
msadoon, line 3: ...this file! Which is simply a wrapper class around our `RemoteConfig.remoteConfig()` ObjC object.
Notice we hide the actual object as a `private var` and then expose the functions through our protocol.
FYI: this works because its' a _Swift_ class not an _ObjC_ class.

So let's go ahead and test this class with out `MockRemoteConfigClient`.

msadoon (changes requested):
Ok, almost good to go!
So I thought it might faster if I merged changes to update the protocol conformance issue.
It's tested and working as expected - I even got the consent management dialog after setting the dashboard flag to true.
Only 2 things remain before merging:
1. Use the `MockRemoteConfigClient` to write tests for `RemoteConfigClient`
2. Let's handle those input signals for Optimizely setup.

We'll talk monday and wrap this one up eod!

Library/RemoteConfig/RemoteConfligClient.swift
scottkicks: ohh makes total sense!

msadoon (approved):
Good to go - if tests pass.
When this ends up going to qa party there is another step related to turning on Realtime Config (which I didn't yet go through with you for Production, but is done on Alpha).

PR #1820, opened 2023-05-11
[MBL-624] XCode 14.3 Update

msadoon:
# 📲 What

We need to upgrade our project from XCode 13.4.1 to some version of XCode 14 in order to keep shipping our app. [Source](https://developer.apple.com/documentation/xcode-release-notes).

# 🤔 Why

Apple requirement, also we should always be up to date on latest Swift versions. In this we support Swift 5.8

# 🛠 How

We saw lots errors of this ilk when first building with Xcode 14.2:

image

**Fix attempt 1**

Initially we thought it was ReactiveSwift/ReactiveExtensions/Prelude.
So we updated each of those for building with XCode 14, imported with SPM, but still there was an error.

Oddly though not all files were breaking for the same errors (ie. something like `.map(first)`) broke in one file but not another.

**Fix attempt 2**

What changed between Xcode 13.4.1 (last 13.0 release) and the next 14 release (14.0) [changelog](https://developer.apple.com/documentation/xcode/customizing-the-build-phases-of-a-target)

Seems like there's a new build system, so might as well have a look at cleaning things up during the build process.

This led to cleaning the the build phases "link binary with libraries"
[relevant learnings info](https://developer.apple.com/documentation/xcode/customizing-the-build-phases-of-a-target)

Basically a Dependency means "we need this framework to be built before we built the target"
And "Link Binary With Libraries" means "we optionally can include these other prebuilt frameworks if the target references any of them".

So you can make a framework a "Dependency" if there's linked frameworks in _it_ that you don't want to include repetitively in "Linked Binary With Libraries".

But still the same reactive swift errors during build, which led to...

** FIx attempt 3 **

Looking at the error itself - ie. It's confused about the type. So maybe explicitly setting the type in the right place makes the compiler happy.

For the screenshot above, turns out the change to get the error to go away is

image

explicitly setting that `let userAttributedChanged` signal to have a type of `Signal<(UserAttribute.Notification?, Bool), Never>` fixes the issue!

Turns out that is the fix for all the other files too.

Maybe new build systems' compiler/syntax reader loses some context when you have too many functions passed into one another.

It's relatively easy to fix... something like `let someEvent = someOtherEvent` needs an explicit type - which will cascade down to all dependent signals.

**Screenshots**
Lots had to be updated for minor visual differences. Technically it's the same simulator, and iOS version as the recording, but after updated my MacOS to Ventura and XCode itself, seems that Apple changed some fonts/spacing on the simulator that made some of our snapshot fail. [reference](https://github.com/pointfreeco/swift-snapshot-testing/issues/450#issuecomment-812147415) 

**Build Settings** Context: Clicking the "Update to Recommended Settings warnings" 

XCode recommends not signing any of our targets so removed those.

Applied the Apple Clang Verifier which is a new build settings called `ENABLE_MODULE_VERIFIER` defined as 
```
Xcode 14.3 includes a new module verifier that generates diagnostics for issues in a framework’s modules. (97345247)
```
but it prevents Stripe framework from linking. So set it to OFF. Not important if all it does is report diagnostics.

I think there were duplicate `STRIP_STYLE` settings that were removed. Here's some [documentation](https://xcodebuildsettings.com/#strip_style) on it.

***XCode 14.2 to XCode 14.3***

The only issue here was updating Kingfisher dependency to support the new version of Swift (5.7 -> 5.8). Noted in this [release](https://github.com/onevcat/Kingfisher/releases/tag/7.6.1).

# ✅ Acceptance criteria

- [x] XCode 14.3 building
- [x] Unit/Snapshot tests passing

# ⏰ TODO

- [x] Fix Unit/snapshot testing
- [x] Run through regression checklist


msadoon:
@scottkicks tests should pass, I just reverted some commits back to when they were passing.

scottkicks (approved):
🥳 Builds, Runs, and tests pass

I can spend time going through the regression checklist as well, but it looks good to me

msadoon:
> 🥳 Builds, Runs, and tests pass
> 
> I can spend time going through the regression checklist as well, but it looks good to me

Awesome! We can start using XCode 14.3 🥳 .
No worries with the regression checklist, we can run through that before releasing. I ran through most steps there anyway, should be fine.

PR #1821, opened 2023-05-14
[MBL-773] CircleCI Workspaces Integration

msadoon:
# 📲 What and  🤔 Why

So our CI pipeline has a lot of repeated steps in it:
Example all of these steps run for every test job (that's 3x for kickstarter-framework-ios, ksapi, library)

```
    - checkout
    - run:
        name: SPM SSH Workaround
        command: *spm_ssh_workaround
    - run:
        name: Store Xcode Version
        command: xcodebuild -version > xcode_version.txt
    - run:
        name: Bootstrap SPM
        command: xcodebuild -resolvePackageDependencies -clonedSourcePackagesDirPath SourcePackages
    - run:
        name: Bootstrap
        command: make bootstrap
    - run:
        name: Download GraphQL Schema
        command: bin/apollo-schema-download.sh
    - run:
        name: Pre-load simulator
        command: *preload_simulator
    - run:
        name: Run tests
        command: make test
        no_output_timeout: "20m" # 20 minutes
   - run:
       name: Install Bundler
       command: bundle check || bundle install
       environment:
          BUNDLE_JOBS: 4
          BUNDLE_RETRY: 3
```
Because these steps are run in parallel, they really don't blow up the time it takes for our suite to finish (approx. 30 min)

However why not run all the steps _once_, incorporate the environment they ran in (for example all the gems from bundle install) and re-use that environment for our test jobs.

Workspaces allows us to do this.

# 🛠 How

Workspaces has two steps essentially:

a `persist_to_workspace` step and `attach_workspace` step.

for our `test_job`  and `distribute_job` we `attach_workspace`. The `build_and_cache` step has `persist_to_a_workspace` step and sets up the environment with gems, swift package dependencies and the code itself.

# 🏎 Performance

**Side-by-side comparison of reduction in time.**

1. Last time an alpha build got deployed:

Screenshot 2023-05-16 at 7 49 48 PM

2. New alpha build deploy

Screenshot 2023-05-16 at 8 01 05 PM

Note the tests do not run until the build and cache step completes.

3. CircleCI reports a 14% decrease in average time to finish.

Screenshot 2023-05-16 at 8 00 55 PM

We might be expected it to be much faster but these jobs were run while CircleCI was in "maintenance" mode, so that probably effected the total time. Let's try this idea for now, and  judge its' efficiency after we get some builds distributed.

In theory it "should" improve total time to build, but we need CI to be out of maintenance mode to see the real results.

## Noted for posterity

Quick smoke test for new ci build time vs old build time (2 suites running same on conditions in CircleCI):

Screenshot 2023-05-17 at 1 10 30 PM

top workflow is using workspaces, bottom workflow using caching.

**Caching is faster for total time taken**
Pros:
- Concurrent workflows run, meaning even with repeated steps, the longest running test suit will be the bottle neck.

Cons:
- Messier `config.yml` file, less maintainable.
- Not using newer circleci options like orbs for MacOS and `prelaunch-simulator` commands.

**Workspaces is faster for individual test suites**
Pros:
- More maintainable `config.yml` file, all downloads/config/gems/code/packages done in `build and cache`. Test job and distribute job just use the workspace.
- Each test job after build-and-cache takes between 3-10 minutes less time.

Cons:
- Consecutive workflows run, adds time. Meaning `build and cache` takes 10 min longer and runs first, before test suites. Bottleneck is `build-and-cache` time + longest test suite time.

# ✅ Acceptance criteria

- [x] CirecleCI steps pass - and a deployment (alpha) should be passing as well.

# ⏰ TODO

- [x] Restore SPM step doesn't have permisions to unarchive cache. 
- [x] Ensure tests passing and alpha gets deployed.


msadoon:
@scottkicks I know you  already approved this, but I just compared two workflows using the old config file and the newer. If you could take a look at `Note for posterity` and let me know if you think its' worth merging this PR.

PR #1824, opened 2023-05-30
[MBL-841] Release 5.8.0 to Main

scottkicks:
This is the release from 5.8.0 to Main

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -9213,7 +9213,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MARKETING_VERSION = 5.7.0;
+				MARKETING_VERSION = 5.8.0;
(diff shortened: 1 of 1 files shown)


scottkicks:
Since Mubarak is on vacation, this is a non reviewed merge.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -9239,6 +9239,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
+				MARKETING_VERSION = 5.8.0;
```
msadoon, line 9242: Do not introduce uneccessary changes here. If it previously was not there, no need to add it. Only need modifications with this. Ie. 5.7.0 -> 5.8.0

msadoon (commented):
Ok, mostly done correctly. 
So next up. I want you to write a separate PR - for removing unneeded additions of `MARKETING_VERSION = 5.7.0`. Only keep the changes that update the number NOT add the marketing version where it previously was not there.

PR #1826, opened 2023-06-12
[MBL-841] Release 5.8.0 to Main 

scottkicks:
This is the release from 5.8.0 to Main with the correct marketing version update.


scottkicks:
@msadoon please verify that the main project and all of its targets have a correct marketing version of 5.8.0

msadoon (approved):
Looks good, but remember as agreed only assign a PR once all tests pass and not in a draft state!

msadoon:
> @msadoon yes just wanted to get this reviewed since most tests cases won't be impacted by these changes

Yep, I just want us to follow the agreed way so we don't get miss steps in the future. Always feel free to tag me in a PR and ask questions without assigning. That way when multiple members are on the team you can always expect a PR to be assigned once it's ready to be merged. In my experience it causes less confusion especially with large PRs, (ie. commits being added while review is happening, random test failures, etc.)

PR #1828, opened 2023-06-15
[MBL-854] Gate Creator Dashboard Behind Feature Flag

scottkicks:
# 📲 What

Gates displaying creator tools (dashboard) behind the [new feature flag](https://github.com/kickstarter/ios-oss/pull/1827).

# 🤔 Why

Since the creator tools (dashboard) is going to be depreciated, and this is a pretty big feature in the app currently, doing so behind a feature flag is safer and gives us more control over the deprecation.

# 🛠 How

There are a couple of entry points to the creator dashboard. 
1. The Dashboard tab in the main tab bar that only shows when a creator is logged in.
2. Deep links handled in RootTabBarViewController & ProjectPageViewController

We can use the `featureCreatorDashboardEnabled()` helper method to gate the following methods responsible for displaying the dashboard tab.

- `RootViewModel.generatePersonalizedViewControllers()`: which view controllers to add to the root tab view.
   - Creates an array with _.dashboard(isMember: Bool)_ and _.profile(isLoggedIn: Bool)_
    - The setViewControllers signal output is then used in the `RootTabBarViewController` to determine which VC to instantiate for each tab.
 - `RootViewModel.tabData(forUser: User)`: what tab data to use for each tab.
      - Signal is called when there is a _currentUser_ and _viewDidLoad_, or _userLocalePreferencesChangedProperty_ (logged in/logged out), is emitted.
      - Based on whether the user is a member (determined by whether the user.stats.memberProjectsCount > 0)

The deep link cases can be found:
- access to switch to the dashboard:  `line 170`  `RootTabBarViewController`
- access to feature of the creator dashboard project activities: `line 231` `RootTabBarViewController`
- access to feature of the creator dashboard creator messages: `line 21`7 `RootTabBarViewController`
- creator dashboard from the project page: `line 264` `ProjectPageViewController`

# 👀 See

| Dashboard Enabled | Dashboard Disabled |
| --- | --- |
| Simulator Screen Recording - iPhone 14 Pro - 2023-06-15 at 09 28 09 | Simulator Screen Recording - iPhone 14 Pro - 2023-06-15 at 09 28 58 |

# ✅ Acceptance criteria

You can log in as a creator using the therealnativesquad@gmail.com account in 1Password under Kickstarter Native Squad

### When Creator Dashboard is _not enabled_
- [x] The creator dashboard tab should not be available regardless of whether you’re logged in as a backer or creator, or logged out.

### When Creator Dashboard is _enabled_
- [x] The creator dashboard tab should be shown as normal to logged in creators only.

# ⏰ TODO

- [x] Discuss the fact that testing cases where RemoteConfig flags are enabled isn't possible right now because the RemoteConfig SDK's obj-c initializer isn't available.


scottkicks:
@msadoon re: our sync conversation around testing cases where our Remote Config flags are true...

See the FIXME comments I've left in `RootViewModelTests`. 
You can step through `func tabData(forUser...)`on `line: 498` in `RootViewModel` to see that even though the remoteConfigClient in AppEnvironment is being set to the Mock with the flag set to true via the tests, the `featureCreatorDashboardEnabled()` call in the view model is returning false.

My current theory is that its related to the commented out tests in `RemoteConfigFeature+HelpersTests` that deal with the RemoteConfigValue not initializing. 

Lmk if you have any thoughts on what might be happening here. Thanks!

msadoon (changes requested):
Overall good job hiding the `DashboardViewController` on first and subsequent launches.

I found that there are scenarios where the app just expects `DashboardViewController` is already shown to the user. Deeplinks for example are routes we support to open the dashboard view controller that can be used even if we've hidden it. So with that in mind let's also gate any events that change tabs when the `DashboardViewController` isn't shown.

The cases found were: 
- access to switch to the dashboard 
  - `line 170` `RootTabBarViewController`
- access to feature of the creator dashboard project activities
  -  `line 231` `RootTabBarViewController`
- access to feature of the creator dashboard creator messages
  -  `line 217` `RootTabBarViewController`
- creator dashboard from the project page
  - `line 264` `ProjectPageViewController`

So my suggestion is to gate these with the `featureCreatorDashboardEnabled` function. Make that `guard` check the first thing each function does before letting control flow through. Some tests might need corrections, no need to write both true and false cases, just make the exist ones to pass using `MockRemoteConfigClient`. 

Which leads to the correction for `MockRemoteConfigClient`

Here is the correction, basically it overrides the value we check `boolValue` from RemoteConfig with a value we can change `bool`. We have to do this because `RemoteConfigValue` has no exposed initializer in the public API from Firebase.

```
private class MockRemoteConfigValue: RemoteConfigValue {
  var bool = false
  
  override var boolValue: Bool { bool }
}
```

All your tests should pass then. I ran them myself and they did. Good job on the tests.

I'll do actual in-app testing on the next review if there's no more code changes. I didn't do it on this one because I'd like to test all the final code changes in app. Let's aim to make the next review about testing the functionality and approved.

Library/ViewModels/RootViewModel.swift
```diff
@@ -488,21 +488,34 @@ private func generateStandardViewControllers() -> [RootViewControllerData] {
...
+      : [
+        .home(index: 0), .activity(index: 1), .search(index: 2),
+        .profile(avatarUrl: (user?.avatar.small).flatMap(URL.init(string:)), index: 3)
+      ]
+  } else {
+    items = [
       .home(index: 0), .activity(index: 1), .search(index: 2),
       .profile(avatarUrl: (user?.avatar.small).flatMap(URL.init(string:)), index: 3)
     ]
+  }
```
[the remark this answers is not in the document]
scottkicks: oh nice! i always forget about using tuples in switch statements

scottkicks:
@msadoon each of your suggestions have been addressed. Thanks for catching those other cases where the dashboard is shown. I _think_ that should be everything. If anyone manages to get to the dashboard in an unforeseen case, the banner i'll be adding next will at least let them know it'll be going away soon.

also...

it looks like gating the deep link handlers doesn't impact any tests. All existing view controller tests still pass after the changes. And i've already covered the switchToDashboard flow, that these view controller methods end up calling, with unit tests in my previous commits. I think we should be good to go.

msadoon (changes requested):
Ok, code changes look good and in-app functionality works as expected for the most part, but I found an inconsistency for the tab items:

https://github.com/kickstarter/ios-oss/assets/4282741/9b435a06-4d36-427d-bb5e-a427010a0450

Here's a recording and you'll notice that being logged in as a creator, if I turn off the beta flag, and switch to the profile tab, the user is shown two profile tabs, but the second profile tab is the Dashboard (it should be hidden). Relaunches fixes the issue.

If logged in as a backer then turning on the beta flag causes my profile tab to change to Dashboard but I don't see an additional tab - for Dashboard. Relaunch fixes the issue.

This bug occurs on production as well, without using the Beta Tools at all. Switching the dashboard values on/off there is a transition period before the tab bar fixes itself (takes subsequent relaunches).

So a small UI bug, but it does effect the production build and will end up confusing users, so we should probably fix it.

Can we investigate this issue and fix it. Big clue is that relaunch corrects the tabs, so maybe the tab bar needs to be updated? Use

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/RootViewModel.swift
```diff
@@ -500,13 +500,13 @@ private func tabData(forUser user: User?) -> TabBarItemsData {
     (user?.stats.memberProjectsCount ?? 0) > 0
   let items: [TabBarItem]
 
-  switch (featureCreatorDashboardEnabled(), isMember) {
-  case (false, _), (true, false):
+  switch isMember {
```
msadoon, line 503: So the idea is that the tabs are based on the user. And they don't need gating because this tab data is only used for styling - the actual setting of the tabs in the tab bar is done in `setViewControllers`.

Kickstarter-iOS/Features/RootTabBar/RootTabBarViewController.swift
```diff
@@ -260,46 +260,70 @@ public final class RootTabBarViewController: UITabBarController, MessageBannerVi
...
+        _ = self?.tabBarItem(atIndex: index)
+          ?|> profileTabBarItemStyle(isLoggedIn: true, isMember: data.isMember)
+          ?|> UITabBarItem.lens.image .~ defaultImage
+          ?|> UITabBarItem.lens.selectedImage .~ selectedImage
       }
+      dataTask.resume()
     }
   }
 
+  fileprivate func isDashboardViewControllerDisplayable() -> Bool {
```
msadoon, line 310: This function just checks if the `DashboardTabViewController` is available to display.

Kickstarter-iOS/Features/RootTabBar/RootTabBarViewController.swift
```diff
@@ -260,46 +260,70 @@ public final class RootTabBarViewController: UITabBarController, MessageBannerVi
...
-              ?|> UITabBarItem.lens.selectedImage .~ selectedImage
-          }
-          dataTask.resume()
-        }
+        setProfileImage(with: data, avatarUrl: avatarUrl, index: index)
+      }
+    }
+  }
+
+  fileprivate func setProfileImage(with data: TabBarItemsData, avatarUrl: URL?, index: Int) {
```
msadoon, line 276: Move all the profile downloading to a separate function.

Kickstarter-iOS/Features/RootTabBar/RootTabBarViewController.swift
```diff
@@ -260,46 +260,70 @@ public final class RootTabBarViewController: UITabBarController, MessageBannerVi
       case let .search(index):
         _ = tabBarItem(atIndex: index) ?|> searchTabBarItemStyle
       case let .dashboard(index):
-        _ = tabBarItem(atIndex: index) ?|> dashboardTabBarItemStyle
+        let featureFlaggedTabBarItemStyle = self
+          .isDashboardViewControllerDisplayable() ? dashboardTabBarItemStyle :
+          profileTabBarItemStyle(isLoggedIn: data.isLoggedIn, isMember: data.isMember)
+        _ = tabBarItem(atIndex: index) ?|> featureFlaggedTabBarItemStyle
```
msadoon, line 266: This function is called as a result of `tabData(forUser` and it checks if that tab is available before styling it. So if we sent "Explore, Activity, Search, Dashboard and Profile" to this function - the view controllers may only be "Explore, Activity, Search and Profile". Which means that `Dashboard` case updates its styling to reflect "Profile" and the `Profile` case doesn't update anything - the tab doesn't exist. `self.tabBar.items` is called under the hood for `tabData(forUser` and reflects the view controllers - which are independently set by `setViewControllers` (whenever that happens).

msadoon: The only trade-off is that we lose the Profile user avatar, if they were a creator logged in with the dashboard hidden. I think its' worth it because this only affects logged in creators and not backers and we're removing creator login functionality anyway.

msadoon (commented):
Ok I uploaded fix for the creator dashboard - but the tests are failing because the `GraphSchema` needs to be updated - will take a look at that in my Add New Card Removal PR.

scottkicks:
@msadoon this makes sense because directly gating tabBarItemsData from the ViewModel can easily cause the setting of the view controllers and the associated tab data to be out of sync. Plus updating from the dashboard can have some lag so this gets around that as well.

Why does the graphschema need to be updated? Since you'll be taking care of that in a different PR this one should wait to be merged in until that is resolved.

Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/RemoteConfigFeatureFlagToolsViewController.swift
```diff
@@ -53,15 +53,6 @@ final class RemoteConfigFeatureFlagToolsViewController: UITableViewController {
...
     self.viewModel.inputs.setFeatureAtIndexEnabled(index: switchControl.tag, isEnabled: switchControl.isOn)
-
-    let (feature, _) = self.features[switchControl.tag]
-    if feature == RemoteConfigFeature.creatorDashboardEnabled {
-      NotificationCenter.default.post(
-        name: Notification.Name.ksr_userLocalePreferencesChanged,
-        object: nil,
-        userInfo: nil
-      )
-    }
```
scottkicks: This still might be worth having for testing. With your solution here, nothing will happen when toggling the feature from the local ff menu

msadoon (approved):
Good to go!

PR #1830, opened 2023-06-23
[MBL-856] Banner For Creator Dashboard Deprecation

scottkicks:
# 📲 What

A banner at bottom of dashboard notifying users of planned deprecation.

# 🤔 Why

We'd want to give users a heads up that features they may be used to seeing/using are going to be removed.

# 🛠 How

- Decided to build this view as a SwiftUI View.
- In order to add a SwiftUI view to a UIKit Storyboard, like the Dashboard, we need to add it via a [UIHostingViewController](https://developer.apple.com/documentation/swiftui/uihostingcontroller) as a subview and add constraints programmatically. This is pretty simple as you'll see in the code.
   - we _could_ add using the existing storyboard, but I found that doing it programmatically like this will make it easier to review and remove once the dashboard is fully deprecated.

### Important
- DashboardViewController is a UITableViewController. My understanding here is that TableViewControllers automatically add views to the scrollable content that it contains. Because we want this banner to be "sticky", my solution is to add it directly to the window. This also means that we need to make sure to manually call `removeFromSuperview` in viewWillDisappear so that it does not show anywhere else when switch tabs or presenting other screens.

# 👀 See

| After 🦋 (iPhone 14) | iPhone 8 (smaller screen) |
| --- | --- |
drawing | drawing |

## Translations
| German | Spanish | French | Japanese |
| --- | --- | --- | --- |
| drawing | drawing |  drawing | drawing |

# ✅ Acceptance criteria

- [x] The banner should match [the designs](https://www.figma.com/file/Fi9KZ4Vd1q1fnvWZxIw0tw/Creator-Tool-Deprecation-Message?type=design&node-id=2-184&mode=design&t=kWJlm6BtE1FXcyTF-0) we have in figma
- [x] banner should be a sticky view. Meaning, it should be constrained to the top of the tab bar and should not move when scrolling the dashboard view behind it.
- [x] Banner should not hide any of the content on the Dashboard view
- [x] Banner should only show when the DashboardViewController is presented. If the Creator Dashboard ff is off, the dashboard _as well as_ this new banner should never be seen.

# ⏰ TODO

- [x] The translations for the copy here need to be added, but in order to keep our PRs smaller and easier to review I'll be making a new one to address that first.


Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -198,6 +205,34 @@ internal final class DashboardViewController: UITableViewController {
     }
   }
 
+  fileprivate func setupDashboardRemovalWarning() {
+    if #available(iOS 15, *) {
+      if let window = UIApplication.shared.windows.filter({ $0.isKeyWindow }).first,
+        let tabBar = self.tabBarController?.tabBar {
+        dashboardRemovalWarningHostingController = UIHostingController(rootView: DashboardRemovalWarning())
+        /// Added to the window because views added to a tableviewcontroller are automatically scrollable and we need this to be a sticky view
+        window.addSubview((dashboardRemovalWarningHostingController?.view)!)
```
scottkicks, line 214: calling out my comment here. anything added to a UITableViewController (DasboardViewController), is automatically added as a scrollable object. To get the banner to stick above the tabbar I had to add it to the window directly.

Because of this, its important to remember to remove the view manually on viewWillDisappear which you can see above.

If you know of an alternative here please point it out. thanks!

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -198,6 +205,34 @@ internal final class DashboardViewController: UITableViewController {
...
+          .isActive = true
+        dashboardRemovalWarningHostingController?.view.widthAnchor.constraint(equalTo: window.widthAnchor)
+          .isActive = true
+        dashboardRemovalWarningHostingController?.view.bottomAnchor.constraint(
+          equalTo: tabBar.topAnchor,
+          constant: 0
+        ).isActive = true
+
+        self.tableView.contentInset
+          .bottom = (tabBar.frame.height + (dashboardRemovalWarningHostingController?.view.frame.height ?? 0))
```
scottkicks, line 231: Makes sure that the banner sits above the tabbar.

scottkicks:
@msadoon Since this view is being added directly to the window, see my comment [here](https://github.com/kickstarter/ios-oss/pull/1830#discussion_r1251113890), the DashboardViewController snapshots will still pass as is since the banner isn't technically being added to it.

I think this is ok in this case because the feature will be removed in a month, but if you have ideas on how to capture this in our snapshot tests please lmk. I'm not sure how to reference the window itself, or if we even want to for this test case.

msadoon (approved):
Ok, try to remember not to assign PR's that are in "draft".

I pushed a commit that basically is all the changes I worked on and tested. It uses the basic structure you have here, but cleaned up the code a lot.

There's a lot of changes here so let's go over them on a call, I think its' more efficient that way. You can also question any of the things I've done and I'll be happy to explain them.

In the future let's start with an engineering plan before diving into the code.

The only thing I missed here was not covering the bottom scrolling of the view. I don't think its' super important, but if you think we should leave that area scrollable, we can work on that in the call together.

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
msadoon: So I made sure not to use the `window`, because its' not really a convention we use in the app to add new views.
You can reference the `self.tabBarController` from within any view controller presented by the `tabBar`.
I also used the `MessageBannerViewController` `configureMessageBannerViewController` method to figure out how to add a temporary view to the existing hierarchy, but instead of anchoring to the bottom of the `tabBarController` `view`, anchored to the `top`.

Well go over it in the call 👍🏿

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -61,6 +64,8 @@ internal final class DashboardViewController: UITableViewController {
     super.viewWillDisappear(animated)
 
     self.viewModel.inputs.viewWillDisappear()
+
+    self.removeDeprecationWarning()
```
msadoon, line 68: I used your idea of view's appearing and disappearing which is the correct approach. Good idea to keep these changes within the `DashboardViewController`. I tried it in `RootTabBarViewController` and it would be messy turning it off there for each tab.

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -198,6 +203,43 @@ internal final class DashboardViewController: UITableViewController {
...
+    dashboardViewControllerView.translatesAutoresizingMaskIntoConstraints = false
+
+    NSLayoutConstraint
+      .activate([
+        dashboardViewControllerView.leftAnchor.constraint(equalTo: tabController.view.leftAnchor),
+        dashboardViewControllerView.rightAnchor
+          .constraint(equalTo: tabController.view.rightAnchor),
+        dashboardViewControllerView.bottomAnchor.constraint(
+          equalTo: tabController.tabBar.safeAreaLayoutGuide.topAnchor)
+      ])
```
msadoon, line 240: I used the `MessageBannerViewController` `MessageBannerViewControllerPresenting` `configureMessageBannerViewController` as a guide here. You'll see a lot of similarities in how the view controller is add as a child and the view is added on top of the `tabController` `view`.

It's the same concept because that toast you might have seen after backing is this `MessageBannerViewController` and it's used everywhere in the app, so it's pretty reliable.

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -198,6 +203,43 @@ internal final class DashboardViewController: UITableViewController {
     }
   }
 
+  fileprivate func removeDeprecationWarning() {
```
msadoon, line 206: Using this function, we don't need a global variable in the view controller, we show and remove the view controller and any subviews every time the dashboard is shown and hidden. Limits the scope, basically a little cleaner.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
...
+import SwiftUI
+
+struct DashboardDeprecationView: View {
+  private let contentPadding = 12.0
+  private let imageSizeMultiplier = 1.5
+  private let deprecationDateText = formatted(dateString: "2023-08-14")
+
+  var body: some View {
+    HStack {
+      if let iconImage = image(named: "fix-icon", inBundle: Bundle.framework) {
```
msadoon, line 11: Always use conventions currently within the app.
In this case we can't be sure that "fix-icon" will be guaranteed to exist, so its' better to check for its' existence before creating the `Image`.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
+import Library
+import SwiftUI
+
+struct DashboardDeprecationView: View {
+  private let contentPadding = 12.0
+  private let imageSizeMultiplier = 1.5
```
msadoon, line 6: Try to use constants for hardcoded values at the top of the file, makes changing the same thing in multiple places easier and is more readable.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
...
+  private let deprecationDateText = formatted(dateString: "2023-08-14")
```
msadoon, line 7: Remember we're support multiple time zones and languages so we should use conventions to add dates to any presented strings.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
...
+
+  var body: some View {
+    HStack {
```
msadoon, line 10: Didn't see a need to use `(spacing: 0)` here. Double checked by running app a couple times.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
...
+      if let iconImage = image(named: "fix-icon", inBundle: Bundle.framework) {
+        Image(uiImage: iconImage)
+          .frame(width: contentPadding * imageSizeMultiplier)
+          .scaledToFit()
```
msadoon, line 14: This will ensure that the view is scaled to fit, so only need to set the width above. Also `resizable()` is no longer needed. SwiftUI modifiers work on the previous modifier's returned view. So `frame` --> `scaledToFit` will scaled a resized image. Before it was `resizable` -> `scaledToFit` -> `foregroundColor` -> `frame`. So could be reorganized to be a little cleaner.

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,50 @@
+import Library
+import SwiftUI
+
```
msadoon, line 3: Also I know I needed `@available(iOS 15)` in my SwiftUI view, but you don't need it. My view needed it because of `@FocusState`.
Careful with adding that availability check because we do still support iOS 14.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -2074,6 +2075,7 @@
 		608E7A5428ABE27400289E92 /* SetYourPasswordViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetYourPasswordViewModel.swift; sourceTree = "<group>"; };
 		60DA50E928B68990002E2DF1 /* SetYourPasswordViewModelTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetYourPasswordViewModelTests.swift; sourceTree = "<group>"; };
 		60DA50EF28B69534002E2DF1 /* SetYourPasswordViewControllerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetYourPasswordViewControllerTests.swift; sourceTree = "<group>"; };
+		60DF50962A434E6B002C771F /* DashboardDeprecationView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DashboardDeprecationView.swift; sourceTree = "<group>"; };
```
msadoon, line 2078: Changed the name of the view just to be a little clearer.

msadoon (changes requested):
Tests not passing. Let's fix in  pairing.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift:50:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift:55:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift:236:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -210,8 +211,7 @@ internal final class DashboardViewController: UITableViewController {
     }
   }
 
-  fileprivate func updateTableViewBottomContentInset() {
-    /// Makes sure that the deprecation warning doesn't cover any table view content
+  private func updateTableViewBottomContentInset() {
     if let tabController = self.tabBarController as? RootTabBarViewController,
       let deprecationWarningHostingController = deprecationWarningHostingController {
       self.tableView.contentInset
```
scottkicks, line 217: we actually do need the tab bar height here

Kickstarter-iOS/Features/DashboardDeprecationBanner/DashboardDeprecationView.swift
```diff
@@ -0,0 +1,52 @@
...
+    guard let date = Format.date(
+      from: dateString,
+      dateFormat: "yyyy-MM-dd",
+      timeZone: UTCTimeZone
+    ) else {
+      fatalError("Unable to parse date format")
+    }
+
+    return date
+  }
```
msadoon, line 51: So the reason tests were failing was because these functions were outside the `struct` but declared as `private`. They had the same name as the functions outside `ManagePledgePaymentMethodsViewModel` which were also `private` and outside the `class`. So `ManagePledgeViewControllerTests.testView_CurrentUser_IsBacker()` is the test that was breaking, it is no longer because these functions don't collide anymore.

Kickstarter-iOS/Features/RootTabBar/RootTabBarViewController.swift
```diff
@@ -260,10 +260,11 @@ public final class RootTabBarViewController: UITabBarController, MessageBannerVi
...
       case let .dashboard(index):
-        let featureFlaggedTabBarItemStyle = self
-          .isDashboardViewControllerDisplayable() ? dashboardTabBarItemStyle :
-          profileTabBarItemStyle(isLoggedIn: data.isLoggedIn, isMember: data.isMember)
-        _ = tabBarItem(atIndex: index) ?|> featureFlaggedTabBarItemStyle
+        let style = self
+          .isTabBarItemLastItem(for: index) ?
+          profileTabBarItemStyle(isLoggedIn: data.isLoggedIn, isMember: data.isMember) :
+          dashboardTabBarItemStyle
+        _ = tabBarItem(atIndex: index) ?|> style
```
msadoon, line 267: I had to do some rework here to configure the logic around getting the tab bar styling, which was still breaking.
Basically the logic is now - if the app is styling the dashboard tab bar item, it should not be the last tab bar item. If it is, then style it as profile.
So the cases are:
1) Logged in as backer (creator dashboard flag ON or OFF):
- The `setTabBarItemStyles` function never gets its' tab bar item set as `dashboard`, because the function that calls this one checks if the user is a `member` first.
2) Logged in as a creator (creator dashboard flag ON):
- If set in Beta tools as ON then go from backer --> creator, then the dashboard is visible.
- Turning this flag OFF in beta tools doesn't reflect the hiding of the dashboard until the user relaunches the app.
3) Logged in as a creator (creator dashboard flag OFF):
- If set in Beta tools as ON then go from backer --> creator, then the dashboard is not visible.
- Turning this flag ON in beta tools doesn't reflect the showing of the dashboard until the user relaunches the app.

Validate my logic here. I tested it, but if you can find a missing step I'm happy to rework this.

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
```diff
@@ -198,6 +215,48 @@ internal final class DashboardViewController: UITableViewController {
     }
   }
 
+  private func updateTableViewBottomContentInset() {
+    if let deprecationWarningHostingController = deprecationWarningHostingController {
+      self.tableView.contentInset
+        .bottom = deprecationWarningHostingController.view.bounds
```
msadoon, line 221: Found that the `depecrationWarningHostingController.view.bounds` always returns the `height` and all other dimensions as `.zero` if `viewWillAppear` is called, so need to wait until `viewDidAppear` is called for the views to have correct dimensions. That's why the `self.tabBarController.tabBar.bounds.height` was needed - because it existed before `viewWillAppear` on this view controller.

msadoon (approved):
Tests should pass - I did write a few commits because this PR was just taking too long.
So my advice here is to look at this commits and really understand what the diffs are between commits and why.
Again always happy to jump on a call and address any questions you have.

Otherwise I think this is good to go.

Like I said earlier let's start with an engineering plan first to avoid all this back and forth.

Kickstarter-iOS/Features/Dashboard/Controller/DashboardViewController.swift
scottkicks: nice

Kickstarter-iOS/Features/RootTabBar/RootTabBarViewController.swift
scottkicks: Interesting that I didn't see this bug, but I've tested the current logic for all cases listed and the app is working as expected. Good to go 👍

PR #1835, opened 2023-07-17
[MBL 873] Implement TriggerThirdPartyEvent Mutation

scottkicks:
# 📲 What

Replacing `TriggerCapiEventInput` calls with [`TriggerThirdPartyEvent`](https://staging.kickstarter.com/graphiql)

# 🤔 Why

We're going to handle Facebook CAPI and Google Analytics calls with a single mutation. 
For more details see the [engineering plan](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2434596865/Facebook+s+Conversions+API+Google+Analytics+Combined+GraphQL+Mutation+Engineering+Plan)

# 🛠 How

Remove TriggerCapiEvent related files and references.

Replace current `.triggerCapiEventInput` calls with `.triggerThirdPartyEvent`
We’re currently triggering this mutation in the following ViewModels:
- PledgePaymentMethodsViewModel
- ProjectPageViewModel
- RewardsCollectionViewModel
- ThanksViewModel

Existing parameter differences below:
- `externalId` → `deviceId`
- `eventName` → `eventName`
- `projectId` → `projectId`
- `appData` → `appData` (we can keep hardcoding these values to true, just like we do now because externalId is required)
- we do not need to pass anything into `firebaseScreen`, `firebasePreviousScreen`
- `userId` →`AppEnvironment.current.currentUser?.id`

New Parameters (These are only passed in on the ThanksViewModel):
We can use this existing code…
```
      if let checkoutDataValues = configData.checkoutData {
        ...
      }
``` 

- `pledgeAmount` → checkoutDataValues.revenueInUsd + checkoutDataValues.bonusAmountInUsd  + checkoutDataValues.addOnsMinimumUsd
- `shipping` → checkoutDataValues.shippingAmountUsd
-  `transactionId` → checkoutDataValues.checkoutId
- clientMutationId → ““

# ✅ Acceptance criteria

- [x] Use Charles Proxy, or something similar, to verify that these calls are being sent at the expected times with success 200 responses


scottkicks:
Since i'll be OOO up until code freeze i'm putting this up here in case the team wants to take over and get this merged in the meantime.
Its 80% of the way there. Just need to test that these calls are being sent. Notes below...

This just needs to be tested via something like Charles Proxy to verify that these calls are being sent at the expected times successfully.
Isa had seen some "200: Bad Request" responses the other day so if we see the same here we can connect with her on that.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -357,36 +357,53 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+                  "",
+                  "",
+                  "",
+                  "",
+                  "",
+                  "",
+                  "",
+                  "",
+                  ""
+                ]
```
scottkicks, line 404: According to the CAPI documentation all values in `extInfo` are required and must be in the given indexed order.
I paired with Isa on this and this format is the only way we can get a successful 200 response for this mutation.

IMO we should only have to pass the used values (i2, OS version) to the backend and _they_ should be responsible for formatting the data correctly. 

Since the docs require this I chose to implement it this way on our end.

Screenshot 2023-07-17 at 2 21 13 PM

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -357,36 +357,53 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
               appData: .init(
                 advertiserTrackingEnabled: true,
                 applicationTrackingEnabled: true,
-                extinfo: ["i2"]
+                extinfo: [
+                  "i2",
+                  "",
+                  "",
+                  "",
+                  "\(ProcessInfo.processInfo.operatingSystemVersion)",
```
scottkicks, line 392: Do we have a helper somewhere already for getting the OS version?

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -198,47 +198,62 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
       if let checkoutDataValues = configData.checkoutData {
-        checkoutUSDAmount = "\(checkoutDataValues.revenueInUsd)" + "\(checkoutDataValues.addOnsMinimumUsd)"
-        checkoutUSDAmount += String(describing: checkoutDataValues.shippingAmountUsd)
-        checkoutUSDAmount += String(describing: checkoutDataValues.bonusAmountInUsd)
+        transactionId = checkoutDataValues.checkoutId
+        shipping = checkoutDataValues.shippingAmountUsd
+        pledgeAmount = Double(truncating: checkoutDataValues.revenueInUsd as NSNumber)
+          + Double(truncating: (checkoutDataValues.bonusAmountInUsd ?? 0) as NSNumber)
+          + checkoutDataValues.addOnsMinimumUsd
       }
```
msadoon, line 217: I didn't verify this - but in case we lose some precision here - this might be more accurate because it doesn't "truncate": 

```
      if let checkoutDataValues = configData.checkoutData {
        transactionId = checkoutDataValues.checkoutId
        shipping = checkoutDataValues.shippingAmountUsd
        pledgeAmount = NSDecimalNumber(decimal: checkoutDataValues.revenueInUsd).doubleValue
          + NSDecimalNumber(decimal: checkoutDataValues.bonusAmountInUsd ?? 0).doubleValue
          + checkoutDataValues.addOnsMinimumUsd
      }
```

msadoon (changes requested):
Ok, this is mostly fine.

I refactored a bit of the parameter passing to keep the code clean.

Normally it would be good to test these network requests wherever they are used in the view model. Oddly we didn't test the CAPI ones. Will enforce this in the eng. plan in the future, so no need to do it here.

The changes are all laid out so make those quickly and I'll approve tomorrow morning. Good work!

scottkicks:
@msadoon I've addressed your comments. thanks for the feedback. I tried your your suggestion for calculating the pledgeAmount and didn't find that it added more accuracy, but i think your solution makes more sense anyway so I've made that change as well.

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -372,36 +372,44 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
           .apiService
-          .triggerCapiEventInput(
+          .triggerThirdPartyEventInput(
             input: .init(
-              projectId: "\(project.id)",
-              eventName: FacebookCAPIEventName.ProjectPageViewed.rawValue,
-              externalId: externalId,
-              userEmail: AppEnvironment.current.currentUserEmail,
+              deviceId: externalId,
+              eventName: ThirdPartyEventInputName.AddNewPaymentMethod.rawValue,
```
msadoon, line 401: This is the wrong event name for this tracking call.

msadoon (changes requested):
There were mistakes here I found, please correct before we can merge.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/ProjectPageViewModel.swift:379:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

msadoon (approved):

PR #1822, opened 2023-05-17
[MBL-736] Upgrade PerimeterX

msadoon:
# 📲 What

We recently saw production issues related to PerimeterX captcha page not appearing when users tried to login.
Upon investigation we found issues with `PBXController` launching on iOS 16.4. This eventually went away, but we are 2 major version back of the dependency, so it's probably a good time to upgrade it.

Going from 1.13.9 to 3.0.3

# 🤔 Why &&  🛠 How

Summary above. For more detail please read [documentation](https://kickstarter.atlassian.net/wiki/spaces/~294637594/pages/2381709313/PerimeterX+Upgrade+on+iOS).

# ✅ Acceptance criteria

- [ ] PerimeterX 3.0.3 is implemented as per SDK instructions, taking into account our custom types and `AppEnvironment` setup.

# ⏰ TODO

- [ ] Ensure we do the required validation as per the SDK instructions. Specific instructions in the document above.

PR #1839, opened 2023-07-26
[MBL-736] Upgrade PerimeterX

msadoon:
# 📲 What

Dependency is 2 major versions back (`1.16.3` -> `3.0.4`) so probably should be upgraded.

# 🤔 Why

We were having login issues related to PerimeterX hopefully this upgrade solves those. We no longer manually launch the PX captcha sheet like in `1.16.3` (see `goToPerimeterXCaptcha` in `AppDelegate` below). `3.0.4` should take care of it automatically. 

This seemed like the issue causing our login button [to not work](https://kickstarter.atlassian.net/browse/SD-2269).

# 🛠 How

Used the simpler PerimeterX upgrade documentation [here](https://docs.perimeterx.com/docs/ios-manual-integration-v3) and the work done on this older, closed [pr](https://github.com/kickstarter/ios-oss/pull/1822).

Was able to reliably to show the captcha sheet on any REST call with these headers: (Solution Architect at PX provided this to us)
```
request.allHTTPHeaderFields = request.allHTTPHeaderFields?.withAllValuesFrom(["x-px-authorization": "4", "x-ks-block": "1"])
```
However, this isn't necessary to do on `xauth` or any other REST call, because PerimeterX has two modes [monitor](https://docs.perimeterx.com/docs/cloudflare-config#monitored-specific-routes) mode and [enforce](https://docs.perimeterx.com/docs/cloudflare-config#enforced-specific-routes) mode.

We can enforce PerimeterX to fully validate each request from the Cloudflare [worker](https://github.com/kickstarter/terraform/blob/main/cloudflare/site/kickstarter/worker-scripts/perimeterx/workerBundle.js).

Suggestion to @kickstarter/infrastructure was to add `xauth/access_token` to `px_enforced_routes` to ensure that route shows captcha (at least on mobile). 

Either way - this boils down to an framework upgrade plain and simple. No extra customization required, let the Captcha sheet show if the SDK decides to.

There's some good cleanup happening as well, with the new SDK instructions which make this implementation much simpler:
Removed:
`PerimeterXBlockResponseType`
`PerimeterXManagerType`
`PXManagerDelegate`
`KsAPINotification`

`PerimeterXClientTests` was also removed, reason being we use `PerimeterXClient` class to wrap the actual SDK, there's no protocol in between to mock. We don't want to be testing the actual SDK.

# ✅ Acceptance criteria

- [x] PerimeterX is updated and correctly configured to send the headers on each REST request.

# ⏰ TODO

- [x] Go through solution with HUMAN Architect and try to identify if there is a way to reliably show the captcha sheet on login.


KsApi/GraphAPI.swift
```diff
@@ -578,182 +578,6 @@ public enum GraphAPI {
     }
```
msadoon, line 578: This file gets re-gened every time `KsAPI` is rebuilt, so can't avoid it being a change.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/AppDelegate.swift:221:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/AppDelegate.swift:224:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

KsApi/ServiceType.swift
```diff
@@ -526,12 +526,7 @@ extension ServiceType {
     headers["X-KICKSTARTER-CLIENT"] = self.serverConfig.apiClientAuth.clientId
     headers["Kickstarter-iOS-App-UUID"] = self.deviceIdentifier
 
-    return headers.withAllValuesFrom(self.pxHeaders)
-  }
-
-  // PerimeterX authorization header
-  internal var pxHeaders: [String: String] {
-    return self.perimeterXClient.headers()
+    return headers
```
msadoon, line 529: We're no longer setting the headers manually, SDK handles that [automatically](https://docs.perimeterx.com/docs/ios-manual-integration-v3). "Understanding the Flow"

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -219,6 +219,12 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
           .messageBannerViewController?.showBanner(with: success ? .success : .error, message: message)
       }
 
+    self.viewModel.outputs.configurePerimeterX
+      .observeValues {
+        AppEnvironment.current.apiService.perimeterXClient
```
msadoon, line 224: This is now where we call `start` on PerimeterX, not in `Service`. Thinking is more along the lines of their official instructions - https://docs.perimeterx.com/docs/ios-how-to-integrate-the-sdk-in-your-application-v3 (Step 5). Although the SDK instructions explicitly say to "You should only call this function once" and `appDidFinishLaunchingWithOptions` is called on every relaunch.
Some ambiguity there.

KsApi/px/PerimeterXClient.swift
```diff
@@ -13,59 +17,39 @@ public class PerimeterXClient: NSObject, PerimeterXClientType {
...
 
-  public func newHeaders(_ httpHeaders: [AnyHashable: Any]!) {
-    print("❎ Perimeter X headers were refreshed: \(String(describing: httpHeaders))")
+  public func handleResponse(data: Data, response: URLResponse) -> Bool {
+    PerimeterX.handleResponse(response: response, data: data) { result in
+      switch result {
+      case .cancelled:
+        print("cancelled")
+      case .solved:
+        print("solved")
```
ifosli, line 49: This doesn't feel submit-ready. Does anything need to happen here? Or, if we just want to log something here, can we do it with more complete messaging?

msadoon: Ah good call out - yes there should not be print statements here. Will remove in my next commit.

msadoon: left them in - previous implementation had some [logs](https://github.com/kickstarter/ios-oss/pull/1839/files#diff-e71c4072e0340759413ae73328b199512c4a430f7ff0ff72e6a863f220296c16L38) that were useful for testing, but removed the force crash `fatalError`.

Kickstarter-iOS/AppDelegate.swift
scottkicks: The instructions also say "It's essential to call this function as early as possible in your application and before any URL request to your server" so this should be fine. I feel like they meant to say "only call this function once _per app launch_" 🤷🏻‍♂️

KsApi/GraphAPI.swift
scottkicks: we might consider putting this in our .gitignore to avoid confusion. We intentionally update this file as needed, when we support a new mutation for example. is there a reason we need to update this file on every build?

scottkicks (approved):
Followed the implementation steps in the docs and this all makes sense to me. App builds and tests pass. Did some general QA as well and everything seems fine 👍

KsApi/extensions/NSURLSession.swift
```diff
@@ -35,7 +35,8 @@ internal extension URLSession {
         guard let response = response as? HTTPURLResponse else { fatalError() }
 
         /// `error` is `nil` or `handleError` returns `false`.
```
ifosli, line 37: Another outdated comment referring to `handleError` - please update. Also, in this comment, can you expand on why we want to process the response when the SDK couldn't handle the response? I would've expected that to be a failure state and the `true` to be a continue state, but I'll admit I don't think the documentation is very clear here. It's also very possible that `true` indicates that the error is fully handled and we don't need to do anything else. Is this something you've been able to test (where the error is non-nil)?

KsApi/px/PerimeterXClient.swift
ifosli: Looks good now, thank you!

ifosli (commented):
I took another pass at this and I think the `handleError` being renamed to `handleResponse` makes some of the comments unclear. I also don't know if the boolean means the same thing now as it did previously, and since I'm not sure how easy that's to test/catch, I wanted to bring it up so you could look into it, if you haven't already.

Kickstarter-iOS/AppDelegate.swift
msadoon: Ah ok makes sense

KsApi/extensions/NSURLSession.swift
msadoon: _What?!_

msadoon: Ok I'm going to delete the inline comment because I don' think we need it. 
In fact - we generally don't write comments in line unless its' `/// TODO:` or `/// FIXME:`
This was just something left in from the previous implementation so I think I'll just remove it.

PR #1844, opened 2023-08-17
[MBL-943] Add Additional Header Support to HTML Parser

msadoon:
# 📲 What

There's been a change on the web to support a new parser called `CKEditor` and it supports additional header tags. Previously we supported only `h1`. Now we support `h1`, `h2`, `h3`, `h4`, `h5`, `h6`.

# 🤔 Why

As stated above, we would eventually need to add new elements/HTML tags supported by more modern project creation tools.

# 🛠 How

Pretty straightforward, taking a look at the Android [pull request](https://github.com/kickstarter/android-oss/pull/1835) that does the exact same thing. Used the font sizes for reference so they should be the same size on both platforms.

Might be nice to screenshot test all the supported HTML components one day...thinking when we remove the parser and get good support for `story` properties on the `ProjectFragment` we can look into this.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [x] Ensure `h1` headers are larger size after the change. Before this change they were size `20`, now they are `28`. Ie. WakeFX Rope-Pal - Automated Waksurf Rope System on staging. (screens above).
- [x] Ensure new headers are supported via sample staging project "CKE Test project for Mobile". Note: Before adding this support `h3` and `h4` would not show. HTML snippet from sample project: `<h3 id=\"h:Header\" class=\"page-anchor\">Header</h3><h4 id=\"h:Subheader\" class=\"page-anchor\">Subheader</h4>`, but now they do. (screens above)

# ⏰ TODO

- [x] Update tests.

### DIFF
--- a/KsApi/lib/HTML Parser/HTMLParserTests.swift
+++ b/KsApi/lib/HTML Parser/HTMLParserTests.swift
@@ -218,27 +218,39 @@ final class HTMLParserTests: XCTestCase {
     XCTAssertEqual(viewElement.embeddedURLContentHeight, 400)
   }
 
-  func testHTMLParser_WithTextHeadline_Success() {
-    let viewElements = self.htmlParser.parse(bodyHtml: HTMLParserTemplates.validHeaderText.data)
+  func testHTMLParser_WithTextHeadlines_Success() {
+    let viewElementsToTextStyleHeaders =
+      [
+        HTMLParserTemplates.validHeader1Text.data: TextComponent.TextStyleType.header1,
+        HTMLParserTemplates.validHeader2Text.data: TextComponent.TextStyleType.header2,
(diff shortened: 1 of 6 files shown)


scottkicks (commented):
LGTM. I see that the headers, `World's First Automated Wakesurf Rope System` and `100% Automatic to the Surfer and Back to Start` are indented unlike the other headers below them.
Could that be an issue with our parser or is that just how the html is initially for this project?

Thanks for linking the android PR. Its good to see that we're maintaining feature parody on with this.

msadoon:
Not sure what - if any - changes are required here.

scottkicks:
@msadoon meant that to be an approval, not a comment. my bad 😅

PR #1850, opened 2023-09-05
[MBL-960] Fix crash by using default hashing and equality for the config

ifosli:
# 📲 What

Update the date formatter to conform to `Hashable` and `Equatable` by default instead of having custom implementations. This has the effect of also relying on the `dateFormat` when comparing two `DateFormatterConfig` objects.

Also add a test that fails without these changes and passes now that the caching is fixed.

# 🤔 Why

Currently, when the `dateFormat` is ignored in comparisons, when we try to parse a different date format, we fetch a cached one expecting a different format, leading to a crash. This happens any time we try to create a formatter for a config that has a different date format but is otherwise identical to a formatter that's already been cached. Specifically, this happens when we show the creator tools deprecation banner and also show credit card expiration dates without killing the app in-between.

# 👀 See

[JIRA ticket](https://kickstarter.atlassian.net/browse/MBL-960)

# ✅ Acceptance criteria

- [ ] Crash no longer repros


Library/Format.swift
```diff
@@ -403,7 +403,7 @@ public enum Format {
 
 public let defaultThresholdInDays = 30 // days
 
-internal struct DateFormatterConfig {
+internal struct DateFormatterConfig: Hashable, Equatable {
```
msadoon, line 406: Cool, yea let's do this - according to the documentation - `Hashable` conformance requires `Equatable`, so you probably only need `Hashable`.

msadoon (approved):
Yea nice clean up.

scottkicks (approved):

Library/Format.swift
ifosli: Cool, I'll delete the explicit Equatable. Thanks!

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/FormatTests.swift:543:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

PR #1853, opened 2023-09-18
[MBL-855] Remove creator dashboard

ifosli:
# 📲 What

Delete all code specific to the creator dashboard now that the creator dashboard is turned off on iOS.

# 🛠 How

This deletes code paths to the creator dashboard, obsolete files, and enums for push notifications that, prior to this change, could only link to the creator dashboard. Note that if the user for some reason were to still receive a notification that we no longer handle, the experience before and after this PR is identical (notification doesn't deep link to anything).

Note: This PR does not try to delete all references to creators or really try to change what server-side information we process. If that's something we want to do, we should do so in a separate PR.

Note: This intentionally does not remove the "Go to dashboard" button on the project page.

# 👀 See

[JIRA](https://kickstarter.atlassian.net/browse/MBL-855)

# ✅ Acceptance criteria

- [x] Logging into a creator account and navigating through the app should be identical to the current (with the creator dashboard turned off via flag) experience

# ⏰ TODO

- [x] Check how these changes affect the dashboard entry points found in [MBL-978](https://kickstarter.atlassian.net/browse/MBL-978) (result: long press shortcut is removed in this pr and button will be removed in a separate pr)

[MBL-978]: https://kickstarter.atlassian.net/browse/MBL-978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/RemoteConfigFeatureFlagToolsViewControllerTests.swift:28:1 | warning: (trailingCommas) Add or remove trailing comma from the last item in a collection literal. |


</p>

Library/ViewModels/RootViewModelTests.swift
ifosli, line 1: Head's up that github's interpretation of added/deleted lines in this file is not great; there's only deleted tests and minor changes to current tests - slight behavior change with the "no creator dashboard" and "memberProjectsCount changing doesn't trigger view controller changes anymore" but most of these lines are just whitespace changes now that we're no longer using remote config in the test.

Library/ShortcutItem.swift
```diff
@@ -1,13 +1,10 @@
 public enum ShortcutItem {
-  case creatorDashboard
```
ifosli, line 2: This gets rid of the long press access to the dashboard bug.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -126,12 +126,6 @@ public protocol AppDelegateViewModelOutputs {
...
   var goToActivity: Signal<(), Never> { get }
 
-  /// Emits when application should navigate to the creator's message thread
-  var goToCreatorMessageThread: Signal<(Param, MessageThread), Never> { get }
-
-  /// Emits when the root view controller should navigate to the creator dashboard.
-  var goToDashboard: Signal<Param?, Never> { get }
-
   /// Emits when the root view controller should navigate to the creator dashboard.
   var goToDiscovery: Signal<DiscoveryParams?, Never> { get }
```
scottkicks, line 130: nit: the comment here is misleading. it should probably say something like 
```suggestion
  /// Emits when the root view controller should navigate to the discovery screen.
  var goToDiscovery: Signal<DiscoveryParams?, Never> { get }
```

Library/ViewModels/RootViewModelTests.swift
scottkicks: The test method `testTabBarItemStyles()`, makes me question whether we need to treat the profile icon differently (backer vs creator) now that we're handling both in essentially the same way. Probably out of scope for this PR, but we could consider updating these tests so that we handle setting the tab bar profile icon the same for all iOS users.

scottkicks (approved):
Nice work! Sorry this took so long. Wow, does it feel good to see so many files cleaned up. 

Left a couple of comments, but everything here looks good other than those entry points your covering in the next PR 👍

Also, Don’t forget to remove the flag from the RemoteConfig [dashboard](https://console.firebase.google.com/u/0/project/kickstarter-ios/config)

Kickstarter-iOS/AppDelegateViewModel.swift
ifosli: Whoops, thanks for catching that!

PR #1859, opened 2023-09-28
[MBL-971] Submit Report Project View

scottkicks:
# 📲 What

New SwiftUI View that displays a form allowing users to report a project.

# 🤔 Why

Our iOS app currently violates the [User Generated Content Policy ](https://developer.apple.com/app-store/review/guidelines/#user-generated-content)in the App Store. Specifically, we need (at the project screen level):

- A method for filtering objectionable content
- A mechanism for users to flag objectionable content
- A mechanism for users to block abusive users

This form should appease Apple

# 🛠 How

- Builds a new SwiftUI view based using [android's implementation](https://kickstarter.slack.com/archives/CG55BBBBL/p1694462280738989?thread_ts=1694459621.003119&cid=CG55BBBBL) as an example.
- Uses our MVVM pattern + Combine to handle state changes and submit the form using our createFlagging graphql mutation
   - We don't have official SwiftUI/Combine standards yet so I mostly followed the existing pattern used in our ChangeEmailView (Also a SwiftUI view)
- The form displays the user's email and projectID for the user's benefit only, and a text box for adding details as to why they are reporting. 
- Displays a Message banner to notify the user when the submission was successful or not
- On successful submission, we auto dismiss these report project views and reload the project page so that it shows the "already reported" label and keeps users from reporting a project more than once.

# 👀 See

Simulator Screen Recording - iPhone 14 - 2023-09-28 at 12 24 09

# ✅ Acceptance criteria

- To view the reports you've submitted log in as an admin and go here: https://staging.kickstarter.com/admin/tasks/.
- Remember to run the app using the AppCenter Beta scheme in order to adjust feature flags.

**Happy Path**
- [x] If the Report Project feature flag is on _and_ the user is logged in, they will see the Report this project option.
- [x] The form displays the logged-in user's email and project URL. This is for the user's benefit and is not editable. 
- [x] Tapping save, submits a report, shows the success message, and then navigates back to the Project page where it will display the new "already reported this project" label
- [x] Users cannot submit a report more than once for the same project at any point

**Sad Path**
- [x] If there is an issue when tapping save, the error message displays but users are given the opportunity to submit again.


ifosli (approved):
I'd like to discuss this more after the cut; I don't think we have time to go into what we want our standards to be right now (I think this should be using Combine instead of reactive swift, for example).

I wasn't able to test this fully (I still don't have an admin password so I can't check if report got sent) but the in-app part looks good. I do think that the "Save" button needs to be called "Send" or "Submit" instead, but I guess that will also need to be a fast follow instead.

So with the caveat that I don't think we have time to explore the best way to do SwiftUI or combine right now, this is good, but I'd love to revisit the decisions made here asap. (I'm not saying the SwiftUI part of it is bad; I am saying it's not following conventions that I'm used to, and I'd need time to dig into why it's structured the way it is to figure out if I agree with it or not. Either way, it's good enough for now.)

PR #1862, opened 2023-10-04
[INF-1630] use m1 resource class

talaris:
# 📲 What

Updates the CircleCI config to use the M1 resources instead of x86.
- Additionally, at the moment Apollo only targets x86 so this also installs Rosetta in CI

# 🤔 Why

CircleCI is retiring the MacOS x86 resource types.

# 👀 See

- [CirleCI m1 resource class](https://circleci.com/docs/configuration-reference/#macos-execution-environment)
- [CirleCI orb for installing Rosetta](https://support.circleci.com/hc/en-us/articles/17402023738651-Installing-Rosetta-on-Apple-Silicon-M1-executors) 

Also note there is a performance improvement, but also a cost increase with moving to M1. The cost is ~2x the x86 cost.

| Before 🐛 | After 🦋 | What |
| --- | --- | --- |
| avg 55 min | 31 min | Total build time |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,7 @@ base_job: &base_job
 
 test_job: &test_job
   steps:
+    - macos/install-rosetta
     - attach_workspace:
         at: .
     - run:
@@ -55,6 +56,7 @@ test_job: &test_job
 
 distribute_job: &distribute_job
   steps:
+    - macos/install-rosetta
     - attach_workspace:
         at: .
     - aws-cli/setup:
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter:
@talaris It looks like there might be an extra commit pulled in here ([cfed005](https://github.com/kickstarter/ios-oss/pull/1862/commits/cfed005218d393397df70d1483469e80a824748e))

talaris:
I believe that was pushed onto the branch to try and fix the apollo errors. I can rebase it out if we don't want it.

amy-at-kickstarter:
@talaris Whoops, you're right! Pardon me 😅

ifosli (approved):
Yeah, can confirm that the graphql changes did fix some warnings (that we thought might be related to the error but weren't). So not strictly necessary in this PR, but definitely still a plus. Thanks for updating this for us!

PR #1866, opened 2023-10-12
[MBL-994] Add Design System Colors From Web

scottkicks:
# 📲 What

Adds colors that the web team currently uses in their design system.

We can use [Android's color mapping](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSColors.kt) as a reference since they've already added these on their side.

# 🤔 Why

Since we don't have a mobile designer, Matthew from Team A is going to dedicate some time to helping us come up with dark-mode designs.

To make this easier for him, and us, we've added a [Design System screen](https://github.com/kickstarter/ios-oss/pull/1849) in our debug menu that acts as a copy/past (styling-wise) of all views/components in our app. We'll use this as a playground where Matthew can see how things are looking on an actual device and it'll give us a reference point for how things should look/behave when we start implementing DM.

We only have the colors that we're currently using on iOS being applied to this screen. This PR adds the rest of what they have on the web. Includes more core colors and defines which should be used for backgrounds, borders, icons, and fonts.

#### Extra context
Currently, in our app, we have a `Colors.swift` file that hard codes color hex values. Nothing there should be impacted by this. 

You'll find that the Design System screen explores a slightly new pattern for applying colors that support their dark mode counterparts. This pattern takes advantage of Xcode asset catalogs since this makes it much easier to apply colors for light/dark/high contrast/etc.

# 🛠 How

1. Move our current color catalogs into a folder titled `colors-old` and all of these new ones into one titled `colors-new`. We can use these folder names as a provided namespace so that we can reference them more directly like this `UIColor(named: "colors-old/white")`
2. Add all of the new colors to the new asset catalog and `DesignSystemCoreColors.swift` for referencing. We don't have any dark mode definitions for these new ones yet.
3. Add enums for font, background, border, and icon colors. These apply the core colors directly so no need to add them in assets.
    - These enums follow the same pattern as core colors. They have a helper method to load the correct UIColor from a specified catalog.  

# ✅ Acceptance criteria

- [x] The Design System screen look/behavior are not impacted
- [ ] Feel free to apply the `adaptive...(...)` helper methods to the components in `DesignSystemViewController.swift` and play around with toggling between Dark and Light mode to see the differences. FYI The new colors don't have dark mode definitions yet.


scottkicks:
There are a lot of file changes because of all of the new colors added

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemBackgroundColors.swift
```diff
@@ -0,0 +1,95 @@
...
+
+  var rawValue: String {
+    switch self {
+    case .backgroundSurfacePrimary:
+      return DesignSystemCoreColors.white.rawValue
+    case .backgroundSurfaceInverse:
+      return DesignSystemCoreColors.grey10.rawValue
+    case .backgroundDisabled:
+      return DesignSystemCoreColors.grey4.rawValue
+    case .backgroundInverse:
```
ifosli, line 38: What does backgroundInverse mean? This is the same as the backgroundSurface color.

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemBackgroundColors.swift
```diff
@@ -0,0 +1,95 @@
...
+      return DesignSystemCoreColors.white.rawValue
+    case .backgroundInverseHover:
```
ifosli, line 40: Do we want to support hover colors on mobile? What situation would we use them in?

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemBackgroundColors.swift
```diff
@@ -0,0 +1,95 @@
...
+      return DesignSystemCoreColors.yellow6.rawValue
+    case .backgroundWarningSubtle:
+      return DesignSystemCoreColors.yellow2.rawValue
+    }
+  }
+}
+
+extension DesignSystemBackgroundColors {
+  public func load(_ colorSet: DesignSystemColorSet) -> UIColor {
+    UIColor(named: "\(colorSet.rawValue)/\(self.rawValue)") ?? .white
```
amy-at-kickstarter, line 88: Might be nice to throw an assert in here if the color is missing (assuming we strip asserts in production).

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemCoreColors.swift
```diff
@@ -79,49 +183,49 @@ public let alertStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
 // MARK: - Buttons
 
 public let adaptiveGreenButtonStyle = baseButtonStyle
-  <> UIButton.lens.titleColor(for: .normal) .~ adaptiveColor(.white)
-  <> UIButton.lens.backgroundColor(for: .normal) .~ adaptiveColor(.create700)
-  <> UIButton.lens.titleColor(for: .highlighted) .~ adaptiveColor(.white)
-  <> UIButton.lens.backgroundColor(for: .highlighted) .~ adaptiveColor(.create700).mixDarker(0.36)
-  <> UIButton.lens.backgroundColor(for: .disabled) .~ adaptiveColor(.create700).mixLighter(0.36)
+  <> UIButton.lens.titleColor(for: .normal) .~ adaptiveColor(.old, .white)
```
amy-at-kickstarter, line 186: If I'm reading this right - for adaptive uses, we're using the old colorset? Will there be a new colorset for adaptive uses?

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemCoreColors.swift
```diff
@@ -42,18 +49,115 @@ public enum DesignSystemColors: String {
...
-public func adaptiveColor(_ style: DesignSystemColors) -> UIColor {
-  style.load()
+public func adaptiveColor(_ colorSet: DesignSystemColorSet, _ style: DesignSystemCoreColors) -> UIColor {
+  style.load(colorSet)
 }
 
+// MARK: Component Styles
+
+/// New component style examples.
+/// Essentially a copy/paste of how we're currently using Prelude to set component styles. These just take advantage of the new adaptiveColor pattern above.
```
amy-at-kickstarter, line 160: A question that is extremely out of scope for this particular PR: what would it look like to move the new design system off of Prelude?

ifosli (changes requested):
Generally looks good, though I have some concerns that are probably out of scope for this PR. Marking this as "Request changes" but for clarity, the only change I think is blocking is marking the correct files as moved vs added (and feel free to push back on that, too, if GitHub doesn't actually give us anything useful between file renames).

I have some high level questions - it's not obvious to me that we want to copy web's colors, especially since things like "black" or "white" get confusing if they're inverted for dark mode. For example, I don't think we need any of the hover colors. Adding purples and stuff is cool, but I think switching out all of our colors is a bigger project than adding dark mode versions to the ones that we currently use, so I think we should be sure we want to go down this path before we commit to it. (Though maybe this is a decision that's already been made that I just don't know about.)

It's also possible that there'll be cases where we want a different dark mode color based on the situation. For example, `textAccentRed` and `backgroundDangerBold` are both `red_06`, so with our current setup, they both map to the same dark mode color. However, it wouldn't surprise me if we'd actually want the text accent to be lighter than the background color in dark mode. We may end up wanting/needing to do the colorsets themselves based on the functional names instead. (Though, since this code is already written, I'm happy to leave this as-is in this PR and revisit later if needed.) I do generally like the design system color names - that'll make the colors nice and readable!

Other than that, I didn't check on the colors-old files (planning to do so, but would prefer them to be correctly marked as moved first).

Finally, also maybe out of scope for this, but if you have time, check out https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Asset-Catalogs. I haven't had time to dig into this, but it looks like asset catalogs will be easier to use going forward if we want to take advantage.

Kickstarter-iOS/Assets.xcassets/colors-new/grey5.colorset/Contents.json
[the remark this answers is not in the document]
scottkicks: I think I see why it's a little confusing in GH. I didn't just move those colors to a `colors-old` folder. I also updated their light and dark color values. I think the best way to review is to look directly in Xcode in this case. I will keep this in mind in the future so files get correctly marked as moved.

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemBackgroundColors.swift
scottkicks: Naming and colors are all taken directly from the web right now. Not all of them apply to our current design system. The intention here is to give us a good starting point (common ground) for working with Designer Matthew when he comes on board. Once dark mode designs are defined, it _should be_ easier to update these existing assets' colors and names to be more mobile-friendly and relevant than doing so from scratch. At least, that's how Android and management decided to approach this. I thought that was fine. 

I can see an argument for only adding color definitions like this once they have been defined by design.

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemBackgroundColors.swift
scottkicks: see my comment above for more context. This will likely change as needed.

scottkicks:
@ifosli I left a comment with this context, but here I'll put it below as well.

Naming and colors are all taken directly from the web right now. Not all of them apply to our current design system. The intention here is to give us a good starting point (common ground) for working with Designer Matthew when he comes on board. Once dark mode designs are defined, it _should be_ easier to update these existing assets' colors and names to be more mobile-friendly and relevant than doing so from scratch. At least, that's how Android and management decided to approach this. I thought that was fine. 

Doing now also gives us a chance to see how the system is being developed and will be referenced. Hopefully, this gives us a head start so that once designs are ready we can get going more quickly. 

Good call out about asset catalogs. I'm working on upgrading to Xcode 15 this week. I think that's only available on that version. I will consider this once I get that done.

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemCoreColors.swift
scottkicks: exactly. I wanted to keep a reference to the current colors that we use (`colors-old`) for now and keep them separate from the new colors being added here (`colors-new`). Those are the names of the color asset folders in Xcode.assets. 

The new colors to have Dark Mode versions yet.

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemCoreColors.swift
scottkicks: It would look like a lot of work. Ideally, that happens as we integrate SwiftUI.

amy-at-kickstarter: If we do ultimately want to move off Prelude, I do think this would be a good time to start thinking about it. (For example, we'd presumably pick a different way of modeling these than using `.template` everywhere.) Like I said, out of scope here, but now feels like as good a time as any to fix it.

One thing I've been thinking about for the SwiftUI transition - we're going to want to fix the app infrastructure before we worry about back-porting individual screens. Design systems are part of that infrastructure, which is why I'm bringing it up now.

PR #1869, opened 2023-10-19
Proof of concept: bridge SwiftUI and Combine in ReportProjectFormView

amy-at-kickstarter:
This is a proof-of-concept exploring how ReactiveSwift might play nicely with Combine and SwiftUI. This would allow us to use more of our existing code, without having to do a lot of boilerplate re-writing from ReactiveSwift to Combine.


Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -16,7 +16,6 @@ struct ReportProjectFormView: View {
   @SwiftUI.Environment(\.dismiss) private var dismiss
   @ObservedObject private var viewModel = ReportProjectFormViewModel()
 
-  @State private var retrievedEmail = ""
```
amy-at-kickstarter, line 19: This state just follows the view model - it should be a binding, instead

PR #1854, opened 2023-09-19
Remove ReactiveSwift from Change Email View [Missing Tests]

msadoon:
# 📲 What

A working change email view replacing ReactiveSwift with Combine.

# 🤔 Why

It's a future proofing of the app and how we can build pages with Apple's SwiftUI and Combine. Third party frameworks are good but aren't as reliable as Apple's frameworks. Also knowing these frameworks are highly transferable and can be applied regardless of iOS project.

# 🛠 How

Functionality is the same as the ReactiveSwift page - except for the one no network bug noted below.

# 👀 See

Identical functionality.

# ✅ Acceptance criteria

- [x] All the functionality is preserved on this page.
- [x] Use should be able to change email, send verification message, see new email when re-entering page.

New bug introduced:
- Sending a correct change email request with "Save" without network connectivity correctly causes error message to show. Then turning network back on and send the same request - the loader doesn’t stop spinning and the request doesn't go through. 

Couldn't test:
- There is a warning message when the email is not deliverable I haven’t tested yet - but it replaces the “The email is unverified” with red text that says “We cannot send the email”. Not sure how to replicate that.

# ⏰ TODO

- [x] Networking
- [x] Refactor business logic
- [x] Validate the unhappy path logic for every network request.
- [ ] Write tests - was locked out of my laptop so you'll have to complete this on your own.

PR #1873, opened 2023-10-26
MBL-1014: Replace ReactiveSwift in ReportProjectFormViewModel with Combine

amy-at-kickstarter:
# 📲 What

This removes ReactiveSwift from `ReportProjectFormViewModel`. As part of that change, I also did some refactoring, to take advantage of as much brevity as I could squeeze out of that change.

# 🤔 Why

This is one of our experiments in moving the app off of ReactiveSwift. The goal here is twofold:

1. Learn more about what that might look like, and how we'd have to architect things
2. Make sure that we can actually ship a Combine + SwiftUI feature without something going terribly wrong

# 👀 See

Because I had to make a lot of architectural choices here, I'm leaving a lot of comments in the code. Take a look and let me know if you agree or disagree! This is all part of figuring out how we want to use Combine in the future.

# ✅ Acceptance criteria

You'll need to turn on the Report Projects feature flag (either in the code directly, or in the app debug interface.) This should have no changes from previous Report Projects behavior, except for a little extra UI when the e-mail is loading.

### DIFF
--- a/Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
+++ b/Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
@@ -7,45 +7,54 @@ enum ReportFormFocusField {
 }
 
 struct ReportProjectFormView: View {
-  let projectID: String
-  let projectURL: String
-  let projectFlaggingKind: GraphAPI.FlaggingKind
   @Binding var popToRoot: Bool
 
   @SwiftUI.Environment(\.dismiss) private var dismiss
-  @ObservedObject private var viewModel = ReportProjectFormViewModel()
+  @ObservedObject private var viewModel: ReportProjectFormViewModel
 
-  @State private var retrievedEmail = ""
-  @State private var details: String = ""
-  @State private var saveEnabled: Bool = false
-  @State private var saveTriggered: Bool = false
   @State private var showLoading: Bool = false
-  @State private var showBannerMessage = false
-  @State private var submitSuccess = false
-  @State private var bannerMessage: MessageBannerViewViewModel?
   @FocusState private var focusField: ReportFormFocusField?
+
+  init(projectID: String,
+       projectURL: String,
+       projectFlaggingKind: GraphAPI.FlaggingKind,
+       popToRoot: Binding<Bool>) {
+    self._popToRoot = popToRoot
+
+    self.viewModel = ReportProjectFormViewModel(
+      projectID: projectID,
+      projectURL: projectURL,
+      projectFlaggingKind: projectFlaggingKind
+    )
+  }
 
   var body: some View {
     GeometryReader { proxy in
       Form {
-        if !retrievedEmail.isEmpty {
-          SwiftUI.Section(Strings.Email()) {
+        SwiftUI.Section(Strings.Email()) {
+          if let retrievedEmail = viewModel.retrievedEmail, !retrievedEmail.isEmpty {
             Text(retrievedEmail)
               .font(Font(UIFont.ksr_body()))
               .foregroundColor(Color(.ksr_support_400))
+              .disabled(true)
+          } else {
+            Text(Strings.Loading())
+              .font(Font(UIFont.ksr_body()))
+              .foregroundColor(Color(.ksr_support_400))
+              .italic()
               .disabled(true)
           }
         }
 
         SwiftUI.Section(Strings.Project_url()) {
-          Text(projectURL)
+          Text(viewModel.projectURL)
             .font(Font(UIFont.ksr_body()))
             .foregroundColor(Color(.ksr_support_400))
             .disabled(true)
         }
 
         SwiftUI.Section {
-          TextEditor(text: $details)
+          TextEditor(text: $viewModel.detailsText)
             .frame(minHeight: 75)
             .font(Font(UIFont.ksr_body()))
             .focused($focusField, equals: .details)
@@ -59,8 +68,8 @@ struct ReportProjectFormView: View {
       .toolbar {
         ToolbarItem(placement: .navigationBarTrailing) {
           LoadingBarButtonItem(
-            saveEnabled: $saveEnabled,
-            saveTriggered: $saveTriggered,
+            saveEnabled: $viewModel.saveButtonEnabled,
+            saveTriggered: $viewModel.saveTriggered,
             showLoading: $showLoading,
             titleText: Strings.Send()
           )
@@ -69,42 +78,21 @@ struct ReportProjectFormView: View {
       .onAppear {
         focusField = .details
         viewModel.inputs.viewDidLoad()
-        viewModel.projectID.send(self.projectID)
-        viewModel.projectFlaggingKind.send(self.projectFlaggingKind)
       }
-      .onChange(of: details) { detailsText in
-        viewModel.detailsText.send(detailsText)
-      }
-      .onChange(of: saveTriggered) { triggered in
-        focusField = nil
-        showLoading = triggered
-        viewModel.saveTriggered.send(triggered)
-      }
-      .onChange(of: bannerMessage) { newValue in
+      .onReceive(viewModel.$bannerMessage) { newValue in
+        showLoading = false
+
         /// bannerMessage is set to nil when its done presenting. When it is done, and submit was successful,  dismiss this view.
-        if newValue == nil, self.submitSuccess {
+        if newValue == nil, viewModel.submitSuccess {
           dismiss()
           popToRoot = true
-        } else if newValue?.bannerBackgroundColor == Color(.ksr_alert) {
-          saveEnabled = true
         }
       }
-      .onReceive(viewModel.saveButtonEnabled) { newValue in
-        saveEnabled = newValue
-      }
-      .onReceive(viewModel.submitSuccess) { _ in
-        submitSuccess = true
-      }
-      .onReceive(viewModel.retrievedEmail) { email in
-        retrievedEmail = email
-      }
-      .onReceive(viewModel.bannerMessage) { newValue in
-        showLoading = false
-        saveEnabled = false
-        bannerMessage = newValue
+      .onReceive(viewModel.$saveTriggered) { triggered in
+        showLoading = triggered
       }
       .overlay(alignment: .bottom) {
-        MessageBannerView(viewModel: $bannerMessage)
+        MessageBannerView(viewModel: $viewModel.bannerMessage)
           .frame(
             minWidth: proxy.size.width,
             idealWidth: proxy.size.width,
--- a/Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
+++ b/Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
@@ -8,25 +8,15 @@ enum ReportFormFocusField {
 
 struct ReportProjectFormView: View {
   @Binding var popToRoot: Bool
+  let projectID: String
+  let projectURL: String
+  let projectFlaggingKind: GraphAPI.FlaggingKind
 
   @SwiftUI.Environment(\.dismiss) private var dismiss
-  @ObservedObject private var viewModel: ReportProjectFormViewModel
+  @StateObject private var viewModel = ReportProjectFormViewModel()
 
   @State private var showLoading: Bool = false
   @FocusState private var focusField: ReportFormFocusField?
-
-  init(projectID: String,
-       projectURL: String,
-       projectFlaggingKind: GraphAPI.FlaggingKind,
-       popToRoot: Binding<Bool>) {
-    self._popToRoot = popToRoot
-
-    self.viewModel = ReportProjectFormViewModel(
-      projectID: projectID,
-      projectURL: projectURL,
-      projectFlaggingKind: projectFlaggingKind
-    )
-  }
 
   var body: some View {
     GeometryReader { proxy in
@@ -47,7 +37,7 @@ struct ReportProjectFormView: View {
         }
 
         SwiftUI.Section(Strings.Project_url()) {
-          Text(viewModel.projectURL)
+          Text(projectURL)
             .font(Font(UIFont.ksr_body()))
             .foregroundColor(Color(.ksr_support_400))
             .disabled(true)
@@ -77,6 +67,10 @@ struct ReportProjectFormView: View {
       }
       .onAppear {
         focusField = .details
+
+        viewModel.projectID = projectID
+        viewModel.projectFlaggingKind = projectFlaggingKind
+
         viewModel.inputs.viewDidLoad()
       }
       .onReceive(viewModel.$bannerMessage) { newValue in
--- a/KsApi/Service.swift
+++ b/KsApi/Service.swift
@@ -1,4 +1,5 @@
 import Apollo
+import Combine
 import Foundation
 import Prelude
 import ReactiveExtensions
@@ -158,6 +159,17 @@ public struct Service: ServiceType {
       .flatMap { _ in
         SignalProducer(value: EmptyResponseEnvelope())
       }
+  }
+
+  public func createFlaggingInputCombine(input: CreateFlaggingInput)
+    -> AnyPublisher<EmptyResponseEnvelope, ErrorEnvelope> {
+    return GraphQL.shared.client
+      .perform(mutation: GraphAPI
+        .CreateFlaggingMutation(input: GraphAPI.CreateFlaggingInput.from(input)))
+      .map { _ in
+        EmptyResponseEnvelope()
+      }
+      .eraseToAnyPublisher()
   }
 
   public func createPassword(input: CreatePasswordInput)
--- a/KsApi/Service.swift
+++ b/KsApi/Service.swift
@@ -363,6 +363,30 @@ public struct Service: ServiceType {
       .flatMap(UserEnvelope<GraphUserEmail>.envelopeProducer(from:))
   }
 
+  public func fetchGraphUserEmailCombine()
+    -> AnyPublisher<UserEnvelope<GraphUserEmail>, ErrorEnvelope> {
+    GraphQL.shared.client
+      .fetch(query: GraphAPI.FetchUserEmailQuery())
+      // TODO: make this a custom extension, we'll want to reuse this pattern
+      .tryMap { (data: GraphAPI.FetchUserEmailQuery.Data) -> UserEnvelope<GraphUserEmail> in
+        guard let envelope = UserEnvelope<GraphUserEmail>.userEnvelope(from: data) else {
+          throw ErrorEnvelope.couldNotParseJSON
+        }
+
+        return envelope
+      }
+      .mapError { rawError in
+
+        if let error = rawError as? ErrorEnvelope {
+          return error
+        }
+
+        return ErrorEnvelope.couldNotParseJSON
+      }
+
+      .eraseToAnyPublisher()
+  }
+
   public func fetchGraphUserSelf()
     -> SignalProducer<UserEnvelope<User>, ErrorEnvelope> {
     return GraphQL.shared.client
--- a/KsApi/combine/CombineTestObserver.swift
+++ b/KsApi/combine/CombineTestObserver.swift
@@ -7,6 +7,7 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
 
   public func observe(_ publisher: any Publisher<Value, Error>) {
     publisher.sink { _ in
+      // TODO(MBL-1017) implement this as part of writing a new test observer for Combine
       fatalError("Errors haven't been handled here yet.")
     } receiveValue: { [weak self] value in
       self?.events.append(value)
--- a/KsApi/combine/CombineTestObserver.swift
+++ b/KsApi/combine/CombineTestObserver.swift
@@ -1,0 +1,16 @@
+import Combine
+import Foundation
+
+public final class CombineTestObserver<Value, Error: Swift.Error> {
+  public private(set) var events: [Value] = []
+  private var subscriptions = Set<AnyCancellable>()
+
+  public func observe(_ publisher: any Publisher<Value, Error>) {
+    publisher.sink { _ in
+      fatalError("Errors haven't been handled here yet.")
+    } receiveValue: { [weak self] value in
+      self?.events.append(value)
+    }
+    .store(in: &self.subscriptions)
+  }
+}
--- a/KsApi/combine/Signal+Combine.swift
+++ b/KsApi/combine/Signal+Combine.swift
@@ -1,0 +1,18 @@
+import Combine
+import Foundation
+import ReactiveSwift
+
+extension Signal where Error == Never {
+  var combinePublisher: AnyPublisher<Value, Never> {
+    let subject = PassthroughSubject<Value, Never>()
+    self.observeValues { value in
+      subject.send(value)
+    }
+
+    return subject.eraseToAnyPublisher()
+  }
+
+  public func assign(toCombine published: inout Published<Value>.Publisher) {
+    self.combinePublisher.assign(to: &published)
+  }
+}
--- a/KsApi/extensions/ApolloClient+RAC.swift
+++ b/KsApi/extensions/ApolloClient+RAC.swift
@@ -1,4 +1,5 @@
 import Apollo
+import Combine
 import Foundation
 import ReactiveSwift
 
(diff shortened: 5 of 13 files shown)


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:6:1 | warning: (blankLinesAtStartOfScope) Remove leading blank line at the start of a scope. |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:24:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:45:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:54:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:61:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/ReportProjectFormViewModelTests.swift:64:1 | warning: (blankLinesAtEndOfScope) Remove trailing blank line at the end of a scope. |


</p>

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
 
 @available(iOS 15.0, *)
 struct ReportProjectFormView: View {
-  let projectID: String
```
amy-at-kickstarter, line 11: I moved these properties into the view model for two reasons:
1) The view model needs `projectId` and `projectFlaggingKind` to submit the report project mutation
2) Just in general, it seems more MVVM-y to me to have the view model own this state, instead of the view

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
...
+  init(projectID: String,
+       projectURL: String,
+       projectFlaggingKind: GraphAPI.FlaggingKind,
+       popToRoot: Binding<Bool>) {
+    self._popToRoot = popToRoot
+
+    self.viewModel = ReportProjectFormViewModel(
+      projectID: projectID,
+      projectURL: projectURL,
+      projectFlaggingKind: projectFlaggingKind
```
amy-at-kickstarter, line 28: Passing these in on init so that `viewModel` has direct access to them. 

One interesting note here - to create a custom init, we needed to include the `Binding<Bool> popToRoot`. The syntax for assigning that binding is weird, hence `self._popToRoot` above.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
...
-        if !retrievedEmail.isEmpty {
-          SwiftUI.Section(Strings.Email()) {
+        SwiftUI.Section(Strings.Email()) {
+          if let retrievedEmail = viewModel.retrievedEmail, !retrievedEmail.isEmpty {
             Text(retrievedEmail)
               .font(Font(UIFont.ksr_body()))
               .foregroundColor(Color(.ksr_support_400))
               .disabled(true)
+          } else {
+            Text(Strings.Loading())
```
amy-at-kickstarter, line 42: Added this because it makes the UI feel a little smoother

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
...
+              .font(Font(UIFont.ksr_body()))
+              .foregroundColor(Color(.ksr_support_400))
+              .italic()
+              .disabled(true)
           }
         }
 
         SwiftUI.Section(Strings.Project_url()) {
-          Text(projectURL)
+          Text(viewModel.projectURL)
```
amy-at-kickstarter, line 51: n.B. that now these are all coming directly from the view model

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
...
             .font(Font(UIFont.ksr_body()))
             .foregroundColor(Color(.ksr_support_400))
             .disabled(true)
         }
 
         SwiftUI.Section {
-          TextEditor(text: $details)
+          TextEditor(text: $viewModel.detailsText)
```
amy-at-kickstarter, line 58: Here's an example of a two-way binding - `TextEditor` can mutate `detailsText`, which will update that property in the viewmodel

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -60,8 +69,8 @@ struct ReportProjectFormView: View {
       .toolbar {
         ToolbarItem(placement: .navigationBarTrailing) {
           LoadingBarButtonItem(
-            saveEnabled: $saveEnabled,
-            saveTriggered: $saveTriggered,
+            saveEnabled: $viewModel.saveButtonEnabled,
+            saveTriggered: $viewModel.saveTriggered,
```
amy-at-kickstarter, line 73: So this is an interesting one - I'm not actually sure that it makes sense for a `triggered` action to be a binding like this. Right now, we're modeling `saveTriggered` as an event stream of `Bools`. For example, if you press the button, you'll get the following stream:
```
false - initial state
true - the button was pressed
false - we set it back to false once the loading is done
```
However, it's not really a _state_, it's more of just a thing that happens. It would make more sense to me to model it as a stream of voids, where only one event occurs when the button is pressed.

I actually like Apple's pattern here with the underlying View class `Button`, where it takes a closure that defines what happens when the button is pressed. This would let us follow more of our old input/output pattern that we used with UIKit:

```
LoadingBarButtonItem(
  saveEnabled: $saveEnabled,
  showLoading: $showLoading,
  titleText: Strings.Send()
) {
  viewModel.didPressSaveButton()
}
```

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -70,42 +79,21 @@ struct ReportProjectFormView: View {
...
-        viewModel.detailsText.send(detailsText)
-      }
-      .onChange(of: saveTriggered) { triggered in
-        focusField = nil
-        showLoading = triggered
-        viewModel.saveTriggered.send(triggered)
       }
-      .onChange(of: bannerMessage) { newValue in
+      .onReceive(viewModel.$bannerMessage) { newValue in
+        showLoading = false
```
amy-at-kickstarter, line 84: It might be cleaner to move the banner logic into the ViewModel, so that the submit, banner display, timeout, and dismiss all happen in the same place. However, that was going to be a bigger piece of refactoring, so I left this as-is.

KsApi/Service.swift
```diff
@@ -351,6 +363,30 @@ public struct Service: ServiceType {
...
+
+        return envelope
+      }
+      .mapError { rawError in
+
+        if let error = rawError as? ErrorEnvelope {
+          return error
+        }
+
+        return ErrorEnvelope.couldNotParseJSON
```
amy-at-kickstarter, line 384: Not clear if there would be a better error to use here. This would be the case when something in userEnvelope(from: data) threw an exception.

KsApi/combine/CombineTestObserver.swift
```diff
@@ -0,0 +1,16 @@
+import Combine
+import Foundation
+
+public final class CombineTestObserver<Value, Error: Swift.Error> {
```
amy-at-kickstarter, line 4: There's another ticket for fleshing this out. This is just the barest-bones stub to get the tests working.

KsApi/combine/Signal+Combine.swift
```diff
@@ -0,0 +1,18 @@
+import Combine
+import Foundation
+import ReactiveSwift
+
+extension Signal where Error == Never {
```
amy-at-kickstarter, line 5: This was an experiment in bridging code that I'm not using any more. Kind of nifty though.

KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -33,6 +34,43 @@ extension ApolloClient: ApolloClientType {
...
   }
 
+  /**
+   Performs a GraphQL fetch request with a given query.
+
+   - parameter query: The `Query` to fetch.
+
+   - returns: A `AnyPublisher` generic over `Query.Data` and `ErrorEnvelope`.
+   */
+  public func fetch<Query: GraphQLQuery>(query: Query) -> AnyPublisher<Query.Data, ErrorEnvelope> {
```
amy-at-kickstarter, line 44: I pulled these changes from Mubarak's PR: https://github.com/kickstarter/ios-oss/pull/1854

KsApi/extensions/ApolloClient+RAC.swift
```diff
@@ -62,4 +100,42 @@ extension ApolloClient: ApolloClientType {
...
+   Performs a GraphQL mutation request with a given mutation.
+
+   - parameter mutation: The `Mutation` to perform.
+
+   - returns: A `AnyPublisher` generic over `Mutation.Data` and `ErrorEnvelope`.
+   */
+  public func perform<Mutation: GraphQLMutation>(
+    mutation: Mutation
+  ) -> AnyPublisher<Mutation.Data, ErrorEnvelope> {
+    let fetchSubject: PassthroughSubject<Mutation.Data, ErrorEnvelope> = .init()
```
amy-at-kickstarter, line 114: One interesting note - `PassthroughSubject` just drops events if nobody is subscribed, and this fires off the networking request immediately. I don't think that will be a problem for these network requests, since presumably they'll be created + subscribed at the same time. But could be an edge case to watch out for.

KsApi/extensions/MockGraphQLClient.swift
```diff
@@ -73,6 +88,18 @@ private func producer<T, E>(for property: Result<T, E>?) -> SignalProducer<T, E>
   }
 }
 
+private func producer<T, E>(for property: Result<T, E>?) -> AnyPublisher<T, E> {
```
amy-at-kickstarter, line 91: This is just copying the old pattern for ReactiveSwift in `MockGraphQLClient`. Keeping it as close as possible.

KsApi/extensions/MockGraphQLClient.swift
```diff
@@ -73,6 +88,18 @@ private func producer<T, E>(for property: Result<T, E>?) -> SignalProducer<T, E>
...
+  switch property {
+  case let .success(data):
+    return CurrentValueSubject(data).eraseToAnyPublisher()
+  case .failure:
+    assertionFailure("Need to implement this behavior. I think the Fail() subject is what we want, possibly with a deferred?")
```
amy-at-kickstarter, line 96: We'll have to implement this when we want to actually test failing a network request.

Library/ViewModels/ReportProjectFormViewModel.swift
```diff
@@ -5,117 +5,108 @@ import ReactiveSwift
...
-    details: String,
-    clientMutationId: String?
-  )
 }
 
-public protocol ReportProjectFormViewModelOutputs {
-  /// Emits the currently logged in user's email
-  var userEmail: Signal<String, Never> { get }
-}
+public protocol ReportProjectFormViewModelOutputs {}
```
amy-at-kickstarter, line 10: So here is an interesting problem: the outputs of our view model are `@Published` properties, and Swift won't let you define a wrapped property like `@Published` in a protocol. That means we can't actually put `userEmail` up here in the outputs, nor any of our other computed properties.

We could potentially define it like this:
```
public protocol ReportProjectFormViewModelOutputs {
   public var retrievedEmailPublisher: Publisher<String?> { get }
   public var retrievedEmail: String?
}
```

but that seems a bit messy. An alternative could also be just making the inputs and outputs a concrete object, instead of a protocol. Or getting rid of them entirely! Thoughts?

Library/ViewModels/ReportProjectFormViewModel.swift
```diff
@@ -5,117 +5,108 @@ import ReactiveSwift
...
+              projectURL: String,
+              projectFlaggingKind: GraphAPI.FlaggingKind) {
+    self.projectID = projectID
+    self.projectURL = projectURL
+    self.projectFlaggingKind = projectFlaggingKind
+
+    /// Only enable the save button if the user has entered detail text
+    self.$detailsText
+      .map { !$0.isEmpty }
+      .assign(to: &$saveButtonEnabled)
```
amy-at-kickstarter, line 45: This is an example of a nice, very brief binding between Combine + SwiftUI. Elegant!

Library/ViewModels/ReportProjectFormViewModel.swift
```diff
@@ -5,117 +5,108 @@ import ReactiveSwift
...
       }
+      .flatMap { $0 }
+      .sink { [weak self] _ in
+        // An error happens
 
-    _ = submitReportEvent.errors()
-      .observeForUI()
-      .observeValues { [weak self] _ in
-        let messageBannerViewViewModel = MessageBannerViewViewModel((
+        self?.saveTriggered = false
```
amy-at-kickstarter, line 72: Some of this logic was in the view; I moved into the view model. These are hard to read in the git diff view so I recommend checking them out in the actual full file view.

Library/ViewModels/ReportProjectFormViewModelTests.swift
```diff
@@ -1,43 +1,65 @@
...
+      projectID: "123", projectURL: "https://www.kickstarter.com/projects/foo/bar",
+      projectFlaggingKind: GraphAPI.FlaggingKind.commentDoxxing
+    )
+
     withEnvironment(apiService: self.userEmailSuccessMockService) {
-      self.vm.inputs.viewDidLoad()
+      let userEmail = CombineTestObserver<String?, Never>()
+      userEmail.observe(vm.$retrievedEmail)
+
+      XCTAssertEqual(userEmail.events.count, 1)
```
amy-at-kickstarter, line 34: Interesting implementation detail: `@Published` properties emit their initial value when a subscriber attaches. So retrieving the e-mail triggers _two_ events - the first containing `nil`, the second containing the actual e-mail.

KsApi/Service.swift
scottkicks: Good question. We could probably use a more verbose error. It looks like the rest of the app currently returns this `couldNotParseJSON` enveloper

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
scottkicks: Agreed. I mentioned in a comment below that we could consider injecting the VM with these properties from the outset instead of having the view be something of a middleman here.

Library/ViewModels/ReportProjectFormViewModel.swift
scottkicks: I could be wrong, but I don't think this input/output protocol pattern is common when using Combine. It seems like more of an RxSwift pattern. I'm happy to explore removing it entirely.  I'm also happy to explore keeping our `@Published` properties more organized in a separate concrete object as you mentioned.

Library/ViewModels/ReportProjectFormViewModel.swift
scottkicks: Oh this is pretty cool how `.assign(to` manages the life cycle of the subscription automatically so we don't have to keep track of an AnyCancellable

Library/ViewModels/ReportProjectFormViewModelTests.swift
```diff
@@ -1,43 +1,65 @@
...
+    XCTAssertEqual(saveButtonEnabled.events.count, 1)
+    XCTAssertEqual(saveButtonEnabled.events.last, false)
+
+    vm.detailsText = "This is my report. I don't like this project very much."
+    XCTAssertEqual(saveButtonEnabled.events.count, 2)
+    XCTAssertEqual(saveButtonEnabled.events.last, true)
+
+    vm.detailsText = ""
+    XCTAssertEqual(saveButtonEnabled.events.count, 3)
+    XCTAssertEqual(saveButtonEnabled.events.last, false)
```
scottkicks, line 63: Since `saveButtonEnabled.events` is an array, we _could_ write these assertions in one line. Def not necessary

```suggestion
    let saveButtonEnabled = CombineTestObserver<Bool, Never>()
    saveButtonEnabled.observe(vm.$saveButtonEnabled)

    XCTAssertEqual(saveButtonEnabled.events, [false])

    vm.detailsText = "This is my report. I don't like this project very much."
    XCTAssertEqual(saveButtonEnabled.events, [false, true])

    vm.detailsText = ""
    XCTAssertEqual(saveButtonEnabled.events, [false, true, false])
```

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,45 +8,54 @@ enum ReportFormFocusField {
...
+  @ObservedObject private var viewModel: ReportProjectFormViewModel
 
-  @State private var retrievedEmail = ""
-  @State private var details: String = ""
-  @State private var saveEnabled: Bool = false
-  @State private var saveTriggered: Bool = false
   @State private var showLoading: Bool = false
-  @State private var showBannerMessage = false
-  @State private var submitSuccess = false
-  @State private var bannerMessage: MessageBannerViewViewModel?
```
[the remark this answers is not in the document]
amy-at-kickstarter: Yeah, I think that's a really interesting question, and I'm not sure I have a great answer for it, either. The way we've structured things here, the viewmodel actually has _no knowledge_ of the view directly, so at  there would be no way for it to even ask the view for its `detailsText`. 

I think `@Binding` kind of raises some interesting questions about ownership in general, since it's really a two-way relationship between the model and the UI. It would be really interesting to try and find some other examples of MVVM in SwiftUI and see what other patterns folks have cooked up.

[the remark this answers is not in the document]
amy-at-kickstarter: At present, my understanding of `@State` is that it's for a variable that is owned by a view, and/or used by its subviews. At first glance to me, it seems as if we don't have much in the way of state that is ever truly owned by a view - unless it was something like, say, a tab collapsing/expanding.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
amy-at-kickstarter: I like that pattern - avoiding the slightly odd custom `init` seems like it would save us some effort.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
amy-at-kickstarter: I could see us passing state back to the viewmodel with an input method like `didSubmitReport(withDetailsText detailsText: String)`. But this also raises another question, which is whether or not it would work with combine and `buttonEnabled`.

Based on what I understand right now - a `@State` variable on a `View` is not a publisher. Right now we're doing this in the viewmodel:
```
    self.$detailsText
      .map { !$0.isEmpty }
      .assign(to: &$saveButtonEnabled)
```
but if `detailsText` is a `@State` and not `@Published`, this pattern doesn't work.

Playing with how this is organized seems like a really good thing to mess around with during a pairing session, it'd be interesting to see what you mean and figure out how else it might work.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
amy-at-kickstarter: Yeah I think that'd be good too

Library/ViewModels/ReportProjectFormViewModel.swift
amy-at-kickstarter: Ah, interesting, so maybe there's no good reason to keep it. I'm open to either. It feels a little verbose to me, but it is nice in terms of auto-complete.

Library/ViewModels/ReportProjectFormViewModel.swift
amy-at-kickstarter: Yeah, I really dig this pattern

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
```diff
@@ -8,33 +8,32 @@ enum ReportFormFocusField {
...
 struct ReportProjectFormView: View {
+  @Binding var popToRoot: Bool
   let projectID: String
   let projectURL: String
   let projectFlaggingKind: GraphAPI.FlaggingKind
-  @Binding var popToRoot: Bool
 
   @SwiftUI.Environment(\.dismiss) private var dismiss
-  @ObservedObject private var viewModel = ReportProjectFormViewModel()
+  @StateObject private var viewModel = ReportProjectFormViewModel()
```
amy-at-kickstarter, line 17: @scottkicks OK, just learned something interesting. In a truly react-like way, `ReportProjectFormView` may be initialized many times - any time something calling it in a closure re-renders, for example. That means that `ReportProjectFormViewModel` was being initialized multiple times, which is a performance drag (and in the case of some re-renders, could also potentially mean we don't have the correct instance of the viewmodel). Apparently the solution is that the `View` should own the `ViewModel` using the `@StateObject` wrapper, which ensures that the view model is only instantiated once for the entire lifecycle of our view.

amy-at-kickstarter: This also means we can't pass in state directly on init of the viewmodel, or pass the view model directly into the view...so I set some properties on the viewmodel in `onAppear`.

Library/ViewModels/ReportProjectFormViewModel.swift
ifosli: I vote we get rid of the inputs and outputs. I think with swiftui/combine bindings (where UI and view model stay in sync almost for us) it's not that useful to distinguish between the inputs and the outputs anymore.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
ifosli: I like the closure approach better, too!

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
ifosli: +1. Creating the view model first and then passing it to the view is also nice for testability. I'm not sure if it matters as much in our codebase, since we typically use mocks for testing, but if we ever want to create a fake of an entire view model, we can use that to make simple snapshot tests.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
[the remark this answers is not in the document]
ifosli: I agree with this - the default should be for the viewmodel to own things (unless it's literally something that only affects the UI). Any time logic can live in the viewmodel instead of the view, it should.

KsApi/extensions/MockGraphQLClient.swift
ifosli: Nit: maybe add a `TODO` here, so we don't lose track of it? I expect we'll want to to be able to test failing network requests, especially if we want to improve how we handle them, so doing that as part of a proof-of-concept sounds like a good idea.

ifosli (approved):
Nice job! I like this a lot! One thing that's not really about this PR at all: it would be nice to link JIRA tickets/add jira ticket numbers to the TODOs. I'm not sure if that's something we actually currently do as a team, but it's a nice way to actually keep track of TODOs and their context (and it's nice to be able to search for the ticket number to find the relevant code if you pick up a backlog bug).

Where the view model is initialized would also be nice to update in this PR, but I like it otherwise (with some expected followup work that you already called out).

amy-at-kickstarter:
> Nice job! I like this a lot! One thing that's not really about this PR at all: it would be nice to link JIRA tickets/add jira ticket numbers to the TODOs. I'm not sure if that's something we actually currently do as a team, but it's a nice way to actually keep track of TODOs and their context (and it's nice to be able to search for the ticket number to find the relevant code if you pick up a backlog bug).
> 
> Where the view model is initialized would also be nice to update in this PR, but I like it otherwise (with some expected followup work that you already called out).

Good point about the `//TODOs`, I'll see about setting up some tickets for those.

Kickstarter-iOS/Features/ReportProject/ReportProjectFormView.swift
amy-at-kickstarter: Cool, I'll throw myself a quick ticket to refactor this. Just to keep this PR a bit more svelte.

KsApi/extensions/MockGraphQLClient.swift
amy-at-kickstarter: Will do

Library/ViewModels/ReportProjectFormViewModel.swift
amy-at-kickstarter: Fair point!

PR #1874, opened 2023-10-30
[MBL-1021] Block User Action Sheet

scottkicks:
# 📲 What

Presents an action sheet using SwiftUI when tapping a creator/backer profile icon or username from the project page, messages, or comments. Providing users with an option to block a creator or backer. Tapping this block action will present the popup confirmation, to be done in the next PR, but for now does nothing. 

# 🤔 Why

Our goal is to implement a user-blocking feature that’s good enough for Apple to accept our app store submission and that’ll provide our users with user blocking functionality. [See here](https://docs.google.com/document/d/1NcccTogym3m8kyr9Ez-PWV3v04FFuTOFCInK54T3yE4/edit#heading=h.tgo967rfn82kConnect) for more details on specific product requirements.

# 🛠 How

Reviewing commit by commit might be helpful

This feature is guarded behind the Block User feature flag and users must be logged in to block users. 

- Built a SwiftUI view that presents an action sheet with a "Block User" option, and a "View Profile" button that will only show if we're presenting this on a project page.
- Adds necessary protocol/delegate relationships to handle tap gestures from comments and messages
- Makes sure to pass the User that is being tapped around so that when we present the Popup confirmation, we can correctly send user data to our backend to be blocked. 

### **important**
The translations team will need screenshots of the new copy being used in the app to have some context for translating. Since we don’t have any designs, we’ll need to do this once the entire UI piece is done. To avoid merging incomplete code into `main` I created a `scott/block-users-ui` feature branch that everything will go in to first. 

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2023-10-30 at 12 50 37

# ✅ Acceptance criteria

_This new action sheet will only be presented if you are logged in and the Block User feature flag is enabled._

- [x] Tapping a creator, or backer, icon/name from a comment, comment reply, or from within a message presents an action sheet with "Block User" as an option.
- [x] Tapping a creator, or backer, icon/name from a project page presents an action sheet with "Block User" and "View Profile" options. 
- [x] Tapping "Block User' in the action sheet currently does nothing. In the next PR, this will present a blocking confirmation alert.


Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +200,33 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
...
+    if #available(iOS 15.0, *) {
+      if let sheet = self.view.filter({ $0.tag == actionSheetTag }) {
+        sheet.removeFromSuperview()
+      }
+
+      let actionSheet =
+        UIHostingController(rootView: BlockUserActionSheetView(
+          blockUser: { self.blockUser() }
+      ))
+      actionSheet.view.tag = actionSheetTag
```
scottkicks, line 218: The action sheet is only being added as a subview as needed. I'm giving it a tag so that we can avoid adding a new instance on every tap. See `line: 210`

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
```diff
@@ -0,0 +1,25 @@
...
+    // Scott TODO: Use localized strings once translations are done
+    EmptyView()
+      .confirmationDialog("", isPresented: .constant(true), titleVisibility: .hidden) {
+        if let viewProfile = viewProfile {
+          Button("View profile") {
+            viewProfile()
+          }
+        }
+
+        Button(role: .destructive) {
```
ifosli, line 18: nit: We should update the styling to be our red color and also a kickstarter font. But that can happen in a separate PR.

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
```diff
@@ -0,0 +1,25 @@
...
+  var blockUser: () -> Void
+  var viewProfile: (() -> Void)?
+
+  var body: some View {
+    // Scott TODO: Use localized strings once translations are done
+    EmptyView()
+      .confirmationDialog("", isPresented: .constant(true), titleVisibility: .hidden) {
+        if let viewProfile = viewProfile {
+          Button("View profile") {
+            viewProfile()
```
ifosli, line 14: No need to use a trailing closure just to call one function; I think `Button("View profile", action: viewProfile)` would be more readable. This is true for the `blockUser` below, too.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +200,33 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
   }
 }
 
+// MARK: - CommentCellDelegate
+
+extension CommentRepliesViewController: CommentCellDelegate {
+  func commentCellDidTapHeader(author _: Comment.Author) {
+    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else { return }
+
+    if #available(iOS 15.0, *) {
```
ifosli, line 209: Is the plan to only make blocking available for iOS 15+? If I remember correctly, we were planning to drop iOS 14 anyways because iOS 17 is out now, so I think we should just drop iOS 14 entirely instead of adding these if available tags.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +200,33 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
...
+      if let sheet = self.view.filter({ $0.tag == actionSheetTag }) {
+        sheet.removeFromSuperview()
+      }
+
+      let actionSheet =
+        UIHostingController(rootView: BlockUserActionSheetView(
+          blockUser: { self.blockUser() }
```
ifosli, line 216: nit: `self.blockUser` instead of closure + calling function

ifosli (changes requested):
Looks pretty good! I left some comments on a few minor things. I think we should fix styling in a separate pr and also fix accessibility in a few different separate prs. The usernames currently aren't clickable using voiceover in messages or comments, but "view replies" isn't clickable in comments either, so I think it's probably out of scope for this pr.

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
scottkicks: Nice. This works for view profile, but I need to control the foregroundColor of the block user label so the syntax for that will need to be something like
```
Button {
          blockUser()
        } label: {
          Text("Block this user")
            .foregroundColor(Color(.ksr_alert))
        }
```

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
scottkicks: Normally yes, but with this SwiftUI view, we need this because `BlockUserActionSheetView` expects `blockUser` to be of type `() -> Void` rather than `()`, which is what self.blockUser is.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
scottkicks: That's right. I think we just haven't made the effort to prioritize that work with all of the release blockers. That could be something we can tackle separately while we wait on the backend though.

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
```diff
@@ -0,0 +1,25 @@
+import SwiftUI
+
+@available(iOS 15.0, *)
+struct BlockUserActionSheetView: View {
+  var blockUser: () -> Void
```
amy-at-kickstarter, line 5: Any reason this couldn't be a `let`?

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -179,6 +186,10 @@ extension CommentRepliesViewController {
       self.viewModel.inputs.didSelectComment(comment)
     }
   }
+
+  override func tableView(_: UITableView, willDisplay cell: UITableViewCell, forRowAt _: IndexPath) {
```
amy-at-kickstarter, line 190: Huh, I'm guessing this is old code, but I'm surprised it's setting up a cell delegate instead of just implementing `tableView:didSelectCell:atIndexPath`

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
amy-at-kickstarter: I have the bandwidth to work on dropping 14 this week, if that'd be helpful.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +200,33 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
...
 
+// MARK: - CommentCellDelegate
+
+extension CommentRepliesViewController: CommentCellDelegate {
+  func commentCellDidTapHeader(author _: Comment.Author) {
+    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else { return }
+
+    if #available(iOS 15.0, *) {
+      if let sheet = self.view.filter({ $0.tag == actionSheetTag }) {
+        sheet.removeFromSuperview()
```
amy-at-kickstarter, line 211: Is this to prevent two sheets from being shown at the same time? If so, I would expect that we would track that state in another way, instead of in tags/the view hierarchy.

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        Button(role: .destructive) {
+          blockUser()
+        } label: {
+          Text("Block this user")
+        }
+      }
+  }
+}
```
amy-at-kickstarter, line 25: Question: Do we need a custom action sheet here, or could we just use a `UIActionSheet`? That'd cut down the amount of code you need pretty substantially.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +200,33 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
...
+      }
+
+      let actionSheet =
+        UIHostingController(rootView: BlockUserActionSheetView(
+          blockUser: { self.blockUser() }
+      ))
+      actionSheet.view.tag = actionSheetTag
+
+      _ = (actionSheet.view, self.view)
+        |> ksr_addSubviewToParent()
```
amy-at-kickstarter, line 221: Question: Why add the view as a child view, instead of just presenting the view controller as a modal?

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
amy-at-kickstarter: See my comments on using `present` and `UIAlertController` instead - that would be the pattern I would recommend here, unless there's a constraint I'm missing in the designs that makes a custom approach necessary.

scottkicks: Yeah, that was my goal here. I didn't see the view being removed from the view hierarchy when dismissing it and wanted to make sure it wasn't being left behind. How would you recommend tracking that state instead?

I'll address your `present` note in your other comment 😄

Kickstarter-iOS/Features/BlockingUsers/BlockUserActionSheetView.swift
scottkicks: We don't need it to be custom, no. I thought it would be more fun to build in SwiftUI here. It's a very simple view and I don't think it adds overhead. 
We definitely could present an action sheet using UIAlertController instead. I'm happy to re-work it if you think there's a strong argument. I can see one for maintaining existing patterns, but is there a performance gain here?

scottkicks:
@ifosli @amy-at-kickstarter Replaced the custom SwiftUI action sheet with `UIAlertController` by adding a helper `blockUserActionSheet()` to our UIAlertController extensions. This is a cleaner and simpler approach. 

Also, @amy-at-kickstarter I'm seeing CI failures that might be related to the recent changes you've made. Could you take a look and see if that's the case?

Library/UIAlertController.swift
```diff
@@ -356,4 +356,37 @@ public extension UIAlertController {
...
+
+    if AppEnvironment.current.device.userInterfaceIdiom == .pad {
+      alertController.modalPresentationStyle = .popover
+      alertController.popoverPresentationController?.sourceView = sourceView
+    }
+
+    // Accessibility
+    // Scott TODO: Use localized strings once translations can be done [mbl-1037](https://kickstarter.atlassian.net/browse/MBL-1037)
+    alertController.view.accessibilityIdentifier = "block_user_action_sheet"
+    alertController.view.accessibilityValue = "User Menu"
```
ifosli, line 388: Soo, I don't think adding extra accessibility stuff here actually helped the way I wanted it to. There's still a sound and an "Alert" announced when the model is presented (and voiceover is active). I didn't see an easy way to turn it off, unfortunately. I was hoping that the "Alert" was only read out as a default title, but it's read out before the title if it's provided. I think we can just get rid of the "User menu" title, unless you think we'll have time to look for a fix here.

Library/UIAlertController.swift
```diff
@@ -356,4 +356,37 @@ public extension UIAlertController {
 
     return alertController
   }
+
+  static func blockUserActionSheet(
+    viewProfileHandler: ((UIAlertAction) -> Void)? = nil,
+    blockUserHandler: ((UIAlertAction) -> Void)? = nil,
```
ifosli, line 362: nit: I don't think this should be optional

ifosli (changes requested):
A couple minor things + voiceover is not ideal. But the good news is it's not more broken, either. I have a hunch that the only way to fix it would be to not use an alert controller at all (and maybe make our own custom UI instead), which I don't recommend right now. Using the alert controller isn't a bad voiceover experience, just a weird one.

Library/UIAlertController.swift
scottkicks: AlertControllers have voice-over support by default, which is the announced "alert" you're referring to. I can remove it for now since this doesn't seem to be replacing it.

.circleci/config.yml
```diff
@@ -99,7 +99,9 @@ jobs:
       - *default_environment
     steps:
       - macos/install-rosetta
-      - git-shallow-clone/checkout
+      - git-shallow-clone/checkout:
+          # This is for Danger, which requires a bit more git history
+          fetch_depth: 100
```
amy-at-kickstarter, line 104: Just merged this into `main`, so you might want to rebase

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -9,6 +9,8 @@ internal final class MessagesViewController: UITableViewController {
   fileprivate let viewModel: MessagesViewModelType = MessagesViewModel()
   fileprivate let dataSource = MessagesDataSource()
 
+  private let actionSheetTag: Int = 1
```
amy-at-kickstarter, line 12: Nit, this is no longer needed

scottkicks: ah shoot

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -189,6 +197,27 @@ extension CommentRepliesViewController: CommentComposerViewDelegate {
...
+        sourceView: cell,
+        isIPad: self.traitCollection.horizontalSizeClass == .regular
+      )
+
+    self.present(actionSheet, animated: true)
+  }
+
+  func commentCellDidTapReply(_: CommentCell, comment _: Comment) {}
+
+  func commentCellDidTapViewReplies(_: CommentCell, comment _: Comment) {}
```
amy-at-kickstarter, line 218: If implementing these everywhere is annoying, you could provide a default empty implementation as an extension next to the protocol. (I don't feel strongly about this, though.)

Kickstarter-iOS/Features/Comments/Views/Cells/CommentCell.swift
```diff
@@ -44,6 +45,10 @@ final class CommentCell: UITableViewCell, ValueCell {
     self.bindViewModel()
 
     self.replyButton.addTarget(self, action: #selector(self.replyButtonTapped), for: .touchUpInside)
+    self.commentCellHeaderStackView.addGestureRecognizer(UITapGestureRecognizer(
```
amy-at-kickstarter, line 48: I wonder if a long press would be a better interaction? But that's really gotta be a question for a new designer...

scottkicks: I considered this, but since users are already used to tapping a creator's icon on project pages as a regular tap I chose to go with what's familiar.

amy-at-kickstarter (approved):
LGTM, with a couple non-blocking suggestions.

ifosli (commented):
I realized one more thing- the `RootCommentCell` should be updated, too. Right now, if you tap view replies, all usernames are tappable except the root comment. (I'm not sure why that's its own class, but I don't think that part is worth fixing right now.)

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/RootCommentCellViewModelTests.swift:92:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

ifosli (approved):
Looks good, thanks!

PR #1878, opened 2023-11-01
Set target deployment to iOS 15.0

amy-at-kickstarter:
# 📲 What

This drops support for iOS 14 and makes our minimum target iOS 15.

# 🤔 Why

We only want to support three iOS versions at a time - so now we have 15, 16 and 17.

# ✅ Acceptance criteria

- [x] Tests pass

*Regression Tests*
- [x] Apple sign-in button on Login page is unchanged
- [x] Still able to go to Login/Signup from `PledgeViewController`
- [x] Still able to report a project 
- [x] Still able to use Change Email view from Settings page
- [x] New password/reset password styles are unchanged

### DIFF
--- a/Library/TestHelpers/TestCase.swift
+++ b/Library/TestHelpers/TestCase.swift
@@ -91,7 +91,7 @@ internal class TestCase: XCTestCase {
     let supportedModels = ["iPhone10,1", "iPhone10,4"] // iPhone 8
     let modelKey = "SIMULATOR_MODEL_IDENTIFIER"
 
-    guard #available(iOS 15, *), supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
+    guard supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
       fatalError("Please only test and record screenshots on an iPhone 8 simulator running iOS 15.5")
     }
   }


Kickstarter-iOS/Features/SettingsAccount/Controller/SettingsAccountViewController.swift
```diff
@@ -189,13 +189,8 @@ extension SettingsAccountViewController {
...
     case .changeEmail:
-      if #available(iOS 15, *) {
-        let changeEmailView = ChangeEmailView()
-
-        return UIHostingController(rootView: changeEmailView)
-      } else {
-        return ChangeEmailViewController.instantiate()
-      }
+      let changeEmailView = ChangeEmailView()
+      return UIHostingController(rootView: changeEmailView)
```
amy-at-kickstarter, line 193: Do we want to clean up/delete the old ChangeEmailViewController, or leave it until SwiftUI is no longer 'experimental'? The new ChangeEmailView is definitely missing some tests, for instance, but it's not like this was behind a feature flag...

Library/TestHelpers/TestCase.swift
```diff
@@ -91,8 +91,8 @@ internal class TestCase: XCTestCase {
     let supportedModels = ["iPhone10,1", "iPhone10,4"] // iPhone 8
     let modelKey = "SIMULATOR_MODEL_IDENTIFIER"
 
-    guard #available(iOS 14.5, *), supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
-      fatalError("Please only test and record screenshots on an iPhone 8 simulator running iOS 14.5")
+    guard #available(iOS 15, *), supportedModels.contains(ProcessInfo().environment[modelKey] ?? "") else {
+      fatalError("Please only test and record screenshots on an iPhone 8 simulator running iOS 15.5")
```
amy-at-kickstarter, line 95: Note that CircleCI was already testing against 15.5, so no updates to snapshots are needed

scottkicks (approved):
LGTM 😄 

There's actually one more `#available(iOS 15, *)` in `TestCase.swift` line: 94 that we could clean up

PR #1879, opened 2023-11-02
[MBL-1021] Fix user blocking voiceOver

ifosli:
# 📲 What

Fix accessibility for comments and messages by splitting the cells into accessibility components. The header/username is annotated as a button when the feature flag is enabled. The accessibility hint for the username button on the project page is removed.

Ideally we should add a "Opens user menu" accessibility hint instead, but that will need to wait until we have such a string translated.

# ♿️ Accessibility 

- [x] Works with VoiceOver

# ✅ Acceptance criteria

- [x] VoiceOver works as expected with feature flag on or off

### DIFF
--- a/Kickstarter-iOS/Features/Comments/Views/Cells/CommentCell.swift
+++ b/Kickstarter-iOS/Features/Comments/Views/Cells/CommentCell.swift
@@ -49,6 +49,8 @@ final class CommentCell: UITableViewCell, ValueCell {
       target: self,
       action: #selector(self.commentCellHeaderTapped)
     ))
+
+    self.configureAccessibilityElements()
   }
 
   required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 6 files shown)


scottkicks (approved):
Nice! 🥳

amy-at-kickstarter (approved):
LGTM (just 👀 , didn't test). Out of curiosity, is this a pattern we'll need to duplicate for future cells, or a specific bug we're working around? (I haven't done a ton of work with voiceover, but I'm used to it being 'good enough' and not needing as much customization.)

ifosli:
> LGTM (just 👀 , didn't test). Out of curiosity, is this a pattern we'll need to duplicate for future cells, or a specific bug we're working around? (I haven't done a ton of work with voiceover, but I'm used to it being 'good enough' and not needing as much customization.)

I think it depends on what kind of behavior we want. In general, it's probably okay to read out a cell as one object (where that's the default), as long as we don't expect the user to interact with different parts of the cell independently. But any time we want the user to be able to click on specific things, especially if we're not using UIButton directly, we should do something like this.

amy-at-kickstarter:
> I think it depends on what kind of behavior we want. In general, it's probably okay to read out a cell as one object (where that's the default), as long as we don't expect the user to interact with different parts of the cell independently. But any time we want the user to be able to click on specific things, especially if we're not using UIButton directly, we should do something like this.

Makes sense. Thanks!

PR #1880, opened 2023-11-06
[MBL-1036] Block User Popup + Banner Message

scottkicks:
# 📲 What

1. Presents a confirmation alert when "Block this user" is selected from the block user action sheet
2. Presents a success or fail Message Banner when tapping block in this new alert. This isn't wired up to an API or anything yet. This behavior is hardcoded for now.

# 🤔 Why

Rather than sending a blocking request from the action sheet, we're going to use this confirmation alert pattern so that we can provide extra information and double-check that they want to block a user. 

# 🛠 How

The key areas are:
- Comments
- CommentReplies
- ProjectPage
- Messages

I'm taking advantage of the existing MessageBannerViewController that we use elsewhere in the app to display API response success and error banners. This means that each ViewController, except ProjectPage, needs to conform to `MessageBannerViewControllerPresenting` and present an instance of the banner when the corresponding ViewModel outputs a response from the API (which isn't wired up yet).

One caveat is that to keep users from spamming the block action sheet and alert, I'm disabling user interaction on the parent view until the success/error banner message is finished displaying. Once the success banner is dismissed on the project page, we auto-dismiss the view since the user doesn't want to see that content anymore. This could also set us up to more easily refresh our app content.
   - we won't dismiss when blocking from comments since users could block a user that doesn't own that project. 

So basically I've:
1. added functionality to display a new UIAlert when tapping the action sheet
2. when tapping Block on this new alert, we trigger an input to the ViewModel (this will eventually kick off our backend request) and then a new `userBlocked` output notifies the ViewController of which banner to display. Success or error.

# 👀 See

| 🦋 |
| --- |
Simulator Screen Recording - iPhone 15 Pro - 2023-11-07 at 10 44 02

# ✅ Acceptance criteria

_**Run the AppCenter Beta build. Make sure you're logged in and that the Block User FF is on.**_

- [x] Tapping the Block this user action sheet, from any of the key areas listed above, presents a UIAlert with an option to block or cancel
- [x] Tapping block user presents a banner message at the bottom of the screen. (this is temporarily controlled by a manual bool in the view model until we have an API to wire this up to)
- [x] If showing the success banner, the parent view is not tappable and dismisses once the banner hides
- [x] If showing the error banner, the parent view does not dismiss and users can try to again to block.

### DIFF
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -75,7 +75,6 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
     self.navigationItem.title = Strings.Replies()
 
     self.messageBannerViewController = self.configureMessageBannerViewController(on: self)
-    self.messageBannerViewController?.delegate = self
     self.tableView.dataSource = self.dataSource
     self.tableView.delegate = self
     self.tableView.registerCellClass(CommentCell.self)
@@ -249,15 +248,6 @@ extension CommentRepliesViewController: CommentCellDelegate {
 extension CommentRepliesViewController: RootCommentCellDelegate {
   func commentCellDidTapHeader(_ cell: RootCommentCell, _ author: Comment.Author) {
     self.handleCommentCellHeaderTapped(in: cell, author)
-  }
-}
-
-// MARK: - MessageBannerViewControllerDelegate
-
-extension CommentRepliesViewController: MessageBannerViewControllerDelegate {
-  func messageBannerViewDidHide(type _: MessageBannerType) {
-    self.commentComposer.isHidden = false
-    self.view.isUserInteractionEnabled = true
   }
 }
 
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -51,12 +51,6 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
     super.bindStyles()
 
     _ = self.tableView |> tableViewStyle
-
-    if let banner = messageBannerViewController?.view {
-      NSLayoutConstraint.activate([
-        banner.widthAnchor.constraint(equalTo: self.view.widthAnchor)
-      ])
-    }
   }
 
   // MARK: Lifecycle
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -51,6 +51,12 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
     super.bindStyles()
 
     _ = self.tableView |> tableViewStyle
+
+    if let banner = messageBannerViewController?.view {
+      NSLayoutConstraint.activate([
+        banner.widthAnchor.constraint(equalTo: self.view.widthAnchor)
+      ])
+    }
   }
 
   // MARK: Lifecycle
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -161,16 +161,22 @@ final class CommentRepliesViewController: UITableViewController {
       }
   }
 
-  private func blockUser() {
-    // Scott TODO: present popup UI [mbl-1036](https://kickstarter.atlassian.net/browse/MBL-1036)
+  private func presentBlockUserAlert(username: String) {
+    let alert = UIAlertController
+      .blockUserAlert(username: username, blockUserHandler: { _ in self.blockUser() })
+    self.present(alert, animated: true)
   }
 
-  private func handleCommentCellHeaderTapped(in cell: UITableViewCell, _: Comment.Author) {
+  private func blockUser() {
+    // Scott TODO: call viewModel.inputs.blockUser
+  }
+
+  private func handleCommentCellHeaderTapped(in cell: UITableViewCell, _ author: Comment.Author) {
     guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else { return }
 
     let actionSheet = UIAlertController
       .blockUserActionSheet(
-        blockUserHandler: { _ in self.blockUser() },
+        blockUserHandler: { _ in self.presentBlockUserAlert(username: author.name) },
         sourceView: cell,
         isIPad: self.traitCollection.horizontalSizeClass == .regular
       )
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -178,6 +178,7 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
         if success {
           self?.messageBannerViewController?
             .showBanner(with: .success, message: "This user has been successfully blocked")
+          self?.view.isUserInteractionEnabled = false
         } else {
           self?.messageBannerViewController?
             .showBanner(with: .error, message: "Your request did not go through. Try again.")
@@ -262,6 +263,7 @@ extension CommentRepliesViewController: RootCommentCellDelegate {
 extension CommentRepliesViewController: MessageBannerViewControllerDelegate {
   func messageBannerViewDidHide(type _: MessageBannerType) {
     self.commentComposer.isHidden = false
+    self.view.isUserInteractionEnabled = true
   }
 }
 
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -10,11 +10,13 @@ private enum Layout {
   }
 }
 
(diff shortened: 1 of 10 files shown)


Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -9,14 +9,22 @@ public protocol MessageBannerViewControllerPresenting {
     -> MessageBannerViewController?
 }
 
+public protocol MessageBannerViewControllerDelegate: AnyObject {
+  func messageBannerViewDidHide(type: MessageBannerType)
```
amy-at-kickstarter, line 13: Nit: it is more canonical to pass the calling object itself back in the delegate method. So the method would look like `messageBannerViewDidHide(view: MessageBannerViewController)` and then `bannerType` could be a `public private(set)`

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -39,6 +43,12 @@ internal final class MessagesViewController: UITableViewController {
 
     _ = self.replyBarButtonItem
       |> UIBarButtonItem.lens.title %~ { _ in Strings.general_navigation_buttons_reply() }
+
+    if let banner = messageBannerViewController?.view {
+      NSLayoutConstraint.activate([
+        banner.widthAnchor.constraint(equalTo: self.view.widthAnchor)
+      ])
```
amy-at-kickstarter, line 50: I'm surprised to see that you need to manipulate `messageBannerViewController?.view` here. Two questions:

1. Why is this extra constraint necessary for this PR but not for other/older uses of `MessageBannerViewController`?
2. Can this be moved into a method on `MessageBannerViewController`? Something like `messageBannerViewController.fixTheBannerWidth(to: self.view.widthAnchor)` would be more self-documenting.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -159,18 +169,35 @@ final class CommentRepliesViewController: UITableViewController {
...
+
+    self.viewModel.outputs.userBlocked
+      .observeForUI()
+      .observeValues { [weak self] success in
+        self?.commentComposer.isHidden = true
+
+        if success {
+          self?.messageBannerViewController?
+            .showBanner(with: .success, message: "This user has been successfully blocked")
+          self?.view.isUserInteractionEnabled = false
```
amy-at-kickstarter, line 181: I suspect that this behavior should be part of `MessageBannerViewController`. Having a banner that blocks interaction seems like a fairly common pattern. You could also add (if it doesn't already exist) a touch gesture that dismisses the banner when any additional area is tapped.

I peeked in `MessageBannerViewController` and it is an older-style class, which is probably way out of scope for any substantial edits, but it might be worthwhile folding in just this bit while you're in there.

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
scottkicks: So for some reason, that I haven't quite figured out yet, when this banner is added to. UITableViewController a right constraint of 1000 is added. If you look in `MessageBannerViewController` you'll see that it should be constrained to the right and left constraint of its parent view, but something is overwriting that. 
I like your idea of moving this to `MessageBannerViewController ` so that it can be more self-documenting. Without knowing the root cause I'm not sure how to approach that though.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
scottkicks: This is interesting. It would require some reworking of `MessageBannerViewController` to hold on to a reference to the parent view, but that might end up being less involved and we might be able to remove the delegate method altogether. I'll look into it.

a touch gesture that dismisses the banner already does exists 👍

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
scottkicks: this method only needs to be aware of the banner type though. what value do we gain from passing the entire object?

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
scottkicks: actually - since this banner is being used in other places in the app I'd rather address any refactoring of this class in a separate PR. This functionality is specific to the blocking flow currently as well and I'd rather not risk introducing regressions elsewhere.

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
amy-at-kickstarter: Ha, we gain nothing right now, but if you _did_ need the dismiss behavior to do something else in the future, it would be simpler to implement. However, that's why I'm thinking of this as a nit; it's really more of a style detail.

You may not need the delegate method at all if you're able to pull the `userInteractionDisabled` stuff into the view controller.

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
amy-at-kickstarter: Very weird! (But intriguing). Have you tried the constraints/view debugger? Sometimes if you print the description of the constraint it'll tell you where it came from.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
amy-at-kickstarter: I think that's fair.

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
scottkicks: I couldn't see why the tableviewcontroller was adding this extra trailing constraint to the banner but, I was able to get around this by setting the banner's widthAnchor rather than using leftAnchor + rightAnchor. 

I tested the other areas of the app where we use this banner and everything still looks good there.

Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -231,6 +252,15 @@ extension CommentRepliesViewController: RootCommentCellDelegate {
   }
 }
 
+// MARK: - MessageBannerViewControllerDelegate
+
+extension CommentRepliesViewController: MessageBannerViewControllerDelegate {
+  func messageBannerViewDidHide(type _: MessageBannerType) {
+    self.commentComposer.isHidden = false
+    self.view.isUserInteractionEnabled = true
```
amy-at-kickstarter, line 260: I took a peek in `MessageBannerViewController`, and it does already reference `self.view.superview`, so it might be safe to move `isUserInteractionEnabled` in there. But that's not a blocker, as previously discussed.

amy-at-kickstarter (approved):
I didn't test on my machine, but 👀 code LGTM.

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -87,6 +91,19 @@ internal final class MessagesViewController: UITableViewController {
...
+    self.viewModel.outputs.userBlocked
+      .observeForUI()
+      .observeValues { [weak self] success in
+
+        if success {
+          self?.messageBannerViewController?
+            .showBanner(with: .success, message: "This user has been successfully blocked")
+        } else {
+          self?.messageBannerViewController?
+            .showBanner(with: .error, message: "Your request did not go through. Try again.")
```
ifosli, line 104: Nit: Add a TODO here and for the other banner messages to update them once translated strings are ready

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -170,16 +189,26 @@ extension MessagesViewController: BackingCellDelegate {
 // MARK: - MessageCellDelegate
 
 extension MessagesViewController: MessageCellDelegate {
-  func messageCellDidTapHeader(_ cell: MessageCell, _: User) {
+  func messageCellDidTapHeader(_ cell: MessageCell, _ author: User) {
```
ifosli, line 192: Nit: I don't think `author` is the right name in this context - just `user` would be better

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -99,6 +108,7 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
 
     if !isHidden {
       self.view.superview?.bringSubviewToFront(self.view)
+      self.view.superview?.isUserInteractionEnabled = false
```
ifosli, line 111: Did you want the message banner class to be responsible for turning on and off user interaction? I'm okay with that (as long as that works for all current call sites), but then the corresponding lines of code in the individual classes need to be deleted.

ifosli (approved):
Looks good! I added some nits that would be nice to fix before merging, but I don't think any of them are complicated enough that they'll require another review pass.

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
scottkicks: Yep - It does work for the other call sites. Good catch. I thought I had gotten them all. Will remove it from the last 2 classes.

PR #1882, opened 2023-11-13
[MBL-957] Add RichPushNotifications push service extension to supportrich push notifications with Braze

amy-at-kickstarter:
# 📲 What

Add a service extension, `RichPushNotifications`, to support rich push notifications with Braze.

# 🤔 Why

This will allow our content marketing team to send push notifications that include images and gifs.

# ✔️ How to test this

To test this locally:

1. Double-check that your provisioning profiles are set up to run the `Beta` scheme for Kickstarter-iOS. You can do this by signing into your developer account in Xcode and checking the automatic code signing switch for the Beta scheme.
2. Double-check that your provisioning profiles can run the `Beta` scheme for RichPushNotifications. The same instructions as above should work - but note that it must be using the *same developer account* as in step 1.
3. (Possibly optional) I can't run the Firebase Crashlytics script on my local machine. Go to Build Phases, find Firebase Crashlytics, and comment out the run command in that script like
```
  echo "Not running this"
  # ${SRCROOT}/bin/firebase-crashlytics-run.sh

```
4. Select the Beta scheme (by default, it's Debug) of Kickstarter-iOS and deploy the app onto your phone. (If this doesn't work, double check steps 1 and 2 or whatever code signing errors you get.)
5. Swap into Staging in the KickBeta app and log into an account.
6. In the Braze dashboard, select 'Segment_Kickstarter_Staging` as your environment.
7. Create a new campaign, target iOS, and add an image to the rich content section.
8. Swap to the 'test' tab. Delete the default testers (probably something like 'Newsletter Test Group' and instead add an individual. Use the e-mail you logged in with in step 5.
9. Hit send. If everything goes well, it will say 'Push successfully sent' and you'll receive your rich push notification shortly!

# ⏰ TODO

To complete this work, we need to create app ids and matching profiles for the `RichPushNotification` target. Because a service extension must include the app identifier of its parent in its own app identifier, we will need four separate app IDs:

`com.kickstarter.kickstarter.RichPushNotifications`
`com.kickstarter.kickstarter.debug.RichPushNotifications`
`com.kickstarter.kickstarter.beta.RichPushNotifications`
`com.kickstarter.kickstarter.kickAlpha.RichPushNotifications`

Once these are set up, we'll add them to our `Fastfile`, like so, and use the various `fastlane match` lanes to update our certs:

```
  desc "Fastlane Match - App Store"
  private_lane :match_appstore do
    match(
      app_identifier: ["com.kickstarter.kickstarter", "com.kickstarter.kickstarter.RichPushNotifications"],
      type: "appstore",
      ...
 ```

I've validated this work by running it for `com.kickstarter.kickstarter.debug.RichPushNotifications`, but I'd like to make sure that it looks good (and we're confident in this) before I do the same `fastlane` changes for our Enterprise and App Store targets.

### CODE REVIEW

File: RichPushNotifications/NotificationService.swift [hunk 1]
```diff
@@ -0,0 +1,102 @@
+import UIKit
```
amy-at-kickstarter, line 1: This is almost 100% copy-pasted from Braze's example code. The only difference is that I changed `print` to my own `printDebug` method to keep us from spamming the console in production. I gave this a read-through and I think this implementation is fine, if a bit verbose and cautious.

.fastlane/Fastfile
```diff
@@ -37,7 +37,8 @@ platform :ios do
   private_lane :match_development do
     match(
       app_identifier: [
-        "com.kickstarter.kickstarter.debug"
+        "com.kickstarter.kickstarter.debug",
+        "com.kickstarter.kickstarter.debug.RichPushNotifications"
```
amy-at-kickstarter, line 41: TODO: Before we ship this, I need to make this change for the Enterprise and App Store targets. But I wanted to get a sign-off from either @ifosli or @scottkicks before I do it - since I'll have to run/update our fastlane certs to do so. I'm fine doing that for the dev target, but want to make sure I don't bury us in certificate problems by doing it for the other app targets!

scottkicks: discussed offline and decided that we should have enough time during this sprint to address any certificate issues that come up.

amy-at-kickstarter: 💯 Yeah my plan is to do the other upgrades on Monday so I've got a full week of (non-Holiday) runway if someone goes haywire

amy-at-kickstarter: Done.

scottkicks (approved):

PR #1883, opened 2023-11-13
[MBL-1026] Filter comments from blocked users

ifosli:
# 📲 What

Filter comments from blocked users by replacing the author name, avatar, and comment text with "Blocked user" placeholders.

Note: This also deletes the `RootCommentCellViewModel` and switches the `RootCommentCell` to using the `CommentCellViewModel` instead. The `RootCommentCellViewModel` code was just a subset of the entire `CommentCellViewModel` and keeping it as a separate copy just meant that we needed to duplicate work to keep them in sync. With this change, all comment cells (`RootCommentCell`, `CommentCell`, `CommentRemovedCell`, `CommentPostFailedCell`) all use the same view model, so the author filtering will apply to all of them.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1026)

image

# ♿️ Accessibility / 🏎 Performance

Not tested; this change doesn't add any new UI.

# ✅ Acceptance criteria

- [x] There are currently no blocked users (and this isn't wired up to an API anyways), so the only way to test that the UI works as intended is to switch the default `isBlocked` value to true or have it read from some other field instead. But when the `isBlocked` value is true, the author's name and avatar as well as the comment text should be hidden (comment functionality is unchanged).
- [x] It should not be possible to block a user that's already blocked.

# ⏰ TODO

- [ ] A future PR will set the `isBlocked` field based on backend info.
- [ ] Strings will be updated in a future PR once translations are ready.
- [ ] When user is blocked, comments should be refreshed. This will be implemented in a separate PR as well. (That way, we can test it properly before merging, once blocking is possible.)


Kickstarter-iOS/Features/Comments/Views/Cells/RootCommentCell.swift
```diff
@@ -34,7 +34,7 @@ final class RootCommentCell: UITableViewCell, ValueCell {
       |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  private let viewModel = RootCommentCellViewModel()
+  private let viewModel = CommentCellViewModel()
```
amy-at-kickstarter, line 37: Nice cleanup here 👍

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -85,13 +85,32 @@ public final class CommentCellViewModel:
       .map { comment, _ in comment }
 
     self.authorImageURL = comment
-      .map(\.author.imageUrl)
+      .map {
+        if $0.author.isBlocked {
+          return "" // Use default avatar.
```
amy-at-kickstarter, line 90: Will this default avatar URL. be hardcoded somewhere? Or is some other piece of code checking for this empty string?

You could also throw this in the model code, instead. It's a little sneaky but means anyone working with/implementing comment code in the future doesn't need to remember this edge case.

Library/ViewModels/CommentCellViewModel.swift
```diff
@@ -85,13 +85,32 @@ public final class CommentCellViewModel:
...
+      }
+    }
 
-    self.authorName = comment.map(\.author.name)
+    self.authorName = comment.map {
+      if $0.author.isBlocked {
+        // TODO(MBL-1026): Use translated string once available.
+        return "Blocked User"
+      } else {
+        return $0.author.name
```
amy-at-kickstarter, line 111: Same comments here, re: putting this directly in the model. Would it make it simpler at all?

amy-at-kickstarter (approved):
LGTM; a few stylistic suggestions, but nothing blocking. Love to see all that code deleted.

Library/ViewModels/CommentCellViewModel.swift
ifosli: I think we may want to replace the current gray circle with an icon instead, but for now empty string -> nil url -> gray placeholder circle. I added the comment since I didn't feel `""` was clear about that by itself. Glad to see I was right about that! Would it be more clear if I said "Keep placeholder avatar" instead?

Library/ViewModels/CommentCellViewModel.swift
ifosli: I don't think this belongs in the model. I think it would complicate things (the model would no longer be encodable which would mess with tests & decoding would be more of a pain). Also, whether a user is blocked or not isn't purely a display thing; it also affects functionality, so it's not like we could do no special handling at all. We also can't consistently always hide blocked user info, since we do want to display usernames of blocked users in messages. Plus, at some point, we might want to add an "unblock" option here, in which case the view model will need to know the real user info.

Library/ViewModels/CommentCellViewModel.swift
amy-at-kickstarter: Yeah, maybe something like `// Use placeholder avatar instead of URL`?

Library/ViewModels/CommentCellViewModel.swift
amy-at-kickstarter: Fair!

PR #1885, opened 2023-11-15
MBL-1029: Refresh Discover page after a user is blocked

amy-at-kickstarter:
# 📲 What

Once you block a user, anywhere in the app, the Discover page will be refreshed.

# 🤔 Why

Once you've blocked someone, we want to stop showing that person's content. Refreshing the Discover page means that we'll re-fetch the list of projects, filtering out any projects by blocked users.

# ⏰ TODO

If I throw in a`logEvent()`, it indicates that my signal is hooked up correctly, but I don't think it's actually refreshing the page like I would expect. I could use a little help debugging this behavior to make sure it's working properly.

### DIFF
--- a/Library/ViewModels/DiscoveryPageViewModelTests.swift
+++ b/Library/ViewModels/DiscoveryPageViewModelTests.swift
@@ -1193,11 +1193,11 @@ internal final class DiscoveryPageViewModelTests: TestCase {
 
       self.projectsAreLoading.assertValueCount(2)
 
-      NotificationCenter.default.post(.init(name: .ksr_blockedUser))
+      self.vm.inputs.blockedUser()
 
       self.scheduler.advance()
 
-      self.projectsAreLoading.assertValueCount(6)
+      self.projectsAreLoading.assertValueCount(4)
     }
   }
 }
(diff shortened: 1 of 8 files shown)


Library/ViewModels/DiscoveryPageViewModelTests.swift
```diff
@@ -1175,4 +1175,29 @@ internal final class DiscoveryPageViewModelTests: TestCase {
...
+
+      self.scheduler.advance()
+
+      self.projectsAreLoading.assertValueCount(2)
+
+      self.vm.inputs.blockedUser()
+
+      self.scheduler.advance()
+
+      self.projectsAreLoading.assertValueCount(4)
```
amy-at-kickstarter, line 1200: There's a similar test for `pullToRefresh` that requires this assert to be `       self.projectsAreLoading.assertValueCount(6)`. But that seems a bit odd to me (why the extra values)? I'm not sure, then, if this is correct or not.

ifosli: I think that might be because there's a loading indicator that shows for `pullToRefresh` (and I don't think we need to show it here). I could be wrong about that, since I didn't look thoroughly at that code. But test looks good to me!

ifosli (changes requested):
This looks great! You're missing the `ProjectPageViewModel`, however. (Probably because that one currently maps to `false` instead of `true`.) Please send out a notification from there as well :)

amy-at-kickstarter:
> This looks great! You're missing the `ProjectPageViewModel`, however. (Probably because that one currently maps to `false` instead of `true`.) Please send out a notification from there as well :)

Whoops, good catch!

ifosli (approved):

PR #1886, opened 2023-11-16
Refresh comments when a user is blocked

ifosli:
# 📲 What

If a user is blocked, comments are reset - refetched and the user is jumped back to top. If a user is blocked from the `CommentRepliesViewController`, they're navigated back to the main `CommentsViewController`.

Note: I did consider refetching comments in place instead, but getting it to work with pagination seemed tricky, so I'm clearing state instead for now. If we have time, it would be nice to revisit this approach and try to let the user keep their place in our comments.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1026)

# ✅ Acceptance criteria

- [ ] Comment state is reset when a user is successfully blocked (from the commentsViewController or the CommentRepliesViewController)

### DIFF
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -10,6 +10,10 @@ private enum Layout {
   }
 }
 
+protocol CommentRepliesViewControllerDelegate: AnyObject {
+  func commentRepliesViewControllerDidBlockUser(_: CommentRepliesViewController)
+}
+
 final class CommentRepliesViewController: UITableViewController, MessageBannerViewControllerPresenting {
   // MARK: Properties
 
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -166,14 +172,16 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
     self.viewModel.outputs.userBlocked
       .observeForUI()
       .observeValues { [weak self] success in
-        self?.commentComposer.isHidden = true
+        guard let self else { return }
```
amy-at-kickstarter, line 175: Huh, nice shortcut. Didn't know you could do that 👍

amy-at-kickstarter (approved):
👍 lgtm

PR #1887, opened 2023-11-22
[MBL-1027] Blocked User Messages UI

scottkicks:
# 📲 What

If a user has blocked someone that they have a message thread with we need to disable the Reply button and inform users that they have blocked said user

# 🤔 Why

If a user has blocked someone they will no longer receive new messages or be able to message them. 

# 🛠 How

Uses the same design that we had for the creator dashboard deprecation warning banner to inform users. 
The isBlocked value is being hardcoded for now but this will be updated so that we get the isBlocked status from the backend instead in MBL-1024.

# 👀 See

| After 🦋 |
| Simulator Screen Recording - iPhone 15 Pro - 2023-11-22 at 12 53 28 |

# ✅ Acceptance criteria

- [ ] banner should show when the participant isBlocked
- [ ] banner should not cover any of the existing message thread

# ⏰ TODO

- get isBlocked status from the backend. [MBL-1024](https://kickstarter.atlassian.net/browse/MBL-1024?atlOrigin=eyJpIjoiZjVkYmQzM2FjMGNkNDA5MGI3ZDU0ZTY2NDRkYjU3YTMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

[MBL-1024]: https://kickstarter.atlassian.net/browse/MBL-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -161,6 +194,48 @@ internal final class MessagesViewController: UITableViewController, MessageBanne
...
+
+  private func showUserBlockedBanner() {
+    let userBlockedBannerHostingController = UIHostingController(rootView: UserBlockedBannerView())
+
+    guard let tabController = self.tabBarController as? RootTabBarViewController,
+      let userBlockedBannerView = userBlockedBannerHostingController.view else { return }
+
+    userBlockedBannerView.tag = -99
+    tabController.addChild(userBlockedBannerHostingController)
+    tabController.view.addSubview(userBlockedBannerView)
```
scottkicks, line 226: adding the banner to the tabController so rather than the viewcontroller directly. This helped to avoid some more specific styling that we would need to do in order to have this banner sit above the tab bar and not block the ViewController's content.

With this approach we can simply call `updateTableViewBottomContentInset` on viewDidAppear.

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -161,6 +194,51 @@ internal final class MessagesViewController: UITableViewController, MessageBanne
...
+        .bottom = userBlockedBannerHostingController.view.bounds
+        .height
+    }
+  }
+
+  private func removeUserBlockedBanner() {
+    guard let bannerHostingController = self.userBlockedBannerHostingController,
+      let tabBarController = self.tabBarController else { return }
+
+    bannerHostingController.removeFromParent()
```
amy-at-kickstarter, line 209: These are missing `willMoveToParent` and `didMoveToParent` (see: [Creating a Custom Container View Controller](https://developer.apple.com/documentation/uikit/view_controllers/creating_a_custom_container_view_controller))

Kickstarter-iOS/Features/UserBlockedBanner/UserBlockedBannerView.swift
```diff
@@ -0,0 +1,27 @@
+import Library
```
amy-at-kickstarter, line 1: Does this need custom UI beyond our normal banner? Why not use a `MessageBannerViewController`?

Kickstarter-iOS/Features/Messages/Controller/MessagesViewController.swift
```diff
@@ -161,6 +194,51 @@ internal final class MessagesViewController: UITableViewController, MessageBanne
...
 
+  private func updateTableViewBottomContentInset() {
+    if let userBlockedBannerHostingController = userBlockedBannerHostingController {
+      self.tableView.contentInset
+        .bottom = userBlockedBannerHostingController.view.bounds
+        .height
+    }
+  }
+
+  private func removeUserBlockedBanner() {
```
amy-at-kickstarter, line 205: Having the presentation/dismissal code live inside `MessagesViewController` feels like the wrong place to me, since this is code that's largely acting on the UITabBarController, not this view controller. I'd recommend putting this into a quick extension on `UITabBarController` - then you get a clean call from here, like ``self.tabBarController.showBlockedBanner()``.

Library/ViewModels/MessagesViewModel.swift
```diff
@@ -57,8 +63,8 @@ public protocol MessagesViewModelOutputs {
   /// Emits when the thread has been marked as read.
   var successfullyMarkedAsRead: Signal<(), Never> { get }
 
-  /// Emits when a request to block a user has been made
-  var userBlocked: Signal<Bool, Never> { get }
+  /// Emits whether arequest to block a user was successful.
```
amy-at-kickstarter, line 66: Nit: typo 😄

Library/ViewModels/MessagesViewModel.swift
```diff
@@ -191,12 +198,16 @@ public final class MessagesViewModel: MessagesViewModelType, MessagesViewModelIn
...
 
-    self.userBlocked.observeValues { didBlock in
+    self.userHasBeenBlocked.observeValues { didBlock in
       guard didBlock == true else { return }
       NotificationCenter.default.post(.init(name: .ksr_blockedUser))
     }
+
+    /// TODO(MBL-1025): Get isBlocked status from the backend instead.
+    self.participantIsBlocked = self.viewWillAppearProperty.signal
+      .mapConst(true)
```
amy-at-kickstarter, line 210: Two things: 

1. Obviously a ``//TODO`` but I would merge this as `false`, since the feature flag defaults to off.
2. Is this redundant with `userHasBeenBlocked`? Why not just emit an initial value of `userHasBeenBlocked` on `viewDidAppear`?

amy-at-kickstarter (changes requested):
There are a couple things in here I have questions about before merging. My specific comments are inline in the code, but here are the three things I'd like to see addressed:

1. New banner view, vs. re-using old banner view
2. Suggestions for improving the code to display a banner from the tab bar controller
3. Clarifying `participantIsBlocked` vs. `userHasBeenBlocked` signals

Kickstarter-iOS/Features/UserBlockedBanner/UserBlockedBannerView.swift
scottkicks: the MessageBannerViewController UX is typically used to show a brief message that is dismissed after a few seconds. Messages like, "Your request was successful", etc. I wanted to have this user-blocked UI be static and didn't want to potentially confuse the UX

Library/ViewModels/MessagesViewModel.swift
scottkicks: That's fair. 
`participantIsBlocked` tells us whether the user is blocked when opening a thread
`userHasBeenBlocked` tells us whether a new request to block someone was successful or not. 

They have different intentions so I opted to keep them separated.

amy-at-kickstarter: Perhaps a naming change would help then - just to throw out the first idea on the top of my head, `participantWasBlockOnLoad` and `didBlockUser`

amy-at-kickstarter: Actually, one more thought here - what is supposed to happen in this case:

1. View messages, no one is blocked
2. You block a user
3. Should reply be disabled? I assume showing the banner is redundant, but perhaps these signals are still kind of linked.

scottkicks:
@amy-at-kickstarter using the existing uikit banner. had to do some rework to control the dismissal of the banner, but i think I've addressed everything now.

Library/ViewModels/MessagesViewModel.swift
```diff
@@ -191,12 +198,16 @@ public final class MessagesViewModel: MessagesViewModelType, MessagesViewModelIn
...
 
-    self.userBlocked.observeValues { didBlock in
+    self.didBlockUser.observeValues { didBlock in
       guard didBlock == true else { return }
       NotificationCenter.default.post(.init(name: .ksr_blockedUser))
     }
+
+    /// TODO(MBL-1025): Get isBlocked status from the backend instead.
+    self.participantPreviouslyBlocked = self.viewWillAppearProperty.signal
+      .mapConst(false)
```
ifosli, line 210: This is already available in `project.creator.isBlocked` (though not wired up to the backend, so not exactly testable). I'd recommend just using it now; then you can update the test as well.

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -19,6 +19,7 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
   @IBOutlet fileprivate var messageLabel: UILabel!
 
   private var bannerType: MessageBannerType?
+  private var dissmissable: Bool = true
```
ifosli, line 22: nit: `dismissible` - here and everywhere this property is used.

ifosli (commented):
I didn't run and test this, but looks good apart from the spelling error! (And you can already use the `isBlocked` field on the creator if you'd like to.)

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -66,7 +67,13 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
...
       .observeForUI()
       .observeValues { [weak self] isHidden in
-        self?.showViewAndAnimate(isHidden)
+        guard let self else { return }
+
+        if isHidden, !self.dissmissable {
+          return
+        }
+
+        self.showViewAndAnimate(isHidden)
```
amy-at-kickstarter, line 76: Obviously this isn't part of this PR, but what a confusing name! Something like `self.animate(showView: !isHidden)` would be clearer to me. Ah well, something for a future fix.

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -95,20 +102,26 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
...
     let duration = isHidden ? AnimationConstants.hideDuration : AnimationConstants.showDuration
 
     let hiddenConstant = self.view.frame.height + (self.view.superview?.safeAreaInsets.bottom ?? 0)
 
     if !isHidden {
       self.view.superview?.bringSubviewToFront(self.view)
-      self.view.superview?.isUserInteractionEnabled = false
+
+      if self.dissmissable {
+        self.view.superview?.isUserInteractionEnabled = false
```
amy-at-kickstarter, line 123: Just to double-check, nowhere else needs this protection, right? Just to make sure we don't end up disabling the whole UI 😆

amy-at-kickstarter (approved):
LGTM. I also like @ifosli's suggestions!

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
scottkicks: correct. Only when the banner is dismissible do we want to disable interaction.

PR #1792, opened 2023-02-13
[WEB-671]: Setup TriggerCapiEventInput GraphQL Mutation

scottkicks:
# 📲 What

Before we can send CAPI events, we need to implement a mutation that will allow us to send the relevant data to our server.
This mutation will be called in the following cases when:
1. Members have given consent for KS to track their data
2. The relevant project has it's `sendMetaCapiEvents` property set to true

**Event Cases:**
- Project page visit
- Select back this project
- Initiate checkout (selects a reward)
- Add payment info tapped (could be swapping existing card or adding anew card)
- Purchase (have reached the thank you page)

# 🤔 Why

If Facebook app users decline permission for FB to track them for advertising purposes, then our KS Facebook Pixel will be rendered useless. Using Facebook's Conversions API (CAPI) is the recommended workaround. 

# 🛠 How

The web team will be handling the actual events to Facebook's Conversions API and has added this new mutation so that we can notify them about what data and which events to send. 

# ✅ Acceptance criteria

- [x] Network call is readily available and usable.

PR #1890, opened 2023-11-27
MBL-1022: Add blocking mutation to iOS client

amy-at-kickstarter:
# 📲 What

Add `BlockUser` mutation to the app.

# 🤔 Why

So we can implement blocking in the UI!


KsApi/Service.swift
```diff
@@ -790,4 +790,13 @@ public struct Service: ServiceType {
       .perform(mutation: GraphAPI.WatchProjectMutation(input: GraphAPI.WatchProjectInput.from(input)))
       .flatMap(WatchProjectResponseEnvelope.producer(from:))
   }
+
+  public func blockUser(userId: String)
+    -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope> {
+    return GraphQL.shared.client
+      .perform(mutation: GraphAPI.BlockUserMutation(input: GraphAPI.BlockUserInput(blockUserId: userId)))
+      .flatMap { _ in
+        SignalProducer(value: EmptyResponseEnvelope())
```
amy-at-kickstarter, line 799: So `BlockUser` _can_ return two fields: `success` and `currentUser { blockedUsers } `. However, based on my current understanding of the client designs, we don't need to read either of these fields: we'll fetch `isBlocked` directly on other `User` objects.

Is that correct? Based on that assumption I figured I'd take the simplest route out and return an `EmptyResponseEnvelope`.

KsApi/Service.swift
```diff
@@ -790,4 +790,13 @@ public struct Service: ServiceType {
       .perform(mutation: GraphAPI.WatchProjectMutation(input: GraphAPI.WatchProjectInput.from(input)))
       .flatMap(WatchProjectResponseEnvelope.producer(from:))
   }
+
+  public func blockUser(userId: String)
+    -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope> {
+    return GraphQL.shared.client
+      .perform(mutation: GraphAPI.BlockUserMutation(input: GraphAPI.BlockUserInput(blockUserId: userId)))
```
amy-at-kickstarter, line 797: n.B. `GraphAPI.BlockUserMutation` wants us to pass it a `GraphQLID` type, which can also be a `String`. It's not obvious to me if we'll need to base64-encode the value like a proper GraphQL identifier, or if it will be happy taking a plain, string-based user ID like `"12345"`.

scottkicks: Let's follow the same "adapter" pattern that we're using elsewhere here. We can create an extension on `GraphAPI.BlockUserInput` for creating the input type. Extracting this allows us to test it more easily. See `GraphAPI.WatchProjectInput.from(_ input:...` for an example

KsApi/Service.swift
scottkicks: Yep exactly. We only need to know if it was successful right now.

KsApi/ServiceType.swift
```diff
@@ -380,6 +380,8 @@ public protocol ServiceType {
 
   func fetchBackedProjects(cursor: String?, limit: Int?)
     -> SignalProducer<FetchProjectsEnvelope, ErrorEnvelope>
+
+  func blockUser(userId: String) -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope>
```
scottkicks, line 384: So, this will be `func blockUser(input: BlockUserInput) -> SignalProducer<EmptyResponseEnvelope, ErrorEnvelope>`

scottkicks (commented):
Nice first run through! There's just a couple of things we need to add to stick to the existing pattern and keep things well tested. I added inline comments and you can use the linked PR below as a guide. 

Let's create a `BlockUserInput` type to inject into our Service method as well as a `.from(_ input...)` helper extension to make it easier on ourselves to bundle up the expected properties once we implement this. Check out [this PR](https://github.com/kickstarter/ios-oss/pull/1792) out as a reference

Oh and we can add some tests around the input and extension as well 😄

KsApi/GraphAPI.swift
```diff
@@ -6716,7 +6841,7 @@ public enum GraphAPI {
       }
 
       public struct Node: GraphQLSelectionSet {
-        public static let possibleTypes: [String] = ["Backing", "Reward", "Photo", "RewardItem", "Project", "Comment", "User", "Address", "Conversation", "Message", "CuratedPage", "Location", "Organization", "UserUrl", "Category", "AiDisclosure", "BusinessAddress", "ProjectFeaturedImage", "Flagging", "Video", "VideoTrack", "VideoTrackCue", "ProjectProfile", "AttachedAudio", "AttachedVideo", "Tag", "CreatorInterview", "InterviewAnswer", "InterviewQuestion", "CreatorPrompt", "FreeformPost", "ShippingRule", "Checkout", "Order", "OrderAddress", "Survey"]
+        public static let possibleTypes: [String] = ["Backing", "Reward", "Photo", "RewardItem", "Project", "Comment", "User", "Address", "Conversation", "Message", "CuratedPage", "Location", "Organization", "UserUrl", "Category", "AiDisclosure", "BusinessAddress", "Flagging", "Video", "VideoTrack", "VideoTrackCue", "ProjectProfile", "AttachedAudio", "AttachedVideo", "Tag", "CreatorInterview", "InterviewAnswer", "InterviewQuestion", "CreatorPrompt", "FreeformPost", "ShippingRule", "Checkout", "Order", "OrderAddress", "Survey"]
```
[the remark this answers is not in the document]
amy-at-kickstarter: One point worth noting - right now we only update the schema when we _need_ to update it on the client. But we should really be updating it any time changes to the web schema are pushed out - otherwise, we risk getting out of sync, and (say) using a feature that no longer actually exists on the web.

Obviously I'd hope that someone working on the web side would point this out to us before merging that change! But nice to enforce the same thing through code, not just socially.

KsApi/models/graphql/adapters/GraphAPI.BlockUserInput+BlockUserInput.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
amy-at-kickstarter, line 1: For what it's worth - I'm happy to follow the pattern for now, but considering how simple this input is, I'm not entirely sure it's worth having this extra boilerplate for the adapter + adapter tests. Just something to consider in the back of our minds as we move forward with other kinds of API refactoring in the future.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/mutations/inputs/BlockUserInput.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
```
amy-at-kickstarter, line 1: @scottkicks Can you explain a little more about this input adapter pattern? Happy to follow it for now, but at least for a very teeny input like this, it doesn't seem obviously beneficial. I'm curious to hear more about why we've been using this pattern instead of using the `GraphAPI.XYZInput` objects directly.

scottkicks:
@amy-at-kickstarter Until we can say this this current mutation is working/testable we might want to hold off on merging this. In case Santiago has to make further changes. ref: https://kickstarter.slack.com/archives/C05HX9GS0CD/p1701219627599909?thread_ts=1701184665.560489&cid=C05HX9GS0CD

scottkicks (approved):

PR #1891, opened 2023-11-29
[MBL-1025] Update User object with server value for isBlocked

ifosli:
# 📲 What

This makes the `isBlocked` bool use the server value, for the `User` model (and the `Author` subset), whether it's fetched using GraphQL or v1. Tests are updated as well.

Note: I also switched `isBlocked` to a required instead of an optional bool, since that feels easier to use. If it's missing from the server for some reason, we default to `false`.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1025)

# ✅ Acceptance criteria

- [x] Comments from a blocked user show as the "Blocked user" placeholder comment (tests `Author`/graphQL)
- [x] Project creator's `isBlocked` corresponds to whether or not the user is actually blocked (tests `User`/v1)
- [x] Verify that posting a comment still works as expected (this is one of the mutations that was changed by adding `isBlocked` to the `Author`


Library/ViewModels/MessagesViewModelTests.swift
```diff
@@ -240,7 +240,7 @@ internal final class MessagesViewModelTests: TestCase {
       self.scheduler.advance()
 
       self.messages.assertValueCount(1)
-      self.replyButtonIsEnabled.assertValues([false, false, true])
+      self.replyButtonIsEnabled.assertValues([false, true, true])
```
ifosli, line 243: This change is because the user template now has the `isBlocked` explicitly set, so the `isBlocked == false` check causes true to be emitted instead. (I also changed isBlocked to be required to avoid weirdness like this in the future; optional bools are weird, especially when we expect them to actually exist.)

scottkicks (approved):
nice once!

Library/ViewModels/MessagesViewModelTests.swift
amy-at-kickstarter: This comment would also be good in the code!

amy-at-kickstarter (approved):
👍

PR #1893, opened 2023-11-30
[MBL-1024] Wire Up Block User Mutation

scottkicks:
# 📲 What

Now that the block mutation has been [added](https://github.com/kickstarter/ios-oss/pull/1890), let's implement it.

# 🤔 Why

So we can block users!

# 🛠 How

We are letting users block someone from the following places:
- Project Page
- Message Threads
- Comments & Comment Replies

When we call this mutation via `AppEnvironment.current.apiService.blockUser...` either an error will occur or it'll be successful. So, in each view model, I've set up an output for each case that will in turn present a success or error message banner. 

We currently don't have the error messages that we can receive from the backend translated and localized so for now we are showing a generic "This didn't work" message. Not ideal. Showing a more useful message like, "You can only block someone that doesn't have a live project you have backed" would be better for obvious reasons so I think this should be a fast follow. Added a //TODO for this inline.

# 👀 See

expects the same UI/UX that we've been working with up to now. 

# ✅ Acceptance criteria

- [ ] We can block someone through each location listed above.
- [ ] Attempting to block someone who has already been blocked will return an error
- [ ] Attempting to block someone without internet will throw an error
- [ ] The UI reflects newly blocked users
   - [ ] Info banner in message threads
   - [ ] Blocked user comments
   - [ ] Once the backend is finished projects by blocked users will be filtered out, so there is no UI to inform you of this directly on a project page.

- you can also double-check that someone has been blocked successfully here https://staging.kickstarter.com/graphiql 

# ⏰ TODO

- [ ] Waiting on the `id` vs `uid` change to be made on prod. Once that's updated we can test that the app reloads after blocking and filters out blocked content correctly
- [x] Add VM tests

### DIFF
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -169,26 +169,28 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
         }
       }
 
-    self.viewModel.outputs.userBlocked
+    self.viewModel.outputs.didBlockUser
       .observeForUI()
-      .observeValues { [weak self] success in
-        guard let self else { return }
-        self.commentComposer.isHidden = true
+      .observeValues { [weak self] _ in
+        guard let self, let messageBanner = self.messageBannerViewController else { return }
 
-        if success {
-          self.messageBannerViewController?
-            .showBanner(with: .success, message: Strings.Block_user_success())
-          self.delegate?.commentRepliesViewControllerDidBlockUser(self)
-        } else {
-          self.messageBannerViewController?
-            .showBanner(with: .error, message: Strings.Block_user_fail())
-        }
+        messageBanner.showBanner(with: .success, message: Strings.Block_user_success())
+      }
+
+    self.viewModel.outputs.didBlockUserError
+      .observeForUI()
+      .observeValues { [weak self] _ in
+        guard let self, let messageBanner = self.messageBannerViewController else { return }
+
+        messageBanner.showBanner(with: .error, message: Strings.Block_user_fail())
       }
   }
 
-  private func presentBlockUserAlert(username: String) {
+  private func presentBlockUserAlert(username: String, userId: String) {
     let alert = UIAlertController
-      .blockUserAlert(username: username, blockUserHandler: { _ in self.viewModel.inputs.blockUser() })
+      .blockUserAlert(username: username, blockUserHandler: { _ in
+        self.viewModel.inputs.blockUser(id: userId)
+      })
     self.present(alert, animated: true)
   }
 
@@ -198,7 +200,7 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
 
     let actionSheet = UIAlertController
       .blockUserActionSheet(
-        blockUserHandler: { _ in self.presentBlockUserAlert(username: author.name) },
+        blockUserHandler: { _ in self.presentBlockUserAlert(username: author.name, userId: author.id) },
         sourceView: cell,
         isIPad: self.traitCollection.horizontalSizeClass == .regular
       )
--- a/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
+++ b/Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
@@ -174,6 +174,9 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
       .observeValues { [weak self] _ in
         guard let self, let messageBanner = self.messageBannerViewController else { return }
 
+        self.commentComposer.isHidden = true
+        self.delegate?.commentRepliesViewControllerDidBlockUser(self)
+
         messageBanner.showBanner(with: .success, message: Strings.Block_user_success())
       }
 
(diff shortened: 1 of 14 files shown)


Kickstarter-iOS/Features/Comments/Controllers/CommentRepliesViewController.swift
```diff
@@ -169,26 +169,28 @@ final class CommentRepliesViewController: UITableViewController, MessageBannerVi
...
-            .showBanner(with: .success, message: Strings.Block_user_success())
-          self.delegate?.commentRepliesViewControllerDidBlockUser(self)
-        } else {
-          self.messageBannerViewController?
-            .showBanner(with: .error, message: Strings.Block_user_fail())
-        }
+      .observeValues { [weak self] _ in
+        guard let self, let messageBanner = self.messageBannerViewController else { return }
+
+        messageBanner.showBanner(with: .success, message: Strings.Block_user_success())
```
amy-at-kickstarter, line 177: It looks like you removed the line `self.commentComposer.isHidden = true` in this refactoring, as well as `self.delegate?.commentRepliesViewControllerDidBlockUser(self)`. Were these deletions intentional?

Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
```diff
@@ -737,9 +741,12 @@ public final class ProjectPageViewController: UIViewController, MessageBannerVie
     }
   }
 
-  private func presentBlockUserAlert(username: String) {
+  private func presentBlockUserAlert(username: String, userId: Int) {
     let alert = UIAlertController
-      .blockUserAlert(username: username, blockUserHandler: { _ in self.viewModel.inputs.blockUser() })
+      .blockUserAlert(username: username, blockUserHandler: { _ in
+        self.viewModel.inputs.blockUser(id: "\(userId)")
```
amy-at-kickstarter, line 747: Should the `viewModel` take an `int` as input instead?

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -248,18 +251,25 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
       .takeWhen(self.dataSourceLoadedProperty.signal)
 
-    // TODO: Call blocking GraphQL mutation
-    self.userBlocked = self.blockUserProperty.signal.map { true }
+    let blockUserEvent = self.blockUserProperty.signal
+      .map(BlockUserInput.init(blockUserId:))
+      .switchMap { input in
+        AppEnvironment.current.apiService
+          .blockUser(input: input)
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
```
amy-at-kickstarter, line 259: Just for my own edition, what's up with `ksr_delay` everywhere when we make network requests?

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -248,18 +251,25 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
 
-    self.userBlocked.observeValues { didBlock in
-      guard didBlock == true else { return }
-      NotificationCenter.default.post(.init(name: .ksr_blockedUser))
-    }
+    self.didBlockUser = blockUserEvent.values().ignoreValues()
+      .map { _ in NotificationCenter.default.post(.init(name: .ksr_blockedUser)) }
+
+    // TODO: Display proper error messaging from the backend
+    self.didBlockUserError = blockUserEvent.errors().ignoreValues()
```
amy-at-kickstarter, line 267: @santiDotIO asked about this, but do we even want to bother with server-driven error messages here? I can think of some that might be nice (this user has already been blocked, for example) but I'm assuming errors here are such a small edge case. Might not be worth the work to plumb them in and translate them.

Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
scottkicks: Well, the mutation expects a string. I could convert it to a string in the view model. Do you have a strong preference here?

scottkicks:
@amy-at-kickstarter Thanks - yea, I have add tests on my to-do list in the description above. I wanted to jump-start the review while I do those.

Kickstarter-iOS/Features/Comments/Controllers/CommentsViewController.swift
```diff
@@ -267,7 +268,7 @@ extension CommentsViewController {
 
 extension CommentsViewController: CommentCellDelegate {
   func commentCellDidTapHeader(_ cell: CommentCell, _ author: Comment.Author) {
-    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else { return }
+//    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else { return }
```
amy-at-kickstarter, line 271: intentional?

scottkicks: 🤦

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/CommentRepliesViewModelTests.swift
```diff
@@ -108,6 +111,67 @@ internal final class CommentRepliesViewModelTests: TestCase {
     }
   }
 
+  func testDidBlockUser_EmitsOnSuccess() {
```
amy-at-kickstarter, line 114: 👍

scottkicks:
> Do we have a pattern for sharing code between view models?

It's a good callout. We could create a BlockUser service that can be injected and used across the app. I opted to stick with what's more common in the app currently because this is more time-sensitive, but I think it could be cleaner.

PR #1897, opened 2023-12-04
[MBL-1075] Prevent blocking for users that are already blocked

ifosli:
# 📲 What

If a user is already blocked, we'll revert to old behavior (clicking on the username on the project page opens the creator page immediately and clicking usernames elsewhere has no effect.)

Note: this required no changes in comments, since that was already checking for blocked users.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1075)

# ✅ Acceptance criteria

- [x] User should not be blockable if already blocked. This is currently only testable on the project page

# ⏰ TODO

- [ ] Messages API doesn't include `isBlocked` info about the sender, so we'll need to test this for messages once that's fixed (probably after this PR is submitted).

### DIFF
--- a/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
+++ b/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
@@ -989,7 +989,11 @@ extension ProjectPageViewController: ProjectPamphletMainCellDelegate {
     _ cell: ProjectPamphletMainCell,
     goToCreatorForProject project: Project
   ) {
-    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else {
+    guard
+      AppEnvironment.current.currentUser != nil,
+      featureBlockUsersEnabled(),
+      !project.creator.isBlocked
+    else {
       self.goToCreatorProfile(forProject: project)
       return
     }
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
```diff
@@ -989,7 +989,11 @@ extension ProjectPageViewController: ProjectPamphletMainCellDelegate {
     _ cell: ProjectPamphletMainCell,
     goToCreatorForProject project: Project
   ) {
-    guard AppEnvironment.current.currentUser != nil, featureBlockUsersEnabled() else {
+    guard
+      AppEnvironment.current.currentUser != nil,
+      featureBlockUsersEnabled(),
+      !project.creator.isBlocked
```
scottkicks, line 995: I wonder if this is necessary since users shouldn't be able to view projects by blocked creators.

ifosli: The projects won't show up in search or discover, but they could still get there from a deeplink. Or if they have a message history with the creator (unusual, but possible), then they can navigate there from messages.

amy-at-kickstarter (approved):
👍

scottkicks (approved):

PR #1900, opened 2023-12-05
MBL-1016: Add custom operators for common API patterns

amy-at-kickstarter:
# 📲 What

Add two custom operators for API patterns that (I expect) will be fairly common in Combine.

# 🤔 Why

As part of the SwiftUI migration work, we want to establish some repeatable patterns for the future. When I ported `ReportProjectFormViewModel` to be completely in Combine, I left two somewhat confusing operations - a way to map fetch results, and a way to handle API failures without cancelling the pipeline. I wanted to make a cleaner interface for these two patterns, since I assume they'll pop up again.

### DIFF
--- a/Kickstarter-iOS/SharedViews/LoadingBarButtonItem.swift
+++ b/Kickstarter-iOS/SharedViews/LoadingBarButtonItem.swift
@@ -1,7 +1,6 @@
 import Library
 import SwiftUI
 
-// TODO(MBL-1039) - Refactor this so that saveTriggered takes a closure, not a binding
 struct LoadingBarButtonItem: View {
   @Binding var saveEnabled: Bool
   @Binding var showLoading: Bool
--- a/KsApi/Publisher+Service.swift
+++ b/KsApi/Publisher+Service.swift
@@ -22,4 +22,27 @@ extension Publisher {
     }
     .eraseToAnyPublisher()
   }
+
+  /// A convenience method for gracefully catching API failures.
+  /// If you handle your API failure in receiveCompletion:, that will actually cancel the entire pipeline, which means the failed request can't be retried.
+  /// This is a wrapper around the .catch operator, which just makes it a bit easier to read.
+  ///
+  /// An example:
+  /// ```
+  /// self.somethingHappened
+  ///    .flatMap() { _ in
+  ///        self.doAnAPIRequest
+  ///            .handleFailureAndAllowRetry() { e in
+  ///               showTheError(e)
+  ///            }
+  ///    }
+
+  public func handleFailureAndAllowRetry(_ onFailure: @escaping (Self.Failure) -> Void)
+    -> AnyPublisher<Self.Output, Never> {
+    return self.catch { e in
+      onFailure(e)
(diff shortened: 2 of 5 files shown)


Kickstarter-iOS/SharedViews/LoadingBarButtonItem.swift
```diff
@@ -1,7 +1,6 @@
 import Library
 import SwiftUI
 
-// TODO(MBL-1039) - Refactor this so that saveTriggered takes a closure, not a binding
```
amy-at-kickstarter, line 4: Did this a while ago but forgot the `//TODO`.

KsApi/Publisher+Service.swift
```diff
@@ -0,0 +1,48 @@
...
+  /// ```
+  /// self.somethingHappened
+  ///    .flatMap() { _ in
+  ///        self.doAnAPIRequest
+  ///            .handleFailureAndAllowRetry() { e in
+  ///               showTheError(e)
+  ///            }
+  ///    }
+
+  public func handleFailureAndAllowRetry(_ onFailure: @escaping (Self.Failure) -> Void)
```
amy-at-kickstarter, line 40: I had an "interesting" learning experience writing this operator. I spent about two hours trying to wrangle the types (with generics) for this function, and got completely stuck. I ended up changing things around a bit so that the types were simpler, which worked - but I think it's worth noting that any custom operators that handle fairly complex type transformations (say, with a `flatMap`) are likely to be beastly to write.

KsApi/Service.swift
```diff
@@ -367,24 +367,9 @@ public struct Service: ServiceType {
...
-      }
-      .mapError { rawError in
-
-        if let error = rawError as? ErrorEnvelope {
-          return error
-        }
-
-        return ErrorEnvelope.couldNotParseJSON
+      .mapFetchResults { (data: GraphAPI.FetchUserEmailQuery.Data) -> UserEnvelope<GraphUserEmail>? in
+        UserEnvelope<GraphUserEmail>.userEnvelope(from: data)
```
amy-at-kickstarter, line 371: Much cleaner!

KsApi/Publisher+Service.swift
```diff
@@ -0,0 +1,48 @@
+import Combine
+import Foundation
+
+extension Publisher {
+  /// A convenience method for mapping the results of your fetch to another data type. Any unknown errors are returned in the error as `ErrorEnvelope.couldNotParseJSON`.
+  func mapFetchResults<NewOutputType>(_ convertData: @escaping ((Output) -> NewOutputType?))
+    -> AnyPublisher<NewOutputType, ErrorEnvelope> {
+    return self.tryMap { (data: Output) -> NewOutputType in
+      guard let envelope = convertData(data) else {
+        throw ErrorEnvelope.couldNotParseJSON
```
amy-at-kickstarter, line 10: We could consider a different generic error here, if we wished, but I think this is fine.

ifosli (approved):
I like it!

PR #1903, opened 2023-12-11
Stop using rawState in exportDataEnvelope

ifosli:
# 📲 What

This is a cleaner way to handle any future server changes; I just didn't understand decoders well enough to to this initially. The actual behavior (unknown strings default to `.unknown` and anything that can't be parsed as a string throws an error) remains the same.

# ✅ Acceptance criteria

- [x] Tests pass
- [x] Project builds & personal data can still be requested


ifosli:
> This LGTM, but I'd like to see a really quickie test just for a valid and an invalid enum value.
> 
> I'm also slightly surprised that the default implementation of `Decodable` for an `Enum` doesn't _already_ do this - what happened before you implemented this change?

If it's just decoded directly (and there's no enum that matches the string found), it throws an error, which means the entire export data envelope doesn't get parsed. (That was the original cause of one of my starter bugs and it caused the "request my data" button to never be clickable.)

I didn't create tests for this initially because this file doesn't have any tests, but good call! Added now.

amy-at-kickstarter:
> > This LGTM, but I'd like to see a really quickie test just for a valid and an invalid enum value.
> > I'm also slightly surprised that the default implementation of `Decodable` for an `Enum` doesn't _already_ do this - what happened before you implemented this change?
> 
> If it's just decoded directly (and there's no enum that matches the string found), it throws an error, which means the entire export data envelope doesn't get parsed. (That was the original cause of one of my starter bugs and it caused the "request my data" button to never be clickable.)
> 
> I didn't create tests for this initially because this file doesn't have any tests, but good call! Added now.

Ah, gotcha, so this is replacing an unsafe/exploding init with a safe one. That seems fine for this use case!

amy-at-kickstarter (approved):
LGTM

PR #1904, opened 2023-12-11
MBL-1016: Implement API V1 request method in Combine

amy-at-kickstarter:
# 📲 What

This implements our most basic request method (used for API V1) in Combine. This is basically a line-for-line rewrite, replacing `SignalProducer` with the equivalent `Publisher`. 

# 🤔 Why

When we start rewriting more of the app in Combine, we will almost certainly run into situations where we don't have the time to port every API v1 call to GraphQL. While obviously this isn't _ideal_, it is a bit more realistic, and I think we'll be happier that the V1->GraphQL and ReactiveSwift->Combine projects won't be tightly coupled.

Thus, we have this funny little set of changes.

# ⏰ TODO

- [x] I need to add some tests to this. I have to double-check if we have tests for these specific sub-methods, or if I should just implement an actual V1 request and test it at a higher level.


KsApi/Service+RequestHelpers.swift
```diff
@@ -24,6 +25,34 @@ extension Service {
...
+
+      switch result {
+      case let .success(value):
+        return value
+      case let .failure(error):
+        throw error
+      }
+    }.compactMap { $0 }
+      .mapError { error in
+        error as! ErrorEnvelope
```
amy-at-kickstarter, line 52: `self.decodeModel` _only_ outputs `ErrorEnvelope` so this forced cast should be safe.

KsApi/MockService.swift
```diff
@@ -1277,6 +1277,10 @@
         .fetchWithResult(query: fetchProjectCommentsQuery, result: projectOnlyResult)
     }
 
+    internal func fetchProject_combine(project _: Project) -> AnyPublisher<Project, ErrorEnvelope> {
+      fatalError("Sorry, this is unimplemented!")
```
amy-at-kickstarter, line 1281: This test method will never be used, so it's just here for protocol conformance.

KsApi/Service+RequestHelpers.swift
```diff
@@ -24,6 +25,35 @@ extension Service {
     .flatMap(self.decodeModel)
   }
 
+  func request<M: Decodable>(_ route: Route) -> AnyPublisher<M, ErrorEnvelope> {
```
amy-at-kickstarter, line 28: This is a fairly faithful line-for-line request of the same `request` method above.

KsApi/ServiceTests.swift
```diff
@@ -120,55 +120,114 @@ final class ServiceTests: XCTestCase {
     XCTAssertNil(result)
   }
 
-  func test_Array_SwiftDecodeModel_ValidModel() {
```
amy-at-kickstarter, line 123: Github is displaying this weirdly because I deleted an old test and added a new test.

KsApi/ServiceTests.swift
```diff
@@ -120,55 +120,114 @@ final class ServiceTests: XCTestCase {
...
+    }
   }
+
+  /*
+   These two tests shouldn't actually *run* in our testing environment, because we don't want to make real network requests
+   in our real tests. But it's a useful little utility for comparing the Combine and ReactiveSwift version of the
+   API V1 call Request.
+   */
+
+  /*
```
amy-at-kickstarter, line 180: I commented these tests out, but ran them locally to ensure that they passed. We don't have a way to mock HTTP at this low of a level - hence these tests that actually _make a real network request_. Obviously something we don't want to do on CircleCI, since it would be a flaky mess.

We could consider adding something like `OHHTTPStubs` for low-level HTTP mocking, or one of those `NSURLProtocol` implementations that are floating around. But this just felt out of scope for this PR. Hence, my compromise option - tests you can turn on yourself if you need them 😉

KsApi/extensions/NSURLSession.swift
```diff
@@ -60,18 +61,39 @@ internal extension URLSession {
       }
   }
 
-  // Converts an URLSessionTask into a signal producer of raw JSON data. If the JSON does not parse
-  // successfully, an `ErrorEnvelope.errorJSONCouldNotParse()` error is emitted.
-  func rac_JSONResponse(_ request: URLRequest, uploading file: (url: URL, name: String)? = nil)
```
amy-at-kickstarter, line 65: Another unused function, deleted.

KsApi/extensions/NSURLSession.swift
```diff
@@ -108,9 +130,8 @@ private let defaultSessionError =
 private let boundary = "k1ck574r73r154c0mp4ny"
 
 extension URLSession {
-  // Returns a producer that will execute the given upload once for each invocation of start().
-  fileprivate func rac_dataWithRequest(_ request: URLRequest, uploading file: URL, named name: String)
-    -> SignalProducer<(Data, URLResponse), Error> {
+  fileprivate func addFileUploadToRequest(_ request: URLRequest, uploading file: URL,
```
amy-at-kickstarter, line 133: A little bit of refactoring in here.

KsApi/extensions/NSURLSession.swift
```diff
@@ -23,7 +24,7 @@ internal extension URLSession {
   func rac_dataResponse(_ request: URLRequest, uploading file: (url: URL, name: String)? = nil,
                         and error: ErrorHandler? = nil)
     -> SignalProducer<Data, ErrorEnvelope> {
-    let producer = file.map { self.rac_dataWithRequest(request, uploading: $0, named: $1) }
+    let producer = file.map { self.rac_requestWithFileUpload(request, uploading: $0, named: $1) }
```
amy-at-kickstarter, line 27: Could use another eye here - pretty sure what this is doing originally was calling one method if `file` exists, and another if it doesn't, but it's confusing to me that it's via `file.map`.

KsApi/Service+DecodeHelpers.swift
```diff
@@ -22,25 +22,18 @@ extension Service {
...
-      }
-  }
-
   func decodeModel<T: Decodable>(data json: Data) ->
     SignalProducer<T?, ErrorEnvelope> {
     return SignalProducer(value: json)
       .map { json in try? JSONDecoder().decode(T.self, from: json) }
   }
+
+  func decodeModel<T: Decodable>(data json: Data, ofType type: T.Type) -> Result<T?, ErrorEnvelope> {
```
ifosli, line 31: optional: I wonder if we should name these `decodeModelToResult` and/or `decodeModelToSignalProducer`. It'd add some extra overhead, especially if we switch all current usages (or replace all future usages of `decodeModelToResult` once that's the only one in use), but it would add a little clarity.

KsApi/extensions/NSURLSession.swift
ifosli: nit: I find this name confusing; since it starts with a verb, I expected it to act on self instead of returning a thing. What do you think of `requestWithFileUpload` or something like that instead?

ifosli (approved):

KsApi/Service+DecodeHelpers.swift
amy-at-kickstarter: Oh, I like that suggestion. Lemmee throw that in the auto-refactorer.

PR #1905, opened 2023-12-13
[MBL-1095] Disable Message Reply Button When User Is Blocked

scottkicks:
# 📲 What

Disables the Reply bar button in message threads if the message participant has been blocked. 

If we don't think this should go out with 5.11.0 I can avoid cutting a new build and this can go out in the next release. I think we can get it in though.

# 🤔 Why

We'll be stopping communication between users when one has been blocked.

# 🛠 How

MessagesViewModel has a `replyButtonIsEnabled` signal that I've updated to emit `true` only when there are messages _and_ the participant isn't blocked.

# ✅ Acceptance criteria

- [ ] If you've blocked someone, you should not be able to tap the reply button from that message thread
- [ ] If you haven't blocked someone, you can still reply to them

### DIFF
--- a/Kickstarter-iOS/Features/Messages/Storyboard/Messages.storyboard
+++ b/Kickstarter-iOS/Features/Messages/Storyboard/Messages.storyboard
@@ -506,7 +506,7 @@
                         </connections>
                     </tableView>
                     <navigationItem key="navigationItem" id="pUe-X9-gXs">
-                        <barButtonItem key="rightBarButtonItem" id="mFf-pV-0Nc">
+                        <barButtonItem key="rightBarButtonItem" enabled="NO" id="mFf-pV-0Nc">
                             <connections>
                                 <action selector="replyButtonPressed" destination="KOm-AI-SFW" id="Cxt-yR-ytl"/>
                             </connections>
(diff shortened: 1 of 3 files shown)


Library/ViewModels/MessagesViewModel.swift
```diff
@@ -168,11 +168,18 @@ public final class MessagesViewModel: MessagesViewModelType, MessagesViewModelIn
...
+
+    self.replyButtonIsEnabled = Signal.combineLatest(
       self.viewDidLoadProperty.signal.mapConst(false),
       self.messages.map { !$0.isEmpty },
-      participant.signal.map { $0.isBlocked == false }
+      self.participantPreviouslyBlocked
     )
+    .map { _, messages, isBlocked in
+      messages && !isBlocked
+    }
```
scottkicks, line 182: I don't really know why `line:176` is necessary tbh

Kickstarter-iOS/Features/Messages/Storyboard/Messages.storyboard
scottkicks, line 1: I noticed that there is a split second where users can tap Reply while we're pulling message and project data, which we use to determine whether to disable or not. Defaulting the reply button's enabled prop to false instead so users can't do that.

scottkicks, line 1: There is a split second where users can tap Reply while we're pulling messages and project data. Defaulting the reply button's enabled prop to false instead, so users can't do that.

Library/ViewModels/MessagesViewModel.swift
```diff
@@ -168,11 +168,17 @@ public final class MessagesViewModel: MessagesViewModelType, MessagesViewModelIn
       }
     )
 
-    self.replyButtonIsEnabled = Signal.merge(
-      self.viewDidLoadProperty.signal.mapConst(false),
+    self.participantPreviouslyBlocked = self.project
+      .map { $0.creator.isBlocked }
```
ifosli, line 172: Nit, kind of out of scope: It feels risky to me to check creator here instead of participant, since theoretically the creator could be the person doing the messaging. Right now, I don't think that's possible (since those messages don't show in the inbox on mobile), but that could (and maybe should) change in the future.

ifosli (approved):
Nice! I added a comment about the `participantPreviouslyBlocked` signal, which I think should be fixed, but I don't mind if that happens in a later PR instead.

PR #1906, opened 2023-12-14
Update deeplinks

ifosli:
# 📲 What

Delete unused deeplink subdomains, update tests, and add support for `kickstarter.com` without the `www.`.

# 🤔 Why

Becca and Ali confirmed in slack that the only email redirect domains we need are `clicks.` and `emails.`.

Note: If anyone tries to test this locally, `www.kickstarter.com` and `kickstarter.com` deeplinks might not work today. This is expected; it takes the `apple-app-site-association` file info about 24 hours to propagate through apple's servers. In the meanwhile, this can be tested by updating the entitlement to `applinks:www.kickstarter.com?mode=developer`

# ♿️ Accessibility 

- [ ] Works with VoiceOver

# ✅ Acceptance criteria

- [ ] staging.kickstarter.com deeplinks work for the beta version of the app
- [ ] www.kickstarter.com and kickstarter.com deeplinks work for the release version of the app
- [ ] email deeplinks (emails. and clicks.) work for the release version of the app

# ⏰ TODO

- [ ] Double-check that deeplinks from Braze are not affected
- [ ] Test non-email deeplinks once the app site association file has time to propagate.

### DIFF
--- a/Kickstarter-iOS/Alpha.entitlements
+++ b/Kickstarter-iOS/Alpha.entitlements
@@ -1,26 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-  <dict>
-    <key>aps-environment</key>
-    <string>production</string>
-    <key>com.apple.developer.associated-domains</key>
-    <array>
-      <string>applinks:www.kickstarter.com</string>
-      <string>applinks:native.dev.kickstarter.com</string>
-      <string>applinks:staging.kickstarter.com</string>
-      <string>applinks:click.e.kickstarter.com</string>
-      <string>applinks:click.em.kickstarter.com</string>
-      <string>applinks:email.kickstarter.com</string>
-      <string>applinks:emails.kickstarter.com</string>
-      <string>applinks:e2.kickstarter.com</string>
-      <string>applinks:e3.kickstarter.com</string>
-      <string>applinks:clicks.kickstarter.com</string>
-      <string>applinks:ea.kickstarter.com</string>
-      <string>applinks:me.kickstarter.com</string>
-      <string>webcredentials:kickstarter.com</string>
-    </array>
-    <key>com.apple.developer.ubiquity-kvstore-identifier</key>
-    <string>$(TeamIdentifierPrefix)$(CFBundleIdentifier)</string>
-  </dict>
+<dict>
+	<key>aps-environment</key>
+	<string>production</string>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:www.kickstarter.com</string>
+		<string>applinks:kickstarter.com</string>
+		<string>applinks:emails.kickstarter.com</string>
+		<string>applinks:clicks.kickstarter.com</string>
+		<string>webcredentials:kickstarter.com</string>
+	</array>
+	<key>com.apple.developer.ubiquity-kvstore-identifier</key>
+	<string>$(TeamIdentifierPrefix)$(CFBundleIdentifier)</string>
+</dict>
 </plist>
--- a/Kickstarter-iOS/Beta.entitlements
+++ b/Kickstarter-iOS/Beta.entitlements
@@ -1,26 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-  <dict>
-    <key>aps-environment</key>
-    <string>production</string>
-    <key>com.apple.developer.associated-domains</key>
-    <array>
-      <string>applinks:www.kickstarter.com</string>
-      <string>applinks:native.dev.kickstarter.com</string>
-      <string>applinks:staging.kickstarter.com</string>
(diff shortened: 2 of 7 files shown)


Kickstarter-iOS/Alpha.entitlements
ifosli, line 1: I'm not sure why xcode wanted to change the spacing in these files, so I decided to let it. If y'all would rather I keep the original spacing, let me know, and I'll edit the actual files.

amy-at-kickstarter: Eh, doesn't bother me. I say let Xcode do what it wants.

Kickstarter-iOS/Beta.entitlements
```diff
@@ -1,26 +1,15 @@
...
-    <key>com.apple.developer.ubiquity-kvstore-identifier</key>
-    <string>$(TeamIdentifierPrefix)$(CFBundleIdentifier)</string>
-  </dict>
+<dict>
+	<key>aps-environment</key>
+	<string>production</string>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:staging.kickstarter.com</string>
+		<string>webcredentials:kickstarter.com</string>
```
amy-at-kickstarter, line 10: Did you find out if it's possible for the beta app to also handle the same links as the production app? Or is it a 1:1 mapping only? It'd be nice to keep the environments equivalent (at least to prevent confusion when we're testing) but I also understand if that would make everything a complete and total mess. Or if Apple doesn't allow it.

Kickstarter-iOS/Release.entitlements
```diff
@@ -1,34 +1,28 @@
...
+	<array>
+		<string>Default</string>
+	</array>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:www.kickstarter.com</string>
+		<string>applinks:kickstarter.com</string>
+		<string>applinks:emails.kickstarter.com</string>
+		<string>applinks:clicks.kickstarter.com</string>
+		<string>webcredentials:kickstarter.com</string>
```
amy-at-kickstarter, line 17: Is this for Apple/Facebook login? Just curious; I noticed it's not in the unit tests or elsewhere in your code changes.

amy-at-kickstarter (approved):
Nice to see all this cleanup. My suggestion would be to add some documentation to `AppDelegateViewModel` or another deeplink related location in the code. (Either a comment linking to a Confluence page, or just tossing some notes inline.) I think it will be easy for us to forget _what_ these links do and when they were last audited, so it might help us from ending up in the same position in another year or two.

scottkicks (approved):
LGTM. I agree with Amy's suggestion for more documentation.

Should be good to go once that last todo is tested.

Kickstarter-iOS/Beta.entitlements
ifosli: So we could definitely do this, but it would require updating our apple-app-site-association files to include the release app id and the beta app id. (If there are multiple app ids provided, Apple will open the first app it finds a match for or maybe the one we specify first; the documentation is lacking and I think it depends on how the file is structured.) I'm a little hesitant to do that because these files are all publicly available and adding references to our Beta app to prod-based files seems weird to me. (But the email links app site association files already include the Alpha app, so maybe it doesn't matter...) Happy to revisit this if you think it'd be helpful!

Kickstarter-iOS/Release.entitlements
ifosli: The webcredentials? I'm not sure; I think it's a login thing but it has nothing to do with deeplinks (other than them both being entitlements) so I didn't touch it. It just looks like I did, since the spacing changed.

Kickstarter-iOS/Beta.entitlements
amy-at-kickstarter: Hmm yeah that's an interesting point! I'm not worried about it (I can't imagine anything you can do just by knowing the identifier of our beta app) - to pick one at random, it looks like Google includes enterprise and dev versions [in their site association](https://google.com/apple-app-site-association).

However, I also don't feel like fixing this is a huge priority - more of a nice-to-have so that the beta and release apps stay at parity. Maybe throw it in a low-priority ticket?

amy-at-kickstarter: (Mostly my thought on the ticket is - if someone from marketing comes to us and asks why it's broken, we can point at it and then maybe decide to prioritize it.)

ifosli:
> Nice to see all this cleanup. My only suggestion would be to add some documentation to `AppDelegateViewModel` or another deeplink related location in the code. (Either a comment linking to a Confluence page, or just tossing some notes inline.) I think it will be easy for us to forget _what_ these links do and when they were last audited, so it might help us from ending up in the same position in another year or two.

My plan is to update the guru cards we already have, but I added a couple extra comments, including a link to one of them.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -315,7 +316,8 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
 
     self.registerPushTokenInSegment = self.deviceTokenDataProperty.signal
 
-    // Deep links
+    // Deep links. For more information, see
+    // https://app.getguru.com/card/cyRdjqgi/How-iOS-Universal-Links-work
```
amy-at-kickstarter, line 320: 👍

Kickstarter-iOS/Beta.entitlements
ifosli: Good call; I created https://kickstarter.atlassian.net/browse/MBL-1107

PR #1907, opened 2023-12-19
MBL-1016: Stub version of paginate that works with Combine

amy-at-kickstarter:
# 📲 What

This is a draft - but I'm working on creating a Combine-compatible version of Paginate.

# 🤔 Why

We'd like free reign to write new code in SwiftUI + Combine. However, given that some large and important sections of our app still use API V1, it's likely we will (at some point) need to write API V1 functionality _in Combine_. This code is a big part of making that possible.

# 🛠 How

By stubbing out a Combine-compatible version of `paginate`, and re-implementing some of the `paginate` test suite to run against it. This will guarantee that the new version behaves the same as the old version.

# ⏰ TODO

This is merely a stub; it isn't the actual implementation.

### DIFF
--- a/KsApi/combine/CombineTestObserver.swift
+++ b/KsApi/combine/CombineTestObserver.swift
@@ -1,12 +1,16 @@
 import Combine
 import Foundation
+// TODO: do I need to move CombineTestObserver into another package, so this isn't imported into KsApi?
+import XCTest
 
-public final class CombineTestObserver<Value, Error: Swift.Error> {
+public final class CombineTestObserver<Value: Equatable, Error: Swift.Error> {
   public enum Event {
     case value(Value)
     case error(Error)
     case finished
   }
+
+  public init() {}
 
   public private(set) var events: [Event] = []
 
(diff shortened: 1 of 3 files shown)


KsApi/combine/CombineTestObserver.swift
```diff
@@ -1,13 +1,17 @@
 import Combine
 import Foundation
+// TODO: do I need to move CombineTestObserver into another package, so this isn't imported into KsApi?
```
amy-at-kickstarter, line 3: I forget whether or not Swift is smart enough not to link a framework added to a file, if that file isn't actually added to a target. We only use `CombineTestObserver` in test packages.

We may need to move this file into its own package.

Library/Paginate.swift
```diff
@@ -107,3 +108,38 @@ public func paginate<Cursor, Value: Equatable, Envelope, ErrorEnvelope, RequestP
...
+  concater _: @escaping (([Value], [Value]) -> [Value]) = (+)
+)
+  ->
+  (
+    paginatedValues: AnyPublisher<[Value], Never>,
+    isLoading: AnyPublisher<Bool, Never>,
+    pageCount: AnyPublisher<Int, Never>,
+    errors: AnyPublisher<ErrorEnvelope, Never>
+  ) {
+  let paginatedValues = Empty(completeImmediately: false, outputType: [Value].self, failureType: Never.self)
```
amy-at-kickstarter, line 130: Non-functioning stub results. Just getting the signature of the method correct.

Library/PaginateTests.swift
```diff
@@ -236,6 +240,116 @@ final class PaginateTests: TestCase {
...
+    let valuesTest = CombineTestObserver<[Int], Never>()
+    valuesTest.observe(values)
+
+    let loadingTest = CombineTestObserver<Bool, Never>()
+    loadingTest.observe(loading)
+
+    valuesTest.assertDidNotEmitValue("No values emit immediately.")
+    valuesTest.assertDidNotFinish()
+    loadingTest.assertDidNotEmitValue("No loading happens immediately.")
+    loadingTest.assertDidNotFinish()
```
amy-at-kickstarter, line 272: My goal here was to do this TDD-style. The original `paginate` method includes a lot of code that isn't 1:1 mappable to Combine, like `switchMap` and `Either`. I think the only way I'm going to be able to re-write it is to go feature-by-feature with the test, and make sure the behavior is the same. 

So the combine test is _literally_ copy-pasted from the original test; I've been adding methods to `CombineTestObserver` so that it can work in the same way.

PR #1908, opened 2024-01-05
[MBL-1031] User Blocking Analytics

scottkicks:
# 📲 What

Adding a new `trackBlockedUser` segment event around in our user blocking flow.
[Event definition doc](https://kickstarter.atlassian.net/wiki/spaces/IN/pages/2548301877/User+Blocking+Flow) from the insights team

I also included an unrelated property name update here that is just a cleanup.

# 🤔 Why

This flow doesn't have any analytic events, and we want to know how useful users find this feature and how often they use it.

# 🛠 How

I've added this new event in `KSRAnalytics`, as well as tests in `KSRAnalyticsTests`
It is then used in these view models:

1. ProjectPageViewModel
2. MessagesViewModel
3. CommentsViewModel & CommentRepliesViewModel

# 👀 See

No UI updates

# 🧪 TESTING

- [x] Check in the [Segment Debugger](https://app.segment.com/kickstarter/sources/ios/debugger) that these events are being received at the appropriate times. The debugger can be, well, buggy and slow. If you don't see events in Staging, try Prod, but remember that it can take some time for new events to come in.

### DIFF
--- a/Library/Tracking/KSRAnalytics.swift
+++ b/Library/Tracking/KSRAnalytics.swift
@@ -43,7 +43,7 @@ public final class KSRAnalytics {
     case onboarding // CategorySelectionViewController, CuratedProjectsViewController
     case pledgeAddNewCard = "pledge_add_new_card"
     case pledgeScreen = "pledge" // PledgeViewController
-    case projectPage = "project" // ProjectPageViewController
+    case project // ProjectPageViewController
     case profile // BackerDashboardProjectsViewController
     case rewards // RewardsViewController
     case search // SearchViewController
@@ -760,7 +760,7 @@ public final class KSRAnalytics {
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
       .withAllValuesFrom(videoProperties(videoLength: videoLength, videoPosition: videoPosition))
-      .withAllValuesFrom(contextProperties(page: .projectPage))
+      .withAllValuesFrom(contextProperties(page: .project))
 
     self.track(
       event: SegmentEvent.videoPlaybackStarted.rawValue,
@@ -808,7 +808,7 @@ public final class KSRAnalytics {
     project: Project
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
-      .withAllValuesFrom(contextProperties(page: .projectPage))
+      .withAllValuesFrom(contextProperties(page: .project))
 
     switch stateType {
     case .pledge:
@@ -1103,7 +1103,7 @@ public final class KSRAnalytics {
     sectionContext: KSRAnalytics.SectionContext
   ) {
     let props = projectProperties(from: project, loggedInUser: self.loggedInUser)
-      .withAllValuesFrom(contextProperties(page: .projectPage, sectionContext: sectionContext))
+      .withAllValuesFrom(contextProperties(page: .project, sectionContext: sectionContext))
 
     self.track(
       event: SegmentEvent.pageViewed.rawValue,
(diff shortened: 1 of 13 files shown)


Library/Tracking/KSRAnalytics.swift
```diff
@@ -1458,6 +1506,19 @@ private func contextProperties(
...
+
+private func interactionProperties(
+  targetUserId: String,
+  prefix: String = "interaction_"
+) -> [String: Any] {
+  var result: [String: Any] = [:]
+
+  result["target_uid"] = targetUserId
+
+  return result.prefixedKeys(prefix)
```
amy-at-kickstarter, line 1519: Why not just set `result["interaction_target_uid"] = targetUserId`?

Library/ViewModels/CommentRepliesViewModel.swift
```diff
@@ -265,6 +265,35 @@ public final class CommentRepliesViewModel: CommentRepliesViewModelType,
...
+            page: .project,
+            sectionContext: .comments,
+            typeContext: .initiate,
+            targetUserId: "\(blockedUserId)"
+          )
+      }
+
+    _ = self.blockUserProperty.signal
+      .combineLatest(with: project)
+      .takeWhen(blockUserEvent.values().ignoreValues())
```
amy-at-kickstarter, line 286: What does this line do? Just an interesting one to see `values().ignoreValues()`

amy-at-kickstarter (approved):
LGTM with two non-blocking questions/comments inline.

Library/ViewModels/CommentRepliesViewModel.swift
scottkicks: So we actually get an EmptyResponseEnvelope from the block user mutation or an error.  `blockUserEvent.values()` tells us that we got a response successfully and not an error. I'm using `.ignoreValues()` here since there isn't anything in the signal. 

Essentially, this block of code is saying, "React when `blockUserProperty` is triggered, combine its value with the late value in `project,` but only complete when the blocking mutation has returned a successful response."

Library/Tracking/KSRAnalytics.swift
scottkicks: We could do that, but I opted to use the same pattern we're using elsewhere just in case we want to add more Interaction specific properties moving forward.

Library/Tracking/KSRAnalytics.swift
```diff
@@ -598,6 +609,43 @@ public final class KSRAnalytics {
     )
   }
 
+  // MARK: - Block User Events
+
+  /**
+   Call when a user attempts or succeeds  to block a user.
```
ifosli, line 615: Nit: extra space and I'm a little unsure of what this comment means. Maybe "Call when a user attempts to block a user and when blocking succeeds" instead?

Library/ViewModels/CommentRepliesViewModelTests.swift
```diff
@@ -172,6 +172,93 @@ internal final class CommentRepliesViewModelTests: TestCase {
...
+
+      XCTAssertEqual(
+        segmentClient.properties(forKey: "session_user_is_logged_in", as: Bool.self),
+        [true, true]
+      )
+      XCTAssertEqual(segmentClient.properties(forKey: "user_uid", as: String.self), ["1", "1"])
+      XCTAssertEqual(segmentClient.properties(forKey: "context_cta"), ["block_user", "block_user"])
+      XCTAssertEqual(segmentClient.properties(forKey: "context_page"), ["project", "project"])
+      XCTAssertEqual(segmentClient.properties(forKey: "context_section"), ["comments", "comments"])
+      XCTAssertEqual(segmentClient.properties(forKey: "context_type"), ["initiate", "confirm"])
```
ifosli, line 257: Since this is checking for both an `initiate` and `confirm` event, do we need the previous test? It feels redundant to me, so unless it's doing something I haven't noticed, I think we should delete it (and rename this test to `testTrackUserBlockedFromComment_EventsEmitted`).

This applies to the other tests as well.

ifosli (approved):
Looks good apart from a couple minor things! I'm going to mark as approved since I don't feel the need to re-review but I really do think we should halve the number of tests.

PR #1909, opened 2024-01-08
[MBL-1093] Messages don't refresh when a user is blocked

scottkicks:
# 📲 What

Refreshes `MessageThreadsViewController` after successfully blocking a user within a message.

# 🤔 Why

Blocking a user through messages doesn’t update the UI until it’s manually refreshed. 

**Steps to reproduce**
1. Block user from messages
2. Get automatically navigated back to inbox
3. Click back in on message thread and see no difference
4. Go back to inbox, pull to refresh, click on message thread again. Notice that message thread now says user is blocked

# 🛠 How

In other parts of this flow we use the `.ksr_blockedUser` notification to update other parts of the app. For this case, the simpler approach is to use the existing refresh functionality in `MessageThreadsViewModel`.

- Added a callback, `var dismissedAfterBlockingUserHandler: (() -> Void)?`, in `MessagesViewController`
- When a user successfully blocks someone from a message, we currently auto-pop the VC. This new callback is triggered in `MessageThreadsViewController,` which calls its `viewModel.inputs.refresh()` and pulls in updated thread data.

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2024-01-08 at 09 54 56

# 🧪 TESTING

- [x] User feels like blocking works immediately (UI is updated the next time the thread is navigated to).

### DIFF
--- a/Kickstarter-iOS/Features/MessageThreads/Controller/MessageThreadsViewController.swift
+++ b/Kickstarter-iOS/Features/MessageThreads/Controller/MessageThreadsViewController.swift
@@ -86,7 +86,9 @@ internal final class MessageThreadsViewController: UITableViewController {
     if let messageThread = self.dataSource[indexPath] as? MessageThread {
       let vc = MessagesViewController.configuredWith(messageThread: messageThread)
 
-      vc.dismissedAfterBlockingUserHandler = { self.viewModel.inputs.refresh() }
+      vc.dismissedAfterBlockingUserHandler = { [weak self] in
+        self?.viewModel.inputs.refresh()
+      }
 
       self.navigationController?.pushViewController(vc, animated: true)
     }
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/MessageThreads/Controller/MessageThreadsViewController.swift
```diff
@@ -85,6 +85,9 @@ internal final class MessageThreadsViewController: UITableViewController {
   ) {
     if let messageThread = self.dataSource[indexPath] as? MessageThread {
       let vc = MessagesViewController.configuredWith(messageThread: messageThread)
+
+      vc.dismissedAfterBlockingUserHandler = { self.viewModel.inputs.refresh() }
```
ifosli, line 89: I'm concerned that this could cause a retain loop. The messagesViewController will now hold a strong reference to the messageThreadViewController. I don't think there's currently a strong reference in the opposite direction, but I'm not 100% sure how the navigation controller works, and it wouldn't surprise me if this gets refactored at some point in the future so that the messageThreadViewController does hold on to the messagesViewController. How would you feel about using weak self in the block or alternatively adding a comment that calls out why this is currently safe (or why it might not be safe in the future)?

scottkicks: It's a good point. I considered this while testing and couldn't see evidence of a retain cycle. As an extra precaution, I'll add a capture list and trigger this callback just before the child VC gets dismissed. That's a good call out.

ifosli (approved):

PR #1910, opened 2024-01-08
MBL-1128: Fix crash when trying to block a user while using an iPad

amy-at-kickstarter:
# 📲 What

Fix [this crash](https://console.firebase.google.com/u/0/project/kickstarter-ios/crashlytics/app/ios:com.kickstarter.kickstarter/issues/e328c5b10cd249143c0846f18587c92f?time=last-twenty-four-hours&versions=5.11.0%20(1702566)&types=crash&sessionEventKey=b69cc5f5cd214edc928e1bb5f75d22d9_1899936616456567754) found in 5.11.0.

To duplicate, try blocking a user from the Comments screen while you're using an iPad in Landscape mode.

# 🤔 Why

At least reading the crash log, it sounds like that when you present a `UIAlertController` on an iPad in a navigation controller, it _forces_ it to be presented as a popover. However, our check `IsIPad` checked the horizontal size class of the view controller. Because we present the comment view controllers in a modal screen - which _isn't the full width_ - the `IsIPad` check turned out `false`, even though we were still presenting as a popover. Hence, the crash.

### DIFF
--- a/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
+++ b/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
@@ -1012,8 +1012,8 @@ extension ProjectPageViewController: ProjectPamphletMainCellDelegate {
           self.presentBlockUserAlert(username: project.creator.name, userId: project.creator.id)
         },
         viewProfileHandler: { _ in self.goToCreatorProfile(forProject: project) },
-        sourceView: cell,
-        isIPad: self.traitCollection.horizontalSizeClass == .regular
+        sourceView: cell.creatorButton,
+        isIPad: self.traitCollection.userInterfaceIdiom == .pad
       )
 
     self.present(actionSheet, animated: true)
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
```diff
@@ -1012,8 +1012,8 @@ extension ProjectPageViewController: ProjectPamphletMainCellDelegate {
           self.presentBlockUserAlert(username: project.creator.name, userId: project.creator.id)
         },
         viewProfileHandler: { _ in self.goToCreatorProfile(forProject: project) },
-        sourceView: cell,
-        isIPad: self.traitCollection.horizontalSizeClass == .regular
+        sourceView: cell.creatorButton,
```
amy-at-kickstarter, line 1015: I had a slight nit that this popover was presented almost at the bottom of the screen, since the source view cell is fairly sizeable:
Screenshot 2024-01-08 at 2 01 26 PM

This makes it look like this, instead:
Screenshot 2024-01-08 at 2 14 01 PM

ifosli (approved):
Interesting! I still think the correct behavior is to not use a popover if iPad screen is small (whether that's how we present it or because you're seeing the app in split screen). But if that's not actually an option, then we should of course use a popover correctly instead of crashing. Thanks for figuring that out!

amy-at-kickstarter:
> Interesting! I still think the correct behavior is to not use a popover if iPad screen is small (whether that's how we present it or because you're seeing the app in split screen). But if that's not actually an option, then we should of course use a popover correctly instead of crashing. Thanks for figuring that out!

Agreed! Not clear why it's forcing the popover, but maybe it's a built-in iOS thing. Quick check on StackOverflow didn't say anything useful about it. 🤷‍♀️

scottkicks (approved):
Nice one! Thanks for jumping on this!

PR #1912, opened 2024-01-16
[MBL-996] Login Flow - Dark Mode

scottkicks:
# 📲 What

Login and Forgot Password screens support dark mode.

# 🤔 Why

Dark Mode is cool

# 🛠 How

We have a `DesignSystemColors` file that adaptively sets colors based on the user's device settings. The main difference between this and how we currently set color style is that it uses colors set in xcassets now.

Basically, if the Dark Mode feature flag is on, components will use the styling set in `DesignSystemColors` which dynamically chooses colors based on the user's dark mode device setting.

The naming of the new colors, defined in the [figma](https://www.figma.com/file/W1FTPNhWGrJGaWzYkVERtN/H1-2024%3A-Dark-Mode-for-Apps?type=design&node-id=15-524&mode=design&t=FEJnaohdJqlkMA9g-0) is subject to change. For now, I named them according to the naming that [Android uses](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSColors.kt), which is also what they're named on web.

# 👀 See

| After 🦋 |
| Simulator Screen Recording - iPhone 15 Pro - 2024-01-11 at 14 24 30 |

# ✅ Acceptance criteria

You'll need to turn the Dark Mode feature flag on _and_ remove the "Appearance" property from the info.plist to see Dark Mode colors.
- [x] Light and Dark mode versions of the Login and Forgot Password screens match the figma design.

# TODO
- [ ] update green color names once design has settled on naming


Kickstarter-iOS/Features/Login/Controller/LoginViewController.swift
```diff
@@ -70,17 +71,29 @@ internal final class LoginViewController: UIViewController {
...
   override func bindStyles() {
-    _ = self |> loginControllerStyle
+    if featureDarkModeEnabled() {
+      _ = self.navigationController?.navigationBar.tintColor = adaptiveColor(.iconPrimary)
+    }
+
+    _ = self
+      |> loginControllerStyle
+
+    if featureDarkModeEnabled() {
```
ifosli, line 81: I think this is going to be a pain to clean up and I'm not sure how much value the flag is adding when the app is in forced light mode. I think we should at least consider not flag guarding these changes; happy to discuss this during a meeting if that's easier!

Kickstarter-iOS/SharedViews/SharedStoryboards/Login.storyboard
```diff
@@ -657,6 +660,9 @@
         </scene>
     </scenes>
     <resources>
+        <namedColor name="backgroundPrimary">
+            <color red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+        </namedColor>
```
ifosli, line 665: I'm confused by the changes to this storyboard. Why are some colors changed and others left alone? Also, does the definition above sync with our backgroundPrimary colorset?

ifosli (changes requested):
Screencast mostly looks good, though a couple things: Disabled button looks like it's getting lighter when I would expect it to be darker/more transparent. And the eye icon is brighter than the other icons in dark mode (color looks like it matches in light mode but isn't updated for dark mode).

I'm excited that we're working on this!

Kickstarter-iOS/Features/Login/Controller/LoginViewController.swift
scottkicks: It should be easy enough to add this attribute to the email and password text fields later, but I hear you. We need to flag-guard these. Otherwise, the adaptive colors are applied immediately and the idea is to fully rely on xcasset colors only once we're ready to roll out the full integration.

Kickstarter-iOS/SharedViews/SharedStoryboards/Login.storyboard
scottkicks: So, the background set in the storyboard was overwriting the programmatic changes I was making. I couldn't find a way to feature flag colors being set directly in storyboards, and since the Light mode version is the same as what's currently being used across the app, I think this is ok.

scottkicks:
@ifosli good catch on the clear all text buttons. will update. Re: the disabled button, Alison hasn't finished defining the new green colors so those will have to be updated later.

scottkicks:
Getting pulled off of this in favor of Post Campaign Pledging. After discussing with the team, we'll be removing the DM feature flag so this will need some updating.

scottkicks:
I'm closing this for now after deciding we'd be better off without a Remote Config feature flag.

PR #1920, opened 2024-01-23
MBL-1017: Add features to CombineTestObserver

amy-at-kickstarter:
# 📲 What

This makes `CombineTestObserver` equivalent to `TestObserver`. Heck, I even just copy-pasted some code over, because it worked fine!

# 🤔 Why

When we write new view models in Combine, we'll want to test them. This means that, if we're porting code, we can re-use the same (or write very similar) tests.

### DIFF
--- a/KsApi/combine/CombineTestObserver.swift
+++ b/KsApi/combine/CombineTestObserver.swift
@@ -73,7 +73,7 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
 
     assert(
       errors.count <= 1,
-      "I'm pretty sure a Combine publisher can only ever emit one error. If this fails, we've learned something new today."
+      "I'm pretty sure a Combine publisher can only ever emit one error. If this fails, I've learned something new today."
     )
 
     return errors.last
--- a/KsApi/combine/CombineTestObserver.swift
+++ b/KsApi/combine/CombineTestObserver.swift
@@ -1,7 +1,13 @@
 import Combine
 import Foundation
+import XCTest
 
+/**
+ A wrapper around a subscription that saves all events to a public array so
+ that assertions can be made on a publisher's behavior.
+ */
 public final class CombineTestObserver<Value, Error: Swift.Error> {
+  /// Represents the state of an event in the publisher's timeline
   public enum Event {
     case value(Value)
     case error(Error)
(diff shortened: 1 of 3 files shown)


KsApi/combine/CombineTestObserver.swift
```diff
@@ -27,4 +33,161 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
...
+      case .error:
+        return true
+
+      default: break
+      }
+      return false
+    }
+  }
+
+  public func assertDidComplete(_ message: String = "Should have completed.",
```
amy-at-kickstarter, line 102: All these asserts are, for the most part, copy-pasted.

KsApi/combine/CombineTestObserver.swift
```diff
@@ -27,4 +33,161 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
...
+  }
+
+  /// `true` if a `.Finished` event has been emitted or a `.Failed` event has been ommitted
+  public var didComplete: Bool {
+    return self.events.contains { event in
+      switch event {
+      case .finished:
+        return true
+      case .error:
+        return true
```
amy-at-kickstarter, line 94: This is one difference. A _failed_  publisher in Combine is 'dead' and complete, whereas I believe that's not true in ReactiveSwift.

KsApi/combine/CombineTestObserver.swift
```diff
@@ -27,4 +33,161 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
...
+      case let .error(e):
+        errors.append(e)
+      default: break
+        // do nothing
+      }
+    }
+
+    assert(
+      errors.count <= 1,
+      "I'm pretty sure a Combine publisher can only ever emit one error. If this fails, we've learned something new today."
```
ifosli, line 76: Nit: Do we want this check in our codebase? I'm okay with it since I'm kind of curious, but it feels a bit unprofessional/informal.

KsApi/combine/CombineTestObserver.swift
```diff
@@ -27,4 +33,161 @@ public final class CombineTestObserver<Value, Error: Swift.Error> {
...
+  public func assertDidNotEmitValue(_ message: String = "Should not have emitted any values.",
+                                    file: StaticString = #file, line: UInt = #line) {
+    XCTAssertEqual(0, self.values.count, message, file: file, line: line)
+  }
+
+  public func assertDidTerminate(
+    _ message: String = "Should have terminated, i.e. completed/failed/interrupted.",
+    file: StaticString = #file, line: UInt = #line
+  ) {
+    XCTAssertTrue(self.didFail || self.didComplete, message, file: file, line: line)
```
ifosli, line 136: Nit: I don't think you need `self.didFail` here - if it failed `self.didComplete` will return true as well. Same for `didNotTerminate` below

ifosli (approved):
Looks good! I'm a little concerned about the different behaviors surrounding failures - for example, if we're waiting for updates from a database and we have a network error, we'd probably not want to stop subscribing to that database; we'd just want to wait for the next update. I'm guessing there are ways to handle that, but it might require more refactoring. Do you know if there is an obvious way to use combine if we want to handle errors differently? Either way, useful callout!

amy-at-kickstarter:
> Looks good! I'm a little concerned about the different behaviors surrounding failures - for example, if we're waiting for updates from a database and we have a network error, we'd probably not want to stop subscribing to that database; we'd just want to wait for the next update. I'm guessing there are ways to handle that, but it might require more refactoring. Do you know if there is an obvious way to use combine if we want to handle errors differently? Either way, useful callout!

Yeah, this is something I noticed early on. I added a helper method for this case specifically: https://github.com/kickstarter/ios-oss/blob/main/KsApi/Publisher%2BService.swift#L40

KsApi/combine/CombineTestObserver.swift
amy-at-kickstarter: Ha, I like throwing a little informality into code sometimes, but if you want me to make the language more neutral I definitely can.

KsApi/combine/CombineTestObserver.swift
amy-at-kickstarter: Legit! This is copy-pasted from the original `TestObserver` so I'm inclined to just leave it for parity.

KsApi/combine/CombineTestObserver.swift
ifosli: I'm used to an "avoid we" policy and I do have a slight preference for more neutral language, but I also think this is kind of fun and it's still clear who "we" is. So up to you.

amy-at-kickstarter: Interesting point on "we"! That's one I hadn't heard before.

PR #1921, opened 2024-01-30
  MBL-1157: Create PKCE code for code verifier and code challenge

amy-at-kickstarter:
# 📲 What

Added a utility for PKCE verification.

# 🤔 Why

PKCE is part of the native apps flow for oAuth - we create a secret (the code verifier) and a challenge (a hashed version) which gets sent to the server. This is used to verify that every step of the oAuth flow is legit. Later in the flow, the client has to prove that it was the one that generated the hash by providing the original secret.

For more, see: https://www.oauth.com/oauth2-servers/mobile-and-native-apps/authorization/

# 🛠 How

I used [this example code](https://kickstarter.atlassian.net/browse/MBL-1168), as well as the documentation above.

# ⏰  TODO
I'd like to add more tests to this, which I'll coordinate with Isa on the Android side. These tests are just to make sure the code is running.

### DIFF
--- a/KsApi/PKCE.swift
+++ b/KsApi/PKCE.swift
@@ -28,7 +28,11 @@ public extension Data {
           throw PKCEError.UnexpectedRuntimeError
         }
 
-        _ = SecRandomCopyBytes(kSecRandomDefault, numBytes, address)
+        let result = SecRandomCopyBytes(kSecRandomDefault, numBytes, address)
+
+        if result != errSecSuccess {
+          throw PKCEError.UnexpectedRuntimeError
+        }
       }
     } catch {
       throw error
--- a/KsApi/PKCE.swift
+++ b/KsApi/PKCE.swift
@@ -1,0 +1,58 @@
+import CryptoKit
+import Foundation
+
+/// PKCE stands for Proof Key for Code Exchange.
+/// The code verifier, and its associated challenge, are used to ensure that an oAuth request is valid.
+/// See this documentation for more details: https://www.oauth.com/oauth2-servers/mobile-and-native-apps/authorization/
+public struct PKCE {
+  /// Creates a random alphanumeric string of the specified length
+  static func createCodeVerifier(ofLength length: Int) throws -> String {
+    let uppercase = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    let lowercase = "abcdefghijklmnopqrstuvwxyz"
+    let numbers = "0123456789"
+    let specials = "-._~"
+
+    let validCharacters = uppercase + lowercase + numbers + specials
+
+    var codeVerifier = ""
+
+    for _ in 0..<length {
+      guard let character = validCharacters.randomElement() else {
+        throw ErrorType.UnableToCreateCodeVerifier
+      }
+
+      codeVerifier += String(character)
+    }
+
+    return codeVerifier
+  }
+
+  /// Creates a base-64 encoded SHA256 hash of the given string.
+  static func createCodeChallenge(fromVerifier verifier: String) throws -> String {
+    guard let stringData = verifier.data(using: .utf8) else {
+      throw ErrorType.UnableToCreateCodeChallenge
+    }
+
+    let hash = SHA256.hash(data: stringData)
+    var hashData: Data?
+
+    hash.withUnsafeBytes { pointer in
+      let dataPointer = pointer.bindMemory(to: UInt8.self)
+      hashData = Data(buffer: dataPointer)
+    }
+
+    guard let unwrappedData = hashData else {
+      throw ErrorType.UnableToCreateCodeChallenge
+    }
+
+    return unwrappedData.base64EncodedString()
+  }
+
+  /// Errors that can occur in the PKCE flow.
+  /// Note that these are exceptional errors - effectively run time errors that we don't ever expect to happen.
+  /// However, because PKCE is a critical part of auth, it's worthwhile to be explicit if the unexpected ever *did* happen.
+  public enum ErrorType: Error {
+    case UnableToCreateCodeChallenge
+    case UnableToCreateCodeVerifier
+  }
+}
--- a/KsApi/PKCE.swift
+++ b/KsApi/PKCE.swift
@@ -1,39 +1,42 @@
 import CryptoKit
 import Foundation
+import Security
 
-/// PKCE stands for Proof Key for Code Exchange.
-/// The code verifier, and its associated challenge, are used to ensure that an oAuth request is valid.
-/// See this documentation for more details: https://www.oauth.com/oauth2-servers/mobile-and-native-apps/authorization/
-public struct PKCE {
-  /// Creates a random alphanumeric string of the specified length
-  static func createCodeVerifier(ofLength length: Int) throws -> String {
-    let uppercase = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-    let lowercase = "abcdefghijklmnopqrstuvwxyz"
-    let numbers = "0123456789"
-    let specials = "-._~"
+enum PKCEError: Error {
+  case UnexpectedRuntimeError
+}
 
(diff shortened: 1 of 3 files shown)


KsApi/PKCE.swift
```diff
@@ -0,0 +1,58 @@
+import CryptoKit
+import Foundation
+
+/// PKCE stands for Proof Key for Code Exchange.
+/// The code verifier, and its associated challenge, are used to ensure that an oAuth request is valid.
+/// See this documentation for more details: https://www.oauth.com/oauth2-servers/mobile-and-native-apps/authorization/
+public struct PKCE {
```
amy-at-kickstarter, line 7: This doesn't need to be a struct; I'm just doing it for the sake of namespace clarity. These methods are called like `PKCE.createCodeVerifier(ofLength:)`.

KsApi/PKCE.swift
```diff
@@ -0,0 +1,58 @@
...
+    let hash = SHA256.hash(data: stringData)
+    var hashData: Data?
+
+    hash.withUnsafeBytes { pointer in
+      let dataPointer = pointer.bindMemory(to: UInt8.self)
+      hashData = Data(buffer: dataPointer)
+    }
+
+    guard let unwrappedData = hashData else {
+      throw ErrorType.UnableToCreateCodeChallenge
```
amy-at-kickstarter, line 45: Somewhat annoyingly - there are a few cases in this code where I need to unwrap a `nil` coming from a system library. In the interest of making this bulletproof, I'm throwing errors if those unwraps fail - but I never expect that to happen, pragmatically. 

If any part of PKCE fails, I think the correct solution is to abort the login process, and just let the user continue in a logged-out state.

KsApi/PKCETest.swift
```diff
@@ -0,0 +1,20 @@
+@testable import KsApi
+import XCTest
+
+final class PKCETest: XCTestCase {
```
amy-at-kickstarter, line 4: Some rough tests - we'll coordinate more examples later, to make sure iOS + Android have the same behavior.

KsApi/PKCE.swift
```diff
@@ -0,0 +1,58 @@
...
+    let lowercase = "abcdefghijklmnopqrstuvwxyz"
+    let numbers = "0123456789"
+    let specials = "-._~"
+
+    let validCharacters = uppercase + lowercase + numbers + specials
+
+    var codeVerifier = ""
+
+    for _ in 0..<length {
+      guard let character = validCharacters.randomElement() else {
```
amy-at-kickstarter, line 20: @Arkariang So this is something I wanted to verify - `randomElement()` uses the system random number generator. Is that going to be _good enough_ for PKCE? The documentation says "cryptographically random" and it's not clear if it needs a more random kind of random than the system provides.

ifosli: I hesitate to rely on the default random generator for crypto - all my crypto classes in college taught "never implement crypto yourself if you can avoid it". However, I did a little digging and it looks fairly safe.

Reasoning: https://developer.apple.com/documentation/swift/systemrandomnumbergenerator claims that the system default random number generator is cryptographically secure whenever possible. https://developer.apple.com/documentation/security/randomization_services, which is cryptographically secure, uses the system default as its source of randomness (unsure what it does if the system default isn't good enough). `.randomElement` uses the system default as well.

The main problem with this approach is that I'm not sure what happens if the system default isn't secure enough (which could happen if there isn't a lot of entropy to draw from - ex new computers booted up at the same time will probably have the same seed and get the same "random" numbers). It's possible that Apple's cryptographically secure random bytes method would throw an error instead in this case, and that that's what makes it actually cryptographically random. However, I don't know enough about PKCE to know if this would actually be a problem we need to handle; I expect our real users would have cryptographically random verifiers.

I think I'd recommend leaving this as-is for now and going to security office hours next week. Or looking for maybe 3rd party solutions for getting guaranteed cryptographic randomness. (Or using https://developer.apple.com/documentation/security/1399291-secrandomcopybytes, but then we can't specify what kinds of bytes we're okay with, as far as I can tell.)

Arkariang: I think it is refered to use some kind of secure hash algorithm, in my case for Android I'll use SHA-256

ifosli (approved):

KsApi/PKCE.swift
```diff
@@ -0,0 +1,83 @@
+import CryptoKit
+import Foundation
+import Security
+
+enum PKCEError: Error {
+  case UnexpectedRuntimeError
+}
+
+public extension Data {
```
amy-at-kickstarter, line 9: These utilities are translated from the examples in OpenID: https://github.com/openid/AppAuth-iOS/blob/aea7b8acd8b3fc261b8a42c998de33d76851f30b/Source/AppAuthCore/OIDTokenUtilities.m#L31-L55

KsApi/PKCE.swift
```diff
@@ -0,0 +1,83 @@
...
+  mutating func fillWithRandomSecureBytes() throws {
+    do {
+      let numBytes = self.count
+      try self.withUnsafeMutableBytes { rawPointer in
+        let pointer = rawPointer.bindMemory(to: UInt8.self)
+        guard let address = pointer.baseAddress else {
+          throw PKCEError.UnexpectedRuntimeError
+        }
+
+        _ = SecRandomCopyBytes(kSecRandomDefault, numBytes, address)
```
amy-at-kickstarter, line 31: @ifosli Thanks for reminding me what on earth this method was called - I was looking for it and couldn't find it 😅

KsApi/PKCE.swift
```diff
@@ -0,0 +1,83 @@
...
+/// PKCE stands for Proof Key for Code Exchange.
+/// The code verifier, and its associated challenge, are used to ensure that an oAuth request is valid.
+/// See this documentation for more details: https://www.oauth.com/oauth2-servers/mobile-and-native-apps/authorization/
+public struct PKCE {
+  /// Creates a random alphanumeric string of the specified length
+  static func createCodeVerifier(byteLength length: Int) throws -> String {
+    do {
+      var buffer = Data(count: length)
+      try buffer.fillWithRandomSecureBytes()
+      return buffer.base64URLEncodedStringWithNoPadding()
```
amy-at-kickstarter, line 64: This is how the RFC for OAuth suggests doing it - you create a shorter, cryptographically random buffer, and use base64 URL encoding to turn it into the valid character set.

KsApi/PKCE.swift
```diff
@@ -0,0 +1,83 @@
...
+    let hash = SHA256.hash(data: self)
+    var hashData: Data?
+
+    hash.withUnsafeBytes { pointer in
+      let dataPointer = pointer.bindMemory(to: UInt8.self)
+      hashData = Data(buffer: dataPointer)
+    }
+
+    guard let unwrappedData = hashData else {
+      throw PKCEError.UnexpectedRuntimeError
```
amy-at-kickstarter, line 48: Alternatively, I could make these methods return `Data?` and handle it further down. Preferences?

KsApi/PKCE.swift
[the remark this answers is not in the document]
amy-at-kickstarter: Good one; fixed!

KsApi/PKCETest.swift
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  func testPKCECreateCodeVerifier_matchesRequirements() {
+    let verifier = try! PKCE.createCodeVerifier(byteLength: 32)
+    XCTAssertEqual(verifier.count, 43)
+
+    let uppercase = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    let lowercase = "abcdefghijklmnopqrstuvwxyz"
+    let numbers = "0123456789"
+    let specials = "-._~"
```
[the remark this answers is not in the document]
amy-at-kickstarter: Good question - the RFC mentioned that these were valid, but you're right, they're not part of base64url 🤷‍♀️

PR #1923, opened 2024-01-30
Move CombineTestObserver files to the test target

ifosli:
# 📲 What

Move combineTestObserver to all test targets and its corresponding test to the `KsApiTests` target. Otherwise, the app crashes on launch.

I think there's probably a better way to ensure `combineTestObserver` is accessible to all test targets without being accessible outside of tests, but I think that would involve creating a separate target just for it, and I'm not sure if that's worth it.

# ✅ Acceptance criteria

- [x] Circle ci tests still pass
- [x] App builds and doesn't crash on launch


amy-at-kickstarter:
I created a follow-up ticket to create a shared test target: https://kickstarter.atlassian.net/browse/MBL-1179

PR #1924, opened 2024-01-31
Clean up RemoteConfigFeature+Helpers and RemoteConfigFeatureFlagToolsViewModel

amy-at-kickstarter:
# 📲 What

Remove boilerplate from our feature flag code.

# 🤔 Why

1. I was about to add a new feature, and the amount of copy-paste was bothering me.
2. This particular copy-pasted code seemed really easy to make mistakes in. We have tests for it, but those tests were also easy to put mistakes in! 

# 👀 See

Note the test runs for the three commits below - I did the code cleanup _first_, waited for the tests to pass, and then did the test cleanup. Just to ensure I didn't break both the code and tests simultaneously.

### DIFF
--- a/Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/RemoteConfigFeatureFlagToolsViewControllerTests.swift
+++ b/Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/RemoteConfigFeatureFlagToolsViewControllerTests.swift
@@ -21,15 +21,10 @@ final class RemoteConfigFeatureFlagToolsViewControllerTests: TestCase {
 
   func testRemoteConfigFeatureFlagToolsViewController() {
     let mockRemoteConfigClient = MockRemoteConfigClient()
-      |> \.features .~ [
-        RemoteConfigFeature.blockUsersEnabled.rawValue: false,
-        RemoteConfigFeature.consentManagementDialogEnabled.rawValue: false,
-        RemoteConfigFeature.darkModeEnabled.rawValue: false,
-        RemoteConfigFeature.facebookLoginInterstitialEnabled
-          .rawValue: false,
-        RemoteConfigFeature.postCampaignPledgeEnabled.rawValue: false,
-        RemoteConfigFeature.reportThisProjectEnabled.rawValue: false
-      ]
+
+    for feature in RemoteConfigFeature.allCases {
+      mockRemoteConfigClient.features[feature.rawValue] = false
+    }
 
     withEnvironment(language: .en, mainBundle: MockBundle(), remoteConfigClient: mockRemoteConfigClient) {
       let controller = RemoteConfigFeatureFlagToolsViewController.instantiate()
--- a/Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
+++ b/Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
@@ -3,10 +3,8 @@ import Prelude
 import XCTest
 
 final class RemoteConfigFeatureHelpersTests: TestCase {
-  func assertFeatureIsFalse(
-    whenRemoteConfigFeatureIsFalse feature: RemoteConfigFeature,
-    checkFeatureFlag: () -> Bool
-  ) {
+  func assert(featureFlagIsFalse checkFeatureFlag: () -> Bool,
+              whenRemoteConfigFeatureIsFalse feature: RemoteConfigFeature) {
     let mockRemoteConfigClient = MockRemoteConfigClient()
       |> \.features .~ [feature.rawValue: false]
 
@@ -15,10 +13,8 @@ final class RemoteConfigFeatureHelpersTests: TestCase {
     }
   }
 
-  func assertFeatureIsTrue(
-    whenRemoteConfigFeatureIsTrue feature: RemoteConfigFeature,
-    checkFeatureFlag: () -> Bool
-  ) {
+  func assert(featureFlagIsTrue checkFeatureFlag: () -> Bool,
+              whenRemoteConfigFeatureIsTrue feature: RemoteConfigFeature) {
     let mockRemoteConfigClient = MockRemoteConfigClient()
       |> \.features .~ [feature.rawValue: true]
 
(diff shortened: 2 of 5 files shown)


Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/RemoteConfigFeatureFlagToolsViewControllerTests.swift
```diff
@@ -21,15 +21,10 @@ final class RemoteConfigFeatureFlagToolsViewControllerTests: TestCase {
...
-        RemoteConfigFeature.darkModeEnabled.rawValue: false,
-        RemoteConfigFeature.facebookLoginInterstitialEnabled
-          .rawValue: false,
-        RemoteConfigFeature.postCampaignPledgeEnabled.rawValue: false,
-        RemoteConfigFeature.reportThisProjectEnabled.rawValue: false
-      ]
+
+    for feature in RemoteConfigFeature.allCases {
+      mockRemoteConfigClient.features[feature.rawValue] = false
+    }
```
amy-at-kickstarter, line 27: You'll still need to re-record the snapshot when you add a new feature, but won't need to modify this code.

Library/TestHelpers/TestCase.swift
```diff
@@ -88,11 +88,11 @@ internal class TestCase: XCTestCase {
   }
 
   internal func preferredSimulatorCheck() {
-    let deviceName = ProcessInfo().environment["SIMULATOR_DEVICE_NAME"]
+    let deviceName = ProcessInfo().environment["SIMULATOR_VERSION_INFO"]
```
amy-at-kickstarter, line 91: @ifosli This was a funny catch. I always name my snapshot simulators something (this one is "New New Snapshot Simulator"). It looks like `SIMULATOR_DEVICE_NAME` would return "New New Snapshot Simulator" whereas `SIMULATOR_VERSION_INFO` is like `CoreSimulator 494.33 - Device: iPhone 8 Plus - Runtime: iOS 11.2 (15C107) - DeviceType: iPhone 8 Plus`.

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
```diff
@@ -3,111 +3,102 @@ import Prelude
 import XCTest
 
 final class RemoteConfigFeatureHelpersTests: TestCase {
-  func testBlockUsers_RemoteConfig_FeatureFlag_False() {
+  func assertFeatureIsFalse(_ featureFlag: Bool,
+                            whenRemoteConfigFeatureIsFalse feature: RemoteConfigFeature) {
```
amy-at-kickstarter, line 7: Normally I'm fine with boilerplate in tests, but in this case, I think the boilerplate was impeding clarity.

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
```diff
@@ -3,111 +3,102 @@ import Prelude
...
 
-  func testBlockUsers_RemoteConfig_FeatureFlag_True() {
+  func assertFeatureIsTrue(_ featureFlag: Bool, whenRemoteConfigFeatureIsTrue feature: RemoteConfigFeature) {
     let mockRemoteConfigClient = MockRemoteConfigClient()
-      |> \.features .~ [RemoteConfigFeature.blockUsersEnabled.rawValue: true]
+      |> \.features .~ [feature.rawValue: true]
 
     withEnvironment(remoteConfigClient: mockRemoteConfigClient) {
-      XCTAssertTrue(featureBlockUsersEnabled())
+      XCTAssertFalse(featureFlag)
```
ifosli, line 21: I don't think these tests are working - we should pass in the `featureFlag` as a function instead so that it can be evaluated after the remoteConfig has been updated. In particular, I think this test should `XCTAssertTrue(featureFlag)` instead, and the fact that it passes the way it's written is concerning.

Also, if we want less boilerplate, thoughts on combining these into one function? Ie a `testFeatureMatchesRemoteConfig` and have it test both the true and the false case? I'm okay with them staying separate, though!

Library/TestHelpers/TestCase.swift
ifosli: Makes sense! I didn't even know you could re-name simulators, so thanks for handling that case!

ifosli (changes requested):
Thanks for making this nicer! It looks good apart from one of the tests that I think is currently not actually testing anything.

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
amy-at-kickstarter: Hahaha, OK, excellent catch here - on both bugs 😆  I see what you mean. Let me take another stab at this.

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
```diff
@@ -3,111 +3,118 @@ import Prelude
...
+    whenRemoteConfigFeatureIsTrue feature: RemoteConfigFeature,
+    checkFeatureFlag: () -> Bool
+  ) {
     let mockRemoteConfigClient = MockRemoteConfigClient()
-      |> \.features .~ [RemoteConfigFeature.blockUsersEnabled.rawValue: true]
+      |> \.features .~ [feature.rawValue: true]
 
     withEnvironment(remoteConfigClient: mockRemoteConfigClient) {
-      XCTAssertTrue(featureBlockUsersEnabled())
+      XCTAssertTrue(checkFeatureFlag())
```
amy-at-kickstarter, line 26: 😅 Great catch @ifosli. Soon as I swapped this to `XCTAssertTrue` the tests started failing!

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
amy-at-kickstarter: As for combining the tests into one method - I just tried it out but honestly, all the closures make it look kind of ugly 😅 So just leaving them in separate test methods for now.

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
ifosli: Looks good now!

ifosli (approved):

Library/RemoteConfig/RemoteConfigFeature+HelpersTests.swift
```diff
@@ -3,102 +3,118 @@ import Prelude
...
-      XCTAssertFalse(featureFlag)
+      XCTAssertTrue(checkFeatureFlag())
     }
   }
 
   func testBlockUsers_RemoteConfig_FeatureFlag_False() {
-    self.assertFeatureIsFalse(featureBlockUsersEnabled(), whenRemoteConfigFeatureIsFalse: .blockUsersEnabled)
+    self.assertFeatureIsFalse(whenRemoteConfigFeatureIsFalse: .blockUsersEnabled) {
+      featureBlockUsersEnabled()
+    }
```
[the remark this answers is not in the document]
amy-at-kickstarter: Ah, you're correct. Will fix, that looks much nicer.

PR #1927, opened 2024-02-01
Update build schema to force you to select correct target when building tests

amy-at-kickstarter:
# 📲 What

Force you to select the correct build schema when running tests via Xcode.

# 🤔 Why

Say you have an arbitrary target selected, and you try and run some tests by hitting the play button. In this example, I'm trying to run a test in `Library-iOS-Tests`, and I have the `Kickstarter-iOS` target selected, because I was running the app.

Screenshot 2024-02-01 at 11 19 09 AM

You might be treated to this linker error. However, look closely - why is it trying and failing to build `KsApiTests`? 
Screenshot 2024-02-01 at 11 19 55 AM

This is odd - while `Library-iOS` depends on `KsApi`, `Library-iOS-Tests` does NOT depend on `KsApiTests`.

Looking into the dependency graphs, and it looks like the app is trying to build `Kickstarter-iOS`, and then as part of that, it _included_ `KsApiTests`. That is also odd. First, why is it building `Kickstarter-iOS` instead of just `Library-iOS` and `Library-iOS-Tests`? And for that matter, why would `Kickstarter-iOS` (the app) depend on `KsApiTests` (a test bundle)?

Screenshot 2024-02-01 at 11 20 46 AM

It turns out the answer was in the schema. `Kickstarter-iOS` was set up so that it was associated with _all_ of our tests, and it would try to run them (all!) if it was selected as the schema when you run any tests. As far as I can tell, this led to general Xcode and linker strangeness, until you either swapped targets or ran a full project clean.

Screenshot 2024-02-14 at 4 12 23 PM

The solution was to uncheck all those test boxes. Now, when you try to run a test in the same scenario, you get this handy pop-up that forces you to swap to the correct test target:

Screenshot 2024-02-01 at 11 21 59 AM

### DIFF
--- a/Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
+++ b/Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
@@ -25,7 +25,7 @@
       </PostActions>
       <BuildActionEntries>
          <BuildActionEntry
-            buildForTesting = "YES"
+            buildForTesting = "NO"
             buildForRunning = "YES"
             buildForProfiling = "YES"
             buildForArchiving = "YES"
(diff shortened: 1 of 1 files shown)


ifosli (approved):
Nice!

PR #1928, opened 2024-02-01
MBL-1158: Add feature flag for logging in with OAuth

amy-at-kickstarter:
# 📲 What

Add a new feature flag for our new OAuth login.

# 🤔 Why

We'll want to ramp this change up.


Library/RemoteConfig/RemoteConfigFeature.swift
```diff
@@ -7,6 +7,7 @@ public enum RemoteConfigFeature: String, CaseIterable {
   case facebookLoginInterstitialEnabled = "facebook_interstitial"
   case postCampaignPledgeEnabled = "post_campaign_pledge"
   case reportThisProjectEnabled = "report_this_project"
+  case loginWithOAuthEnabled = "ios_oauth"
```
amy-at-kickstarter, line 10: This is called `"android_oath"` on Android, so just keeping it parallel

ifosli (approved):
Looks good to me but I still don't know our remote config flag creation well, so feel free to wait for Scott to take a look if you'd prefer

scottkicks (approved):

PR #1930, opened 2024-02-01
Make RemoteConfigFeatureFlagToolsViewModelTests resistant to adding new features

amy-at-kickstarter:
# 📲 What

This is a follow up to my PR #1924, making one more test future-proof.

# 🤔 Why

Adding feature flags is unnecessarily difficult, requiring a lot of copy-paste. This is one more place we shouldn't have to paste a new flag.

### DIFF
--- a/Library/ViewModels/RemoteConfigFeatureFlagToolsViewModelTests.swift
+++ b/Library/ViewModels/RemoteConfigFeatureFlagToolsViewModelTests.swift
@@ -21,15 +21,10 @@ final class RemoteConfigFlagToolsViewModelTests: TestCase {
 
   func testReloadWithData_AllFeaturesEnabled() {
     let mockRemoteConfigClient = MockRemoteConfigClient()
-      |> \.features .~ [
-        RemoteConfigFeature.blockUsersEnabled.rawValue: true,
-        RemoteConfigFeature.consentManagementDialogEnabled.rawValue: true,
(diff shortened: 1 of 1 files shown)


Library/ViewModels/RemoteConfigFeatureFlagToolsViewModelTests.swift
```diff
@@ -45,14 +41,11 @@ final class RemoteConfigFlagToolsViewModelTests: TestCase {
...
-        RemoteConfigFeature.blockUsersEnabled.rawValue: false,
-        RemoteConfigFeature.consentManagementDialogEnabled.rawValue: true,
-        RemoteConfigFeature.darkModeEnabled.rawValue: false,
-        RemoteConfigFeature.facebookLoginInterstitialEnabled.rawValue: true,
-        RemoteConfigFeature.postCampaignPledgeEnabled.rawValue: false,
-        RemoteConfigFeature.reportThisProjectEnabled.rawValue: false
-      ]
+
+    for feature in RemoteConfigFeature.allCases {
+      let trueOrFalse = feature.rawValue.hashValue % 2 == 1
```
amy-at-kickstarter, line 46: This is so each flag will get a stable true or false value, but we don't have to keep adding them to this list.

ifosli (approved):
I like it! Thanks for simplifying things!

scottkicks (approved):

PR #1932, opened 2024-02-05
[MBL-1146] Update transcend privacy url to use current locale

ifosli:
# 📲 What

Update the url to use the language code for the current locale. We're doing this instead of a translatable string because this is how they construct urls on web, so they explicitly ask translators to not translate urls.

# 👀 See

[JIRA](https://kickstarter.atlassian.net/browse/MBL-1146)

# ✅ Acceptance criteria

- [x] Url (request my data or delete my account button) opens the modal in the app language for all the languages we support

### DIFF
--- a/Library/ViewModels/SettingsDeleteOrRequestCellViewModel.swift
+++ b/Library/ViewModels/SettingsDeleteOrRequestCellViewModel.swift
@@ -23,8 +23,8 @@ public final class SettingsDeleteOrRequestCellViewModel: SettingsDeleteOrRequest
   public init() {
     self.notifyDeleteAccountTapped = self.deleteAccountTappedProperty.signal
       .map {
-        // TODO(MBL-1063): Replace hardcoded URL with translatable string.
-        URL(string: "https://legal.kickstarter.com/policies/en/?modal=take-control")
+        let code = Locale.current.languageCode ?? "en"
+        return URL(string: "https://legal.kickstarter.com/policies/\(code)/?modal=take-control")
       }.skipNil()
   }
 


amy-at-kickstarter (changes requested):
This makes sense, but there's a gotcha - your phone can be in _any_ language, and we don't support every single language code on the Kickstarter web side. (For an example: Arabic is `ar` but `https://legal.kickstarter.com/policies/ar/` is a 404.) Can we either fix it on the client or server side so that this link always (at least) goes to a valid default page, instead of a 404?

I'm also just a little confused about the behavior here, has this web page been translated? Ones I expect to work - `es` or `de` - still show the page 99% in English.

ifosli:
> This makes sense, but there's a gotcha - your phone can be in _any_ language, and we don't support every single language code on the Kickstarter web side. (For an example: Arabic is `ar` but `https://legal.kickstarter.com/policies/ar/` is a 404.) Can we either fix it on the client or server side so that this link always (at least) goes to a valid default page, instead of a 404?
> 
> I'm also just a little confused about the behavior here, has this web page been translated? Ones I expect to work - `es` or `de` - still show the page 99% in English.

There's a reason I wanted to do this with translations instead... I did check what happens if your phone is in a language we don't support and the locale that apple gives us is the locale *of the app* not the phone default. And then I checked all the languages that our app supports and they all open the modal in the corresponding language.

The "take control" modal is translated into the different languages (ex: https://legal.kickstarter.com/policies/es/?modal=take-control) so even though the policies aren't translated, I still think it's worth using the localized url.

Interestingly enough, I also checked what happens if we provide a language code that isn't supported and the modal still opens (in english) - it's just the policy it renders on top of that doesn't show. This will not happen unless we change the languages we support, but it's not a worst-case scenario if it does.

amy-at-kickstarter:
> I did check what happens if your phone is in a language we don't support and the locale that apple gives us is the locale _of the app_ not the phone default. And then I checked all the languages that our app supports and they all open the modal in the corresponding language.

Aha! This isn't what I expected/remembered. Sounds fine to me then.

amy-at-kickstarter (approved):
Swapping to approved after discussion

PR #1935, opened 2024-02-06
MBL-1158: Set up correct URLs for OAuth authorization endpoint

amy-at-kickstarter:
# 📲 What

This sets up the actual authorization URLs we will be using for OAuth, instead of the previous placeholder.

# 🤔 Why

Now that the web change is live, we can start testing!


Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -402,14 +402,7 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
...
   fileprivate func pushLoginViewController() {
-    if featureLoginWithOAuthEnabled() {
-      let session = ASWebAuthenticationSession(
-        url: OAuth.authorizationURL(),
-        callbackURLScheme: OAuth.redirectScheme
-      ) { _, _ in
-        // TODO: MBL-1159: Get required information from the callback; call mutation to login.
-      }
-
+    if featureLoginWithOAuthEnabled(), let session = OAuth.createAuthorizationSession() {
```
amy-at-kickstarter, line 405: Moving this into the OAuth struct, I don't think we'll need any logic that lives in `LoginToutViewController`.

Library/OAuth.swift
```diff
@@ -1,12 +1,52 @@
...
+        url: url,
+        callbackURLScheme: OAuth.redirectScheme
+      ) { _, _ in
+        // TODO: MBL-1159: Exchange information in callback for credentials, then login.
+      }
+
+      return session
+
+    } catch {
+      // TODO: Is there a way we can log/monitor these errors?
```
amy-at-kickstarter, line 48: @scottkicks Do we have a place to log generic errors in iOS? In this case, I only expect this catch to happen if something _really odd_ occurs, i.e. we fail to create a URL or a hash.

ifosli (approved):
I have no idea how error logging works in our codebase unfortunately, but looks good for now!

Library/OAuth.swift
amy-at-kickstarter: Added this to our iOS sync agenda.

PR #1938, opened 2024-02-08
[MBL-1176] ValidateCheckout Query Support

scottkicks:
# 📲 What

Add support for the new ValidateCheckout query. 

# 🤔 Why

We'll need to validate late pledges just before calling the Stripe's [confirmPayment](https://stripe.dev/stripe-ios/stripe/documentation/stripe/stppaymenthandler/confirmpayment(_:with:completion:)) method. For more context checkout [the initial proposal](https://kickstarter.atlassian.net/wiki/spaces/PAYM/pages/2586443778/Validations+Approach+1+Separate+validations+from+Checkouts+Apply+service+client-side+confirmation#Mobile-proposal)

# 👀 See

No UI changes


KsApi/Service.swift
```diff
@@ -805,4 +805,24 @@ public struct Service: ServiceType {
         SignalProducer(value: EmptyResponseEnvelope())
       }
   }
+
+  /**
+   Post Campaign Pledge Validation.
+   - parameter checkoutId: The Checkout Id..
```
ifosli, line 811: nit: Extra period

ifosli (approved):

PR #1939, opened 2024-02-08
MBL-1016: Create Paginator for pagination in SwiftUI/Combine

amy-at-kickstarter:
# 📲 What

This adds a new class, `Paginator`, inspired by our existing `paginate` method. `Paginator` is designed to fill the same functionality, but in a way that integrates neatly with SwiftUI and Combine-based code.

# 🤔 Why

When we move over to SwiftUI and Combine, we'll need an easy way to page through long lists of API results.

# 🛠 How

Originally I wanted to create a 1-for-1 re-write of `paginate` in Combine. The nice thing about the original `paginate` method is that it's _stateless_, nothing but a bunch of pure functions. However, I spent a significant amount of time trying to duplicate it, and was unable to make any traction. Writing stateless functional code is significantly more challenging (to me at least), and I don't think the benefits of its elegance outweigh the difficult I had modifying and understanding it. Throw in a few more variables, and I think `Paginator` will be easier to modify and understand in the long run.

`paginate` came with a great deal of tests, which I used to understand its behavior, and that became my inspiration for writing `Paginator`.

Note that `paginate` includes some niche features - you can pass in a customer `concater` for example - that I chose not to implement. If we need those, we can implement them in the future. Most of those niche features are only used on one or two screens.

# 👀 See

I created `PaginationExampleView` and `PaginationExampleViewModel` as examples (and my own test bed!) for how `Paginator` could be used in the app. You can try it yourself by throwing in this somewhere in the app: `presentViewController(UIHostingController(rootView: PaginationExampleView(), animated: true))`

https://github.com/kickstarter/ios-oss/assets/146007185/14e06c76-3356-4b4a-8a4a-9b47b19e97c3

### DIFF
--- a/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
+++ b/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
@@ -14,8 +14,8 @@ private struct PaginationExampleProjectList: View {
   @Binding var showProgressView: Bool
   @Binding var statusText: String
 
-  let onRefresh: @Sendable() -> Void
-  let onDidShowProgressView: @Sendable() -> Void
+  let onRefresh: () -> Void
+  let onDidShowProgressView: () -> Void
 
   var body: some View {
     HStack {
@@ -52,17 +52,15 @@ public struct PaginationExampleView: View {
   @StateObject private var viewModel = PaginationExampleViewModel()
 
   public var body: some View {
-    let capturedViewModel = viewModel
-
     PaginationExampleProjectList(
       projectIdsAndTitles: $viewModel.projectIdsAndTitles,
       showProgressView: $viewModel.showProgressView,
       statusText: $viewModel.statusText,
       onRefresh: {
-        capturedViewModel.didRefresh()
+        viewModel.didRefresh()
       },
       onDidShowProgressView: {
-        capturedViewModel.didShowProgressView()
+        viewModel.didShowProgressView()
       }
     )
   }
--- a/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
+++ b/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
@@ -52,6 +52,9 @@ public struct PaginationExampleView: View {
   @StateObject private var viewModel = PaginationExampleViewModel()
 
   public var body: some View {
+    //Note that PaginationExampleProjectList is decoupled from the view model;
+    //all the information it needs is passed in via bindings.
+    //This makes it easy to write a preview!
     PaginationExampleProjectList(
       projectIdsAndTitles: $viewModel.projectIdsAndTitles,
       showProgressView: $viewModel.showProgressView,
--- a/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
+++ b/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
@@ -52,9 +52,9 @@ public struct PaginationExampleView: View {
   @StateObject private var viewModel = PaginationExampleViewModel()
 
   public var body: some View {
-    //Note that PaginationExampleProjectList is decoupled from the view model;
-    //all the information it needs is passed in via bindings.
-    //This makes it easy to write a preview!
+    // Note that PaginationExampleProjectList is decoupled from the view model;
+    // all the information it needs is passed in via bindings.
+    // This makes it easy to write a preview!
     PaginationExampleProjectList(
       projectIdsAndTitles: $viewModel.projectIdsAndTitles,
       showProgressView: $viewModel.showProgressView,
--- a/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
+++ b/Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
@@ -1,0 +1,82 @@
+import KsApi
+import SwiftUI
+
+private struct PaginationExampleProjectCell: View {
+  let title: String
+  var body: some View {
+    Text(title)
+      .padding(.all, 10)
+  }
+}
+
+private struct PaginationExampleProjectList: View {
+  @Binding var projectIdsAndTitles: [(Int, String)]
+  @Binding var showProgressView: Bool
+  @Binding var statusText: String
+
+  let onRefresh: @Sendable() -> Void
+  let onDidShowProgressView: @Sendable() -> Void
+
+  var body: some View {
+    HStack {
+      Spacer()
+      Text("👉 \(statusText)")
+      Spacer()
+    }
+    .padding(EdgeInsets(top: 20, leading: 0, bottom: 20, trailing: 0))
+    .background(Color.yellow)
+    List {
+      ForEach(projectIdsAndTitles, id: \.0) {
+        let title = $0.1
+        PaginationExampleProjectCell(title: $0.1)
+      }
+      if showProgressView {
+        HStack {
+          Spacer()
+          Text("Loading 😉")
+            .onAppear {
+              onDidShowProgressView()
+            }
+          Spacer()
+        }
+        .background(Color.yellow)
+      }
+    }
+    .refreshable {
+      onRefresh()
+    }
+  }
+}
+
+public struct PaginationExampleView: View {
+  @StateObject private var viewModel = PaginationExampleViewModel()
+
+  public var body: some View {
+    let capturedViewModel = viewModel
+
+    PaginationExampleProjectList(
+      projectIdsAndTitles: $viewModel.projectIdsAndTitles,
+      showProgressView: $viewModel.showProgressView,
+      statusText: $viewModel.statusText,
+      onRefresh: {
+        capturedViewModel.didRefresh()
+      },
+      onDidShowProgressView: {
+        capturedViewModel.didShowProgressView()
+      }
+    )
+  }
+}
+
+#Preview {
+  PaginationExampleProjectList(
+    projectIdsAndTitles: .constant([
+      (1, "Cool project one"),
+      (2, "Cool project two"),
+      (3, "Cool project three")
+    ]),
+    showProgressView: .constant(true),
+    statusText: .constant("Example status text"),
+    onRefresh: {}, onDidShowProgressView: {}
+  )
+}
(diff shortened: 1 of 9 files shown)


Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+    .refreshable {
+      onRefresh()
+    }
+  }
+}
+
+public struct PaginationExampleView: View {
+  @StateObject private var viewModel = PaginationExampleViewModel()
+
+  public var body: some View {
```
amy-at-kickstarter, line 54: Note that I very intentionally split up `PaginationExampleView` from `PaginationExampleProjectList`. `PaginationExampleProjectList` is _decoupled from the view model_, basically containing only display logic for simple types. This made it really easy to write the SwiftUI preview; instead of having to mock out an entire environment with a view model and stubbed networking on so on, I could just pass in some strings.

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+private struct PaginationExampleProjectCell: View {
+  let title: String
+  var body: some View {
+    Text(title)
+      .padding(.all, 10)
+  }
+}
+
+private struct PaginationExampleProjectList: View {
+  @Binding var projectIdsAndTitles: [(Int, String)]
```
amy-at-kickstarter, line 13: See my note below - the actual display view is just taking simple types. This _could_ take a list of `Project`s, and that's what I did originally, but it got a little annoying - I had to use `Project.template`, and that's marked as private to `KsApi`. So I chose to just keep it to strings-and-ints for now.

Library/Paginator.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    self.values = []
+    self.isLoading = true
+    self.error = nil
+
+    let request = self.requestFromParams(params)
+    self.handleRequest(request)
+  }
+
+  public func requestNextPage() {
```
amy-at-kickstarter, line 113: I find this kind of declarative implementation much easier to grok than the original functional implementation of `paginate`.

Library/PaginatorTests.swift
```diff
@@ -0,0 +1,315 @@
...
+}
+
+struct ConcreteError: Error {}
+
+final class PaginatorTests: XCTestCase {
+  let valuesFromEnvelope: (TestEnvelope) -> [Int] = { $0.values }
+  let cursorFromEnvelope: (TestEnvelope) -> Int? = { $0.cursor }
+
+  func waitTinyInterval() {
+    _ = XCTWaiter.wait(for: [expectation(description: "Wait a tiny interval of time.")], timeout: 0.05)
```
amy-at-kickstarter, line 21: Easy/cheaters' method for testing asynchronous code 🙈

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+    .background(Color.yellow)
+    List {
+      ForEach(projectIdsAndTitles, id: \.0) {
+        let title = $0.1
+        PaginationExampleProjectCell(title: $0.1)
+      }
+      if showProgressView {
+        HStack {
+          Spacer()
+          Text("Loading 😉")
```
amy-at-kickstarter, line 36: I noticed a weird bug (?) where `ProgressView` sometimes disappeared while scrolling. Out of scope to figure out why, so you get a winky face instead of a spinner.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+  let onDidShowProgressView: @Sendable() -> Void
+
+  var body: some View {
+    HStack {
+      Spacer()
+      Text("👉 \(statusText)")
+      Spacer()
+    }
+    .padding(EdgeInsets(top: 20, leading: 0, bottom: 20, trailing: 0))
+    .background(Color.yellow)
```
scottkicks, line 27: nit: you can remove `Color` and just use `.yellow`.

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+    List {
+      ForEach(projectIdsAndTitles, id: \.0) {
+        let title = $0.1
+        PaginationExampleProjectCell(title: $0.1)
```
scottkicks, line 31: looks like we can remove this line, or use it as the title argument in `PaginationExampleProjectCell`
```suggestion
        PaginationExampleProjectCell(title: $0.1)
```

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+private struct PaginationExampleProjectList: View {
+  @Binding var projectIdsAndTitles: [(Int, String)]
+  @Binding var showProgressView: Bool
+  @Binding var statusText: String
+
+  let onRefresh: @Sendable() -> Void
+  let onDidShowProgressView: @Sendable() -> Void
```
scottkicks, line 18: @Sendable allows for these methods to run concurrently, yes?

KsApi/Service+RequestHelpers.swift
```diff
@@ -64,4 +51,35 @@ extension Service {
       .rac_dataResponse(preparedRequest(forURL: paginationUrl), and: self.perimeterXClient)
       .flatMap(self.decodeModelToSignal)
   }
+
+  func requestPaginationDecodable<M: Decodable>(_ paginationUrl: String)
```
scottkicks, line 55: What does `M` represent here?

Library/Paginator.swift
```diff
@@ -0,0 +1,136 @@
...
+ * `RequestParams`: The type that allows us to make a request for values without a cursor.
+
+ - parameter valuesFromEnvelope:   A function to get an array of values from the results envelope.
+ - parameter cursorFromEnvelope:   A function to get the cursor for the next page from a results envelope.
+ - parameter requestFromParams:    A function to get a request for values from a params value.
+ - parameter requestFromCursor:    A function to get a request for values from a cursor value.
+
+  You can observe the results of `values`, `isLoading`, `error` and `state` to access the loaded data.
+
+ */
```
scottkicks, line 27: Awesome documentation!

Library/Paginator.swift
```diff
@@ -0,0 +1,136 @@
...
+      .assertNoFailure()
+      .handleEvents(receiveCompletion: { [weak self] _ in
+        self?.isLoading = false
+      }, receiveCancel: { [weak self] in
+        self?.isLoading = false
+      })
+      .sink(receiveValue: { [weak self] envelope in
+        guard let self = self else {
+          return
+        }
```
scottkicks, line 83: nit: latest swift simplifies this guard. just fyi 😄 

```suggestion
        guard let self else { return }
```

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
```diff
@@ -0,0 +1,82 @@
...
+      onRefresh()
+    }
+  }
+}
+
+public struct PaginationExampleView: View {
+  @StateObject private var viewModel = PaginationExampleViewModel()
+
+  public var body: some View {
+    let capturedViewModel = viewModel
```
scottkicks, line 55: Could you explain why we need to capture the viewModel here instead of using the @StateObject var directly in ` PaginationExampleProjectList`?

scottkicks (commented):
Awesome work! I just have some questions that are mostly due to my lack of swiftui/combine knowledge

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
amy-at-kickstarter: I initially put this tag here because Xcode insisted that I did (IIRC, something like these properties were implicitly `@Sendable` and I needed to make them explicit). I've never used `@Sendable` before, so I took a peek in the docs, and you are correct that it allows a function to be used "across concurrency domains." My guess is that here, pragmatically, it's enforcing something at compile time that's important - perhaps that every captured variable included in `onRefresh` and `onDidShowProgressView` won't be de-allocated? But we should probably learn more if this comes up frequently.

KsApi/Service+RequestHelpers.swift
amy-at-kickstarter: `M` is the generic type placeholder for the model that will be decoded, for example, a `Project` or a `Backing`.

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
amy-at-kickstarter: It's a side-effect of these two closures being marked `@Sendable`. Without the capture, you get useful, but not terribly clear error:

```Main actor-isolated property 'viewModel' can not be referenced from a Sendable closure```

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
amy-at-kickstarter: Huh, y'know what? It compiles fine without it. Maybe this was an artifact of an earlier version of this code. I'm going to clean this up.

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
amy-at-kickstarter: Fixed!

Kickstarter-iOS/Features/PaginationExample/PaginationExampleView.swift
amy-at-kickstarter: Fixed, see note above about removing `@Sendable`

Library/Paginator.swift
amy-at-kickstarter: I take very little credit, this is largely copy-pasted from the original `paginate` 🤫

KsApi/Service+RequestHelpers.swift
scottkicks: Got it. I figured! I would prefer Model here but totally up to you. Just a nit 🙂

scottkicks (approved):
👍🏻 just some failing ci tests to work out

PR #1940, opened 2024-02-12
Add a nonfatal error; delete another nonfatal error

amy-at-kickstarter:
# 📲 What

1. Stop logging remote configuration errors to Crashlytics when a user boots the app with no connectivity.
2. Log when OAuth setup fails.

# 🤔 Why

This is to take better advantage of our non-fatal error logging in Crashlytics. Removing a noisy error, adding one I care about.

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -416,9 +416,16 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         return
       }
 
-      print("🔴 Remote Config SDK Activation Failed with Error: \(remoteConfigActivationError.localizedDescription)")
+      let errorAsNSError = remoteConfigActivationError as NSError
 
-      Crashlytics.crashlytics().record(error: remoteConfigActivationError)
+      if errorAsNSError.domain == RemoteConfigErrorDomain,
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -416,9 +416,16 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         return
       }
 
-      print("🔴 Remote Config SDK Activation Failed with Error: \(remoteConfigActivationError.localizedDescription)")
+      let errorAsNSError = remoteConfigActivationError as NSError
 
-      Crashlytics.crashlytics().record(error: remoteConfigActivationError)
+      if errorAsNSError.domain == RemoteConfigErrorDomain,
+        errorAsNSError.code == RemoteConfigError.internalError.rawValue {
```
amy-at-kickstarter, line 422: Dug a bit into the Firebase code - this is the correct domain and error for what they're sending us (error `8003`). The naming is a bit confusing, so it's possible we'll accidentally catch an error we may have been interested in.

amy-at-kickstarter: cc @ifosli since you hated this error!

scottkicks (approved):

PR #1941, opened 2024-02-13
Update GraphQL schema and CreatePaymentIntentMutation to match

amy-at-kickstarter:
# 📲 What

Update the GraphQL schema to the latest version.

# 🤔 Why

This fixes some build errors that were occurring due to a schema mis-match.


scottkicks (approved):
👍  thanks for updating the amountDollars -> amount change too

scottkicks (approved):
👍  thanks for updating the amountDollars -> amount change too

PR #1942, opened 2024-02-13
Explicitly add Prelude to Link Libraries phase of Kickstarter-Framework

amy-at-kickstarter:
# 📲 What

Explicitly link Prelude as part of the build phases for Kickstarter-Framework.
Screenshot 2024-02-13 at 10 33 39 AM

# 🤔 Why

One of our more persistent (and annoying) build issues is this linker error, in which it fails to link to Prelude: 
Screenshot 2024-02-13 at 10 34 18 AM

My best guess (based on a little reading and this helpful [StackOverflow](https://stackoverflow.com/questions/56855766/xcode-build-fails-due-to-undefined-symbol-swift-force-load-swiftuikit) post) is that Xcode is sometimes clever enough to resolve these dependencies automatically, but not consistently. Note that for our other targets, we also have libraries (including Prelude) explicitly linked, so it seems probable that this was just an omission.

Adding this resolved that particular instance of that Prelude linker error that was I was seeing. Hopefully this reduces the overall number of linker errors we get on a regular basis.


ifosli (approved):
Looks promising!

PR #1943, opened 2024-02-13
[MBL-1147] Track push notifications

ifosli:
# 📲 What

Track push notifications by telling braze/appboy when they're opened.

Note: In order to do this smoothly, `Appboy.sharedInstance()` is moved out of the helper function we had it in. I've verified that the shared instance that exists in that helper is different from the one that exists in AppDelegate (but within each file, they're consistent). I'm not sure why that's the case, but I'd theorize it has something to do with different targets or the helper function being static or that it lives within a SEG class and is itself a SEG class. In any case, moving these lines of code out means that `Appboy.sharedInstance()` is consistent from when we add it to the configuration to when we need it for reporting push notifications.

(In the original commit, I stored appboyHelper instead to make it consistent between the config and the push notification report, since I hadn't tried to investigate why they were different yet. This solution is much cleaner.)

# 🤔 Why

Direct opens for push notifications weren't being tracked; now they will be!

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1147)

# ✅ Acceptance criteria

- [x] Direct opens are now tracked

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -27,8 +27,6 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
   internal var rootTabBarController: RootTabBarViewController? {
     return self.window?.rootViewController as? RootTabBarViewController
   }
-  
-  internal var appBoyHelper: SEGAppboyHelper? = nil
 
   func application(
     _ application: UIApplication,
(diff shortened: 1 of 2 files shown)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/AppDelegate.swift:234:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/AppDelegate.swift:472:1 | warning: (redundantParens) Remove redundant parentheses. |
| Library/Tracking/Segment.swift:9:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

amy-at-kickstarter (approved):
LGTM. Yeah, the question about why the instance was broken is interesting - it looks like it's using `dispatch_once` under the hood, which I would expect to be immutable? But regardless of why it was broken, this change makes sense. I suppose if you wanted to be extra cautious you could just have the app delegate keep its own reference to the instance you need to recycle.

PR #1944, opened 2024-02-14
 MBL-1161: Add code to handle storing/retrieving/deleting Keychain items 

amy-at-kickstarter:
# 📲 What

This PR adds `Keychain.swift`, which contains a utility methods for interacting with the system keychain.

# 🤔 Why

Currently we store our authentication tokens in `UserDefaults`, which is insecure. As part of migrating to OAuth, both iOS and Android want to use their equivalent system keychains, which is the correct/recommended implementation for login data.

This is the first step in that - writing code that allows us to actually interact with the keychain.

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -1501,7 +1501,9 @@
 		E10D06652AD48C9C00470B5C /* FetchBackerProjectsQueryRequestForTests.graphql_test in Resources */ = {isa = PBXBuildFile; fileRef = E10D06642AD48C9C00470B5C /* FetchBackerProjectsQueryRequestForTests.graphql_test */; };
 		E10F75E82B6937FA00024AD1 /* PKCETest.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1EEED2A2B686829009976D9 /* PKCETest.swift */; };
 		E113BD782B7D1B7700D3A809 /* Keychain.swift in Sources */ = {isa = PBXBuildFile; fileRef = E113BD762B7D1B7700D3A809 /* Keychain.swift */; };
-		E113BD7A2B7D1B8600D3A809 /* KeychainTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E113BD772B7D1B7700D3A809 /* KeychainTests.swift */; };
+		E113BD842B7D255000D3A809 /* Library_Keychain_iOSTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E113BD832B7D255000D3A809 /* Library_Keychain_iOSTests.swift */; };
+		E113BD852B7D255000D3A809 /* Library.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A755113C1C8642B3005355CF /* Library.framework */; platformFilter = ios; };
+		E113BD8D2B7D255A00D3A809 /* KeychainTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E113BD772B7D1B7700D3A809 /* KeychainTests.swift */; };
 		E118351F2B75639F007B42E6 /* PaginationExampleViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = E118351E2B75639F007B42E6 /* PaginationExampleViewModel.swift */; };
 		E11CFE4B2B6C42CE00497375 /* OAuth.swift in Sources */ = {isa = PBXBuildFile; fileRef = E11CFE492B6C41B400497375 /* OAuth.swift */; };
 		E170B9112B20E83B001BEDD7 /* MockGraphQLClient+CombineTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E170B9102B20E83B001BEDD7 /* MockGraphQLClient+CombineTests.swift */; };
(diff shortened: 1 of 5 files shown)


Library/Keychain.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+}
+
+public struct Keychain {
+  private static var serviceName: String {
+    return Bundle.main.bundleIdentifier ?? "com.kickstarter.kickstarter"
+  }
+
+  public static func deleteAllPasswords() throws {
```
amy-at-kickstarter, line 24: Many of these functions were inspired by the example code @chris-laidlaw-kickstarter sent, here: https://www.advancedswift.com/secure-private-data-keychain-swift

Library/Keychain.swift
```diff
@@ -0,0 +1,115 @@
...
+    let status = SecItemDelete(query as CFDictionary)
+
+    if status == errSecItemNotFound {
+      return
+    } else if status != errSecSuccess {
+      throw KeychainError.unexpectedStatus(status)
+    }
+  }
+
+  public static func storePassword(_ password: String, forAccount account: String) throws {
```
amy-at-kickstarter, line 55: Probably worth noting - I decided to keep the nomenclature of this code fairly close to the way keychain actually works, which uses "accounts" and "passwords". In our case, we're going to be storing an oAuth token against some bit of user data (probably a user e-mail or a user ID). I plan on writing a quick little extension when we integrate this code with the rest of the app to make that clearer:

```
extension Keychain {
  public static func storeOAuthToken(_ token: String, forUserID id: Int) { 
     self.storePassword(token, forAccount: "\(id)")
  }
}
```

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -9128,6 +9153,7 @@
 				SDKROOT = iphoneos;
 				SUPPORTS_MACCATALYST = NO;
 				SWIFT_VERSION = 5.0;
+				TEST_HOST = "$(BUILT_PRODUCTS_DIR)/KickDebug.app/$(BUNDLE_EXECUTABLE_FOLDER_PATH)/KickDebug";
```
amy-at-kickstarter, line 9156: This is the change that was required to get the unit tests to work. It's fairly innocuous looking in Xcode. Annoyingly, I can't find any good documentation around this switch, and why we didn't have it switched on before - my guess is that it might increase the runtime for `Library-Tests`, which I'll keep an eye on.
Screenshot 2024-02-14 at 11 05 18 AM

Library/Keychain.swift
```diff
@@ -0,0 +1,115 @@
...
+      return errString ?? "Unknown keychain error"
+    case .unableToDecodePasswordData:
+      return "Unable to decode password data from keychain."
+    }
+  }
+}
+
+public struct Keychain {
+  private static var serviceName: String {
+    return Bundle.main.bundleIdentifier ?? "com.kickstarter.kickstarter"
```
amy-at-kickstarter, line 21: The documentation for this is poor; I think it can be an arbitrary string, but the bundle ID seems like a nice value to use.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -7256,6 +7324,7 @@
 				A7D1F9441C850B7C000D41D5 /* Kickstarter-iOS */,
 				A755113B1C8642B3005355CF /* Library-iOS */,
 				A75511441C8642B3005355CF /* Library-iOSTests */,
+				E113BD802B7D255000D3A809 /* Library-Keychain-iOSTests */,
```
amy-at-kickstarter, line 7327: OK, to get these tests to work, I had to make a _new testing bundle_, which is hosted, which just contains a single test, `KeychainTests`. Using an app host for `Library-iOSTests` breaks several tests, but without using the app host, `KeychainTests` will fail.

These changes look like this in the GUI:
Screenshot 2024-02-14 at 11 53 14 AM
Screenshot 2024-02-14 at 11 53 26 AM

Kickstarter.xcodeproj/xcshareddata/xcschemes/Library-iOS.xcscheme
```diff
@@ -47,6 +47,16 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "E113BD802B7D255000D3A809"
+               BuildableName = "Library-Keychain-iOSTests.xctest"
+               BlueprintName = "Library-Keychain-iOSTests"
+               ReferencedContainer = "container:Kickstarter.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
```
amy-at-kickstarter, line 59: This schema change is so that when you test `Library-iOS`, it runs both `Library-iOSTests and Library-Keychain-iOSTests`

Screenshot 2024-02-14 at 11 52 59 AM

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ifosli (approved):
Looks good! Thanks for explaining all the things!

PR #1946, opened 2024-02-14
Remove old legacyOauthToken code (from 2016)

amy-at-kickstarter:
# 📲 What

Remove a switch in the code that checks for a legacy OAuth token.

# 🤔 Why

As part of OAuth (or as a fast follow), we're going to swap the apps over to save our OAuth tokens in the keychain. I noticed while poking around our login code that we have this very, very old migration. I look at the commit 9e53ab9 where this was added, and it was back in 2016. If I'm reading our tags right, it was shipped around version 3.0.2, which is so old it's not even tracked in Firebase.

I think it's resoundingly safe to delete this, and it will make the code cleaner when we add a _new_ migration for the _new_ way to store OAuth tokens. 

Worst case scenario is that someone upgrading from a truly ancient version of the app will have to log in again. 🤷‍♀️ 

# ✔️ TODO
- [x] Login before the commit, and again afterwards, to double-check I didn't break anything

### DIFF
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -298,11 +298,6 @@ public struct AppEnvironment: AppEnvironmentType {
     if let oauthToken = data["apiService.oauthToken.token"] as? String {
       // If there is an oauth token stored in the defaults, then we can authenticate our api service
       service = service.login(OauthToken(token: oauthToken))
-      removeLegacyOauthToken(fromUserDefaults: userDefaults)
-    } else if let oauthToken = legacyOauthToken(forUserDefaults: userDefaults) {
-      // Otherwise if there is a token in the legacy user defaults entry we can use that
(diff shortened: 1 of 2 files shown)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:397:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:398:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:399:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

ifosli (approved):
Looks good! Do the login test in the TODOs before submitting, please, if you haven't yet :)

PR #1947, opened 2024-02-14
[MBL-1197] Show unavailable rewards last

ifosli:
# 📲 What

Order rewards so that unavailable ones are shown last. Note that we will need to update the `rewardIsAvailable` utility method to include post campaign backings, but outside of that, this sorting will work for both.

This also deletes unnecessary snapshots for the rewards tests in the first commit. The first commit looks pretty large, but it's mostly indentation changes and deleted snapshots. The second commit is the actual sorting code, with some test changes to make the tests more relevant. I recommend reviewing this one commit at a time.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1197) and snapshot tests.

# ✅ Acceptance criteria

- [x] Unavailable rewards show last
- [x] Rewards can still be selected and interacted with without any weirdness


Kickstarter-iOS/Features/RewardsCollection/Controller/RewardsCollectionViewControllerTests.swift
```diff
@@ -44,6 +44,7 @@ final class RewardsCollectionViewControllerTests: TestCase {
   func testRewards_NonBacker_LiveProject_Landscape() {
     let project = Project.cosmicSurgery
       |> Project.lens.state .~ .live
+      |> Project.lens.rewardData.rewards %~ { Array($0[1...3]) }
```
ifosli, line 47: This just reduces the number of items so that the unavailable one shows as well, even when it's last

Kickstarter-iOS/Features/RewardsCollection/Controller/RewardsCollectionViewControllerTests.swift
```diff
@@ -77,24 +74,23 @@ final class RewardsCollectionViewControllerTests: TestCase {
           |> Backing.lens.rewardId .~ reward.id
           |> Backing.lens.shippingAmount .~ 10
           |> Backing.lens.amount .~ 700.0
+          |> Backing.lens.addOns .~ []
```
ifosli, line 77: Opting out of add ons here ensures that this reward will be shown as unavailable when it's not selected. The default addOn has the same reward id as one of the rewards, which messes with our count for the reward.

scottkicks (approved):
LGTM 👍

PR #1948, opened 2024-02-14
Add a return after the assert

ifosli:
# 📲 What

In our beta build (and probably release build, though I didn't check), having an assert inside a guard block with no other return causes the build to not compile. I'm not 100% sure, but I think the assert gets optimized out or at least doesn't crash the binary in these cases.

# ✅ Acceptance criteria

- [x] Our project builds!


amy-at-kickstarter (approved):
Whoops. Surprised it didn't yell at me in the PR when I did this.

amy-at-kickstarter:
@ifosli yeah, I believe asserts are stripped as part of our build settings:
Screenshot 2024-02-15 at 10 43 24 AM

PR #1950, opened 2024-02-20
MBL-1159: Exchange temporary token for OAuth token and login

amy-at-kickstarter:
# 📲 What

This implements the MVP of login with OAuth. 

It works like this:

1. Open a web authentication session
2. If it completes successful, the website redirects to a URL which includes a tempoary authorization code
3. We exchange that code and a secret for the real OAuth token
4. We use that OAuth token to fetch the user's information
5. ...and then save that to the app environment login. 

# 🤔 Why

We are porting the app to use OAuth via a web authentication session to improve security.

# 👀 See

This snazzy demo!

https://github.com/kickstarter/ios-oss/assets/146007185/19803d96-04c0-4e1c-b596-c9527bc63d8e

### DIFF
--- a/Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
+++ b/Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
@@ -402,12 +402,32 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
   }
 
   fileprivate func pushLoginViewController() {
-    if featureLoginWithOAuthEnabled(), let session = OAuth.createAuthorizationSession() {
+    if featureLoginWithOAuthEnabled(), let session = createAuthorizationSession() {
       session.presentationContextProvider = self
       session.start()
     } else {
       self.navigationController?.pushViewController(LoginViewController.instantiate(), animated: true)
       self.navigationItem.backBarButtonItem = UIBarButtonItem.back(nil, selector: nil)
+    }
+  }
+
+  fileprivate func createAuthorizationSession() -> ASWebAuthenticationSession? {
+    return OAuth.createAuthorizationSession { [weak self] result in
+      switch result {
+      case .loggedIn:
+        self?.viewModel.inputs.environmentLoggedIn()
+      case let .failure(errorMessage):
+        let alert = UIAlertController(
+          title: Strings.login_errors_title(),
+          message: errorMessage,
+          preferredStyle: .alert
+        )
+        alert.addAction(UIAlertAction(title: Strings.login_errors_button_ok(), style: .cancel))
+        self?.present(alert, animated: true)
+      case .cancelled:
+        // Do nothing
+        break
+      }
     }
   }
 
--- a/KsApi/MockService.swift
+++ b/KsApi/MockService.swift
@@ -1399,6 +1399,15 @@
       return SignalProducer(value: self.fetchUserSelfResponse ?? .template)
     }
 
+    func fetchUserSelf_combine(withToken _: String) -> AnyPublisher<User, ErrorEnvelope> {
+      if let error = fetchUserSelfError {
+        return Fail(outputType: User.self, failure: self.fetchUserSelfError!).eraseToAnyPublisher()
+      }
+
+      return Just(self.fetchUserSelfResponse ?? .template).setFailureType(to: ErrorEnvelope.self)
+        .eraseToAnyPublisher()
+    }
+
     internal func fetchSurveyResponse(surveyResponseId id: Int)
       -> SignalProducer<SurveyResponse, ErrorEnvelope> {
       if let response = fetchSurveyResponseResponse {
@@ -1742,6 +1751,12 @@
     }
 
     func fetchDiscovery_combine(paginationUrl _: String) -> AnyPublisher<DiscoveryEnvelope, ErrorEnvelope> {
+      Empty(completeImmediately: false).eraseToAnyPublisher()
+    }
+
+    func exchangeTokenForOAuthToken(params _: OAuthTokenExchangeParams)
+      -> AnyPublisher<OAuthTokenExchangeResponse, ErrorEnvelope> {
+      // TODO: Implement for testing.
       Empty(completeImmediately: false).eraseToAnyPublisher()
     }
   }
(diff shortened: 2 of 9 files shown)


Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -411,6 +411,26 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
     }
   }
 
+  fileprivate func createAuthorizationSession() -> ASWebAuthenticationSession? {
+    return OAuth.createAuthorizationSession { [weak self] result in
+      switch result {
+      case .loggedIn:
+        self?.viewModel.inputs.environmentLoggedIn()
```
amy-at-kickstarter, line 418: I went back-and-forth on where to put which pieces of logic. I settled on having the actual login call live in `OAuth.swift`, but the view controller is responsible for updating the UI when the login is complete.

KsApi/lib/Route.swift
```diff
@@ -243,6 +248,9 @@ internal enum Route {
       case .userSelf:
         return (.GET, "/v1/users/self", [:], nil)
 
+      case let .userSelfWithToken(token):
+        return (.GET, "/v1/users/self", ["oauth_token": token], nil)
```
amy-at-kickstarter, line 252: Parameters passed in here will override the defaults, which in this case will be missing the `oauth_token`.

Library/OAuth.swift
```diff
@@ -44,8 +54,47 @@ public struct OAuth {
...
         callbackURLScheme: OAuth.redirectScheme
-      ) { _, _ in
-        // TODO: MBL-1159: Exchange information in callback for credentials, then login.
+      ) { url, error in
+        guard error == nil else {
+          if let authenticationError = error as? ASWebAuthenticationSessionError,
+            authenticationError.code == .canceledLogin {
+            onComplete(.cancelled)
+          } else {
+            onComplete(.failure(errorMessage: Strings.Something_went_wrong_please_try_again()))
```
amy-at-kickstarter, line 63: These are all errors that aren't user-solvable, so giving them a generic error message.

KsApi/MockService.swift
```diff
@@ -1744,6 +1753,12 @@
     func fetchDiscovery_combine(paginationUrl _: String) -> AnyPublisher<DiscoveryEnvelope, ErrorEnvelope> {
       Empty(completeImmediately: false).eraseToAnyPublisher()
     }
+
+    func exchangeTokenForOAuthToken(params _: OAuthTokenExchangeParams)
+      -> AnyPublisher<OAuthTokenExchangeResponse, ErrorEnvelope> {
+      // TODO: Implement for testing.
+      Empty(completeImmediately: false).eraseToAnyPublisher()
```
amy-at-kickstarter, line 1760: I need to come back and evaluate what of this flow I can test. It's all fairly tightly integrated with Apple's authentication controller. I could add tests just for these new endpoints, but since network tests go against these mocks anyways, it doesn't seem super valuable to me...

amy-at-kickstarter: @ifosli @scottkicks I'm open to suggestions for what seems worthwhile to test here.

amy-at-kickstarter:
Converting to draft while I write some tests.

Library/OAuth.swift
```diff
@@ -55,4 +65,67 @@ public struct OAuth {
       return nil
     }
   }
+
+  internal static func handleRedirect(
```
amy-at-kickstarter, line 69: Pulled this out into its own method to make it testable, without needing to deal with `ASWebAuthenticationSession` in the tests.

Library/OAuth.swift
```diff
@@ -55,4 +65,67 @@ public struct OAuth {
...
+    redirectURL url: URL?,
+    error: Error?,
+    verifier: String,
+    onComplete: @escaping (OAuthAuthorizationResult) -> Void
+  ) {
+    guard error == nil else {
+      if let authenticationError = error as? ASWebAuthenticationSessionError,
+        authenticationError.code == .canceledLogin {
+        DispatchQueue.main.async {
```
amy-at-kickstarter, line 78: These dispatches are here to prevent us [from releasing Zalgo](https://blog.izs.me/2013/08/designing-apis-for-asynchrony/) - the completion block is _only ever called asynchronously_.

Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -411,6 +411,26 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
...
+      case let .failure(errorMessage):
+        let alert = UIAlertController(
+          title: Strings.login_errors_title(),
+          message: errorMessage,
+          preferredStyle: .alert
+        )
+        alert.addAction(UIAlertAction(title: Strings.login_errors_button_ok(), style: .cancel))
```
amy-at-kickstarter, line 425: Interestingly these strings existed, but were unused 🤷‍♀️

Library/OAuthTests.swift
```diff
@@ -0,0 +1,94 @@
+import AuthenticationServices
+@testable import KsApi
+@testable import Library
+import XCTest
+
+final class OAuthTests: XCTestCase {
```
amy-at-kickstarter, line 6: These tests are all a little tautological (they really test the specific internal implementation), but I feel like I'd rather have them than not.

ifosli (approved):

Library/OAuth.swift
```diff
@@ -55,4 +65,67 @@ public struct OAuth {
...
+        let token = response.token
+        // TODO: This would be neater if we can just return the V1 user from the exchange endpoint.
+
+        // Return a publisher that emits a tuple of (token, user) when the user request completes
+        return Just(token).setFailureType(to: ErrorEnvelope.self)
+          .zip(AppEnvironment.current.apiService.fetchUserSelf_combine(withToken: token))
+      }
+      .sink { result in
+        if case let .failure(error) = result {
+          let message = error.errorMessages.first ?? Strings.login_errors_unable_to_log_in()
```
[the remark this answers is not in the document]
amy-at-kickstarter: Good question. I assumed that server error messages are localized, but I've never actually checked - is that assumption incorrect?

PR #1951, opened 2024-02-20
[MBL-1207] Pledge Button on Project Page

scottkicks:
this applies to PCP campaigns that backers have already pledged to

# 📲 What

Updates the pledge button on project pages to say "Back this project" when a campaign has Post-Campaign Pledging (PCP) enabled and is actively accepting late pledges.

In addition, I also made sure that the correct pledge button shows on rewards when in a PCP state. Right now, if a project isn't live or has already been backed, we don't show anything. Now, the pledge button will show when in an active PCP state _if the user hasn't already backed_. 
If the user has already backed it, they'll see the normal "view your pledge" CTA that presents users with their pledge info.

# 🤔 Why

If PCP is disabled and a user has backed the campaign, we show a “View Rewards” button. For this feature, we want backers to be able to pledge post a successful campaign. It's important to note that backers should never be able to pledge more than once to **any campaign**

# 🛠 How

Let’s use the feature flag and the [new PCP properties](https://github.com/kickstarter/ios-oss/pull/1937) on the project fragment to gate this new behavior.

# 👀 See

**[Designs](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=54-5016&mode=design&t=a4GepSxu18KcroSW-0)**

# ✅ Acceptance criteria

- [x] If a campaign has PCP enabled _and_ is actively accepting late pledges _and_ users have not backed already, then the project should show the "Back this project" CTA
- [x]  If a campaign has PCP enabled _and_ is actively accepting late pledges _and_ users have already pledged, then the project should show the "view your pledge" pledge CTA.
- [x] If a campaign has PCP disabled _or_ is not actively accepting late pledges, then the normal pledge flow should be active


Library/SharedFunctions.swift
```diff
@@ -380,6 +380,15 @@ public func rewardsCarouselCanNavigateToReward(_ reward: Reward, in project: Pro
   let isAvailableForNewBacker = rewardIsAvailable(project: project, reward: reward) && !isBacking
   let isAvailableForExistingBackerToEdit = (isBacking && reward.hasAddOns)
 
+  if featurePostCampaignPledgeEnabled(), project.postCampaignPledgingEnabled,
+    project.isInPostCampaignPledgingPhase {
+    return [
+      project.state == .successful,
+      isAvailableForNewBacker || isAvailableForExistingBackerToEdit
```
ifosli, line 387: Can post campaign pledges be edited? I thought they couldn't, since they're charged immediately

Library/RewardCellProjectBackingStateType.swift
```diff
@@ -5,6 +5,7 @@ public enum RewardCellProjectBackingStateType: Equatable {
   public enum ProjectState {
     case nonLive
     case live
+    case successful
```
ifosli, line 8: It looks like we're using `successful` here to mean `post campaign pledging available`, which seems like a bad idea. That makes it easy to confuse with successful projects that don't have post campaign pledging enabled in the future. I think we should be consistent with the backend naming here and call this `inPostCampaignPledgingPhase` instead, which is a little long but much clearer.

Library/RewardCellProjectBackingStateType.swift
```diff
@@ -13,6 +14,11 @@ public enum RewardCellProjectBackingStateType: Equatable {
 
   static func state(with project: Project) -> RewardCellProjectBackingStateType {
     guard let backing = project.personalization.backing else {
+      if featurePostCampaignPledgeEnabled(), project.postCampaignPledgingEnabled,
+        project.isInPostCampaignPledgingPhase {
```
ifosli, line 18: nit: I don't think we need to check if the `project.postCampaignPledgingEnabled` - I don't think that can be false if it's in the post campaign pledging phase.

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -210,6 +210,11 @@ private func pledgeCTA(project: Project, backing: Backing?) -> PledgeStateCTATyp
       return PledgeStateCTAType.viewYourRewards
     }
 
+    if featurePostCampaignPledgeEnabled(), project.postCampaignPledgingEnabled,
+      project.isInPostCampaignPledgingPhase {
```
ifosli, line 214: same here; I think we only need to check if it's in the post campaign pledging phase and if the flag is flipped.

Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -112,10 +112,12 @@ private func pledgeButtonTitle(project: Project, reward: Reward) -> String? {
   case (.nonBacked(.live), _, true):
     return Strings.Select()
   case (.backed(.nonLive), false, _),
-       (.nonBacked(.nonLive), _, _):
+       (.nonBacked(.nonLive), _, _), (.backed(live: .successful), _, _):
```
ifosli, line 115: Nit: I don't think we need the `live:` here. Also, I would find this a little easier to read if it were spread over three lines instead of one line with one statement and one line with two

Library/ViewModels/RewardCardContainerViewModel.swift
```diff
@@ -146,6 +148,8 @@ private func buttonStyleType(project: Project, reward: Reward) -> ButtonStyleTyp
     return .none
   case .nonBacked(.nonLive):
     return .none
+  case .nonBacked(live: .successful), .backed(live: .successful):
```
ifosli, line 151: nit: For consistency, no `live:` here either would be more readable

ifosli (changes requested):
I like where this is headed, but I have some concerns. I've left comments. Also, I'm confused about this line: "If the user has already backed it, they'll see the normal "View Rewards" CTA" - if a user has backed a thing, they shouldn't see a "view rewards" CTA - they should see a "view your pledge" button. Can you make sure you test this scenario, if you haven't already?

Library/ViewModels/RewardCardContainerViewModel.swift
scottkicks: 👍 happy to update. I'm not really a fan of using one method to handle all of these different pledge states

scottkicks:
@ifosli, your suggestions also pointed out that we'll probably want to show the manage pledge button and update the manage pledge view as well. [Posted in Slack](https://kickstarter.slack.com/archives/C03J6NRA8/p1708526931929909)for some team/design feedback and will finish this PR once I get confirmation on that.

Library/ViewModels/PledgeCTAContainerViewViewModel.swift
```diff
@@ -205,21 +205,29 @@ private func pledgeCTA(project: Project, backing: Backing?) -> PledgeStateCTATyp
...
   // NB: Add error case back once correctly returned
-  switch (project.state, projectBacking.status) {
-  case (.live, _):
+  switch (project.state, projectBacking.status, isInPostCampaign) {
+  case (.live, _, _):
     return .manage
-  case (.successful, .errored):
+  case (.successful, _, true):
+    return .viewBacking
+  case (.successful, .errored, _):
```
ifosli, line 228: I could be wrong here, but I think a user could have an errored backing even if the project is in a post campaign state. If that's the case, we'd want to check the errored case first, in which case I don't think we actually need any custom post campaign logic at all?

ifosli (approved):
Just one question but otherwise it looks good!

PR #1952, opened 2024-02-21
[MBL-1209] Select Your Reward Title

scottkicks:
# 📲 What

Adds "Select your reward." to Rewards views _if post campaign feature flag is **on**_
- We decided that this will get rolled out to all projects so I didn't also gate this behind the pcp project properties.

# 🤔 Why

With the new PCP flow having an additional "Confirm details" screen, design has decided that it'd be beneficial to title each page to provide users with more context.

# 🛠 How

Adds a header to the RewardsCollectionView.

Note that there is a touch more space between the rewards options and the title than the designs suggest. This is because each reward moves up slightly when scrolling through their details. Designs didn't catch that edge case and will update the [figma](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=54-5016&mode=design&t=ya5v8NrLJGO5Z169-0) soon.

The title string is hardcoded for now and will need to be updated once all copy translations are in. I've added a comment with the [ticket](https://kickstarter.atlassian.net/browse/MBL-1217) to address that.

# 👀 See
Simulator Screen Recording - iPhone 15 Pro - 2024-02-21 at 13 12 58

# ✅ Acceptance criteria

- [x] New title appears when viewing rewards and matches the [designs](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=73-5153&mode=design&t=hlCDBHT4ARSgwjnt-0)


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/RewardsCollection/Controller/RewardsCollectionViewController.swift:238:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

Kickstarter-iOS/Features/RewardsCollection/Controller/RewardsCollectionViewController.swift
```diff
@@ -203,6 +236,13 @@ final class RewardsCollectionViewController: UICollectionViewController {
       self.collectionView.topAnchor.constraint(equalTo: self.view.topAnchor)
     ])
 
+    if featurePostCampaignPledgeEnabled() {
+      self.headerView.leftAnchor.constraint(equalTo: self.view.leftAnchor).isActive = true
+      self.headerView.rightAnchor.constraint(equalTo: self.view.rightAnchor).isActive = true
+      self.headerView.topAnchor.constraint(equalTo: self.view.topAnchor).isActive = true
+      self.headerView.widthAnchor.constraint(equalTo: self.view.widthAnchor).isActive = true
```
ifosli, line 243: nit: Does the width anchor do anything when we already have left and right? Also, the height is underconstrained - does it give a build warning? If it does, can you add a height constraint and just set it to the intrinsic content size?

Kickstarter-iOS/SharedViews/HeaderFooterViews/RewardsCollectionViewHeaderView.swift
```diff
@@ -0,0 +1,60 @@
...
+      bottom: Styles.grid(1),
+      right: Styles.grid(3)
+    )
+    |> \.backgroundColor .~ .clear
+}
+
+private let labelStyle: LabelStyle = { label in
+  label
+    |> \.numberOfLines .~ 1
+    |> \.lineBreakMode .~ .byTruncatingTail
```
ifosli, line 57: Are you sure we should be truncating this? My general ux experience says that strings we provide should never be truncated. Happy to follow what UX recommends, but please double check to make sure. I really think we shouldn't truncate it at all. (Though it might be worth asking Alison what she thinks the UX should look like in ex german with the largest font size - does adding a title make the screen unusable, especially in landscape mode?)

ifosli (approved):
A few things, none of which need to block this pr, but which I think we need to iron out before the feature is rolled out. I'm particularly concerned that adding a title will make landscape mode unusable for this screen, especially for accessible font sizes.

Kickstarter-iOS/Features/RewardsCollection/Controller/RewardsCollectionViewController.swift
scottkicks: I was seeing issues originally, but it looks like now that I'm setting the collectionView's topAnchor directly, that issue has been fixed 🤷 .

scottkicks:
@ifosli I tested landscape mode in a few device sizes but didn't consider testing different languages or larger accessibility fonts. Thanks for calling that out! I think I addressed all of your concerns. Next time, I'll try to remember to add a screen recording of landscape so that it's clear.

PR #1953, opened 2024-02-22
[MBL-1218] Create RefInfo struct and use it where needed

ifosli:
# 📲 What

Create a `RefInfo` struct consisting of the `RefTag` enum. In a future pr, this struct will be updated to also hold a deeplink url. We need this in order to catch the deeplink url and pass it along whenever the project page is opened from a deeplink.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1218)

# ✅ Acceptance criteria

- [x] Functionality is unchanged


scottkicks (approved):
👍 just update with main before merging

PR #1954, opened 2024-02-22
MBL-1214: Send OAuth token only in headers, not in request parameters

amy-at-kickstarter:
# 📲 What

Send OAuth tokens only in the request headers, not in the request parameters.

# 🤔 Why

This is a security concern which we wanted to address with the MVP of OAuth. Moving the authorization parameters into headers is a best practice.

See: https://github.com/kickstarter/kickstarter/pull/26425 and https://github.com/kickstarter/kickstarter/pull/26432

# 🛠 How

Note that this isn't behind any flags. The server change to accept the OAuth token in the headers was made globally, for all V1 requests, and already applied for GraphQL. We decided a ramp-up would be difficult, due to how deep the headers are buried in our networking stack, and of limited use, compared to testing this change manually.

Also note that we are sometimes sending _redundant headers_, like both `Authorization` and `X-Auth` with the OAuth token. To the best of my knowledge, both V1 and GraphQL discard extra headers, so there's no harm in this - and it makes our code significantly simpler.

### DIFF
--- a/KsApi/Service+RequestHelpers.swift
+++ b/KsApi/Service+RequestHelpers.swift
@@ -41,6 +41,26 @@ extension Service {
     ).handle_combine_dataResponse(service: self)
   }
 
+  func requestWithAuthentication<Model: Decodable>(_ route: Route,
+                                                   oauthToken: String) -> AnyPublisher<Model, ErrorEnvelope> {
+    let properties = route.requestProperties
+
+    guard let URL = URL(string: properties.path, relativeTo: self.serverConfig.apiBaseUrl as URL) else {
+      fatalError(
+        "URL(string: \(properties.path), relativeToURL: \(self.serverConfig.apiBaseUrl)) == nil"
+      )
+    }
+
+    var preparedRequest = preparedRequest(forURL: URL, method: properties.method, query: properties.query)
+    addV1AuthenticationToRequest(&preparedRequest, oauthToken: oauthToken)
+
+    return Service.session.combine_dataResponse(
+      preparedRequest,
+      uploading: properties.file.map { ($1, $0.rawValue) },
+      and: self.perimeterXClient
+    ).handle_combine_dataResponse(service: self)
+  }
+
   func requestPaginationDecodable<Model: Decodable>(_ paginationUrl: String)
     -> SignalProducer<Model, ErrorEnvelope> {
     guard let paginationUrl = URL(string: paginationUrl) else {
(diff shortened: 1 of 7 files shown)


KsApi/Service+RequestHelpers.swift
```diff
@@ -41,6 +41,26 @@ extension Service {
     ).handle_combine_dataResponse(service: self)
   }
 
+  func requestWithAuthentication<Model: Decodable>(_ route: Route,
```
amy-at-kickstarter, line 44: The amount of copy-paste here makes me unhappy. However, I poked around with a few ideas for refactoring, and realized it was going to take significantly _more_ refactoring to make a dent in this. I decided that (for the moment) this was an OK evil. 

Because this is only used by precisely one route, I would much rather have one ugly copy-pasted method than add an extra "authentication header" parameter throughout our whole networking stack. And I would rather ship this now than embark in a larger refactoring project for `Route`.

KsApi/lib/Route.swift
```diff
@@ -248,9 +247,6 @@ internal enum Route {
       case .userSelf:
         return (.GET, "/v1/users/self", [:], nil)
 
-      case let .userSelfWithToken(token):
```
amy-at-kickstarter, line 251: Unnecessary; handling this at the Service level.

Library/OAuth.swift
```diff
@@ -97,15 +97,15 @@ public struct OAuth {
...
       .flatMap { response in
         let token = response.token
         // TODO: This would be neater if we can just return the V1 user from the exchange endpoint.
 
         // Return a publisher that emits a tuple of (token, user) when the user request completes
         return Just(token).setFailureType(to: ErrorEnvelope.self)
-          .zip(AppEnvironment.current.apiService.fetchUserSelf_combine(withToken: token))
+          .zip(AppEnvironment.current.apiService.fetchUserSelf_combine(withOAuthToken: token))
       }
+      .receive(on: RunLoop.main)
```
amy-at-kickstarter, line 108: Whoops, this wasn't applying to the `fetchUserSelf_combine` call because it happened before. 🤦‍♀️

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| KsApi/ServiceTypeTests.swift:329:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

KsApi/Service+RequestHelpers.swift
ifosli: Just to confirm: this is necessary because we're calling this as soon as we have the token; before the token has been set (since we set the token when we have the user as well)? If I'm understanding this correctly, I support this as an OK evil, especially if we still think we may get the user object returned with the token in the future.

ifosli (approved):
Very nice!

PR #1957, opened 2024-02-23
[MBL-1220] Add deeplink url to refInfo when opening proj page

ifosli:
# 📲 What

Add a deeplink url to the RefInfo struct. This is captured for deeplink opens to anything in the project page. Once the attribution event is ready, this will be included there.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1220)

# ✅ Acceptance criteria

- [x] Test by setting a breakpoint in `configureWith(projectOrParam: refInfo:)` in `ProjectPageViewModel` and confirm that opening the project page using a deeplink causes the deeplink to be available there

### DIFF
--- a/Library/Navigation.swift
+++ b/Library/Navigation.swift
@@ -553,7 +553,7 @@ private func parsedParams(url: URL, fromTemplate template: String) -> RouteParam
   }
 
   // If the deeplink opens the project page, track the url.
-  if urlComponents[0] == "projects" {
+  if urlComponents.first == "projects" {
     object[RouteParamsDecoded.CodingKeys.deeplinkUrl.rawValue] = url.absoluteString
   }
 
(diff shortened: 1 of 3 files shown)


Library/Navigation.swift
```diff
@@ -552,6 +552,11 @@ private func parsedParams(url: URL, fromTemplate template: String) -> RouteParam
     object[key] = value
   }
 
+  // If the deeplink opens the project page, track the url.
+  if urlComponents[0] == "projects" {
```
amy-at-kickstarter, line 556: Nit: is the URL guaranteed to have a path component? Otherwise you might want to do `urlComponents.first` instead.

ifosli: Yeah, it is right now but that might change in the future, so good call out! Fixed.

amy-at-kickstarter (approved):
LGTM

PR #1958, opened 2024-02-23
Update schema for attribution

ifosli:
# 📲 What

Run `bin/apollo-schema-download.sh` and update the validate checkout mutation to align with changes.

I decided to do this in a separate pr from the event attribution mutation creation since updating the graphql-schema caused me to need to fix the build.

# ✅ Acceptance criteria

- [ ] Build still works


ifosli:
@scottkicks I think this is an uneventful change since I don't think this mutation is actually being used yet, but it might affect your work. Let me know if you'd rather own this or if there's anything else I should test before merging this.

scottkicks (approved):
The implementation of the new PCP related mutations haven't been started so this is fine.

PR #1959, opened 2024-02-26
MBL-1161: Run AppEnvironmentTests in both hosted and unhosted test bundles

amy-at-kickstarter:
# 📲 What

Add `AppEnvironmentTests` to the hosted test bundle, `Library-Keychain-iOSTests`. This also includes some bug fixes to make that change work. 

# 🤔 Why

As part of OAuth, we want to start storing our OAuth token in the keychain, instead of in user defaults. This has led to some interesting problems with our test environment. The keychain _only works_ if your unit tests are "hosted" within an application, which means the tests will actually boot our app before running the unit tests. However, many of our unit tests break in a hosted environment. This is because the app has its own state - it pushes/pops/manipulates a lot of global environment, and that overlaps with tests.

# 🛠 How

By adding `AppEnvironmentTests` to both the hosted and the unhosted test bundles, I'll be able to test the `AppEnvironment` both with and without the keychain. When I moved the test over, I noticed certain things break - state from the app was leaking into `AppEnvironmentTests`.

The solution I landed on was pretty simple; I just added some extra code to `setUp` to wipe out any state from `AppEnvironment` before running the tests. Interestingly, I noticed that doing this broke a completelely unrelated test `PledgeDisclaimerViewModelTests`. This was because `PledgeDisclaimerViewModelTests` had an unmatched environment push/pop. 

# 👀 See

Screenshot 2024-02-26 at 10 44 45 AM

Screenshot 2024-02-26 at 10 53 20 AM

### DIFF
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -133,7 +133,19 @@ public struct AppEnvironment: AppEnvironmentType {
       ubiquitousStore: next.ubiquitousStore,
       userDefaults: next.userDefaults
     )
+
+    // If there are no more items in the stack,
+    // then the next call to AppEvironment.current will fail.
+    assert(self.stack.count > 0)
+
     return last
+  }
+
+  internal static func resetStackForUnitTests() {
+    while self.stack.popLast() != nil {}
+
+    let next = Environment()
+    self.pushEnvironment(next)
   }
 
   // Replace the current environment with a new environment.
--- a/Library/AppEnvironmentTests.swift
+++ b/Library/AppEnvironmentTests.swift
@@ -5,6 +5,10 @@ import Prelude
 import XCTest
 
 final class AppEnvironmentTests: XCTestCase {
+  override func setUp() {
+    AppEnvironment.resetStackForUnitTests()
+  }
+
   func testPushAndPopEnvironment() {
     let lang = AppEnvironment.current.language
 
--- a/Library/ViewModels/PledgeDisclaimerViewModelTests.swift
+++ b/Library/ViewModels/PledgeDisclaimerViewModelTests.swift
@@ -9,6 +9,7 @@ final class PledgeDisclaimerViewModelTests: TestCase {
   private let notifyDelegateLinkTappedWithURL = TestObserver<URL, Never>()
 
   override func setUp() {
+    super.setUp()
     self.vm.outputs.notifyDelegateLinkTappedWithURL
       .observe(self.notifyDelegateLinkTappedWithURL.observer)
   }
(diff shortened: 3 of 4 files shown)


Library/AppEnvironment.swift
```diff
@@ -133,9 +133,21 @@ public struct AppEnvironment: AppEnvironmentType {
...
     )
+
+    // If there are no more items in the stack,
+    // then the next call to AppEvironment.current will fail.
+    assert(self.stack.count > 0)
+
     return last
   }
 
+  internal static func resetStackForUnitTests() {
```
amy-at-kickstarter, line 144: Sadly, no way to mark this as only visible for testing, so I just made it obvious in the name.

Library/AppEnvironmentTests.swift
```diff
@@ -5,6 +5,10 @@ import Prelude
 import XCTest
 
 final class AppEnvironmentTests: XCTestCase {
+  override func setUp() {
+    AppEnvironment.resetStackForUnitTests()
```
amy-at-kickstarter, line 9: This keeps any state from the actual application host from leaking into `AppEnvironmentTests`.

Library/ViewModels/PledgeDisclaimerViewModelTests.swift
```diff
@@ -9,6 +9,7 @@ final class PledgeDisclaimerViewModelTests: TestCase {
   private let notifyDelegateLinkTappedWithURL = TestObserver<URL, Never>()
 
   override func setUp() {
+    super.setUp()
```
amy-at-kickstarter, line 12: `PledgeDisclaimerViewModelTests` extends `TestCase`, which sets up a default environment. Because this test wasn't calling `super.setUp()` but is (implicitly) calling `super.tearDown()`, it was unintentionally popping an extra environment off the stack.

Library/AppEnvironment.swift
```diff
@@ -133,9 +133,21 @@ public struct AppEnvironment: AppEnvironmentType {
       ubiquitousStore: next.ubiquitousStore,
       userDefaults: next.userDefaults
     )
+
+    // If there are no more items in the stack,
+    // then the next call to AppEvironment.current will fail.
+    assert(self.stack.count > 0)
```
amy-at-kickstarter, line 139: AFAIK we disable assertions for release builds. Plus, we already had a force unwrap when you call `AppEnvironment.current`, so the worst this would do would be to crash slightly sooner.

ifosli (approved):

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:409:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:410:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:411:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

PR #1961, opened 2024-02-27
[MBL-1223] Track attribution event

ifosli:
# 📲 What

Create an event attribution tracking event the first time a project page loads.  This will ensure the event is sent only once per project (since we create a new view model when we configure the view controller with a new project).

Note that we can't simply send the event when the view model is configured, since we need the graphQL id of the project, which we don't have until the fresh project fetch completes.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1223)
[Spike](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2605711496/Event+attribution+Spike#Client-side-implementation-details-based-on-API)

# ✅ Acceptance criteria

- [x] Event gets sent with empty url whenever the project page is opened from within the app
- [x] Event gets sent with a context page url whenever the project page is opened from a deeplink


Library/Tracking/AttributionTracking.swift
```diff
@@ -0,0 +1,20 @@
...
+    let sessionRefTag = refInfo?.refTag?.stringTag
+    let contextPageUrl = refInfo?.deeplinkUrl
+    var props = [String: String]()
+    if let sessionRefTag {
+      props["session_ref_tag"] = sessionRefTag
+    }
+    props["context_page_url"] = contextPageUrl ?? ""
+
+    let propsData = try? JSONSerialization.data(withJSONObject: props)
+    return String(data: propsData!, encoding: .utf8)
```
amy-at-kickstarter, line 18: Is the force unwrap here necessary? Looks like you're already returning `String?` so I assume a `nil` would be harmless.

amy-at-kickstarter (approved):
One nit/suggestion but otherwise lgtm

Library/Tracking/AttributionTracking.swift
ifosli: Good catch! I like to force unwrap things when I'm testing but I didn't mean to leave it in. Fixed.

PR #1962, opened 2024-02-27
MBL-1233: Use correct string for 'Sign up or Log in' button

amy-at-kickstarter:
# 📲 What

Change the string I'm using for a button.

# 🤔 Why

This should have its own translation, instead of using the translation for the similar button in Discover.

We'll have to update the strings files once the translations percolate through the system.

### DIFF
--- a/Kickstarter-iOS/Locales/de.lproj/Localizable.strings
+++ b/Kickstarter-iOS/Locales/de.lproj/Localizable.strings
@@ -72,13 +72,13 @@
 "Big_Kickstarter_news_and_events_near_you" = "Ein paar Mal pro Jahr: Kickstarter-Neuigkeiten und Events in deiner Nähe.";
 "Big_Kickstarter_news_plus_occasional_projects_and_events_chosen_just_for_you" = "Das Neueste von Kickstarter plus gelegentlich für dich ausgewählte Projekte und Events.";
 "Biography" = "Biografie";
-"Block" = "Block";
+"Block" = "Blockieren";
 "Block_this_user" = "Diesen Nutzer blockieren";
 "Block_user_fail" = "Deine Anfrage konnte nicht verarbeitet werden. Bitte versuche es noch einmal.";
 "Block_user_success" = "Dieser Nutzer wurde erfolgreich blockiert";
 "Block_username" = "%{username} blockieren?";
 "Blocked_user" = "Blockierter Nutzer";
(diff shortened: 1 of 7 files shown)


Kickstarter-iOS/Locales/de.lproj/Localizable.strings
```diff
@@ -1467,6 +1467,7 @@
 "login_tout.generic_intent.title" = "Hoppla! Um fortzufahren, musst du dich auf Kickstarter einloggen. Du kannst dich unten mit deinem Facebook-Konto oder deiner E-Mail-Adresse anmelden oder dich neu registrieren.";
 "login_tout.generic_intent.traditional_login_button" = "Anmelden";
 "login_tout.generic_intent.traditional_signup_button" = "Registrieren";
+"login_tout.generic_intent.traditional_signup_or_login_button" = "Sign up or Log in";
```
amy-at-kickstarter, line 1470: @scottkicks @ifosli Do we wait for translations before cutting a release? How do I know if translations are done?

ifosli: In general, I like to make sure any untranslated strings are behind a flag. That flag doesn't get rolled out until I confirm all translations are in. When a string isn't behind a flag, I'd personally wait to use it until it's fully translated. (Ie don't do this pr and mark the ticket as blocked until translations are in.) However, if using the correct string here would be a release blocker, you could just merge this and create a 1 pointer to verify strings are translated before the release is cut.

At my old job, we had automated tooling to check that all strings had translations, which we don't have here, so I just check each of the language files to make sure it's not the original english.

ifosli: We seldom have new strings outside of a flag-guarded new feature, though, so it wouldn't surprise me if there isn't a process for it

ifosli: Wait, sorry, this is totally behind a flag! My personal method then is to just check translations before flipping the flag. If you think getting translations in will be tight and that having the correct string isn't a release blocker, I'd still recommend waiting to update the string until translations are in. Technically, we're supposed to run the make strings script before we cut a release, but I don't think we have a process for who's responsible for checking that translations are actually in.

amy-at-kickstarter: Aha, good to know. OK. I'll make myself a 1-point ticket to remember it - worst case scenario, I can always copy-paste the old strings by hand into the strings from (from the old button) and let it be overwritten when the actual translations come in.

PR #1963, opened 2024-02-27
MBL-1161: Read and store OAuth token from keychain

amy-at-kickstarter:
# 📲 What

Store the OAuth token in the keychain, instead of in user defaults. Additional code moves the old token out of defaults and into the keychain if you were already logged in.

# 🤔 Why

Leaving the token in user defaults is insecure. 

# 🛠 How

We store the OAuth token per-user, using a pseudo-account-name like `kickstarter_12345` for user with id `12345`. The change is gated behind a feature flag, and hopefully fairly robust to any keychain failures or exceptions.

# ⏰ TODO
- [x] Solicit feedback about storing token per-user vs. just one for the entire app
- [ ] Run testing on-device

### DIFF
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -296,12 +296,12 @@ public struct AppEnvironment: AppEnvironmentType {
     )
   }
 
-  private static func accountNameForUserId(_ userId: Int) -> String {
+  internal static func accountNameForUserId(_ userId: Int) -> String {
     return "kickstarter_\(userId)"
   }
 
   private static func storeOAuthTokenToKeychain(_ oauthToken: String, forUserId id: Int?) -> Bool {
-    guard featureLoginWithOAuthEnabled(), let userId = id
+    guard featureUseKeychainForOAuthTokenEnabled(), let userId = id
     else {
       return false
     }
@@ -317,7 +317,7 @@ public struct AppEnvironment: AppEnvironmentType {
   }
 
   private static func fetchOAuthTokenFromKeychain(forUserId id: Int?) -> String? {
-    guard featureLoginWithOAuthEnabled(), let userId = id
+    guard featureUseKeychainForOAuthTokenEnabled(), let userId = id
     else {
       return nil
     }
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -351,9 +351,6 @@ public struct AppEnvironment: AppEnvironmentType {
       service = service.login(OauthToken(token: oauthToken))
     } else if let oauthToken = data["apiService.oauthToken.token"] as? String {
       service = service.login(OauthToken(token: oauthToken))
-
-      // Move it over to the keychain if we can
-      _ = self.storeOAuthTokenToKeychain(oauthToken, forUserId: userFromDefaults?.id)
     }
 
     // Try restoring the client id for the api service
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -1,4 +1,5 @@
 import FBSDKCoreKit
+import FirebaseCrashlytics
 import Foundation
 import KsApi
 import Prelude
@@ -295,6 +296,41 @@ public struct AppEnvironment: AppEnvironmentType {
     )
   }
 
+  private static func accountNameForUserId(_ userId: Int) -> String {
+    return "kickstarter_\(userId)"
+  }
+
+  private static func storeOAuthTokenToKeychain(_ oauthToken: String, forUserId id: Int?) -> Bool {
+    guard featureLoginWithOAuthEnabled(), let userId = id
+    else {
+      return false
+    }
+
+    do {
+      try Keychain.storePassword(oauthToken, forAccount: self.accountNameForUserId(userId))
+      return true
+    } catch {
+      Crashlytics.crashlytics().record(error: error)
+    }
+
+    return false
+  }
+
+  private static func fetchOAuthTokenFromKeychain(forUserId id: Int?) -> String? {
+    guard featureLoginWithOAuthEnabled(), let userId = id
+    else {
+      return nil
+    }
+
+    do {
+      return try Keychain.fetchPassword(forAccount: self.accountNameForUserId(userId))
+    } catch {
+      Crashlytics.crashlytics().record(error: error)
+    }
+
+    return nil
+  }
+
   // Returns the last saved environment from user defaults.
   public static func fromStorage(
     ubiquitousStore _: KeyValueStoreType,
@@ -303,13 +339,21 @@ public struct AppEnvironment: AppEnvironmentType {
     let data = userDefaults.dictionary(forKey: self.environmentStorageKey) ?? [:]
 
     var service = self.current.apiService
-    var currentUser: User?
+    var currentUser: User? // Will only be set if an OAuth token is also set
     let configDict: [String: Any]? = data["config"] as? [String: Any]
     let config: Config? = configDict.flatMap(Config.decodeJSONDictionary)
 
-    if let oauthToken = data["apiService.oauthToken.token"] as? String {
-      // If there is an oauth token stored in the defaults, then we can authenticate our api service
+    let userFromDefaults: User? = data["currentUser"].flatMap(tryDecode)
+
+    // If there is an oauth token stored, then we can authenticate our api service
+
+    if let oauthToken = fetchOAuthTokenFromKeychain(forUserId: userFromDefaults?.id) {
       service = service.login(OauthToken(token: oauthToken))
+    } else if let oauthToken = data["apiService.oauthToken.token"] as? String {
+      service = service.login(OauthToken(token: oauthToken))
+
+      // Move it over to the keychain if we can
+      _ = self.storeOAuthTokenToKeychain(oauthToken, forUserId: userFromDefaults?.id)
     }
 
     // Try restoring the client id for the api service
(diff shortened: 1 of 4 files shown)


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:460:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:461:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:462:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

Library/AppEnvironment.swift
```diff
@@ -295,6 +296,41 @@ public struct AppEnvironment: AppEnvironmentType {
     )
   }
 
+  internal static func accountNameForUserId(_ userId: Int) -> String {
+    return "kickstarter_\(userId)"
```
amy-at-kickstarter, line 300: This is something I went back-and-forth on. The keychain is designed to store passwords for accounts. However, the way our app works right now, we only ever store one OAuth token in the user defaults - we store it for the current user, and remove it when you log out. 

So I had two choices here. I could name the keychain account something like `kickstarter_current_user` and only have one, or I could name it after the _specific_ user who is logged in. 

I decided to keep it linked to the specific user, but there's an argument to be made that YAGNI and I'm over-complicating things.

Library/AppEnvironment.swift
```diff
@@ -303,13 +339,21 @@ public struct AppEnvironment: AppEnvironmentType {
     let data = userDefaults.dictionary(forKey: self.environmentStorageKey) ?? [:]
 
     var service = self.current.apiService
-    var currentUser: User?
+    var currentUser: User? // Will only be set if an OAuth token is also set
```
amy-at-kickstarter, line 342: Something that is a little confusing now is that you need both the `User` and the token to log in, but they're stored in different places 🤯

Library/AppEnvironment.swift
```diff
@@ -303,13 +339,21 @@ public struct AppEnvironment: AppEnvironmentType {
...
     let configDict: [String: Any]? = data["config"] as? [String: Any]
     let config: Config? = configDict.flatMap(Config.decodeJSONDictionary)
 
-    if let oauthToken = data["apiService.oauthToken.token"] as? String {
-      // If there is an oauth token stored in the defaults, then we can authenticate our api service
+    let userFromDefaults: User? = data["currentUser"].flatMap(tryDecode)
+
+    // If there is an oauth token stored, then we can authenticate our api service
+
+    if let oauthToken = fetchOAuthTokenFromKeychain(forUserId: userFromDefaults?.id) {
```
amy-at-kickstarter, line 350: n.B. the feature flag check is inside this method

Library/AppEnvironment.swift
```diff
@@ -404,7 +448,14 @@ public struct AppEnvironment: AppEnvironmentType {
     var data: [String: Any] = [:]
 
     // swiftformat:disable wrap
-    data["apiService.oauthToken.token"] = env.apiService.oauthToken?.token
+
+    if let oauthToken = env.apiService.oauthToken?.token {
+      // Try to save to the keychain, but if that fails, save to user defaults
+      if !self.storeOAuthTokenToKeychain(oauthToken, forUserId: env.currentUser?.id) {
```
amy-at-kickstarter, line 454: Again, this will fail if the feature flag is off, and it will resume saving it to user defaults.

Library/AppEnvironmentTests+OAuthInKeychain.swift
```diff
@@ -0,0 +1,422 @@
+@testable import KsApi
+@testable import Library
+import XCTest
+
+final class AppEnvironmentTests_OAuthInKeychain: XCTestCase {
```
amy-at-kickstarter, line 5: These are all tests run in the hosted environment, so they can actually test the keychain. I included a bunch of duplicates of tests in `AppEnvironmentTests` that just swap the feature flag to 'on' to correctly test that behavior.

Library/AppEnvironmentTests+OAuthInKeychain.swift
```diff
@@ -0,0 +1,422 @@
...
+  func setFeatureUseKeychainEnabled(_ setting: Bool) {
+    let mockConfigClient = MockRemoteConfigClient()
+    mockConfigClient.features = [
+      RemoteConfigFeature.useKeychainForOAuthToken.rawValue: setting
+    ]
+
+    /* Awkward bit of condition here - our feature flag values implicitly come from
+     AppEnvironment.current, which in some cases, is the PREVIOUS environment in the stack.
+     If you're pushing/saving a new stack, that can be confusing. */
+    AppEnvironment.updateRemoteConfigClient(mockConfigClient)
```
amy-at-kickstarter, line 20: This was an interesting mess I got myself into - what does it mean to test a feature flag from within `AppEnvironment`, which is _where we store the feature flags_?

Library/AppEnvironmentTests.swift
```diff
@@ -81,7 +93,9 @@ final class AppEnvironmentTests: XCTestCase {
     AppEnvironment.popEnvironment()
   }
 
-  func testFromStorage_WithNothingStored() {
+  func testFromStorage_WithNothingStored_featureUseKeychainEnabledIsFalse() {
+    self.setFeatureUseKeychainEnabled(false)
```
amy-at-kickstarter, line 97: So now we have tests running in three different states:

`AppEnvironmentTests` running in the un-hosted test environment, with the feature flag OFF. This lets us run the original `AppEnvironmentTests` without breaking all of them.
`AppEnvironmentTests` running in the hosted test environment, with the feature flag OFF. This simulates the app with the feature flag off.
`AppEnvironmentTests+OAuthInKeychain` running in the hosted test environment, with the feature flag ON. This simulates the app with the feature flag on.

ifosli (approved):
I have some questions but overall, looks good!

Library/AppEnvironment.swift
amy-at-kickstarter: Yeah, I'm not sure I see a need any more, either. My original thinking was "hey, this way we can support multiple users in the future!" but I'm starting to think that's not ... particularly useful. You could definitely talk me into making it a single key. 

As for the naming scheme, great question. We do [use the bundle ID in Keychain.swift](https://github.com/kickstarter/ios-oss/blob/main/Library/Keychain.swift#L21), which should identify it uniquely, but I haven't actually tested to make sure that it doesn't overlap between build environments. I think you have to do _extra_ work to enable keychain sharing between app IDs but yeah, let's validate that.

Library/AppEnvironment.swift
```diff
@@ -303,13 +339,21 @@ public struct AppEnvironment: AppEnvironmentType {
...
+
+    // If there is an oauth token stored, then we can authenticate our api service
+
+    if let oauthToken = fetchOAuthTokenFromKeychain(forUserId: userFromDefaults?.id) {
       service = service.login(OauthToken(token: oauthToken))
+    } else if let oauthToken = data["apiService.oauthToken.token"] as? String {
+      service = service.login(OauthToken(token: oauthToken))
+
+      // Move it over to the keychain if we can
+      _ = self.storeOAuthTokenToKeychain(oauthToken, forUserId: userFromDefaults?.id)
```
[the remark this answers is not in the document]
amy-at-kickstarter: Hmmm very good point about just keeping it in `saveEnvironment`! Let me check, that would be cleaner.

Library/AppEnvironmentTests+OAuthInKeychain.swift
```diff
@@ -0,0 +1,422 @@
...
+
+    XCTAssertEqual("http://api.ksr.com", result["apiService.serverConfig.apiBaseUrl"] as? String)
+    XCTAssertEqual("cafebeef", result["apiService.serverConfig.apiClientAuth.clientId"] as? String)
+    XCTAssertEqual("http://ksr.com", result["apiService.serverConfig.webBaseUrl"] as? String)
+    XCTAssertEqual("en", result["apiService.language"] as? String)
+    XCTAssertEqual(User.template.id, (result["currentUser"] as? [String: AnyObject])?["id"] as? Int)
+
+    XCTAssertEqual(
+      nil, ubiquitousStore.string(forKey: AppEnvironment.oauthTokenStorageKey),
+      "No token stored."
```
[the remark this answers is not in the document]
amy-at-kickstarter: Hoo boy, that's a good question. As far as I know it was _never_ stored in the `ubiquitousStore`, only in `userDefaults`? So this must be an ancient failsafe for that case that I copy-pasted. I can fix the message for sure.

Library/AppEnvironment.swift
amy-at-kickstarter: Double-checked and yes, we're already using the bundle ID when we save the key. So there won't be any overlap.

Library/AppEnvironment.swift
```diff
@@ -142,10 +143,12 @@ public struct AppEnvironment: AppEnvironmentType {
...
 
   internal static func resetStackForUnitTests() {
-    while self.stack.popLast() != nil {}
+    while self.stack.count > 1 {
+      _ = self.stack.popLast()
+    }
 
     let next = Environment()
-    self.pushEnvironment(next)
+    self.replaceCurrentEnvironment(next)
```
amy-at-kickstarter, line 151: This fixes a crash in `resetStackForUnitTests` because it will never be truly empty. If it's empty, calls to `isFeatureEnabled` will fail.

amy-at-kickstarter:
Regarding testing on device: I'm going to deploy the beta and then run the tests that way. I won't mark the ticket as complete until I'm happy with on-device testing.

PR #1965, opened 2024-03-04
MBL-1211: Pass down pledge total to thank you page for new page copy

amy-at-kickstarter:
# 📲 What

Pass the pledge total into the thank you page.

# 🤔 Why

The new string for post-campaign backings on the thank-you page will reference the pledge amount, so we need it in the view model.

### DIFF
--- a/Library/ViewModels/PledgeViewModel.swift
+++ b/Library/ViewModels/PledgeViewModel.swift
@@ -905,7 +905,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         isApplePay: isApplePay
       )
 
-      return (data.project, baseReward, checkoutPropsData)
+      return (data.project, baseReward, checkoutPropsData, data.pledgeTotal)
     }
     .skipNil()
 
--- a/Library/ViewModels/ThanksViewModel.swift
+++ b/Library/ViewModels/ThanksViewModel.swift
@@ -7,7 +7,8 @@ import UIKit
 public typealias ThanksPageData = (
   project: Project,
   reward: Reward,
-  checkoutData: KSRAnalytics.CheckoutPropertiesData?
+  checkoutData: KSRAnalytics.CheckoutPropertiesData?,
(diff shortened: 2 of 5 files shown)


Library/ViewModels/ThanksViewModel.swift
```diff
@@ -77,21 +78,25 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
       .skipNil()
       .map { ($0.reward, $0.checkoutData) }
 
-    self.backedProjectText = project.map {
-      let string = Strings.You_have_successfully_backed_project_html(
-        project_name: $0.name
-      )
+    self.backedProjectText = self.configureWithDataProperty.signal
+      .skipNil()
+      .map { project, _, _, _ in
```
amy-at-kickstarter, line 89: Fourth arg here will be the pledge total.

amy-at-kickstarter:
Whoops, converting to draft while I fix tests.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -905,7 +905,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
         isApplePay: isApplePay
       )
 
-      return (data.project, baseReward, checkoutPropsData)
+      return (data.project, baseReward, checkoutPropsData, data.pledgeTotal)
```
amy-at-kickstarter, line 908: One line to change it, another 40 lines to fix the tests about it 😆

scottkicks (changes requested):
oops disregard this

Library/ViewModels/ThanksViewModel.swift
scottkicks: Is the plan to then gate the string from within this output?

amy-at-kickstarter: @scottkicks Correct; this is just plumbing to get all the data where it needs to go.

PR #1966, opened 2024-03-04
MBL-1211: Rewire current user email in AppEnvironment/AppDelegateViewModel

amy-at-kickstarter:
# 📲 What

Fix `AppEnvironment.current.currentUserEmail`, which did not appear to be correctly saving the currently logged-in user's e-mail.

# 🤔 Why

We need the user's e-mail on the Thanks page for post-campaign backings. I noticed that we had this `currentUserEmail` property on `Environment`, but it wasn't working correctly. I did some digging and I figured it's because the signal is being deallocated before the value is set.

An interesting question is why `currentUserEmail` isn't just part of `currentUser`. Apparently the V1 `User` object doesn't appear to have an e-mail attached to it, and the e-mail is coming from GraphQL, instead. It's a bit messy but I wanted to keep this change fairly low-profile instead of doing more refactoring.

### DIFF
--- a/Library/AppEnvironment.swift
+++ b/Library/AppEnvironment.swift
@@ -107,6 +107,7 @@ public struct AppEnvironment: AppEnvironmentType {
       apiService: AppEnvironment.current.apiService.logout(),
       cache: type(of: AppEnvironment.current.cache).init(),
       currentUser: nil,
+      currentUserEmail: nil,
       ksrAnalytics: self.current.ksrAnalytics |> KSRAnalytics.lens.loggedInUser .~ nil
     )
   }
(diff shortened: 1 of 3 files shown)


Library/AppEnvironment.swift
```diff
@@ -107,6 +107,7 @@ public struct AppEnvironment: AppEnvironmentType {
       apiService: AppEnvironment.current.apiService.logout(),
       cache: type(of: AppEnvironment.current.cache).init(),
       currentUser: nil,
+      currentUserEmail: nil,
```
amy-at-kickstarter, line 110: Clear it on logout, too.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/AppEnvironment.swift:467:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:468:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/AppEnvironment.swift:469:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

scottkicks (approved):
We should make note of this change for when we regression test the release since AppDelegate changes have a wide impact

PR #1968, opened 2024-03-05
MBL-1211: Add new thank-you page copy for post-campaign pledges

amy-at-kickstarter:
# 📲 What

Adds new copy to the thank you page for post-campaign pledges.

# 👀 See
Screenshot 2024-03-05 at 5 22 21 PM

### DIFF
--- a/Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
+++ b/Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
@@ -796,6 +796,7 @@
 "You_canceled_your_pledge_for_this_project" = "You canceled your pledge for this project.";
 "You_cant_use_this_credit_card_to_back_a_project_from_project_country" = "You can’t use this credit card to back a project from %{project_country}.";
 "You_have_successfully_backed_project_html" = "You have successfully backed <b>%{project_name}</b>. This project is now one step closer to a reality, thanks to you. Spread the word!";
+"You_have_successfully_backed_project_post_campaign_html" = "<p>You have successfully backed <b>%{project_name}</b>. Your pledge of <b>%{pledge_total}</b> has been collected.</p>\n<p>You’ll receive a confirmation email at %{user_email} when your rewards are ready to fulfill so that you can finalize and pay shipping and tax.</p>\n<p>This project is now one step closer to a reality, thanks to you. Spread the word!</p>";
 "You_launched_this_project_on_launch_date" = "You launched this project on %{launch_date}. For creator tools, head to our website!";
 "You_need_to_pledge_at_least_reward_minimum_for_this_reward" = "You need to pledge at least %{reward_minimum} for this reward.";
 "You_pledged_on_date" = "<b>You pledged</b> on %{pledge_date}";
--- a/Library/ViewModels/ThanksViewModel.swift
+++ b/Library/ViewModels/ThanksViewModel.swift
@@ -86,11 +86,25 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs, 
 
     self.backedProjectText = self.configureWithDataProperty.signal
       .skipNil()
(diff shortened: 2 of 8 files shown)


Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
```diff
@@ -796,6 +796,7 @@
 "You_canceled_your_pledge_for_this_project" = "You canceled your pledge for this project.";
 "You_cant_use_this_credit_card_to_back_a_project_from_project_country" = "You can’t use this credit card to back a project from %{project_country}.";
 "You_have_successfully_backed_project_html" = "You have successfully backed <b>%{project_name}</b>. This project is now one step closer to a reality, thanks to you. Spread the word!";
+"You_have_successfully_backed_project_post_campaign_html" = "<p>You have successfully backed <b>%{project_name}</b>. Your pledge of <b>%{pledge_total}</b> has been collected.</p>\n<p>You’ll receive a confirmation email at %{user_email} when your rewards are ready to fulfill so that you can finalize and pay shipping and tax.</p>\n<p>This project is now one step closer to a reality, thanks to you. Spread the word!</p>";
```
amy-at-kickstarter, line 799: Note the `<p>` tags. Because we interpret this string as HTML, raw newlines (`\n`) are stripped out. Instead, we have to actually treat it like HTML to get the formatting we want! 

I think `NSAttributedString` has some tools for changing newline spacing, if we need to, but I think the default behavior is fine for now.

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -86,11 +86,25 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
-          project_name: project.name
-        )
+      .map { project, _, _, pledgeTotal in
+
+        let string: String
+
+        if featurePostCampaignPledgeEnabled(),
+          project.isInPostCampaignPledgingPhase,
+          let email = AppEnvironment.current.currentUserEmail {
+          let formattedTotal = Format.formattedCurrency(pledgeTotal, country: project.country)
```
amy-at-kickstarter, line 96: @scottkicks This is my first foray into formatting monetary strings in the app. Am I pulling the correct fields? `project.country` seemed like the more immediate/correct option, versus `project.stats.country` but let me know if I'm missing something.

amy-at-kickstarter: As a note for future refactoring, we should really be passing around something like `Money` type that includes the currency with the amount. The raw double being passed around makes me nervous.

Library/ViewModels/ThanksViewModel.swift
```diff
@@ -86,11 +86,25 @@ public final class ThanksViewModel: ThanksViewModelType, ThanksViewModelInputs,
...
-        let string = Strings.You_have_successfully_backed_project_html(
-          project_name: project.name
-        )
+      .map { project, _, _, pledgeTotal in
+
+        let string: String
+
+        if featurePostCampaignPledgeEnabled(),
+          project.isInPostCampaignPledgingPhase,
+          let email = AppEnvironment.current.currentUserEmail {
```
amy-at-kickstarter, line 95: This dependency on `currentUserEmail` is a little bit awkward - but it also seemed like a shame to add another fetch to what is currently a static screen, especially when `currentUserEmail` already was supposed to be there...

Library/ViewModels/ThanksViewModel.swift
scottkicks: Yes, this is how currency formatting is being handled elsewhere in the checkout flow. I agree it's not the best. I'm pushing to get some time allocated after the first release of PCP to make refactors like this. It's a little too in the weeds right now, I think.

scottkicks (approved):
Nice work!

PR #1969, opened 2024-03-06
[MBL-1208] Part 1: Initial ConfirmDetailsViewController | Shipping Location + Pledge/Bonus Steppers

scottkicks:
# 📲 What
Part 1 in the breakdown of adding a new [Confirm Details Screen](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=73-5208&mode=design&t=30TTqFlNfFrutqCM-0)
When a user selects an option in the rewards screen, navigate to the new screen. 
- To keep these PRs small, the new view controller, right now, only displays the shipping location selector and pledge/bonus steppers. See the [designs](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=73-5038&mode=design&t=30TTqFlNfFrutqCM-0) for reference.

# 🤔 Why

This screen will allow backers to confirm their pledge info before reaching the final checkout screen. This allows us to call PCP validation mutations before creating stipe intents and initializing the payment sheet.

# 🛠 How

This initiative is time-sensitive and is part of our more complicated and very important checkout flow. With that in mind, I'm choosing to follow existing patterns closely rather than spending more time refactoring or using new patterns like SwiftUI or Combine.

Navigation sources: `RewardAddOnSelectionViewController` and `RewardsCollectionViewController`

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2024-03-06 at 09 28 46

# ✅ Acceptance criteria

- [x] Navigating to new VC when selecting a $1 pledge with no reward only shows the pledge stepper
- [x] Navigating to new VC when selecting a pledge with a digital reward and no add-ons shows only the bonus stepper
- [x] Navigating to new VC when selecting a pledge with a physical reward and no add-ons shows the shipping dropdown selector and bonus stepper
- [x] Navigating to new VC when selecting a pledge with a physical reward that includes add-ons shows the static dropdown info and bonus stepper

# ⏰ TODO

- [x] Unit tests for the view model. This is pretty involved so I wanted to get the review on this started in the meantime
Part 2 will be all about adding the [pledge total section](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=73-5025&mode=design&t=Nz3VCG5iDaQPeLjn-4) https://github.com/kickstarter/ios-oss/pull/1970

PR #1970, opened 2024-03-07
[MBL-1208] Part 2: Pledge Total Summary Section

scottkicks:
# 📲 What

Displays the pledge total, with conversion, under the "bonus" or "update pledge" inputs

# 🤔 Why

When there isn't a reward, or the reward is digital and has no shipping info, we want to display this "total summary" section instead of the pledge summary table (see [designs](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=296-633&mode=design&t=ENjUnnfpltWps4tL-4). This table will be added in the following PR).

# 🛠 How

We already display this view on the current checkout screen (PledgeViewController). I created a new one titled `PostCampaignPledgeSummaryViewController` that is almost identical. Having new components within the PostCampaignPledge context will make it easier to remove the old flow in the future and help this flow be as clearer.

# 👀 See
[Figma Designs](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=296-633&mode=design&t=ENjUnnfpltWps4tL-4)
Simulator Screen Recording - iPhone 15 Pro - 2024-03-07 at 09 42 27

# ✅ Acceptance criteria

- [x] Total summary section displays when no reward ($1 pledge) is selected
- [x] Total summary section displays when a digital reward with no add-ons is selected
- [x] Total summary section **does not** display when a physical reward, with _or_  without add-ons is selected

# ⏰ TODO
- [x] Update once https://github.com/kickstarter/ios-oss/pull/1969 is merged.
- Next up will be the Pledge Rewards Summary Table View


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/ConfirmDetails/Controllers/PostCampaignPledgeSummaryViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+  func textView(
+    _: UITextView, shouldInteractWith url: URL, in _: NSRange,
+    interaction _: UITextItemInteraction
+  ) -> Bool {
+    self.viewModel.inputs.tapped(url)
+    return false
+  }
+}
+
+// MARK: - Styles
```
ifosli, line 115: Nit: I find styles to be more readable and they compile faster when we don't use prelude to define them. Thoughts on using standard UI kit (ex using helper functions) to style them instead?

Kickstarter-iOS/Features/ConfirmDetails/Controllers/PostCampaignPledgeSummaryViewController.swift
```diff
@@ -0,0 +1,171 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+final class PostCampaignPledgeSummaryViewController: UIViewController {
```
ifosli, line 6: I don't understand why we need a separate post campaign version of this class. The styling looks the same. If it's just hiding the "If the project reaches its funding goal" string that's the difference, we should just add an option to show/hide that instead of making a separate copy of the class. If there are other notable changes between this file and the standard pledge summary, can you call them out?

scottkicks: Yea there aren't more notable differences. I mostly wanted to maintain context and keep all of the PCP related components isolated so that removing the old/current flow would be easier

ifosli: I don't think we'd ever be able to remove the old version, though; the "If the project reaches its funding goal" text is still going to be relevant in the normal pledge flow, even if we do update the rest of the styling to match the PCP flow. So unless we want to keep two almost identical versions of this class in sync forever (which I really don't), we should keep the old class and instead modify it to make it support both cases.

ifosli (changes requested):

ifosli (approved):

PR #1972, opened 2024-03-07
MBL-1268: Check for 'canceled' param in login redirect

amy-at-kickstarter:
# 📲 What

Check if redirect was canceled.

# 🤔 Why

Otherwise, if the user cancels somewhere inside the webview, we would show an error message, which looks a little silly.
https://github.com/kickstarter/ios-oss/assets/146007185/2a073c11-be58-4217-9385-f29c608d19b6


Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -432,7 +432,7 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
         )
         alert.addAction(UIAlertAction(title: Strings.login_errors_button_ok(), style: .cancel))
         self?.present(alert, animated: true)
-      case .cancelled:
+      case .canceled:
```
amy-at-kickstarter, line 435: Renamed for sanity since the web param is also `canceled` with one `l`

scottkicks (approved):

PR #1974, opened 2024-03-11
MBL-1226: Add stripeCardId to UserStoredCardsFragment and CreditCardFragment

amy-at-kickstarter:
# 📲 What

Add `stripeCardId` to `UserStoredCardsFragment`

# 🤔 Why

From the ticket:
> When we end up calling Stripe.confirmPayment to process late pledges, we need to pass in a [paymentMethodId](https://stripe.dev/stripe-ios/stripe/documentation/stripe/stppaymentintentparams/paymentmethodid). This is accessible to us via the stripeCardId field that we can get in the UserStoredCardsFragment.

# 🛠 How

For the example IDs in all the templates in the tests, I used test cards I had on staging and pulled their IDs from GraphQL.

### DIFF
--- a/KsApi/models/DeletePaymentMethodEnvelopeTests.swift
+++ b/KsApi/models/DeletePaymentMethodEnvelopeTests.swift
@@ -13,13 +13,15 @@ class DeletePaymentMethodEnvelopeTests: XCTestCase {
                   "expirationDate": "2055-11-01",
                   "id": "1057",
                   "lastFour": "1111",
-                  "type": "VISA"
+                  "type": "VISA",
+                  "stripeCardId": "pm_1OtGFX4VvJ2PtfhK3Gp00SWK",
                 },
                 {
                   "expirationDate": "2044-04-01",
                   "id": "1055",
                   "lastFour": "4444",
-                  "type": "MASTERCARD"
+                  "type": "MASTERCARD",
+                  "stripeCardId": "pm_1OpDEC4VvJ2PtfhKftPrpgJ2",
                 }
               ]
             }
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -16858,6 +16858,7 @@ public enum GraphAPI {
           id
           lastFour
           type
+          stripeCardId
         }
         totalCount
       }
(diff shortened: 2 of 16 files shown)


KsApi/models/DeletePaymentMethodEnvelopeTests.swift
```diff
@@ -13,13 +13,15 @@ class DeletePaymentMethodEnvelopeTests: XCTestCase {
                   "expirationDate": "2055-11-01",
                   "id": "1057",
                   "lastFour": "1111",
-                  "type": "VISA"
+                  "type": "VISA",
+                  "stripeCardId": "pm_1OtGFX4VvJ2PtfhK3Gp00SWK",
```
amy-at-kickstarter, line 17: I erred on adding this field everywhere that references `storedCards`, even if the test didn't break without the field. I'd rather the data was accurate to what we'll get back from GraphQL.

scottkicks (changes requested):
Nice one! I'm actually realizing now that we need to do this in the `CreditCardFragment` too. 
When a user is adds a new payment method, `CreatePaymentSource` is called. It returns this fragment that we can use to pass `stripeCardId` to Stripe.confirmPayment. Sorry it wasn't in the ticket.

```swift
fragment CreditCardFragment on PaymentSource {
  ... on CreditCard {
    expirationDate
    id
    lastFour
    paymentType
    state
    type
    ---------
    stripeCardId
    ---------
  }
}
```

amy-at-kickstarter:
> Nice one! I'm actually realizing now that we need to do this in the `CreditCardFragment` too. When a user is adds a new payment method, `CreatePaymentSource` is called. It returns this fragment that we can use to pass `stripeCardId` to Stripe.confirmPayment. Sorry it wasn't in the ticket.
> 
> ```swift
> fragment CreditCardFragment on PaymentSource {
>   ... on CreditCard {
>     expirationDate
>     id
>     lastFour
>     paymentType
>     state
>     type
>     ---------
>     stripeCardId
>     ---------
>   }
> }
> ```

Ah, will do and then will ping for an updated review.

amy-at-kickstarter:
@scottkicks OK, added it to the other fragment.

FWIW, there are two GraphQL adapters for these two fragments - but I want to wait and see if we need to add this code to the adapters before modifying those. That should be a pretty obvious/easy change when we get there.

scottkicks (approved):
Thanks!

PR #1975, opened 2024-03-12
MBL-1204: Add explicit dependencies for Apollo code generation

amy-at-kickstarter:
# 📲 What

Add explicit input and output dependencies for our Apollo code generation script.

# 🤔 Why

By default, the Apollo code gen script ran every build. It would alter/touch `KsApi/GraphAPI.swift`, which would trigger the build pipeline to assume that the `KsApi` target had changed in some way. This extra build step could cascade into extra building, slowing down our build times.

Adding explicit dependencies means the Apollo script only runs when it actually needs to run, and should speed up our incremental builds.

# 🛠 How

This change has three parts:

1. Adding an input dependency `$(PROJECT_DIR)/KsApi/GraphQLFiles.xcfilelist` to the Apollo Code generation build step
2. Adding an output dependency `$(PROJECT_DIR)/KsApi/GraphAPI.swift` to the Apollo Code generation build step.
3. Adding a new build phase script, "Make list of GraphQL dependencies", which puts a list of all our `.graphql` files into the file list `$(PROJECT_DIR)/KsApi/GraphQLFiles.xcfilelist`

Step 3 is useful because Xcode has no way of using file wildcards (like `$PROJECT_DIR/KsApi/mutations/*.graphql`) for input dependencies. Instead, the documentation recommends making an `.xcfilelist` and updating that list. 

# 👀 See
To give it a whirl, try out the following!

1. Just hit build a bunch of times. It should feel much faster!
2. Add a field to a fragment! It should re-run the Apollo script. Then build again ... it shouldn't re-run the Apollo script!
3. Copy a query file (any query file). It should re-run the Apollo script. Then build again ... it shouldn't re-run the Apollo script! (Sometimes for me it builds the second time, but not the third - I'm happy with this; I'd rather err on over- rather than under- building.)


bin/find-graphql-files.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+GRAPHQL_FILE_LIST="$DERIVED_FILE_DIR/GraphQLFiles.xcfilelist"
+
+#Find all GraphQL files in KsApi.
+#Replace the literal $SRCROOT with the token $(SRCROOT)
+find $SRCROOT -name "*.graphql" | sed "s#^$SRCROOT#\$(SRCROOT)#" > "$GRAPHQL_FILE_LIST.tmp"
```
amy-at-kickstarter, line 7: This replace step isn't strictly necessary, but when I was debugging, I found it odd to have a list of fully qualified files in it. I like this slightly better. 🤷‍♀️

amy-at-kickstarter:
Whoops, moving back to draft while I fix CI.

bin/find-graphql-files.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+GRAPHQL_FILE_LIST="$PROJECT_DIR/KsApi/GraphQLFiles.xcfilelist"
+
+#Find all GraphQL files in KsApi.
+#Replace the literal $PROJECT_DIR with the token $(PROJECT_DIR)
+find $PROJECT_DIR -name "*.graphql" | sed "s#^$PROJECT_DIR#\$(PROJECT_DIR)#" > "$GRAPHQL_FILE_LIST.tmp"
```
amy-at-kickstarter, line 7: Replacing the actual project directory (like `/Users/amy/ios-oss/`) with the token means we're not always overwriting this file for one another.

ifosli (approved):
🎉 🎉 🎉

PR #1976, opened 2024-03-13
Fix regressions found in the login with OAuth flow

amy-at-kickstarter:
# 📲 What

1. Make the old login, old sign up, and new OAuth "signup or login" buttons reactive.
2. Make the default value of the `featureLoginWithOAuthEnabled` true.

# 🤔 Why

1. One copy of our login screen is created as a root view of the app - in the Profile tab in the tab bar. This copy of the screen was always getting the _first_ (ie cached or default value) of `featureLoginWithOAuthEnabled()`. This means it was never updating, even after new values of the OAuth feature flag loaded.  Fixing this bug ensures we can correctly use the feature flag for OAuth, in case we need to ramp down.
2. If the remote config hasn't loaded yet, we want to show the new flow, not the old flow.

### DIFF
--- a/Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
+++ b/Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
@@ -37,6 +37,10 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
     |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
+  private lazy var signupOrLoginWithOAuthButton = { UIButton(type: .custom)
+    |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
   private lazy var loginContextStackView = { UIStackView() }()
   private lazy var logoImageView = { UIImageView(frame: .zero) }()
   internal var processingView: ProcessingView? = ProcessingView(frame: .zero)
@@ -138,15 +142,12 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
       }
       |> UILabel.lens.text %~ { _ in Strings.Get_notified_when_your_friends_back_and_launch_projects() }
 
-    if self.viewModel.outputs.loginWithOAuthEnabled {
-      // TODO: Add and translate a new version of this string for this page.
-      _ = self.loginButton |> greenButtonStyle
-      self.loginButton
-        .setTitle(Strings.login_tout_generic_intent_traditional_signup_or_login_button(), for: .normal)
-    } else {
-      _ = self.loginButton |> greyButtonStyle
-      self.loginButton.setTitle(Strings.login_tout_back_intent_traditional_login_button(), for: .normal)
-    }
+    _ = self.loginButton |> greyButtonStyle
+    self.loginButton.setTitle(Strings.login_tout_back_intent_traditional_login_button(), for: .normal)
+
+    _ = self.signupOrLoginWithOAuthButton |> greenButtonStyle
+    self.signupOrLoginWithOAuthButton
+      .setTitle(Strings.login_tout_generic_intent_traditional_signup_or_login_button(), for: .normal)
 
     _ = self.loginContextStackView
       |> UIStackView.lens.spacing .~ Styles.gridHalf(1)
@@ -188,6 +189,12 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
       .observeForControllerAction()
       .observeValues { [weak self] _ in
         self?.pushSignupViewController()
+      }
+
+    self.viewModel.outputs.startOAuthSignupOrLogin
+      .observeForControllerAction()
+      .observeValues { [weak self] _ in
+        self?.pushOAuthFlow()
       }
 
     self.viewModel.outputs.logIntoEnvironmentWithApple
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -37,6 +37,10 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
     |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
+  private lazy var loginOrSignUpWithOAuthButton = { UIButton(type: .custom)
```
amy-at-kickstarter, line 40: Most of this boiler-plate code is splitting out the "Sign up or Log in" button from the "Log In" button, where it was stuffed earlier.

Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -410,13 +420,17 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
   }
 
   fileprivate func pushLoginViewController() {
-    if self.viewModel.outputs.loginWithOAuthEnabled, let session = createAuthorizationSession() {
-      session.presentationContextProvider = self
-      session.start()
-    } else {
-      self.navigationController?.pushViewController(LoginViewController.instantiate(), animated: true)
-      self.navigationItem.backBarButtonItem = UIBarButtonItem.back(nil, selector: nil)
+    self.navigationController?.pushViewController(LoginViewController.instantiate(), animated: true)
```
amy-at-kickstarter, line 423: This is now only called by the traditional login button.

Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -410,13 +420,17 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
...
+    self.navigationItem.backBarButtonItem = UIBarButtonItem.back(nil, selector: nil)
+  }
+
+  fileprivate func pushOAuthFlow() {
+    guard featureLoginWithOAuthEnabled(), let session = createAuthorizationSession() else {
```
amy-at-kickstarter, line 428: This is only called by the new login button.

Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -298,6 +305,10 @@ public final class LoginToutViewController: UIViewController, MFMailComposeViewC
     self.contextLabel.rac.text = self.viewModel.outputs.logInContextText
     self.bringCreativeProjectsToLifeLabel.rac.hidden = self.viewModel.outputs.headlineLabelHidden
 
+    self.loginButton.rac.hidden = self.viewModel.outputs.showLoginWithOAuth
```
amy-at-kickstarter, line 308: Either show the new, combined button, or the two old buttons.

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -1,10 +1,12 @@
...
 
-private func featureEnabled(feature: RemoteConfigFeature) -> Bool {
-  return AppEnvironment.current.userDefaults
-    .remoteConfigFeatureFlags[feature.rawValue] ??
-    (AppEnvironment.current.remoteConfigClient?
-      .isFeatureEnabled(featureKey: feature) ?? false)
+  let valueFromRemoteConfig = AppEnvironment.current.remoteConfigClient?
+    .isFeatureEnabled(featureKey: feature)
+
+  return valueFromDefaults ?? valueFromRemoteConfig ?? defaultValue
```
amy-at-kickstarter, line 9: Slight refactoring to support new default values, and to make this easier to read and debug.

Library/ViewModels/LoginToutViewModelTests.swift
```diff
@@ -117,6 +121,43 @@ final class LoginToutViewModelTests: TestCase {
     )
   }
 
+  func testStartSignupOrLoginWithOAuth() {
```
amy-at-kickstarter, line 124: Some simple/boilerplate tests, just for sanity.

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -1,10 +1,12 @@
...
+  let valueFromDefaults = AppEnvironment.current.userDefaults
+    .remoteConfigFeatureFlags[feature.rawValue]
 
-private func featureEnabled(feature: RemoteConfigFeature) -> Bool {
-  return AppEnvironment.current.userDefaults
-    .remoteConfigFeatureFlags[feature.rawValue] ??
-    (AppEnvironment.current.remoteConfigClient?
-      .isFeatureEnabled(featureKey: feature) ?? false)
+  let valueFromRemoteConfig = AppEnvironment.current.remoteConfigClient?
+    .isFeatureEnabled(featureKey: feature)
```
ifosli, line 7: nit: Unlike before, `valueFromRemoteConfig` will now be calculated each time. I wouldn't expect that to be a problem (it's probably cached in the remote config client and thus not an expensive operation) but just in case it is actually expensive or always does a network call or something, I think we should stick with early returns. Ex use `if let` blocks with returns instead of just having one return statement at the end.

ifosli (approved):
Looks good! I just have one minor concern in the remote config code

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
amy-at-kickstarter: Ah, good point - will fix then merge.

PR #1978, opened 2024-03-14
MBL-1181: Update ApplePay payment total string for post campaign pledges

amy-at-kickstarter:
# 📲 What

Update the payment total string in ApplePay for post campaign pledges.

# 🤔 Why

It says "Pay Kickstarter (if funded)" by default. For PCP, it can just say "Pay Kickstarter".

# 👀 See
Screenshot 2024-03-14 at 11 12 29 AM
Screenshot 2024-03-14 at 11 11 23 AM

### DIFF
--- a/Library/PKPaymentRequest+Helpers.swift
+++ b/Library/PKPaymentRequest+Helpers.swift
@@ -9,12 +9,11 @@ extension PKPaymentRequest {
     allRewardsTotal: Double,
     additionalPledgeAmount: Double,
     allRewardsShippingTotal: Double,
-    merchantIdentifier: String,
-    env: Environment = AppEnvironment.current
+    merchantIdentifier: String
   ) -> PKPaymentRequest {
     let request = PKPaymentRequest()
     request.merchantIdentifier = merchantIdentifier
-    request.supportedNetworks = env.applePayCapabilities.supportedNetworks(for: project)
+    request.supportedNetworks = AppEnvironment.current.applePayCapabilities.supportedNetworks(for: project)
     request.merchantCapabilities = .capability3DS
     request.countryCode = project.country.countryCode
     request.currencyCode = projectCountry(forCurrency: project.stats.currency)?.currencyCode ?? project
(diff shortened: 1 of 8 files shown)


Library/PKPaymentRequest+Helpers.swift
```diff
@@ -9,35 +9,34 @@ extension PKPaymentRequest {
     allRewardsTotal: Double,
     additionalPledgeAmount: Double,
     allRewardsShippingTotal: Double,
-    merchantIdentifier: String,
-    env: Environment = AppEnvironment.current
```
amy-at-kickstarter, line 13: This was always the default value, so just making it explicit. I'm also checking a feature flag in here which implicitly uses `AppEnvironment.current`.

Library/PKPaymentRequest+Helpers.swift
```diff
@@ -9,35 +9,34 @@ extension PKPaymentRequest {
...
-      allRewardsShippingTotal: allRewardsShippingTotal
+      allRewardsShippingTotal: allRewardsShippingTotal,
+      postCampaignPledgingActive: featurePostCampaignPledgeEnabled() && project.isInPostCampaignPledgingPhase
     )
 
     return request
   }
 
   private static func paymentSummaryItems(
-    forProject _: Project,
```
amy-at-kickstarter, line 36: Removed unused variable

Library/PKPaymentRequestHelpersTests.swift
```diff
@@ -7,8 +7,10 @@ import XCTest
 
 final class PKPaymentRequestHelpersTests: XCTestCase {
   func testPaymentRequest_NoShipping() {
-    let project = Project.template
-      |> Project.lens.country .~ .us
+    var project = Project.template
```
amy-at-kickstarter, line 10: Project is mutable, so just pulling out a little Prelude while I'm in here.

Library/Strings.swift
```diff
@@ -2269,18 +2269,18 @@ Haz clic para volver a intentarlo."
     )
   }
   /**
-   "Continue with %{quantity_count} add-ons"
+   "Continue with %{quantity_count} add-on"
```
amy-at-kickstarter, line 2272: These are weird. Did something happen to muck with plurals?

ifosli: I've seen similar changes before that weirded me out. Basically, we have multiple versions of this string depending on the count. The change here is in the comment that's just for our sake; not in the actual string presented to users. I have no idea what makes the comment choose the singular vs plural version, though. I assume the change to the default value is similar (not actually going to have an effect since we're not relying on the default), but I'm not sure.

ifosli (approved):

Library/Strings.swift
amy-at-kickstarter: Ooh, that makes much more sense. Cool!

PR #1980, opened 2024-03-18
[MBL-1208] Part 4: Continue CTA Section

scottkicks:
# 📲 What

Adds the CTA section to the bottom of the screen. 
Includes a continue button that will navigate to the final checkout screen and displays the current pledge total.

# 🤔 Why

We need a way to confirm pledge details.

# 🛠 How

creates a new `ConfirmDetailsContinueCTAView` and adds it to `ConfirmDetailsViewController` following the designs: https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=73-5025&mode=design&t=4iAXblqSygfASSvB-0

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2024-03-18 at 08 29 04

# ✅ Acceptance criteria

- [ ] CTA section matches designs
- [ ] Continue button doesn't do anything yet
- [ ] The pledge total is only displayed if the rewards summary table is displayed
- [ ] The pledge total updates appropriately when bonus, pledge, and shipping updates are made

### DIFF
--- a/Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
+++ b/Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
@@ -1,0 +1,163 @@
+import KsApi
+import Library
+import Prelude
+import UIKit
+
+public typealias ConfirmDetailsContinueCTAViewData = (
+  project: Project,
+  total: Double
+)
+
+private enum Layout {
+  enum Button {
+    static let minHeight: CGFloat = 48.0
+  }
+}
+
+protocol ConfirmDetailsContinueCTAViewDelegate: AnyObject {
+  func continueButtonTapped()
+}
+
+final class ConfirmDetailsContinueCTAView: UIView {
+  // MARK: - Properties
+
+  private lazy var rootStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private(set) lazy var titleAndAmountStackView: UIStackView = { UIStackView(frame: .zero) }()
+  private lazy var titleLabel: UILabel = { UILabel(frame: .zero) }()
+  private lazy var amountLabel: UILabel = { UILabel(frame: .zero) }()
+
+  private(set) lazy var continueButton: UIButton = {
+    UIButton(type: .custom)
+      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+  }()
+
+  public weak var delegate: ConfirmDetailsContinueCTAViewDelegate?
+
+  // MARK: - Lifecycle
+
+  override init(frame: CGRect) {
+    super.init(frame: frame)
+
+    self.configureSubviews()
+    self.setupConstraints()
+
+    self.continueButton.addTarget(
+      self, action: #selector(self.continueButtonTapped),
+      for: .touchUpInside
+    )
+  }
+
+  required init?(coder _: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: - Styles
+
+  override func bindStyles() {
+    super.bindStyles()
+
+    _ = self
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
```diff
@@ -0,0 +1,163 @@
...
+import UIKit
+
+public typealias ConfirmDetailsContinueCTAViewData = (
+  project: Project,
+  total: Double
+)
+
+private enum Layout {
+  enum Button {
+    static let minHeight: CGFloat = 48.0
```
amy-at-kickstarter, line 13: Is this not in our design system somewhere?

Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
```diff
@@ -0,0 +1,163 @@
...
+
+    _ = self.titleLabel
+      |> self.titleLabelStyle
+
+    _ = self.amountLabel
+      |> self.amountLabelStyle
+
+    _ = self.continueButton
+      |> greenButtonStyle
+      |> UIButton.lens.title(for: .normal) %~ { _ in Strings.Continue() }
```
amy-at-kickstarter, line 89: Nit/non-blocking: Mentioned this in an earlier PR but I'd prefer us to move away from Prelude code in `bindStyles` whenever we can avoid it.

Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
```diff
@@ -0,0 +1,163 @@
...
+    ])
+  }
+
+  func configure(with data: ConfirmDetailsContinueCTAViewData) {
+    if let attributedAmount = attributedCurrency(with: data.project, total: data.total) {
+      self.amountLabel.attributedText = attributedAmount
+    }
+  }
+
+  private func attributedCurrency(with project: Project, total: Double) -> NSAttributedString? {
```
amy-at-kickstarter, line 118: Extremely teeny nit: This method name would be clearer if it was `withProject:` instead of just `with:`

Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
```diff
@@ -0,0 +1,163 @@
...
+    let defaultAttributes = checkoutCurrencyDefaultAttributes()
+      .withAllValuesFrom([.foregroundColor: UIColor.ksr_support_700])
+    let projectCurrencyCountry = projectCountry(forCurrency: project.stats.currency) ?? project.country
```
amy-at-kickstarter, line 121: Do we expect this fall-through to ever have to happen? I wonder if we should just change the type of `project.currency` to be a forced non-nil value, since it would be pretty odd if we didn't have that information from the server.

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -310,6 +311,12 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
       pledgeTotalSummaryData
     )
     }
+
+    self.configureCTAWithPledgeTotal = Signal.combineLatest(
+      project,
+      pledgeTotal
+    )
+    .map { project, total in (project, total) }
```
amy-at-kickstarter, line 319: Nit: this map is a no-op, I think.

amy-at-kickstarter (approved):
LGTM with a few non-blocking suggestions.

PR #1981, opened 2024-03-18
MBL-1177: Add CompleteOnSessionCheckoutMutation to GraphAPI

amy-at-kickstarter:
# 📲 What

Add CompleteOnSessionCheckoutMutation to GraphAPI.

# 🤔 Why

This is a new mutation for post-campaign pledging. I modeled it after the mutation it will replace, `createBacking`, having it return a `CheckoutFragment`.

# 👀 See

[Trello, screenshots, external resources?](https://kickstarter.atlassian.net/wiki/spaces/PAYM/pages/2586443778/Validations+Approach+1+Separate+validations+from+Checkouts+Apply+service+client-side+confirmation#Mobile-proposal)


KsApi/MockService.swift
```diff
@@ -1892,3 +1913,9 @@ private extension Result {
     }
   }
 }
+
+extension GraphAPI.CompleteOnSessionCheckoutMutation.Data: Decodable {
+  public init(from _: Decoder) throws {
+    fatalError("The test code should not actually be decoding this object.")
```
amy-at-kickstarter, line 1919: This is required for us to use `performWithResult`. I'm fairly sure the `Decodable` requirement is for Apollo, not for the `MockGraphQLClient`, which doesn't actually do any decoding.

KsApi/ServiceType.swift
```diff
@@ -77,6 +77,9 @@ public protocol ServiceType {
   func createBacking(input: CreateBackingInput) ->
     SignalProducer<CreateBackingEnvelope, ErrorEnvelope>
 
+  func completeOnSessionCheckout(input: GraphAPI.CompleteOnSessionCheckoutInput) ->
+    SignalProducer<GraphAPI.CompleteOnSessionCheckoutMutation.Data, ErrorEnvelope>
```
amy-at-kickstarter, line 81: I'd like to move away from wrapper objects unless we really need them - implementing this as its raw GraphQL types for now; if we need to write a wrapper later, I'm OK doing that just-in-time.

scottkicks (approved):

PR #1982, opened 2024-03-18
[MBL-1273] Implement Create Checkout Mutation

scottkicks:
# 📲 What

When the CTA on Confirm Details is pressed, call CreateCheckout

# 🤔 Why

The ValidateCheckout Mutation that will be called on the checkout screen needs a `checkoutId` returned from this mutation. We'll call it here and if successful navigate to the checkout screen, passing along the `checkoutId`.

# 🛠 How

The checkout screen isn't complete yet so there's nothing to navigate to, but we can wire this new mutation up in the meantime.

# ✅ Acceptance criteria

- [x] Call mutation on CTA Tap
- [x] If there was an error, show the error message banner
- [x] If successful, print for now. We'll navigate later

# ⏰ TODO

- [ ] Navigate to Checkout screen when [mbl-1210](https://kickstarter.atlassian.net/browse/MBL-1210) is ready


Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -313,10 +317,51 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
+    )
+
+    let createCheckoutEvents = pledgeDetailsData
+      .takeWhen(self.continueCTATappedProperty.signal)
+      .map { project, rewards, pledgeTotal, refTag in
+        let rewardsIDs = rewards.map { $0.graphID }
+
+        return CreateCheckoutInput(
+          projectId: project.graphID,
+          amount: "\(pledgeTotal)",
```
amy-at-kickstarter, line 337: Presumably `pledgeTotal` is a double - does it need to be formatted in a specific way for the server? My concern would be that the automatic printer might output "7" or "7.0" or "7.000000001" and the server might expect a specific number of decimal places.

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -313,10 +317,51 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
+          locationId: "\(project.location.id)",
```
amy-at-kickstarter, line 338: Unconcerned about this one because it's an int.

amy-at-kickstarter (approved):
This LGTM, but only assuming that the print for `pledgeTotal` gives you something consistent/sane. 

I'd recommend using a more specific print formatter like `String.format("%.2d", pledgeTotal)` or `NumberFormatter` so that you're guaranteed consistent results.

PR #1983, opened 2024-03-18
Create PaginatingLazyVStack as a generic scrollable list container

amy-at-kickstarter:
# 📲 What

Add `PaginatingLazyVStack` to our component library.

# 🤔 Why

As part of SwiftUI work, I'm working on re-writing the old backer dashboard using SwiftUI. (Even if it never ships, I think the work will be educational.)

The main component of this page is going to be a scrolling list of cards. That seems like fairly generic functionality, so I wrote this paginating/refreshable `VStack` as a generic implementation of it.

# 👀 See

Here's what the example view looks like - it supports infinite scroll and pull-to-refresh.

paginating vstack example

### DIFF
--- a/Library/PaginatingLazyVStack.swift
+++ b/Library/PaginatingLazyVStack.swift
@@ -1,0 +1,120 @@
+import Foundation
+import SwiftUI
+
+// MARK: Implementation
+
+public struct PaginatingLazyVStack<Data: Identifiable, Cell: View>: View {
+  @Binding var data: [Data]
+  @Binding var canShowProgressView: Bool
+
+  let onRefresh: () async -> Void
+  let onDidShowProgressView: () -> Void
+  let configureCell: (Data) -> Cell
+
+  public init(
+    data: Binding<[Data]>,
+    canShowProgressView: Binding<Bool>,
+    onRefresh: @escaping () async -> Void,
+    onDidShowProgressView: @escaping () -> Void,
+    configureCell: @escaping (Data) -> Cell
+  ) {
+    self._data = data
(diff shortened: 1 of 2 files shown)


Library/PaginatingLazyVStack.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  private func refresh() async {
+    if #available(iOS 16.0, *) {
+      try? await Task.sleep(for: .seconds(1))
+    } else {}
+
+    self.data = []
+    self.addMore()
```
amy-at-kickstarter, line 114: I learned some interesting things about `Task` and `.refreshable` by doing this. Any modification to data that is bound to the `PaginatingLazyVStack` causes a re-render, which will cancel any tasks that are currently pending in `.refreshable`. That's what `self.data = []` has to happen after this `Task.sleep`; otherwise it's instantly re-rendered and canceled.

Library/PaginatingLazyVStack.swift
```diff
@@ -0,0 +1,120 @@
...
+
+// MARK: Implementation
+
+public struct PaginatingLazyVStack<Data: Identifiable, Cell: View>: View {
+  @Binding var data: [Data]
+  @Binding var canShowProgressView: Bool
+
+  let onRefresh: () async -> Void
+  let onDidShowProgressView: () -> Void
+  let configureCell: (Data) -> Cell
```
msrutek-paylocity, line 12: these should all be `private`

stevestreza-ksr:
Closing this as a revised version of this PR was merged in #2180.

PR #1985, opened 2024-03-19
Fix CircleCI error caused by GraphQL change to GraphAPI.FlaggingKind

amy-at-kickstarter:
# 📲 What

* Add graphql-schema.json to dependency list for Apollo update script
* Change type of CreateFlaggingInput from GraphAPI.FlaggingKind to GraphAPI.NonDeprecatedFlaggingKind

# 🤔 Why

Ingerid encountered this issue in a CircleCI build. Looks like we were running the Apollo update script on Circle, and it was making changes to `GraphAPI.swift` which broke her build. This fixes both the change to the schema, as well as makes sure that future updates to the schema will correctly cause a re-run of the Apollo build script.

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -483,7 +483,7 @@ public enum GraphAPI {
     ///   - kind
     ///   - details
     ///   - clientMutationId: A unique identifier for the client performing the mutation.
-    public init(contentId: GraphQLID, kind: FlaggingKind, details: Swift.Optional<String?> = nil, clientMutationId: Swift.Optional<String?> = nil) {
+    public init(contentId: GraphQLID, kind: NonDeprecatedFlaggingKind, details: Swift.Optional<String?> = nil, clientMutationId: Swift.Optional<String?> = nil) {
       graphQLMap = ["contentId": contentId, "kind": kind, "details": details, "clientMutationId": clientMutationId]
     }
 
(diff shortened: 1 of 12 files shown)


KsApi/GraphAPI.swift
```diff
@@ -810,7 +805,7 @@ public enum GraphAPI {
       }
     }
 
-    public static var allCases: [FlaggingKind] {
+    public static var allCases: [NonDeprecatedFlaggingKind] {
```
amy-at-kickstarter, line 808: This rename is the source of the problem.

ifosli (approved):

Library/ReportProjectInfoListItem.swift
```diff
@@ -84,7 +84,7 @@ let reportSpamSubListItems = [
   ),
   ReportProjectInfoListItem(
     type: .child,
-    flaggingKind: .abuse,
+    flaggingKind: .guidelinesAbuse,
```
amy-at-kickstarter, line 87: @scottkicks This is the one material change here - `NonDeprecatedFlaggingKind` does not include `.abuse` but it does include `.guidelinesAbuse`. What are these enums ultimately used for? Does it matter?

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ReportProjectInfoListItem.swift
amy-at-kickstarter: I just tested it on staging and was able to submit a report - I tried IP and Abuse. Both look sane:
Screenshot 2024-03-19 at 4 48 28 PM

scottkicks: They're used to determine the category reason why users are reporting. I think this should be fine

scottkicks (approved):

PR #1986, opened 2024-03-20
Rename PaymentMethodsViewController to PaymentMethodSettingsViewController for clarity

amy-at-kickstarter:
# 📲 What

Rename `PaymentMethodsViewController` to `PaymentMethodSettingsViewController`

# 🤔 Why

While looking at post-campaign backing changes, I got pretty confused why we have `PaymentMethodsViewController` as well as `PledgePaymentMethodsViewController`. Once I learned that `PaymentMethodsViewController` is only used in your user profile, it made much more sense. This rename should make that more obvious.


scottkicks (approved):
💯

PR #1987, opened 2024-03-20
[MBL-1217] Post Campaign Pledge Copy Translations

scottkicks:
# 📲 What

- Runs `make strings` to pull in new translations
- Updates the hardcoded strings to use the new localized ones

### DIFF
--- a/Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
+++ b/Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
@@ -109,6 +109,7 @@
 "Check_out_these_handpicked_projects" = "Check out these handpicked projects just for you.";
 "Check_your_inbox_to_complete_this_simple_step" = "Check your inbox to complete this simple step.";
 "Check_your_payment_details" = "Check your payment details";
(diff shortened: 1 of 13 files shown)


Kickstarter-iOS/Locales/Base.lproj/Localizable.strings
```diff
@@ -109,6 +109,7 @@
 "Check_out_these_handpicked_projects" = "Check out these handpicked projects just for you.";
 "Check_your_inbox_to_complete_this_simple_step" = "Check your inbox to complete this simple step.";
 "Check_your_payment_details" = "Check your payment details";
+"Checkout" = "Checkout";
```
ifosli, line 112: Do you want to update `PostCampaignCheckoutViewController` to use this string instead of the hardcoded `Checkout` in this pr, since that pr is merged now?

ifosli (approved):
Some of these strings aren't translated yet for all languages, so we'll need to double check to make sure translations are done and in our binary before we release. I'm good with updating them now anyways, though. Just update the "checkout" string to the localizable version, too, before submitting!

PR #1989, opened 2024-03-20
[MBL-1210] Update checkout pledge summary table

ifosli:
# 📲 What

Update the pledge summary table & navigation to the checkout screen to support no shipping and the no rewards case.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1210)

| No reward | Pledge with bonus |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] No rewards case shows correct cost
- [x] Shipping shows only when set
- [x] Bonus shows whenever it's set (and > 0).


scottkicks (approved):
Excellent

PR #1990, opened 2024-03-20
MBL-1123: Refactor PaymentSourceSelected to be an enum

amy-at-kickstarter:
# 📲 What

Refactor `PaymentSourceSelected` to be an `enum`, instead of a `struct`.

# 🤔 Why

For `MBL-1123`, I'm adding another type of payment source to `PledgeViewModel`. We start to create `PaymentIntent` secrets, which are a separate kind of client secret from our existing `SetupIntent` secrets. 

Before, the different between payment types was modeled by storing the payment source ID and a boolean, which flagged whether or not the string was a stripe token or a Kickstarter ID. 

After, this is modeled as an enum, so that it should be much clearer what kind of payment source we're passing around the codebase. As a bonus, this actually deleted a few lines of redundant code, since it makes the plumbing a little simpler.

When I finish this ticket, we'll have an enum like this:
```
public enum PaymentSourceSelected: Equatable {
  case paymentSourceId(String)
  case setupIntentClientSecret(String)
  case paymentIntentClientSecret(String)
}
```

### DIFF
--- a/Library/ViewModels/PaymentSourceSelected.swift
+++ b/Library/ViewModels/PaymentSourceSelected.swift
@@ -4,17 +4,19 @@ public enum PaymentSourceSelected: Equatable {
   case setupIntentClientSecret(String)
 
   public var paymentSourceId: String? {
-    if case let .paymentSourceId(value) = self {
+    switch self {
+    case let .paymentSourceId(value):
       return value
-    } else {
+    default:
       return nil
     }
   }
 
   public var setupIntentClientSecret: String? {
-    if case let .setupIntentClientSecret(value) = self {
+    switch self {
+    case let .setupIntentClientSecret(value):
       return value
-    } else {
+    default:
       return nil
     }
   }
--- a/Library/ViewModels/PaymentSourceSelected.swift
+++ b/Library/ViewModels/PaymentSourceSelected.swift
@@ -1,0 +1,21 @@
+import Foundation
+public enum PaymentSourceSelected: Equatable {
+  case paymentSourceId(String)
+  case setupIntentClientSecret(String)
+
+  public var paymentSourceId: String? {
+    if case let .paymentSourceId(value) = self {
+      return value
+    } else {
+      return nil
+    }
+  }
+
+  public var setupIntentClientSecret: String? {
+    if case let .setupIntentClientSecret(value) = self {
+      return value
+    } else {
+      return nil
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


Library/ViewModels/PaymentSourceSelected.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+public enum PaymentSourceSelected: Equatable {
+  case paymentSourceId(String)
```
amy-at-kickstarter, line 3: Suggestions for better names welcome - as far as I know, this is an ID that comes from GraphQL and represents a particular payment methods stored in our own DB.

Library/ViewModels/PaymentSourceSelected.swift
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  public var setupIntentClientSecret: String? {
+    if case let .setupIntentClientSecret(value) = self {
+      return value
+    } else {
+      return nil
+    }
+  }
+}
```
scottkicks, line 21: any particular reason you didn't want to use switch statements here?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -259,15 +259,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
-            paymentSourceId: selectedPaymentSheetPaymentMethodCardId,
-            isSetupIntentClientSecret: true
-          )
+          return PaymentSourceSelected.setupIntentClientSecret(selectedPaymentSheetPaymentMethodCardId)
         case let (.some(selectedPaymentMethodCardId), .none):
-          return PaymentSourceSelected(
-            paymentSourceId: selectedPaymentMethodCardId,
-            isSetupIntentClientSecret: false
-          )
+          return PaymentSourceSelected.paymentSourceId(selectedPaymentMethodCardId)
```
scottkicks, line 264: so much clearer!

scottkicks (approved):
These changes look good. Great cleanup. I suggest running through our normal pledge flow to make sure we're all good there before merging. Even though tests are passing and this is more of a cleanup, changing our checkout code makes me paranoid

Library/ViewModels/PaymentSourceSelected.swift
amy-at-kickstarter: Hm, not really. Just tried it out and it's a little neater looking, so I'll swap these.

amy-at-kickstarter:
Ran two manual tests on staging - I made a pledge with an existing card, and another pledge that was adding a new card.

PR #1991, opened 2024-03-20
[MBL-1210] Fix logged out and login flow for post campaign checkout

ifosli:
# 📲 What

Make "continue" button launch the login flow. Payment methods sheet should be hidden in logged out case and show up again once the user is logged in.

This makes the checkout VC a delegate of the pledge CTA and implements methods other than those corresponding to the "pledge" and "apple pay" buttons.

Note: This still hasn't added the extra warning to the pledge CTA. Since that can probably be added without touching the checkout VC, I'm going to do that in another pr to minimize merge conflicts.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1210)

Simulator Screen Recording - iPhone 15 Pro Max - 2024-03-20 at 16 02 58

# ✅ Acceptance criteria

- [x] Checkout screen looks good when the user is logged out
- [x] Checkout screen is back to normal when the user logs in by hitting "continue" button

### DIFF
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -216,7 +216,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
 
   // MARK: - Public functions
 
-  func setShouldCancelPaymentSheetAppearance(hidden: Bool) {
+  func cancelModalPresentation(_ hidden: Bool) {
     self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: hidden)
   }
 }
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -216,8 +216,8 @@ final class PledgePaymentMethodsViewController: UIViewController {
 
   // MARK: - Public functions
 
-  func cancelPaymentSheetAppearance() {
-    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: true)
+  func setShouldCancelPaymentSheetAppearance(hidden: Bool) {
+    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: hidden)
   }
 }
 
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -213,6 +213,12 @@ final class PledgePaymentMethodsViewController: UIViewController {
       self.tableView.reloadSections([PaymentMethodsTableViewSection.addNewCard.rawValue], with: .none)
     }
   }
+
+  // MARK: - Public functions
+
+  func cancelPaymentSheetAppearance() {
+    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: true)
+  }
 }
 
 // MARK: - UITableViewDelegate
@@ -229,13 +235,5 @@ extension PledgePaymentMethodsViewController: UITableViewDelegate {
     tableView.deselectRow(at: indexPath, animated: true)
 
     self.viewModel.inputs.didSelectRowAtIndexPath(indexPath)
-  }
-}
-
-// MARK: - PaymentSheetAppearanceDelegate
-
-extension PledgePaymentMethodsViewController: PaymentSheetAppearanceDelegate {
-  func pledgeViewControllerPaymentSheet(_: PledgeViewController, hidden: Bool) {
-    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: hidden)
   }
 }
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -234,22 +234,22 @@ extension PostCampaignCheckoutViewController: PledgeDisclaimerViewDelegate {
 
 extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate {
   func goToLoginSignup() {
-    self.paymentMethodsViewController.setShouldCancelPaymentSheetAppearance(hidden: true)
+    self.paymentMethodsViewController.cancelModalPresentation(true)
     self.viewModel.inputs.goToLoginSignupTapped()
   }
 
   func applePayButtonTapped() {
-    self.paymentMethodsViewController.setShouldCancelPaymentSheetAppearance(hidden: true)
+    self.paymentMethodsViewController.cancelModalPresentation(true)
     // TODO: Respond to button tap
   }
 
   func submitButtonTapped() {
-    self.paymentMethodsViewController.setShouldCancelPaymentSheetAppearance(hidden: true)
+    self.paymentMethodsViewController.cancelModalPresentation(true)
     // TODO: Respond to button tap
   }
 
   func termsOfUseTapped(with helpType: HelpType) {
-    self.paymentMethodsViewController.setShouldCancelPaymentSheetAppearance(hidden: true)
+    self.paymentMethodsViewController.cancelModalPresentation(true)
     self.viewModel.inputs.termsOfUseTapped(with: helpType)
   }
 }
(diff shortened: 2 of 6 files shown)


Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -213,6 +213,12 @@ final class PledgePaymentMethodsViewController: UIViewController {
       self.tableView.reloadSections([PaymentMethodsTableViewSection.addNewCard.rawValue], with: .none)
     }
   }
+
+  // MARK: - Public functions
+
+  func setShouldCancelPaymentSheetAppearance(hidden: Bool) {
+    self.viewModel.inputs.shouldCancelPaymentSheetAppearance(state: hidden)
```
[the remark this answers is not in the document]
amy-at-kickstarter: Ah, or is this more like a dismissal?

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -79,6 +81,10 @@ final class PostCampaignCheckoutViewController: UIViewController {
     self.viewModel.inputs.viewDidLoad()
   }
 
+  deinit {
+    self.sessionStartedObserver.doIfSome(NotificationCenter.default.removeObserver)
```
amy-at-kickstarter, line 85: Nifty, first time I've seen this in this codebase. Some kind of fancy RAC unsubscribe?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -289,6 +289,13 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
     self.shouldCancelPaymentSheetAppearance <~ updatedCards.signal
       .mapConst(true)
 
+    // Only ever use the value if the view model is configured and the payment sheet could exist.
+    let safeShouldCancelPaymentSheet = Signal.combineLatest(
+      self.shouldCancelPaymentSheetAppearance.signal,
+      configureWithValue
+    )
+    .map(first)
```
amy-at-kickstarter, line 297: Nit: Most patterns I've seen in our codebase call `configureWithValue` very early, like in `onAppear`. When does this issue happen? Just seems a bit odd to need the protection.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -52,16 +57,24 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
       .compactMap { data -> PledgePaymentMethodsValue? in
         guard let user = AppEnvironment.current.currentUser else { return nil }
         guard let reward = data.rewards.first else { return nil }
 
         return (user, data.project, reward, data.context, data.refTag)
       }
 
-    // TODO: Respond to login flow.
-    let isLoggedIn = initialData.ignoreValues()
+    self.goToLoginSignup = Signal.combineLatest(initialData, self.goToLoginSignupSignal)
```
amy-at-kickstarter, line 73: Is `Signal.combineLatest(a, b).map(first)` effectively just a `takeWhen`? If so, I find `takeWhen` clearer.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -70,11 +83,21 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        isLoggedIn: isLoggedIn,
+        isEnabled: true, // Pledge button never needs to be disabled on checkout page.
+        context: context,
+        willRetryPaymentMethod: false // Only retry in the `fixPaymentMethod` context.
+      )
     }
 
-    self.showWebHelp = self.pledgeDisclaimerViewDidTapLearnMoreSignal.mapConst(.trust)
+    self.paymentMethodsViewHidden = Signal.combineLatest(isLoggedIn, context)
+      .map { !$0 || $1.paymentMethodsViewHidden }
```
amy-at-kickstarter, line 95: Nit: Slightly hard to read with `$0`, would prefer that these variables were named so we had `!isLoggedIn || context.paymentMethodsViewHidden`

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -101,12 +124,27 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
     self.configureWithDataProperty.value = data
   }
 
+  private let (goToLoginSignupSignal, goToLoginSignupObserver) = Signal<Void, Never>.pipe()
```
amy-at-kickstarter, line 127: First time I've seen this, what's the goal of using `Signal<Void, Never>.pipe()`?

amy-at-kickstarter (approved):
Some nits and commentary, but nothing blocking. Otherwise LGTM.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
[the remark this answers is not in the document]
ifosli: I think this name is more clear than it was before but you're right - when I named it, I didn't know what it did either 😭. You're pretty much correct, though. Basically when you add a new card to the payment sheet, there's a loading indicator for a bit before the sheet pops up. Setting this to false (the only way it's used outside of tests) cancels that pop up (usually because we've presented a different modal instead). I _think_ it might also dismiss the sheet itself if it's already presented by the time this is called, but I'm not sure about that one since that would only happen as a race condition. If you have any ideas for a different name, though, let me know!

Library/ViewModels/PledgePaymentMethodsViewModel.swift
ifosli: It happens in the logged out state! If the user is logged out, the payment sheet doesn't actually get configured, but we still say "hey, please don't show that extra modal" if the user interacts with the pledge cta (ex: if they want to log in). Leaving it as it was causes the app to crash when that happens.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: I don't think so, because we want to update whenever either a or b has a new event. We could do a `Signal.merge(a, a.takeWhen(b))` instead, though I don't really think that's cleaner. I'm okay with switching to that if you like it better? I was just following our usual pattern for the `configureWithData` signal here (and copy-pasting from the normal pledge VC). Or if you have any better ideas for a thing that would make this doable in one command, let me know!

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: Good point, I'll update it!

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: Hmm, to be honest, this is also just copy-pasted from the standard pledge version of this VC/VM. But I think the point is that we want to react to button clicks - so we want to track whenever new events happen but we don't actually need it to pass along any data.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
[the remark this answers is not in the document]
amy-at-kickstarter: Hmmm...what about something like `dontShowPaymentSheet`?

Library/ViewModels/PledgePaymentMethodsViewModel.swift
amy-at-kickstarter: Aha, that makes sense! This would be a good comment, actually.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
amy-at-kickstarter: Nah, you're right, it's not any clearer, then. I guess it would be nice to have an operator like `take(value: someSignal, whenAnyChange: [someSignal, anotherSignal, aThirdSignal])` but that is vastly out of scope here.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
amy-at-kickstarter: Aah, gotcha. I think we mix `Void` and `()` throughout the codebase. Seems fine.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: You were right about this one, actually! I just mixed up use-cases and made us both confused 😅

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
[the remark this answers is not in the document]
ifosli: I'm just going to name it `cancelModalPresentation` I think. That's clear enough while still being general enough that you don't need to understand the paymentMethodsVC to understand what the goal is.

PR #1992, opened 2024-03-21
MBL-1123: Refactor PledgePaymentMethodsViewModel to use PaymentSourceSelected

amy-at-kickstarter:
# 📲 What

Refactor `PledgePaymentMethodsViewModel` to use `PaymentSourceSelected`, instead of managing separate values for a `selectedSetupIntent` and a `selectedCreditCard`. 

# 🤔 Why
For MBL-1123, I need to pass the final selected value (whether it's from a `SetupIntent`, a `PaymentIntent` or an existing `CreditCard`) back to the delegate of the `PledgePaymentMethodsViewController`. This refactor makes that significantly easier - we'll be able to keep the declarative nature of the screen, but be explicit about what kind of payment method is selected.

The alternative would be to keep some kind of state in the `PledgePaymentMethodsViewController` that indicated whether or not it should use a `SetupIntent` or a `PaymentIntent`, and that felt messy in a different way.

### DIFF
--- a/Library/ViewModels/PledgePaymentMethodsViewModel.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModel.swift
@@ -27,8 +27,7 @@ public typealias PaymentSheetSetupData = (
 public typealias PledgePaymentMethodsAndSelectionData = (
   paymentMethodsCellData: [PledgePaymentMethodCellData],
   paymentSheetPaymentMethodsCellData: [PaymentSheetPaymentMethodCellData],
-  selectedCard: UserCreditCards.CreditCard?,
-  selectedSetupIntent: String?,
+  selectedPaymentMethod: PaymentSourceSelected?,
   shouldReload: Bool,
   isLoading: Bool
 )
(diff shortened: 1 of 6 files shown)


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -253,19 +254,7 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
-        switch (selectedPaymentMethodCardId, selectedPaymentSheetPaymentMethodCardId) {
-        case let (.none, .some(selectedPaymentSheetPaymentMethodCardId)):
-          return PaymentSourceSelected.setupIntentClientSecret(selectedPaymentSheetPaymentMethodCardId)
-        case let (.some(selectedPaymentMethodCardId), .none):
-          return PaymentSourceSelected.paymentSourceId(selectedPaymentMethodCardId)
-        default:
-          return nil
-        }
-      }
+      .map { $0.selectedPaymentMethod }
```
amy-at-kickstarter, line 257: All this refactoring was to get this simplification!

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -49,11 +49,12 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
       .observe(self.reloadPaymentMethodsProjectCountry.observer)
     self.vm.outputs.reloadPaymentMethods.map { $0.1 }
       .observe(self.reloadPaymentSheetPaymentMethodsCards.observer)
-    self.vm.outputs.reloadPaymentMethods.map { $0.2 }.observe(self.reloadPaymentMethodsSelectedCard.observer)
-    self.vm.outputs.reloadPaymentMethods.map { $0.3 }
+    self.vm.outputs.reloadPaymentMethods.map { data in data.selectedPaymentMethod?.paymentSourceId }
```
amy-at-kickstarter, line 52: Instead of re-writing these tests to use the new-and-better output, just shimming the refactored code into the old test. We can make these tests prettier later, if we want.

scottkicks (approved):
LGTM 👍

PR #1996, opened 2024-03-25
Confirm Details UI Cleanup

scottkicks:
# 📲 What

- Removes Prelude elements from ConfirmDetailsViewController and it's CTA ViewController
- Expands Pledge Rewards Summary Table to full width
- Ensure that the Pledge Summary Section make sure the pledge summary is only shown in IsNoReward states when pledging
   - We need to hid this regardless of whether there is shipping enabled for a reward or not

# 🤔 Why

Cleanup is nice, and we need to match the figma

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2024-03-25 at 11 52 18

### DIFF
--- a/Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
+++ b/Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
@@ -56,37 +56,45 @@ final class ConfirmDetailsContinueCTAView: UIView {
   override func bindStyles() {
     super.bindStyles()
 
-    _ = self
-      |> \.layoutMargins .~ .init(all: Styles.grid(3))
+    self.layoutMargins = .init(all: Styles.grid(3))
 
-    _ = self.layer
-      |> checkoutLayerCardRoundedStyle
-      |> \.backgroundColor .~ UIColor.ksr_white.cgColor
-      |> \.shadowColor .~ UIColor.ksr_black.cgColor
-      |> \.shadowOpacity .~ 0.12
-      |> \.shadowOffset .~ CGSize(width: 0, height: -1.0)
-      |> \.shadowRadius .~ CGFloat(1.0)
-      |> \.maskedCorners .~ [
-        CACornerMask.layerMaxXMinYCorner,
-        CACornerMask.layerMinXMinYCorner
-      ]
+    self.layer.cornerRadius = 16.0
+    self.layer.backgroundColor = UIColor.ksr_white.cgColor
+    self.layer.shadowColor = UIColor.ksr_black.cgColor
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -88,9 +88,6 @@ final class PostCampaignCheckoutViewController: UIViewController {
   // MARK: - Configuration
 
   private func configureChildViewControllers() {
-    _ = (self.rootScrollView, self.view)
-      |> ksr_addSubviewToParent()
-
```
scottkicks, line 93: We don't need this anymore because rootScrollView is being added to the view in the line below

Kickstarter-iOS/Features/ConfirmDetails/Views/ConfirmDetailsContinueCTAView.swift
```diff
@@ -56,37 +56,45 @@ final class ConfirmDetailsContinueCTAView: UIView {
...
+
+    self.continueButton.setTitle(Strings.Continue(), for: .normal)
+    self.continueButton.setTitleColor(.ksr_white, for: .normal)
+    self.continueButton.setTitleColor(.ksr_white, for: .highlighted)
+    self.continueButton.setBackgroundColor(.ksr_create_700, for: .normal)
+    self.continueButton.setBackgroundColor(UIColor.ksr_create_700.mixDarker(0.36), for: .highlighted)
+    self.continueButton.setBackgroundColor(UIColor.ksr_create_700.mixLighter(0.36), for: .disabled)
+    self.continueButton.clipsToBounds = true
+    self.continueButton.layer.masksToBounds = true
+    self.continueButton.layer.cornerRadius = Styles.grid(2)
```
ifosli, line 97: I don't think this is the right way to move away from prelude. This copy-pastes code from `baseButtonStyle` and `greenButtonStyle`. In the long run, I think we want to convert these common styles into a different kind of utility method (ex static functions that take the button as input or maybe extensions on the UIButton class). For now, however, I think we should still use the prelude helpers, and just switch the additional styling unique to this class to standard UIKit. If we copy-paste it like this, we'll probably lose track of this by the time we switch the green button style, and it'll be more of a hassle to change that in the future (ex when we add dark mode support).

So I love the idea here and I think any styles that were just defined in this class (or prelude used directly) is worth getting rid of, but I don't want us to start copy-pasting common styling throughout the app.

ifosli (changes requested):
Looks good, apart from the fact that you're redefining some of the common styles in a way where I think fully migrating off of prelude and keeping these in sync will be more complicated than if we just use them for now.

PR #1999, opened 2024-03-26
[MBL-1194] Implement Validate Checkout 

scottkicks:
# 📲 What

When tapping pledge, we need to run this mutation so the backend can do some extra validations before we ask Stripe to do its thing.

**For New Payment Methods**
- We'll use the payment intent that is passed back to the VM via PaymentSourceSelected
- We can then call Stripe's STPAPIClient.shared.retrievePaymentIntent to get the required paymentSourceId (paymentMethodId in Stripe)

**For Pre-Existing Payment Methods that were originally created with a setup intent**
- We'll need to call Create Payment Intent again to pass to this mutation
- We'll also need to use the selected existing card's id to filter the user's stored cards and grab the `stripeCardId`

I opted to make separate calls to this mutation depending on the case since they require enough distinct information to warrant it. There might be an opportunity to remove some duplicate code so if there are any nits please call them out!

### The mutation accepts the following params:
- the **checkoutId** (from the https://github.com/kickstarter/ios-oss/pull/1982 on ConfirmDetailsViewController). One update to this was made. Validate Checkout requires that the checkoutId be decoded to base 64.
- the **paymentMethodId** that we can pull in using, `STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret)` for new cards, or the `stripeCardId` property on existing cards.
- and the **paymentIntentClientSecret** that we create using the backend.

### Extras
I also updated PostCampaignCheckoutViewController to use our `MessageBannerViewControllerPresenting` to show error messages when an API call fails.
   - It's important to note that in the case of a failure, the expected UX is to navigate the backer back to the ConfirmDetails screen so that we can being the process of creating a checkout Id/validating/etc. before attempting another pledge. This is because once we confirm the payment with Stripe, the backend will need to invalidate that request, essentially processing a refund to the backer and invalidating our previous validations.

# 🤔 Why

Gotta validate, ya dig?

# ✅ Acceptance criteria

- [x] Tapping Pledge, after adding a card or selecting an existing card, successfully runs the validateCheckout mutation with the correct values.

# ⏰ TODO

- [ ] Next step is to confirm the payment intent via Stripe.confirmPayment()

### DIFF
--- a/Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
+++ b/Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
@@ -135,11 +135,8 @@ final class ConfirmDetailsViewController: UIViewController, MessageBannerViewCon
   // MARK: - Configuration
 
   private func configureChildViewControllers() {
-    _ = (self.rootScrollView, self.view)
-      |> ksr_addSubviewToParent()
-
-    _ = (self.continueCTAView, self.view)
-      |> ksr_addSubviewToParent()
+    self.view.addSubview(self.rootScrollView)
+    self.view.addSubview(self.continueCTAView)
 
     _ = (self.rootStackView, self.rootScrollView)
       |> ksr_addSubviewToParent()
@@ -152,26 +149,21 @@ final class ConfirmDetailsViewController: UIViewController, MessageBannerViewCon
       self.pledgeRewardsSummaryViewController
     ]
 
-    let arrangedSubviews = [
-      self.rootInsetStackView
-    ]
+    self.rootStackView.addArrangedSubview(self.rootInsetStackView)
 
     let arrangedInsetSubviews = [
       [self.titleLabel],
       self.inputsSectionViews,
-      self.pledgeAmountSummarySectionViews,
-      [self.pledgeRewardsSummaryViewController.view]
+      self.pledgeAmountSummarySectionViews
     ]
     .flatMap { $0 }
     .compact()
 
-    arrangedSubviews.forEach { view in
-      self.rootStackView.addArrangedSubview(view)
-    }
-
     arrangedInsetSubviews.forEach { view in
       self.rootInsetStackView.addArrangedSubview(view)
     }
+
+    self.rootStackView.addArrangedSubview(self.pledgeRewardsSummaryViewController.view)
 
     childViewControllers.forEach { viewController in
       self.addChild(viewController)
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -19,7 +19,8 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
 
   private lazy var paymentMethodsViewController = {
     PledgePaymentMethodsViewController.instantiate()
-    // TODO: Add self as delegate and add support for delegate methods.
+      |> \.messageDisplayingDelegate .~ self
+      |> \.delegate .~ self
   }()
 
   private lazy var pledgeCTAContainerView: PledgeViewCTAContainerView = {
@@ -296,6 +297,17 @@ extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewController
     default:
       break
     }
+// MARK: - PledgeViewControllerMessageDisplaying
+
+extension PostCampaignCheckoutViewController: PledgeViewControllerMessageDisplaying {
+  func pledgeViewController(_: UIViewController, didErrorWith message: String) {
+    self.messageBannerViewController?.showBanner(with: .error, message: message)
+  }
+
+  func pledgeViewController(_: UIViewController, didSucceedWith message: String) {
+    self.messageBannerViewController?.showBanner(with: .success, message: message)
+  }
+}
     return self
   }
 }
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -9,8 +9,10 @@ private enum PostCampaignCheckoutLayout {
   }
 }
 
-final class PostCampaignCheckoutViewController: UIViewController {
+final class PostCampaignCheckoutViewController: UIViewController, MessageBannerViewControllerPresenting {
   // MARK: - Properties
+
+  internal var messageBannerViewController: MessageBannerViewController?
 
   private lazy var titleLabel = UILabel(frame: .zero)
 
@@ -70,6 +72,8 @@ final class PostCampaignCheckoutViewController: UIViewController {
       |> \.extendedLayoutIncludesOpaqueBars .~ true
 
     self.title = Strings.Back_this_project()
+
+    self.messageBannerViewController = self.configureMessageBannerViewController(on: self)
 
     self.configureChildViewControllers()
     self.setupConstraints()
@@ -196,6 +200,12 @@ final class PostCampaignCheckoutViewController: UIViewController {
       .observeValues { [weak self] helpType in
         guard let self = self else { return }
         self.presentHelpWebViewController(with: helpType, presentationStyle: .formSheet)
+      }
+
+    self.viewModel.outputs.showErrorBannerWithMessage
+      .observeForControllerAction()
+      .observeValues { [weak self] errorMessage in
+        self?.messageBannerViewController?.showBanner(with: .error, message: errorMessage)
       }
 
     self.sessionStartedObserver = NotificationCenter.default
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -299,7 +299,7 @@ extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewController
         }
 
         let stripeCardId = intent.stripeId
-        let paymentIntentClientSecret = paymentSource.paymentIntentClientSecret ?? ""
+        let paymentIntentClientSecret = paymentSource.paymentIntentClientSecret!
 
         self.viewModel.inputs.creditCardSelected(with: (stripeCardId, paymentIntentClientSecret))
       }
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -300,13 +300,11 @@ extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewController
     switch paymentSource {
     case let .paymentIntentClientSecret(clientSecret):
       return STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret) { intent, _ in
-        guard let intent = intent else {
+        guard let intent = intent, let paymentMethodId = intent.paymentMethodId else {
           self.messageBannerViewController?
             .showBanner(with: .error, message: Strings.Something_went_wrong_please_try_again())
           return
         }
-
-        let paymentMethodId = intent.paymentMethodId!
 
         self.viewModel.inputs
           .creditCardSelected(
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -263,3 +263,14 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
     self.viewModel.inputs.termsOfUseTapped(with: helpType)
   }
 }
+
+// MARK: - PledgePaymentMethodsViewControllerDelegate
+
+extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewControllerDelegate {
+  func pledgePaymentMethodsViewController(
+    _: PledgePaymentMethodsViewController,
+    didSelectCreditCard paymentSource: PaymentSourceSelected
+  ) {
+    self.viewModel.inputs.creditCardSelected(with: paymentSource)
+  }
+}
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -291,8 +291,9 @@ extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewController
   ) {
     switch paymentSource {
     case let .paymentIntentClientSecret(clientSecret):
-      return STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret) { intent, _ in
+      return STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret) { intent, error in
         guard let intent = intent else {
+          print(error?.localizedDescription)
           self.messageBannerViewController?
             .showBanner(with: .error, message: Strings.Something_went_wrong_please_try_again())
           return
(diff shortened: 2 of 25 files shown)


Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -353,7 +353,16 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
+      .map { values in
+        var checkoutId: String?
+
+        if let decoded = decodeBase64(values.checkout.id), let range = decoded.range(of: "Checkout-") {
+          let id = decoded[range.upperBound...]
+          checkoutId = String(id)
+        }
+
+        return checkoutId
+      }
```
scottkicks, line 365: These will always decode to include "Checkout-", but I'm not a big fan of hardcoding strings like this

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -253,11 +272,61 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
...
+// MARK: - PledgePaymentMethodsViewControllerDelegate
+
+extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewControllerDelegate {
+  func pledgePaymentMethodsViewController(
+    _: PledgePaymentMethodsViewController,
+    didSelectCreditCard paymentSource: PaymentSourceSelected
+  ) {
+    switch paymentSource {
+    case let .paymentIntentClientSecret(clientSecret):
+      return STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret) { intent, _ in
```
amy-at-kickstarter, line 293: It would follow our existing patterns better if this call to Stripe's API went into `PostCampaignPledgeViewModel`, instead - that way, all the business logic stays in the VM, including the call back out to the error banner.

amy-at-kickstarter: 🤔 Unless we explicitly keep stripe stuff out of VMs so that the VMs are more testable. Could go either way on this one, now that I say it.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -253,11 +272,61 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
...
+        guard let intent = intent else {
+          self.messageBannerViewController?
+            .showBanner(with: .error, message: Strings.Something_went_wrong_please_try_again())
+          return
+        }
+
+        let stripeCardId = intent.stripeId
+        let paymentIntentClientSecret = paymentSource.paymentIntentClientSecret ?? ""
```
amy-at-kickstarter, line 301: This can safely be `paymentSource.paymentIntentClientSecret!` because it should always exist in this case.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -253,11 +272,61 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
...
+
+        self.viewModel.inputs.creditCardSelected(with: (stripeCardId, paymentIntentClientSecret))
+      }
+    default:
+      break
```
amy-at-kickstarter, line 306: This also needs to handle the `.savedCreditCard` case - otherwise, we're not completing the pledge if you re-use an old card.

Library/ViewModels/ConfirmDetailsViewModel.swift
amy-at-kickstarter: Huh, is this decoding a GraphID sent back from GraphQL? You should ask the server-side folks to give you a plain ID in the mutation. Those Base64 IDs are sort of intended to be opaque and shouldn't need to be read like this.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +136,35 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+    let validateCheckout = Signal.combineLatest(checkoutId, stripeCardIdAndPaymentIntent)
+      .takeWhen(self.submitButtonTappedProperty.signal)
+      .switchMap { checkoutId, stripeCardIdAndPaymentIntent in
+        let (stripeCardId, paymentIntent) = stripeCardIdAndPaymentIntent
+
+        return AppEnvironment.current.apiService
+          .validateCheckout(
+            checkoutId: checkoutId,
+            paymentSourceId: stripeCardId,
+            paymentIntentClientSecret: paymentIntent
```
amy-at-kickstarter, line 153: Do we need a payment intent to validate? What if you're using a saved card?

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +136,35 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+          )
+          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+          .materialize()
+      }
+
+    self.validateCheckoutSuccess = stripeCardIdAndPaymentIntent
+      .takeWhen(validateCheckout.values().ignoreValues())
+      .map { stripeCardIdAndPaymentIntent in
```
amy-at-kickstarter, line 161: Nit: Can be slightly more compact like
`.map { stripeCardId, paymentIntent  in paymentIntent }`

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +136,35 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        let (_, paymentIntent) = stripeCardIdAndPaymentIntent
+        return paymentIntent
+      }
+
+    self.showErrorBannerWithMessage = validateCheckout.errors()
+      .map { _ in Strings.Something_went_wrong_please_try_again() }
```
amy-at-kickstarter, line 167: Nit: Will we show any of the errors to the user, or always this one?

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
scottkicks: Sheesh, thanks for the call-out! I got a little lost in the sauce with this one, and save cards slipped my mind. 

For saved cards I think we'll need to grab the stripeCardId on the selected stored card in the VM to pass to this mutation.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
scottkicks: Unfortunately, we don't have any neat server-side error messaging. I was told generic messaging is fine for now 😢

scottkicks:
Looking into your other comments.

> Why do we need to decode the `GraphID` of a Checkout? Can this be resolved at the mutation level, instead of doing it in the client?

I spoke with the backend, and they would rather we did this. It sounded like they could (they do in other cases), but they aren't willing to spend the time updating the different pieces involved right now.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
amy-at-kickstarter: Sounds good, OK.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
amy-at-kickstarter: Yeah, I think that's an argument for updating the enum to be 
```
case .savedCreditCard(String, String)
```

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -140,26 +139,53 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
         return AppEnvironment.current.apiService
           .validateCheckout(
             checkoutId: checkoutId,
             paymentSourceId: stripeCardId,
-            paymentIntentClientSecret: paymentIntent
+            paymentIntentClientSecret: paymentIntent ?? ""
           )
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
           .materialize()
       }
```
scottkicks, line 180: still working on verifying what the expected value is for paymentIntentClientSecret for existing payment methods

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -211,7 +211,7 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
 
     /// Hide when there is a reward and shipping is enabled (accounts for digital rewards), and in a pledge context
     self.pledgeSummaryViewHidden = Signal.zip(baseReward, context).map { baseReward, context in
-      (baseReward.isNoReward == false && baseReward.shipping.enabled) && context == .pledge
+      baseReward.isNoReward == false && context == .pledge
```
scottkicks, line 214: unrelated fix for showing the summary view on details

Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
scottkicks, line 1: unrelated cleanup of details screens. makes sure table view expands to the full width of the view

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -199,6 +205,28 @@ final class PostCampaignCheckoutViewController: UIViewController {
...
+
+        // TODO: Confirm paymentIntent using Stripe.confirmPayment()
+      }
+
+    self.viewModel.outputs.validateCheckoutExistingCardSuccess
+      .observeForControllerAction()
+      .observeValues { [weak self] _ in
+        guard let self else { return }
+
+        // TODO: Confirm paymentIntent using Stripe.confirmPayment()
```
amy-at-kickstarter, line 221: Is this ``// TODO:`` correct for the existing card case, too?

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -253,11 +281,70 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
...
+    switch paymentSource {
+    case let .paymentIntentClientSecret(clientSecret):
+      return STPAPIClient.shared.retrievePaymentIntent(withClientSecret: clientSecret) { intent, _ in
+        guard let intent = intent else {
+          self.messageBannerViewController?
+            .showBanner(with: .error, message: Strings.Something_went_wrong_please_try_again())
+          return
+        }
+
+        let paymentMethodId = intent.paymentMethodId!
```
amy-at-kickstarter, line 309: Nit: Could also throw this `let` into the above `guard` statement, just in case.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -253,11 +281,70 @@ extension PostCampaignCheckoutViewController: PledgeViewCTAContainerViewDelegate
...
+        self.viewModel.inputs
+          .creditCardSelected(
+            source: paymentSource,
+            paymentMethodId: paymentMethodId,
+            isNewPaymentMethod: true
+          )
+      }
+    case let .savedCreditCard(savedCardId):
+      self.viewModel.inputs
+        .creditCardSelected(source: paymentSource, paymentMethodId: savedCardId, isNewPaymentMethod: false)
```
amy-at-kickstarter, line 320: 👍

KsApi/models/graphql/UserCreditCards.swift
```diff
@@ -9,6 +9,7 @@ public struct UserCreditCards: Decodable {
     public var id: String
     public var lastFour: String
     public var type: CreditCardType?
+    public var stripeCardId: String?
```
amy-at-kickstarter, line 12: Ah, yeah, I only added it to the GraphQL object. Good find.

Library/ViewModels/ConfirmDetailsViewModel.swift
amy-at-kickstarter: 👍

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -353,7 +353,16 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
+      .map { values in
+        var checkoutId = values.checkout.id
+
+        if let decoded = decodeBase64(checkoutId), let range = decoded.range(of: "Checkout-") {
+          let id = decoded[range.upperBound...]
+          checkoutId = String(id)
+        }
+
+        return checkoutId
+      }
```
amy-at-kickstarter, line 365: Any resolution on why this needs to be decoded from GraphQL? At a minimum, this should really have a ``//TODO: `` explaining the hack.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +137,99 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        AppEnvironment.current.apiService
+          .fetchGraphUser(withStoredCards: true)
+          .ksr_debounce(.seconds(1), on: AppEnvironment.current.scheduler)
+          .map { envelope in (envelope, false) }
+          .prefix(value: (nil, true))
+          .materialize()
+      }
+
+    let storedCardsValues = storedCardsEvent.values()
+      .filter(second >>> isFalse)
```
amy-at-kickstarter, line 157: What does this do?

amy-at-kickstarter: Nit: Would prefer this to be a non-Prelude operator for clarity.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +137,99 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+      .map(first)
+      .skipNil()
+      .map { $0.me.storedCards.storedCards }
+
+    let selectedExistingCard = self.creditCardSelectedProperty.signal.skipNil()
+      .filter { $0.isNewPaymentMethod == false }
+
+    let newPaymentIntentEvent = initialData
```
amy-at-kickstarter, line 165: This only needs to happen when you're _not_ adding a new card via the payment method selector, since in that case, we've already made a PI for you. Should this be attached to a different event than `initialData`?

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -130,6 +137,99 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+    self.validateCheckoutExistingCardSuccess = paymentIntentClientSecret
+      .takeWhen(validateCheckoutExistingCard.values())
+      .map { $0 }
+
+    // MARK: - Validate New Cards
+
+    let selectedNewCreditCard = self.creditCardSelectedProperty.signal.skipNil()
+      .filter { $0.isNewPaymentMethod }
+
+    // Runs validation for new cards that were created with payment intents.
```
amy-at-kickstarter, line 210: Loving these comments

amy-at-kickstarter (approved):
LGTM with some questions and nits

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
scottkicks: Yes we'll have to confirm the payment either way using the intent

Library/ViewModels/ConfirmDetailsViewModel.swift
scottkicks: Ah yea i left an unfortunate [comment](https://github.com/kickstarter/ios-oss/pull/1999#issuecomment-2020596093) earlier.

scottkicks: its silly that they would give us a string, that we only use to send to one of their mutations, and not format it correctly for us

PR #2001, opened 2024-03-26
[MBL-1282] Include setup intent context

ifosli:
# 📲 What

Include the setup intent context in setup intents.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1282)

# ✅ Acceptance criteria

- [x] Adding new payment methods still works in pledge context and in settings.


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -309,8 +310,13 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
 
     switch paymentSheetType {
     case .setupIntent:
+      let setupIntentContext = pledgeContext == .latePledge
+        ? GraphAPI.StripeIntentContextTypes.postCampaignCheckout
+        : GraphAPI.StripeIntentContextTypes.crowdfundingCheckout
       clientSecretSignal = AppEnvironment.current.apiService
-        .createStripeSetupIntent(input: CreateSetupIntentInput(projectId: project.graphID))
+        .createStripeSetupIntent(
+          input: CreateSetupIntentInput(projectId: project.graphID, context: setupIntentContext)
```
amy-at-kickstarter, line 318: Nit: I believe `paymentSheetType` should never be `.setupIntent` in the `.latePledge` context, but this is harmless.

amy-at-kickstarter (approved):
LGTM, one nit

Library/ViewModels/PledgePaymentMethodsViewModel.swift
ifosli: I'll just leave it for now, and we can clean it up later if we need to! Specifically, this is for the "Add card" option on the payment sheet; it's not about scheduling a specific payment or anything like that. I don't know enough about the other calls to know if we use a different method to add new cards for late pledges, so I'd rather leave this for now, and we can clean it up later once we have all the backend calls in place, if it makes sense to do so.

amy-at-kickstarter: I'm totally fine with that. The way it's set up is confusing anyways!

PR #2002, opened 2024-03-26
MBL-1285: Add ApplePay into post-campaign pledge controller

amy-at-kickstarter:
# 📲 What

Add ApplePay into post-campaign pledge controller

# 🤔 Why

This is another piece we needed before we can set up the final checkout call - we'll need ApplePay parameters (i.e. the token) to complete the payment.

# 👀 See

Example with a plain pledge:
Screenshot 2024-03-26 at 4 55 46 PM
Screenshot 2024-03-26 at 4 55 48 PM

Example with shipping and a reward:
Screenshot 2024-03-26 at 4 55 19 PM

### DIFF
--- a/Library/PKPaymentRequest+Helpers.swift
+++ b/Library/PKPaymentRequest+Helpers.swift
@@ -3,14 +3,7 @@ import KsApi
 import PassKit
 
 extension PKPaymentRequest {
-  public static func paymentRequest(
-    for project: Project,
-    reward: Reward,
-    allRewardsTotal: Double,
-    additionalPledgeAmount: Double,
-    allRewardsShippingTotal: Double,
-    merchantIdentifier: String
-  ) -> PKPaymentRequest {
+  private static func paymentRequest(for project: Project, merchantIdentifier: String) -> PKPaymentRequest {
     let request = PKPaymentRequest()
     request.merchantIdentifier = merchantIdentifier
     request.supportedNetworks = AppEnvironment.current.applePayCapabilities.supportedNetworks(for: project)
@@ -19,15 +12,34 @@ extension PKPaymentRequest {
     request.currencyCode = projectCountry(forCurrency: project.stats.currency)?.currencyCode ?? project
       .country.currencyCode
     request.shippingType = .shipping
+    return request
+  }
+
+  public static func paymentRequest(
+    for project: Project,
+    reward: Reward,
+    allRewardsTotal: Double,
+    additionalPledgeAmount: Double,
+    allRewardsShippingTotal: Double,
+    merchantIdentifier: String
+  ) -> PKPaymentRequest {
+    let request = self.paymentRequest(for: project, merchantIdentifier: merchantIdentifier)
 
     request.paymentSummaryItems = self.paymentSummaryItems(
       reward: reward,
       allRewardsTotal: allRewardsTotal,
       additionalPledgeAmount: additionalPledgeAmount,
-      allRewardsShippingTotal: allRewardsShippingTotal,
-      postCampaignPledgingActive: featurePostCampaignPledgeEnabled() && project.isInPostCampaignPledgingPhase
+      allRewardsShippingTotal: allRewardsShippingTotal
     )
 
+    return request
+  }
+
+  public static func paymentRequest(
+    for data: PostCampaignPaymentAuthorizationData
+  ) -> PKPaymentRequest {
+    let request = self.paymentRequest(for: data.project, merchantIdentifier: data.merchantIdentifier)
+    request.paymentSummaryItems = self.paymentSummaryItems(for: data)
     return request
   }
 
@@ -35,8 +47,7 @@ extension PKPaymentRequest {
     reward: Reward,
     allRewardsTotal: Double,
     additionalPledgeAmount: Double,
-    allRewardsShippingTotal: Double,
-    postCampaignPledgingActive: Bool
+    allRewardsShippingTotal: Double
   ) -> [PKPaymentSummaryItem] {
     var paymentSummaryItems: [PKPaymentSummaryItem] = []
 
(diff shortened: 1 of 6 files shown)


Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -231,6 +238,53 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
     self.showErrorBannerWithMessage = Signal
       .combineLatest(validateCheckoutExistingCard.errors(), validateCheckoutNewCard.errors())
       .map { _ in Strings.Something_went_wrong_please_try_again() }
+
+    let paymentAuthorizationData: Signal<PaymentAuthorizationData, Never> = self.configureWithDataProperty
+      .signal
+      .skipNil()
+      .map { (data: PostCampaignCheckoutData) -> PaymentAuthorizationData in
+        let reward = data.rewards.first! // What happens if you select no reward?
```
ifosli, line 246: Nit: While I don't think it's possible for there to be no reward (selecting "no reward" means we create a special "no reward" reward), can we fail gracefully if this were to happen instead of crashing the program?

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -231,6 +238,53 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        return PaymentAuthorizationData(
+          project: data.project,
+          reward: reward,
+          // TODO: PKPaymentRequests+Helper switches on reward.isNoReward for displaying the total
+          allRewardsTotal: reward.isNoReward ? 0 : data.total,
+          additionalPledgeAmount: reward.isNoReward ? data.total : 0,
```
ifosli, line 252: I think this is an example of how keeping two separate flows is adding tech debt. In the old checkout flow, the "no rewards" case reward never gets updated, so we need to check the "bonus" amount to find out what amount is currently associated with it. But in the new flow I updated the code that configures the data for the checkout screen to update the "no rewards" reward with its final amount (and set the bonus to 0 so it doesn't get counted twice) in order to make it display nicely in the table without needing the table to know about the "no rewards" case. So this might be something that we just leave kind of ugly until we manage to unify our checkout flows.

ifosli (approved):

Library/ViewModels/PostCampaignCheckoutViewModel.swift
scottkicks: Yeah technically there is a special reward with its own rewardID that we should be aware of.

scottkicks (approved):

Library/ViewModels/PostCampaignCheckoutViewModel.swift
amy-at-kickstarter: Agreed, with regards to the tech debt. And thanks for the explanation - I was kind of confused how I ended up in this position, so it's interesting to hear what the root cause is.

It also doesn't help that the code to create the ApplePay configuration takes these (somewhat opaque) parameters; it's unclear when you examine the method signature what it's actually going to do with the various values.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -231,6 +248,62 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        return PostCampaignPaymentAuthorizationData(
+          project: data.project,
+          hasNoReward: firstReward.isNoReward,
+          subtotal: firstReward.isNoReward ? firstReward.minimum : calculateAllRewardsTotal(
+            addOnRewards: data.rewards,
+            selectedQuantities: data.selectedQuantities
+          ),
+          bonus: data.bonusAmount ?? 0,
+          shipping: data.shipping?.total ?? 0,
+          total: data.total,
```
amy-at-kickstarter, line 271: @ifosli This came out of your comment. Here's what I did:
1) Created a new data struct, just for ApplePay for late pledges, which is much more clear about what the different named fields do (subtotal, bonus, shipping, total).
2) Check whether the reward `isNoReward`. Only calculate the subtotal if there is a "real" reward.

Library/ViewModels/PostCampaignCheckoutViewModelTests.swift
```diff
@@ -0,0 +1,163 @@
...
+    Never
+  >()
+
+  override func setUp() {
+    self.vm.goToApplePayPaymentAuthorization.observe(self.goToApplePayPaymentAuthorization.observer)
+  }
+
+  func testApplePayAuthorization_noReward_isCorrect() {
+    let project = Project.cosmicSurgery
+    let reward = Reward.noReward |> Reward.lens.minimum .~ 5
```
amy-at-kickstarter, line 20: I cried a single tear when I imported Prelude 🥲

Library/PKPaymentRequest+Helpers.swift
```diff
@@ -3,14 +3,7 @@ import KsApi
...
 extension PKPaymentRequest {
-  public static func paymentRequest(
-    for project: Project,
-    reward: Reward,
-    allRewardsTotal: Double,
-    additionalPledgeAmount: Double,
-    allRewardsShippingTotal: Double,
-    merchantIdentifier: String
-  ) -> PKPaymentRequest {
+  private static func paymentRequest(for project: Project, merchantIdentifier: String) -> PKPaymentRequest {
```
amy-at-kickstarter, line 6: Splitting up the bit that is shared between regular/late pledges.

Library/PKPaymentRequest+Helpers.swift
```diff
@@ -84,13 +95,56 @@ extension PKPaymentRequest {
...
+
+    paymentSummaryItems.append(
+      PKPaymentSummaryItem(
+        label: Strings.Kickstarter_payment_summary(),
+        amount: NSDecimalNumber(value: data.total),
+        type: .final
+      )
+    )
+
+    return paymentSummaryItems
```
amy-at-kickstarter, line 148: This is all much simpler. The actual logic to calculate the subtotal happens in the view model, instead of here.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -19,6 +19,16 @@ public struct PostCampaignCheckoutData: Equatable {
   public let checkoutId: String
 }
 
+public struct PostCampaignPaymentAuthorizationData: Equatable {
```
amy-at-kickstarter, line 22: New struct just for this use case.

ifosli (approved):

PR #2003, opened 2024-03-26
[MBL-1271]Stripe's confirmPayment 

scottkicks:
# 📲 What

Calls Stripe.confirmPayment after validating checkout

# 🤔 Why

We need to confirm payment using the new payment intent with Stripe so that we can charge these cards immediately

# 🛠 How

When validation is successful for both a new card and an existing card, call this method!

**There's a key difference for existing cards:**
- We need to set the payment param's `paymentMethodId` property to be the `stripeCardId` that's on the stored card.
- When a pre-existing card (one that was created with a setup intent) is used in late pledges, we use the backend to create a payment intent and use the returned intent to validate and then confirm payment with Stripe. The issue there is that the new payment intent isn't associated with any payment method in Stripe yet so Stripe.confirmPayment will throw an error. Setting the `paymentMethodId` will associate it!

# 🪲
- There's just one issue I noticed last minute in this PR where the outputs for validating new cards and existing cards occasionally trigger at the same time. Have a feeling this is a simple cross of signals, but if someone could take a look when they review that would be really helpful!

# ✅ Acceptance criteria

- [x] When stripe confirms successfuly the charge show up in Stripe

Screenshot 2024-03-26 at 5 45 07 PM


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift:269:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift:274:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/PostCampaignCheckoutViewModel.swift:316:1 | warning: (blankLinesAroundMark) Insert blank line before and after MARK: comments. |


</p>

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -259,11 +251,35 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
     )
 
     let navigationController = UINavigationController(rootViewController: loginSignupViewController)
-    let navigationBarHeight = navigationController.navigationBar.bounds.height
 
     self.present(navigationController, animated: true)
   }
 
+  private func confirmPayment(with validation: PaymentSourceValidation) {
+    guard validation.requiresConfirmation else {
```
amy-at-kickstarter, line 259: This could probably be done in a highly clever was in the view model, but I felt that this was going to be the easiest to follow. This way, new and existing cards go through the same input/output flow, all the time.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -176,10 +184,11 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
       .map { $0.me.storedCards.storedCards }
 
-    let selectedExistingCard = self.creditCardSelectedProperty.signal.skipNil()
-      .filter { $0.isNewPaymentMethod == false }
+    let selectedExistingCard = selectedCard.filter { (_, _, isNewPaymentMethod: Bool) in
+      !isNewPaymentMethod
+    }
 
-    let newPaymentIntentEvent = initialData
+    let newPaymentIntentForExistingCards = initialData
```
amy-at-kickstarter, line 191: Renaming for clarity.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -194,13 +203,16 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
       }
 
-    let paymentIntentClientSecret = newPaymentIntentEvent.values()
+    let paymentIntentClientSecretForExistingCards = newPaymentIntentForExistingCards.values()
       .map { $0.clientSecret }
 
     // Runs validation for pre-existing cards that were created with setup intents originally but require payment intents for late pledges.
     let validateCheckoutExistingCard = Signal
-      .combineLatest(checkoutId, selectedExistingCard, paymentIntentClientSecret, storedCardsValues)
+      .combineLatest(checkoutId, selectedCard, paymentIntentClientSecretForExistingCards, storedCardsValues)
```
amy-at-kickstarter, line 211: @ifosli Did I potentially just undo your bug fix from earlier?

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
ifosli: Yeah, I like this

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: No, you're good - the important part is that we're looking for the most recently selected card in this line and that the filter to check if it's a new or existing card happens after (in line 213-215).

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -241,9 +264,16 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+        paymentIntentClientSecret: paymentSource.paymentIntentClientSecret!,
+        selectedCardStripeCardId: nil,
+        requiresConfirmation: false // Newly added cards are confirmed in PledgePaymentMethodsViewController
+      ) }
+
+    self.validateCheckoutSuccess = Signal
+      .merge(validateCheckoutNewCardSuccess, validateCheckoutExistingCardSuccess)
 
     self.showErrorBannerWithMessage = Signal
       .combineLatest(validateCheckoutExistingCard.errors(), validateCheckoutNewCard.errors())
```
ifosli, line 279: nit: Do you want to swap this to a `merge` while you're here?

ifosli (approved):

Library/ViewModels/PostCampaignCheckoutViewModel.swift
amy-at-kickstarter: Will do

PR #2004, opened 2024-03-27
[MBL-1298] Use correct project country for currency symbol

ifosli:
# 📲 What

Use `project.country` instead of `project.stats.currentCountry` to determine currency labels, since the latter refers to the user's location, not the projects. It also simplifies `PostCampaignCheckoutData` to avoid sending `project` fields as extra params.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1298)

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] Currency symbols are now correct!


Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -294,25 +294,11 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
-      )
-    }
-
-    return (
+    (
       PostCampaignRewardsSummaryViewData(
         rewards: rewards,
         selectedQuantities: selectedQuantities,
-        projectCountry: projectCurrencyCountry,
+        projectCountry: project.country,
```
amy-at-kickstarter, line 301: Huh, interesting. What's the difference between `project.country` and `project.stats.currentCountry`?

amy-at-kickstarter (approved):
LGTM

Library/ViewModels/ConfirmDetailsViewModel.swift
ifosli: `currentCountry` is based on the user's location! It's what we use to determine what to convert currency to, for example. It's the opposite of what we want for the actual project's country

PR #2005, opened 2024-03-28
MBL-1285: Implement CompleteOnSessionCheckoutMutation for new and existing cards

amy-at-kickstarter:
# 📲 What

Implement `CompleteOnSessionCheckoutMutation` for new and existing cards in the late-pledge flow

# 🤔 Why

This is the last mutation call required to actually complete the pledge! 

This PR is _intentionally_ incomplete, to keep the changes a bit more manageable. Follow-up PRs will do the following:

- Implement `CompleteOnSessionCheckoutMutation` for ApplePay
- Route successful transactions to the Thanks page
- Route unsuccessful transactions back to the Select Reward page

# 👀 See
Screenshot 2024-03-28 at 11 05 56 AM

### DIFF
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -238,6 +238,33 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
       .observeValues { [weak self] paymentAuthorizationData in
         self?.goToPaymentAuthorization(paymentAuthorizationData)
       }
+
+    self.viewModel.outputs.checkoutComplete
+      .observeForUI()
+      .observeValues { [weak self] _ in
+        let alert = UIAlertController(
+          title: "Wow!",
+          message: "It worked! Your checkout is done. This should push us to the Thanks page.",
+          preferredStyle: .alert
+        )
+        alert.addAction(UIAlertAction.init(title: "OK", style: .cancel))
+        self?.present(alert, animated: true)
+      }
+
+    self.viewModel.outputs.checkoutError
+      .observeForUI()
+      .observeValues { [weak self] error in
+
+        #if DEBUG
+          let serverError = error.errorMessages.first ?? ""
+          let message = "\(Strings.Something_went_wrong_please_try_again())\n\(serverError)"
+        #else
+          let message = Strings.Something_went_wrong_please_try_again()
+        #endif
+
+        self?.messageBannerViewController?
+          .showBanner(with: .error, message: message)
+      }
   }
 
   // MARK: - Functions
--- a/KsApi/models/graphql/adapters/GraphAPI.ApplePay+ApplePayParams.swift
+++ b/KsApi/models/graphql/adapters/GraphAPI.ApplePay+ApplePayParams.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-extension GraphAPI.ApplePayInput {
+public extension GraphAPI.ApplePayInput {
   static func from(_ input: ApplePayParams?) -> GraphAPI.ApplePayInput? {
     guard let input = input else { return nil }
     return GraphAPI.ApplePayInput(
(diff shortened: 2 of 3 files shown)


Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -239,6 +239,33 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
...
+        self?.present(alert, animated: true)
+      }
+
+    self.viewModel.outputs.checkoutError
+      .observeForUI()
+      .observeValues { [weak self] error in
+
+        #if DEBUG
+          let serverError = error.errorMessages.first ?? ""
+          let message = "\(Strings.Something_went_wrong_please_try_again())\n\(serverError)"
```
ifosli, line 261: I like this pattern!

KsApi/models/graphql/adapters/GraphAPI.ApplePay+ApplePayParams.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-extension GraphAPI.ApplePayInput {
+public extension GraphAPI.ApplePayInput {
```
ifosli, line 3: nit: Should this change be in this pr?

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -335,7 +343,57 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
 
     // MARK: CompleteOnSessionCheckout
 
-    // TODO: Call .completeOnSessionCheckout when self.confirmPaymentSuccessfulProperty.signal is called
+    let completeCheckoutWithExistingCardInput: Signal<GraphAPI.CompleteOnSessionCheckoutInput, Never> = Signal
+      .combineLatest(self.confirmPaymentSuccessfulProperty.signal.skipNil(), checkoutId, selectedExistingCard)
```
ifosli, line 347: I think this is going to have the same issue as before. Let's say I first select an existing card, then add a new card. Now both `selectedExistingCard` and `selectedNewCard` will have values set (the existing one is outdated, but it doesn't know that). Once `confirmPaymentSuccessfulProperty` has an updated value, both this and the new card version of it will fire. I'm pretty sure that, in order to prevent this from happening, we need to add a `filter` _after_ the `combineLatest`.

Alternatively, if we want to be able to just use `selectedNewCard`/`selectedExistingCard` we could map the opposite version to `nil` and filter on that (or return early/map to nil in the map/ect). Ie make sure that these signals won't both have valid cards at the same time.

For this specific call, though, I think it'd be easiest to just have one `completeCheckoutWithCardInput` method and switch on `card.isNewPaymentMethod` for the one field that's different.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -335,7 +343,57 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+      .merge(
+        completeCheckoutWithExistingCardInput,
+        completeCheckoutWithNewCardInput
+        // completeCheckoutWithApplePayInput
+      )
+      .switchMap { input in
+        AppEnvironment.current.apiService.completeOnSessionCheckout(input: input).materialize()
+      }
+
+    self.checkoutComplete = checkoutCompleteSignal.signal.ignoreValues()
```
ifosli, line 395: nit: I'm not sure about this, but I think this might need to be `checkoutCompleteSignal.signal.values().ignoreValues()` in order to prevent it from also firing if there's an error. Will you test it if you haven't already?

ifosli (changes requested):
I have one major concern about possibly duplicating complete checkout; otherwise just a couple nits. We're almost there!

Library/ViewModels/PostCampaignCheckoutViewModel.swift
amy-at-kickstarter: Ha, yup, you're right. Happened to me a few times.

KsApi/models/graphql/adapters/GraphAPI.ApplePay+ApplePayParams.swift
amy-at-kickstarter: Nope! Fixed.

ifosli (approved):

PR #2006, opened 2024-03-28
[old] Show processing view while payment is processing

ifosli:
# 📲 What

Show processing view while payment is processing. This prevents further user interaction until it's dismissed.

Note: It's very important that we make sure that regardless of how checkout goes/where it fails, the processing view will get dismissed. It will lock the app otherwise. The original pledge VC ties the processing view to specific API calls, but we can't do that for post campaign pledges, since we require multiple different API calls to finish the payment flow.

Since the payment flow isn't done yet, the processing view dismisses early (and doesn't show at all for apple pay).

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1307)

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

- [x] Processing overlay shows when it should and hides again when it should

# ⏰ TODO

- [ ] I might leave this PR open for now and wait to update & submit it until the payment flow is done.
- [ ] Add tests

### DIFF
--- a/Library/ViewModels/PostCampaignCheckoutViewModel.swift
+++ b/Library/ViewModels/PostCampaignCheckoutViewModel.swift
@@ -34,6 +34,7 @@ public struct PaymentSourceValidation {
 }
 
 public protocol PostCampaignCheckoutViewModelInputs {
+  func checkoutTerminated()
   func configure(with data: PostCampaignCheckoutData)
   func confirmPaymentSuccessful(clientSecret: String)
   func creditCardSelected(source: PaymentSourceSelected, paymentMethodId: String, isNewPaymentMethod: Bool)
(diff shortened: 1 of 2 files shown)


Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -336,10 +342,24 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
     // TODO: Call .completeOnSessionCheckout when self.confirmPaymentSuccessfulProperty.signal is called
+
+    self.processingViewIsHidden = Signal.merge(
+      initialData.mapConst(true),
+      // TODO: show processing view when Apple pay is initiated
+      self.submitButtonTappedProperty.signal.mapConst(false),
+      validateCheckoutError.mapConst(true),
+      self.checkoutTerminatedProperty.signal.mapConst(true),
+      self.confirmPaymentSuccessfulProperty.signal.mapConst(true) // TODO: update to actual end of flow
+    )
```
amy-at-kickstarter, line 353: Ha, cool pattern.

amy-at-kickstarter (approved):
LGTM.

scottkicks:
@ifosli You'll see some merge conflicts with the https://github.com/kickstarter/ios-oss/pull/2008 changes I've merged in. lmk if I can help resolve.

PR #2007, opened 2024-03-28
[MBL-1305] Don't dismiss view controller when payment fails

ifosli:
# 📲 What

Instead of dismissing the checkout vc when there's an issue and the "Something went wrong, try again" banner shows, stay in the checkout vc in order to let the user actually try again.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1305)

| Before 🐛 | After 🦋 |
| --- | --- |
| b | Simulator Screen Recording - iPhone 15 Pro Max - 2024-03-28 at 14 21 40 |

# ✅ Acceptance criteria

- [x] Error banner shows but doesn't dismiss the view controller


amy-at-kickstarter (approved):
Wow, that's simple! LGTM.

PR #2008, opened 2024-04-01
Navigate To Thank You Page On Successful Late Pledge

scottkicks:
# 📲 What

Navigates To Thank You Page On Successful Late Pledge

# 🤔 Why

So we can formally thank users for their pledge, show them verification, and suggest other projects to them

# 🛠 How

Using data we already have in the PostCampaignCheckoutViewModel, gather the data needed for the Thanks Page.

# 👀 See

I can't remove backings rn and only have 2 accounts that I used to test already so I can't screen record this.

# ✅ Acceptance criteria

- [x] On successful backing, users are navigated to the Thanks Page which displays [the appropriate info ](https://www.figma.com/file/sFfDKxlJ2tiuq1xgIsaoiI/Late-Campaign-Backings?type=design&node-id=54-5016&mode=design&t=oQ3DObf65PfyXZKz-0)

### DIFF
--- a/Library/ViewModels/PostCampaignCheckoutViewModel.swift
+++ b/Library/ViewModels/PostCampaignCheckoutViewModel.swift
@@ -65,7 +65,7 @@ public protocol PostCampaignCheckoutViewModelOutputs {
   var showWebHelp: Signal<HelpType, Never> { get }
   var validateCheckoutSuccess: Signal<PaymentSourceValidation, Never> { get }
   var goToApplePayPaymentAuthorization: Signal<PostCampaignPaymentAuthorizationData, Never> { get }
-  var checkoutComplete: Signal<(), Never> { get }
+  var checkoutComplete: Signal<ThanksPageData, Never> { get }
   var checkoutError: Signal<ErrorEnvelope, Never> { get }
 }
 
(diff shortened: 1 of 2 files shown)


Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -369,7 +370,17 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
         AppEnvironment.current.apiService.completeOnSessionCheckout(input: input).materialize()
       }
 
-    self.checkoutComplete = checkoutCompleteSignal.signal.values().ignoreValues()
+    let thanksPageData = Signal.combineLatest(initialData, baseReward)
+      .map { initialData, baseReward -> ThanksPageData? in
+        guard let reward = baseReward else { return nil }
+
+        return (initialData.project, reward, nil, initialData.total)
+      }
```
scottkicks, line 378: It didn't look like we needed to pass any CheckoutData in this case so I passed nil instead.

amy-at-kickstarter: Huh. Do regular pledges require it, or can `CheckoutData` be cleaned up from the `ThanksPageData` type?

amy-at-kickstarter (approved):
LGTM, one suggestion if it's possible.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
scottkicks: Yea I have a note to look at that. I couldn't see find what it was needed for honestly.

PR #2011, opened 2024-04-01
[MBL-1305] Revert banner changes

ifosli:
# 📲 What

Revert banner changes as we want to start the checkout flow over from the beginning if checkout goes wrong, and the confirm checkout VC is designed to create the checkout id.

Hopefully we'll restructure this for a better user experience in the future, but for now, this was the flow that eng across platforms decided on.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1305)


scottkicks (approved):
Thanks. Sorry, again, for the confusion around this

PR #2012, opened 2024-04-01
MBL-1316: Use isAvailable to filter visible add-ons

amy-at-kickstarter:
# 📲 What

Use the GraphQL field `isAvailable` to filter visible add-ons

# 🤔 Why

Our existing implementation did not take late pledges into account, so it was filtering out add-ons that were actually still available for late pledges. I spoke with @zan-rosenthal and apparently the gold-standard solution is to use the `available` field from GraphQL.

Most of this PR is updating plumbing to get `available` into `Reward`.

### DIFF
--- a/KsApi/models/lenses/RewardLenses.swift
+++ b/KsApi/models/lenses/RewardLenses.swift
@@ -24,7 +24,8 @@ extension Reward {
         shippingRulesExpanded: $1.shippingRulesExpanded,
         startsAt: $1.startsAt,
         title: $1.title,
-        localPickup: $1.localPickup
+        localPickup: $1.localPickup,
+        isAvailable: nil
       ) }
     )
 
@@ -49,7 +50,8 @@ extension Reward {
         shippingRulesExpanded: $1.shippingRulesExpanded,
         startsAt: $1.startsAt,
         title: $1.title,
-        localPickup: $0
+        localPickup: $0,
+        isAvailable: nil
       ) }
     )
 
@@ -74,7 +76,8 @@ extension Reward {
         shippingRulesExpanded: $1.shippingRulesExpanded,
         startsAt: $1.startsAt,
         title: $1.title,
-        localPickup: $1.localPickup
+        localPickup: $1.localPickup,
+        isAvailable: nil
       ) }
     )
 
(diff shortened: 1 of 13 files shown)


KsApi/models/lenses/RewardLenses.swift
```diff
@@ -24,7 +24,8 @@ extension Reward {
         shippingRulesExpanded: $1.shippingRulesExpanded,
```
amy-at-kickstarter, line 24: I'm still formulating some thoughts about how we can replace the `Foo.template.lens |> ...` pattern, so I used this as an opportunity to actually get into the guts of one to learn more about it.

Library/ViewModels/RewardAddOnSelectionViewModel.swift
```diff
@@ -468,17 +467,7 @@ private func addOnIsAvailable(_ addOn: Reward, in project: Project) -> Bool {
...
-  // Assuming the user has not backed the addOn, we only display if it's within range of the start and end date
-  let hasNoTimeLimitOrIsWithinRange = isStartDateBeforeToday(for: addOn) && isEndDateAfterToday(for: addOn)
-
-  return [
-    project.state == .live,
-    hasNoTimeLimitOrIsWithinRange,
-    isUnlimitedOrAvailable
-  ]
-  .allSatisfy(isTrue)
+  return addOn.isAvailable ?? false
```
amy-at-kickstarter, line 470: Replace all the start time logic with `isAvailable`.

Library/ViewModels/RewardAddOnSelectionViewModelTests.swift
```diff
@@ -467,6 +484,7 @@ final class RewardAddOnSelectionViewModelTests: TestCase {
       |> Reward.lens.remaining .~ nil
       |> Reward.lens.startsAt .~ (MockDate().timeIntervalSince1970 + 30)
       |> Reward.lens.endsAt .~ (MockDate().timeIntervalSince1970 + 60)
+      |> Reward.lens.isAvailable .~ false
```
amy-at-kickstarter, line 487: This test actually includes some unavailable rewards, so updating it to reflect that.

scottkicks (approved):
Nice! Sounds like this change is actually a long time coming anyway. 

We should note this change in our regression testing since impacts all rewards

PR #2013, opened 2024-04-01
[MBL-1307] Show processing view while payment is processing

ifosli:
# 📲 What

Show processing view while payment is processing. This prevents further user interaction until it's dismissed.

Note: It's very important that we make sure that regardless of how checkout goes/where it fails, the processing view will get dismissed. It will lock the app otherwise. The original pledge VC ties the processing view to specific API calls, but we can't do that for post campaign pledges, since we require multiple different API calls to finish the payment flow. Please help me double check that there aren't edge cases that would leave the processing screen up indefinitely!

Note: The processing screen does hide immediately once apple pay is dismissed - I have no idea why the gif is showing it weird.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1307)

| Pledge flow | Apple pay flow |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro Max - 2024-04-01 at 16 01 41 | Simulator Screen Recording - iPhone 15 Pro Max - 2024-04-01 at 16 04 20 |

# ✅ Acceptance criteria

- [x] Processing overlay shows when it should and hides again when it should

# ⏰ TODO

- [x] Add tests

### DIFF
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -447,7 +447,6 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
     completion: @escaping StripeApplePay.STPIntentClientSecretCompletionBlock
   ) {
     guard let paymentIntentClientSecret = self.applePayPaymentIntent else {
-      self.viewModel.inputs.checkoutTerminated()
       completion(
         nil,
         PostCampaignCheckoutApplePayError.missingPaymentIntent("Missing PaymentIntent")
@@ -457,7 +456,6 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
 
     guard let paymentDisplayName = payment.token.paymentMethod.displayName,
       let paymentNetworkName = payment.token.paymentMethod.network?.rawValue else {
-      self.viewModel.inputs.checkoutTerminated()
       completion(
         nil,
         PostCampaignCheckoutApplePayError
@@ -475,7 +473,6 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
       }
 
       guard let tokenId = token?.tokenId else {
-        self.viewModel.inputs.checkoutTerminated()
         completion(
           nil,
           PostCampaignCheckoutApplePayError.missingToken("Unable to retrieve token from Stripe")
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -11,7 +11,9 @@ private enum PostCampaignCheckoutLayout {
   }
 }
 
-final class PostCampaignCheckoutViewController: UIViewController, MessageBannerViewControllerPresenting {
+final class PostCampaignCheckoutViewController: UIViewController,
+  MessageBannerViewControllerPresenting,
+  ProcessingViewPresenting {
   // MARK: - Properties
 
   internal var messageBannerViewController: MessageBannerViewController?
@@ -41,6 +43,8 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
   private lazy var pledgeRewardsSummaryViewController = {
     PostCampaignPledgeRewardsSummaryViewController.instantiate()
   }()
+
+  internal var processingView: ProcessingView? = ProcessingView(frame: .zero)
 
   private lazy var rootScrollView: UIScrollView = {
     UIScrollView(frame: .zero)
@@ -91,6 +95,7 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
   }
 
   deinit {
+    self.hideProcessingView()
     self.sessionStartedObserver.doIfSome(NotificationCenter.default.removeObserver)
   }
 
@@ -186,6 +191,16 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
       .observeForUI()
       .observeValues { [weak self] value in
         self?.pledgeCTAContainerView.configureWith(value: value)
+      }
+
+    self.viewModel.outputs.processingViewIsHidden
+      .observeForUI()
+      .observeValues { [weak self] isHidden in
+        if isHidden {
+          self?.hideProcessingView()
+        } else {
+          self?.showProcessingView()
+        }
       }
 
     self.viewModel.outputs.configurePaymentMethodsViewControllerWithValue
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -296,6 +311,7 @@ final class PostCampaignCheckoutViewController: UIViewController, MessageBannerV
         guard error == nil, status == .succeeded else {
           self.messageBannerViewController?
             .showBanner(with: .error, message: Strings.Something_went_wrong_please_try_again())
+          self.viewModel.inputs.checkoutTerminated()
```
ifosli, line 314: I don't like how many different places we're calling `checkoutTerminated` in this - it feels messy. But I can't think of a better way to implement this. Let me know if y'all have any ideas.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/PostCampaignCheckoutViewModelTests.swift:15:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/PostCampaignCheckoutViewModelTests.swift:269:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/PostCampaignCheckoutViewModelTests.swift:288:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -391,10 +403,25 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
+
+    self.processingViewIsHidden = Signal.merge(
+      initialData.mapConst(true),
+      self.submitButtonTappedProperty.signal.mapConst(false),
+      self.applePayButtonTappedSignal.mapConst(false),
+      newPaymentIntentForApplePayError.mapConst(true), // Apple pay payment intent error.
+      validateCheckoutError.mapConst(true), // Card validation error terminates flow.
+      self.checkoutTerminatedProperty.signal.mapConst(true), // Error/cancellation in VC.
+      checkoutCompleteSignal.signal.mapConst(true) // Checkout completed.
+    )
```
amy-at-kickstarter, line 415: I see what you mean about worrying about edge cases in here. I don't see any obvious errors - but for readability, you could sort the `merge` into clearer groups of `true` and `false`.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -391,10 +403,25 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
       .map { $0 }
 
     self.checkoutError = checkoutCompleteSignal.signal.errors()
+
+    self.processingViewIsHidden = Signal.merge(
+      initialData.mapConst(true),
```
amy-at-kickstarter, line 408: Because this property starts as `true` I don't think this one is necessary.

Library/ViewModels/PostCampaignCheckoutViewModelTests.swift
```diff
@@ -12,11 +12,14 @@ final class PostCampaignCheckoutViewModelTests: XCTestCase {
     Never
```
amy-at-kickstarter, line 12: Can you add more tests for the processing view in here?

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
amy-at-kickstarter: You could write a wrapper method like 

```
checkoutTerminated(withErrorMessage message: String?) {
    self.messageBannerViewController?
            .showBanner(with: .error, message: message ?? Strings.Something_went_wrong_please_try_again())
    self.viewModel.inputs.checkoutTerminated()
}
```

which would consolidate two of these.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -430,6 +447,7 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
     completion: @escaping StripeApplePay.STPIntentClientSecretCompletionBlock
   ) {
     guard let paymentIntentClientSecret = self.applePayPaymentIntent else {
+      self.viewModel.inputs.checkoutTerminated()
```
amy-at-kickstarter, line 450: These calls aren't needed - AFAIK everywhere that `completion(nil, error)` is called will end up below in `case: .error` of `func applePayContext(didCompleteWith:error:)`. Tested this with a breakpoint and by breaking one of the calls.

amy-at-kickstarter (approved):
Some suggestions for clarity but overall LGTM

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
ifosli: Oh, thanks! That makes this nicer :) I wasn't sure how to generate one of these errors so I really appreciate that!

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
ifosli: That feels more confusing to me actually, since then the `checkoutTerminated` would sometimes be nested and sometimes not. I think I'm going to leave this as-is if you're okay with it. I also think it's clearer now that I got rid of the extraneous calls, so thanks for that call-out!

The other thing I tried was having the processing view tied directly to api calls the way the PledgeViewModel does, but it did disappear and reappear between linked calls, so I say that very much did not work.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
ifosli: Done!

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
amy-at-kickstarter: Ha, me neight, but I just set something to `nil` and checked what happened 😆

Library/ViewModels/PostCampaignCheckoutViewModelTests.swift
ifosli: Yeah, of course! I wanted to make sure we were on the same page about going with this approach before I wrote proper tests. But I think we're good now!

scottkicks (approved):

PR #2015, opened 2024-04-02
Add init(jsonString:) to GraphQLSelectionSet

amy-at-kickstarter:
# 📲 What

Add a convenience initializer `init(jsonString:)` to `GraphQLSelectionSet`

# 🤔 Why

This makes it easier to create new GraphQL Data objects in test environments.


ifosli (approved):
Nice!

PR #2018, opened 2024-04-03
Remove swiftformat:disable directives that SwiftFormat was ignoring

amy-at-kickstarter:
# 📲 What

Remove `swiftformat:disable` directives that weren't working.

# 🤔 Why

When all of us upgraded our Macbooks to Sonoma 14.4, `swiftformat` suddenly ceased to respect all of our `disable` and `enable` statements. Frustratingly, one file `ServiceTypeTests.swift` would get trapped in an infinite loop, in which swiftformat _always_ insisted on editing it.

We've yet to find an actual fix for this issue, but to be blunt, the `ignore`s were mostly just there to make certain files that shouldn't be wrapped a bit prettier. In this PR I'm removing all of the `ignore`s that I can, and fixing the one in `ServiceTypeTest` that is actually broken. 

I'll file a follow-up ticket to investigate further, but it's more important that the team stops being blocked by these unnecessary/unwanted formatting changes than it is for us to have these files be beautiful.

### DIFF
--- a/KsApi/ServiceTypeTests.swift
+++ b/KsApi/ServiceTypeTests.swift
@@ -319,14 +319,10 @@ final class ServiceTypeTests: XCTestCase {
   }
 }
 
-// swiftformat:disable wrap
 private func userAgent() -> String {
-  return """
-  xctest/\(testToolBuildNumber()) (\(UIDevice.current.model); iOS \(UIDevice.current.systemVersion) Scale/\(UIScreen.main.scale))
-  """
+  // swiftformat:disable:next wrap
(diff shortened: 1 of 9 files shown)


KsApi/ServiceTypeTests.swift
```diff
@@ -319,15 +319,11 @@ final class ServiceTypeTests: XCTestCase {
   }
 }
 
-// swiftformat:disable wrap
 private func userAgent() -> String {
-  return """
-  xctest/\(testToolBuildNumber()) (\(UIDevice.current.model); iOS \(UIDevice.current.systemVersion) Scale/\(UIScreen.main.scale))
-  """
+  // swiftformat:disable:next wrap
```
amy-at-kickstarter, line 323: This one _does_ work, but I was unable to apply it to other files ❓😕❓ Something strange is going on here and I don't have the bandwidth to dive into it.

PR #2021, opened 2024-04-03
[MBL-1124] Add tracking for late pledges

ifosli:
# 📲 What

Track when the checkout page is viewed and pledge (or apple pay) button is clicked in the late pledge flow. We also add [requested late pledge information to the project properties](https://kickstarter.atlassian.net/wiki/spaces/IN/pages/2687762661/Late+Pledge+Event+Instrumentation).

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1124)

# ✅ Acceptance criteria

- [x] Segment events are sent in late pledge checkout flow with `project_state` set to `post_campaign`
- [x] Segment events outside the late pledge checkout flow are unchanged except for having one extra property
- [x] `project_late_pledge_enabled` is set to true if a project has late pledges enabled, regardless of if it's in the late pledging state or not


scottkicks (approved):
LGTM

PR #2024, opened 2024-04-04
DRAFT: Move CreateCheckout to final pledge page

amy-at-kickstarter:
# 📲 What

WIP: Move CreateCheckout to final pledge page.

# 🤔 Why

If you are logged out, we need to re-fetch a new `Checkout` for you. The original `Checkout` we create is not associated with your user.

# 🛠 How

This is an experimental/potential way to fix this problem. Right now, the code works, but tests are failing and will take some additional work to make them pass. Additionally, this adds yet another layer of complexity to `PostCampaignCheckoutViewModel`, which is already fairly complex. 

# 📓  TODO
- [ ] Update tests with `CreateCheckoutResult` mocks so that they pass correctly

### DIFF
--- a/Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
+++ b/Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
@@ -12,12 +12,10 @@ public enum ConfirmDetailsLayout {
 
 protocol ConfirmDetailsViewControllerDelegate: AnyObject {}
 
-final class ConfirmDetailsViewController: UIViewController, MessageBannerViewControllerPresenting {
+final class ConfirmDetailsViewController: UIViewController {
   // MARK: - Properties
 
   public weak var delegate: ConfirmDetailsViewControllerDelegate?
-
-  internal var messageBannerViewController: MessageBannerViewController?
 
   private lazy var titleLabel = UILabel(frame: .zero)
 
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
```diff
@@ -12,13 +12,11 @@ public enum ConfirmDetailsLayout {
 
 protocol ConfirmDetailsViewControllerDelegate: AnyObject {}
 
-final class ConfirmDetailsViewController: UIViewController, MessageBannerViewControllerPresenting {
+final class ConfirmDetailsViewController: UIViewController {
   // MARK: - Properties
 
   public weak var delegate: ConfirmDetailsViewControllerDelegate?
 
-  internal var messageBannerViewController: MessageBannerViewController?
```
amy-at-kickstarter, line 20: This page would no longer need to show any server errors 🤷‍♀️

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -30,6 +30,7 @@ public func featureFacebookLoginInterstitialEnabled() -> Bool {
 }
 
 public func featurePostCampaignPledgeEnabled() -> Bool {
+  return true
```
amy-at-kickstarter, line 33: Don't commit this, ha

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -379,14 +347,11 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
-        context: initialData.context,
-        checkoutId: checkoutId
+        context: initialData.context
       )
     }
 
-    // TODO: [MBL-1217] Update string once translations are done
-    self.showErrorBannerWithMessage = createCheckoutEvents.errors()
-      .map { _ in Strings.Something_went_wrong_please_try_again() }
+    self.showErrorBannerWithMessage = .never
```
amy-at-kickstarter, line 354: This output could be cleaned up

PR #2026, opened 2024-04-04
[MBL-1350] Remove Thanks Page String

scottkicks:
# 📲 What

It was decided that the string doesn't contextually make a lot of sense during a late pledge and that we should remove it entirely for the beta.

[Slack Conversation](https://kickstarter.slack.com/archives/C03J6NRA8/p1712256586541839)

I created a [fast follow ticket](https://kickstarter.atlassian.net/browse/MBL-1351) to address getting a new string defined/translated.

# 🛠 How

Sets the current string to "" and adds a TODO comment so we remember to get this updated.

# 👀 See

# ✅ Acceptance criteria

- [x] The only thing on this page is "Thank you!", share this project button, and the "Other projects you might like" section


scottkicks:
Merging because the team is OOO

PR #2027, opened 2024-04-04
[MBL-1339] Include multiple copies of add-ons if selected

ifosli:
# 📲 What

Send all copies of each reward to the backend. Note that this doesn't fix the bonus support being weird, but that matches a web bug that backend/web people are already working on.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1339)

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot 2024-04-04 at 2 29 58 PM | Screenshot 2024-04-04 at 2 32 31 PM |
| Screenshot 2024-04-04 at 2 30 06 PM | Screenshot 2024-04-04 at 2 32 40 PM |

# ✅ Acceptance criteria

- [x] Reward cards now match actual pledge
- [x] Total reward cost and total shipping cost is now correct in the pledge summary

### DIFF
--- a/Library/ViewModels/ConfirmDetailsViewModel.swift
+++ b/Library/ViewModels/ConfirmDetailsViewModel.swift
@@ -314,14 +314,22 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
     let pledgeDetailsData = Signal.combineLatest(
       project,
       rewards,
+      selectedQuantities,
       pledgeTotal,
       refTag
     )
 
     let createCheckoutEvents = pledgeDetailsData
       .takeWhen(self.continueCTATappedProperty.signal)
-      .map { project, rewards, pledgeTotal, refTag in
-        let rewardsIDs = rewards.first?.isNoReward == true ? [] : rewards.map { $0.graphID }
+      .map { project, rewards, selectedQuantities, pledgeTotal, refTag in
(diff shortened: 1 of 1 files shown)


Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -314,14 +314,22 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
...
-        let rewardsIDs = rewards.first?.isNoReward == true ? [] : rewards.map { $0.graphID }
+      .map { project, rewards, selectedQuantities, pledgeTotal, refTag in
+        let rewardsIDs: [String] = rewards.first?.isNoReward == true
+          ? []
+          : rewards.flatMap { reward -> [String] in
+            guard let count = selectedQuantities[reward.id] else {
+              return []
+            }
+            return [String](repeating: reward.graphID, count: count)
+          }
```
scottkicks, line 332: might be a good idea to add a test for this

scottkicks (approved):

Library/ViewModels/ConfirmDetailsViewModel.swift
ifosli: Discussed offline - yeah, I'd be nice, but our test suite isn't designed to test that network calls are constructed correctly. So I'm going to just leave this for now.

PR #2029, opened 2024-04-08
[MBL-1355] Move Login/Signup Logic To Confirm Pledge Details

scottkicks:
# 📲 What

Addresses [MBL-1347](https://kickstarter.atlassian.net/browse/MBL-1347) issue by moving login/signup logic from `PostCampaignCheckoutViewController` to `ConfirmDetailsViewController`

# 🤔 Why

By gating the checkout screen and checkout id creation behind login/signup, we can keep UX much simpler and more straightforward for users _and_ code perspective.

# 🛠 How

1. Remove login logic from `PostCampaignCheckoutViewController`
2. Add the same login logic to `ConfirmDetailsViewController` but attach to continue button tap.
3. Call `.createCheckout` only if the user is logged in. The same goes for navigating to checkout. Users should never reach that screen if they aren't logged in.
4. Navigating back to Confirm, from Checkout, should create a new checkout ID and should all for successful late pledging. 

# 👀 See

https://github.com/kickstarter/ios-oss/assets/110618242/3c8d1b16-085e-4de1-a5b7-709ab79d7a19

# ✅ Acceptance criteria

Original test [scenarios that were failing](https://kickstarter.slack.com/archives/C05HX9GS0CD/p1712269943445659)

- [x] Login is presented after tapping continue on ConfirmDetails Screen _if logged out_
- [x] Users can not reach the Checkout screen _if logged out_
- [x] _if logged in_ users can confirm pledge details and complete late pledge flow successfully. This is true even if users navigate back and forth around the late pledge flow.

[MBL-1347]: https://kickstarter.atlassian.net/browse/MBL-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
```diff
@@ -301,6 +318,18 @@ final class ConfirmDetailsViewController: UIViewController, MessageBannerViewCon
 
   // MARK: - Functions
 
+  private func goToLoginSignup(with intent: LoginIntent, project: Project, reward: Reward?) {
+    let loginSignupViewController = LoginToutViewController.configuredWith(
+      loginIntent: intent,
+      project: project,
+      reward: reward
+    )
```
scottkicks, line 326: At first glance, it looks like LoginToutViewController doesn't actually use the project or reward being passed in, but I found that they're specifically for analytics.

Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -319,9 +333,11 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
       refTag
     )
 
-    let createCheckoutEvents = pledgeDetailsData
+    let createCheckoutEvents = Signal.combineLatest(isLoggedIn, pledgeDetailsData)
```
ifosli, line 336: Should we call this on userSessionStarted as well? I think that would probably fix the main issue I have with this flow (having to hit "continue" twice)

ifosli (approved):
Code looks good! I still think we should fix the "user needs to tap continue twice" to log in and proceed to the checkout page, but I'll let you decide if that's something you have time to do in this pr or not, since it's release-blocking.

Library/ViewModels/ConfirmDetailsViewModel.swift
scottkicks: What's a good RxSwift way to say "do this when userSessionStarted or if continueCTATappedProperty has been tapped by a logged in user"? I don't think just adding self.userSessionStarted to this combineLatest would work.

Kickstarter-iOS/Features/ConfirmDetails/Controllers/ConfirmDetailsViewController.swift
amy-at-kickstarter: I've noticed this in quite a few places - a a view model requiring way more information than seems initially necessary, because it gets passed through to analytics. This is something we might consider for future refactoring. The way we pass around "contexts" in some places feels like a better model for this.

Library/ViewModels/ConfirmDetailsViewModel.swift
amy-at-kickstarter: Although I wonder if that might double-fire, hm.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -431,13 +407,12 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
     self.configurePledgeViewCTAContainerView = Signal.combineLatest(
-      isLoggedIn,
       pledgeButtonEnabled,
       context
     )
-    .map { isLoggedIn, pledgeButtonEnabled, context in
+    .map { pledgeButtonEnabled, context in
       PledgeViewCTAContainerViewData(
-        isLoggedIn: isLoggedIn,
+        isLoggedIn: true, // Users should always be logged in when they get to the Checkout screen.
```
amy-at-kickstarter, line 415: Nit: It would be good to throw an assertion in this view model to this effect.

PR #2032, opened 2024-04-10
[MBL-1334] Bind styles when PostCampaignPledgeRewardsSummaryTotalViewController loads

ifosli:
# 📲 What

Call `bindStyles` in `viewDidLoad`. We usually swizzle view controllers to call `bindStyles` on `viewWillAppear` and `traitCollectionDidChange`, but for some reason, on iOS 16, neither of these trigger `bindStyles` for this class. I think embedding a view controller's view in a different view controller is preventing `viewWillAppear` from triggering it, but I'm not sure why `traitCollectionDidChange` calls `bindStyles` on iOS 17 but not on iOS 16. Regardless of why, calling `bindStyles` in `viewDidLoad` fixes the problem and seems significantly safer.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1334)

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] Total row in the pledge summary table looks good on iOS 16 as well

### DIFF
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PostCampaignPledgeRewardsSummaryTotalViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PostCampaignPledgeRewardsSummaryTotalViewController.swift
@@ -24,6 +24,7 @@ final class PostCampaignPledgeRewardsSummaryTotalViewController: UIViewControlle
     super.viewDidLoad()
 
     self.configureSubviews()
+    self.bindStyles()
 
     self.viewModel.inputs.viewDidLoad()
   }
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter (approved):
LGTM. I'd double-check if this view controller was being correctly added as a child of its parent container, and see if that fixes it, but this change seems harmless anyways.

ifosli:
> LGTM. I'd double-check if this view controller was being correctly added as a child of its parent container, and see if that fixes it, but this change seems harmless anyways.

Ahh, good call! that was the problem! I've added it as a child of the parent and I left the bindStyles call in as well.

PR #2033, opened 2024-04-11
[MBL-1329 Calculate total cost for each add-on

ifosli:
# 📲 What

When users add multiple copies of an add-on, whether in the standard pledge flow or in the late pledge flow, multiply the line item cost for that add-on by the count.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1329)

| Before 🐛 | After 🦋 |
| --- | --- |
| todo | todo |
| todo | todo |

# ✅ Acceptance criteria

- [ ] Line item cost matches multiplier
- [ ] Items with just one copy are unchanged

### DIFF
--- a/Library/ViewModels/PledgeExpandableRewardsHeaderViewModel.swift
+++ b/Library/ViewModels/PledgeExpandableRewardsHeaderViewModel.swift
@@ -151,11 +151,11 @@ private func items(
     guard let title = reward.title else { return nil }
 
     let quantity = selectedQuantities[reward.id] ?? 0
-
     let itemString = quantity > 1 ? "\(Format.wholeNumber(quantity)) x \(title)" : title
 
+    let amount = quantity > 1 ? reward.minimum * Double(quantity) : reward.minimum
     let amountAttributedText = attributedRewardCurrency(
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter (approved):
LGTM. Have you double-checked ApplePay, as well? IIRC we just have one generic "rewards" line item there, we don't call out add-ons, but it'd be good to double-check.

ifosli:
> LGTM. Have you double-checked ApplePay, as well? IIRC we just have one generic "rewards" line item there, we don't call out add-ons, but it'd be good to double-check.

Good idea! Just did. Yeah, apple pay has one "Reward" item that includes the cost of all rewards (including multiple of each add-on, if relevant). So that one is already good (and unchanged)!

amy-at-kickstarter:
> > LGTM. Have you double-checked ApplePay, as well? IIRC we just have one generic "rewards" line item there, we don't call out add-ons, but it'd be good to double-check.
> 
> Good idea! Just did. Yeah, apple pay has one "Reward" item that includes the cost of all rewards (including multiple of each add-on, if relevant). So that one is already good (and unchanged)!

Perfect. Maybe some day we can make ApplePay nicer, too, but might as well leave it as-it-was for now.

PR #2041, opened 2024-04-18
MBL-1375: Change example data in TwoFactorViewModelTests

amy-at-kickstarter:
# 📲 What

Remove example data that includes a real-looking Kickstarter address and password.

# 🤔 Why

Reported as possible security leak.

### DIFF
--- a/Library/ViewModels/TwoFactorViewModelTests.swift
+++ b/Library/ViewModels/TwoFactorViewModelTests.swift
@@ -37,7 +37,7 @@ final class TwoFactorViewModelTests: TestCase {
 
     self.isFormValid.assertValues([false])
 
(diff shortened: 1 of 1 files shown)


Library/ViewModels/TwoFactorViewModelTests.swift
```diff
@@ -228,7 +228,7 @@ final class TwoFactorViewModelTests: TestCase {
 
     withEnvironment(apiService: MockService(resendCodeResponse: error)) {
       self.vm.inputs.viewWillAppear()
-      self.vm.inputs.facebookToken("2934ohhailisa")
```
scottkicks, line 231: 😂 what a token

scottkicks (approved):

PR #2042, opened 2024-04-18
MBL-1371: Only show 'Coming Soon' if project is actually pre-launch

amy-at-kickstarter:
# 📲 What

Check both `displayPrelaunch` and `prelaunchActivated` flags on `Project` before showing `"Coming Soon"` in the project status.

# 🤔 Why

When I swapped this page to use GraphQL, it started returning values in `displayPrelaunch` and `prelaunchActivated` which were formerly `nil`. This caused the status badges to be incorrect.

On a little digging, the existing logic to check for prelaunch status didn't make sense:

- `prelaunchActivated` refers to whether the prelaunch page is available 
- `displayPrelaunch` is true if the project has not actually launched yet 

I surmised that checking for both, not either, is the correct behavior.

I kept the behavior that checks the launch date - turns out that is used for detecting prelaunch projects fetched via `/v1/discover` i.e. search.

# 👀 See

I checked this by comparing the status badges for my saved projects on prod with the badges in this version. I also created a prelaunch project on staging and verified that it shows "Coming soon" until I actually hit "launch".

Screenshot 2024-04-18 at 5 14 51 PM
Screenshot 2024-04-18 at 5 16 13 PM

### DIFF
--- a/Kickstarter-iOS/Features/Search/Controller/SearchViewControllerTests.swift
+++ b/Kickstarter-iOS/Features/Search/Controller/SearchViewControllerTests.swift
@@ -114,6 +114,7 @@ internal final class SearchViewContollerTests: TestCase {
         |> Project.lens.photo.full .~ ""
         |> Project.lens.photo.med .~ ""
         |> Project.lens.displayPrelaunch .~ true
+        |> Project.lens.prelaunchActivated .~ true
     }
 
     let discoveryResponse = .template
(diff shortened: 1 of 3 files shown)


amy-at-kickstarter:
OK, I did some additional spot checks here - 

I tested search (and found and fixed a bug). Now the cards here should display the same information as cards on web for search, including for prelaunch projects:
Screenshot 2024-04-22 at 10 26 25 AM
Screenshot 2024-04-22 at 10 26 38 AM

I also created my own prelaunch project on staging, just to verify that it looks like I expect:
Screenshot 2024-04-22 at 10 12 18 AM

ifosli (approved):

PR #2043, opened 2024-04-22
MBL-1381: Use Projects query instead of UserSavedProjectsConnection for saved projects

amy-at-kickstarter:
# 📲 What

Use `projects` GraphQL search query to fetch saved projects, instead of using `me { savedProjects } `.

# 🤔 Why

The `savedProjects` attached to the `me` query uses a specific, hardcoded sort methodology, which doesn't work for the apps. We need the saved projects to be sorted by end date (descending).

This is a slightly imperfect solution. The `projects` search query doesn't return buried projects, which would otherwise show up in your saved projects. However, it _does_ show late pledge projects, in the correct order, which was the primary motivation for swapping this endpoint from V1 to GraphQL. Since the old V1 endpoint _also_ didn't show buried projects, this is not a regression.

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -8717,23 +8717,20 @@ public enum GraphAPI {
     public let operationDefinition: String =
       """
       query FetchMySavedProjects($first: Int = null, $after: String = null, $withStoredCards: Boolean = false) {
-        me {
+        projects(first: $first, after: $after, starred: true, sort: END_DATE) {
           __typename
-          savedProjects(first: $first, after: $after) {
+          nodes {
             __typename
-            nodes {
-              __typename
-              ...ProjectFragment
-            }
-            pageInfo {
-              __typename
-              hasNextPage
-              endCursor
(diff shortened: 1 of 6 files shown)


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ifosli (approved):
Looks good to me! Are we able to do something similar for backed projects as well?

amy-at-kickstarter:
> Looks good to me! Are we able to do something similar for backed projects as well?

In this case, the backed projects sort looks totally fine using the `me { backedProjects }` query - so I left it as-is. I think the saved projects query is potentially a special case.

Here's an example comparing my first 5 on prod:
Screenshot 2024-04-23 at 2 34 17 PM
Screenshot 2024-04-23 at 2 34 25 PM

PR #2044, opened 2024-04-23
Fix missing spinner on BackerDashboardProjectsViewController

amy-at-kickstarter:
# 📲 What

Fix a missing loading spinner on `BackerDashboardProjectsViewController`.

# 🤔 Why

I noticed this while fixing other bugs - the spinner was frequently missing, and it made the loading feel much slower than it effectively was.

# 🛠 How

- Change `viewWillAppear` to `viewDidAppear` to fix a warning about using the refresh controller while off screen
- Add extra `endRefreshing()` call to `viewDidDisappear`, per a suggestion from StackOverflow. Seems like the refresh controller is strangely sensitive to view lifecycle methods, and this fixes the problem.


scottkicks (approved):
Noice

PR #2045, opened 2024-04-23
[MBL-1358] Add Privacy Manifest File

scottkicks:
… type and reason

# 📲 What

Creates a new Privacy Manifest file and includes the missing reasons for our use of the called out NSPrivacyAccessedAPICategory types

# 🤔 Why

From Apple:

```
Missing API declaration - Your app’s code in the “Kickstarter” file references one or more APIs that require reasons, including the following API categories: 
- NSPrivacyAccessedAPICategoryUserDefaults
- NSPrivacyAccessedAPICategoryFileTimestamp
```

Screenshot 2024-04-23 at 10 50 41 AM

We need to provide appropriate reasoning for these or else we won't be accepted in the App Store.

# 🛠 How

- Creates a new PrivacyInfo.xcprivacy following [Apple's docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files)
- Adds the NSPrivacyAccessedAPICategoryUserDefaults type and reason following these [apple docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

We aren't using the NSPrivacyAccessedAPICategoryFileTimestamp type internally. Rather, it looks like Kingfisher does. Specifically, they use:  [modificationDate, creationDate, contentModificationDateKey, and creationDateKey](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api) 

- I _think_ the solution here is to update Kingfisher to >= 7.10.0 (the version where they updated their privacy manifest) instead of updating our manifest. We're reaching out to Apple to hopefully get some clarity on that. It might not be a big deal to update the version anyway.

# 👀 See
Screenshot 2024-04-23 at 10 59 04 AM


amy-at-kickstarter (approved):
LGTM. Makes sense, we just use defaults for our own local storage.

PR #2046, opened 2024-04-23
[MBL-1358] Update Kingfisher

scottkicks:
# 📲 What

Updates min version to 7.10.0 up to next major version 7.11.0 (Latest).

# 🤔 Why

[Main reference for what's going on](https://github.com/kickstarter/ios-oss/pull/2045)

But, Apple requires us to give reasons for using certain APIs. Reasons that need to be added to a Privacy Manifest file (added in that PR ☝️ )

One of the APIs that our app is missing a reason for is `NSPrivacyAccessedAPICategoryFileTimestamp`. 
We aren't actually using this API internally, Kingfisher is.

Kingfisher added privacy support for Apple's new requirement in 7.10.0. So, by updating the version we're using we should now be meeting the requirement.
- I'm not certain that we don't still need to include a reason for how the 3rd party is using this API as well, but we've reached out to apple support for clarification. For now we're going to update anyway since it has minimal impact.

# 🛠 How

[set minimum version to 7.10.0 (version that they added privacy support) up to next major version (7.11.0)](https://github.com/kickstarter/ios-oss/commit/d3f06d36fc5bc5ab99d922bd344cb3afd407e6fd)

# 👀 See

Kingfisher's Privacy Manifest which contains the File Timestamp type and reason that Apple is warning us about.

Screenshot 2024-04-23 at 11 59 16 AM

# ✅ Acceptance criteria

- [x] The app runs as normal. Specifically image loading. I ran through som regression test steps as well to cover more cases.
- [x] All tests pass

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -10785,7 +10785,7 @@
 			repositoryURL = "https://github.com/onevcat/Kingfisher";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 7.3.2;
+				minimumVersion = 7.10.0;
 			};
 		};
 		60DA511028C96865002E2DF1 /* XCRemoteSwiftPackageReference "SwiftSoup" */ = {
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter (approved):
LGTM. Will be interesting to hear what Apple says.

PR #2047, opened 2024-04-25
[MBL-1373] ValidateCheckout Before Completing Checkout With ApplePay

scottkicks:
# 📲 What

Calls ValidateCheckout mutation after ApplePay Context completes and before CompleteOnSessionCheckout

# 🤔 Why

We should be validating the current late pledge checkout for ApplePay pledges like we do for Stripe pledges.

# 🛠 How

When the ApplePay sheet completes, `didCreatePaymentMethod` is called and we pass our ApplePayParams object back to the ViewModel to complete the checkout. This method returns a `StripePayments.STPPaymentMethod` with an id (`stripeId`) that we need to validate the checkout. So, we'll use that ID to call ValidateCheckout before trying to complete the checkout.

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] ApplePay Late Pledges still work
- [x] If ValidateCheckout fails, an error message banner is shown
- [x] The processing view is still shown until the backing is complete

### DIFF
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
@@ -416,7 +416,7 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
   func applePayContext(
     _: StripeApplePay.STPApplePayContext,
     didCreatePaymentMethod paymentMethod: StripePayments.STPPaymentMethod,
-    paymentInformation payment: PKPayment,
+    paymentInformation _: PKPayment,
     completion: @escaping StripeApplePay.STPIntentClientSecretCompletionBlock
   ) {
     guard let paymentIntentClientSecret = self.applePayPaymentIntent else {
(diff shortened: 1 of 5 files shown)


KsApi/mutations/inputs/ApplePayParams.swift
```diff
@@ -1,17 +1,20 @@
 import Foundation
 
 public struct ApplePayParams: Encodable, Equatable {
+  public let paymentMethodId: String?
```
scottkicks, line 4: Xcode was yelling at me that this property was 'inaccessible' due to the struct being 'internal', even though it and its init are both marked as public. Explicitly labeling this method as public was the only way to get around it for me.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -497,6 +497,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
     )
     .map { paymentData, token in
       (
+        nil,
```
ifosli, line 500: Do we validate checkout in the normal pledge flow? Because if we do, we probably need to do so for apple pay here, too.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -440,6 +440,7 @@ extension PostCampaignCheckoutViewController: STPApplePayContextDelegate {
     let transactionId = payment.token.transactionIdentifier
 
     let params = ApplePayParams(
+      paymentMethodId: paymentMethod.stripeId,
```
ifosli, line 443: I think it would make more sense to pass the `paymentMethodId` as its own thing to the view model, especially if we don't need it for normal pledges. It's not used where the other params are used, and in fact, if I remember correctly, we don't actually use the `ApplePayParams` in the late pledge flow anymore anyways (they're just not cleaned up properly since this was part of that quick patch for fixing apple pay). It'd be nice but not strictly necessary to clean up the apple pay params as part of this work & only pass the info the view model needs.

ifosli (approved):
A couple comments but not blocking for this pr. However, please make sure you actually test this in prod if you haven't already! With the release timeline, I'm okay with just testing that in the testflight build if that's easier and doing a rollback if we need to, but let's make sure any apple pay changes are thoroughly tested before they're released to our users.

Library/ViewModels/PledgeViewModel.swift
scottkicks: Nope. ValidateCheckout is just for late pledge as of now.

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
amy-at-kickstarter: You are correct, we no longer need any info from the `ApplePayParams` in the VM: https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/PostCampaignCheckoutViewModel.swift#L348

amy-at-kickstarter (approved):
Also LGTM.

scottkicks:
@ifosli Noted in regression that we should thoroughly test this in TestFlight before releasing. I also cleaned up the ApplePayParams stuff. Good call out!

PR #2048, opened 2024-04-25
MBL-1390: Add Stripe Link

amy-at-kickstarter:
# 📲 What

Implement Stripe Link for iOS.

# 🤔 Why

Stripe Link allows backers to store their card information with Stripe, and then auto-complete it when they use the Stripe payment sheet in Kickstarter. We're implementing this because it's a nicer user experience in checkout.

# 🛠 How

When Stripe Link returns, it gives us payment display information specific to Stripe Link - i.e. it was giving us a the Link icon and the label "Stripe Link". 

Screenshot 2024-04-25 at 9 53 49 AM

To work around this, Stripe suggested we fetch the underlying payment method associated with the link, which allows us to grab the last four digits of the card and its card image. This makes Stripe Link function the way a manually added card does, and looks better in our UI.

Screenshot 2024-05-06 at 11 37 03 AM

# 👀 See

https://github.com/kickstarter/ios-oss/assets/146007185/5c1f5b79-22ec-405f-9b23-0ddb8cef99bc

# ✅  Regression testing before merging
- [ ] Pay with Stripe Link in live pledge flow
- [ ] Pay with Stripe Link in late pledge flow
- [ ] Pay with manually added card in live pledge flow
- [ ] Pay with manually added card in late pledge flow

### DIFF
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -282,37 +282,11 @@ final class PledgePaymentMethodsViewController: UIViewController {
       return nil
     }
 
-    let formattedLabel = self.formatLinkLabel(paymentMethod.label)
+    let formattedLabel = KSRStripeLink.formatLinkLabel(paymentMethod.label)
     return PaymentSheetPaymentOptionsDisplayData(
       image: paymentMethod.image,
       label: formattedLabel ?? paymentMethod.label
     )
-  }
-
-  private func formatLinkLabel(_ label: String) -> String? {
-    // Link gives us a label like "Visa 1234"; reformat it to match our UI
-    do {
-      // Find 4 digits in the string
-      let regex = try NSRegularExpression(pattern: "\\d{4}")
-      let matches = regex.matches(
-        in: label,
-        range: NSRange(location: 0, length: label.count)
-      )
-
-      guard let match = matches.first else {
-        return nil
-      }
-
-      guard let range = Range(match.range, in: label) else {
-        return nil
-      }
-
-      let lastFour = label[range]
-      return "•••• \(lastFour)"
-
-    } catch {
-      return nil
-    }
   }
 
   private func updateAddNewPaymentMethodButtonLoading(state: Bool) {
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -292,6 +292,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
   private func formatLinkLabel(_ label: String) -> String? {
     // Link gives us a label like "Visa 1234"; reformat it to match our UI
     do {
+      // Find 4 digits in the string
       let regex = try NSRegularExpression(pattern: "\\d{4}")
       let matches = regex.matches(
         in: label,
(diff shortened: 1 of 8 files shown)


Library/TestHelpers/Stripe+PaymentMethod.swift
```diff
@@ -22,15 +22,15 @@ extension STPPaymentMethod {
   ])
 
   static let sampleStringPaymentOption: (STPPaymentMethod) -> PaymentSheet.PaymentOption = { paymentMethod in
-    PaymentSheet.PaymentOption.saved(paymentMethod: paymentMethod)
+    PaymentSheet.PaymentOption.saved(paymentMethod: paymentMethod, confirmParams: nil)
```
amy-at-kickstarter, line 25: Small change due to the Stripe SDK upgrade

Library/ViewModels/PaymentMethodSettingsViewModel.swift
```diff
@@ -157,6 +157,8 @@ public final class PaymentMethodSettingsViewModel: PaymentMethodsViewModelType,
             var configuration = PaymentSheet.Configuration()
             configuration.merchantDisplayName = Strings.general_accessibility_kickstarter()
             configuration.allowsDelayedPaymentMethods = true
+            configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
```
amy-at-kickstarter, line 160: This is how Stripe looks up the user for Stripe Link

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -223,6 +222,98 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
+
+    // For link payment types, fetch the underlying payment method so we can get at the card type and label
+
+    switch paymentSheetType {
+    case .setupIntent:
+      STPAPIClient.shared
+        .retrieveSetupIntent(
+          withClientSecret: clientSecret,
+          expand: ["payment_method"]
+        ) { [weak self] intent, _ in
```
amy-at-kickstarter, line 252: The callbacks for `retrieveSetupIntent` and for `retrievePaymentIntent` have different types, so they can't be shared.

amy-at-kickstarter:
Converting to draft so I can add a feature flag

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -223,6 +222,98 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
+    return PaymentSheetPaymentOptionsDisplayData(
+      image: paymentMethod.image,
+      label: formattedLabel ?? paymentMethod.label
+    )
+  }
+
+  private func formatLinkLabel(_ label: String) -> String? {
+    // Link gives us a label like "Visa 1234"; reformat it to match our UI
+    do {
+      let regex = try NSRegularExpression(pattern: "\\d{4}")
```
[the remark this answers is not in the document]
scottkicks: Also we may want this to be in a helper function so that the settings payment method view can use it as well.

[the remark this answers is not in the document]
amy-at-kickstarter: Haha I'll add a comment. It's just looking for 4 digits. As for the settings page, in this case it doesn't need it, because we never actually display/fetch the payment method there. On that page, it just stores the card immediately and we fetch the stored card from Kickstarter instead of from Stripe.

scottkicks (approved):
👍

PR #2050, opened 2024-04-29
[MBL-1393] Moves Stripe Intent Logic Into It's Own Service

scottkicks:
# 📲 What

Creates a new StripeIntentService to handle creating Payment and Setup Intents

# 🤔 Why

This logic is used in crowdfunding, late pledges, and account settings payment method flows. 
Creating a shared/reusable service makes things cleaner and easier to maintain. 

Hopefully, it makes consolidating late pledges and crowdfunding easier as well.

# 🛠 How

Creates a pretty simple StripeIntentService to handle both intent types.

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] No changes to late pledge flow
- [x] No changes to crowdfunding flow
- [x] No changes to managing payment method in account settings


Library/StripeIntentService.swift
scottkicks, line 1: Another option would be to make this a class and inject it into PostCampaignCheckoutViewModel's init. This approach felt cleaner, but maybe dependency injection would make for better testing?

scottkicks: We also don't really use dependency injection in any of our View Models

amy-at-kickstarter: I'd say before you go too deep into dependency injection, write some tests for `StripeIntentService` as well as for the VMs that are using it. I think that'll tell you if what you're doing works well enough for our purposes, or if we need to do something more complex with DI.

amy-at-kickstarter (changes requested):
This change looks perfectly good to me on its own, but since we're experimenting with new patterns, I have two thoughts to take it further:

1. `StripeIntentService` should have its own tests
2. This change helps us to not repeat the API call in multiple classes, which is great, but it still only encapsulates a single signal (the API request). What would it look like if we wrote a service that was a higher-order encapsulation of multiple signals, or signal transformations? What might take this further?

scottkicks:
@amy-at-kickstarter I added tests for `StripeIntentService`. My goal is to ensure that intent requests are made with the correct data and called only the expected number of times. 

I tried a bit of a different pattern than what we use elsewhere in the app. One callout I have is that we have tests behind the actual API request already that make sure the correct data is being passed in, so what I have might be a bit overkill. Lmk what you think.

> 2. This change helps us to not repeat the API call in multiple classes, which is great, but it still only encapsulates a single signal (the API request). What would it look like if we wrote a service that was a higher-order encapsulation of multiple signals, or signal transformations? What might take this further?  One small idea that comes to mind is that it looks like we really just want the `clientSecret` from this API call, so you could probably fold at least a `map` operation into it.

I'm not sure about this. For one thing, we need the result of the Signal (the envelope with the client secret _or_ the ErrorEnvelope), not just the client secret. I feel like keeping the `.map` in the calling method makes more sense so that it can handle the result appropriately.

Library/TestHelpers/MockStripeIntentService.swift
```diff
@@ -0,0 +1,51 @@
...
+@testable import Library
+import ReactiveSwift
+
+public struct StripeIntentRequestTestData: Equatable {
+  let projectId: String?
+  let pledgeTotal: Double?
+  let paymentIntentEnvelope: PaymentIntentEnvelope?
+  let context: GraphAPI.StripeIntentContextTypes?
+  let clientSeretEnvelope: ClientSecretEnvelope?
+}
```
scottkicks, line 11: One data structure to compare felt cleaner than having one for payment intents and one for setup intents. Also this way we can guarantee that only one intent request type (payment or setup) is being made at a time.

Library/TestHelpers/MockStripeIntentService.swift
```diff
@@ -0,0 +1,51 @@
...
+
+public class MockStripeIntentService: StripeIntentServiceType {
+  public private(set) var intentRequests = [StripeIntentRequestTestData]()
```
amy-at-kickstarter, line 14: Is this ever read, or only ever set? I don't see it used anywhere

Library/TestHelpers/XCTestCase+AppEnvironment.swift
```diff
@@ -71,6 +72,7 @@ extension XCTestCase {
         pushRegistrationType: pushRegistrationType,
         remoteConfigClient: remoteConfigClient,
         scheduler: scheduler,
+        stripeIntentService: stripeIntentService,
```
amy-at-kickstarter, line 75: Interesting idea to move this to `AppEnvironment`! Will that be sustainable when we have a lot of these services?

Library/ViewModels/PostCampaignCheckoutViewModelTests.swift
```diff
@@ -455,7 +455,7 @@ final class PostCampaignCheckoutViewModelTests: TestCase {
       self.scheduler.run()
 
       self.goToApplePayPaymentAuthorization.assertDidNotEmitValue()
-      self.processingViewIsHidden.assertValues([false, true])
+      self.processingViewIsHidden.assertValues([true, false])
```
amy-at-kickstarter, line 458: Was this intentional? Surprised to see this swapped

Library/TestHelpers/MockStripeIntentService.swift
scottkicks: I'm appending to this array in the create methods below and then referencing it in my test assertions in `StripeIntentServiceTests`. [👀 ](https://github.com/kickstarter/ios-oss/pull/2050/files#diff-085de37a469a9dbfb6ea20e854e7a1aef83a6345c17037e4ef7841d28ee9ce6bR41)

Library/TestHelpers/XCTestCase+AppEnvironment.swift
scottkicks: Ohh that's a good point. _Probably not_ seeing as how AppEnvironment is already pretty large. That could be a good argument for just injecting this class into the ViewModel's that need it.

Library/ViewModels/PostCampaignCheckoutViewModelTests.swift
scottkicks: whoops. I can fix this. Although the tests all passed, so that's interesting...

Library/TestHelpers/MockStripeIntentService.swift
amy-at-kickstarter: Ah, I see, it's used in the tests. I'm curious why you're tracking this separately instead of just using the test methods that test signals?

scottkicks: The main reason is that keeping track of each signal request in an array helps us guarantee that only the expected number of requests are made and with the correct request types/data.

Do we have test helpers that do this for us? 😅

amy-at-kickstarter: Hm, it might just be that the tests seem a bit tautological to me - you call the mock service, then validate that the mock service was called, instead of validating its output. An alternative could be to use a mock `apiService` in your mock service, to mock out the networking calls, and then that would validate the whole networking chunk?

amy-at-kickstarter: How many times can I say "mock" in one comment 😆

Library/TestHelpers/XCTestCase+AppEnvironment.swift
amy-at-kickstarter: Yeah, I'm OK with just injecting it where we need it. Unless we're starting to see issues with the tests for the VMs that use the `stripeIntentService` I'm not sure we need this level of DI

scottkicks:
@amy-at-kickstarter, I did the following from our Slack discussion:
- Removed the service from `Environment` and injected it only into the ViewModels needing it.
- Updated `MockStripeIntentService` so that it uses apiService and keeps track of the number of requests made
- Added test assertions in all relevant ViewModels that guarantee that we're only requesting the right intents and making those requests the correct number of times.

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodSettingsViewController.swift
```diff
@@ -13,7 +13,8 @@ protocol PaymentMethodSettingsViewControllerDelegate: AnyObject {
 internal final class PaymentMethodSettingsViewController: UIViewController,
   MessageBannerViewControllerPresenting {
   private let dataSource = PaymentMethodsDataSource()
-  private let viewModel: PaymentMethodsViewModelType = PaymentMethodSettingsViewModel()
+  private let viewModel: PaymentMethodsViewModelType =
+    PaymentMethodSettingsViewModel(stripeIntentService: StripeIntentService())
```
amy-at-kickstarter, line 17: I like this pattern for dependencies!

Library/TestHelpers/MockStripeIntentService.swift
```diff
@@ -0,0 +1,46 @@
...
+  public init() {}
+
+  public func createPaymentIntent(
+    for projectId: String,
+    pledgeTotal: Double
+  ) -> SignalProducer<PaymentIntentEnvelope, ErrorEnvelope> {
+    assert(
+      AppEnvironment.current.apiService as? MockService != nil,
+      "AppEnvironment.current.apiService should be a MockService when used in test."
+    )
```
amy-at-kickstarter, line 18: 👍 nice assert

amy-at-kickstarter (approved):
LGTM, I like how this shaped up.

PR #2059, opened 2024-05-14
[MBL-1439] Remove 'Block Users' Feature Flag

scottkicks:
# 📲 What

Removes Block Users Feature Flag.

# 🤔 Why

It has been on at 100% for some time now. It's time to say goodbye.
Also I want to focus on cleaning our flags up more consistently. 

# 🛠 How

- Remove flag from `RemoteConfigFeature` and all subsequent tests and helpers
- Remove any guards that use this feature flag.

# 👀 See

No Changes

# ♿️ Accessibility 

No changes

# 🏎 Performance

Same

# ✅ Acceptance criteria

- [x] Blocking users still works as expected. Areas where we can block users:
   - Creator profile icon, Comments, Comment Replies, Messages

# ⏰ TODO

- [x] Create an [Epic](https://kickstarter.atlassian.net/browse/MBL-1440) to track cleaning up the other flags that can be removed


ifosli (approved):
Nice! Once this is in, we should also update the flag in remote config to stop sending it to new builds (and maybe eventually delete it entirely, assuming we're okay with reverting block user functionality for the builds that depended on the flag)

scottkicks:
> Nice! Once this is in, we should also update the flag in remote config to stop sending it to new builds (and maybe eventually delete it entirely, assuming we're okay with reverting block user functionality for the builds that depended on the flag)

@ifosli Agreed! I have a condition ready to be published once that says that anyone on the current 5.15.0 version or below will respond as if this flag is set to true/enabled. 
Screenshot 2024-05-14 at 1 23 00 PM

Screenshot 2024-05-14 at 1 24 52 PM

PR #2061, opened 2024-05-15
Fix Stripe Link remote configuration flag

amy-at-kickstarter:
# 📲 What

Fix the Stripe Link remote configuration flag, so that it _actually_ controls whether or not Stripe Link is enabled.

# 🤔 Why

Apparently setting the email address was insufficient to control the appearance of Stripe Link. 🤷‍♀️

### DIFF
--- a/Library/ViewModels/PaymentMethodSettingsViewModel.swift
+++ b/Library/ViewModels/PaymentMethodSettingsViewModel.swift
@@ -159,10 +159,6 @@ public final class PaymentMethodSettingsViewModel: PaymentMethodsViewModelType,
             configuration.merchantDisplayName = Strings.general_accessibility_kickstarter()
             configuration.allowsDelayedPaymentMethods = true
 
-            if featureStripeLinkEnabled() {
-              configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
-            }
-
             let data = PaymentSheetSetupData(
               clientSecret: envelope.clientSecret,
               configuration: configuration,
--- a/Library/ViewModels/PledgePaymentMethodsViewModel.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModel.swift
@@ -354,6 +354,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
 
           if featureStripeLinkEnabled() {
             configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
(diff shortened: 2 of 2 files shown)


Library/ViewModels/PaymentMethodSettingsViewModel.swift
```diff
@@ -159,10 +159,6 @@ public final class PaymentMethodSettingsViewModel: PaymentMethodsViewModelType,
             configuration.merchantDisplayName = Strings.general_accessibility_kickstarter()
             configuration.allowsDelayedPaymentMethods = true
 
-            if featureStripeLinkEnabled() {
```
amy-at-kickstarter, line 162: Per @tedPen, we don't need to show Stripe Link from the payment method settings screen - only in the pledge flow. Removing this for clarity's sake, since it would never show anyways.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -354,6 +354,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
 
           if featureStripeLinkEnabled() {
             configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
+            configuration.billingDetailsCollectionConfiguration.email = .automatic
+          } else {
+            configuration.billingDetailsCollectionConfiguration.email = .always
```
amy-at-kickstarter, line 359: @ycheng-kickstarter Can you confirm this is what Android is doing? Stripe's documentation is stunningly unclear on this, but it does work for me.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -354,6 +354,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
 
           if featureStripeLinkEnabled() {
             configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
+            configuration.billingDetailsCollectionConfiguration.email = .automatic
```
amy-at-kickstarter, line 357: `.automatic` is the default but setting it explicitly for clarity

scottkicks (approved):

Library/ViewModels/PledgePaymentMethodsViewModel.swift
ycheng-kickstarter: Yup we match

ycheng-kickstarter (approved):

PR #2065, opened 2024-05-21
MBL-1421: Fix animated webp images

amy-at-kickstarter:
# 📲 What

Animated `.webp` images on the Campaign page remained, stubbornly, static. This adds the `KingfisherWebp` package, using a webp-compatible image serializer and processor.

# 🤔 Why

It turns out that our helper method `ksr_cacheImageWith(url:completionHandler)` was returning only the first frame of any animated `webp` files. I did a lot of digging into Kingfisher (our image display library), and figured out that Kingfisher was _downloading_ the entire image, but when it instantiated the data, dropped all but the first frame. 

Frustratingly, Kingfisher allegedly supports `webp` out of the box, as does iOS; [but it was clear on reading the code](https://github.com/onevcat/Kingfisher/blob/master/Sources/Image/Image.swift#L389) that `webp` was falling into an 'other' bucket and that's presumably why we couldn't read these images.

# 👀 See

One thing I realized was that we're pre-fetching all the image data for the entire project page; for this particular test project (which has many webp images) that's like >100MB of data.

Screenshot 2024-05-22 at 9 03 09 AM

I tested this code before-and-after, and the allocations/leaks looked the same, so I don't believe my changes have caused any regressions. But I did file a new ticket to address that issue, including all my investigation: https://kickstarter.atlassian.net/browse/MBL-1506

# 👀 See also

Finally, I've been seeing an odd issue where (on this one project), sometimes the animated webp images are pixelated, and sometimes they are not. Scrolling past the pixelated image and scrolling back frequently causes it to reset to a crisper version. 

blurry-and-not-blurry

I spent a lot of time trying to see if it was some issue with the images being re-fetched, but couldn't identify the cause. I'm not sure this is worth blocking this PR over (since IMO animated-and-pixelly is better than unanimated-and-crisp), but thoughts and further investigation are welcome!


Kickstarter-iOS/Library/UIImageView+URL.swift
```diff
@@ -36,11 +37,20 @@ extension UIImageView {
 
   public static func ksr_cacheImageWith(
     _ url: URL,
-    serializer _: CacheSerializer = DefaultCacheSerializer(),
     completionHandler: @escaping (UIImage?) -> Void
   ) {
+    /* WebPProcessor and WebPSerializer fall back to
+     DefaultImageProcessor and DefaultCacheSerializer, respectively,
+     for non-webp images */
```
amy-at-kickstarter, line 44: See: 

https://github.com/yeatse/KingfisherWebP/blob/main/Sources/WebPProcessor.swift#L29
https://github.com/yeatse/KingfisherWebP/blob/main/Sources/WebPSerializer.swift#L37

ifosli (approved):
Good job! This was (and still is if we want to keep making this better) a tricky one!

amy-at-kickstarter:
> Good job! This was (and still is if we want to keep making this better) a tricky one!

And like all good tricky bugs, ends up being a tiny two-line fix 😆

PR #2072, opened 2024-05-29
Turn PledgePaymentMethodsAndSelectionData into a struct

amy-at-kickstarter:
# 📲 What

`PledgePaymentMethodsAndSelectionData` is a model object which is used to describe the state of the `PledgePaymentMethodsViewController` and `ViewModel`. 

I changed it from a large tuple into a `struct`. 

# 🤔 Why

I'm doing a lot of refactoring in this view controller and view model - this was the first, and most straightforward, step. This change makes it substantially easier to rename and change the types of data used by `PledgePaymentMethodsViewController`.


Library/ViewModels/PledgePaymentMethodsAndSelectionData.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+public struct PledgePaymentMethodsAndSelectionData {
+  public let paymentMethodsCellData: [PledgePaymentMethodCellData]
+  public let paymentSheetPaymentMethodsCellData: [PaymentSheetPaymentMethodCellData]
+  public let selectedPaymentMethod: PaymentSourceSelected?
+  public let shouldReload: Bool
+  public let isLoading: Bool
+
+  public func updateWithPaymentMethodsData(_ newPaymentMethodsData: [PledgePaymentMethodCellData])
```
amy-at-kickstarter, line 10: These methods replace the use of the Prelude templating for this tuple - they return new copies of the object, mutating the correct variable.

amy-at-kickstarter: There is probably a smarter and less verbose way to do all of these 🤷‍♀️

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -227,16 +224,14 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
 
-          let selectionUpdatedData = updatedData
-            |> \.paymentMethodsCellData .~ cellData(data.paymentMethodsCellData, selecting: nil)
-            |> \.paymentSheetPaymentMethodsCellData .~ selectedAllSheetPaymentMethods
-            |> \.selectedPaymentMethod .~
-            (
-              usePaymentIntent ? .paymentIntentClientSecret(clientSecret) :
-                .setupIntentClientSecret(clientSecret)
-            )
+          var selectionUpdatedData = updatedData
```
amy-at-kickstarter, line 231: One thing I want to poke at a bit while we're talking about deprecating Prelude - this pattern does lose us the ability to use `let` here (like `let selectionUpdatedData = updatedData...`), but because `selectionUpdatedData` is a struct, it effectively does the same thing. We're still creating a new copy of `updatedData` with a couple mutated properties, and we're doing it in substantially less code.

Thoughts?

amy-at-kickstarter: cc @ifosli since you're not on the review

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -202,13 +194,18 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
-        let updatedData = data
-          |> \.paymentMethodsCellData .~ []
-          |> \.paymentSheetPaymentMethodsCellData .~ []
-          |> \.selectedPaymentMethod .~ nil
-          |> \.isLoading .~ false
-          |> \.shouldReload .~ true
+      .map { (
+        data: PledgePaymentMethodsAndSelectionData,
+        indexPath: IndexPath,
+        paymentSheetType: PledgePaymentSheetType
```
scottkicks, line 200: Nice I really appreciate seeing the types in line like this

scottkicks (approved):
Nice!

PR #2074, opened 2024-05-30
MBL-1449: Implement createOrUpdateBackingAddress mutation

amy-at-kickstarter:
# 📲 What

Add the `createOrUpdateBackingAddress` to the app.

# 🤔 Why

This mutation is used to confirm a backer's address for a given backing; it will be used in the PPO dashboard.


KsApi/mutations/ConfirmBackingAddressMutationTests.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+@testable import KsApi
+import XCTest
+
+final class ConfirmBackingAddressMutationTests: XCTestCase {
```
amy-at-kickstarter, line 5: These tests are a little silly since they're just exercising the `MockService` 🤷‍♀️. But at least they provide a little documentation of what the Service call can do.

ifosli (approved):

PR #2075, opened 2024-05-31
MBL-1488 & MBL-1360: Refactor Payment Sheet

amy-at-kickstarter:
# 📲 What

Refactor `PledgePaymentMethodsViewController` and `PledgePaymentMethodsViewModel` so that we always use a `SetupIntent` and store the card immediately, regardless of which pledge flow we're using.

# 🤔 Why

Late Pledges use a `PaymentIntent` in Stripe. When I implemented Late Pledges for the payment method sheet, I added a lot of plumbing to make the sheet use a `PaymentIntent` or a `SetupIntent` depending on context.

However, we have since learned that this behavior leads to bugs - `PaymentIntent` and `SetupIntent` lead to subtly different behavior with the payment sheet. For example, when you confirm a payment sheet made with a `PaymentIntent` it actually _charges the card_, ultimately leading to an issue with "dangling" Stripe charges that aren't attached to a pledge.

This refactor not only undoes all that, but greatly simplifies the payment sheet by making it _always_ save a card and _always_ returned a card that has been saved to the Kickstarter backend. 

# 🛠 How

Each commit should be fairly atomic (although they can't be shipped separately as a functioning feature). I recommend reviewing this PR commit-by-commit.

# ✅ Acceptance criteria

I will run through _all_ this acceptance criteria before merging this change.

Verify that adding a card in both pledge flow succeeds, the pledge is marked with the correct card on the KSR backend, and the card is saved to the KSR server:

- [x] Add a new card in the pledge flow, charge using that card
- [x] Add a new card in the pledge flow, select an existing card, charge using that card
- [x] Charge using an existing card in the pledge flow
- [x] Add a new card in the late pledge flow, charge using that card
- [x] Add a new card in the late pledge flow, select an existing card, charge using that card
- [x] Charge using an existing card in the late pledge flow
- [x] Charge using a 3DS challenge card in the live pledge flow - Worked but didn't ask for confirmation
- [x] Charge using a 3DS challenge card in the late pledge flow
- [x] Add a new card in the live pledge flow which is an ineligible card type for that pledge
- [x] Add a new card in the late pledge flow which is an ineligible card type for that pledge - Note that this fails and does not save the card, whereas live pledge flow DOES save the card in this case.

Verify that the late pledge "double charge" is fixed
- [x] Add a new card in the late pledge flow. Before hitting 'pledge', tap Back, then confirm again. You should be able to select a card and complete the pledge; you should _not_ see an error 

Verify context:
- [x] Verify that the setup intent pledge context is still set correctly for a live pledge
- [x] Verify that the setup intent pledge context is still set correctly for a late pledge

Verify that the payment sheet works correctly:
- [x] Able to select an existing card
- [x] Able to select a new card 
- [x] Able to add multiple new cards

Verify that the payment sheet works correctly when modifying a pledge:
- [x] Change payment method for a pledge, adding a new card
- [x] Change payment method for a pledge, selecting an existing card
- [x] Previously selected payment method cannot be selected when modifying the pledge

Verify that manage card works correctly:
- [x] Manage card settings, add a new card
- [x] Manage card settings, delete a card


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/PledgeView/Controllers/PostCampaignCheckoutViewController.swift
```diff
@@ -354,28 +354,7 @@ extension PostCampaignCheckoutViewController: PledgePaymentMethodsViewController
     _: PledgePaymentMethodsViewController,
     didSelectCreditCard paymentSource: PaymentSourceSelected
   ) {
-    switch paymentSource {
```
amy-at-kickstarter, line 357: @amy-at-kickstarter reminder for Wednesday - can you reinstate this? Makes more sense than fetching from the graph IMO.

Kickstarter-iOS/Features/PaymentMethods/Controller/PaymentMethodSettingsViewController.swift
```diff
@@ -207,12 +207,8 @@ internal final class PaymentMethodSettingsViewController: UIViewController,
 
       switch paymentResult {
       case .completed:
-        let paymentDisplayData = PaymentSheetPaymentOptionsDisplayData(
-          image: existingPaymentOption.image,
-          label: existingPaymentOption.label
-        )
         strongSelf.viewModel.inputs
-          .paymentSheetDidAdd(newCard: paymentDisplayData, setupIntent: clientSecret)
+          .paymentSheetDidAdd(setupIntent: clientSecret)
```
amy-at-kickstarter, line 211: This data was already unused; `PaymentMethodSettingsViewController` never read `paymentDisplayData`. It just used the `clientSecret`.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -49,7 +49,6 @@ final class PledgePaymentMethodsViewController: UIViewController {
       |> ksr_addSubviewToParent()
 
     self.tableView.registerCellClass(PledgePaymentMethodCell.self)
-    self.tableView.registerCellClass(PledgePaymentSheetPaymentMethodCell.self)
```
amy-at-kickstarter, line 52: n.B. this view controller still displays the list of payment methods in two sections, with the top section being 'new' cards and the bottom section being 'existing' cards. However, they use the same cell type and same backing data (a `UserCreditCards.CreditCard`).

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -80,8 +79,8 @@ final class PledgePaymentMethodsViewController: UIViewController {
       .observeValues { [weak self] data in
         guard let self = self else { return }
 
-        let cards = data.paymentMethodsCellData
-        let paymentSheetCards = data.paymentSheetPaymentMethodsCellData
+        let cards = data.existingPaymentMethods
```
amy-at-kickstarter, line 82: Renamed these properties for (hopeful) clarity.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -180,50 +173,43 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
-
       switch paymentResult {
       case .completed:
-        let paymentDisplayData = PaymentSheetPaymentOptionsDisplayData(
-          image: existingPaymentOption.image,
-          label: existingPaymentOption.label
-        )
-        strongSelf.viewModel.inputs
-          .paymentSheetDidAdd(newCard: paymentDisplayData, clientSecret: clientSecret)
+        // Fetch the stripe payment method ID
```
amy-at-kickstarter, line 195: This is a somewhat annoying workaround. When the view model calls `addPaymentSheetPaymentSource` to store the credit card, it does not return a `stripeCardId` in the `UserCreditCard.CreditCard` in the payload. Payments has opened a ticket about this - but for now, we fetch the `stripeCardId` directly from the Stripe API, and pass it back up to the view model.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -180,50 +173,43 @@ final class PledgePaymentMethodsViewController: UIViewController {
...
+            // The payment method ID is only required for late pledges.
+            strongSelf.viewModel.inputs
+              .paymentSheetDidAdd(
+                clientSecret: clientSecret,
+                paymentMethod: intent?.paymentMethodID
+              )
+          }
       case .canceled:
-        // User cancelled intentionally so do nothing.
-        break
```
amy-at-kickstarter, line 225: huh, guess our linter strips `break` statements

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewControllerTests.swift
```diff
@@ -51,9 +51,7 @@ final class PledgePaymentMethodsViewControllerTests: TestCase {
           checkoutId: "checkoutId",
           reward: reward,
           context: .pledge,
-          refTag: nil,
-          pledgeTotal: Double.nan,
-          paymentSheetType: .setupIntent
```
amy-at-kickstarter, line 56: n.B. that this stuff doesn't need to be passed around any more, because the payment sheet no longer needs a total or a type.

Kickstarter-iOS/Features/PledgePaymentMethods/Datasource/PledgePaymentMethodsDataSourceTests.swift
```diff
@@ -42,18 +42,18 @@ final class PledgePaymentMethodsDataSourceTests: XCTestCase {
   func testLoad_PaymentSheetCardValues() {
     let paymentSheetData = [
       (
-        image: UIImage(),
```
amy-at-kickstarter, line 45: Data source now uses the same data type for new cards and existing cards, as mentioned.

KsApi/models/templates/graphql/GraphCreditCardTemplate.swift
```diff
@@ -5,56 +5,64 @@ extension UserCreditCards {
...
+    stripeCardId: "pm_fake_jcb"
   )
 
   public static let discover = UserCreditCards.CreditCard(
     expirationDate: "2022-03-12",
     id: "5",
     lastFour: "4242",
-    type: .discover
+    type: .discover,
+    stripeCardId: "pm_fake_discover"
```
amy-at-kickstarter, line 41: We need these or else some of my new asserts will get triggered.

Library/ViewModels/PaymentMethodSettingsViewModel.swift
```diff
@@ -86,25 +85,6 @@ public final class PaymentMethodSettingsViewModel: PaymentMethodsViewModelType,
     )
 
     let newSetupIntentCards = self.newSetupIntentCreditCardProperty.signal.skipNil()
-      .map { data -> PaymentSheetPaymentMethodCellData? in
```
amy-at-kickstarter, line 89: This was all unused!

Library/ViewModels/PaymentMethodSettingsViewModelTests.swift
```diff
@@ -127,7 +127,6 @@ internal final class PaymentMethodSettingsViewModelTests: TestCase {
 
       self.vm.inputs
         .paymentSheetDidAdd(
-          newCard: paymentOptionsDisplayData,
           setupIntent: "seti_1LVlHO4VvJ2PtfhK43R6p7FI_secret_MEDiGbxfYVnHGsQy8v8TbZJTQhlNKLZ"
```
amy-at-kickstarter, line 130: @ifosli I hope these are test client secrets, otherwise we may have another thing to clean up 😅

Library/ViewModels/PaymentSourceSelected.swift
```diff
@@ -1,33 +1,26 @@
...
-      return nil
-    }
-  }
+/* PledgePaymentMethodsViewController used to return multiple types of selected payments.
+ Now it can only return stored credit cards. Leaving this enum here in case we ever add
+ more payment types in the future.
+ */
+
+public typealias StripePaymentMethodID = String
+public typealias KSRCreditCardId = String
```
amy-at-kickstarter, line 9: Just trying to make things clearer than `case savedCreditCard(String, String?)` which is what it really is.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -519,45 +512,53 @@ private func pledgePaymentSheetMethodCellDataAndSelectedCardSetupIntent(
...
+    updatePaymentMethodData.newPaymentMethods = data
+
+    if newestPaymentSheetPaymentMethod.isEnabled {
+      if let stripeCardId = newestPaymentSheetPaymentMethod.card.stripeCardId {
+        updatePaymentMethodData
+          .selectedPaymentMethod = .savedCreditCard(newestPaymentSheetPaymentMethod.card.id, stripeCardId)
+      } else {
+        assert(false, "Expected stripeCardId to be set. Late pledges may fail if this value is missing.")
+        updatePaymentMethodData
+          .selectedPaymentMethod = .savedCreditCard(newestPaymentSheetPaymentMethod.card.id, "")
```
amy-at-kickstarter, line 559: My goal was to make sure if `stripeCardId` failed anywhere, the code would still _mostly_ work, since  `stripeCardId` is only important for late pledges.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -110,19 +100,46 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
     )
     .map { ($0.0, $0.1, $0.2, false) }
 
-    let newSetupIntentCards = self.newStripeIntentCreditCardProperty.signal.skipNil()
-      .map { data -> [PaymentSheetPaymentMethodCellData] in
-        let (displayData, setupIntent) = data
+    let newlyAddedCardProducer = self.newStripeIntentCreditCardProperty.signal.skipNil()
```
amy-at-kickstarter, line 103: This is the beefiest part of the change - when a new card is selected, add it to the Kickstarter backend, then populate the payment sheet with the actual `UserCreditCards.CreditCard` returned by that mutation.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -110,19 +100,46 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
-        return [(
-          image: displayData.image,
-          redactedCardNumber: displayData.label,
-          clientSecret: setupIntent,
-          isSelected: false,
-          isEnabled: true
-        )]
+    let newCardsData: Signal<([PledgePaymentMethodCellData], UserCreditCards.CreditCard?), Never> = Signal
+      .combineLatest(
+        newlyAddedCardProducer.values(),
```
amy-at-kickstarter, line 131: I haven't added anything for `newlyAddedCardProducer.errors()`, not sure what to do with that. I guess display to the user?

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -183,80 +162,41 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
           paymentIntentClientSecret: paymentIntentClientSecret,
-          selectedCardStripeCardId: selectedCardStripeCardId,
+          selectedCardStripeCardId: selectedCard.stripePaymentMethodId,
           requiresConfirmation: true
         )
       }
 
-    // MARK: - Validate New Cards
-
-    let validateCheckoutNewCardInput = Signal.combineLatest(checkoutId, selectedCard)
```
amy-at-kickstarter, line 228: All of this can go away, because all cards are now "existing" cards (i.e. stored to the KSR backend)

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -126,15 +126,20 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
           .materialize()
       }
 
+    let allNewlyAddedCards: Signal<[UserCreditCards.CreditCard], Never> = newlyAddedCardProducer.values()
+      .scan(into: []) { array, card in
+        array.insert(card, at: 0)
```
amy-at-kickstarter, line 131: I forgot the `scan` at first - this turns a signal like `[a, b, c]` into a signal like `[[a], [a, b], [a, b, c]]`.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -100,28 +100,46 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+        return AppEnvironment.current.apiService.addPaymentSheetPaymentSource(input: input)
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
-          .map { (envelope: CreatePaymentSourceEnvelope) in envelope.createPaymentSource }
+          .map { (envelope: CreatePaymentSourceEnvelope) -> UserCreditCards.CreditCard? in
+            guard envelope.createPaymentSource.isSuccessful else {
+              return nil
+            }
+
+            // There is a backend bug where the stripeCardId isn't being refreshed
+            // after the add card mutation is called. This adds it back in.
```
ifosli, line 117: nit: It'd be nice to link to a ticket here so we can check if the bug still exists the next time we touch this class, especially if we want to remove this code.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -146,7 +164,12 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
 
         var selectedPaymentMethod: PaymentSourceSelected?
         if let selectedCardId = selectedCard?.id {
-          selectedPaymentMethod = .savedCreditCard(selectedCardId)
+          if let stripeCardId = selectedCard?.stripeCardId {
+            selectedPaymentMethod = .savedCreditCard(selectedCardId, stripeCardId)
+          } else {
+            assert(false, "Expected stripeCardId to be set. Late pledges may fail if this value is missing.")
```
ifosli, line 175: Just confirming: this assert only crashes debug builds? Would we want to log this as an error to firebase for release builds? Up to you if you think that's worth doing or not. (Probably out of scope for this, but I think it'd be nice to have an assert helper function that crashed debug builds and logged to firebase for beta/release builds.)

Library/ViewModels/PledgePaymentMethodsViewModel.swift
ifosli: I like that! nit: I do think it'd be nice to pull this out into a helper function, though, considering we're doing basically the same checks in about 5 places.

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -179,37 +162,37 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
     // Runs validation for pre-existing cards that were created with setup intents originally but require payment intents for late pledges.
     let validateCheckoutExistingCard = validateCheckoutExistingCardInput
       .takeWhen(self.submitButtonTappedProperty.signal)
-      .switchMap { checkoutId, selectedExistingCreditCard, paymentIntentClientSecret, storedCards in
-        let selectedStoredCard = storedCards.first { $0.id == selectedExistingCreditCard.savedCreditCardId }
-        let selectedCardStripeCardId = selectedStoredCard?.stripeCardId ?? ""
+      .switchMap { checkoutId, selectedCard, paymentIntentClientSecret in
+
+        assert(
+          selectedCard.stripePaymentMethodId != nil,
```
ifosli, line 174: nit: Maybe check that the string is non-empty here instead of just non-nil (just in case we default to empty string ever in the future)?

ifosli (commented):
I've only reviewed 4/7 commits (hence the "comment" instead of "approve") but I'll get to the rest on monday, but since you're probably starting work before me I figured I'd give you something to work with if you'd like. So far I think it looks good - just some nits that you're welcome to ignore if you disagree.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -519,45 +509,49 @@ private func pledgePaymentSheetMethodCellDataAndSelectedCardSetupIntent(
   }()
 
   // First ensure all existing payment sheet payment methods are not selected.
-  let preexistingPaymentSheetCardDataUnselected: [PaymentSheetPaymentMethodCellData] = {
-    var updatedPaymentSheetPaymentMethodData = paymentMethodData.paymentSheetPaymentMethodsCellData
+  let preexistingPaymentSheetCardDataUnselected: [PledgePaymentMethodCellData] = {
```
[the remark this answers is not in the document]
amy-at-kickstarter: Will do!

ifosli (approved):
Looked through the last 3 commits and they look good!

Library/ViewModels/PledgePaymentMethodsViewModel.swift
amy-at-kickstarter: To the best of my knowledge, yes - we have assertions disabled at the compiler level for production builds:
Screenshot 2024-06-10 at 11 54 18 AM

I loooove the idea of something that has both an assertion and a log to firebase! Let me think about that.

Library/ViewModels/PledgePaymentMethodsViewModel.swift
amy-at-kickstarter: Good suggestion, fixed!

Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -110,19 +100,46 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
...
+            // after the add card mutation is called. This adds it back in.
+            var card = envelope.createPaymentSource.paymentSource
+            if card.stripeCardId == nil {
+              card.stripeCardId = paymentMethodId
+            }
+
+            return card
+          }
+          .skipNil()
+          .materialize()
```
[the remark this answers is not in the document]
amy-at-kickstarter: Will do.

PR #2076, opened 2024-06-10
MBL-1390: Re-enable Stripe Link

amy-at-kickstarter:
# 📲 What

Turn on Stripe Link.

# 🛠 How

This is _much_ simpler now that the payments sheet has been refactored, and since we also decided to nix feature flags for Stripe Link! All we need to do is upgrade the Stripe SDK and pass in the user's e-mail address.

# 👀 See
Screenshot 2024-06-10 at 4 19 59 PM

### DIFF
--- a/Library/ViewModels/PledgePaymentMethodsViewModel.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModel.swift
@@ -325,6 +325,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
           configuration.merchantDisplayName = Strings.general_accessibility_kickstarter()
           configuration.allowsDelayedPaymentMethods = true
 
+          // Enable Stripe Link
+          configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
+
           let data = PaymentSheetSetupData(
             clientSecret: clientSecret,
             configuration: configuration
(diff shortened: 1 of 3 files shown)


Library/ViewModels/PledgePaymentMethodsViewModel.swift
```diff
@@ -325,6 +325,9 @@ public final class PledgePaymentMethodsViewModel: PledgePaymentMethodsViewModelT
           configuration.merchantDisplayName = Strings.general_accessibility_kickstarter()
           configuration.allowsDelayedPaymentMethods = true
 
+          // Enable Stripe Link
+          configuration.defaultBillingDetails.email = AppEnvironment.current.currentUserEmail
```
ifosli, line 329: Nit: If enabling stripe link is now this easy, is it a change we could/would want to flag guard after all?

ifosli (approved):

Library/ViewModels/PledgePaymentMethodsViewModel.swift
amy-at-kickstarter: This is what happens if you don't pass in the e-mail - Stripe link is still there, i
Screenshot 2024-06-10 at 4 37 49 PM

This issue with the feature flag is that Stripe Link is effectively controlled by server-level settings, there's no way to enable/disable it purely on the client (without doing a ton of extra work on both front-end and back-end).

amy-at-kickstarter:
Whoops, pushed a fix for the tests. Those were failing due to API changes in the SDK upgrade for Stripe.

PR #2077, opened 2024-06-10
[MBL-1450] Survey webview

ifosli:
# 📲 What

Update the survey webview to make all links clickable (excluding the profile links). This adds `/edit` and `/edit_address` survey links to our deeplinks; these'll simply open the survey webview as well.

Note: As part of this, I cleaned up some code that we're no longer relying on; the "survey submitted" alert doesn't show anymore anyways and we had a workaround for iOS versions below iOS 13, which we no longer support.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1450)

[Followup ticket for the profile links](https://kickstarter.atlassian.net/browse/MBL-1532)

| Open survey | Closed survey |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro Max - 2024-06-10 at 14 53 33 | Simulator Screen Recording - iPhone 15 Pro Max - 2024-06-10 at 14 56 35 |

# ✅ Acceptance criteria

- [x] Links are clickable and open relevant views over the current view hierarchy
- [x] Survey can be dismissed using the close button
- [x] Survey can be filled out
- [x] `/edit` and `/edit_address` deeplinks also open the survey webview


Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -64,37 +57,59 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
...
+            (param, project, updateId)
+          }
+      }
+      .switchMap { (param: Param, project: Project, updateId: Int) in
+        AppEnvironment.current.apiService.fetchUpdate(updateId: updateId, projectParam: param)
+          .demoteErrors()
+          .map { update -> (Project, Update) in
+            (project, update)
+          }
+      }
```
amy-at-kickstarter, line 109: Nit: will this show a spinner or some other UI while it makes two requests to the Kickstarter backend? Seems like it might be slow.

amy-at-kickstarter (approved):
LGTM.

My only thought is that it might make more sense to push to the update controller (or at least push some kind of loading UI) before fetching update data - I assume two chained requests will be somewhat slow on an actual phone.

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: That's a good point but I can't make the update page responsible for this unless I refactor the UpdateViewModel (very doable) and the ShareViewModel (looks kind of terrible, since the Project and Update are associated types for the config enum)... And adding a loading indicator between two webviews is not something I know how to do. I'll file a followup ticket to refactor the UpdateViewModel and ShareViewModel to make this possible.

amy-at-kickstarter: Solid, that makes sense.

ifosli: Ticket here: https://kickstarter.atlassian.net/browse/MBL-1533 for reference

PR #2080, opened 2024-06-12
MBL-1456: Stub PPO view and container

amy-at-kickstarter:
# 📲 What

Stub code for our new PPO dashboard.
ppo stub

# 🤔 Why

PPO development is moving full steam ahead; this will give us a place to put the work we're doing.

# 🛠 How

This consists of a few separate pieces:
*  `PagedContainerViewController` and `PagedContainerViewModel`, a container `UIViewController`. It displays a segmented control (which will need to be replaced with the _actual_ control). I realized I had to write this in UIKit, because the container needs to also hold the old `ActivitiesViewController`, and it seemed like it would be a bit of compatibility wizardry to get a SwiftUI view to contain a `UIViewController`. 
I did discover while doing this the `SortPagerViewController`, which has the selector control we do want - but that'll have to be either abstracted out or re-written, since that view controller is fairly specific to the `Discover` flow.
* `PPOContainerViewController`, a `PagedContainerViewController` which has tabs for the PPO view and Activity
* A swift view called `PPOView` and matching `PPOViewModel` to contain all our PPO content
* A stub test for `PPOViewModel`


Library/PagedContainer/PagedContainerViewController.swift
```diff
@@ -0,0 +1,138 @@
+import Combine
+import Foundation
+import UIKit
+
+open class PagedContainerViewController: UIViewController {
+  private weak var activeController: UIViewController? = nil
+  private var subscriptions = Set<AnyCancellable>()
+  private let viewModel = PagedContainerViewModel()
```
amy-at-kickstarter, line 8: Philosophically, I put everything that seemed like `UIKit` logic in the `ViewController`, and everything that seemed like "how it's displayed" logic in the view model. However, that ended up meaning the view model was _very small_ since it's a pretty simple concept for a page 🤷‍♀️

Library/PagedContainer/PagedContainerViewController.swift
```diff
@@ -0,0 +1,138 @@
+import Combine
```
amy-at-kickstarter, line 1: 🎉

Library/PagedContainer/PagedContainerViewController.swift
```diff
@@ -0,0 +1,138 @@
...
+import Foundation
+import UIKit
+
+open class PagedContainerViewController: UIViewController {
+  private weak var activeController: UIViewController? = nil
+  private var subscriptions = Set<AnyCancellable>()
+  private let viewModel = PagedContainerViewModel()
+
+  // TODO: Use the correct page control, per the designs.
+  // This may exist already in SortPagerViewController, or we can write one in SwiftUI.
```
amy-at-kickstarter, line 11: Worth calling out - we have another ticket for implementing the correct page control. I leave it up to someone else to determine if we should write a new one in SwiftUI or steal/refactor out the existing one.

Library/PagedContainer/PagedContainerViewController.swift
```diff
@@ -0,0 +1,138 @@
...
+    return false
+  }
+
+  public override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+
+    if self.toggle.selectedSegmentIndex == UISegmentedControl.noSegment {
+      self.viewModel.didSelectPage(atIndex: 0)
+    } else if let activeController = self.activeController {
+      activeController.beginAppearanceTransition(true, animated: animated)
```
amy-at-kickstarter, line 47: OK, so all this elaborate `appearanceTransition` stuff. This exists so that the child view controllers will receive `viewWillAppear` with `animated = true`. The reason we need _that_ is because `ActivitiesViewController` filters out calls that are unanimated. I was going to clean up `ActivitiesViewController`, instead, but it turns out that we implicitly use that `animated` as a filter for whether or not the `ActivitiesViewController` is being instantiated by a screenshot test; basically, it was rabbithole, and it seemed clearer to fix it this way.

Library/PagedContainer/PagedContainerViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+import UIKit
+
+public class PagedContainerViewModel {
+  // Internal
+  private var childViewControllers: [UIViewController] = []
+
+  // Inputs
+  func viewWillAppear() {
+    if self.selectedIndex.value == nil {
+      self.didSelectPage(atIndex: 0)
```
amy-at-kickstarter, line 12: Philosophically, normally we do this sort of thing by chaining signals on init...but is there any reason _not_ to do it like this? 🤔

amy-at-kickstarter: I could re-write this whole thing to be done on init, if anyone has a strong opinion about it. I didn't do it originally because I refactored this view model out of the view controller, so it's a little less FRP in design.

Library/PagedContainer/PagedContainerViewModel.swift
```diff
@@ -0,0 +1,48 @@
...
+        return nil
+      }
+    }.compactMap { $0 }
+    .eraseToAnyPublisher()
+  }
+
+  // Inputs
+  func viewWillAppear() {
+    if self.selectedIndex.value == nil {
+      self.didSelectPage(atIndex: 0)
```
amy-at-kickstarter, line 31: This tiny bit of state (in `self.selectedIndex`, a `CurrentValueSubject` instead of a straight `PassthroughSubject`), saves a lot of lines of confusing stateless code. I'm OK with that.

Library/PagedContainer/PagedContainerViewModel.swift
```diff
@@ -0,0 +1,48 @@
...
+public class PagedContainerViewModel {
+  // Internal
+  private var subscriptions = Set<AnyCancellable>()
+
+  init() {
+    self.pageTitles = self.configureWithChildrenSubject.map { controllers in
+      controllers.map { $0.title ?? "Page " }
+    }.eraseToAnyPublisher()
+
+    self.displayChildViewControllerAtIndex = Publishers.CombineLatest(
```
amy-at-kickstarter, line 14: I feel like this should really be a `takeWhen` but I can't find the Combine equivalent. We want the index + children when the index changes.

Kickstarter-iOS/Features/PostPledgeOverview/PPOView.swift
```diff
@@ -0,0 +1,12 @@
+import SwiftUI
```
amy-at-kickstarter, line 1: 🎉

Library/PagedContainer/PagedContainerViewController.swift
ifosli: Oh, I think that animated thing is what broke our surveys not showing up in activity, too. So it definitely might be worth looking into at some point but I agree that it's complicated and maybe not worth doing right now (and I definitely think it's not worth doing if there's a chance that we'll rewrite the view controller to stop basically being a webview).

ifosli: I do also think it might be worth leaving a comment in the code about this; it doesn't need to be as detailed, though!

Library/PagedContainer/PagedContainerViewModel.swift
ifosli: I don't think there is an equivalent? I don't think it matters for this one since I doubt the other signal can change. But if we really end up needing a `takeWhen` we may need to build our own. (Or, alternatively, store more state so that we don't need to grab everything from publishers?)

Library/PagedContainer/PagedContainerViewModel.swift
ifosli: I like it!

ifosli (approved):
Nice!! This feels like it's actually getting us started and I'm excited! Left a couple comments but everything looks pretty good apart from the "PostPledgeOverview" folder name being confusing (or at least new to me)

Kickstarter-iOS/Features/PostPledgeOverview/PPOContainerViewController.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import Library
+import SwiftUI
+
+public class PPOContainerViewController: PagedContainerViewController {
+  public override func viewDidLoad() {
+    super.viewDidLoad()
+
+    // TODO: Translate these strings
```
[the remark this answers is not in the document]
amy-at-kickstarter: Good idea, I'll add one.

Library/PagedContainer/PagedContainerViewController.swift
amy-at-kickstarter: Good idea, I'll just pull in this comment and edit it a bit.

Library/PagedContainer/PagedContainerViewModel.swift
amy-at-kickstarter: Fair point that the other signal (probably) won't ever change! The only situation I could see would be if we ever added a new page of content to this tab, and that page of content was affected by logged in/logged out state.

Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -4768,6 +4780,7 @@
 				1937A6F028C92F0000DD732D /* PledgeAmountSummary */,
 				19A97D3928C802230031B857 /* PledgePaymentMethods */,
 				1937A70428C9392600DD732D /* PledgeShippingLocation */,
+				E1BAA0332C1A18DA004F8B06 /* PostPledgeOverview */,
```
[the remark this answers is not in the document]
amy-at-kickstarter: Hahaha no you are correct, I keep mixing that up again and again 😆  Will fix.

PR #2081, opened 2024-06-18
MBL-1550: Don't change selected card when unavailable card is added to payment methods

amy-at-kickstarter:
# 📲 What

Don't change the selection on the Payments sheet if the newly added card is disabled.

# 🤔 Why

Now that we're saving cards immediately, it's possible to add a card that's not available for the current project. The previous logic would try and select the card, even if it was unavailable; this short-circuits that edge case. Fixing this prevents a bunch of unfortunate behavior as documented in the ticket.

### DIFF
--- a/Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
@@ -1975,7 +1975,7 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
 
       self.scheduler.run()
 
-      // There are now twos card, in the new section
+      // There are now two cards, in the new section
       XCTAssertEqual(self.reloadPaymentSheetPaymentMethodsCards.lastValue?.count, 2)
       XCTAssertEqual(self.reloadPaymentMethodsCards.lastValue?.count, 0)
 
--- a/Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
+++ b/Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
@@ -1786,4 +1786,231 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
       XCTAssertTrue(allowedDelayedPaymentMethods)
     }
   }
+
+  // A test for MBL-1550
+  func testAddNewUnavailablePaymentMethod_userAlreadyHasCard_selectsValidPaymentMethod() {
+    let cards = UserCreditCards.withCards([
+      UserCreditCards.visa
+    ])
+
+    let graphUser = GraphUser.template |> \.storedCards .~ cards
+    let graphUserResponse = UserEnvelope<GraphUser>(me: graphUser)
+    let addPaymentSheetResponse = CreatePaymentSourceEnvelope.paymentSourceSuccessTemplate
+
+    let mockService = MockService(
+      addPaymentSheetPaymentSourceResult: .success(addPaymentSheetResponse),
+      fetchGraphUserResult: .success(graphUserResponse)
+    )
+    let project = Project.template
+      |> \.availableCardTypes .~ ["VISA"]
+
+    withEnvironment(apiService: mockService, currentUser: User.template) {
+      self.vm.inputs
+        .configure(with: (
+          User.template,
+          project,
+          "checkoutID",
+          Reward.template,
+          .pledge,
+          .discovery
+        ))
+      self.vm.inputs.viewDidLoad()
+
+      self.scheduler.run()
+
+      // There is one card, it's enabled, and it's selected
+      XCTAssertEqual(self.reloadPaymentMethodsIsSelected.lastValue, [true])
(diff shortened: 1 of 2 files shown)


Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -1786,4 +1786,231 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
       XCTAssertTrue(allowedDelayedPaymentMethods)
     }
   }
+
+  // A test for MBL-1550
+  func testAddNewUnavailablePaymentMethod_userAlreadyHasCard_selectsValidPaymentMethod() {
```
amy-at-kickstarter, line 1791: One line to fix the bug, 229 lines to test it.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -1786,4 +1786,231 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+        )
+
+      self.scheduler.run()
+
+      // There are now twos card, in the new section
+      XCTAssertEqual(self.reloadPaymentSheetPaymentMethodsCards.lastValue?.count, 2)
+      XCTAssertEqual(self.reloadPaymentMethodsCards.lastValue?.count, 0)
+
+      // The discover is disabled and not selected.
+      let discoverCardData = self.reloadPaymentSheetPaymentMethodsCards.lastValue?.first(where: { (
```
amy-at-kickstarter, line 1983: Can someone remind me if there's a more beautiful way to do this search? Maybe with a keypath?

amy-at-kickstarter: FWIW I did this because the cards weren't ordered the way I expected, but I also realized it doesn't _matter_ what order two disabled cards are in.

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -1786,4 +1786,231 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+      // Add another new card (the discover)
+      self.vm.inputs
+        .paymentSheetDidAdd(
+          clientSecret: "si_fake",
+          paymentMethod: "pm_fake"
+        )
+
+      self.scheduler.run()
+
+      // There are now twos card, in the new section
```
ifosli, line 1978: minor nit: "two cards" instead of "twos card"

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
ifosli: I don't know a nicer way 😭 Tbh, if it's deterministic, I think I'd just go by the index (which would mean that the test would need to be changed if we change the order in the future, so I think I do like this search better).

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
```diff
@@ -1786,4 +1786,231 @@ final class PledgePaymentMethodsViewModelTests: TestCase {
...
+
+      // There is now one card, in the new section
+      XCTAssertEqual(self.reloadPaymentSheetPaymentMethodsCards.lastValue?.count, 1)
+      XCTAssertEqual(self.reloadPaymentMethodsCards.lastValue?.count, 0)
+
+      // The newly added card is disabled and not selected
+      let newCardData = self.reloadPaymentSheetPaymentMethodsCards.lastValue?.first
+      XCTAssertEqual(newCardData?.isEnabled, false)
+      XCTAssertEqual(newCardData?.isSelected, false)
+      XCTAssertNil(self.reloadPaymentMethodsSelectedCardId.lastValue ?? nil)
```
ifosli, line 1901: Nit: I haven't looked through all the tests again so this may be unnecessary, but should we also test that the pledge button doesn't get enabled in this test?

ifosli (approved):

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
scottkicks: ➕

scottkicks (approved):

Library/ViewModels/PledgePaymentMethodsViewModelTests.swift
amy-at-kickstarter: Ahh that's a great idea _but_ this test doesn't have the pledge button! It's just the isolated payment sheet. The test for pledges don't have access to the payment sheet itself. So I'm not sure we actually _can_ test the complete behavior of the bug.

amy-at-kickstarter: The `PledgeViewController` watches the signal/delegate call for when a card is selected, so testing that behavior should hopefully be sufficient.

PR #2082, opened 2024-06-18
[MBL-1553] Configure survey webview with url instead of SurveyResponse object

ifosli:
# 📲 What

Configure survey webview using a url string instead of an entire SurveyResponse object. This allows us to open deeplinks to new surveys as well as old surveys.

# 🛠 How

I'm reconstructing the url here using the current host and the path from the deeplink; I thought this had a good balance of being clear as well as definitely working. (I often do `ksr:// `deeplinks for testing, for example, and these would not be displayable in the webview directly if we were just using `.absoluteString` on the raw deeplink.)

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1553)

# ✅ Acceptance criteria

- [x] New surveys are openable from activity and deeplinks
- [x] Old surveys are openable from activity and deeplinks

# ⏰ TODO

Note: This pr does not fix push notifications, which will be fixed in a later pr.
Once push notifications depend on the url as well, I'll follow up to see if we can just delete the user survey path.

### DIFF
--- a/Library/Navigation.swift
+++ b/Library/Navigation.swift
@@ -47,7 +47,7 @@ public enum Navigation: Equatable {
     case pledge(Navigation.Project.Pledge)
     case updates
     case update(Int, Navigation.Project.Update)
-    case survey(Int)
+    case survey(Int, String)
 
     public enum Checkout: Equatable {
       case thanks(racing: Bool?)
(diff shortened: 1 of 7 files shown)


Library/Navigation.swift
```diff
@@ -47,7 +47,7 @@ public enum Navigation: Equatable {
     case pledge(Navigation.Project.Pledge)
     case updates
     case update(Int, Navigation.Project.Update)
-    case survey(Int)
+    case survey(Int, String)
```
amy-at-kickstarter, line 50: Are both guaranteed to be set? I thought in this case the surveyParam was `0`.

Library/NavigationTests.swift
```diff
@@ -126,7 +126,10 @@ public final class NavigationTests: XCTestCase {
       navigation: .update(2, .commentThread("dead", "beef"))
     )
 
-    self.assertProjectMatch(path: "/projects/creator/project/surveys/3", navigation: .survey(3))
+    self.assertProjectMatch(
+      path: "/projects/creator/project/surveys/3",
+      navigation: .survey(3, "https://www.kickstarter.com/projects/creator/project/surveys/3")
```
amy-at-kickstarter, line 131: What happens on staging/beta builds?

amy-at-kickstarter (approved):
I haven't touched our deeplinks before, so you may want to wait for more 👀 from @scottkicks. But overall LGTM, left a few non-blocking questions.

Library/Navigation.swift
amy-at-kickstarter: If not, it might make sense to split this enum into a type for survey-with-id and survey-with-url.

PR #2084, opened 2024-06-20
[MBL-1553] Make sure surveyVC is created on the main thread

ifosli:
# 📲 What

Make sure the survey view controller is created on the main thread. When I added the url support, I accidentally skipped this step when we created the view controller directly from the url. I'm not sure why this crash didn't show up in the simulator 🤷 

# ✅ Acceptance criteria

- [x] Opening a deeplink to the survey doesn't cause the app to crash


amy-at-kickstarter (approved):

amy-at-kickstarter:
Love a good crash. Nice find.

PR #2087, opened 2024-06-25
[MBL-1554] Fix survey push notifications for new surveys

ifosli:
# 📲 What

Allow new surveys to be opened from push notifications by using the url path provided instead of relying on the survey response id.

# 🤔 Why

New surveys don't have a unique id.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1554)

No screen recording of this because push notifications are a pain to work with, but I've verified that it works for new surveys and that push notification for old surveys have the exact same data format.

# ✅ Acceptance criteria

- [x] Clicking a push notification for a new survey opens the survey in the app.


scottkicks (approved):
LGTM 👍

PR #2088, opened 2024-06-25
[MBL-1547] Encapsulate Late Pledge Apple Pay Logic

scottkicks:
# 📲 What

Creates a new ApplePayCheckoutService to encapsulate late pledge Apple Pay Logic.

# 🤔 Why

Encapsulating checkout View Model dependencies that are shared across checkout to hopefully make consolidating late pledges and crowdfunding, and adding a new View Model for Pledge Redemption, easier and more maintainable.

This logic is used in crowdfunding and late pledges, but for now, we're just encapsulating it in the late pledge flow. 

Ideally, this gets shared in the crowdfunding VM, but the way ApplePay is being used in that flow there are a lot of differences, so I think that should be done separately.

# 🛠 How

Creates a new ApplePayCheckoutService and updates PostCampaignCheckoutViewModel to use it.

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] No changes to late pledge flow via ApplePay pledges
- [x] Make a note to test ApplePay late pledges in the next TestFlight build on Prod.


Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -197,92 +213,18 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
-      .skipNil()
+    /// See ApplePayCheckoutViewModel.swift
 
-    let validateCheckoutWithApplePay = Signal.combineLatest(
-      newPaymentIntentForApplePay,
-      checkoutId,
-      self.applePayPaymentMethodIdSignal.signal
+    let applePayErrors = Signal.merge(
+      self.applePayViewModel.createNewPaymentIntentError,
+      self.applePayViewModel.validateCheckoutError
```
ifosli, line 222: Nit: I kind of think having these two separate error outputs makes the PostCampaignCheckoutViewModel a little too responsible for knowing how the applePayViewModel works. What do you think of combining these? Or, alternatively, having one ex `silentError` and one `showError`? (Assuming I'm reading the code right and that not telling the user about the payment intent error is intentional?)

Library/ViewModels/PostCampaignCheckoutViewModel.swift
```diff
@@ -197,92 +213,18 @@ public class PostCampaignCheckoutViewModel: PostCampaignCheckoutViewModelType,
...
 
-    let validateCheckoutError = Signal
-      .merge(
-        validateCheckoutExistingCard.errors(),
-        validateCheckoutWithApplePay.errors()
-      )
+    let validateCheckoutExistingCardError = validateCheckoutExistingCard.errors()
 
-    self.showErrorBannerWithMessage = validateCheckoutError
+    self.showErrorBannerWithMessage = self.applePayViewModel.validateCheckoutError
```
ifosli, line 227: It looks like this will prevent errors from showing unless they're apple card errors - Shouldn't we merge the `validateCheckoutExistingCardError` and `self.applePayViewModel.validateCheckoutError` signals instead?

Library/ApplePayCheckoutViewModel.swift
```diff
@@ -0,0 +1,165 @@
...
+   2) Create a new payment intent
+   3) Present the payment authorization form
+   4) Payment authorization form calls applePayContextDidCreatePayment with the payment method Id
+   5) Payment authorization form calls paymentAuthorizationDidFinish
+   6) Validate checkout using the checkoutId, payment source id, and payment intent
+   7) Notify the parent View Model that Apple Pay checkout is complete
+   */
+
+  public init(
+    withConfigurationSignal configureWithData: Signal<PostCampaignCheckoutData, Never>,
```
ifosli, line 37: Have you explored just giving the ApplePayViewModel the parameters it actually needs, instead of passing along the entire `PostCampaignCheckoutData`? I think just using the parameters it needs would both be clearer and it would hopefully allow us to just use it in the crowdfunding checkout.

ifosli (approved):
Exciting!! I'm so curious to see if this will work nicely for crowdfunding checkouts as well! I left some comments including one that I think is blocking (an error that I think is getting dropped now), but I'm still marking this as approved so that you don't have to wait another 24 hours for me to review it again! If you make significant changes and want me to take another look, feel free to re-request the review.

PR #2089, opened 2024-06-26
Cleanup Risk Messaging Feature

scottkicks:
# 📲 What

Removes the unused Optimizely Risk Messaging Experiment Feature. 

# 🤔 Why

This extra, unused screen during checkout reminded people that backings weren't guaranteed and should have been removed when Optimizely was deprecated.

# 🛠 How

- Removes `RiskMessagingViewController/ViewModel`
- Updates`PledgeViewModel` checkout code that references `isNativeRiskMessagingControlEnabled` 
   - This boolean had been defaulting to true for some time
   -  Cleanup touches ApplePay and regular pledge code
- Removes `isNativeRiskMessagingControlEnabled` from SharedFunctions
- Removes `func trackPledgeConfirmButtonClicked()` from our analytics events because it was only used in the Risk Messaging flow.

# 👀 See

There should be no visible changes to the app.

# ✅ Acceptance criteria

- [x] Crowdfunding pledge flow experiences no changes for regular or ApplePay pledging
   - I've tested many projects in as many pledge scenarios as possible with multiple success/failure stripe cards. It will still be important to focus on testing this in our next release QA to make doubly sure I haven't broken something unexpected.
- [x] Updating Pledge via change payment and edit reward still works as expected


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -405,22 +403,25 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
       .ignoreValues()
 
-    // If the Optimizely risk messaging experiment is set to the control AND the Pay With Apple button is tapped
-    // Or if the Optimizely risk messaging experiment is set to the variant, and we are changing the payment method to Pay With Apple
-    // Or if the Optimizely risk messaging experiment is set to the variant and it is dismissed, this emits
-    let applePayButtonTappedOrRiskMessagingModalDismissed = Signal.merge(
-      self.applePayButtonTappedSignal.filter(isNativeRiskMessagingControlEnabled),
-      applePayButtonTappedAndIsChangingPaymentMethod,
-      self.riskMessagingViewControllerDismissedProperty.signal.skipNil().filter(isTrue).ignoreValues()
+    let applePayButtonTappedAndOrIsChangingPaymentMethod = Signal.merge(
```
ifosli, line 412: I think this entire signal should just be `self.applePayButtonTappedSignal` instead (since we're always in the control of the experiment now). The way it's currently written, if the apple pay button is tapped in the change pledge context, the signal looks like it'll emit twice.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -405,22 +403,25 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs, Pledge
...
+      self.applePayButtonTappedSignal,
+      changingApplePayPaymentMethod
     )
 
+    let goToApplePayPaymentAuthorization = pledgeAmountIsValid
+      .takeWhen(applePayButtonTappedAndOrIsChangingPaymentMethod)
```
ifosli, line 418: Specifically, I think we should just use `.takeWhen(self.applePayButtonTappedSignal)` here (and for the `showApplePayAlert` below

ifosli (changes requested):
Mostly looks good but I'm concerned that the behavior has changed a little (see comments for details). Will you make sure you also test this in the change pledge context if you haven't already? (Ie pledge with a fake card and go try to change your payment method to apple pay? Requires finding a project in staging that has real time left though or testing in prod)

Library/ViewModels/PledgeViewModel.swift
scottkicks: I hear you. There are XCTAssertEquals assertions in the tests for this that compare arrays, so if it was emitting twice, I would think those would catch it. 

I think you're suggestion reads more clearly anyway tho. I'll add this to the next commit.

scottkicks:
Thanks @ifosli! I've thoroughly tested changing payment and the rest of the manage pledge flow on top of pledging. I like your suggestions and will get those in today and re-test 👍

PR #2090, opened 2024-06-26
[MBL-1422] Handle errors without always dismissing

ifosli:
# 📲 What

Don't dismiss the post campaign VC if there's an error caused by trying to add a new card.

Note: There are other categories of errors that also don't need to dismiss the VC (for example, if a user fails to authenticate a 3DS card). However, because we're currently planning to combine our checkout flows and get rid of the "confirm details" page, I didn't think it was worth spending a lot of extra time on this now. 

# 🤔 Why

Needing to restart checkout in these cases is a little annoying.

# 🛠 How

I refactored the messageBannerView to optionally be in charge of dismissing the vc. That way, we can decide if the error should lead to dismissing the VC when a specific error is displayed to the user.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1422)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro Max - 2024-06-26 at 11 42 14 | Simulator Screen Recording - iPhone 15 Pro Max - 2024-06-26 at 11 38 05 |

# ✅ Acceptance criteria

- [x] Errors while adding a new card don't restart checkout
- [x] All other errors during late pledge will still restart checkout

### DIFF
--- a/Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
+++ b/Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
@@ -29,7 +29,7 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
     case bannerOnly
     // Banner can be dismissed by the user and will be dismissed automatically.
     // Once the banner is dismissed, the presenting view controller is popped.
-    case bannerAndParentVC
+    case bannerAndViewController
   }
 
   private var dismissType: DismissType = .bannerOnly
@@ -182,7 +182,7 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
             )
           }
         } else {
-          if self?.dismissType == .bannerAndParentVC {
+          if self?.dismissType == .bannerAndViewController {
             self?.navigationController?.popViewController(animated: true)
             return
           }
--- a/Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
+++ b/Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
@@ -10,6 +10,7 @@ public protocol MessageBannerViewControllerPresenting {
 }
 
 public protocol MessageBannerViewControllerDelegate: AnyObject {
+  // Called when banner view hides if dismissType is `.bannerOnly`.
   func messageBannerViewDidHide(type: MessageBannerType)
 }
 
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -19,7 +20,19 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
...
+    case persist
+    // Banner can be dismissed by the user and will be dismissed automatically.
+    // Once the banner is dismissed, the delegate method `messageBannerViewDidHide` is called.
+    case bannerOnly
+    // Banner can be dismissed by the user and will be dismissed automatically.
+    // Once the banner is dismissed, the presenting view controller is popped.
+    case bannerAndParentVC
+  }
+
+  private var dismissType: DismissType = .bannerOnly
```
ifosli, line 35: I didn't add any new tests around dismissType - I don't think our code is structured to make it easy to capture when view controllers dismiss in tests (please correct me if I'm wrong about that!). Alternatively, I could move all the dismiss logic into the view controller, but that would be a bigger refactor than I was hoping for. Open to suggestions here!

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
```diff
@@ -19,7 +20,19 @@ public final class MessageBannerViewController: UIViewController, NibLoading {
...
+
+  public enum DismissType {
+    // Banner cannot be dismissed and will not be dismissed automatically.
+    case persist
+    // Banner can be dismissed by the user and will be dismissed automatically.
+    // Once the banner is dismissed, the delegate method `messageBannerViewDidHide` is called.
+    case bannerOnly
+    // Banner can be dismissed by the user and will be dismissed automatically.
+    // Once the banner is dismissed, the presenting view controller is popped.
+    case bannerAndParentVC
```
scottkicks, line 32: minor nit: maybe `bannerToParentVC` is a clearer name for this case?

scottkicks (approved):
LGTM 

Are we confident that we don't need to create a new checkout from Confirm Details after this error. I'm like 90% that's true, but I tend to get paranoid in this flow 😅

ifosli:
> LGTM
> 
> Are we confident that we don't need to create a new checkout from Confirm Details after this error. I'm like 90% that's true, but I tend to get paranoid in this flow 😅

I double-checked the code and we don't even give stripe the checkout id when we're adding a new card (we give the project and late pledge vs crowdfunding pledge context), so there's no reason to go recreate the checkout id if adding the card goes wrong!

Kickstarter-iOS/Features/MessageBanner/Controller/MessageBannerViewController.swift
ifosli: I've renamed this to just `bannerAndViewController` - I think that we were thinking of the "parent vc" differently (in my mind it's the view controller that's getting dismissed, since it's the parent of the message banner). I think the new name is clearer for both of us!

PR #2092, opened 2024-07-08
[MBL-1457] Create empty state view for PPO

ifosli:
# 📲 What

Add an empty state view to the pledged projects overview.

Note: I didn't add a view model for the empty state view since it's so simple that I don't feel like it would add value. If anyone disagrees, however, let me know!

Note: This only kind of supports dynamic type: if the trait collection changes after the view has rendered, it doesn't re-render. The rest of our app is already pretty inconsistent about behavior when the trait collection changes, so I don't think this is worse than the current app experience. We may want to revisit this if/when we re-do the design system, though.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1457), [figma](https://www.figma.com/design/r0WKSECmMTCJTSKQskt2SQ/Backed-Projects-Dashboard?node-id=722-10751&t=sskwwFxurGDam09Y-0)

| Screen recording | Large text |
| -- | -- |
| Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-08 at 06 50 28 | Simulator Screen Recording - iPhone SE (3rd generation) - 2024-07-08 at 07 20 51 |

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Empty state view looks good
- [x] Empty state button works

(we're not testing that the empty state view shows up when it should yet)

# ⏰ TODO

- [x] Test voiceover
- Pull button styling out into a helper as part of https://kickstarter.atlassian.net/browse/MBL-1452 (not planned for this pr)


scottkicks (approved):
I agree that a View Model for this view doesn't add enough value to warrant one. There also aren't any @State objects to manage either.

LGTM 👍  Yay SwiftUI!

PR #2094, opened 2024-07-10
[MBL-1500] Update and use banners for ppo

ifosli:
# 📲 What

Update the swiftUI banners and use a sample one in the PPO view.

Note: I've added a test helper for snapshots of a swiftui view instead of our current view controllers. This doesn't match the format of our current tests (with a lot of different device sizes), but we could easily extend this to add a default iPad frame and a default landscape frame as well.

I do think this way of testing swiftui might make it harder to test different Environments (ex different font sizes or color schemes) but I also don't think it's worth investing a lot of time into that until we start actually working on dark mode. I'm open to suggestions, though!

Also an intentional difference from the way we often test: I'm hardcoding strings here instead of using known strings and testing different languages. The goal of these tests isn't to test our translations, so having a short string and a long string for each layout should be sufficient and makes sure we don't have extraneous snapshots. If I've missed anything/there are other reasons to use translatable strings and test all/multiple languages, let me know!

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1500)

Also see snapshots!

# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Banner is readable/useable and matches designs/our UIKit banners
- [x] Voiceover switches focus to the banner when it appears


Kickstarter-iOS/Features/MessageBanner/Views/MessageBannerView.swift
```diff
@@ -3,24 +3,31 @@ import SwiftUI
...
+          .frame(
+            maxWidth: .infinity,
+            minHeight: 40,
+            alignment: self.alignmentFromTextAlignment(vm.messageTextAlignment)
+          )
+      }
+      .foregroundColor(vm.messageTextColor)
+      .padding(EdgeInsets(top: 16, leading: 9, bottom: 16, trailing: 9))
+      .background {
+        RoundedRectangle(cornerRadius: 6)
```
ifosli, line 24: This corner radius (and possibly the padding) matches the figma for PPO but is different from the UIKit banners. I'm going to confirm with Alison if we want to match the figma or if we want to be consistent with the app before I submit the pr.

scottkicks: What do you think about having some kind of MessageBannerViewLayout enum to avoid hardcoding the padding, cornerRadius, minHeight, and spacing values?

Kickstarter-iOS/Features/MessageBanner/Views/MessageBannerViewTests.swift
```diff
@@ -0,0 +1,52 @@
+@testable import Kickstarter_Framework
+// @testable import KsApi
```
scottkicks, line 2: looks like we can remove this import for now

Kickstarter-iOS/Features/MessageBanner/Views/__Snapshots__/MessageBannerViewTests/testBannerInfo_longString.1.png
scottkicks, line 1: Is the image icon supposed to show in these snapshots? Either the checkmark or "i" info icon?

scottkicks (approved):
LGTM 

Just a couple of suggestions. I like your point about keeping snapshots simple and testing only what's important in the given context!

Kickstarter-iOS/Features/MessageBanner/Views/__Snapshots__/MessageBannerViewTests/testBannerInfo_longString.1.png
ifosli: No, we're not using an info icon - that's why the info banners are centered instead of aligned to the left. I think it's a bit of a weird choice, but it matches what our other banners do so I'm pretty sure it's intentional.

Kickstarter-iOS/Features/MessageBanner/Views/MessageBannerViewTests.swift
ifosli: Good catch, thanks!

Kickstarter-iOS/Features/MessageBanner/Views/MessageBannerView.swift
ifosli: Done! I wanted to do a shared enum between the swiftUI and UIKit versions, but the UIKit version is primarily a storyboard, so I gave up on that pretty quickly... I've also changed some of the values (and re-recorded screenshots) after talking to Alison and determining that PPO styling should match current standards, not our Figma mocks.

PR #2095, opened 2024-07-10
[MBL-1599] Adds the Shipping Dropdown Selector to the Rewards Carousel.

scottkicks:
### Includes snapshot test updates hence all of the files changed

# 📲 What

Adds the Shipping Dropdown Selector to the Rewards Carousel.

The `PledgeShippingLocationViewController` currently works by passing in the selected reward and fetching the `ShippingRule` associated with it via our V1 endpoint `.fetchRewardShippingRules(project.id, reward.id)`

Adding this to Rewards means that we won't have a specific reward to fetch `ShippingRule`s for, only the Project directly, and instead need to do the following:
- Filter all of the project's rewards and check for a reward with an `unrestricted` shipping preference. If there is one, we’ll use it to fetch the shipping rules using the V1 fetch above. This will return ALL available locations.
- For the case where there **isn’t** an `unrestricted`  reward, we’ll call the query for every reward with a `restricted` shipping preference to capture all possible shipping locations.

_This isn't ideal performance or networking and I don't want to write bootstrap code to work with our V1 API when the end game is to move completely to GraphQL._

_I've made a ticket and discussed this with Santiago, who is looking into the lift for adding all possible shipping locations directly to the Project Graphql model. This would solve both of my concerns. We can update this piece later if we decide to._

# 🛠 How

- Updates `PledgeShippingLocationViewModel` using the new logic above
- Adds the shipping dropdown to `RewardsCollectionViewController` 
- Updates Snapshot Tests

**Up Next**
1. Filter the displayed projects based on location
2. Remove the dropdown from both checkout screens and the Confirm Details screen
3. Unify checkout screen UI

# 👀 See

Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-15 at 14 32 20

# 🏎 Performance

# ✅ Acceptance criteria
- [x] The shipping dropdown shows all available shipping locations for the current project's shippable rewards
- [x] If there are no shippable rewards, the dropdown doesn't show
- [x] Late pledge/crowdfunding checkouts still function as expected. They still show the standard shipping selection as well.


Library/ViewModels/PledgeShippingLocationViewModel.swift
```diff
@@ -52,17 +52,44 @@ public final class PledgeShippingLocationViewModel: PledgeShippingLocationViewMo
...
+      
+//    let shippingRulesEvent = Signal.zip(project, reward)
+//      .filter { _, reward in reward.shipping.enabled }
+//      .switchMap { project, reward -> SignalProducer<Signal<[ShippingRule], ErrorEnvelope>.Event, Never> in
+//        AppEnvironment.current.apiService.fetchRewardShippingRules(projectId: project.id, rewardId: reward.id)
+//          .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
+//          .map(ShippingRulesEnvelope.lens.shippingRules.view)
+//          .retry(upTo: 3)
+//          .materialize()
+//      }
```
scottkicks, line 68: This is the current app implementation that fetches the Shipping Rules for a single reward.

Library/ViewModels/PledgeShippingLocationViewModel.swift
```diff
@@ -52,17 +52,44 @@ public final class PledgeShippingLocationViewModel: PledgeShippingLocationViewMo
...
+          )
           .ksr_delay(AppEnvironment.current.apiDelayInterval, on: AppEnvironment.current.scheduler)
           .map(ShippingRulesEnvelope.lens.shippingRules.view)
           .retry(upTo: 3)
           .materialize()
+
+          queryResult = queryResult?.merge(with: fetchResult) ?? fetchResult
+        }
+
+        return queryResult ?? SignalProducer(value: .value([]))
```
scottkicks, line 92: This solution of grabbing the IDs of the rewards that will help us determine what Shipping Rules to get, then looping through and calling the V1 endpoint for each one makes sense in theory, but I'm wondering if this is the best RxSwift pattern.

I would definitely love some help refactoring this if possible.

scottkicks: I need a way to capture each returned SignalProducer from each `fetchRewardShippingRules` call and then essentially combine them into one resulting `Signal<[ShippingRule], ErrorEnvelope>` hence the `.merge(with:)` on line: 89.

I'm not sure that this is very RxSwifty. It requires `queryResult` to be `nil` which  _might_ be ok since there shouldn't be a case where the project has no rewards. If the `rewardIDsToQuery` is empty, it would likely break things.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

scottkicks:
I'll be merging this once tests pass since all other iOS devs are OOO for the next 2 weeks

PR #2096, opened 2024-07-16
[MBL-1616] Send backingId to CreatePaymentIntent for late pledges

scottkicks:
# 📲 What

Exposes `backingId` in our `CreateCheckout` GraphQL Mutation Response in order to pass it to `CreatePaymentIntent`.

# 🤔 Why

As a backer, if my late pledge checkout errors, the payment should be reversed or the checkout rescued asynchronously (handled by Rosie (backend) if `backingId` is present)

# 🛠 How

- Expose `backingId` in `CreateCheckout.graphql`
- Pass newly exposed `backingId` to `CreatePaymentIntent` GraphQl call

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] backingId is successfully passed along to the backend in the CreatePaymentIntent call
- [x] Late pledge checkout flow works as normal. (creating new and using existing payment methods specifically)


Kickstarter.xcodeproj/project.pbxproj
scottkicks, line 1: Our apollo script wasn't pulling in the latest GraphAPI changes at first. Re-creating our `CreateCheckout.graphql` file seemed to clear whatever cache was holding that up.

It's the almost the exact same file that we currently have, other than the new object that exposes the needed backing id

KsApi/GraphAPI.swift
scottkicks, line 1: Auto updated file based on the .graphql changes I've made. This file is updated via our `bin/apollo-schema-download.sh` script

PR #2103, opened 2024-07-24
[MBL-1599] Add Shipping Location Dropdown to Rewards Carousel

scottkicks:
# 📲 What

Adds the Shipping Dropdown Selector to the Rewards Carousel.
Gated behind **No Shipping At Checkout** feature flag.

# 🛠 How

The strategy here is to create duplicate classes with updates specific to this feature. In this flow, gating behind a feature flag is needed to make it more straightforward and still an effective rollout strategy. The alternative would be a tangled mess because of the refactoring required to remove shipping costs from checkout.

The new "duplicate" classes for this piece are: `RewardsWithShippingCollectionViewController` `RewardsWithShippingCollectionViewModel` and `RewardsWithShippingCollectionViewHeaderView`

### Implementation Details
- Updates `PledgeShippingLocationViewModel` using the new logic above
- Adds the shipping dropdown to a new `RewardsWithShippingCollectionViewController` with associated `RewardsWithShippingCollectionViewModel` and `RewardsWithShippingCollectionViewHeaderView`
- Updates Snapshot Tests

The `PledgeShippingLocationViewController` currently works by passing in the selected reward and fetching the `ShippingRule` associated with it via our V1 endpoint `.fetchRewardShippingRules(project.id, reward.id)`

Adding the shipping dropdown to Rewards means we won't have a specific reward for fetching `ShippingRule`s. 

Instead, need to do the following when the feature flag is enabled:
- Filter all of the project's rewards and check for a reward with an `unrestricted` shipping preference. If there is one, we’ll use it to fetch the shipping rules using the V1 fetch above. This will return ALL available locations.
- For the case where there **isn’t** an `unrestricted`  reward, we’ll call the query for every reward with a `restricted` shipping preference to capture all possible shipping locations.

_This isn't ideal performance or networking and I don't want to write bootstrap code to work with our V1 API when the end game is to move completely to GraphQL._

_I've made a ticket and discussed this with Santiago, who is looking into the lift for adding all possible shipping locations directly to the Project GraphQL model. This would solve both of my concerns. We can update this piece later if we decide to._

# 👀 See

Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-24 at 11 14 11

# 🏎 Performance

# ✅ Acceptance criteria
- [x] The shipping dropdown shows all available shipping locations for the current project's shippable rewards
- [x] If there are no shippable rewards, the dropdown doesn't show
- [x] Late pledge/crowdfunding checkouts still function as expected. They still show the standard shipping selection as well.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #2110, opened 2024-07-31
Swift UI styling

ifosli:
# 📲 What

Create and use swiftui button styles to match our UIKit styling for buttons. Note that this is done with `SwiftUI.ButtonStyles`, not using view modifiers. This gives us access to the `isPressed` field, which looks equivalent to the UIKit highlighted state. 

I've only copied over the styles that I think we need. If I've missed anything, either for a different kind of button or something from BaseStyles, we can add it when we need it. (Or let me know, and I can add it here.)

Note that tests don't cover the `isPressed` state - I couldn't find a way to do this. If it's possible and anyone knows how to do it, let me know, and I'll add a "highlighted" state snapshot test.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1452)

Button styles can be seen in the snapshot tests.

# ✅ Acceptance criteria

- [ ] Button styles match our prelude styles and are easy to use.


Kickstarter-iOS/SharedViews/SwiftUIHelpers/ButtonStyles.swift
```diff
@@ -0,0 +1,74 @@
...
+// See `Library/Styles/ButtonStyles.swift` for the Prelude version of these styles.
+// These files should be kept in sync.
+
+struct GreenButtonStyle: SwiftUI.ButtonStyle {
+  @SwiftUI.Environment(\.isEnabled) private var isEnabled
+  
+  func makeBody(configuration: Configuration) -> some View {
+    configuration.label
+      .frame(maxWidth: .infinity)
+      .font(Font(UIFont.boldSystemFont(ofSize: 16)))
```
stevestreza-ksr, line 13: Should these be `ksr_body` or one of the other defined fonts? (Same question for the other button styles in this file)

scottkicks (approved):
LGTM 👍 

I think adding the other button styles as we need them makes sense!

PR #2114, opened 2024-08-06
[MBL-1601] Filter Rewards By Selected Shipping Location

scottkicks:
# 📲 What

Filters the displayed rewards based on the selected shipping location.

# 🤔 Why

We only want users to select rewards that they can actually get.

# 🛠 How

The shipping dropdown is already on this screen. We need to filter based on the currently selected location.
The types of rewards are:
- No-reward/Digital only
- Local Pickup only
- Ships anywhere in the world: 
- Ships to selected countries

Right now, we pull in a project's rewards using this GraphQL Query `FetchProjectRewardsById`.
This returns all rewards for a project and has an option to `includeShippingRules`.
Flipping this to `true` allows us to compare individual reward shipping info to the selected location and filter accordingly.

# 👀 See

I created a test project, A Really Cool Proj, that has the following reward types.
1. _No Reward:_ Should always show
2. _Digital:_ Should always show
3. _Local:_ Only show if the selected country matches the pickup country (United States)
4. _Ships to select countries:_ Shows if the selected country is United States, Aruba, or Antarctica
5. _Ships anywhere:_ Should always show

Simulator Screen Recording - iPhone 15 Pro Max - 2024-08-06 at 15 05 46

# ✅ Acceptance criteria

- [x] When there are only digital rewards, only the no-reward and digital-only rewards are shown, and no shipping location dropdown is shown.
- [x] Rewards that ship anywhere are always shown if enabled.
- [x] When the selected country matches the local pickup reward's country, or a country in a restricted rewards shipping location list, those rewards + the no reward/digital rewards are shown


Library/ViewModels/WithShippingRewardsCollectionViewModel.swift
```diff
@@ -429,5 +438,56 @@ private func backingAndShippingTotal(for project: Project, and reward: Reward) -
...
+  _ reward: Reward
+) -> Bool {
+  guard let selectedLocationId = locationId else { return false }
+
+  let shippingLocationIds: Set<Int> = Set(
+    reward.shippingRulesExpanded?.map(\.location).map(\.id) ?? []
+  )
+
+  return shippingLocationIds.contains(selectedLocationId)
+}
```
scottkicks, line 493: My current issue. This logic is taken from the Add-Ons screen that currently filters addons based on location, but I'm not sure how that really works because from what I can tell...

Rewards with restricted shipping don't have location or shipping rules associated with them; their values are `nil,` so I don't know how to compare them to the selected shipping location in the dropdown. Working on seeing what's available on the backend and if the team as any insights.

KsApi/Service.swift
```diff
@@ -549,7 +549,7 @@ public struct Service: ServiceType {
     let query = GraphAPI
       .FetchProjectRewardsByIdQuery(
         projectId: projectId,
-        includeShippingRules: false,
+        includeShippingRules: true,
```
scottkicks, line 552: Tested around the app to make sure this doesn't impact anything unexpectedly. All this does is include shipping rule/location data in each reward for a project. 

In theory, this might cause a slower project load time. I haven't seen a notable difference, but I will keep an eye on it.

Library/ViewModels/WithShippingRewardsCollectionViewModel.swift
```diff
@@ -429,5 +438,55 @@ private func backingAndShippingTotal(for project: Project, and reward: Reward) -
...
+        }
+      }
+      // If restricted shipping, compare against selected shipping location.
+    } else if isRestrictedShippingReward {
+      shouldDisplayReward = rewardShipsTo(selectedLocation: shippingRule?.location.id, reward)
+    }
+
+    return shouldDisplayReward
+  }
+}
```
scottkicks, line 476: This logic is mostly from the existing AddOns screen filtering logic with minor adjustments. 

If there are any suggestions for simplifying or making this easier to understand, please suggest them!

ifosli (approved):

PR #2115, opened 2024-08-06
[MBL-1462] Clean up OAuth flag

ifosli:
# 📲 What

Clean up oauth flag (defaulting to oauth behavior) and delete dead code.

I filed https://kickstarter.atlassian.net/browse/MBL-1656 as a followup so that we can make sure the `LoginViewController` and related classes cannot and should not be cleaned up as well. (They're currently accessible from the facebook login flow, but I'm not sure if that's intentional or if it could be an oversight.)

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1462)

# ✅ Acceptance criteria

- [x] Tests pass and app runs without visible changes (apart from the remote config controls)


Kickstarter-iOS/SharedViews/SharedStoryboards/Login.storyboard
ifosli, line 1: Note: these changes are the automatic changes xcode made when I deleted the signup screen.

scottkicks (approved):
👍 Byeeee!

PR #2116, opened 2024-08-07
Expose EstimatedMin and EstimatedMax in ShippingRule Fragment

scottkicks:
# 📲 What

Exposes EstimatedMin and EstimatedMax in ShippingRule Fragment

# 🤔 Why

We're going to display the estimated shipping costs in our checkout flows now that shipping won't be charged at the time of the pledge.

See [the designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=19-852&t=0OnrI8sWDKkTxgma-4)

# 🛠 How

- Add properties to our ShippingRuleFragment
- Update the model and wrapper 
- Update tests and Prelude lens

# 👀 See

no changes

# ✅ Acceptance criteria

- [x] Everything works as normal!

PR #2117, opened 2024-08-12
[MBL-1605] Display Estimated Shipping on Rewards and Add-Ons

scottkicks:
# 📲 What

**I recommend reviewing this commit by commit**

Displays estimated shipping range, if there are any set, on reward card and add-on card views.

# 🤔 Why

[See jira ticket](https://kickstarter.atlassian.net/browse/MBL-1605)

# 🛠 How

Updates `RewardCardView` and `RewardAddOnView`
Uses the same styling as the other sections of the cards.

`ShippingRule` structs contain the [newly exposed estimatedMin and estimatedMax values](https://github.com/kickstarter/ios-oss/pull/2116) that are set in the project config.
Since a reward can have multiple ShippingRules, we need to pass the `ShippingRule` that is selected in the location dropdown to the `RewardCardView` so that we can determine which `ShippingRule` info to display.

This only applies to rewards cards because we already pass in the relevant ShippingRule to the AddOns screen.

From there it's just getting the new UI elements in place and making sure this is only shown when the feature flag is on and when there is actually estimated shipping values to show.
   - Estimated shipping isn't required by creators so if there is none, we don't show anything.

# 👀 See

Simulator Screen Recording - iPhone 15 Pro - 2024-08-12 at 14 05 21

# ✅ Acceptance criteria

- [x] Estimated shipping is shown on reward cards and addon cards when feature flag is on an values are set 

# ⏰ TODO

- We'll get the hardcoded strings translated and updated later before release.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

scottkicks:
FYI this has some extra commits that will be on main by now. Formatting that Ingerid fixed + exposing estimated shipping properties stuff that I did. These should be cleaned up once main is up to date.

Library/SharedFunctions.swift
```diff
@@ -660,3 +660,28 @@ public func isRewardDigital(_ reward: Reward?) -> Bool {
...
+        estimatedMin > 0 || estimatedMax > 0 else {
+    return ""
+  }
+
+  /// Drop digits after the decimal.
+  let formattedMin = String(format: "%.0f", estimatedMin)
+  let formattedMax = String(format: "%.0f", estimatedMax)
+
+  // TODO: Update string with translations
+  return "About $\(formattedMin)-$\(formattedMax)"
```
ifosli, line 686: I think this needs to check if `formattedMin == formattedMax` and just use ex `About $\(formattedMin)$` if they're equal. Will you fix and maybe add tests for this function specifically, especially if it's not already covered by other tests?

Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardView.swift
ifosli, line 1: Can you add snapshot tests to this where the estimated shipping shows? I think one where the min and max are different and a different one where they're equal should cover it, but feel free to cover other edge cases if I'm missing something!

ifosli (approved):
Looks pretty good but please make sure you handle the "estimated min and max are equal" case and add snapshot tests for this before submitting. Marking this as approved because I don't think I need to re-review it, but please lmk if you want me to take a look at any changes!

PR #2121, opened 2024-08-14
[MBL-1458 pt 2] Add address summary for PPO

stevestreza-ksr:
# 📲 What

This PR implements the address summary for the card view for pledged projects, as well as the relevant subviews. It adds a view model to cover handling actions like tapping buttons and sending the creator a message.

# 🛠 How

Views are implemented as SwiftUI views, broken up into a few pieces, including individual flags, a project details view, a creator summary view, an address summary view, and buttons. Where possible, views used existing styling functions but did not use Prelude.

Snapshot testing is already complete and will be included in the card view PR.

# 👀 See

# ♿️ Accessibility 

# 🏎 Performance

# ✅ Acceptance criteria

# ⏰ TODO

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
@@ -3,6 +3,7 @@ import SwiftUI
 
 struct PPOAddressSummary: View {
   let address: String
+  let leadingColumnWidth: CGFloat
 
   var body: some View {
     HStack(alignment: .firstTextBaseline) {
@@ -11,7 +12,7 @@ struct PPOAddressSummary: View {
         .lineLimit(nil)
         .font(Font(Constants.labelFont))
         .foregroundStyle(Color(Constants.labelColor))
-        .frame(width: Constants.labelWidth, alignment: .leading)
+        .frame(width: self.leadingColumnWidth, alignment: .leading)
         .multilineTextAlignment(.leading)
 
       Text(self.address)
@@ -25,7 +26,6 @@ struct PPOAddressSummary: View {
   }
 
   private enum Constants {
-    static let labelWidth: CGFloat = 85
     static let labelFont = UIFont.ksr_caption1().bolded
     static let labelColor = UIColor.ksr_black
     static let addressFont = UIFont.ksr_caption1()
@@ -35,12 +35,17 @@ struct PPOAddressSummary: View {
 
 #Preview {
   VStack(spacing: 28) {
-    PPOAddressSummary(address: """
-      Firsty Lasty
-      123 First Street, Apt #5678
-      Los Angeles, CA 90025-1234
-      United States
-    """)
+    GeometryReader(content: { geometry in
+      PPOAddressSummary(
+        address: """
+          Firsty Lasty
+          123 First Street, Apt #5678
+          Los Angeles, CA 90025-1234
+          United States
+        """,
+        leadingColumnWidth: geometry.size.width / 4
+      )
+    })
   }
   .padding(28)
 }
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
@@ -1,0 +1,46 @@
+import Foundation
+import SwiftUI
+
+struct PPOAddressSummary: View {
+  let address: String
+
+  var body: some View {
+    HStack(alignment: .firstTextBaseline) {
+      // TODO: Localize
+      Text("Shipping address")
+        .lineLimit(nil)
+        .font(Font(Constants.labelFont))
+        .foregroundStyle(Color(Constants.labelColor))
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
```diff
@@ -0,0 +1,46 @@
...
+    HStack(alignment: .firstTextBaseline) {
+      // TODO: Localize
+      Text("Shipping address")
+        .lineLimit(nil)
+        .font(Font(Constants.labelFont))
+        .foregroundStyle(Color(Constants.labelColor))
+        .frame(width: Constants.labelWidth, alignment: .leading)
+        .multilineTextAlignment(.leading)
+
+      Text(self.address)
```
ankushkushwaha, line 17:
```suggestion
      Text(address)
```

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
```diff
@@ -0,0 +1,46 @@
...
+import SwiftUI
+
+struct PPOAddressSummary: View {
+  let address: String
+
+  var body: some View {
+    HStack(alignment: .firstTextBaseline) {
+      // TODO: Localize
+      Text("Shipping address")
+        .lineLimit(nil)
```
ankushkushwaha, line 11: We do not need `.lineLimit(nil)`, as default `lineLimit` for text component is `nil`

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
```diff
@@ -0,0 +1,46 @@
...
+        .multilineTextAlignment(.leading)
+        .font(Font(Constants.addressFont))
+        .foregroundStyle(Color(Constants.addressColor))
+        .frame(maxWidth: .infinity, alignment: .leading)
+    }
+    .frame(maxWidth: .infinity)
+  }
+
+  private enum Constants {
```
ankushkushwaha, line 27: Is there any reason we are using enum here instead of struct?

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
ifosli, line 1: nit: if we're seeing a lot of similar styling across views, we could define view modifiers for them to reuse.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
```diff
@@ -0,0 +1,46 @@
...
+    static let labelWidth: CGFloat = 85
```
ifosli, line 28: What's the benefit of hardcoding the label width? I have a feeling we'd want it to be able to grow larger for larger font sizes (you're welcome to check with accessibility sizes in your simulator!) or in landscape/on iPad. Maybe calculate the width based on the screen size instead? (I think GeometryReader allows this but lmk if I'm wrong and I'd be happy to help brainstorm.)

Feel free to file a followup ticket (doesn't necessarily need to block launch) if you'd rather revisit layouts for large text later - we can follow up with UX.

ifosli (approved):
Looks pretty good! My main concern is that I think we'd want that first column width to depend on the screen size to make this look good on iPad. Please either look into that or file a followup ticket before submitting!

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
stevestreza-ksr: Our style rules currently require use of `self`.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
stevestreza-ksr: `enum`s without cases cannot be instantiated, so they are good for use as namespaces

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
stevestreza-ksr: Injecting this value now to calculate the result from the parent card.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
ankushkushwaha: Okay. Thanks for letting me know. Looks good then

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOAddressSummary.swift
ankushkushwaha: Okay, sounds good.

PR #2123, opened 2024-08-14
[MBL-1458 pt 4] Add project details for PPO

stevestreza-ksr:
# 📲 What

This PR implements the project detail summary view for the card view for pledged projects, as well as the relevant subviews. It adds a view model to cover handling actions like tapping buttons and sending the creator a message.

# 🛠 How

Views are implemented as SwiftUI views, broken up into a few pieces, including individual flags, a project details view, a creator summary view, an address summary view, and buttons. Where possible, views used existing styling functions but did not use Prelude.

Snapshot testing is already complete and will be included in the card view PR.

# 👀 See

# ♿️ Accessibility 

# 🏎 Performance

# ✅ Acceptance criteria

# ⏰ TODO

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
@@ -1,0 +1,64 @@
+import Foundation
+import Kingfisher
+import KsApi
+import SwiftUI
+
+struct PPOProjectDetails: View {
+  let imageUrl: URL?
+  let title: String?
+  let pledge: GraphAPI.MoneyFragment
+
+  var body: some View {
+    HStack {
+      KFImage(self.imageUrl)
+        .resizable()
+        .clipShape(Constants.imageShape)
+        .aspectRatio(Constants.imageAspectRatio, contentMode: .fit)
+        .frame(width: Constants.imageWidth)
+      Spacer()
+        .frame(width: Constants.spacing)
+      VStack {
+        if let title {
+          Text(title)
+            .font(Font(Constants.titleFont))
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .lineLimit(2)
+        }
+        // TODO: Localize
+        Text("\(self.pledge.symbol ?? "")\(self.pledge.amount ?? "") pledged")
+          .font(Font(Constants.subtitleFont))
+          .foregroundStyle(Color(Constants.subtitleTextColor))
+          .frame(maxWidth: .infinity, alignment: .leading)
+          .lineLimit(1)
+      }
+    }
+    .fixedSize(horizontal: false, vertical: true)
+    .frame(maxWidth: .infinity)
+  }
+
+  private enum Constants {
+    static let imageShape = RoundedRectangle(cornerRadius: 4)
+    static let imageAspectRatio: CGFloat = 16 / 9
+    static let imageWidth: CGFloat = 85
+    static let spacing: CGFloat = 8
+    static let titleFont = UIFont.ksr_caption1().bolded
+    static let subtitleFont = UIFont.ksr_footnote()
+    static let subtitleTextColor = UIColor.ksr_support_400
+  }
+}
+
+#Preview {
+  VStack {
+    PPOProjectDetails(
+      imageUrl: URL(string: "http:///")!,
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+    .fixedSize(horizontal: false, vertical: true)
+    .frame(maxWidth: .infinity)
+  }
+
+  private enum Constants {
+    static let imageShape = RoundedRectangle(cornerRadius: 4)
+    static let imageAspectRatio: CGFloat = 16 / 9
+    static let imageWidth: CGFloat = 85
```
ifosli, line 42: If we want this in sync with the address columns, we should definitely pull that out into a shared "styles"  file. I still think we might want both of these to be a calculation based on the screen width, but maybe that's a question for design, too.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -0,0 +1,64 @@
...
+        .frame(width: Constants.spacing)
+      VStack {
+        if let title {
+          Text(title)
+            .font(Font(Constants.titleFont))
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .lineLimit(2)
+        }
+        // TODO: Localize
+        Text("\(self.pledge.symbol ?? "")\(self.pledge.amount ?? "") pledged")
```
ifosli, line 28: nit: Should this also have an `if let` and only show if there is a pledge symbol and amount? " pledged" would look pretty weird, but I'm okay with it since it really shouldn't happen regardless.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -0,0 +1,64 @@
...
+        .aspectRatio(Constants.imageAspectRatio, contentMode: .fit)
+        .frame(width: Constants.imageWidth)
+      Spacer()
+        .frame(width: Constants.spacing)
+      VStack {
+        if let title {
+          Text(title)
+            .font(Font(Constants.titleFont))
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .lineLimit(2)
```
ifosli, line 25: nit: this should probably have a defined text color, too

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -0,0 +1,64 @@
...
+          .frame(maxWidth: .infinity, alignment: .leading)
+          .lineLimit(1)
+      }
+    }
+    .fixedSize(horizontal: false, vertical: true)
+    .frame(maxWidth: .infinity)
+  }
+
+  private enum Constants {
+    static let imageShape = RoundedRectangle(cornerRadius: 4)
```
ifosli, line 40: nit: prefer to use the `cornerRadius` in our `BaseStyles.swift` file instead of hardcoding the `4`.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -0,0 +1,64 @@
...
+    static let imageAspectRatio: CGFloat = 16 / 9
+    static let imageWidth: CGFloat = 85
+    static let spacing: CGFloat = 8
```
ifosli, line 43: nit: Not super important if it'll lead to a lot of extra work, but I think we should try to stick to using `Styles.grid()` and `Styles.gridHalf()` for spacing, just so that the app looks cohesive.

ifosli (approved):
A few nits but overall really nice! I do think it looks worth pulling out shared styles/constants into a "PPO styles" file, especially if we want the address label and the image to be aligned.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
stevestreza-ksr: Similar to the address summary PR, moved this to be an injected property which will calculate based on the card size

PR #2127, opened 2024-08-20
[MBL-1606] New Crowdfund Checkout UI

scottkicks:
# 📲 What

## commit by commit review preferred

Updates the UI according to the [new designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=19-852&t=wSSXdx5F7wOTdtCM-4)

This doesn't include the new estimated shipping view yet. I'll do that in a separate PR

# 🤔 Why

[Some background context on why the change is needed.](https://kickstarter.atlassian.net/browse/MBL-1606)

# 🛠 How

- Mostly code deletion and re-ordering views
The Pledge Summary Table also needed some updates. It now more closely matches the design of the Late Pledge Checkout screen. That being said I opted to reuse some of the PostCampaign components here since late pledge will be updated to have basically the same UI. 
- The new CTA view that holds the pledge buttons also needs to display the pledge amount. `ConfirmDetailsViewController` does this, so I created a new CTA container class to keep it behind our feature flag.

# 👀 See

[Before comparison here](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2939388143/Consolidate+Checkout+Flows#Crowdfunding-Updates)

Simulator Screen Recording - iPhone 15 Pro - 2024-08-14 at 11 19 17

# ✅ Acceptance criteria

- [x] Crowdfund checkout works as expected
- [x] New UI matches new designs
- [x] This is all behind the no shipping at checkout feature flag.

# ⏰ TODO
- [x] fix tests
- Add the new estimated shipping view in a separate PR
- Rename the PostCampaign views that are being used in both flows now to a more generic name. Maybe `Checkout...`


nativeksr:
## Danger has errored
[!] Invalid `Dangerfile` file: Error running SwiftFormat:
Error: [0mRunning SwiftFormat...
(lint mode - no files will be changed.)
error: Unexpected end of file at 705:1 in /Users/distiller/kickstarter/Library/SharedFunctions.swift.[0m. Updating the Danger gem might fix the issue. Your Danger version: 9.4.2, latest Danger version: 9.5.0

```
 #  from Dangerfile:8
 #  -------------------------------------------
 #  swiftformat.exclude = %w(Library/Strings.swift Library/Styles/Colors.swift)
 >  swiftformat.check_format
 #  -------------------------------------------
```

</p>

Library/ViewModels/NoShippingPledgeViewModelTests.swift
scottkicks, line 1: These are minor updates to existing tests since `configurePledgeViewCTAContainerView` now listens to the latest changes to the project and pledge total signals and emits an extra signal.

ifosli: Will you also add a test to make sure the reward summary gets constructed correctly? These changes look like they're deleting tests for old behavior but I didn't notice any new tests for the new behavior.

ifosli (changes requested):
We talked offline about doing further cleanup separately, so I'm okay with leaving in the bonus amount classes for now. This needs tests, though (covering the new classes and the new pledge flow - I'd like to see both snapshot and unit tests). When the code is this complicated and is actively introducing tech debt, it's especially valuable to add tests now to make sure behavior works and stays consistent.

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_HasAddOns_ShippingSelected.lang_de_device_phone4_7inch_LoggedOut.png
ifosli, line 1: nit: I don't think the table is supposed to have margins on the left and right

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_PledgeContext_FixPaymentMethod_ErroredCard.lang_en_device_phone4_7inch.png
ifosli, line 1: nit: I think this should be "Pledge without a reward" instead of "Bonus support"

Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPledgeViewControllerTests.swift
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  func testView_PledgeContext_UnavailableStoredCards() {
+    let response = UserEnvelope<GraphUser>(me: self.userWithCards)
+    let mockService = MockService(fetchGraphUserResult: .success(response))
+    let project = Project.template
+      |> \.availableCardTypes .~ [CreditCardType.discover.rawValue]
+
+    orthogonalCombos(Language.allLanguages, Device.allCases).forEach { language, device in
+      withEnvironment(apiService: mockService, currentUser: User.template, language: language) {
```
ifosli, line 36: Will you turn the no shipping flag on in the environment for these tests? I'm not seeing the headings in the snapshots and I'm guessing that's why. (Also, there are a lot of them so I'm totally fine with doing minimal tests in this file if you'd rather and flesh it out once we get rid of the other tests).

ifosli (approved):
A few weird things in the tests but overall looks good! I think it might also be valuable to add a snapshot where the "estimated shipping" shows - this might also require the feature flag to be on in the environment the test is running in.

PR #2128, opened 2024-08-20
[MBL-1606] Display Estimated Shipping Cost at Checkout (Crowdfund)

scottkicks:
# 📲 What

Adds a new SwiftUI to display estimated shipping cost

# 🤔 Why

[JIRA TICKET](https://kickstarter.atlassian.net/browse/MBL-1606)

# 🛠 How

- Created a new SwiftUI View `EstimatedShippingCheckoutView`
- Used the estimatedMn and estimatedMax properties in the selected Reward's ShippingRule to get the estimated shipping cost and calculate the conversion to the user's preferred currency.
- If no estimated shipping or if shipping is not enabled for the selected reward we don't show this new view.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro - 2024-08-20 at 13 06 45 | with-view |

# ✅ Acceptance criteria

- [x] New view matches [the new designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=105-8312&t=x8MhTSiTedcRIb7a-4)
- [x] If there is no estimated shipping or if shipping is not enabled for the selected reward, we don't show this new view.


Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
```diff
@@ -0,0 +1,42 @@
...
+    public static let verticalPadding = Styles.grid(2)
+    public static let horizontalPadding = Styles.grid(4)
+  }
+
+  // TODO: Update strings with translations [mbl-1667](https://kickstarter.atlassian.net/browse/MBL-1667)
+  var body: some View {
+    HStack(alignment: .top) {
+      VStack(alignment: .leading, spacing: Constants.defaultSpacing) {
+        Text("Estimated Shipping")
+          .font(Font(UIFont.ksr_subhead(size: 16)))
```
ankushkushwaha, line 20: I think it would be helpful to define typography and define enum sizes to avoid hardcoded numbers.

ankushkushwaha (approved):

Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
```diff
@@ -0,0 +1,42 @@
...
+        .foregroundStyle(Color(UIColor.ksr_support_400))
+      }
+
+      VStack(alignment: .trailing, spacing: Constants.defaultSpacing) {
+        Text(self.estimatedCost)
+          .font(Font(UIFont.ksr_subhead()))
+          .bold()
+        Text(self.aboutConversion)
+          .font(Font(UIFont.ksr_subhead(size: 12)))
+          .foregroundStyle(Color(UIColor.ksr_support_400))
```
stevestreza-ksr, line 35: Would be good to move these fonts and colors into Constants

Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
```diff
@@ -0,0 +1,42 @@
...
+
+  // TODO: Update strings with translations [mbl-1667](https://kickstarter.atlassian.net/browse/MBL-1667)
+  var body: some View {
+    HStack(alignment: .top) {
+      VStack(alignment: .leading, spacing: Constants.defaultSpacing) {
+        Text("Estimated Shipping")
+          .font(Font(UIFont.ksr_subhead(size: 16)))
+          .bold()
+        Text(
+          "This is meant to give you an idea of what shipping might cost. Once the creator is ready to fulfill your reward, you’ll return to pay shipping and applicable taxes."
```
stevestreza-ksr, line 23: Assuming this should be localized, a todo with a link to MBL-1667 would be good here so it doesn't get missed

Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -306,6 +308,22 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
       self.pledgeDisclaimerViewDidTapLearnMoreSignal.mapConst(.trust)
     )
 
+    self.configureEstimatedShippingView = Signal.combineLatest(project, baseReward, selectedShippingRule)
+      .map { project, reward, shippingRule in
+        guard let rule = shippingRule else { return ("", "") }
```
stevestreza-ksr, line 313: Does it make sense to make this Signal return an Optional and return a nil here, rather than a pair of empty strings? Answer might be no, just safety checking

Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
scottkicks: Yep there's a TODO comment at the top of this class

Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPledgeViewControllerTests.swift
```diff
@@ -170,10 +170,56 @@ final class NoShippingPledgeViewControllerTests: TestCase {
...
+      |> Reward.lens.shipping.enabled .~ true
+      |> Reward.lens.shippingRules .~ [shippingRule]
+      |> Reward.lens.id .~ 99
+    let mockConfigClient = MockRemoteConfigClient()
+    mockConfigClient.features = [
+      RemoteConfigFeature.noShippingAtCheckout.rawValue: true
+    ]
+
+    orthogonalCombos(Language.allLanguages, [Device.phone4_7inch, Device.pad], [nil, User.template])
+      .forEach { language, device, _ in
```
ifosli, line 187: Since we're not using the user in this test. can you delete it from the `forEach`?

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_HasAddOns.lang_de_device_phone4_7inch_LoggedOut.png
ifosli, line 1: There's a weird margin where the Add-ons cell gets cut off in the new screenshot. This is happening in the screencast, too. Will you take a look and see if you can fix it?

Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
```diff
@@ -0,0 +1,54 @@
...
+struct EstimatedShippingCheckoutView: View {
+  let estimatedCost: String
+  let aboutConversion: String
+
+  private enum Constants {
+    /// Spacing & Padding
+    public static let defaultSpacing = Styles.grid(1)
+    public static let verticalPadding = Styles.grid(2)
+    public static let horizontalPadding = Styles.grid(4)
+    public static let maxViewHeight = 125.0
```
ifosli, line 13: How does this max height work for accessibility large sizes? I'm concerned that the text will get truncated enough that it won't be usable.

Library/SharedFunctions.swift
```diff
@@ -683,11 +683,62 @@ public func estimatedShippingText(for reward: Reward, selectedShippingRule: Ship
...
+  /// Make sure the current reward has shipping rules and that one of them matches the selected shipping rule (from the locations dropdown).
+  guard let shippingRules = reward.shippingRules,
+        let selectedShippingRule = shippingRules
+        .first(where: { $0.location.country == selectedShippingRule.location.country })
+  else {
+    return ""
+  }
+
+  guard let estimatedMin = selectedShippingRule.estimatedMin?.amount.rounded(.towardZero),
+        let estimatedMax = selectedShippingRule.estimatedMax?.amount.rounded(.towardZero),
```
ifosli, line 709: Why are you rounding these? They're getting rounded by the format function anyways. Rounding first and then converting just makes the final result less accurate.

Library/ViewModels/RewardAddOnCardViewModel.swift
```diff
@@ -103,8 +103,11 @@ public final class RewardAddOnCardViewModel: RewardAddOnCardViewModelType, Rewar
       .map(itemsLabelAttributedText)
       .skipNil()
 
+    // TODO: Update string with translations [mbl-1667](https://kickstarter.atlassian.net/browse/MBL-1667)
     self.estimatedShippingLabelText = Signal.combineLatest(reward, shippingRule)
-      .map { reward, shippingRule in estimatedShippingText(for: reward, selectedShippingRule: shippingRule) }
+      .map { reward, shippingRule in
+        "About \(estimatedShippingText(for: reward, selectedShippingRule: shippingRule))"
```
ifosli, line 109: Do you know if ranges are represented by `min - max` in all the languages we translate into? Specifically, is that true for Japanese? If not, we should have two separate translated strings; one "About {value}" and a separate "About {min_value} - {max_value}".

ifosli (changes requested):
Looks pretty good! Just a few minor things plus that weird extra padding above the pledge CTA.

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_HasAddOns.lang_de_device_phone4_7inch_LoggedOut.png
scottkicks: That's because of the spacing between the CTA container and the scroll view. I'll look into dynamically updating the constraint based on whether the estimated shipping view is being shown. This one may end up being tech debt tho.

Kickstarter-iOS/Features/PledgeView/Views/EstimatedShippingCheckoutView.swift
scottkicks: Ah good call. Removed. Here's how it looks with large accessibility font sizes 
Simulator Screen Recording - iPhone 15 Pro - 2024-08-26 at 08 58 36

Library/ViewModels/RewardAddOnCardViewModel.swift
scottkicks: This is and interesting point. I'm reaching out to verify

scottkicks: If we need to update our strings, I should do that in a separate PR. I've reached out to design and the team so that the conversation can start ASAP.

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_HasAddOns.lang_de_device_phone4_7inch_LoggedOut.png
ifosli: Yeah, you're right - I agree this doesn't need to be launch blocking. Will you file a ticket for it, though? I definitely don't want us to lose track of the new tech debt we introduce.

ifosli (approved):
Looks good, thank you! Just file a ticket for that weird spacing issue, please!

PR #2129, opened 2024-08-21
[MBL-1604] Show bonus in add-ons

ifosli:
# 📲 What

Show bonus support selection on the add-ons screen. Note that the continue button is only enabled if the amount is valid. The confirm details screen uses the bonus support passed in as its initial selection.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1604)

| Invalid amount case | Navigating to/from confirm details |
| --- | --- |
| Screenshot 2024-08-21 at 10 57 53 AM | Simulator Screen Recording - iPhone 15 - 2024-08-21 at 13 08 18 |

# ✅ Acceptance criteria

- [x] Bonus support shows on the add-ons screen and is propagated to the next step of checkout. So far, it's only actually used in the confirm details screen.

# ⏰ TODO

-  I didn't test voiceover for this but since it's only moving known components, I'm planning to skip that for now and test it during QA instead.


Kickstarter-iOS/Features/RewardAddOnSelection/Controller/__Snapshots__/RewardAddOnSelectionNoShippingViewControllerTests/testView.lang_de_device_phone4inch.png
ifosli, line 1: I'm not sure why these snapshots don't translate correctly to other languages - the same is true for our current add-ons snapshots. Because it's not a new change, my vote is to leave them for now and file a low priority follow-up ticket to fix this later (at least once we confirm that the screen itself translates correctly in prod).

Kickstarter-iOS/Features/RewardAddOnSelection/Controller/RewardAddOnSelectionNoShippingViewControllerTests.swift
ifosli, line 1: This is mainly just copied from our current tests - take a look commit by commit to just see the changes

Library/ViewModels/PledgeAmountViewModel.swift
```diff
@@ -267,12 +267,10 @@ public final class PledgeAmountViewModel: PledgeAmountViewModelType,
 // MARK: - Functions
 
 private func initialPledgeAmount(
-  from project: Project,
+  from _: Project,
   currentAmount: Double,
   minValue: Double
 ) -> Double {
-  guard userIsBackingProject(project) else { return minValue }
-
```
scottkicks, line 275: Why don't we need this guard here anymore?

scottkicks (approved):
LGTM 👍 Just a couple of minor comments

Library/ViewModels/PledgeAmountViewModel.swift
ifosli: I don't think we ever needed this guard. It looks like it's here to protect against developer error more than anything. (If the user isn't backing the project, we always set bonus support to `0` to start.)

The reason I removed it is because the thing it's saying ("if the user isn't already backing the project, change the bonus support to start at the min value") is wrong in a world where the confirm details screen uses the bonus support set on the add-ons screen as its initial value. Keeping it would reset the bonus support we see on the confirm details page. Deleting this guard means that the confirm details page, while not actually providing value to the UI, isn't doing anything silly either.

I'm okay with reverting this change and keeping the guard if you'd rather! But note that if we do that, then getting rid of bonus support on the confirm details page (or getting rid of the entire confirm details page) is definitely launch blocking.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -39,11 +39,55 @@ public typealias PKPaymentData = (displayName: String, network: String, transact
...
   public let rewards: [Reward]
+  public let bonusSupport: Double?
   public let selectedShippingRule: ShippingRule?
   public let selectedQuantities: SelectedRewardQuantities
   public let selectedLocationId: Int?
   public let refTag: RefTag?
   public let context: PledgeViewContext
+
+  // Convenience initializer to allow `bonusSupport` to default to `nil`.
+  // TODO: Delete this when cleaning up the noShippingAtCheckout feature.
```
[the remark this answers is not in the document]
ifosli: Done, good call!

PR #2130, opened 2024-08-21
[MBL-1458 pt 5] Card view for pledged projects overview

stevestreza-ksr:
# 📲 What

This PR implements the card view for pledged projects, as well as the relevant subviews. It adds a view model to cover handling actions like tapping buttons and sending the creator a message. This builds on previous PRs to add the individual pieces.

# 🛠 How

The project card is built from the individual pieces added in previous PRs.

Each card view gets its own view model instance, which is a container that holds the relevant data and can forward messages from the card view to a caller via Combine.

Views were tested via snapshot testing of the card in various configurations listed in Figma. View model inputs and outputs have unit tests.

# 👀 See

# ♿️ Accessibility 

# 🏎 Performance

# ✅ Acceptance criteria

# ⏰ TODO


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
```diff
@@ -0,0 +1,177 @@
...
+final class PPOProjectCardViewModel: PPOProjectCardViewModelType {
+  internal private(set) var isUnread: Bool
+  internal private(set) var alerts: [Alert]
+  internal private(set) var imageURL: URL
+  internal private(set) var title: String
+  internal private(set) var pledge: GraphAPI.MoneyFragment
+  internal private(set) var creatorName: String
+  internal private(set) var address: String?
+  internal private(set) var actions: (Action, Action?)
+  internal private(set) var parentSize: CGSize
```
ifosli, line 27: Using vars here instead of subjects is inconsistent with how we usually do things in this codebase. I don't personally have a strong opinion about one or the other, so I like to go with whatever we usually do. However, if you think that using vars simplifies things or makes things better in a way that's more valuable than keeping the patterns we usually use, I'm fine with leaving this as it is.

ifosli (approved):
Overall I really like how clean this is. Just one thing that I think is worth thinking twice about that deviates from our typical patterns. (Also, I added me and Scott to the reviewers!)

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
stevestreza-ksr: I've updated these in 80b1694 to be Publisher-based, they don't actually need to change yet so the vars were premature, but they are reactive now, and if that changes later then we can replace that in the view model and the view should update without issue.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift:27:1 | warning: (trailingClosures) Use trailing closure syntax where applicable. |
| Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift:31:1 | warning: (trailingClosures) Use trailing closure syntax where applicable. |


</p>

PR #2139, opened 2024-08-29
PPO: View model GraphQL loading and pagination support

stevestreza-ksr:
# 📲 What

This PR adds support for loading, reloading, and fetching more data for the PPO view. (Integration forthcoming)

# 🤔 Why

As part of PPO, we need the view to support loading paginated data from GraphQL and converting that into view models for individual cards.

# 🛠 How

This PR does two things:

1. The Paginator class was updated to use an enum with associated values, rather than separate optional values. This prevents invalid states from being representable by the compiler.
2. The PPOViewModel was updated to connect a Paginator to the GraphQL endpoint, with methods to trigger first load, pull to refresh, and loading the next page.

# ✅ Acceptance criteria

- [x] Unit tests should pass


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/Paginator.swift
```diff
@@ -54,84 +118,75 @@ public class Paginator<Envelope, Value: Equatable, Cursor, SomeError: Error, Req
...
-
-        if self.values.count == 0 {
-          self.state = .empty
-        } else if cursor == nil || newValues.count == 0 {
-          self.state = .allLoaded
+        let nextCursor = self.cursorFromEnvelope(envelope)
+        var allValues = shouldClear ? [] : previousResults.values
+        allValues.append(contentsOf: newValues)
+
+        let results: Results = if allValues.count == 0 {
```
ifosli, line 142: nit: I haven't seen this this kind of syntax before and I find it kind of hard to read. Would you be opposed to rewriting this as a typical nested `else if` thing with early returns instead?

ifosli (approved):
Looks pretty good - I left a few minor comments and I'm a little confused by some of the code that got deleted from the PPOView. The "show banner" button was intended to be temporary but the ability to show a banner was not. If you have a different plan for how to show a banner, I'm okay with deleting that code, but then I think we should just delete the geometry reader entirely. Marking this as approved but please fix the banner stuff before submitting!

Library/Paginator.swift
stevestreza-ksr: Sure; this is [new syntax](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md) as of Swift 5.9.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
```diff
@@ -92,6 +93,19 @@ final class PPOProjectCardViewModel: PPOProjectCardViewModelType {
 
   typealias Action = PPOProjectCardAction
   typealias Alert = PPOProjectCardAlert
+
+  // Add this static function at the end of the class
```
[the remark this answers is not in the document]
stevestreza-ksr: Adding a comment; it's there because of the `actions` tuple, for some reason Xcode won't synthesize with it (even though it's doing so anyway for `lhs.actions == rhs.actions`).

PR #2140, opened 2024-08-29
[MBL-1684] Remove Redundant No Shipping At Checkout Feature Flag Checks

scottkicks:
# 📲 What

Removes Redundant `featureNoShippingAtCheckout` Checks

Also does a quick fix for an issue found in QA for the Stripe Sheet not showing on `NoShippingPledgeViewController`.

# 🤔 Why

The `ProjectPageVIewController` determines whether to start the "No Shipping At Checkout" flow or the current flow in Production. 

We have unnecessary `if featureNoShippingAtCheckout()` conditions in the "No Shipping At Checkout" View Controllers. 

Since they'll never be shown when the flag is disabled, I'm removing them. They should never have been here tbh. 
This all cleanup work. 

# 🛠 How

Goes through each of these VCs and removes unused code and unnecessary feature flag checks:
- WithShippingRewardsCollectionVC + RewardsCollectionVC
- RewardAddOnSelectionNoShippingVC + RewardAddOnSelectionVC

# ✅ Acceptance criteria

- [x] Both Checkout flows function as expected when `featureNoShippingAtCheckout` is **OFF**
- [x] Both Checkout flows function as expected when `featureNoShippingAtCheckout` is **ON**
- [x] All tests pass

### DIFF
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
@@ -159,6 +159,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
 
         assert(
           topViewController is PledgeViewController ||
+            topViewController is NoShippingPledgeViewController ||
             topViewController is PostCampaignCheckoutViewController,
           "PledgePaymentMethodsViewController is only intended to be presented as part of a pledge flow."
         )
--- a/Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
+++ b/Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
@@ -180,25 +180,15 @@ final class WithShippingRewardsCollectionViewController: UICollectionViewControl
     self.viewModel.outputs.goToAddOnSelection
       .observeForControllerAction()
       .observeValues { [weak self] data in
-        if featureNoShippingAtCheckout() {
-          self?.goToNoShippingAddOnSelection(data: data)
-        } else {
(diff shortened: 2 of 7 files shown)


Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PledgePaymentMethodsViewController.swift
```diff
@@ -159,6 +159,7 @@ final class PledgePaymentMethodsViewController: UIViewController {
 
         assert(
           topViewController is PledgeViewController ||
+            topViewController is NoShippingPledgeViewController ||
             topViewController is PostCampaignCheckoutViewController,
           "PledgePaymentMethodsViewController is only intended to be presented as part of a pledge flow."
```
scottkicks, line 164: We need to add this type-check here so that the Stripe payment sheet is presented on the `NoShippingPledgeViewController`

Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
```diff
@@ -180,25 +180,15 @@ final class WithShippingRewardsCollectionViewController: UICollectionViewControl
...
-        if featureNoShippingAtCheckout() {
-          self?.goToNoShippingAddOnSelection(data: data)
-        } else {
-          self?.goToAddOnSelection(data: data)
-        }
+        self?.goToNoShippingAddOnSelection(data: data)
       }
 
-    self.viewModel.outputs.goToPledge
+    self.viewModel.outputs.goToCustomizeYourReward
```
scottkicks, line 186: Updated the name of this output because this is called anytime a nonshippable reward is selected taking the user to the screens that only have the pledge amount or bonus amount selectors shown.

ifosli (approved):
Nice!

PR #2143, opened 2024-09-03
[MBL-1688] Estimated Shipping Range Should Include Add-Ons Estimates

scottkicks:
Also fixes https://kickstarter.atlassian.net/browse/MBL-1683 since I can't seem to reproduce it with these changes.

# 🤔 Why

Right now estimated shipping on the checkout screen only reflects estimates for the main reward. It should include add-on estimates if there are any.

So if the base reward has a range of `$2-$7` and the backer also selects an add-on with a range of `$1-$5`, the displayed estimated shipping on the checkout screen should be `$1-$7`.

# 🛠 How

- Pass in all of the selected rewards during pledge to the estimated shipping text and conversion text helpers in SharedFunctions.
- Include those in their calculations.

# 👀 See

Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-03 at 11 50 10

# ✅ Acceptance criteria

- [x] if the base reward has a range of $2 - $7 and the backer also selects an add-on with a range of $1-$5, the displayed estimated shipping on the checkout screen should be $1 - $7.
- [x] Ranges for individual rewards still show correctly on Rewards Cards and Add-On Cards


Library/SharedFunctions.swift
```diff
@@ -764,3 +748,47 @@ public func attributedCurrency(withProject project: Project, total: Double) -> N
...
+  from rewards: [Reward],
+  selectedShippingRule: ShippingRule
+) -> (Double, Double) {
+  var min: Double = 0
+  var max: Double = 0
+
+  rewards.forEach { reward in
+    guard reward.shipping.enabled, let shippingRules = reward.shippingRules else { return }
+
+    /// if the current reward's shipping is unrestricted then
```
ifosli, line 762: Delete this comment?

Library/SharedFunctions.swift
```diff
@@ -764,3 +748,47 @@ public func attributedCurrency(withProject project: Project, total: Double) -> N
...
+
+    var shippingRule: ShippingRule?
+
+    /// If the reward's shipping prefernce is Anywhere in the world, us it's first and only shipping rule.
```
ifosli, line 766: nit: "use its first" instead of "us it's first"

Library/SharedFunctions.swift
```diff
@@ -764,3 +748,47 @@ public func attributedCurrency(withProject project: Project, total: Double) -> N
...
+    /// Else use the rule that  matches the selected shipping rule (from the locations dropdown).
+    shippingRule = reward.shipping.preference == .unrestricted
+      ? shippingRules.first
+      : shippingRules
+      .first(where: { $0.location.country == selectedShippingRule.location.country })
+
+    guard let shipping = shippingRule else { return }
+
+    /// Verify there are estimated amounts greater than 0
```
ifosli, line 775: Can you confirm that min estimated shipping needs to be greater than zero? Ie that it's not possible for creators to ex say shipping is 0-1 dollars? I don't think it's very likely that a creator would charge shipping for something like a postcard or letter during pledge redemption, but I think it could be possible

Library/SharedFunctions.swift
```diff
@@ -764,3 +748,47 @@ public func attributedCurrency(withProject project: Project, total: Double) -> N
...
+
+    /// Only update min if the current rewards esitmatedMin is smaller.
+    if min == 0 || estimatedMin < min {
+      min = estimatedMin
+    }
+
+    /// Only update max if the current rewards esitmatedMax is bigger.
+    if max == 0 || estimatedMax > max {
+      max = estimatedMax
+    }
```
ifosli, line 790: I think we should be adding the maxes together to get the final estimated max, not just taking the biggest one. Ex if I'm pledging to a reward with estimated shipping and I have an add-on with additional shipping, I expect to pay shipping for both. This is also true for min. It's worth confirming that web and android are calculating these the same way, though; we want the total estimated shipping to be consistent across platforms.

Library/SharedFunctionsTests.swift
```diff
@@ -671,4 +671,32 @@ final class SharedFunctionsTests: TestCase {
...
+
+    let project = Project.template
+
+    let estimatedShipping = estimatedShippingText(
+      for: [reward, addOn],
+      project: project,
+      selectedShippingRule: .template
+    )
+
+    XCTAssertEqual(estimatedShipping, "$1-$7")
```
ifosli, line 700: I like how clear this test is about expected behavior! Re: my comment above, I would expect this to be "$3-$12" instead, so please confirm with our other clients what this should say and update the test accordingly if needed.

ifosli (changes requested):
A couple minor nits and I'm concerned this might not be expected behavior. Please confirm across clients a) if minimum estimated shipping is guaranteed to be more than zero and b) how to combine estimated shipping for multiple rewards/add-ons. Or let me know if you want me to help take that on; I'd be happy to help!

Library/SharedFunctions.swift
scottkicks: great call! you're correct. we should be adding these together

Library/SharedFunctions.swift
scottkicks: I can confirm that you cannot have a minimum shipping value of 0.

Library/SharedFunctions.swift
```diff
@@ -759,11 +759,9 @@ private func estimatedMinMax(
   rewards.forEach { reward in
     guard reward.shipping.enabled, let shippingRules = reward.shippingRules else { return }
 
-    /// if the current reward's shipping is unrestricted then
-
     var shippingRule: ShippingRule?
 
-    /// If the reward's shipping prefernce is Anywhere in the world, us it's first and only shipping rule.
+    /// If the reward's shipping prefernce is Anywhere in the world, use it's first and only shipping rule.
```
ifosli, line 764: nit: "its" instead of "it's"

Library/SharedFunctions.swift
```diff
@@ -779,15 +777,8 @@ private func estimatedMinMax(
...
-    if min == 0 || estimatedMin < min {
-      min = estimatedMin
-    }
-
-    /// Only update max if the current rewards esitmatedMax is bigger.
-    if max == 0 || estimatedMax > max {
-      max = estimatedMax
-    }
+    min = min == 0 ? estimatedMin : min + estimatedMin
+    max = max == 0 ? estimatedMax : max + estimatedMax
```
ifosli, line 781: Nit: Do we need to check for `0` here? Why not just `min += estimatedMin` and `max += estimatedMax`? In the case where ex `min == 0`, `min + estimatedMin` and `estimatedMin` have the same value. I could be missing something, but I find `min += estimatedMin` to be more common/predicable and thus quicker to understand.

Library/SharedFunctions.swift
ifosli: Great, thanks!

ifosli (approved):
A few nits still but overall looks great!

PR #2144, opened 2024-09-03
[MBL-1680] Shipping Info Missing From Rewards Summary

scottkicks:
… if the shipping total is 0

# 📲 What

For some projects (at least one late pledge project on the confirm details screen), shipping information is missing from the pledge summary table. The total is still correct (includes the shipping).

# 🤔 Why

For projects that have shipping and have been active already, and since the pledge total includes this amount, we need to show shipping info in the summary tables. 

# 🛠 How

When creating the rewards summary table we have a condition that, If the flag is off, we hide shipping info. 

We should update this condition to only hide shipping for projects that don’t have shipping data _or_ if shipping is 0 as it will be for new projects that are created during this estimated shipping rollout.

# 👀 See

| Project the bug was found in (now shows shipping) | New projects with the flag on (continues to hide shipping) |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-03 at 12 43 38 | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-03 at 12 47 04 |

# ✅ Acceptance criteria

- [x] Shipping shows for projects that have project shipping data
- [x] Shipping is hidden for projects that don’t have shipping data _or_ if shipping is 0 as it will be for new projects that are created during this estimated shipping rollout


Library/ViewModels/PostCampaignPledgeRewardsSummaryViewModel.swift
```diff
@@ -189,10 +189,10 @@ private func items(
   }
   var items = [headerItem] + rewardItems
 
-  if featureNoShippingAtCheckout() == false {
-    // MARK: Shipping
+  if let shipping = data.shipping {
+    if shipping.total > 0 {
```
ifosli, line 193: nit: This is a little nicer to read if you do `if let shipping = data.shipping, shipping.total > 0 {` instead of nesting two `if`s.

ifosli (approved):
Just a little style thing that'd be nice to fix but otherwise looks good!

PR #2147, opened 2024-09-05
[MBL-1687] Fix continue button on confirm details page not working

ifosli:
# 📲 What

We accidentally introduced a bug where the late pledge flow can't be completed because the location id isn't correctly transformed to a string. This fixes that bug and the "no shipping" test that I thought might be related that ended up not being related after all.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1687)

| With flag off | With flag on |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Plus - 2024-09-05 at 12 59 12 | Simulator Screen Recording - iPhone 15 Plus - 2024-09-05 at 13 01 18 |

# ✅ Acceptance criteria

- [x] Checkout id is created successfully and user can navigate from confirm details to checkout screen. Tested with and without flag on and with and without shipping

PR #2148, opened 2024-09-05
[MBL-1687] Fix continue button on confirm details page not working #2147

ifosli:
# 📲 What

We accidentally introduced a bug where the late pledge flow can't be completed because the location id isn't correctly transformed to a string. This fixes that bug and the "no shipping" test that I thought might be related that ended up not being related after all.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1687)

| With flag off | With flag on |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Plus - 2024-09-05 at 12 59 12 | Simulator Screen Recording - iPhone 15 Plus - 2024-09-05 at 13 01 18 |

# ✅ Acceptance criteria

- [x] Checkout id is created successfully and user can navigate from confirm details to checkout screen. Tested with and without flag on and with and without shipping

### DIFF
--- a/Library/ViewModels/ConfirmDetailsViewModel.swift
+++ b/Library/ViewModels/ConfirmDetailsViewModel.swift
@@ -362,10 +362,14 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
             return [String](repeating: reward.graphID, count: count)
           }
 
+        let locationId = selectedShippingRule == nil
+          ? nil
+          : String(selectedShippingRule!.location.id)
+
         return CreateCheckoutInput(
           projectId: project.graphID,
           amount: String(format: "%.2f", pledgeTotal),
-          locationId: "\(selectedShippingRule?.location.id)",
+          locationId: locationId,
           rewardIds: rewardsIDs,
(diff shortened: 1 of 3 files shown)


Library/ViewModels/ConfirmDetailsViewModel.swift
```diff
@@ -362,10 +362,14 @@ public class ConfirmDetailsViewModel: ConfirmDetailsViewModelType, ConfirmDetail
             return [String](repeating: reward.graphID, count: count)
           }
 
+        let locationId = selectedShippingRule == nil
```
stevestreza-ksr, line 365: Suggestion: You could do `selectedShippingRule.flatMap { String($0.location.id) }` to make this one line and remove the `!` below (and do the same for `NoShippingConfirmDetailsViewModel`

stevestreza-ksr (approved):

scottkicks (approved):

Library/ViewModels/ConfirmDetailsViewModel.swift
ifosli: I like that! Thanks!

PR #2149, opened 2024-09-05
[MBL-1682] Late Pledge UI ScrollView Not Scrolling

scottkicks:
# 📲 What

The pledge summary hides behind the pledge cta on the late pledge checkout screen, making it difficult (though not impossible) to see what you’re paying.

# 🛠 How

The rootStackView's wasn't being constrained to it's parent, rootScrollView, properly

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-05 at 15 53 39 | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-05 at 15 57 57 |

# ✅ Acceptance criteria

- [x] Backers can view the pledge summary in its entirety.


ifosli (approved):
Nice! Just checking that we don't also want to/need to make these changes for the other version of this class?

scottkicks:
> Nice! Just checking that we don't also want to/need to make these changes for the other version of this class.

Nope. Tested and it looks good.
That other class is the same as what's currently on Prod.

PR #2150, opened 2024-09-06
[MBL-1712] No Reward Option Not Shown in Edit Reward Flow

scottkicks:
# 📲 What

Bug found in QA
- When pledging with a "no reward" selection this option wasn't being shown in the "Edit Reward" flow.
- It also wasn't selectable/editable

# 🤔 Why

- We should be showing the currently selected reward
- We also need them to be selectable so that backers can update their pledge/bonus amounts. Since those steppers were moved to the rewards screen we need to update this behavior.

# 🛠 How

Instead of only allowing rewards in the "Edit Rewards" flow to be editable if they have addons we should update this so that this condition is true if it has add ons _and/or_ the no shipping at checkout feature flag is on.

# 👀 See

Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-09-06 at 13 22 31

# ✅ Acceptance criteria

- [ ] Backers can select their current pledged reward and adjust their bonus/pledge amounts
- [ ] Backers can change rewards/add-ons etc. successfully
- [ ] When the feature flag is off, "edit reward" and "manage pledge" flows work as normal

### DIFF
--- a/Library/ViewModels/WithShippingRewardsCollectionViewModel.swift
+++ b/Library/ViewModels/WithShippingRewardsCollectionViewModel.swift
@@ -476,8 +476,8 @@ private func filteredRewardsByLocation(
     let isUnrestrictedShippingReward = reward.isUnRestrictedShippingPreference
     let isRestrictedShippingReward = reward.isRestrictedShippingPreference
 
-    // return all rewards that are digital or ship anywhere in the world.
-    if isRewardLocalOrDigital || isUnrestrictedShippingReward {
+    // return all rewards that are, no reward, digital, or ship anywhere in the world.
+    if rewards.first?.id == reward.id || isRewardLocalOrDigital || isUnrestrictedShippingReward {
       shouldDisplayReward = true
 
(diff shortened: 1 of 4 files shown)


Library/ViewModels/WithShippingRewardsCollectionViewModel.swift
```diff
@@ -476,8 +476,8 @@ private func filteredRewardsByLocation(
     let isUnrestrictedShippingReward = reward.isUnRestrictedShippingPreference
     let isRestrictedShippingReward = reward.isRestrictedShippingPreference
 
-    // return all rewards that are digital or ship anywhere in the world.
-    if isRewardLocalOrDigital || isUnrestrictedShippingReward {
+    // return all rewards that are, no reward, digital, or ship anywhere in the world.
```
ifosli, line 479: nit: delete the extra comma after `are`

ifosli (approved):
Nice! I'm surprised this automatically passes existing tests - it'd be nice to add a test to cover this case as part of this pr!

PR #2161, opened 2024-09-25
[MBL-1501] PPO view model navigation events

stevestreza-ksr:
# 📲 What
  
Implements most of the logic for navigation events in the PPO view model and the tab bar integration. 
  
  # 🛠 How
  
Each user interaction was added to the PPOViewModel separately, with separate subjects to represent each type of user interaction. This will let us handle navigation events and analytics separately. 

A `PPONavigationEvent` enum was added to represent different locations the user might navigate to. Then, the individual subjects are mapped to their respective navigation events, and those get merged into a navigationEvents publisher. 

Contextual information for routing locations (e.g. which project was backed) and handling of routing will be added once we connect this to a navigation service in MBL-1451).

Unit tests were added to cover these inputs and their navigation events.
  
  # ✅ Acceptance criteria
  
All tests should pass.

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
@@ -4,6 +4,7 @@ import Library
 import SwiftUI
 
 public class PPOContainerViewController: PagedContainerViewController<PPOContainerViewController.Page> {
+  private let viewModel = PPOContainerViewModel()
   private var ppoViewModel = PPOViewModel()
 
   public override func viewDidLoad() {
@@ -26,13 +27,19 @@ public class PPOContainerViewController: PagedContainerViewController<PPOContain
 
     let tabBarController = self.tabBarController as? RootTabBarViewController
 
-    self.ppoViewModel.$results.sink { [weak self] results in
-      let badge: TabBarBadge = results.total.flatMap { count in .count(count) } ?? .none
-      self?.setPagedViewControllers([
-        (.projectAlerts(badge), ppoViewController),
-        (.activityFeed(.dot), activitiesViewController)
-      ])
-    }.store(in: &self.subscriptions)
+    let projectAlertsBadge = self.ppoViewModel.$results
+      .map { results -> TabBarBadge in
+        results.total.flatMap { count in .count(count) } ?? .none
+      }
+
+    Publishers.CombineLatest(projectAlertsBadge, self.viewModel.activityBadge)
+      .sink { [weak self] projectAlerts, activity in
+        self?.setPagedViewControllers([
+          (.projectAlerts(projectAlerts), ppoViewController),
+          (.activityFeed(activity), activitiesViewController)
+        ])
+      }
+      .store(in: &self.subscriptions)
 
     ppoView.viewModel.navigationEvents.sink { nav in
       switch nav {
@@ -43,6 +50,11 @@ public class PPOContainerViewController: PagedContainerViewController<PPOContain
         break
       }
     }.store(in: &self.subscriptions)
+  }
+
+  public override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    self.viewModel.viewWillAppear()
   }
 
   public enum Page: TabBarPage {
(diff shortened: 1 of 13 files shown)


Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
```diff
@@ -20,6 +23,26 @@ public class PPOContainerViewController: PagedContainerViewController<PPOContain
       (.projectAlerts(.count(5)), ppoViewController),
       (.activityFeed(.dot), activitiesViewController)
     ])
+
+    let tabBarController = self.tabBarController as? RootTabBarViewController
+
+    self.ppoViewModel.$results.sink { [weak self] results in
+      let badge: TabBarBadge = results.values.count > 0 ? .count(results.values.count) : .none
```
ifosli, line 30: If a user has enough results that this paginates, won't the badge count be off? I think we should either use a `totalCount` which I think graphql gives us or just do a dot here for now and change it to a count once the backend updates to give us that count (not scoped for v1).

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
```diff
@@ -20,6 +23,26 @@ public class PPOContainerViewController: PagedContainerViewController<PPOContain
...
+      self?.setPagedViewControllers([
+        (.projectAlerts(badge), ppoViewController),
+        (.activityFeed(.dot), activitiesViewController)
```
ifosli, line 33: The activity feed count, however, is one that we can (and should) get. I think it's stored on the user object. If you can't find it, let me know, and I'll help look! (We should be using it for the tabbar count when the PPO flag is off.) If this gets complicated, it might make more sense to do the badges in a separate pr.

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
```diff
@@ -20,6 +23,26 @@ public class PPOContainerViewController: PagedContainerViewController<PPOContain
...
+      ])
+    }.store(in: &self.subscriptions)
+
+    ppoView.viewModel.navigationEvents.sink { nav in
+      switch nav {
+      case .backingPage:
+        tabBarController?.switchToProfile()
```
ifosli, line 40: Can we rename this case if it's supposed to refer to switching to the profile page? There's a separate thing called backing details or backing details page, which refers to the backing details within the survey webview. To me, `backingPage` sounds like it should be opening that, not the profile.

ifosli (changes requested):
I like the way navigation is set up! I don't think I've seen multiple views access the same view model in our code before, but it's clean and simple and I like it (apart from the one name that I think is confusing). I'm concerned about the alert dot/badges; if you want to either tag `TODO`s for those or split those changes into a separate pr, I'd be happy to approve this one.

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
stevestreza-ksr: Added support for the totalCount in https://github.com/kickstarter/ios-oss/pull/2161/commits/504282fedb6642d2565693aa0a1299428ee980cb

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOContainerViewController.swift
stevestreza-ksr: I added a new PPOContainerViewModel and added activity feed badge in https://github.com/kickstarter/ios-oss/pull/2161/commits/549d151211a739f3b273fd6d6217aeff25c30526 using similar patterns for handling user updates I'd seen elsewhere in the app.

Library/PaginatorTests.swift
```diff
@@ -18,6 +18,7 @@ struct ConcreteError: Error {}
 final class PaginatorTests: XCTestCase {
   let valuesFromEnvelope: (TestEnvelope) -> [Int] = { $0.values }
   let cursorFromEnvelope: (TestEnvelope) -> Int? = { $0.cursor }
+  let totalFromEnvelope: (TestEnvelope) -> Int? = { _ in 42 }
```
ifosli, line 21: This feels weird to me. Would it make more sense to add the total to the `TestEnvelope` itself? Or maybe just add a "Hardcoded because the total won't change during a fetch" comment if that's why it's hardcoded?

ifosli (approved):
Nice! I did add a few questions but nothing that needs to be blocking

Library/PaginatorTests.swift
stevestreza-ksr: Updated tests to not use a hardcoded value

PR #2167, opened 2024-10-03
[MBL-1460] PPO analytics events in PPOViewModel; refactor of PPO data loading types

stevestreza-ksr:
Sorry about the size, this one got away from me a bit. 😓 The biggest parts are just a refactor and some data plumbing.

Builds on the work in #2161. Should not merge before that PR.

# 📲 What

This PR adds telemetry events for various PPO actions that the user might take. 

# 🛠 How

These events are added in PPOViewModel, with corresponding code added to KSRAnalytics and the various Contexts that are needed.

Most of these actions require us sending information about the project that the card relates to. So the GraphQL query was updated to include the `ProjectFragment` needed to generate those properties.

That ProjectFragment needed to make its way from GraphQL, through various models and view models, to the PPOViewModel where it could be sent to analytics.

Because that also includes unit tests and the SwiftUI preview window, we needed some way to create a Project template and pass it into those objects. Previously I had built parsing into the PPOProjectCardViewModel, but we ultimately needed a type in KsApi, as that's where the existing template objects were.

*Sigh. Deep breath.*

That meant refactoring the parsing into a new struct type, PledgedProjectsOverviewCard, which handles parsing and data transfer from GraphQL. This type lives in KsApi, so I also added templates for the different card types that were present in the SwiftUI previews and in the unit tests. These templates can therefore use the Project template.

Then I refactored everything that parsed directly into PPOProjectCardViewModel and had those places return PledgedProjectOverviewCard types instead.

And once all of that was done… the Project type was available to pass to analytics events. And then the events got connected to their respective PassthroughSubject signals in PPOViewModel.

# ✅ Acceptance criteria

Tests should pass. No UI is hooked up yet.


ifosli (commented):
Talked about re-organizing this async so I'll review properly later, but you can have this one little nit now.

stevestreza-ksr:
Closing as this will get rebuilt on top of #2170

PR #2168, opened 2024-10-03
[MBL-1721] Crowdfunding Checkout Pledge Summary Table View Refactor

scottkicks:
# 📲 What

Instead of making the header text of each section a part of each cell, it'd be better to use actual table view header cells. We currently use a custom data source implementation that is overly complicated, somewhat difficult to understand, and not set up to easily define table view sections with headers.

After trying a couple of solutions, I chose to go with building a new data source class that implements the `UITableViewDiffableDataSource` API. This approach comes with three main wins for us:
- Simplifies the UI State

I've only implemented this to crowdfunding under the "No Shipping At Checkout" Feature Flag here, but I want to update late pledge to use this same table once merged.

# 🛠 How

- Rename `NoShippingPledgeRewardsSummaryTotalViewController` to `NoShippingPledgeRewardsSummaryViewController` so i makes sense
- Sets this new data source on the crowdfunding checkout table view
   - defines sections for the table as `header`, `reward`, `addOns`,  and `bonusSupport`

# 👀 See

| --- | --- | --- | --- |
| No Reward | reward (no addon) | reward + addon | reward + bonus support |

# ✅ Acceptance criteria

- [x] crowdfunding pledges work as expected
- [x] crowdfunding pledge summary table matches [designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=35-1532&node-type=frame&t=Gqnx64udPplFjYSb-0)

# ⏰ TODO
- [ ] Add snapshot tests before merging once the separator issue is figured out
- Update late pledge to reuse this table. Further simplifying our checkout flows and getting closer to unification
- Add section separators


Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
scottkicks, line 1: Reviewing commit by commit may be easier. I renamed this file so git will show it as a new file making it hard to see the specific changes I made.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+  private func tableViewStyle(_ tableView: UITableView) {
+    tableView.separatorInset = .zero
+    tableView.contentInsetAdjustmentBehavior = .never
+    tableView.isScrollEnabled = false
+    tableView.delegate = self
+    tableView.rowHeight = UITableView.automaticDimension
+    tableView.separatorStyle = .none
+    tableView.backgroundColor = .ksr_white
+    tableView.translatesAutoresizingMaskIntoConstraints = false
+  }
```
scottkicks, line 168: I'm setting `tableView.separatorStyle = .none` here because without it, the table view adds a separator line below each cell _and_ section header. 

I haven't found a good way to remove the lines from the section headers, yet. Here's a screenshot of what this looks like.

simulator_screenshot_B3AAAF1E-A06F-446D-B8DE-DA16E7A9EECA

If anyone has ideas short of adding a new separator view manually within each cell, I'd love to hear them.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+
+  private func setEntireViewToIsHidden(_ isHidden: Bool) {
+    self.view.isHidden = isHidden
+    self.pledgeTotalViewController.view.isHidden = isHidden
+    self.separatorView.isHidden = isHidden
+  }
+
+  // MARK: Apply DataSource Snapshot
+
+  private func applySnapshotToDataSource(from data: [PostCampaignRewardsSummaryItem]) {
```
scottkicks, line 214: This method defines the sections and data used in the table via `snapshot.appendSections` and `snapshot.appendItems`, respectively. Then, it updates the table all at once using `apply.`

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+    /// Decipher header vs reward objects from the `[PostCampaignRewardsSummaryItem]` data object.
+    let headerItemData = data.compactMap { item -> PledgeExpandableHeaderRewardCellData? in
+      guard case let .header(data) = item else { return nil }
+      return data
+    }
+
+    let rewardData = data.compactMap { item -> PledgeExpandableHeaderRewardCellData? in
+      guard case let .reward(data) = item else { return nil }
+      return data
+    }
```
scottkicks, line 226: This follows the pattern that the current data source class uses. I tried to focus only on implementing this new DiffableDataSource API and avoid changing how the table data is created and sent to this view controller from the view model.

I think there's an argument for this logic to live in the VM.

PR #2169, opened 2024-10-08
[MBL-1721] Crowdfunding Checkout Pledge Summary Table View Refactor

scottkicks:
# 📲 What

Instead of making the header text of each section a part of each cell, it'd be better to use actual table view header cells. We currently use a custom data source implementation that is overly complicated, somewhat difficult to understand, and not set up to easily define table view sections with UITableView headers.

After trying a couple of solutions, I built a new data source class that implements the `UITableViewDiffableDataSource` API. This approach comes with three main wins for us:
- Simplifies the UI State

I've only implemented this to crowdfunding under the "No Shipping At Checkout" Feature Flag here, but I want to update late pledge to use this same table once we've decided that this is a good solution.

# 🛠 How

- Rename `NoShippingPledgeRewardsSummaryTotalViewController` to `NoShippingPledgeRewardsSummaryViewController` so i makes sense
- Sets this new data source on the crowdfunding checkout table view
   - defines sections for the table as `header`, `reward`, `addOns`,  and `bonusSupport`

# 👀 See

| --- | --- | --- | --- |
| No Reward | reward (no addon) | reward (addon) | reward (addon + bonus support) |

# ✅ Acceptance criteria

- [x] crowdfunding pledges work as expected
- [x] crowdfunding pledge summary table matches [designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=35-1532&node-type=frame&t=Gqnx64udPplFjYSb-0)

# ⏰ TODO
- [ ] Update late pledge to reuse this table and late pledge snapshots
- [ ] Add shipping section to data source

### DIFF
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
@@ -306,7 +306,9 @@ extension NoShippingPledgeRewardsSummaryViewController: UITableViewDelegate {
   }
 
   func tableView(_: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+    let section = PledgeRewardsSummarySection.allCases[section]
+
     /// Hides the first section header because we're using our own UITableCell here.
-    return section == 0 ? 0 : UITableView.automaticDimension
+    return section == .header || section == .bonusSupport ? 0 : UITableView.automaticDimension
   }
 }
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
@@ -233,22 +233,20 @@ final class NoShippingPledgeRewardsSummaryViewController: UIViewController {
 
     // MARK: Reward Section
 
-    if rewardData.count > 0 {
-      let baseReward = rewardData[0]
-      snapshot.appendSections([.reward])
-      snapshot.appendItems([.reward(baseReward)], toSection: .reward)
+    let baseReward = rewardData[0]
+    snapshot.appendSections([.reward])
+    snapshot.appendItems([.reward(baseReward)], toSection: .reward)
 
-      // MARK: Add-Ons
+    // MARK: Add-Ons
 
-      let addOns = rewardData
-        .filter { reward in reward != baseReward && reward.text != Strings.Bonus_support() }
-        .map { PledgeRewardsSummaryRow.addOns($0) }
+    let addOns = rewardData
+      .filter { reward in reward != baseReward && reward.text != Strings.Bonus_support() }
+      .map { PledgeRewardsSummaryRow.addOns($0) }
 
-      /// We only want to add an add-ons section if any have been selected
-      if addOns.count > 0 {
-        snapshot.appendSections([.addOns])
-        snapshot.appendItems(addOns, toSection: .addOns)
-      }
+    /// We only want to add an add-ons section if any have been selected
+    if addOns.count > 0 {
+      snapshot.appendSections([.addOns])
+      snapshot.appendItems(addOns, toSection: .addOns)
     }
 
     // MARK: Bonus Support
--- a/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
+++ b/Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
@@ -214,8 +214,6 @@ final class NoShippingPledgeRewardsSummaryViewController: UIViewController {
   private func applySnapshotToDataSource(from data: [PostCampaignRewardsSummaryItem]) {
     var snapshot = NSDiffableDataSourceSnapshot<PledgeRewardsSummarySection, PledgeRewardsSummaryRow>()
 
-    // MARK: Header + Reward Sections
-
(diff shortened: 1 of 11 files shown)


Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// Decipher header vs reward objects from the `[PostCampaignRewardsSummaryItem]` data object.
+    let headerItemData = data.compactMap { item -> PledgeExpandableHeaderRewardCellData? in
+      guard case let .header(data) = item else { return nil }
+      return data
+    }
+
+    let rewardData = data.compactMap { item -> PledgeExpandableHeaderRewardCellData? in
+      guard case let .reward(data) = item else { return nil }
+      return data
+    }
```
scottkicks, line 226: This follows the pattern that the current data source class uses. I tried to focus only on implementing this new DiffableDataSource API and avoid changing how the table data is created and sent to this view controller from the view model.

I think there's an argument for this logic to live in the VM.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -0,0 +1,310 @@
...
+      .filter { reward in reward.text == Strings.Bonus_support() }
+      .map { PledgeRewardsSummaryRow.bonusSupport($0) }
+
+    if let bonusSupport = bonusSupportReward.first {
+      snapshot.appendSections([.bonusSupport])
+      snapshot.appendItems([bonusSupport], toSection: .bonusSupport)
+    }
+
+    /// Applys the changes above as a snapshot of all of the data needed to render the `UITableView `.
+    self.dataSource.apply(snapshot, animatingDifferences: true)
```
scottkicks, line 264: This method defines the sections and data used in the table via snapshot.appendSections and snapshot.appendItems, respectively. Then, it updates the table all at once using apply.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
scottkicks, line 1: Reviewing commit by commit will be easiest.
I renamed this file so git shows it as an entirely new file, making it hard to see the individual changes that I've made.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -184,6 +189,69 @@ final class NoShippingPledgeRewardsSummaryViewController: UIViewController {
...
+      guard case let .header(data) = item else { return nil }
+      return data
+    }
+
+    let rewardData = data.compactMap { item -> PledgeExpandableHeaderRewardCellData? in
+      guard case let .reward(data) = item else { return nil }
+      return data
+    }
+
+    let baseReward = rewardData[0]
```
stevestreza-ksr: Sanity checking, is it guaranteed that this array will always have at least one item in it?

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
stevestreza-ksr: Moving this to the VM seems like the right call to me, maybe having the VM produce a Publisher/Signal of snapshots that can be consumed in the VC and passed to the diffable data source.

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -184,6 +189,69 @@ final class NoShippingPledgeRewardsSummaryViewController: UIViewController {
...
+
+    // MARK: Header + Reward Sections
+
+    snapshot.appendSections([.header, .reward])
+    snapshot.appendItems([.header(headerItemData[0])], toSection: .header)
```
stevestreza-ksr, line 232: Sanity checking, is it guaranteed that this array will always have at least one item in it?

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/NoShippingPledgeRewardsSummaryViewController.swift
```diff
@@ -205,21 +208,32 @@ final class NoShippingPledgeRewardsSummaryViewController: UIViewController {
...
+      .map { PledgeRewardsSummaryRow.addOns($0) }
     /// We only want to add an add-ons section if any have been selected
     if addOns.isEmpty == false {
       snapshot.appendSections([.addOns])
       snapshot.appendItems(addOns, toSection: .addOns)
     }
 
+    // MARK: Bonus Support
+
+    let bonusSupportReward = rewardData.filter { $0.text == Strings.Bonus_support() }
```
stevestreza-ksr: Any way to make these filter on something more semantic than text equality?

stevestreza-ksr (commented):
I love using snapshotted data sources, they fit great with reactive programming and view models. Overall the pattern seems to work great here. Agreed on moving the snapshot generation to the view model.

PR #2171, opened 2024-10-09
[MBL-1613] Handle errors for late pledges without dismissing the view controller

ifosli:
# 📲 What

Handle errors in the late pledge flow without a confirm details page by just showing an error banner (not dismissing the vc). The error banner will persist if the error happens when creating the checkout. If it's an error that happens in a part of the flow the user can retry (after hitting "pledge" or adding a new card) the banner disappears.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1613)

Simulator Screen Recording - iPhone 16 Pro Max - 2024-10-09 at 15 37 42

# ✅ Acceptance criteria

- [x] The same errors show as before, but they no longer navigate backwards
- [x] If in debug, the server error always shows in addition to the general "Something went wrong"
- [x] I tested the "checkout error causes banner to persist" by hardcoding the view model to say it happened. So I've tested all the UI changes, but I haven't tried to generate new (real) errors.

### DIFF
--- a/Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
+++ b/Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
@@ -34,7 +34,7 @@ public protocol NoShippingPostCampaignCheckoutViewModelOutputs {
   var goToLoginSignup: Signal<(LoginIntent, Project, Reward), Never> { get }
   var paymentMethodsViewHidden: Signal<Bool, Never> { get }
   var processingViewIsHidden: Signal<Bool, Never> { get }
-  var showErrorBannerWithMessage: Signal<(String, /* persist: */ Bool), Never> { get }
+  var showErrorBanner: Signal<(message: String, persist: Bool), Never> { get }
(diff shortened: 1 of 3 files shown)


Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
```diff
@@ -34,12 +34,11 @@ public protocol NoShippingPostCampaignCheckoutViewModelOutputs {
   var goToLoginSignup: Signal<(LoginIntent, Project, Reward), Never> { get }
   var paymentMethodsViewHidden: Signal<Bool, Never> { get }
   var processingViewIsHidden: Signal<Bool, Never> { get }
-  var showErrorBannerWithMessage: Signal<String, Never> { get }
+  var showErrorBannerWithMessage: Signal<(String, /* persist: */ Bool), Never> { get }
```
scottkicks, line 37: nit: Adding parameter names instead of an inline comment might clarify that this output has the option to persist.

scottkicks (approved):
LGTM 👍

PR #2172, opened 2024-10-10
[MBL-1777] Register To Receive APNs Token On App Launch

scottkicks:
# 📲 What

Adds the [registerForRemoteNotifications()](https://developer.apple.com/documentation/uikit/uiapplication/1623078-registerforremotenotifications) method of [UIApplication](https://developer.apple.com/documentation/uikit/uiapplication) to`AppDelegate.didFinishLaunchingWithOptions` to request the APNs device token on app launch.

# 🤔 Why

Tarveen and E have called out that they haven’t been getting Push Notifications for a while. 

The running theory is that their APNs token may have been replaced by the APNs Server for one of the following reasons:
- restoring the phone from a backup
- transferring data from one phone to another
- not having reinstalled the app or logged out for an extended period of time

Currently, we're only registering the app and receiving the token at log-in when users opt-in to push notifications, so... 

This PR adds that method above so that the app registers and then receives the current APNs token on app launch in the `UIApplication.didRegisterForRemoteNotificationsWithDeviceToken` callback.

From there, we send the token to our `v1/push_tokens` server endpoint as expected.

# 👀 See

No Visible change

# ✅ Acceptance criteria

It isn't easy to test this, but I've verified that this callback is now called on app launch _and_ when opting into notifications.
- For more info, check out the Apple recommended approach for [Registering an app and retrieving it's device token](https://developer.apple.com/documentation/usernotifications/registering-your-app-with-apns#Register-your-app-and-retrieve-your-apps-device-token)

# ⏰ TODO

- [ ] Check in with Tarveen and E once this goes out to make sure they are getting notifications again


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -263,6 +263,8 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
 
     UNUserNotificationCenter.current().delegate = self
 
+    UIApplication.shared.registerForRemoteNotifications()
```
stevestreza-ksr, line 266: It looks like we have a bunch of stuff in AppDelegateViewModel and PushRegistration already which this seems to bypass, can we use that object for this and update any signals if they aren't firing, rather than call this manually here?

scottkicks: The `PushRegistration.register` method is only triggered once the user interacts with the request permissions dialog. 

I can add a `registerOnAppLaunch` method to `PushRegistration` and set up a new input that then allows us to call `UIApplication.shared.registerForRemoteNotifications()` from the view model. This initially felt a little over-engineered to me, though.

scottkicks: Here's a [solution](https://github.com/kickstarter/ios-oss/pull/2172/commits/e10a7e890a773e8ee3b31637dce8831970386aff) that uses the existing PushRegistration class. Lmk what ya think!

stevestreza-ksr: - In cases where the user has already authorized notifications, does the `UIApplication.shared.registerForRemoteNotifications()` call happen in PushRegistration?
- In cases where the user has not authorized, does this trigger the permission dialog? I suspect without the call to `UNUserNotificationCenter.current().requestAuthorization` that it might fail. But either way we should control when the dialog appears.

I don't think we need a new method for PushRegistration but we can just stick something into app launch signals to trigger the existing code path.

scottkicks: Yeah the existing code path registers for remote notifications, but only after requesting authorization, and we present the request alert once the user has logged in and opted in to receive notifications via a custom KS alert that is shown just before the official Apple-generated permissions dialog.

The commit I just put up makes the call to register via the initial app launch signal like we want it to. I like this because it keeps the logic that requests PN authorization separate from registering.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -276,12 +276,16 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
 
     // Push notifications
 
-    let pushNotificationsPreviouslyAuthorized = self.applicationLaunchOptionsProperty.signal
-      .flatMap { _ in AppEnvironment.current.pushRegistrationType.hasAuthorizedNotifications() }
+    let pushNotificationsPreviouslyAuthorizedAndRegistered = self.applicationLaunchOptionsProperty.signal
+      .flatMap { _ in
+        AppEnvironment.current.pushRegistrationType.registerForPushTokenOnAppLaunch()
```
ifosli, line 281: Does this get called on app launch even for people that haven't explicitly opted into notifications? That's what it looks like to me, though I could definitely be wrong since I'm not very familiar with this code. If it's getting called before people opt in to notifications, that's a problem.

scottkicks: it's a good call out @ifosli . 

So, after clearing and resetting my simulator and testing on my physical device, I'm now seeing that `UIApplication.shared.registerForRemoteNotifications()` is already being called consistently in our existing `PushRegistration.register(for)` method on app launch when logged in or logged out. That is if the user has previously opted in (expected behavior).

Seeing this new behavior shows that this solution isn't going to solve the initial problem.

Next steps should be to update the current ticket, but we'll need to keep digging to determine why Tarveen and E aren't getting PNs.

scottkicks:
https://github.com/kickstarter/ios-oss/pull/2172#discussion_r1799896847

PR #2174, opened 2024-10-10
[MBL-1721] Pledge Rewards Summary Table View Refactor

scottkicks:
# 📲 What

Instead of making the header text of each section a part of each cell, it'd be better to use actual table view header cells. We currently use a custom data source implementation that is overly complicated, somewhat difficult to understand, and not set up to easily define table view sections with UITableView headers.

I've built a new data source class that implements the `UITableViewDiffableDataSource` API. 
This approach comes with three main wins for us:
- Simplifies the UI State

These changes are behind the "No Shipping At Checkout" Feature Flag

# 🛠 How
- Implement the new data source. This applies a snapshot that defines sections for the table as `header`, `reward`, `addOns`,  and `bonusSupport`
- Update table view cell styling. Includes some Prelude cleanup.

# 👀 See

| --- | --- | --- | --- |
| No Reward | reward (no addon) | reward (addon) | reward (addon + bonus support) |

# ✅ Acceptance criteria

- [x] crowdfund and late pledge pledges work as expected
- [x]  tables match our [designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=35-1532&node-type=frame&t=Gqnx64udPplFjYSb-0)

# ⏰ TODO
- [ ] Add snapshot tests onces issues are resolved


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

scottkicks:
I have two outstanding issues that are giving me trouble that I could use extra eyes on if anyone has time.

1. The AddOns section header text shows when no add-ons have been selected, only when bonus support is added.
drawing

Setting breakpoints in the snapshot definition, [here](https://github.com/kickstarter/ios-oss/blob/b8d1978699b7afbda0d9193785ee0913d24b6676/Kickstarter-iOS/Features/PledgePaymentMethods/Datasource/NoShippingPledgeRewardsSummaryDiffableDataSource.swift#L96), shows that in this case we're not actually appending `.addOns` to the sections. I'm trying to understand why the data source think that it has.

2. When adding multiple different addOns the ones after the first one are squished.
drawing

The tricky part is that the cell styling is set based on its relation to the section header. The cells that come after the first don't need to be so conditionally rendering those is a bit of a curveball.

stevestreza-ksr (approved):
This looks pretty straightforward. It might make sense to move some of the constants in those classes out into a constants block.

Kickstarter-iOS/Features/PledgePaymentMethods/Datasource/NoShippingPledgeRewardsSummaryDiffableDataSource.swift
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+  }
+}
+
+/**
+ Creates `NSDiffableDataSourceSnapshot` to apply to the pledge summary table view data source on our crowdfunding and late pledge checkout screens.
+
+ - parameter headerData: The `PledgeExpandableHeaderRewardCellData`that represents the tables custom header `UITableViewCell`.
+ - parameter rewards: The `[PledgeExpandableHeaderRewardCellData]`that represents the tables reward data (baseReward, addOns, and bonusSupport).
```
ifosli, line 61: nit `table's` instead of `tables` in this line and the line above.

Kickstarter-iOS/Features/PledgePaymentMethods/Views/Cells/PostCampaignPledgeRewardsSummaryCell.swift
```diff
@@ -39,17 +46,16 @@ final class PostCampaignPledgeRewardsSummaryCell: UITableViewCell, ValueCell {
 
     self.amountLabel.adjustsFontForContentSizeCategory = true
 
-    _ = self.rootStackView
-      |> rootStackViewStyle(self.traitCollection.preferredContentSizeCategory > .accessibilityLarge)
+    self.applyContainerStackViewStyle(self.containerStackView)
```
ifosli, line 49: This kind of naming convention works, too! I'm not sure we have enough style helpers that aren't in prelude to have a project-wide naming convention yet, but I would still lean toward `styleContainerStackView` and ect instead, just because it's a little shorter. Definitely not blocking, though!

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPostCampaignViewControllerTests/testView_AddOns_LoggedOut.lang_en_device_pad.png
ifosli, line 1: There's a reward missing from this table - the second add on doesn't show. Can you please look into this and make sure that the table is capable of showing multiple rows of add-ons, both in tests and in the app?

Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPostCampaignViewControllerTests/testView_PledgeContext_UnavailableStoredCards.lang_de_device_phone4inch.png
ifosli, line 1: This one's concerning, too - it's showing the "Add-ons" title for the shipping section, which I thought was fixed. Can you fix this for the post campaign version, too, if you haven't already? Or look into this and see what's going on if it should be fixed?

Library/ViewModels/PostCampaignPledgeRewardsSummaryViewModel.swift
```diff
@@ -48,7 +48,10 @@ public protocol PostCampaignPledgeRewardsSummaryViewModelInputs {
 
 public protocol PostCampaignPledgeRewardsSummaryViewModelOutputs {
   var configurePledgeTotalViewWithData: Signal<PledgeSummaryViewData, Never> { get }
-  var loadRewardsIntoDataSource: Signal<[PostCampaignRewardsSummaryItem], Never> { get }
+  var loadRewardsIntoDataSource: Signal<
+    (headerData: PledgeExpandableHeaderRewardCellData?, rewards: [PledgeExpandableHeaderRewardCellData]),
```
ifosli, line 52: Nit: If `PledgeExpandableHeaderRewardCellData` is the data type that's used for every cell, including rewards cells, can we please rename it?

ifosli (changes requested):
Some of these snapshots reveal issues - primarily "add-ons" is showing as a title for the shipping section for one of the post campaign vcs (immediate example for why I think it'd be nice to combine these table classes, but I do still understand if that's out of scope) and tables that should have two separate add-ons is only showing one. I left a bunch of other minor comments (mostly typos and naming convention stuff for styling helper functions) but do what you want with those. The table looks good, though! Overall I like this a lot!

Kickstarter-iOS/Features/PledgePaymentMethods/Controller/PostCampaignPledgeRewardsSummaryViewController.swift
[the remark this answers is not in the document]
scottkicks: Agreed. I've noted it in the ticket that tackles unifying late pledge and crowdfunding checkout so that this PR doesn't get too big.

ifosli (approved):
Nice!

PR #2175, opened 2024-10-15
Allowlist all stripe requests, not just stripe elements

ifosli:
# 📲 What

Allowlist all requests from `*.stripe.com`, `*.stripe.network`, or `stripecdn.com` in the survey/backing details webview.

# 🤔 Why

The webview got stuck in an infinite loading state earlier today. I'm not sure if it's related to our policy not allowing some of these requests through, but it definitely might be. I noticed `m.stripe.network` requests that we were blocking when I was debugging earlier. [Aidan did a little digging and found that there's also `m.stripe.com` and `r.stripe.com` domains on the checkout page](https://kickstarter.slack.com/archives/C06EKGN4MHU/p1729019704602539?thread_ts=1729007161.056579&cid=C06EKGN4MHU).

I let this idle for a while and noticed `stripecdn.com` requests as well, for invisible captchas, so I've added those as well. It required reformatting the helper function for this to make sure that each domain is only allowed if it has the correct tld.

# 🛠 How

I'm doing a general component match that allows all subdomains - I think that's safe since they all have to be owned by stripe, and being general in webviews is nice for backwards compatibility. If anyone has concerns about this approach, though, definitely let me know!

# ✅ Acceptance criteria

- [x] Tests pass

### DIFF
--- a/Library/ViewModels/SurveyResponseViewModel.swift
+++ b/Library/ViewModels/SurveyResponseViewModel.swift
@@ -164,10 +164,11 @@ private func isSurvey(request: URLRequest) -> Bool {
   return true
 }
 
-// Returns true if the url host is of the form *.stripe.com or *.stripe.network.
+// Returns true if the url host is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
 private func isStripeRequest(_ request: URLRequest) -> Bool {
+  let stripeDomains = ["stripe.com", "stripe.network", "stripecdn.com"]
+
   guard let host = request.url?.host?.lowercased() else { return false }
-  let components = host.split(separator: ".")
-  return components.count == 3 && components[1] == "stripe" &&
-    (components[2] == "com" || components[2] == "network")
+  let withoutSubdomain = host.split(separator: ".").suffix(2).joined(separator: ".")
+  return stripeDomains.contains(withoutSubdomain)
 }
--- a/Library/ViewModels/SurveyResponseViewModel.swift
+++ b/Library/ViewModels/SurveyResponseViewModel.swift
@@ -107,7 +107,7 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
 
     self.policyDecisionProperty <~ newRequest
       .map { request in
-        if isStripeElement(request) {
+        if isStripeRequest(request) {
           return true
         }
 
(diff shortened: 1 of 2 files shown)


Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -164,6 +164,10 @@ private func isSurvey(request: URLRequest) -> Bool {
...
 }
 
-private func isStripeElement(_ request: URLRequest) -> Bool {
-  return request.url?.host == "js.stripe.com"
+// Returns true if the url host is of the form *.stripe.com or *.stripe.network.
+private func isStripeRequest(_ request: URLRequest) -> Bool {
+  guard let host = request.url?.host?.lowercased() else { return false }
+  let components = host.split(separator: ".")
+  return components.count == 3 && components[1] == "stripe" &&
+    (components[2] == "com" || components[2] == "network")
```
scottkicks, line 172: _not merge blocking:_ I'm not sure if we have other methods that check for a URL host like this, but we could consider setting up an enum to check against here to avoid hardcoding the strings and make it easier to verify other URLs in the future.

This could be too specific of a use case for that, though, so your call!

scottkicks (approved):
LGTM

Just one non-blocking comment

Library/ViewModels/SurveyResponseViewModel.swift
scottkicks: a unit test for this might also be helpful

ifosli: I don't think it's worth setting up constants for this, to be honest. We never handle third party hosts outside of this webview, and I don't expect us to do so in the future either (unless we decide to add more webviews 😭 ). I'll add tests, though! Good call!

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -164,6 +164,11 @@ private func isSurvey(request: URLRequest) -> Bool {
...
 }
 
-private func isStripeElement(_ request: URLRequest) -> Bool {
-  return request.url?.host == "js.stripe.com"
+// Returns true if the url host is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
+private func isStripeRequest(_ request: URLRequest) -> Bool {
+  let stripeDomains = ["stripe.com", "stripe.network", "stripecdn.com"]
+
+  guard let host = request.url?.host?.lowercased() else { return false }
+  let withoutSubdomain = host.split(separator: ".").suffix(2).joined(separator: ".")
```
stevestreza-ksr, line 172: Not necessary for this but it might be useful to make a category on URL/String that returns a "base domain" if this kind of pattern is used anywhere else

stevestreza-ksr (approved):
Lgtm!

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: I don't think it is (the only other hosts we care about are kickstarter ones), but agreed - if I see this elsewhere or find myself needing it in the future, I'll pull this out into a utility function then!

PR #2178, opened 2024-10-16
[MBL-1783 Unhide rewards immediately if there's no shippable rewards

ifosli:
# 📲 What

We currently hide all rewards until the shipping location selector is done loading. If there aren't any shippable rewards, unhide the rewards immediately.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1783)

# ✅ Acceptance criteria

- [x] Rewards without any shippable rewards show their rewards correctly

### DIFF
--- a/Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
+++ b/Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
@@ -193,11 +193,10 @@ final class WithShippingRewardsCollectionViewController: UICollectionViewControl
 
     self.viewModel.outputs.rewardsCollectionViewIsHidden
       .observeForUI()
-      .observeValues { [weak self] shimmerLoadingViewIsHidden in
+      .observeValues { [weak self] rewardsCollectionViewIsHidden in
         guard let self else { return }
 
-        self.collectionView.isHidden = !shimmerLoadingViewIsHidden
-        self.collectionView.layoutIfNeeded()
+        self.collectionView.isHidden = rewardsCollectionViewIsHidden
       }
 
     self.viewModel.outputs.rewardsCollectionViewFooterIsHidden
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
```diff
@@ -193,11 +193,10 @@ final class WithShippingRewardsCollectionViewController: UICollectionViewControl
 
     self.viewModel.outputs.rewardsCollectionViewIsHidden
       .observeForUI()
-      .observeValues { [weak self] shimmerLoadingViewIsHidden in
+      .observeValues { [weak self] rewardsCollectionViewIsHidden in
         guard let self else { return }
 
-        self.collectionView.isHidden = !shimmerLoadingViewIsHidden
-        self.collectionView.layoutIfNeeded()
```
ifosli, line 200: I deleted this because now that we're not always waiting for the location view to load, it gets called before the view hierarchy is ready. I did test a bunch of random projects (hence the local pickup bug I found) and didn't see any UI weirdness caused by removing this line.

Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
```diff
@@ -193,11 +193,10 @@ final class WithShippingRewardsCollectionViewController: UICollectionViewControl
...
+        self.collectionView.isHidden = rewardsCollectionViewIsHidden
```
stevestreza-ksr, line 199: Making sure the intent was to invert this isHidden check

stevestreza-ksr (approved):
Can't really be sure that this behavior is fully correct as I haven't been in this area much, but changes seem to make sense to me!

Kickstarter-iOS/Features/RewardsCollection/Controller/WithShippingRewardsCollectionViewController.swift
ifosli: Yeah, I found the naming to be super confusing before, so I've inverted it (both here and in the view model) to be more intuitive. It's still in sync with the underlying value; it's just the output signal value that's inverted. Thanks for checking though!

scottkicks (approved):

PR #2179, opened 2024-10-17
[MBL-1783] Don't filter local pickup based on shipping rule

ifosli:
# 📲 What

Previously, we handled local pickup rewards similar to shippable rewards, and required them to match the shipping rule's location. Now we handle them similar to digital rewards instead (they'll always show as a pledge option).

This pr also the reward summary table, which was waiting for a non-nil shipping rule before showing up.

# 🤔 Why

With the previous setup, we don't include local pickup rewards when determining what locations are selectable. If a project had only local pickup rewards, we filtered out all rewards except the "no reward" option on iOS, with no way to unhide them (the shipping location selector only shows if there's at least one shippable reward).

# ✅ Acceptance criteria

- [x] A local pickup can be selected when there's no shipping rule or a mismatch between the selected shipping rule and the local pickup location.
- [x] In both of these cases, a user's pledge goes through. (Tested in staging - I didn't find a project in staging that only has local pickup rewards, so I hardcoded the shipping rule to nil.)

# ⏰ TODO

- ~~Not in this pr, I'm going to file a ticket (if we don't already have one) to add the local pickup view back in to the pledge screen UI. The current UI does not make the "local pickup" part clear enough.~~ Created [MBL-1784](https://kickstarter.atlassian.net/browse/MBL-1784) to track this.

[MBL-1784]: https://kickstarter.atlassian.net/browse/MBL-1784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Kickstarter-iOS/Features/PledgeView/Controllers/__Snapshots__/NoShippingPledgeViewControllerTests/testView_PledgeContext_FixPaymentMethod_ErroredCard.lang_en_device_phone4_7inch.png
ifosli, line 1: I'm a little concerned that we didn't catch that the pledge summary table was dependent on the shipping rule sooner, considering what this screenshot looked like before, but at least it's fixed now!

stevestreza-ksr (approved):
Would be really good to have a test to cover this, but otherwise LGTM!

PR #2181, opened 2024-10-22
Update Firebase to 11.4.0

stevestreza-ksr:
# 📲 What

Updates the Firebase package to 11.4.0.

# 🤔 Why

Fixing a crash apparently due to Firebase SDK being out of date.

# 🛠 How

Did the upgrade to 11.4 and made sure it built. Sanity checked against the [release notes](https://firebase.google.com/support/release-notes/ios#11.4.0) and didn't see anything breaking between 11.4.0 and 10.11.0 that affects us. Building for release works, and different Firebase systems appear to be activating.


ifosli (approved):
Looks great! Thanks for taking this on!

PR #2182, opened 2024-10-23
[MBL-1821] Get shipping info from backing if there's no shipping rule selected

ifosli:
# 📲 What

With the "no shipping at pledge" flag on, trying to update the payment method of a reward with shipping doesn't work. We didn't show the shipping in the reward summary table and we defaulted the shipping amount to zero. This pr fixes both of these and adds a test to ensure that'll stay true.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1821)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 - 2024-10-23 at 11 45 16 | Simulator Screen Recording - iPhone 16 - 2024-10-23 at 12 12 44 |

# ✅ Acceptance criteria

- [x] Changing payment method for reward with shipping with "no shipping at pledge" flag on works
- [x] I also tested changing the reward, including the shipping location, and everything looks good!

### DIFF
--- a/Library/ViewModels/NoShippingPledgeViewModel.swift
+++ b/Library/ViewModels/NoShippingPledgeViewModel.swift
@@ -124,6 +124,13 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
     )
     .map(calculateAllRewardsTotal)
 
+    let initialShippingTotal = project.map { project in
+      guard let backing = project.personalization.backing else {
+        return 0.0
+      }
+      return backing.shippingAmount ?? 0.0
+    }
+
     let calculatedShippingTotal = Signal.combineLatest(
       selectedShippingRule.skipNil(),
       rewards,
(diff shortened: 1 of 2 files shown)


stevestreza-ksr (approved):
As a fix to unblock the release, I think this is fine.

Beyond the immediate needs though, there should be a refactor to condense the shipping information into one Signal that can be consumed by `initialShippingTotal` and `shippingLocation`. That way there is only one source of truth for shipping. I'll bring this up in postmortem.

ifosli:
> As a fix to unblock the release, I think this is fine.
> 
> Beyond the immediate needs though, there should be a refactor to condense the shipping information into one Signal that can be consumed by `initialShippingTotal` and `shippingLocation`. That way there is only one source of truth for shipping. I'll bring this up in postmortem.

Yeah, I agree it'd be nice to make sure that the shipping total and location are always from the same source. This class is still (at least partially) set up to respond to shipping location changes, since the old pledge VC included that and pledge redemption work was rushed enough that getting rid of dead code/simplifying code based on the current UI was considered a tech debt thing instead of a "required for release" thing. I'm hoping we can refactor the class into something simpler (including how we calculate shipping) as part of the unification of pledge flows.

PR #2192, opened 2024-11-06
Make apple pay button do nothing if we don't have checkout created

ifosli:
# 📲 What

Currently, we do three things when the apple pay button is tapped during the late pledge flow:
1. We try to start the apple pay checkout
2. We track that apple pay checkout is started
3. We show the "Processing..." overlay

If checkout isn't successfully created, however, we only 2. and 3. Because we never actually start apple pay checkout, checkout never completes (or errors), so the "Processing..." overlay is never dismissed.

I'm updating the "Processing..." overlay to only appear when the apple pay button is tapped if the signals the apple pay flow relies on have values to emit. 

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1845)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Plus - 2024-11-05 at 16 59 18 | Simulator Screen Recording - iPhone 16 Plus - 2024-11-05 at 17 00 51 |

# ✅ Acceptance criteria

- [x] Processing overlay does not show when apple pay button is tapped when there's an error
- [x] Processing overlay shows when apple pay is initiated from normal checkouts


scottkicks (approved):
Looks good! I've had kickstarter-tests time-out on my last couple of PRs. I'm not sure why yet, but re-running from circle ci has fixed it so you may need to do the same.

stevestreza-ksr (approved):

PR #2194, opened 2024-11-12
[MBL-1855] Fix Estimated Shipping Conversion

scottkicks:
# 📲 What

Fixes the estimated shipping text on reward cards so that it displays based on the user's selected currency settings and converts the values when the user's chosen currency differs from the project's currency configuration. 

# 🛠 How

- Updates our `estimatedShippingText` shared function to use the project's currency.
- Update the label to use the conversion if the user's chosen currency differs from the project's currency.

# 👀 See

In this example, my user settings have the British pound currency selected. The project is based in the US.

The expected behavior is that the estimated shipping on rewards cards is shown in/converted to pounds, and on the checkout screen's estimated shipping view, the range in the US is shown with the conversion to pounds displayed underneath it.

Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-11-12 at 11 41 53

# ✅ Acceptance criteria

- [ ] Estimated shipping on Reward cards and checkout are displayed as the expected behavior above and match [the designs](https://www.figma.com/design/z4faZM4BQwbyiDe7y8tbpP/Pledge-Redemption?node-id=0-1&node-type=canvas&t=W7XKlVYXEAq16uCY-0)

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
@@ -396,7 +396,8 @@ extension PPOProjectCardModel {
 
     let projectAnalyticsFragment = backing?.project?.fragments.projectAnalyticsFragment
 
-    if let image, let title, let pledge, let creatorName, let projectAnalyticsFragment, let backingDetailsUrl {
+    if let image, let title, let pledge, let creatorName, let projectAnalyticsFragment,
+       let backingDetailsUrl {
       self.init(
         isUnread: true,
         alerts: alerts,
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
@@ -61,7 +61,10 @@ final class PPOProjectCardViewModel: PPOProjectCardViewModelType {
 
   // MARK: - Outputs
 
-  var viewBackingDetailsTapped: AnyPublisher<(), Never> { self.viewBackingDetailsSubject.eraseToAnyPublisher() }
+  var viewBackingDetailsTapped: AnyPublisher<(), Never> {
+    self.viewBackingDetailsSubject.eraseToAnyPublisher()
+  }
+
   var sendMessageTapped: AnyPublisher<(), Never> { self.sendCreatorMessageSubject.eraseToAnyPublisher() }
   var actionPerformed: AnyPublisher<Action, Never> { self.actionPerformedSubject.eraseToAnyPublisher() }
 
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -687,19 +687,15 @@ public func estimatedShippingText(
 
   guard estimatedMin > 0, estimatedMax > 0 else { return nil }
 
-  let currentCountry = project.stats.currentCountry ?? Project.Country.us
-
   let formattedMin = Format.currency(
     estimatedMin,
-    country: currentCountry,
-    omitCurrencyCode: project.stats.omitUSCurrencyCode,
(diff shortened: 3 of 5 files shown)


Library/SharedFunctions.swift
```diff
@@ -687,19 +687,15 @@ public func estimatedShippingText(
 
   guard estimatedMin > 0, estimatedMax > 0 else { return nil }
 
-  let currentCountry = project.stats.currentCountry ?? Project.Country.us
```
stevestreza-ksr, line 690: Nit: Since `project.country` is used in multiple places, keep this variable and just assign it to `project.country` instead.

Library/ViewModels/RewardCardViewModel.swift
```diff
@@ -132,7 +132,9 @@ public final class RewardCardViewModel: RewardCardViewModelType, RewardCardViewM
       .map { reward, project, shippingRule in
         guard let locationId = shippingRule?.location.id else { return nil }
 
-        return estimatedShippingText(for: [reward], project: project, locationId: locationId)
+        return project.stats.needsConversion
+          ? estimatedShippingConversionText(for: [reward], project: project, locationId: locationId)
+          : estimatedShippingText(for: [reward], project: project, locationId: locationId)
```
stevestreza-ksr, line 137: Looks like `estimatedShippingConversionText` does its own check for `project.stats.needsConversion`, does it make sense to move this call to `estimatedShippingText(…)` into that function? (It might not)

stevestreza-ksr (approved):
Left some suggestions but otherwise seems fine to me!

Library/SharedFunctions.swift
```diff
@@ -687,19 +687,15 @@ public func estimatedShippingText(
...
-
   let formattedMin = Format.currency(
     estimatedMin,
-    country: currentCountry,
-    omitCurrencyCode: project.stats.omitUSCurrencyCode,
```
ifosli, line 695: I think we still need to know if we should omit the us currency code. Don't we want to include `USD` if someone ex has their preferred currency set to canadian dollars and the project's currency is in usd? And not include it if the project and the user's currency are both usd?

ifosli (changes requested):
I agree with Steve's comments. Also, I think we really do need the `omitUSCurrencyCode` still. Looks good other than that!

And outside the scope of this pr maybe, but I think this only solves half the bug. Are you planning to fix the "estimated shipping reacts to country in location selector" piece separately?

Library/SharedFunctions.swift
scottkicks: The designs don't include the currency code as well as the currency symbol so I removed this option. I can double check with Alison, unless you've had a conversation about this already.

scottkicks:
> And outside the scope of this pr maybe, but I think this only solves half the bug. Are you planning to fix the "estimated shipping reacts to country in location selector" piece separately?

the estimated shipping is now reacting to shipping location changes via the drop down. See the gif above. Is there more to that bug that I'm missing?

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
scottkicks, line 1: formatter caught these

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
scottkicks, line 1: more formatting

Library/SharedFunctions.swift
scottkicks: Yeah it looks like we have inconsistencies in a few places when it comes to showing or not showing currency codes. Both in the app and in the designs even. I'll put `omitUSCurrencyCode` back in here since this is the behavior that has been, but also make a note to call this out to alison and ask for more direction.

ifosli: Sounds good!

ifosli (approved):

PR #2196, opened 2024-11-13
[MBL-1459] Tab bar badging for PPO

stevestreza-ksr:
# 📲 What

This PR connects the badge on the tab bar to the User property that says there is a pending PPO action.

# 🛠 How

The User object has a `ppoHasAction` property now. When this field is true and the PPO feature flag is enabled, we show a dot badge rather than a count by passing an empty string.

There was a bunch of existing badging support that already worked with a string. This PR uses that existing badging with the same signals, and extends it where the PPO feature flag is enabled.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ✅ Acceptance criteria

Tests should pass.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/ViewModels/RootViewModel.swift:272:1 | warning: (trailingClosures) Use trailing closure syntax where applicable. |
| Library/ViewModels/RootViewModel.swift:274:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |
| Library/ViewModels/RootViewModelTests.swift:617:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/RootViewModelTests.swift:618:1 | warning: (indent) Indent code in accordance with the scope level. |
| Library/ViewModels/RootViewModelTests.swift:619:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

jovaniks (approved):
LGTM

Library/ViewModels/RootViewModel.swift
```diff
@@ -457,7 +470,12 @@ private func tabData(forUser user: User?) -> TabBarItemsData {
 extension TabBarItemsData: Equatable {}
 extension TabBarItem: Equatable {}
 
-private func activitiesBadgeValue(with value: Int?) -> String? {
+private func activitiesBadgeValue(with value: Int?, hasPPOAction: Bool) -> String? {
```
ifosli, line 473: I like this interpretation, but my understanding was that if PPO is enabled, we never want to show a count. So ex if I have PPO enabled, no PPO actions, but I do have activity notifications, I was expecting the dot to show with no count. Would you mind checking with Alison what she intended before you submit this?

Library/ViewModels/RootViewModelTests.swift
```diff
@@ -611,6 +611,50 @@ final class RootViewModelTests: TestCase {
...
+      RemoteConfigFeature.pledgedProjectsOverviewEnabled.rawValue: false
+    ]
+
+    self.setBadgeValueAtIndexValue.assertValues([])
+    self.setBadgeValueAtIndexIndex.assertValues([])
+
+    withEnvironment(currentUser: user, remoteConfigClient: remoteConfig) {
+      self.vm.inputs.viewDidLoad()
+
+      self.setBadgeValueAtIndexValue.assertValues([nil])
```
ifosli, line 654: I'm very confused by this test - shouldn't the value here be `54`, not nil? Would you mind explaining why this is intended or looking into why the test isn't working?

ifosli (changes requested):
Code looks good, I just have a couple concerns around intended behavior (of the dot and of that last test). Let me know if I'm wrong about changes being needed and I'd be happy to approve.

Library/ViewModels/RootViewModelTests.swift
stevestreza-ksr: Good call, tests updated to include testing for a badge.

Library/ViewModels/RootViewModel.swift
stevestreza-ksr: Alison says we should show the count in the case where PPO is enabled, there is no PPO action, and there is activity count.

stevestreza-ksr:
Feel free to merge on my behalf once a second approval lands. 😄

ifosli (approved):
Nice!

PR #2197, opened 2024-11-13
[MBL-1494] Wire up contact creator action in PPO

ifosli:
# 📲 What

When a user takes the contact creator action in PPO, present a MessageDialogViewController.

Note: The first commit updates the `MessageDialogViewController` to be easier to use in two ways. First, I got rid of an unused enum case (that I think was used back when our app supported creators as well). Second, I've updated the `MessageSubject` to consist only of the fields it needs (project id and name) instead of an entire `Project` object.

Once that was done, wiring it up was straightforward.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1494)

Simulator Screen Recording - iPhone 16 Pro Max - 2024-11-13 at 16 37 07

# ✅ Acceptance criteria

- [x] Message to creator is successfully sent (see recording)
- [x] Message to creator can be dismissed
- [x] Messages can still be sent elsewhere in the app (I tested from manage pledge and creator profile)


stevestreza-ksr (approved):
Love it when we remove more LoC than we add! 🤩

scottkicks (approved):

PR #2198, opened 2024-11-14
[MBL-1558] Localization for PPO

stevestreza-ksr:
# 📲 What

Adding localized strings for PPO.

# ✅ Acceptance criteria

Tests should pass.


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift:398:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

jovaniks (commented):
Looks good

jovaniks (approved):

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
```diff
@@ -8,8 +8,7 @@ struct PPOProjectCreator: View {
 
   var body: some View {
     HStack(alignment: .firstTextBaseline) {
-      // TODO: Localize
-      Text("Created by **\(self.creatorName)**")
+      Text("\(Strings.project_menu_created_by()) **\(self.creatorName)**")
```
ifosli, line 11: I don't think this is blocking release since it looks like we got lucky, but we should never combine translated strings like this; instead, the localized string should take in the creator name as input (there's no guarantee that this order of words makes sense in other languages).  See https://github.com/kickstarter/ios-oss/blob/459b37f241b0bf0fa0f898c6c8c6402be013a7f2/Library/Strings.swift#L25476 for an example of what this should look like.

Would you file a ticket for this? I think this one might get a little extra complicated since we also want the creator name formatted differently.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -38,8 +38,7 @@ struct PPOProjectDetails: View {
             .lineLimit(Constants.titleLineLimit)
         }
         if let symbol = pledge.symbol, let amount = pledge.amount {
-          // TODO: Localize
-          Text("\(symbol)\(amount) pledged")
+          Text(Strings.Pledge_amount_pledged(pledge_amount: "\(symbol)\(amount)"))
```
ifosli, line 41: Nit: Should this be using a `currency(` or `formattedCurrency(` helper (in Format.swift) to construct the "\(symbol)\(amount)" string? Ex for backers that don't have USD as their default currency, we usually show "US$" as the currency symbol, and I doubt we're handling that correctly right now. Maybe the view model could expose a `formattedAmount` string that we construct there?

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/__Snapshots__/PPOProjectCardTests/testCardAuthPledgeDropped.cardAuthPledgeDropped.png
ifosli, line 1: It looks like the localized strings for these buttons use title case, which is not what the designs say. And the "Take Survey" string is also not what I expected. Will you check with Alison or with Android if these strings are intentionally different from the figma? If not, I think we'll need to add new strings with the correct capitalization. (I'm okay with using these strings for now, though.)

ifosli (approved):
There are a few things I found concerning, but nothing that needs to be handled in this pr.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/__Snapshots__/PPOProjectCardTests/testCardAuthPledgeDropped.cardAuthPledgeDropped.png
stevestreza-ksr: I will follow up with Alison next week; these strings ARE consistent with [Android](https://github.com/kickstarter/android-oss/blob/81c088c1b9de96b51bcfbaad8cb6bba447d8a2db/app/src/test/java/com/kickstarter/ui/activities/compose/PPOCardViewKtTest.kt#L116) as of now, though.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
stevestreza-ksr: [Ticketed](https://kickstarter.atlassian.net/browse/MBL-1878) this for now. The model uses GraphAPI.MoneyFragment internally, so this would require making a bunch of changes just to support this formatting type. I'd recommend instead adding a formatter for MoneyFragment, which would need testing.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
stevestreza-ksr: Good callout. [Ticketed](https://kickstarter.atlassian.net/issues/MBL-1879).

PR #2200, opened 2024-11-18
[MBL-1891] Estimated Shipping Doesn't Update On Location Dropdown Change 

scottkicks:
# 📲 What

This PR makes sure that estimated shipping on reward cards is updated based on the selected shipping location. Specifically for when the reward has a "Ships anywhere in the world" preference _and_ has shipping rules set for specific countries.

# 🤔 Why

Existing logic was written assuming that unrestricted shipping rewards would only have shipping costs for the rest of the world. It turns out they can also have shipping costs for specific countries.

This means that a reward can have an "unrestricted" shipping preference with shipping rules for "Rest of World" _and_ other specific countries.

# 🛠 How

We need to check first to see if the reward has a shipping rule matching the selected location ID.
If it _doesn't _and_ the reward has an unrestricted shipping preference, use the shipping rule with an id of 1. This is always the  "Anywhere in the world" rule.
If that doesn't exist either, we know the reward has no shipping rule information.

# 👀 See

The test project I used has the following shipping config:
Screenshot 2024-11-18 at 12 34 12 PM

See that the estimated shipping values update as expected when the location changes.

Simulator Screen Recording - iPhone 15 Pro 17 5 - 2024-11-18 at 12 26 40

# ✅ Acceptance criteria

- [x] Estimated shipping on rewards cards matches the selected location or the costs associated with "Rest of World" shipping rules if the reward has unrestricted shipping

### DIFF
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -778,12 +778,18 @@ private func estimatedMinMax(
   var max: Double = 0
 
   rewards.forEach { reward in
-    guard reward.shipping.enabled, let shippingRules = reward.shippingRulesExpanded,
-          let shipping = shippingRules.first(where: { $0.location.id == locationId }) else { return }
+    guard reward.shipping.enabled,
+          let shippingRules = reward.shippingRulesExpanded ?? reward.shippingRules else { return }
+
+    var shipping: ShippingRule? = shippingRules.first(where: { $0.location.id == locationId })
+
+    if shipping == nil && reward.shipping.preference == .unrestricted {
+      shipping = shippingRules.first
+    }
 
     /// Verify there are estimated amounts greater than 0
-    guard let estimatedMin = shipping.estimatedMin?.amount,
-          let estimatedMax = shipping.estimatedMax?.amount,
+    guard let estimatedMin = shipping?.estimatedMin?.amount,
+          let estimatedMax = shipping?.estimatedMax?.amount,
           estimatedMin > 0 || estimatedMax > 0 else {
       return
     }
--- a/Library/SharedFunctions.swift
+++ b/Library/SharedFunctions.swift
@@ -778,22 +778,8 @@ private func estimatedMinMax(
   var max: Double = 0
 
   rewards.forEach { reward in
-    guard reward.shipping.enabled, let shippingRules = reward.shippingRules else {
-      return
-    }
-
-    var shippingRule: ShippingRule?
-
-    /// First check to see if the reward has a shipping rule that matches the selected location id.
-    /// If it doesn't AND the reward has an unrestricted shipping preference, grab the "Anywhere in the world' shipping rule by id. (id is always 1) .
-    /// Otherwise, we know that the reward doesn't have any shipping rule information associated with it.
-    if let matchingShippingRule = shippingRules.first(where: { $0.location.id == locationId }) {
-      shippingRule = matchingShippingRule
-    } else if reward.shipping.preference == .unrestricted {
-      shippingRule = shippingRules.first(where: { $0.location.id == 1 })
-    }
-
-    guard let shipping = shippingRule else { return }
+    guard reward.shipping.enabled, let shippingRules = reward.shippingRulesExpanded,
+          let shipping = shippingRules.first(where: { $0.location.id == locationId }) else { return }
 
     /// Verify there are estimated amounts greater than 0
     guard let estimatedMin = shipping.estimatedMin?.amount,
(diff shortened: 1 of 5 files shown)


Library/SharedFunctions.swift
```diff
@@ -778,16 +778,20 @@ private func estimatedMinMax(
   var max: Double = 0
 
   rewards.forEach { reward in
-    guard reward.shipping.enabled, let shippingRules = reward.shippingRules else { return }
+    guard reward.shipping.enabled, let shippingRules = reward.shippingRules else {
```
ifosli, line 781: Use `shippingRulesExpanded` here - this will have a shipping rule for each country. If `shippingRulesExpanded` isn't always available where we estimate shipping, we could use `shippingRules` as a backup for a "mostly working temp solution" but only if we're okay getting estimated shipping amounts wrong sometimes.

Library/SharedFunctions.swift
```diff
@@ -778,16 +778,20 @@ private func estimatedMinMax(
...
-      ? shippingRules.first
-      : shippingRules
-      .first(where: { $0.location.id == locationId })
+    /// First check to see if the reward has a shipping rule that matches the selected location id.
+    /// If it doesn't AND the reward has an unrestricted shipping preference, grab the "Anywhere in the world' shipping rule by id. (id is always 1) .
+    /// Otherwise, we know that the reward doesn't have any shipping rule information associated with it.
+    if let matchingShippingRule = shippingRules.first(where: { $0.location.id == locationId }) {
+      shippingRule = matchingShippingRule
+    } else if reward.shipping.preference == .unrestricted {
+      shippingRule = shippingRules.first(where: { $0.location.id == 1 })
```
ifosli, line 793: Once we switch to using `shippingRulesExpanded` we no longer need this line. (And this logic won't always work, which is why we needed to fetch expanded shipping rules for filtering anyways; ex shipping a reward within the EU when there's a shipping rule for all the EU will not show the correct amount.)

ifosli (changes requested):
Good job finding the problem! Let's just update this to use expanded shipping rules, so that matching by location id always works.

Library/SharedFunctions.swift
```diff
@@ -778,18 +778,8 @@ private func estimatedMinMax(
...
-
-    /// If the reward's shipping prefernce is Anywhere in the world, use its first and only shipping rule.
-    /// Else use the rule that  matches the selected shipping rule (from the locations dropdown).
-    shippingRule = reward.shipping.preference == .unrestricted
-      ? shippingRules.first
-      : shippingRules
-      .first(where: { $0.location.id == locationId })
-
-    guard let shipping = shippingRule else { return }
+    guard reward.shipping.enabled, let shippingRules = reward.shippingRulesExpanded,
```
ifosli, line 781: I know `shippingRulesExpanded` are fetched in the rewards carousel. Will you double check that this still works well anywhere else we use the helper function? For example, estimated shipping will show fine on the checkout page during checkout, but does it show on the "manage pledge" screen? And "update payment method" flow?

If it doesn't, we either need to make sure we start fetching `shippingRulesExpanded` (at least for the relevant `locationId`) or we could do a temporary `let shippingRules = reward.shippingRulesExpanded ?? reward.shippingRules` (which wouldn't work for the EU shipping rule).

ifosli (approved):
Looks good. Just please make sure you test all possible flows before submitting! I'm concerned that some of them will expect to show estimated shipping but will not already be fetching expanded shipping rules.

stevestreza-ksr (approved):
LGTM pending Ingerid's comments

Library/SharedFunctions.swift
scottkicks: @ifosli  So I actually found that we still need the logic that chooses the first shipping rule in the following case.

If a reward "Ships anywhere" and also has shipping rules for US and Japan, but the backer selects and then pledges with a different country like Vietnam for example, then on the manage pledge screen `shippingRulesExpanded` is nil. And since `shippingRules` won't contain a rule for Vietnam, we'd still need to grab the first shipping rule (if the preference is unrestricted) which will be the "Rest of World" rule.

Hopefully that makes sense. These edge cases are starting to get to me lol

ifosli: Yeah. In the long run, we need to fetch `shippingRulesExpanded` there as well (ex if shipping to the eu is cheaper than worldwide, and there's no specific rule for denmark, the current solution will say that estimated shipping for denmark is the worldwide shipping). But I'm okay with this as a temporary patch, since estimated shipping is just estimated anyways. Will you file a follow-up/tech debt ticket to make sure we fetch `shippingRulesExpanded` anywhere we need to calculate shipping for a specific location, though?

PR #2188, opened 2024-10-30
[MBL-1492] View backing details/survey webview action

ifosli:
# 📲 What

Wire up three separate navigation actions to the survey webview: "view backing details" (newly added), "edit address," and "complete survey." I'm matching android's behavior here and linking all these actions to the survey tab of the backing details webview. For a project without a survey, this will redirect to the backing details page instead.

Note: In order for this format of survey url to be visible in the webview, I needed to allowlist it. Instead of just adding `project/backing/survey_responses` to the list, I've updating our current urls of this type to simply be `project/backing/:backing_tab`. This will allow any current and future tabs on the backing page to be displayed in the webview.

Note: I did regenerate graphql-schema as part of this work, which accounts for the majority of the lines added/removed. I didn't look thoroughly into these graphql changes.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1492)

UI test that fakes the initial event that a card would send to the PPOViewModel:
Simulator Screen Recording - iPhone 16 - 2024-10-30 at 16 57 02

# ✅ Acceptance criteria

- [x] Survey webview opens when edit address, complete survey, or view backing details events are triggered
- [x] The survey webview loads the survey tab if available and defaults to the backing details tab if not
- [x] Backer can complete the survey and edit address through the webview when opened from PPO

# ⏰ TODO

-  We need to test complete edit address, complete survey, and view backing details flows once cards are ready as well.

PR #2201, opened 2024-11-19
[MBL-1829] Pledge Redemption Push Notification Support

scottkicks:
# 📲 What

Adds support for pledge redemption push notifications

# 🤔 Why

pledge redemption

# 🛠 How

backend push data structure can be seen here: https://kickstarter.atlassian.net/browse/BEE-1513

I referenced how we added support for survey push notifications here.

This is my first time adding PN support, and since I need to test this in either TestFlight or our Beta when it's pointing to production, I need to get this merged before I can do so.

My understanding of this task is that the pledge redemption PN can be opened in the survey web view, so that's what I'm doing here.

# ✅ Acceptance criteria

- [ ] App receives pledge redemption notifications
- [ ] Pledge redemption flow opens in a web view when being tapped.


KsApi/models/PushEnvelope.swift
```diff
@@ -34,6 +35,13 @@ public struct PushEnvelope {
     public let projectId: Int
   }
 
+  public struct PledgeRedemption {
+    public let id: Int
+    public let backingId: Int
```
scottkicks, line 40: I'm actually not sure that we need to decode backingId. Adding it for now while I wait for confirmation.

ifosli: Yeah, I don't think we need the backingId for the webview - my guess is this is included because we would've needed it for a native implementation.

Library/Navigation.swift
```diff
@@ -132,6 +132,7 @@ private let allRoutes: [String: (RouteParamsDecoded) -> Navigation?] = [
   "/signup": signup,
   "/projects/:creator_param/:project_param": project,
   "/projects/:creator_param/:project_param/backing/:backing_tab": projectSurvey,
+  "/projects/:creator_param/:project_param/backing/redeem": projectRedeem,
```
ifosli, line 135: We don't need this url added separately; it's already covered by line 134.

Library/Navigation.swift
```diff
@@ -400,6 +401,18 @@ private func pledgeNew(_ params: RouteParamsDecoded) -> Navigation? {
   return nil
 }
 
+private func projectRedeem(_ params: RouteParamsDecoded) -> Navigation? {
```
ifosli, line 404: Same here; since we handle all backings urls in the "survey webview", we should treat them all as `projectSurvey` urls. (The webview itself checks if a url is of type `projectSurvey` and only allows it to be loaded if it is.)

ifosli (commented):
Push notification itself looks good and looks like it's matching backend. We don't need to change anything in Navigation.swift, though!

Library/Navigation.swift
scottkicks: so ":backing_tab" and "redeem" result in the same url param?

Library/Navigation.swift
ifosli: The `:` makes it a parameter name. So /projects/[creator id or name]/[project id or slug]/backing/[any alphanumeric string] will be interpreted as a `projectSurvey` url and be opened in the webview. A malformed url (where the `backing_tab` is something that doesn't exist) would show a `404` in the webview - not ideal but something I'm okay with living with considering how many backing tabs we need to support. Let me know if you disagree, though!

(I updated this fairly recently, in https://github.com/kickstarter/ios-oss/pull/2188/files#diff-001c28f2becf20e1b939e1f60059a4a4dea67c0fc5bd21850ffd7213bfa10390 instead of adding a 3rd url - my hope was that this change would also mean that web will be able to make further changes to the survey without us needing to add local changes to support them. But I don't think it's even in prod yet, so happy to revisit if you'd rather we list the urls out)

ifosli (approved):

jovaniks (approved):
Nice, good job!

PR #2203, opened 2024-11-25
[MBL-1852] Update CreateBackingInput

scottkicks:
# 📲 What

Updates the `CreateBackingInput` GraphQL object to support its new `incremental: Bool?` property.

Actual implementation will be done separately in [mbl-1853](https://kickstarter.atlassian.net/browse/MBL-1853).

# 🤔 Why

This flag will let the backend know if backers have selected the Pledge Over Time option at checkout. 

# 🛠 How

- Updated GraphAPI.swift via our apollo script
- Updates the existing adapters and structs to include this new property.
- Passes `false` to call points for now since implementation will be done in [mbl-1853](https://kickstarter.atlassian.net/browse/MBL-1853)

# 👀 See

Tested that the new property is being sent as false in when attempting a crowdfund pledging:

Screenshot 2024-11-25 at 10 23 49 AM

# ✅ Acceptance criteria

- [x] All checkout flows work as expected
- [x] `incremental` is being added to the `CreateBackingInput` and is false for now.
- [x] Late pledge flow experiences no change and is unaware of this new property.

### DIFF
--- a/KsApi/mutations/inputs/CreateBackingInput.swift
+++ b/KsApi/mutations/inputs/CreateBackingInput.swift
@@ -16,7 +16,7 @@ public struct CreateBackingInput: GraphMutationInput, Encodable {
 
    - parameter amount: The amount.
    - parameter applePay: The optional ApplePayParams.
-   - parameter incremental: The optional Bool indicating whether pledge over time has been selected..
+   - parameter incremental: The optional Bool indicating whether pledge over time has been selected.
    - parameter locationId: The optional ID of the ShippingRule's Location.
    - parameter paymentSourceId: The optional ID of the PaymentSource.
    - parameter projectId: The GraphID of the Project.
--- a/KsApi/mutations/inputs/CreateBackingInput.swift
+++ b/KsApi/mutations/inputs/CreateBackingInput.swift
@@ -3,6 +3,7 @@ import Foundation
 public struct CreateBackingInput: GraphMutationInput, Encodable {
   let amount: String
   let applePay: ApplePayParams?
+  let incremental: Bool?
   let locationId: String?
   let paymentSourceId: String?
   let projectId: String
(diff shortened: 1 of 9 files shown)


KsApi/mutations/inputs/CreateBackingInput.swift
```diff
@@ -15,6 +16,7 @@ public struct CreateBackingInput: GraphMutationInput, Encodable {
 
    - parameter amount: The amount.
    - parameter applePay: The optional ApplePayParams.
+   - parameter incremental: The optional Bool indicating whether pledge over time has been selected..
```
ifosli, line 19: nit: extra `.`

Library/CreateBackingConstructorTests.swift
ifosli, line 1: Should there be a test where `incremental` is `true`?

ifosli (approved):
Looks good! Just one typo. We will want tests at some point where pledge over time is enabled, but I trust your judgement for whether that's something that should be tested now or if it should wait.

Library/CreateBackingConstructorTests.swift
scottkicks: Definitely. My plan is to add tests when I get going on the implementation piece.

Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -521,7 +521,8 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
         paymentSourceId: paymentSourceId,
         setupIntentClientSecret: nil,
         applePayParams: applePayParams,
-        refTag: refTag
+        refTag: refTag,
+        incremental: false // TODO: implementation in [mbl-1853](https://kickstarter.atlassian.net/browse/MBL-1853)
```
stevestreza-ksr, line 525: It looks like `incremental` is always false in this PR, is the idea to make it true sometimes in that ticket?

stevestreza-ksr (approved):

amy-at-kickstarter:
Re-read this PR and tested it in simulator; I think this was just left open by accident. I'll go ahead and merge.

PR #2204, opened 2024-11-26
[MBL-1871] Firebase consent signals

stevestreza-ksr:
# 📲 What

This PR adds Firebase consent signals to the existing app tracking and app launch pipeline.

# 🛠 How

Our existing ATT infrastructure already tracks the user's authorization status. This updates ATT to make a SignalProducer that emits the current value when observed + any changes. Then it connects that signal through AppDelegateViewModel to listen to app launch signals and update the Firebase consent settings based on the ATT authorization.

AppDelegateViewModel uses ReactiveSwift already, so that's what I went with.

# 👀 See

https://developers.google.com/tag-platform/security/guides/app-consent?platform=ios&consentmode=advanced

# ✅ Acceptance criteria

Allow app tracking and search the logs for `ad_`. You should see this log: `ad_storage is granted. analytics_storage is granted. ad_user_data is granted. ad_personalization is granted.`

Disallow app tracking and search the logs for `ad_`. You should see this log: `ad_storage is denied. analytics_storage is denied. ad_user_data is denied. ad_personalization is denied.`


nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/AppDelegate.swift:184:1 | warning: (redundantSelf) Insert/remove explicit self where applicable. |
| Kickstarter-iOS/AppDelegate.swift:410:1 | warning: (trailingCommas) Add or remove trailing comma from the last item in a collection literal. |
| Kickstarter-iOS/AppDelegateViewModel.swift:771:1 | warning: (braces) Wrap braces in accordance with selected style (K&R or Allman). |
| Kickstarter-iOS/AppDelegateViewModel.swift:773:1 | warning: (indent) Indent code in accordance with the scope level. |
| Kickstarter-iOS/AppDelegateViewModelTests.swift:86:1 | warning: (redundantSelf) Insert/remove explicit self where applicable. |
| Kickstarter-iOS/AppDelegateViewModelTests.swift:86:1 | warning: (wrap) Wrap lines that exceed the specified maximum width. |


</p>

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -388,6 +392,21 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     task.resume()
   }
 
+  private func updateFirebaseConsent(status: AppTrackingAuthorization) {
+    // https://developers.google.com/tag-platform/security/guides/app-consent?platform=ios&consentmode=advanced#set_up_consent_mode
+    let consentStatus: ConsentStatus = if case .authorized = status {
+      .granted
+    } else {
+      .denied
+    }
```
scottkicks: Nice. I appreciate the link to the docs. Also, still getting used to reading this new `if case` syntax!

Library/TestHelpers/MockAppTrackingTransparency.swift
```diff
@@ -1,19 +1,45 @@
...
   }
 
   func shouldRequestAuthorizationStatus() -> Bool {
     self.shouldRequestAuthStatus
   }
+
+  private var authorizationStatusProperty = MutableProperty<AppTrackingAuthorization>(.notDetermined)
+  var authorizationStatus: SignalProducer<AppTrackingAuthorization, Never> {
+    self.authorizationStatusProperty.producer
+      .replayLazily(upTo: 1)
```
scottkicks, line 42: how does this work? I've never seen .replayLazily

scottkicks (approved):
LGTM 👍 

FYI Ingerid is out for the rest of the week. If this needs to get merged this week you might want to request review from Jovani for a 2nd approval.

Library/TestHelpers/MockAppTrackingTransparency.swift
stevestreza-ksr: The behavior is that when someone observes the values from this observable thing, they should get the current authorization value immediately and then any changes later.

With ReactiveSwift, a Signal is "hot", meaning that the pipeline only runs once, and everyone observing it gets the same values at the same time. That means that you can't have a "return the current value immediately" behavior using a Signal, because observing doesn't start anything, and since every observer gets the same values, subscribing anywhere would need to re-emit the current value to all the other observers who wouldn't need it.

Instead this is a SignalProducer, which is "cold", meaning that the pipeline doesn't start until you observe it, and every observation gets a separate pipeline. This means you CAN have a "return the current value immediately" behavior, because when you observe a SIgnalProducer, it creates a Signal, making it "hot" for the lifetime of its Disposable, and that can emit a value immediately.

Sometimes you want a hybrid of these approaches, where one SignalProducer acts like a Signal and runs a single pipeline with values shared to multiple observers, but where you can also modify the pipeline for each individual observer. `replayLazily(upTo:)` does this by buffering up to a certain number of values, returning those immediately to new observers. And then any values after that are shared to all observers.

So that's what's going on here. When someone observes this initially, they get the current buffered value from the property, and then any subsequent values.

Kickstarter-iOS/AppDelegate.swift
stevestreza-ksr: It's functionally the same as:

```swift
let consentStatus: ConsentStatus
if case .authorized = status {
  consentStatus = .granted
} else {
  consentStatus = .denied
}
```

I think it's a little cleaner because it means that the variable is never in that weird in-between uninitialized state, but if we don't like it we can always undo these or make a formatting rule for them.

ifosli (approved):

PR #2207, opened 2024-12-02
[MBL-1905] Revert "[MBL-1613]"

scottkicks:
# 📲 What

Reverts https://github.com/kickstarter/ios-oss/pull/2171

# 🤔 Why

As our estimated shipping flag has been rolling out, a backer experienced an error when attempting to late pledge with Apple Pay. They were shown an error banner and weren't able to retry their pledge.

This looks to be an error where re-creating a payment intent is failing because some backend state changed.
See original thread between mobile and payments for more context: https://kickstarter.slack.com/archives/C3LS41US0/p1732652825528529

# 🛠 How

By reverting these changes, the checkout screen will dismiss forcing a new checkout object to be created for the user so that a new payment intent can be created and validated when there is an error like this.

# ✅ Acceptance criteria

Tested this by manually forcing validate checkout and onCompleteSessionCheckout to fail on Staging. Testing the Apple Pay flow on Prod to verify that is working as expected will be important also.

- [x] Validate Checkout and Checkout errors show the error banner
- [x] The checkout screen dismisses when after the error banner is shown
- [x] Attempting to pledge after this error scenario results in a successful attempt when there are not validation or checkout errors.

# ⏰ TODO

- [ ] Test Apple Pay flow on Test Flight


Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
```diff
@@ -456,7 +457,6 @@ public class NoShippingPostCampaignCheckoutViewModel: NoShippingPostCampaignChec
 
     let completeCheckoutWithCreditCardInput: Signal<GraphAPI.CompleteOnSessionCheckoutInput, Never> = Signal
       .combineLatest(self.confirmPaymentSuccessfulProperty.signal.skipNil(), checkoutId, selectedCard)
-      .takeWhen(self.confirmPaymentSuccessfulProperty.signal.skipNil())
```
ifosli, line 459: I think we should still keep this `takeWhen` - it's probably not needed since a user shouldn't be able to change the selected card again once we dismiss this vc if checkout fails, but having "change selected card" kick off checkout is not a risk I'd like to take.

ifosli (approved):
Looks good; just one concern

Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPostCampaignCheckoutViewController.swift
```diff
@@ -306,6 +306,21 @@ final class NoShippingPostCampaignCheckoutViewController: UIViewController,
         let thanksVC = ThanksViewController.configured(with: thanksPageData)
         self?.navigationController?.pushViewController(thanksVC, animated: true)
       }
+
+    self.viewModel.outputs.checkoutError
+      .observeForUI()
+      .observeValues { [weak self] error in
+
+        #if DEBUG
+          let serverError = error.errorMessages.first ?? ""
```
stevestreza-ksr, line 315: Nit: Maybe put a default value (`Unknown server error` or something) instead of an empty string here, since this is for debugging

stevestreza-ksr (approved):

ifosli (approved):

PR #2210, opened 2024-12-03
[MBL-1905] Create One Payment Intent Per Late Pledge Checkout 

scottkicks:
# 📲 What

Makes sure that we only call `CreatePayment` Mutation once in our late pledge checkout

# 🤔 Why

As our estimated shipping flag has been rolling out, a backer experienced an error when attempting to late pledge with Apple Pay. They were shown an error banner and weren't able to retry their pledge.

This looks to be an error where re-creating a payment intent is failing because some backend state changed.
See original thread between mobile and payments for more context: https://kickstarter.slack.com/archives/C3LS41US0/p1732652825528529

One thing we can do to make sure we avoid conflicting Payment Intent creation calls is only calling .createPaymentIntent once, instead of for existing payment methods _and_ apple pay.

# How

Instead of calling `apiService.createPaymentIntent` for exiting CC and when Apple Pay is tapped, I've setup a variable that calls this once when checkout loads and then updated both flows to use the that payment intent.

If creating a new payment intent fails, the error banner is shown and checkout is dismissed. Similar to how we handle validate and checkout errors. Since checkout rely on a payment intent in both cases, if we fail to make one we'll need to try again before checkout can completed. 

# ✅ Acceptance criteria

- [x] All NoShippingPostCampaignViewModel tests pass
- [x] Late pledge with an existing cc succeeds
- [x]  Late pledge with an new cc succeeds
- [x]  If creating a payment intent fails an error banner is shown and checkout is dismissed
- [ ]  Late pledge with apple pay succeeds (Test on TestFlight)


amy-at-kickstarter:
Can you throw in a test to validate this bug?

Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
```diff
@@ -290,21 +290,14 @@ public class NoShippingPostCampaignCheckoutViewModel: NoShippingPostCampaignChec
...
-
-    // MARK: Validate Existing Cards
-
-    let newPaymentIntentForExistingCards = Signal.combineLatest(
+    let paymentIntentEvent = Signal.combineLatest(
       project,
       pledgeTotal,
       checkoutId,
       backingId
     )
```
amy-at-kickstarter, line 300: Just getting my brain back into things - is one of these linked to a `viewDidLoad` event? It would be good to guarantee that this can still only be called once per page load, and none of these properties make that obvious.

ifosli: All 4 of these fields should only have one value; `checkoutId`/`backingId` are generated by createCheckout (which is called on load) and `pledgeTotal` is calculated on load as well. However, this class is a bit of a mess- `pledgeTotal` here is also parsed from the backing, if there is any (which there can't be as it's a late pledge...), so it's definitely not clear from the signals that (if) `pledgeTotal` is only set once. So this should happen exactly once, but I agree that it'd be worth making that more clear and cleaning this class up a bit.

At a minimum, we may want to add some `skipNil`s (if needed - I can't tell from a quick look at the code if these are nullable) and a `takeFirst`, either to this signal or to its inputs.

ifosli (approved):
I think this should work. It'd be nice to make the code a little clearer, though. I think the initial paymentIntentEvent was created when all its inputs were calculated by the confirm details page, so it's probably worth adding a guarantee here that they can't be changed once set.

Library/ViewModels/NoShippingPostCampaignCheckoutViewModel.swift
amy-at-kickstarter: I like the idea of adding a `takeFirst`, just to make it very explicit!

scottkicks: None of these signals emit optional values so I don't think we need any `skipNil`s here. I'm not super familiar with how `.take(first:)` functions, but adding it directly to this signal blocks it from completing. 

Do either of you know more about why that is? `checkoutId/backingId` rely on the CreateCheckout API call and the others are parsed out from the `PledgeViewData` that this VM is configured with. I'm not seeing where this would be called multiple times.

amy-at-kickstarter: Looks like `.take(first:)` means
> Take up to `n` values from the signal and then complete.
so that should be doing what we want (unless the completion itself is an issue). What happens if you try taking the first 2 or 3 etc?

ifosli: That's weird. Does `take(first: 1)` not work but `take(first: 2)` does? That definitely sounds like something getting called twice that shouldn't be (though I agree with Amy that it's possible that completing the signal itself is causing issues somehow). Let me know if you want help debugging this! I definitely don't think we should leave a `2` here without a comment explaining why; that's more misleading than not including a `take(first:)` at all.

scottkicks: Also, my [test](https://github.com/kickstarter/ios-oss/pull/2210/commits/8f1dcc3a073fbef25d453e72f30dfaeff5157f97) asserts that only one payment intent request is being emitted. I don't see why `take(first: 1)` isn't working, tbh. My breakpoints/debugger aren't telling me much, unfortunately.

ifosli: I logged the events to look at them and here's what I think happens: `take(first: 1)` does exactly what it says it will, but the signal is completed, terminated, and disposed after. So the client secret is fetched and does exist, but it doesn't stick around. So in the existing card flow, for example, we see
```
let paymentIntentClientSecretForExistingCards = paymentIntentClientSecret
      .takeWhen(selectedCard)
```
but the `paymentIntentClientSecret` signal doesn't exist anymore when that card is selected. (I looked this up and `take(first:}` literally works by sending the completion event early.

I'm not sure if there's something similar to `take(first: 1)` that we can use instead that doesn't complete, but I definitely didn't find anything, so I think this is fine without it.

ifosli (approved):

PR #2218, opened 2024-12-10
[MBL-1930] Delete ppoHasAction from UserFragment

ifosli:
# 📲 What

The `ppoHasAction` field caused basically all graphQL fetches to fail for logged out users. (They probably only fail if we're fetching a user, but ex fetching a project includes a user fetch because of the project creator.) In order to unblock the build, we'll just delete this field from graphQL.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1930)
[Jira for next steps (currently blocked since the notification dot is broken)](https://kickstarter.atlassian.net/browse/MBL-1931)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro Max - 2024-12-10 at 15 28 31 | Simulator Screen Recording - iPhone 16 Pro Max - 2024-12-10 at 15 26 37 |

# ✅ Acceptance criteria

- [x] Project pages load normally for logged out users

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -20206,7 +20206,6 @@ public enum GraphAPI {
           topic
         }
         optedOutOfRecommendations
-        ppoHasAction
         showPublicProfile
         savedProjects {
           __typename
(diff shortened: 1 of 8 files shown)


scottkicks (approved):
Is this field not being used anywhere? Or is the plan to use it in upcoming PPO work that we haven't gotten to yet?

ifosli:
> Is this field not being used anywhere? Or is the plan to use it in upcoming PPO work that we haven't gotten to yet?

It is being used (only for ppo). But most of the time we fetch it through the v1 API, which still works even with this change. I'm not sure if we need to fetch it through graphQL at all (even with the PPO flag on), so I filed a ticket to look into that. It's currently blocked since the `ppoHasAction` is broken right now (always returns false). So these changes might break the PPO notification dot a little for logged-in users, but there's already a new ticket to figure that out that's it's unblocked. (PPO is not launched yet for the record.)

jovaniks (approved):
Perfect 💯

PR #2224, opened 2024-12-17
Remove generic Constants file that was only used in one place

amy-at-kickstarter:
# 📲 What

Remove a little-used file.

# 🤔 Why

This came up in a PR for PLOT - @jovaniks made a class named `Constants` and I noticed the overlap. This old `Constants` was only used in one place and never expanded upon; removing it is just a teeny step towards clearer code. I don't feel bad about hard-coding a single image size in a single mutation.


ifosli (approved):
Nice!

scottkicks (approved):

jovaniks (approved):

PR #2225, opened 2024-12-19
MBL-1838: Wire up PLOT query to UI in no-shipping checkout flow

amy-at-kickstarter:
# 📲 What

Hook up the `BuildPaymentPlanQuery` to the new PLOT UI.

# 🤔 Why

This means PLOT will start showing live data, instead of mock data.

# 🛠 How

* I created some helpers (plus tests) to map between the GraphAPI response object and @jovaniks view model objects. While I didn't write an intermediary envelope object, those helpers serve a similar role. As an aside, I've been noticing that Apollo/GraphQL objects are painful to test, which is a downside to avoiding envelope objects. 
* I had to update our ISO8061 date formatter to work in both directions, since we need to parse dates.
* I removed some fields we weren't using from the `PledgePaymentIncrement` fragment.

# 👀 See
https://github.com/user-attachments/assets/d7eb3348-d5ce-4b18-b99a-2a7c03924b25

# 📓  Todos
- [ ] Fix failing screenshot tests
- [ ] Discuss loading states. Do we want to block pledge until the PLOT module loads? What should it look like while it's loading?

### DIFF
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPledgeViewControllerTests.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPledgeViewControllerTests.swift
@@ -298,8 +298,14 @@ final class NoShippingPledgeViewControllerTests: TestCase {
   }
 
   func testView_ShowCollectionPlans() {
-    let response = UserEnvelope<GraphUser>(me: self.userWithCards)
-    let mockService = MockService(fetchGraphUserResult: .success(response))
+    let userResponse = UserEnvelope<GraphUser>(me: self.userWithCards)
+    let paymentPlanResponse = try! GraphAPI.BuildPaymentPlanQuery
+      .Data(jsonString: buildPaymentPlanQueryJson_Eligible)
+    let mockService = MockService(
+      buildPaymentPlanResult: .success(paymentPlanResponse),
+      fetchGraphUserResult: .success(userResponse)
+    )
+
     let project = Project.template
       |> \.availableCardTypes .~ [CreditCardType.discover.rawValue]
       |> Project.lens.isPledgeOverTimeAllowed .~ true
@@ -346,8 +352,14 @@ final class NoShippingPledgeViewControllerTests: TestCase {
   }
 
   func testView_ShowCollectionPlans_Ineligible() {
-    let response = UserEnvelope<GraphUser>(me: self.userWithCards)
-    let mockService = MockService(fetchGraphUserResult: .success(response))
+    let userResponse = UserEnvelope<GraphUser>(me: self.userWithCards)
+    let paymentPlanResponse = try! GraphAPI.BuildPaymentPlanQuery
+      .Data(jsonString: buildPaymentPlanQueryJson_Ineligible)
+    let mockService = MockService(
+      buildPaymentPlanResult: .success(paymentPlanResponse),
+      fetchGraphUserResult: .success(userResponse)
+    )
+
     let project = Project.template
       |> \.availableCardTypes .~ [CreditCardType.discover.rawValue]
       |> Project.lens.isPledgeOverTimeAllowed .~ true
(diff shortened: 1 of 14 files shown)


KsApi/queries/BuildPaymentPlanQuery.graphql
```diff
@@ -8,12 +8,7 @@ query BuildPaymentPlan($slug: String!, $amount: String!) {
           amount
           currency
         }
-        id
```
amy-at-kickstarter, line 11: We never used these fields, cleaned them up to make testing slightly easier on myself.

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -625,3 +626,20 @@ private func distinctRewards(_ rewards: [Reward]) -> [Reward] {
     return !rewardIds.contains(reward.id)
   }
 }
+
+// TODO: Remove this when implementing the API  [MBL-1851](https://kickstarter.atlassian.net/browse/MBL-1851)
```
amy-at-kickstarter, line 630: Moved this and changed the ticket number. It's no longer needed in `NoShippingPledgeViewModel`, but still needed here in `ManagePledgeViewModel`.

Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -987,32 +987,53 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
...
+      > in
+        let nf = NumberFormatter()
+        let amount = nf.string(from: NSNumber(value: pledgeTotal)) ?? ""
+        return AppEnvironment.current.apiService.buildPaymentPlan(
+          projectSlug: project.slug,
+          pledgeAmount: amount
+        ).materialize()
+      }
+
+    self.showPledgeOverTimeUI = Signal.merge(
```
amy-at-kickstarter, line 1012: @jovaniks My assumption is that, since this feature is so small, there's no designs for a separate error state. Is that correct? My approach was just to hide the entire element if an error occurs in the `BuildPaymentPlan` query.

Library/ViewModels/NoShippingPledgeViewModelTests.swift
```diff
@@ -5285,4 +5287,97 @@ final class NoShippingPledgeViewModelTests: TestCase {
       self.showPledgeOverTimeUI.assertValues([false])
     }
   }
+
+  func testPledgeOverTimeConfigData_LoadsAfterBuildPledgeQuery() {
```
amy-at-kickstarter, line 5291: These tests are pretty bare-bones. I'd love to test that the mock query is actually making a GraphQL request for the right amount, but I don't think we have that capability with our current mock infrastructure.

Library/ViewModels/NoShippingPledgeViewModel.swift
jovaniks: That’s correct, we don’t have error handling planned for this component at the moment. Do you think it would make sense to add a `TODO` to revisit this discussion once everyone is back in January

amy-at-kickstarter: I'll set myself a reminder in Slack for January! And I'll raise it in the next PLOT meeting.

KsApi/queries/BuildPaymentPlanQuery.graphql
scottkicks: We should keep `state`. We'll use that to display the correct "badge" component on the view pledge details screen.

Good call with removing the others. I _pretty sure_ we won't need `stateReason` either for now.

[Designs](https://www.figma.com/design/vyTjzt9Tb67oHYUVbV2zKh/PLOT-V1-Designs?node-id=1039-17930&t=bJD5dny7BysKQLrz-4)

Screenshot 2025-01-02 at 9 37 39 AM

Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -987,32 +987,53 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
...
 
-    self.pledgeOverTimeConfigData = Signal.combineLatest(
-      self.showPledgeOverTimeUI,
-      project,
-      pledgeTotal
+    let pledgeOverTimeQuery = Signal.combineLatest(project, pledgeTotal)
+      // Only make the query when PLOT is enabled
+      .filterWhenLatestFrom(pledgeOverTimeUIEnabled, satisfies: {
+        $0 == true
+      })
```
scottkicks, line 997: love how verbose this filter is

Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -987,32 +987,53 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
...
+      // Only call the query once
+      .take(first: 1)
+      .switchMap { (project: Project, pledgeTotal: Double) -> SignalProducer<
+        Signal<GraphAPI.BuildPaymentPlanQuery.Data, ErrorEnvelope>.Event,
+        Never
+      > in
+        let nf = NumberFormatter()
```
scottkicks, line 1004: minor nit: `formatter` or `numFormatter` could be a better name.

Kickstarter-iOS/Features/PledgeView/Controllers/NoShippingPledgeViewControllerTests.swift
```diff
@@ -393,3 +405,51 @@ final class NoShippingPledgeViewControllerTests: TestCase {
       }
   }
 }
+
+private let buildPaymentPlanQueryJson_Eligible = """
```
amy-at-kickstarter, line 409: Of note: if we upgrade Apollo, we'll get access to their mocks API. Then it will be really, really easy to mock GraphQL response objects in our tests without all of this verbose json nonsense.

KsApi/queries/BuildPaymentPlanQuery.graphql
amy-at-kickstarter: Noted! I'll put state back.

jovaniks (approved):
Nice work!

PR #2223, opened 2024-12-16
[MBL-1899] Checkout screen PLOT Badge

jovaniks:
# 📲 What

This PR adds a **Pledge Over Time (PLOT) badge** under the "Your pledge" section when PLOT is selected in the collection plan. It also updates the payment label and status label copy for PLOT-enabled pledges.

# 🤔 Why

Some background context on why the change is needed.

# 🛠 How

- Added a **Pledge Over Time badge** to be displayed under the "Your pledge" section.
- Updated the **payments label** to show "charged as 4 payments."
- Updated the **status label copy** to display dynamic text:

   > "If the project reaches its funding goal, the first charge of $20 will be collected on March 15, 2024."

# 👀 See

- [MBL-1899](https://kickstarter.atlassian.net/browse/MBL-1899)

Simulator Screen Recording - iPhone SE (3rd generation) - 2024-12-16 at 10 19 42

# ✅ Acceptance criteria

- [ ] The **Pledge Over Time badge** is displayed under "Your pledge" when PLOT is selected.
- [ ] The payments label reads: "charged as 4 payments."
- [ ] The status label copy displays the correct amount and scheduled collection date
    - Example: "If the project reaches its funding goal, the first charge of $20 will be collected on March 15, 2024."

# ⏰ TODO

- [ ] Using real data from API when availble ([MBL-1838](https://kickstarter.atlassian.net/browse/MBL-1838))

[MBL-1899]: https://kickstarter.atlassian.net/browse/MBL-1899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MBL-1838]: https://kickstarter.atlassian.net/browse/MBL-1838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2226, opened 2025-01-03
Move PLOT logic out of NoShippingPledgeViewModel and into its own VM

amy-at-kickstarter:
# 📲 What

Move all PLOT logic out of `NoShippingPledgeViewModel` and into its own view model, `PLOTPledgeViewModel`.

# 🤔 Why

The complexity of checkout is a significant thorn in our side. Moving the PLOT logic out of the main pledge VM and into its own accomplishes three things:

1. Moves us a small step closer to checkout being modular, shorter, and easier to understand.
2. Allows us to integrate PLOT into other branches of checkout, if necessary.
3. Makes PLOT easier to test.

# 🛠 How

The PLOT VM is fairly simple - this is mostly copy-pasted from `NoShippingPledgeViewModel`. It follows the same structure as our other view models, with one notable exception: it takes its inputs as signals passed during `init`, instead of having an input function like a view model linked directly to a View Controller. 

# 👀 See

Implementing this allowed me to uncover a gnarly little bug in PLOT, which I've called out below!

### DIFF
--- a/Library/ViewModels/NoShippingPledgeViewModel.swift
+++ b/Library/ViewModels/NoShippingPledgeViewModel.swift
@@ -987,53 +987,7 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
 
     // MARK: Pledge Over Time
 
-    let pledgeOverTimeUIEnabled = project.signal
-      .map { ($0.isPledgeOverTimeAllowed ?? false) && featurePledgeOverTimeEnabled() }
-
-    let pledgeOverTimeQuery = Signal.combineLatest(project, pledgeTotal)
-      // Only make the query when PLOT is enabled
-      .filterWhenLatestFrom(pledgeOverTimeUIEnabled, satisfies: {
-        $0 == true
-      })
-      // Only call the query once
-      .take(first: 1)
-      .switchMap { (project: Project, pledgeTotal: Double) -> SignalProducer<
-        Signal<GraphAPI.BuildPaymentPlanQuery.Data, ErrorEnvelope>.Event,
-        Never
-      > in
-        let amountFormatter = NumberFormatter()
-        let amount = amountFormatter.string(from: NSNumber(value: pledgeTotal)) ?? ""
-        return AppEnvironment.current.apiService.buildPaymentPlan(
-          projectSlug: project.slug,
-          pledgeAmount: amount
-        ).materialize()
-      }
-
-    self.showPledgeOverTimeUI = Signal.merge(
-      // Hide PLOT if the feature flag is off on either client or server
-      pledgeOverTimeUIEnabled,
-      // Hide PLOT if an error occurs
-      pledgeOverTimeQuery.errors().map(value: false)
-    )
-
-    self.pledgeOverTimeConfigData = pledgeOverTimeQuery
-      .values()
-      .compactMap { $0.project?.paymentPlan }
-      .combineLatest(with: project)
-      .map { paymentPlan, project in
-
-        // TODO: Temporary placeholder to simulate the ineligible state for plans.
-        // The `thresholdAmount` will be retrieved from the API in the future.
-        // See [MBL-1838](https://kickstarter.atlassian.net/browse/MBL-1838) for implementation details.
-        let thresholdAmount = 125.0
-
-        return PledgePaymentPlansAndSelectionData(
-          withPaymentPlanFragment: paymentPlan,
-          selectedPlan: .pledgeInFull,
-          project: project,
-          thresholdAmount: thresholdAmount
-        )
-      }
+    self.plotViewModel = PLOTPledgeViewModel(project: project, pledgeTotal: pledgeTotal)
   }
 
   // MARK: - Inputs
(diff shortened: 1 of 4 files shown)


Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -1157,11 +1111,20 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
...
+  }
+
+  public var pledgeOverTimeConfigData: Signal<PledgePaymentPlansAndSelectionData, Never> {
+    return self.plotViewModel.outputs.pledgeOverTimeConfigData
+  }
 
   public var inputs: NoShippingPledgeViewModelInputs { return self }
   public var outputs: NoShippingPledgeViewModelOutputs { return self }
+
+  // MARK: - Internal view models
```
amy-at-kickstarter, line 1125: There's probably a better name for these. Component view models? Sub view models?

Library/ViewModels/PLOTPledgeViewModel.swift
```diff
@@ -0,0 +1,73 @@
...
+      // Only call the query once
+      .take(first: 1)
+      // Only make the query when PLOT is enabled
+      .filterWhenLatestFrom(pledgeOverTimeUIEnabled, satisfies: {
+        $0 == true
+      })
+      .map { (project: Project, pledgeTotal: Double) -> (String, String) in
+        let amountFormatter = NumberFormatter()
+        amountFormatter.minimumFractionDigits = 2
+        amountFormatter.maximumFractionDigits = 2
```
amy-at-kickstarter, line 28: Added this when I moved the code over. We weren't correctly formatting the cents value of the pledge total.

Library/ViewModels/PLOTPledgeViewModel.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+import KsApi
+import ReactiveSwift
+
+public protocol PLOTPledgeViewModelOutputs {
+  var showPledgeOverTimeUI: Signal<Bool, Never> { get }
+  var pledgeOverTimeConfigData: Signal<PledgePaymentPlansAndSelectionData, Never> { get }
+
+  // Visible only for testing
+  var buildPaymentPlanInputs: Signal<(String, String), Never> { get }
```
amy-at-kickstarter, line 10: Exposing this makes it really easy to test that we're calling `BuildPaymentPlanQuery` with the correct input values.

Library/ViewModels/PLOTPledgeViewModelTests.swift
```diff
@@ -0,0 +1,72 @@
...
+  let buildPaymentPlanInputs = TestObserver<(String, String), Never>()
+
+  override func setUp() {
+    super.setUp()
+
+    self.vm = PLOTPledgeViewModel(project: self.projectSignal, pledgeTotal: self.pledgeTotalSignal)
+    self.vm!.outputs.buildPaymentPlanInputs.observe(self.buildPaymentPlanInputs.observer)
+  }
+
+  func testViewModel_callsBuildPaymentPlanQuery_withCorrectSlugAndPledgeTotal() {
```
amy-at-kickstarter, line 23: Test for the bug I found.

Library/ViewModels/NoShippingPledgeViewModel.swift
ifosli: I like "component view models" but I think "internal view models" works okay, too.

Library/ViewModels/PLOTPledgeViewModel.swift
ifosli: Just wanted to confirm we should always be using cents here? I have a vague memory of there being some discussion around currencies that don't have cents (ex JPY) and how we want to handle those. If this is consistent with how we already display pledge totals elsewhere, I'm good with keeping it!

ifosli (approved):
Nice work! I like the overall strategy a lot! Just one concern around cents, but I'll let you and Scott figure that out.

Library/ViewModels/PLOTPledgeViewModel.swift
amy-at-kickstarter: In this case, I believe it would be harmless - if you had a pledge of JPY it would send "100.00" instead of "100" as the string.

I may actually have spoken too soon that this was a real bug - is it even _possible_ to have a pledge total with some cents amount? It looks like we always make pledges whole dollar/currency amounts anyways, although it's possible that shipping changes that.

But either way, still harmless!

scottkicks (approved):
love it!

jovaniks (approved):
Nice, I need these changes in my [PR](https://github.com/kickstarter/ios-oss/pull/2223) because `pledgeOverTimeConfigData` must emit values even when Pledge Over Time is not enabled. This is necessary as the confirmation text label depends on the selected payment plan (none, pledge in full, or plot).

amy-at-kickstarter:
@jovaniks Noted. Just merged, let me know if you need me to make that change (or if you have code you made already in your PR that you want me to integrate.)

PR #2230, opened 2025-01-06
MBL-1853: Send 'incremental' in CreateBacking if PLOT is selected

amy-at-kickstarter:
# 📲 What

Set `incremental = true` in `CreateBacking` in `NoShippingPledgeViewModel` if pledge over time is selected. 

# 🤔 Why

Allows us to actually create a pledge-over-time pledge!

# 🛠 How

This change adds another required signal to `CreateBacking`. I wanted to be sure that the signal was pretty well understood - and that, in all the failure modes I could think of, it would default to `.pledgeInFull`.

Because of that, I did some refactoring in `PLOTPledgeViewModel` to simplify how and when the config data was created. See comments inline.

### DIFF
--- a/Library/ViewModels/NoShippingPledgeViewModel.swift
+++ b/Library/ViewModels/NoShippingPledgeViewModel.swift
@@ -486,7 +486,14 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
       applePayParams.wrapInOptional()
     )
 
+    // MARK: Pledge Over Time
+
+    self.plotViewModel = PLOTPledgeViewModel(project: project, pledgeTotal: pledgeTotal)
+
     // MARK: - Create Backing
+
+    let selectedPaymentPlan = self.plotViewModel.pledgeOverTimeConfigData
+      .map { $0?.selectedPlan ?? .pledgeInFull }
 
     let createBackingData = Signal.combineLatest(
       project,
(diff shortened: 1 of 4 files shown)


Library/ViewModels/NoShippingPledgeViewModel.swift
```diff
@@ -486,8 +486,15 @@ public class NoShippingPledgeViewModel: NoShippingPledgeViewModelType, NoShippin
       applePayParams.wrapInOptional()
     )
 
+    // MARK: Pledge Over Time
+
+    self.plotViewModel = PLOTPledgeViewModel(project: project, pledgeTotal: pledgeTotal)
```
amy-at-kickstarter, line 491: Moving higher up again.

Library/ViewModels/NoShippingPledgeViewModelTests.swift
```diff
@@ -5337,7 +5337,7 @@ final class NoShippingPledgeViewModelTests: TestCase {
 
       self.showPledgeOverTimeUI.assertValues([true])
       self.pledgeOverTimeConfigData.assertDidEmitValue()
-      self.plotSelectedPlan.assertValues([.pledgeOverTime])
+      self.plotSelectedPlan.assertValues([.pledgeInFull, .pledgeOverTime])
```
amy-at-kickstarter, line 5340: @jovaniks I did a little refactoring so that it will send a default `pledgeInFull` right after it loads. Is this going to cause any problems on your end?

jovaniks: Your refactor looks super cool! 🤩 I ran a couple of tests, and everything works perfectly on my end. You're a rockstar! 🚀🎉

jovaniks (approved):
LGTM

PR #2233, opened 2025-01-08
[MBL-1878] Add GraphAPI.MoneyFragment currency formatter

stevestreza-ksr:
# 📲 What

Following up on PR comments in #2198, this PR adds a currency formatter to Format.swift to turn a `GraphAPI.MoneyFragment` into a localized string. 

# 🤔 Why

Being able to format the `GraphAPI.MoneyFragment` in a consistent and tested way is good

# 🛠 How

A new API was added which takes a `GraphAPI.MoneyFragment` and formats it using a `NumberFormatter` similar to our other currency formatters. Tests were added to generate a few strings from different fragments in different currencies and locales.

This API is nullable, because if the fragment is invalid, we should not show an incorrect string. Since this is the case, it made sense to update how this was used in backings dashboard. This check was moved to the `PPOProjectCardModel` constructor which already turns a bunch of other GraphQL data into the model object along with validation. Then the site where it's used already knows this field is valid.

# 👀 See

See #2198 for comments on why this ticket was created. 

# ♿️ Accessibility 

N/A, nothing user-facing changes

# ✅ Acceptance criteria

Tests should pass.

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
@@ -366,7 +366,7 @@ extension PPOProjectCardModel {
     let projectName = ppoProject?.name
     let projectId = ppoProject?.pid
     let pledgeFragment = backing?.amount.fragments.moneyFragment
-    let formattedPledge = pledgeFragment.flatMap({ Format.currency($0) })
+    let formattedPledge = pledgeFragment.flatMap { Format.currency($0) }
     let creatorName = ppoProject?.creator?.name
 
     let address: String? = backing?.deliveryAddress.flatMap { deliveryAddress in
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
@@ -221,7 +221,7 @@ extension PPOProjectCardModel {
       .init(type: .warning, icon: .time, message: "Address locks in 8 hours")
     ],
     image: .network(URL(string: "https:///")!),
-    projectName: "Sugardew Island - Your cozy farm shop let's pretend this is a way way way longer title",
+    projectName: "Sugardew Island - Your cozy farm shop let’s pretend this is a way way way longer title",
     projectId: 12_345,
     pledge: "$50.00",
     creatorName: "rokaplay truncate if longer than",
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
@@ -9,7 +9,7 @@ public struct PPOProjectCardModel: Identifiable, Equatable, Hashable {
   public let image: Kingfisher.Source
   public let projectName: String
   public let projectId: Int
-  public let pledge: GraphAPI.MoneyFragment
+  public let pledge: String
   public let creatorName: String
   public let address: String?
   public let actions: (Action, Action?)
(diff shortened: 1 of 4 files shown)


Library/FormatTests.swift
```diff
@@ -357,6 +357,58 @@ final class FormatTests: TestCase {
...
+    withEnvironment(locale: Locale(identifier: "en-GB")) {
+      let fragment = GraphAPI.MoneyFragment(amount: "1000.00", currency: .gbp, symbol: "£")
+      XCTAssertEqual(Format.currency(fragment), "£1,000.00")
+    }
+  }
+
+  func testCurrencyMoneyFragment_JPY() {
+    withEnvironment(locale: Locale(identifier: "ja")) {
+      let fragment = GraphAPI.MoneyFragment(amount: "1000.00", currency: .jpy, symbol: "¥")
+      XCTAssertEqual(Format.currency(fragment), "¥1,000.00")
```
amy-at-kickstarter, line 384: JPY should always format [with no decimal value](https://en.wikipedia.org/wiki/ISO_4217).

amy-at-kickstarter (changes requested):
I love this - big improvement, and it would be great to handle all `Money` fragments in the same way.

One requested change, which is fixing the decimal value for JPY. IIRC if you set the `currency` and `numberStyle = .currency` you should get the correct number of decimals for free.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift:369:1 | warning: (trailingClosures) Use trailing closure syntax where applicable. |
| Library/Format.swift:423:1 | warning: (indent) Indent code in accordance with the scope level. |


</p>

ifosli (approved):
I like it! Just make sure to rerun the format script if you haven't already

stevestreza-ksr:
Formatting code was rewritten after getting rabbit hole'd on weird quirks about currency formatting and locales. Apple seems to insist that when the user's locale is "ja" (for Japanese language), the yen SHOULD be formatted with decimals, and only when the language region is set to Japan ("ja_JP") does it remove the decimals.

For some reason though, this isn't true when using our cached number formatters. I haven't isolated why, but creating a new NumberFormatter from scratch resolved this correctly.

... but in the end, Apple also recommends Swift developers to use `Decimal.FormatStyle` for currency formatting as a more idiomatic way to do this, and claims they cache this internally. This also avoids the use of a Double. After discussion with Amy, we decided this would be ok to go with. So the code is now rewritten to use that instead.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -223,7 +223,7 @@ extension PPOProjectCardModel {
     image: .network(URL(string: "https:///")!),
     projectName: "Sugardew Island - Your cozy farm shop let’s pretend this is a way way way longer title",
     projectId: 12_345,
-    pledge: .init(amount: "50.00", currency: .usd, symbol: "$"),
+    pledge: "$50.00",
     creatorName: "rokaplay truncate if longer than",
```
amy-at-kickstarter, line 227: Ha, I was about to say something like "some of these test strings should be annoyingly long to test truncation", and then I saw you already got there!

Library/FormatTests.swift
```diff
@@ -357,6 +357,86 @@ final class FormatTests: TestCase {
...
+    withEnvironment(locale: Locale(identifier: "en_GB")) {
+      let fragment = GraphAPI.MoneyFragment(amount: "1000.33", currency: .gbp, symbol: "£")
+      XCTAssertEqual(Format.currency(fragment), "£1,000.33")
+    }
+  }
+
+  func testCurrencyMoneyFragment_JPY() {
+    withEnvironment(locale: Locale(identifier: "ja_JP")) {
+      let fragment = GraphAPI.MoneyFragment(amount: "1000.00", currency: .jpy, symbol: "¥")
+      XCTAssertEqual(Format.currency(fragment), "¥1,000")
```
amy-at-kickstarter, line 412: 👍

amy-at-kickstarter (approved):
Nice, LGTM!

Non-blocking suggestion: add a test that documents what the formatter does if we get a `MoneyFragment` with an amount that has more than 2 decimal places. Just as a bit of defensive programming.

PR #2235, opened 2025-01-08
[MBL-1490] Handle fix pledge action

ifosli:
# 📲 What

Handle the fix pledge action by opening the "manage pledge" view when the "fix pledge" button is tapped. This is consistent with behavior elsewhere in the app (primarily activity) and is significantly simpler than trying to open the pledge UI directly.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1490)

| Before bug fix | After bug fix |
| --- | --- |
| Simulator Screen Recording - iPhone 16 - 2025-01-08 at 16 33 25 | Simulator Screen Recording - iPhone 16 - 2025-01-09 at 11 06 34 | 

# ✅ Acceptance criteria

- [x] Tapping the "fix payment" button launches the fix pledge flow
- [x] Tests pass and have sufficient coverage

# ⏰ TODO

- Tested end to end flow and it works as expected. Saw a weird bug the first time I initiated the fix payment flow - will investigate if I see it again. And I'm not sure if we want to refresh automatically once an action is completed? (Did it manually here) - if so, we'll want a separate ticket for this.


scottkicks (approved):
LGTM 👍

stevestreza-ksr (approved):
Would be good to have a test that shows that sending the PPOContainerViewController a `fixPaymentMethod` event caused a `ManagePledgeViewController` to be presented, but otherwise lgtm 👍

PR #2237, opened 2025-01-09
[MBL-1958] Use New PLOT Disclaimer Copy On Project Fragment

scottkicks:
# 📲 What

Exposes a new `pledgeOverTimeMinimumExplanation` property on the Project fragment and implements it.

The threshold should be **$125**

# 🤔 Why

To keep [this string](https://www.figma.com/design/vyTjzt9Tb67oHYUVbV2zKh/PLOT-V1-Designs?node-id=914-18906&t=qqJCJuhfCyhmbsAG-4) dynamic in case the plot eligibility threshold changes in the future, payments added it for us on the Project fragment.

# 🛠 How

- exposes the new string
- uses it in the PLOT UI checkout component

# 👀 See

plot component

# ✅ Acceptance criteria

- [x] New string is being used when PLOT is ineligible for the current pledge


amy-at-kickstarter (approved):
👍 Nice! Great to have this moved to the server.

Library/ViewModels/PledgePaymentPlansOptionViewModel.swift
```diff
@@ -123,7 +123,7 @@ public final class PledgePaymentPlansOptionViewModel:
 
     self.ineligibleBadgeText = configData
       .filterWhenLatestFrom(ineligible, satisfies: { $0 == true })
-      .map { getIneligibleBadgeText(with: $0.project, thresholdAmount: $0.thresholdAmount) }
```
jovaniks, line 126: Since `thresholdAmount` will no longer be used, should we remove all references to it? Also, what about `PledgePaymentPlansAndSelectionData.thresholdAmount`? It might be a good idea to clean those up if they're not needed anymore.

jovaniks (changes requested):
LGTM overall! Just one small concern: could we clean up the unused `thresholdAmount` fields?

jovaniks (approved):
Nice!!!

PR #2241, opened 2025-01-13
MBL-1985: Remove unused .update context from NoShippingPledgeViewModel

amy-at-kickstarter:
# 📲 What

Remove `.update` context from `NoShippingPledgeViewModel`.

# 🤔 Why

This context is never used in the No Shipping flow. See: https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/ManagePledgeViewModel.swift#L534-L537


stevestreza-ksr (approved):
Seems straightforward enough, not sure what this was there for but if it's not being used then no sense keeping it around 👍

amy-at-kickstarter:
> Seems straightforward enough, not sure what this was there for but if it's not being used then no sense keeping it around 👍

For context, `.update` was a mode where you could generically update your pledge (the amount, the payment method, the reward). It's been replaced by specific modes like `.changePaymentMethod` and `.updateReward`.

PR #2242, opened 2025-01-14
[MBL-1990] Update PaymentIncrement fragment and implement the formatted amounts

jovaniks:
…mounts

# 📲 What

Update `PaymentIncrementFragment` fragment to include the following new formatted amounts:

- `amountAsFloat`
- `amountFormattedInProjectNativeCurrency`

# 🤔 Why

The `PaymentIncrementFragment` is updated to include formatted amounts

# 🛠 How

Updating the GraphQL fragment `PaymentIncrementFragment`

# 👀 See

- [Jira](https://kickstarter.atlassian.net/browse/MBL-1990)

# ✅ Acceptance criteria

- [ ] The currency amount should be shown as expected.

# ⏰ TODO

- [ ] Fix unit test


jovaniks:
I have a couple of points I'd like to address regarding the formatting of the amounts:

1. According to the design, when the increment amount is "0" cents, the amount should be shown without the cents (e.g., "$250" instead of "$250.00"). However, `amountFormattedInProjectNativeCurrency` is coming through as "$250.00". Do we want to display this amount without cents (e.g., "$250") instead?

2. The amounts in the Payment Schedule are attributed text, as the currency symbol and cents are displayed in a smaller size than the full amount (see attached screenshot). For this case, should we use `amountAsFloat` since we are applying the format to build the attributed text, or would we prefer to create another attributed text builder that uses `amountFormattedInProjectNativeCurrencyWithCurrencyCode`? This field returns values like "USD $250.00".

- [Checkout screen](https://www.figma.com/design/vyTjzt9Tb67oHYUVbV2zKh/PLOT-V1-Designs?node-id=914-19032&t=hoyetROBOhzI5LwP-4)
- [Payment schedule](https://www.figma.com/design/vyTjzt9Tb67oHYUVbV2zKh/PLOT-V1-Designs?node-id=1039-17930&t=hoyetROBOhzI5LwP-4)

| PLOT Charges | PLOT Disclaimer | Payment schedule |
| --- | --- | --- |
| image | image  | image |

scottkicks:
> 1. According to the design, when the increment amount is "0" cents, the amount should be displayed without the cents (e.g., "$250" instead of "$250.00"). However, `amountFormattedInProjectNativeCurrency` currently displays "$250.00". Do we need to adjust this behavior to match the design by showing amounts without cents (e.g., "$250")?

I think we should match designs as closely as we can. Is there a scenario where the increment amount would be 0?

scottkicks (approved):
If design comes back with and says it needs to be changed. We'll handle as a fast follow. For now, let's get this in

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
```diff
@@ -2,18 +2,20 @@ import Foundation
...
+    guard let amountAsDouble = Double(amountAsString) else {
       return nil
     }
 
     guard let intervalAsTime = TimeInterval.from(ISO8601DateTimeString: fragment.scheduledCollection) else {
       return nil
     }
 
-    self.amount = PledgePaymentIncrementAmount(amount: amountAsDouble, currency: currency)
+    self.amount = PledgePaymentIncrementAmount(
```
amy-at-kickstarter, line 14: Does `PledgePaymentIncrementAmount` still need `amount` and `currency` now that it has `amountFormattedInProjectNativeCurrency`?

amy-at-kickstarter (approved):
LGTM. One non-blocking suggestion for cleanup - if nothing uses `currency` or `amount` then those should be deleted.

nativeksr:
### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| KsApi/models/PledgePaymentIncrement+GraphAPITests.swift:41:1 | warning: (trailingSpace) Remove trailing space at end of a line. |


</p>

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
jovaniks: yeah, we need them to build the attributed text for Payment schedule component
image

PR #2246, opened 2025-01-15
[MBL-1961] Update PaymentIncrement State & StateReason GraphQL Types

scottkicks:
# 📲 What

Updates the PaymentIncrementFragment to include `stateReason` and updates the handling of it that and `state` now that they are enums.

Supports "canceled" and "Authentication required" states.

# 🤔 Why

- `state` and `stateReason` are now enums. 
- we need to support both new state/stateReason types

# 🛠 How

- Updates fragment and decoding/usage of said fragment
- updates the view to handle the new states.
   - "Authentication required" is handled when the `state == .errored` _and_ `stateReason == .requiresAction`
   - "Errored payment" is handled when `state == .errored` _and_ `stateReason != .requiresAction`. Right now there is no other possible `stateReason`.

# 👀 See

| Collected | Collected + Scheduled | Errored Payment |
| --- | --- | --- |
| plot-all-collected-badges | plot-collected-scheduled-badges | plot-erroredpayment-badge |

# ✅ Acceptance criteria

- [x] All payment increment states are handled and each plot payment in the schedule matches the API response states/stateReasons


Library/ViewModels/PledgeOverTimePaymentScheduleViewModel.swift
```diff
@@ -117,3 +117,15 @@ private func attributedCurrency(
...
   return attributedCurrency
 }
+
+private func stateLabelText(from increment: PledgePaymentIncrement) -> String {
+  let state = increment.state
+  let stateReason = increment.stateReason
+
+  switch (state, stateReason) {
+  case (.errored, .requiresAction):
+    return Strings.Authentication_required()
```
amy-at-kickstarter, line 127: Is authentication _always_ what's going to need to happen to an errored PLOT payment?

amy-at-kickstarter (approved):
LGTM. Looks like all that `PaymentIncrement` and `PaymentIncrementFragment` refactoring is paying off!

One thing that isn't clear to me - `ERRORED` and `REQUIRES_ACTION` don't really map 1:1 to `"Authentication Required"`. Does the `ACTION` always imply authentication? Or should that enum state actually be something like `REQUIRES_AUTHENTICATION`? Mostly I'm wondering if eventually we'll be adding other error states, and the copy we're showing will become out-of-date.

Library/ViewModels/PledgeOverTimePaymentScheduleViewModel.swift
scottkicks: only if state == .errored and stateReason == .requiresAction. if state == .errored and stateReason == anything else, we show the Errored payment badge.

Right now stateReason can only be .requiresAction or nil.

amy-at-kickstarter: Gotcha. As long as this is future-proofed against future `stateReason`s, it seems good to me.

jovaniks (approved):
LGTM

PR #2250, opened 2025-01-20
[MBL-1493] Handle confirm address action

ifosli:
# 📲 What

Handle confirm address by wiring up confirm address button to the `createOrUpdateBacking` mutation.

This also fixes a minor bug in the way the address is displayed (and updates tests to catch that bug), and updates the confirm address card to only be displayed/created if the backend gives us an address.

Note: while the confirm address mutation does successfully get called, this does not cause the card to disappear. A backend ticket exists to look into what's going on.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1493)

| Happy path | Error |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Plus - 2025-01-21 at 15 11 09 | Simulator Screen Recording - iPhone 16 Plus - 2025-01-21 at 15 15 17 |

# ♿️ Accessibility 

untested

# ✅ Acceptance criteria

- [x] Tapping the confirm address button causes the expected mutation to be called and a success banner shows (assuming no errors)
- [x] If there are errors, tapping confirm address button causes error banner to show.

# ⏰ TODO

- We probably want a loading state for this button, too. I didn't add it in this PR because I didn't want to overcomplicate things (and it'd be nice to have a general loading state for buttons; maybe tied to the ppo view refreshing itself?). It especially didn't feel worth spending extra time on this when confirming the address doesn't actually work (and I'm not sure if a backend solution is actually feasible). Happy to file a ticket to add a loading state to the button if y'all would like one.


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -402,33 +422,37 @@ extension PPOProjectCardModel {
...
       secondaryAction = nil
       tierType = .fixPayment
-    case (PPOProjectCardModelConstants.confirmAddress, _):
-      primaryAction = .confirmAddress
+    case let (PPOProjectCardModelConstants.confirmAddress, _, .some(address), .some(addressId)):
+      primaryAction = .confirmAddress(address: address, addressId: addressId)
       secondaryAction = .editAddress
       tierType = .confirmAddress
-    case (PPOProjectCardModelConstants.completeSurvey, _):
+    case (PPOProjectCardModelConstants.confirmAddress, _, _, _):
```
ifosli, line 436: I'm not sure about this pattern - I'm trying to be consistent with the existing code, but I'm tempted to just move the address nullability checks into the case itself. I definitely don't want to add 4 separate cases (3 that map to nil; if one is nil, if the other is nil, or if they're both nil). I'm especially interested in if you have a better way to do this, Steve, since you felt strongly about the way this currently works for the client secret. (The address nullability isn't relevant for the other cards, but neither is the client secret for cards that aren't the auth card.)

stevestreza-ksr: After thinking on this overnight, I tried a refactor where each parser was moved into helper functions, which should keep things clean while also not overloading the switch statement. Thanks for pushing back on it, let me know how it looks now. 😄

stevestreza-ksr:
Taking this PR over while Ingerid is out, and thus swapping out myself as a reviewer

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -16,6 +16,7 @@ public struct PPOProjectCardModel: Identifiable, Equatable, Hashable {
   public let tierType: TierType
   public let backingDetailsUrl: String
   public let backingId: Int
+  public let backingGraphId: String
```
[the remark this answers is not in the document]
amy-at-kickstarter: Ha, this burned me in late pledges, too. Out of scope here, but perhaps we should make our own type like
```
struct GraphQLID {
   var base64Encoded { get }
   var integer { get }
   //...
}
```
instead of trying to guess all the time.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -232,10 +233,18 @@ extension PPOProjectCardModel {
       Los Angeles, CA 90025-1234
       United States
     """,
-    actions: (.confirmAddress, .editAddress),
```
amy-at-kickstarter, line 235: Also a little out of scope, but I find it slightly confusing to see template data in a production file. Would be nice to see in a more appropriately-named file.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -375,60 +388,60 @@ extension PPOProjectCardModel {
...
-      return fields.compactMap { $0 }.joined(separator: "\n")
+      // Create address from all fields that are not nil and not the empty string.
+      return fields.compactMap { ($0 ?? "").isEmpty ? nil : $0 }.joined(separator: "\n")
+    }
+    let addressWithName = addressWithoutName.flatMap { address in
+      guard let name = backing?.deliveryAddress?.recipientName else {
+        return address
+      }
+      return name + "\n" + address
     }
```
amy-at-kickstarter, line 413: This is a good candidate for text that should be formatted on the back end. IIRC addressing formatting is a Hard Problem that can get increasingly complicated for other countries.

Fine here for now, but something to make a note of for v2.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -375,60 +388,60 @@ extension PPOProjectCardModel {
...
+      actionResult = Self.actionsForConfirmAddress(address: addressWithoutName, addressId: addressId)
+    case PPOProjectCardModelConstants.completeSurvey:
+      actionResult = Self.actionsForSurvey()
+    case PPOProjectCardModelConstants.authenticationRequired:
+      actionResult = Self.actionsForAuthentication(clientSecret: backing?.clientSecret)
+    default:
       return nil
     }
 
+    guard let (primaryAction, secondaryAction, tierType) = actionResult else { return nil }
```
amy-at-kickstarter, line 433: I find tuples longer than 2 items hard to reason about and debug later. Can you make this a `struct` instead?

amy-at-kickstarter (approved):
I have some non-blocking suggestions, but mostly LGTM.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
stevestreza-ksr: Yeah, this file is getting a little out of hand, so I'll split out both the templates and the parsing into their own files. I probably should've done that sooner, thanks for the nudge.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

stevestreza-ksr:
Most of the LoC in the changes are now related to splitting up PPOProjectCardModel, and aren't actual "changes".

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel+Parsing.swift
```diff
@@ -0,0 +1,145 @@
...
+    case PPOProjectCardModelConstants.paymentFailed:
+      actions = Self.actionsForPaymentFailed()
+    case PPOProjectCardModelConstants.confirmAddress:
+      actions = Self.actionsForConfirmAddress(address: addressWithoutName, addressId: addressId)
+    case PPOProjectCardModelConstants.completeSurvey:
+      actions = Self.actionsForSurvey()
+    case PPOProjectCardModelConstants.authenticationRequired:
+      actions = Self.actionsForAuthentication(clientSecret: backing?.clientSecret)
+    default:
+      return nil
```
stevestreza-ksr, line 73: This section was refactored to use this new PPOParsedAction struct and the helper functions at the bottom of the file.

stevestreza-ksr:
Since this PR went through a weird path it doesn't have 2 approvals, but with one approval + the handoff + Ingerid's signing off, I'm calling that two approvals and merging once CI passes.

PR #2251, opened 2025-01-20
[MBL-1996] Don't fetch errored pledges if PPO is enabled

ifosli:
# 📲 What

Since PPO will live right next to activities, it feels redundant to show the same information in both places. We will still show surveys in both places, however, at least while some surveys show in Activity without being available in PPO.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-1996)

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] When PPO flag is on, errored pledges are not included in the activity count and errored pledges don't show in Activity
- [x] When PPO flag is off, errored pledges are included in both

# ⏰ TODO

- We will likely want to hide surveys as well, once all surveys are available as part of PPO, but that's been removed from this ticket.
- If we decide to only roll out to superbackers after all, this change will need to be rolled back, too.

### DIFF
--- a/Library/ViewModels/ActivitiesViewModel.swift
+++ b/Library/ViewModels/ActivitiesViewModel.swift
@@ -235,8 +235,7 @@ public final class ActivitiesViewModel: ActivitiesViewModelType, ActitiviesViewM
           self.surveyResponseViewControllerDismissedProperty.signal
         )
       )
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter (approved):
LGTM, love the comments.

stevestreza-ksr (approved):

PR #2255, opened 2025-01-22
MBL-1972: Add PaymentMethodsUseCase and tests

amy-at-kickstarter:
# 📲 What

Add `PaymentsMethodsUseCase` and `PaymentMethodsUseCaseTests`.

n.B. that this does not hook up the use case - I'm holding off on installing it until our next release is cut. 

# 🤔 Why

This puts all of the code related to creating/configuring/displaying a `PledgePaymentMethodsViewController` in one place.

This also means Late + Live pledges will be able to start sharing some of the code that's currently duplicated.


Library/ViewModels/PaymentMethodsUseCase.swift
```diff
@@ -0,0 +1,57 @@
...
+ Outputs:
+  * `configurePaymentMethodsViewControllerWithValue` - Configuration data for `PledgePaymentMethodsViewController`. May be sent never (if payment methods are disabled, or if the user is logged out), once, or many times (if the user logs out and in again.)
+  * `paymentMethodsViewHidden` - Whether or not to show the payment methods view. Sent at least once after `initialData` is sent.
+  * `selectedPaymentSource` - The currently selected credit card, or `nil` if no card is selected. Sent at least once after `initialData` is sent.
+  */
+public final class PaymentMethodsUseCase {
+  init(initialData: Signal<PledgeViewData, Never>, userSessionStarted: Signal<(), Never>) {
+    let project = initialData.map(\.project)
+    let baseReward = initialData.map(\.rewards).map(\.first).skipNil()
+    let refTag = initialData.map(\.refTag)
```
amy-at-kickstarter, line 18: As a fun note - pulling this out into its own component makes it a much more obvious that we're passing around extra data we don't need. The pledge payment methods page doesn't need to know about `refTag` or `baseReward`, as far as I know. Once this is hooked up to both pledge controllers, refactoring those redundancies will be easier!

Library/ViewModels/PaymentMethodsUseCase.swift
```diff
@@ -0,0 +1,57 @@
...
+    let configurePaymentMethodsViewController = Signal.merge(
+      initialDataUnpacked,
+      initialDataUnpacked.takeWhen(userSessionStarted)
+    )
+
+    self.configurePaymentMethodsViewControllerWithValue = configurePaymentMethodsViewController
+      .filter { !$3.paymentMethodsViewHidden }
+      .compactMap { project, reward, refTag, context -> PledgePaymentMethodsValue? in
+        guard let user = AppEnvironment.current.currentUser else { return nil }
+        return (user, project, "", reward, context, refTag)
```
amy-at-kickstarter, line 36: I removed a defunct comment about `NaN` here.

Library/ViewModels/PaymentMethodsUseCaseTests.swift
```diff
@@ -0,0 +1,129 @@
...
+      initialData: self.initialDataSignal,
+      userSessionStarted: self.userSessionStartedSignal
+    )
+
+    self.useCase.paymentMethodsViewHidden.observe(self.paymentMethodsViewHidden.observer)
+    self.useCase.configurePaymentMethodsViewControllerWithValue.observe(self.configureWithValue.observer)
+    self.useCase.selectedPaymentSource.observe(self.selectedPaymentSource.observer)
+  }
+
+  func test_LoggedInUser_SendsConfigValue_AndShowsPaymentMethods() {
```
amy-at-kickstarter, line 34: To @ifosli's point, I'm trying to make these tests less of [change detectors](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) than our old pledge tests. They're not perfectly comprehensive, but they should exercise most of the important bits of the use case.

Library/ViewModels/PaymentMethodsUseCase.swift
```diff
@@ -0,0 +1,70 @@
...
+ Inputs:
+  * `initialData` - Pledge configuration data. Must send at least one event for any outputs to send.
+  * `userSessionStarted` - Empty signal that indicates when a user logs in.
+  * `creditCardSelected(with:)` - Call this when the user selects a new card.
+ Outputs:
+  * `configurePaymentMethodsViewControllerWithValue` - Configuration data for `PledgePaymentMethodsViewController`. May be sent never (if payment methods are disabled, or if the user is logged out), once, or many times (if the user logs out and in again.)
+  * `paymentMethodsViewHidden` - Whether or not to show the payment methods view. Sent at least once after `initialData` is sent.
+  * `selectedPaymentSource` - The currently selected credit card, or `nil` if no card is selected. Sent at least once after `initialData` is sent.
+  */
+public final class PaymentMethodsUseCase: PaymentMethodsUseCaseInputs, PaymentMethodsUseCaseOutputs {
```
jovaniks, line 24: Let's follow the `VieModelType` protocol pattern here

jovaniks (changes requested):
This is great work! It's a nice step towards making the view models cleaner and splitting them into smaller pieces focused on single responsibility components. 

To align with existing patterns, I’d recommend following the `ViewModelType` protocol pattern.

Library/ViewModels/PaymentMethodsUseCase.swift
```diff
@@ -0,0 +1,70 @@
...
+  init(initialData: Signal<PledgeViewData, Never>, userSessionStarted: Signal<(), Never>) {
+    let project = initialData.map(\.project)
+    let baseReward = initialData.map(\.rewards).map(\.first).skipNil()
+    let refTag = initialData.map(\.refTag)
+    let context = initialData.map(\.context)
+
+    let isLoggedIn = Signal.merge(initialData.ignoreValues(), userSessionStarted)
+      .map { _ in AppEnvironment.current.currentUser }
+      .map { $0 != nil }
```
stevestreza-ksr, line 33: Nit: Could probably merge these

stevestreza-ksr (approved):

Library/ViewModels/PaymentMethodsUseCase.swift
amy-at-kickstarter: Whoops! Missed that again, thanks for the eagle eyes

Library/ViewModels/PaymentMethodsUseCase.swift
```diff
@@ -0,0 +1,57 @@
...
+  * `initialData` - Pledge configuration data. Must send at least one event for any outputs to send.
+  * `userSessionStarted` - Empty signal that indicates when a user logs in.
+  * `creditCardSelected(with:)` - Call this when the user selects a new card.
+ Outputs:
+  * `configurePaymentMethodsViewControllerWithValue` - Configuration data for `PledgePaymentMethodsViewController`. May be sent never (if payment methods are disabled, or if the user is logged out), once, or many times (if the user logs out and in again.)
+  * `paymentMethodsViewHidden` - Whether or not to show the payment methods view. Sent at least once after `initialData` is sent.
+  * `selectedPaymentSource` - The currently selected credit card, or `nil` if no card is selected. Sent at least once after `initialData` is sent.
+  */
+public final class PaymentMethodsUseCase {
+  init(initialData: Signal<PledgeViewData, Never>, userSessionStarted: Signal<(), Never>) {
```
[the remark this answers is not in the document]
amy-at-kickstarter: Hmm, yeah, fair point. The only catch is that there's an extra input (`creditCardSelected`) which doesn't live on init, so I don't want to document them there. Were you thinking the comments would make more sense to put on `PaymentMethodsUseCaseInputs` and `PaymentMethodUseCaseOutputs`?

jovaniks (approved):
Nice, great job

scottkicks (approved):
🎉 Nice! Its refreshing to see these behaviors broken into smaller more digestible modules.

PR #2256, opened 2025-01-22
Rename PLOTPledgeViewModel to PledgeOverTimeUseCase

amy-at-kickstarter:
# 📲 What

Rename PLOTPledgeViewModel to PledgeOverTimeUseCase (and associated variables, tests, etc etc.)

# 🤔 Why

Android calls these sub-modules 'use cases', which is a great name. Renaming this makes us more consistent, and will be consistent with other use cases I'm writing.

# 🛠 How

Find and replace, woohoo


scottkicks (approved):
Oh nice. Yea I like UseCase better

Library/ViewModels/PledgeOverTimeUseCase.swift
```diff
@@ -35,7 +35,7 @@ public protocol PLOTPledgeViewModelOutputs {
   - `pledgeOverTimeIsLoading`: Whether the PLOT module is loading. Sends one or more events.
  */
 
-public final class PLOTPledgeViewModel: PLOTPledgeViewModelInputs, PLOTPledgeViewModelOutputs {
+public final class PledgeOverTimeUseCase: PledgeOverTimeUseCaseInputs, PledgeOverTimeUseCaseOutputs {
```
jovaniks, line 38: I'm not sure if it's relevant at this point, but generally, we define another protocol, like "ViewModelType", for view models. It could look something like this:

```swift
public protocol PledgeOverTimeUseCaseType {
  var inputs: PledgeOverTimeUseCaseInputs { get }
  var outputs: PledgeOverTimeUseCaseOutputs { get }
}
```

jovaniks (approved):
LGTM. Just make sure we want to omit the "ViewModelType" pattern implementation here.

Library/ViewModels/PledgeOverTimeUseCase.swift
amy-at-kickstarter: Yes, in this case I'm just renaming it from `...ViewModelInputs` to `...UseCaseInputs`

amy-at-kickstarter: Oh, hm. I forgot about that pattern. I'll add it.

PR #2258, opened 2025-01-22
Create ApplePayTokenUseCase and tests

amy-at-kickstarter:
# 📲 What

Create a use case that will handle ApplePay in the regular (live) pledge flow.

# 🤔 Why

This puts a lot of ApplePay code in one place, so we can pull it out of `NoShippingPledgeViewModel` in the next release.

It's worth noting that ApplePay in late + live pledges is fundamentally different - they use different Stripe APIs. It would be nice to consolidate that, but that would be out of scope without further API changes. This is just a refactoring for niceness and modularity, instead of code re-use.

### DIFF
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -1595,6 +1595,8 @@
 		E13D76812D42CBA400FB58CE /* LoginSignupUseCaseTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E13D767E2D42C52900FB58CE /* LoginSignupUseCaseTests.swift */; };
 		E13D76712D4011BF00FB58CE /* PaymentMethodsUseCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = E13D76702D4011BF00FB58CE /* PaymentMethodsUseCase.swift */; };
 		E13D76742D404F1600FB58CE /* PaymentMethodsUseCaseTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E13D76722D40438100FB58CE /* PaymentMethodsUseCaseTests.swift */; };
+		E13D76762D41698500FB58CE /* ApplePayTokenUseCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = E13D76752D41698000FB58CE /* ApplePayTokenUseCase.swift */; };
+		E13D76792D419A5900FB58CE /* ApplePayTokenUseCaseTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E13D76772D4195E700FB58CE /* ApplePayTokenUseCaseTests.swift */; };
 		E16794282B7EAA5200064063 /* OAuthTokenExchange.swift in Sources */ = {isa = PBXBuildFile; fileRef = E16794272B7EAA5200064063 /* OAuthTokenExchange.swift */; };
 		E167942A2B85136900064063 /* OAuthTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E16794292B85136900064063 /* OAuthTests.swift */; };
 		E16ECA702C245A34002A1D25 /* PagedContainerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E16ECA6F2C245A34002A1D25 /* PagedContainerViewController.swift */; };
@@ -3314,6 +3316,8 @@
 		E13D767E2D42C52900FB58CE /* LoginSignupUseCaseTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LoginSignupUseCaseTests.swift; sourceTree = "<group>"; };
 		E13D76702D4011BF00FB58CE /* PaymentMethodsUseCase.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PaymentMethodsUseCase.swift; sourceTree = "<group>"; };
 		E13D76722D40438100FB58CE /* PaymentMethodsUseCaseTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PaymentMethodsUseCaseTests.swift; sourceTree = "<group>"; };
+		E13D76752D41698000FB58CE /* ApplePayTokenUseCase.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ApplePayTokenUseCase.swift; sourceTree = "<group>"; };
+		E13D76772D4195E700FB58CE /* ApplePayTokenUseCaseTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ApplePayTokenUseCaseTests.swift; sourceTree = "<group>"; };
 		E16794272B7EAA5200064063 /* OAuthTokenExchange.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OAuthTokenExchange.swift; sourceTree = "<group>"; };
 		E16794292B85136900064063 /* OAuthTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OAuthTests.swift; sourceTree = "<group>"; };
 		E16ECA6F2C245A34002A1D25 /* PagedContainerViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PagedContainerViewController.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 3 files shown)


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -279,7 +279,6 @@
 		33AF01BC2D2D74B30085A153 /* PledgePaymentIncrementState+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33AF01BB2D2D748B0085A153 /* PledgePaymentIncrementState+Extension.swift */; };
 		33AF01BF2D2E72430085A153 /* PledgeOverTimePaymentScheduleViewModelTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33AF01BD2D2E71E30085A153 /* PledgeOverTimePaymentScheduleViewModelTest.swift */; };
 		33AF01C22D2E973A0085A153 /* PledgeOverTimePaymentScheduleViewControllerTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33AF01C02D2E95FD0085A153 /* PledgeOverTimePaymentScheduleViewControllerTest.swift */; };
-		33C9F0BE2CF5104C00B62E14 /* PledgePaymentPlanOptionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33C9F0BC2CED6DEF00B62E14 /* PledgePaymentPlanOptionView.swift */; };
```
amy-at-kickstarter, line 282: Must have been an automated change, I didn't touch this.

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+    displayName: String?,
+    network: String?,
+    transactionIdentifier: String
+  ))
+  func paymentAuthorizationViewControllerDidFinish()
+  func stripeTokenCreated(token: String?, error: Error?) -> PKPaymentAuthorizationStatus
+}
+
+public protocol ApplePayTokenUseCaseOutputs {
+  var goToApplePayPaymentAuthorization: Signal<PKPaymentRequest, Never> { get }
```
amy-at-kickstarter, line 18: Slight tweak; I made this a `PKPaymentRequest` instead of `PaymentAuthorizationData`.

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+
+ Inputs:
+ - `initialData` - An `initialData` event is required for any other signals to send.
+
+ Outputs:
+ - `applePayAuthorizationStatus` - Sends an event indicating whether the ApplePay flow succeeded or failed.
+  */
+
+public final class ApplePayTokenUseCase: ApplePayTokenUseCaseInputs, ApplePayTokenUseCaseOutputs {
+  init(initialData: Signal<PaymentAuthorizationData, Never>) {
```
amy-at-kickstarter, line 44: This is largely copy-pasted from `NoShippingPledgeViewModel`.

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+    .mapConst(PKPaymentAuthorizationStatus.success)
+
+    let applePayStatusFailure = Signal.merge(
+      self.stripeErrorSignal.skipNil().ignoreValues(),
+      self.stripeTokenSignal.filter { $0 == nil }.ignoreValues(),
+      pkPaymentData.filter { $0 == nil }.ignoreValues()
+    )
+    .mapConst(PKPaymentAuthorizationStatus.failure)
+
+    self.createApplePayBackingStatusProperty <~ Signal.merge(
```
amy-at-kickstarter, line 85: TIL this is a ReactiveSwift operator that means `bind`.

Kickstarter.xcodeproj/project.pbxproj
stevestreza-ksr: Interesting, the same change showed up on my branch as well, maybe it was a merge issue or something as it appeared twice in the pbxproj. Agreed that it appears to be automated, should be fine to include it IMO

jovaniks (changes requested):
Same as other UseCase components, just implement the `ViewModelType` protocol pattern.

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+/**
+ A use case for ApplePay transactions in the regular (not post!) pledge flow.
+
+ To complete an ApplePay payment, the use case should be used in this order:
+ - `input.applePayButtonTapped()` - The ApplePay button has been tapped
+ - `output.goToApplePayPaymentAuthorization` - The view controller should display a `PKPaymentAuthorizationViewController` with the sent `PKPaymentRequest`
+ - `input.paymentAuthorizationDidAuthorizePayment(paymentData:)` - The `PKPaymentAuthorizationViewController` successfully authorized a payment
+ - `input.stripeTokenCreated(token:error:)` - Stripe successfully turned the `PKPayment` into a Stripe token. Returns a status, which is also sent by the `applePayAuthorizationStatus` signal.
+ - `input.paymentAuthorizationViewControllerDidFinish()` - The `PKPaymentAuthorizationViewController` was dismissed
+ - `output.applePayParams` - Sends parameters which can be used in `CreateBacking` or `UpdateBacking`. Sends an initial `nil`value, by default.
```
stevestreza-ksr, line 32: Not necessary for this PR, but long term it might be good to build a little enum-based state machine for this if these must be called in order. Then the compiler can enforce they are called correctly.

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+      .takeWhen(self.applePayButtonTappedSignal)
+
+    let pkPaymentData = self.pkPaymentSignal
+      .map { pkPayment -> PKPaymentData? in
+        guard let displayName = pkPayment.displayName, let network = pkPayment.network else {
+          return nil
+        }
+
+        return (displayName, network, pkPayment.transactionIdentifier)
+      }
```
stevestreza-ksr, line 69: Looks like all the downstream signal processing is skipping nils. You could move the `skipNils` to this step or use `compactMap` instead of `map` here.

stevestreza-ksr (approved):

Library/ViewModels/ApplePayTokenUseCase.swift
```diff
@@ -0,0 +1,160 @@
...
+}
+
+public protocol ApplePayTokenUseCaseOutputs {
+  var goToApplePayPaymentAuthorization: Signal<PKPaymentRequest, Never> { get }
+  var applePayParams: Signal<ApplePayParams?, Never> { get }
+  var applePayAuthorizationStatus: Signal<PKPaymentAuthorizationStatus, Never> { get }
+}
+
+/**
+ A use case for ApplePay transactions in the regular (not post!) pledge flow.
```
scottkicks, line 24: nit: I know that sometimes we use `live` instead of `post`. IMO `late` is a better naming convention, but either way It'd be a good idea to align on one and stick to it. You used `live` in your [other PR](https://github.com/kickstarter/ios-oss/pull/2255) so maybe we can go with that here.

Library/ViewModels/ApplePayTokenUseCase.swift
scottkicks: is it fundamentally the same as `=`? I guess I'm wondering why use a different approach suddenly for this output.

scottkicks (approved):
LGTM 

Just a couple minor comments.

Library/ViewModels/ApplePayTokenUseCase.swift
amy-at-kickstarter: Not quite - here's the comment on it in ReactiveSwift: https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Sources/UnidirectionalBinding.swift#L68

My guess is that we're doing this because `Signal.merge` outputs a `Signal`, but `createApplePayBackingStatusProperty` is a `MutableProperty` - so you couldn't assign this with `=`. 

More broadly, though, it's not clear to me whether using `MutableProperty` everywhere is appropriate in our codebase. I don't think we ever read the current value of a mutable property, we just observe them as if they were signals. This is something I'd like to understand better about the guts of ReactiveSwift.

Library/ViewModels/ApplePayTokenUseCase.swift
amy-at-kickstarter: Love this idea. Would be great to make this enforced in code instead of with hopes and comments.

Library/ViewModels/ApplePayTokenUseCase.swift
amy-at-kickstarter: Ah, there's a few lines in there that are filtering _for_ nil values for purposes of error processing. Slightly clunky pattern unfortunately.

jovaniks (approved):
Oh nice, this new pattern is amazing.

PR #2261, opened 2025-01-23
MBL-1967: Create LoginSignupUseCase and install in pledge flows

amy-at-kickstarter:
# 📲 What

Create use case for logging in and signing up.

# 🤔 Why

This puts signup/login code in one place, so we can pull it out of NoShippingPledgeViewModel and NoShippingPostCampaignCheckoutViewModel in the next release.


Library/ViewModels/LoginSignupUseCase.swift
```diff
@@ -0,0 +1,57 @@
...
+  var outputs: LoginSignupUseCaseOutputs { get }
+}
+
+public final class LoginSignupUseCase: LoginSignupUseCaseType, LoginSignupUseCaseInputs,
+  LoginSignupUseCaseOutputs {
+  init(withLoginIntent loginIntent: LoginIntent) {
+    self.goToLoginSignup = self.goToLoginSignupSignal
+      .mapConst(loginIntent)
+
+    self.sessionStartedObserver = NotificationCenter.default
```
amy-at-kickstarter, line 24: Typically we put these observers in the view controller, but I see no practical reason why they can't live directly in a use case or view model. It cuts down on some inputs/outputs.

amy-at-kickstarter: Aha, when trying to use this later, I realized some good reasons we _don't_ do this - namely it mucks up testability. I'm going to go back to the drawing board here, and will re-open the PR when I'm done.

amy-at-kickstarter:
OK, I'm now done tweaking this pattern and fixing tests. This one is good to go.

scottkicks (approved):
👍 
Are you going to start updating the checkout VMs with the other refactor work you have PRs for now as well?

Library/ViewModels/LoginSignupUseCase.swift
```diff
@@ -0,0 +1,82 @@
...
+  var goToLoginSignup: Signal<LoginIntent, Never> { get }
+}
+
+public protocol LoginSignupUseCaseDataOutputs {
+  var userSessionChanged: Signal<Void, Never> { get }
+  var isLoggedIn: Signal<Bool, Never> { get }
+}
+
+/**
+ A use case for logging in and signing up during the pledge flow.
```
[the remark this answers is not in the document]
amy-at-kickstarter: Good point, will fix. This one is for both flows.

Library/ViewModels/NoShippingPledgeViewModelTests.swift
```diff
@@ -12,6 +12,12 @@ private struct MockStripePaymentHandlerActionStatus: StripePaymentHandlerActionS
   let status: StripePaymentHandlerActionStatus
 }
 
+/*
+ The vm is subscribing to this thing
+ it accepts an event in the test
+ then user session started always has a value??
+ */
+
```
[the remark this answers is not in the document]
amy-at-kickstarter: Hahaha whoops. This is less embarrassing than my notes-to-self usually are!

Library/ViewModels/LoginSignupUseCase.swift
```diff
@@ -0,0 +1,82 @@
...
+    return self.userSessionDidChangeSignal
+  }
+
+  public let isLoggedIn: Signal<Bool, Never>
+
+  // MARK: - Type
+
+  public var uiInputs: any LoginSignupUseCaseUIInputs { return self }
+  public var uiOutputs: any LoginSignupUseCaseUIOutputs { return self }
+  public var dataOutputs: any LoginSignupUseCaseDataOutputs { return self }
```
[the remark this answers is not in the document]
amy-at-kickstarter: They did indeed!

Library/ViewModels/LoginSignupUseCase.swift
```diff
@@ -0,0 +1,82 @@
...
+  func goToLoginSignupTapped()
+  func userSessionDidChange()
+}
+
+public protocol LoginSignupUseCaseUIOutputs {
+  var goToLoginSignup: Signal<LoginIntent, Never> { get }
+}
+
+public protocol LoginSignupUseCaseDataOutputs {
+  var userSessionChanged: Signal<Void, Never> { get }
```
ifosli, line 20: Nit: this seems like an unnecessary output, since it's just reflecting the input. I'm okay with keeping it if it makes things easier, though.

ifosli (approved):
Nice! When I looked at the code for the use case, I didn't expect to like this change - it seemed like a lot of code just for an input to trigger an output basically. But in the view models that use it, it's super clear and it doesn't feel like it's actually adding more code. I'm excited to be able to do similar things for more complicated use cases!

stevestreza-ksr (approved):

amy-at-kickstarter:
> Are you going to start updating the checkout VMs with the other refactor work you have PRs for now as well?

Yup! I'm going to do them one at a time, since they all conflict. This is just the first!

Library/ViewModels/LoginSignupUseCase.swift
amy-at-kickstarter: I feel this way about a lot of reactive code we use to hook up buttons 🤷‍♀️

PR #2270, opened 2025-01-30
Replace assertValues with single value to assertLastValue in pledge view model tests

amy-at-kickstarter:
# 📲 What

Replace calls to `assertValues([true])` to `assertLastValue(true)` and `assertValues([false])` to `assertLastValue(false)`.

# 🤔 Why

These tests frequently operate as change detectors - they break under small, trivial refactors. To whit, much of my attempts to refactor checkout code have caused harmless signal side effects which break the tests.

For example, if a signal emits `[true, true]` instead of `[true]`, is it actually a bug? The UI would have no notion of that change, nor would the user. And while I wish I could make all of my refactoring purely side-effect free, it hasn't been realistic, unfortunately.

Fixing this will make it easier to refactor our checkout code, without breaking tests, and adding minimal risk.

# 🛠 How

Find-n-replace.


scottkicks (approved):

jovaniks (approved):
LGTM! Replacing `assertValues` with `assertLastValue` makes sense to avoid unnecessary test failures due to minor refactors. This is especially relevant since what we actually want to verify is the last emitted state, not the entire stream. 🚀

PR #2276, opened 2025-02-10
[MBL-2044] New design System - Custom font

jovaniks:
# 📲 What

This PR introduces a new Typography (**Inter**) that defines the updated typography styles based on the new design system.

# 🛠 How

- Created a new Typography file to define the updated typography styles according to the new design specifications.
- Added the new styles in the `Font.swift` file.
- A new protocol `CustomFont` is created to implement custom fonts.
- The custom font is under feature flag **new_design_system**.
- The new font should support Dynamic Type and Bold Text

# 👀 See

- [Jira](https://kickstarter.atlassian.net/browse/MBL-2044)

|  |  | |
| --- | --- | --- |
|  Simulator Screenshot - iPhone SE (3rd generation) - 2025-02-10 at 07 40 59  |  Simulator Screenshot - iPhone SE (3rd generation) - 2025-02-10 at 07 41 07  |  Simulator Screenshot - iPhone SE (3rd generation) - 2025-02-10 at 07 41 13 |

# ✅ Acceptance criteria

- [ ] The new typography should be applied to the UI when the feature flag is ON.
- [ ] All UI test should pass
- [ ] Dynamic type should be supported.

# ⏰ TODO

- [ ] Add the new **new_design_system** in remote config
- [ ] Set the font to all of the text components (UILabels, UItextfield, etc) in UIStoryborads, Xib and UIViews in a another PR.
- [ ] Add UI for the new font.


Library/Styles/Fonts/InterFont.swift
```diff
@@ -0,0 +1,92 @@
...
+    case .bodyMD: return 14
+    case .bodySM: return 12
+    case .bodyXS: return 11
+    case .bodyXXS: return 10
+    }
+  }
+
+  var textStyle: UIFont.TextStyle {
+    switch self {
+    // Old Design System
```
jovaniks, line 66: the current styles maintain the same text styles

Library/Styles/Fonts/SystemFont.swift
```diff
@@ -0,0 +1,58 @@
+import UIKit
```
jovaniks, line 1: Current font styles

jovaniks (commented):
In a new task: make sure all of the text components like `UILabel`, `UItextField`, etc, are using `UIFont.kst_{type}`

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -34,3 +34,7 @@ public func featurePledgedProjectsOverviewEnabled() -> Bool {
 public func featurePledgeOverTimeEnabled() -> Bool {
   featureEnabled(feature: .pledgeOverTime)
 }
+
+public func featureNewDesignSystemEnabled() -> Bool {
+  featureEnabled(feature: .newDesignSystem)
```
stevestreza-ksr, line 39: Open question: Since these fonts are going to be used *everywhere*, this function will be getting called a lot, especially in rendering code. Does it make sense to have something like this:

```swift
public func featureNewDesignSystemEnabled() -> Bool {

#if DESIGN_SYSTEM_ENABLED
  featureEnabled(feature: .newDesignSystem)
#else
  false
#endif

}
```

And then only define that macro in debug builds until we're closer to shipping it?

Library/Styles/Fonts/SystemFont.swift
```diff
@@ -0,0 +1,58 @@
...
+
+  func font(size: CGFloat?) -> UIFont {
+    let font = UIFont.preferredFont(
+      forTextStyle: self.textStyle,
+      compatibleWith: .current
+    )
+    let descriptor = font.fontDescriptor
+    return UIFont(
+      descriptor: descriptor,
+      size: ceil(font.pointSize / self.defaultSize * (size ?? self.defaultSize))
```
stevestreza-ksr, line 23: This looks like the same implementation as `defaultSystemFont(…)`, can these be unified?

stevestreza-ksr (commented):
Aside from the comments, seems straightforward. Will look once more in the AM with fresh eyes.

Library/Styles/Fonts/Fonts.swift
```diff
@@ -0,0 +1,207 @@
...
+  public static func ksr_bodyXXS(size: CGFloat? = nil) -> UIFont {
+    return InterFont.bodyXXS.font(size: size)
+  }
+}
+
+extension UIFont {
+  /// Returns a bolded version of `self`.
+  public var bolded: UIFont {
+    return self.fontDescriptor.withSymbolicTraits(.traitBold)
+      .map { UIFont(descriptor: $0, size: 0.0) } ?? self
```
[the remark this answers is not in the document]
jovaniks: We set `0.0` to get the new font with the same size as the original one.

[more info](https://developer.apple.com/documentation/uikit/uifont/init%28descriptor%3Asize%3A%29)

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
jovaniks: That sounds a good idea, I'll go for it

Library/Styles/Fonts/SystemFont.swift
jovaniks: Ah good catch, I think it is possible, I'll make this change

stevestreza-ksr (approved):

Library/Styles/Fonts/Fonts.swift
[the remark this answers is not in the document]
amy-at-kickstarter: This would be good as a code comment

amy-at-kickstarter (approved):
LGTM 👍

scottkicks (approved):

Library/Styles/Fonts/Fonts.swift
```diff
@@ -0,0 +1,207 @@
+import UIKit
```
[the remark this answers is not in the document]
jovaniks: Yeah, I moved it to have all of font related in the same path

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -34,3 +34,11 @@ public func featurePledgedProjectsOverviewEnabled() -> Bool {
 public func featurePledgeOverTimeEnabled() -> Bool {
   featureEnabled(feature: .pledgeOverTime)
 }
+
+public func featureNewDesignSystemEnabled() -> Bool {
+  #if DESIGN_SYSTEM_ENABLED
```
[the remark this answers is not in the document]
jovaniks: As @stevestreza-ksr suggested, we should enable the compiler flag when shipping this feature to specific build targets such as `AppCenter Alpha`, `AppCenter Beta`, `Release`, etc.

Currently, it is only flagged in the `Debug` target.

image

PR #2279, opened 2025-02-10
MBL-2020: Use BorderlessButtonStyle for message creator button

amy-at-kickstarter:
# 📲 What
Use `BorderlessButtonStyle` for message creator button.

# 🤔 Why

Without this style, the button action hijacks other tap events in the same `HStack`. In our case, tapping anywhere that _wasn't_ a button in the card would open the Message Creator view. I was unable to find a satisfying explanation as to why this happens - a quirk of SwiftUI of some kind. However, this one liner conveniently solves the problem.

send-message-bug

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
@@ -4,7 +4,6 @@ import SwiftUI
 
 struct PPOProjectCreator: View {
   let creatorName: String
-  var onSendMessage: (() -> Void)? = nil
 
   var body: some View {
     HStack(alignment: .firstTextBaseline) {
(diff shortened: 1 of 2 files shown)


ifosli (commented):
Huh, that's interesting. Nice find! I do think we want any tap in the "creator" cell (including on the > or on the creator name or the whitespace right around it) to cause the send message action. Is this still happening with the changed button style?

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
```diff
@@ -28,6 +28,7 @@ struct PPOProjectCreator: View {
       .foregroundStyle(Color(Constants.sendMessageColor))
       .frame(alignment: Constants.buttonAlignment)
       .lineLimit(Constants.textLineLimit)
+      .buttonStyle(BorderlessButtonStyle())
```
stevestreza-ksr, line 31: Nit: Leave a comment explaining why this is here, so it doesn't get refactored out accidentally

stevestreza-ksr (approved):

amy-at-kickstarter:
I'm closing this and opening a new version, since I ended up taking a different approach. New PR incoming shortly!

PR #2281, opened 2025-02-11
MBL-2020: Use borderless buttons for project creator and details views

amy-at-kickstarter:
# 📲 What

Use borderless buttons for project creator and details views.

# 🤔 Why

1. Wrapping these views in a `Button`, instead of using `onTapGesture`, gets us the button-like interaction experience that users crave. The whole view will respond to touch, with a pleasant animation and a generous tap target. `onTapGesture` was a bit fussy about detecting taps in transparent views.
2. Adding `BorderlessButtonStyle()` fixes an issue where the button action is triggered by tapping on areas outside of the button. I was unable to find a satisfying explanation as to why this happens - a quirk of SwiftUI of some kind. However, this one liner conveniently solves the problem.

# 🛠 How

A little refactoring and rejiggering.

# 👀 See

This is what the bug looked like before:
send-message-bug

This is what these nice buttons look like now:
nice-buttons


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCard.swift
```diff
@@ -79,27 +79,33 @@ struct PPOProjectCard: View {
...
-      image: self.viewModel.card.image,
-      title: self.viewModel.card.projectName,
-      pledge: self.viewModel.card.pledge,
-      leadingColumnWidth: leadingColumnWidth
-    )
-    .padding([.horizontal])
-    .onTapGesture {
-      self.viewModel.viewBackingDetails()
+    Button { [weak viewModel] () in
+      viewModel?.viewBackingDetails()
```
amy-at-kickstarter, line 83: I noticed us using this pattern in `baseButton(for:)` and did the same - but I'm actually really curious if retain-cycle-management matters the same way in SwiftUI that it does in UIKit.

ifosli (approved):
Love this!

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCard.swift
stevestreza-ksr: Since the viewModel is a `@StateObject` it gets managed by the SwiftUI runtime. In theory it should also manage the Button instance, so it shouldn't be necessary. My concern is that under the hood SwiftUI uses UIKit for a lot of things, so there might be some case where it attaches this closure to a UIButton under the hood which doesn't get cleaned up in the same way.

So having it is really more about defensive use than anything.

I found [this article](https://medium.com/@cobainjosue/reducing-memory-leaks-in-swiftui-when-using-stateobject-and-navigationview-4db91095d5af) which suggests doing it, but it's listed as defensive here too.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardTests.swift
```diff
@@ -78,4 +78,18 @@ final class PPOProjectCardTests: TestCase {
     try? await Task.sleep(nanoseconds: 10_000_000)
     assertSnapshot(matching: card, as: .image, named: "surveyAvailable")
   }
+
+  @MainActor
+  func testShortTemplateText() async {
```
amy-at-kickstarter, line 83: Adding a test for shorter strings to test string alignment.

amy-at-kickstarter: For what it's worth, I only touched _multiline_ string alignment, but I feel a little safer knowing this is here.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectDetails.swift
```diff
@@ -36,6 +36,7 @@ struct PPOProjectDetails: View {
               alignment: Constants.textAlignment
             )
             .lineLimit(Constants.titleLineLimit)
+            .multilineTextAlignment(Constants.multilineTextAlignment)
```
amy-at-kickstarter, line 39: Without this, the title was centered when it was multiline. Some implicit behavior of wrapping it in a button? 🤷‍♀️

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCard.swift
amy-at-kickstarter: Interesting article, thanks!

PR #2282, opened 2025-02-12
[MBL-2021] Disable "authenticate card" after completion of 3DS auth in PPO

stevestreza-ksr:
# 📲 What

After the "Authenticate Card" flow is complete, the button now disables.

# 🛠 How

PPOActionState was already being propagated into the PPOProjectCardViewModel, so I added a new "isComplete" flag to that, which is reacted to by the PPOProjectCard.


amy-at-kickstarter (approved):

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardViewModel.swift
```diff
@@ -35,6 +35,7 @@ typealias PPOProjectCardViewModelType = Equatable & Hashable & Identifiable & Ob
 final class PPOProjectCardViewModel: PPOProjectCardViewModelType {
   @Published private(set) var card: PPOProjectCardModel
   @Published var isLoading: Bool = false
+  @Published var isComplete: Bool = false
```
ifosli, line 38: Nit: Since these booleans are interconnected (at least, I'd assume the card can't be loading and complete at the same time), what do you think of doing an enum "state" here as well, instead of two separate bools? I'm okay with this as-is, though, if you prefer.

ifosli (approved):

PR #2283, opened 2025-02-12
PPO Fix missing empty address section

stevestreza-ksr:
# 📲 What

If a PPO card comes down in an "address locked" state but has no address, it currently does not render at all. This change makes it appear in the "take survey" state.

We don't expect this state to happen in production, but if it does, then we should show something rather than nothing, and filter this out server-side.

# 🛠 How

Previously we were setting no action for the card, which caused the initializer to fail, which hides it in the list. This changes the behavior to show the "take survey" action instead, which allows the initializer to succeed and the item ends up in the list.


ifosli (approved):
I'm okay with this workaround, but please follow along with https://kickstarter.slack.com/archives/C071SRX82K1/p1739472842233529/sync with Leigh so we can make sure iOS and Android are handling this in the same way. (Especially since we know the backend sends us these cards anytime the project enters address locking before the user completes their survey. It's a backend bug, but not one that's currently prioritized to be fixed.)

jovaniks (approved):

PR #2288, opened 2025-02-13
MBL-2111: Update tab bar on app foreground if PPO config flag has changed

amy-at-kickstarter:
# 📲 What

Update the tab bar when the app returns to foreground if the PPO config flag has changed.

# 🤔 Why

Our config flags are pulled from Firebase some time after app launch. For these flags to take effect, an app reboot was required, and that's not something we normally expect users to do. This meant users might not see PPO until a long time after they were bucketed into the rampup.

This refreshes the tab bar on app foreground. I chose to do this on app foreground, _instead_ of when the config is loaded, because we have no control over exactly when the new config data comes down from Firebase. This way, the tabs will be refreshed a little more frequently, but only when the user isn't looking.

https://github.com/user-attachments/assets/6358d81d-e74d-453f-bea3-9f910fcd945a


Library/ViewModels/RootViewModel.swift
```diff
@@ -88,7 +88,9 @@ public protocol RootViewModelInputs {
   /// Call when we should switch to the search tab.
   func switchToSearch()
 
-  /// Call when the a user locale preference has changed
+  /// Call when the a user locale preference has changed.
+  /// Used to blow away and re-generate all the tabs, so the entire app reflects changes in user language or currency.
+  /// See PR #576.
```
amy-at-kickstarter, line 93: I was wondering why this code was here - conveniently, blame mode pointed me at this very old PR.

Library/ViewModels/RootViewModel.swift
```diff
@@ -281,10 +297,14 @@ public final class RootViewModel: RootViewModelType, RootViewModelInputs, RootVi
       .filter { prev, next in prev == next }
       .map { $1 }
 
+    // FIXME: This signal always needs to happen when setViewControllers happens.
+    // Should they be combined into one signal with a tuple? Explicitly made dependent on one another?
+    // See https://kickstarter.atlassian.net/browse/MBL-2053
```
amy-at-kickstarter, line 302: I made some updates to this ticket - making it more of a general "what's up with `RootViewModel`" ticket. Some of the interdependent signals in here are thorny.

Library/ViewModels/RootViewModelTests.swift
```diff
@@ -427,15 +427,71 @@ final class RootViewModelTests: TestCase {
     self.vm.outputs.setViewControllers.map(extractRootNames)
       .observe(viewControllerNames.observer)
 
-    self.vm.inputs.viewDidLoad()
+    withEnvironment(language: .en, locale: Locale(identifier: "en")) {
```
amy-at-kickstarter, line 430: Making this test a little more true-to-life.

Library/ViewModels/RootViewModel.swift
```diff
@@ -149,18 +151,37 @@ public final class RootViewModel: RootViewModelType, RootViewModelInputs, RootVi
...
+
     let standardViewControllers = loginState.map { isLoggedIn -> [RootViewControllerData] in
       generateViewControllers(isLoggedIn: isLoggedIn)
     }
 
+    // We could detect when this feature changes in more places - like listening for notifications
+    // that the remote config loaded - but that might change the tab bar at a strange time.
+    // Updating this just on app foreground keeps the change invisible to the user.
+    // Conveniently, the activity badge is also updated on app foreground.
+    let ppoEnabledChanged = self.applicationWillEnterForegroundSignal.signal
```
stevestreza-ksr, line 167: Should this be `lifecycleEvents`?

stevestreza-ksr (approved):
LGTM pending one thing

Library/ViewModels/RootViewModel.swift
amy-at-kickstarter: I think the net effect would be the same (since `lifecycleEvents` just adds `viewDidLoad`), but I could update this for clarity/consistency.

amy-at-kickstarter: Whoops! Famous last words - this actually adds an extra event on `viewDidLoad`. So this is actually correct, it should be just `applicationWillEnterForeground` and _not_ `lifecycleEvents`. I just moved those up top to make the file a little more readable.

scottkicks (approved):
👍 nice updated tests!

PR #2290, opened 2025-02-18
Rename RewardAddOnNoShippingSelectionViewController, NoShippingPledgeRewardsSummaryDataSource, and WithShippingRewardsCollectionViewController

scottkicks:
# 🤔 Why

`NoShipping` classes are the standard and the only checkout classes now so we can remove this preface.

# 🛠 How

- Rename the view controllers and re-run view controller tests, if there are any, so that the snapshots have the correct naming.
- Rename view models and view model tests
   - Includes any protocols and objects these classes use with `NoShipping` in their name.

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] Tests pass


amy-at-kickstarter (approved):

scottkicks:
opening a new PR with a cleaner commit history. ran into some pbxproj wonkiness when rebasing `main`

PR #2293, opened 2025-02-18
[MBL-2118] Voiceover fixes for PPO

stevestreza-ksr:
# 📲 What

Changes some accessibility elements for PPO so the screen reads better with voiceover.

# 🛠 How

Some SwiftUI elements were marked as accessibility elements combining their children, and had a few traits specified.


amy-at-kickstarter (approved):
👍

ifosli (approved):
Nice! Just double-checking: you did test this on device? If not, can you do so before merging?

stevestreza-ksr:
Thanks! And yep they were tested on device with VoiceOver prior to the PR going up.

PR #2294, opened 2025-02-18
[MBL-2117] Add minimal large font support

ifosli:
# 📲 What

Prevent hardcoded strings from truncating at all and add more reasonable line limits to strings with user provided content where needed. Also, prevent tab bar titles from growing dynamically at all.

Note: This does NOT change layouts or add tests. I spent some time trying to do each of these things with limited success, so I'm going to submit this pr as is, and hope we'll have time to make this better as a fast follow.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2117)

| Default font size | Largest normal size | Largest accessible size |
| --- | --- | --- |
| image | image | Simulator Screen Recording - iPhone 16 - 2025-02-18 at 17 00 53 |
| image | image | Simulator Screen Recording - iPhone 15 Pro - 2025-02-18 at 17 05 23 |

# ♿️ Accessibility 

- [x] Supports Dynamic Type 

# ✅ Acceptance criteria

- [x] Tests pass
- [x] Large font experience is better than it was before (tested the cards that showed up for me, which is survey (incl one with an address showing) and fix payment) 

# ⏰ TODO

In a follow-up pr (possibly a separate ticket), I'd like to:
- Work with UX to figure out what we want accessible layouts to be (I'm thinking fewer hstacks) and implement these. 
- Add snapshot tests for dynamic sizes (neither the sizeCategory env variable nor the `dynamicTypeSize` modifiers did anything, so we'll need to do some framework level work to implement dynamic type support in our UI tests - possibly by explicitly use a hosting view controller and setting the size in UIKit.)
- I'd like us to have font size actually update dynamically instead of needing to restart the app to see changes in this view. Not sure if it's worth our time to implement if it ends up being tricky, but definitely worth looking into for an hour or so.

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCreator.swift
@@ -4,6 +4,7 @@ import SwiftUI
 
 struct PPOProjectCreator: View {
   let creatorName: String
+  @SwiftUI.Environment(\.sizeCategory) var sizeCategory
 
   var body: some View {
     HStack(alignment: .firstTextBaseline) {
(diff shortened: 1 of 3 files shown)


stevestreza-ksr (approved):

amy-at-kickstarter (approved):
LGTM.

Can you add a screenshot test with large font sizes to `PPOProjectCardTests`?

stevestreza-ksr:
I tried to add screenshot tests but due to how these views are setup, this isn't trivial, so I've [ticketed it](https://kickstarter.atlassian.net/browse/MBL-2122) for now.

PR #2295, opened 2025-02-19
Rename RewardAddOnNoShippingSelectionViewController, NoShippingPledgeRewardsSummaryDataSource, and WithShippingRewardsCollectionViewController

scottkicks:
# 🤔 Why

`NoShipping` classes are the standard and the only checkout classes now so we can remove this preface.

# 🛠 How

- Rename the view controllers and re-run view controller tests, if there are any, so that the snapshots have the correct naming.
- Rename view models and view model tests
   - Includes any protocols and objects these classes use with `NoShipping` in their name.

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] Tests pass


Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
amy-at-kickstarter, line 1: Is this change intentional

amy-at-kickstarter (approved):
LGTM, but can you double-check that swift package change? I assume it's something Xcode did automatically, but if not it should be taken out of this PR.

Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
scottkicks: oh that's weird. good catch!

jovaniks (approved):
LGTM

PR #2299, opened 2025-02-21
[MBL-2105] Rename WithShippingRewardsCollectionViewHeaderView, WithShippingRewardsCollectionViewController, WithShippingRewardsCollectionViewController & WithShippingRewardsCollectionViewModel

scottkicks:
# 🤔 Why

`WithShipping` classes are the standard and the only checkout classes now, so we can remove this preface.

# 🛠 How

- Rename views and remove any `WithShipping` prefaces

# 👀 See

No visible changes

# ✅ Acceptance criteria

- [x] Tests pass


Kickstarter-iOS/SharedViews/HeaderFooterViews/RewardsCollectionViewHeaderView.swift
ifosli, line 1: This one would also have been nice to have marked as deleted (and the "deleted" file marked as moved) but I don't care as much about this one, since it's a short file.

Kickstarter-iOS/Features/RewardAddOnSelection/Controller/RewardAddOnSelectionViewControllerTests.swift
```diff
@@ -49,6 +49,7 @@ final class RewardAddOnSelectionViewControllerTests: TestCase {
         let data = PledgeViewData(
           project: project,
           rewards: [reward],
+          bonusSupport: nil,
```
ifosli, line 52: Nit: not sure why we don't have them yet, but it'd be nice to have bonus support set for one of these tests, or add an extra test that makes sure bonus support is displayed correctly, if it's not too complicated to add.

ifosli (approved):

amy-at-kickstarter (approved):

Kickstarter-iOS/Features/RewardAddOnSelection/Controller/RewardAddOnSelectionViewControllerTests.swift
scottkicks: There are actually a few that do. I haven't dug into them so it's definitely possible that more would be valuable. The can make a note to check this out and address in a different ticket is so!

PR #2301, opened 2025-02-25
Fix empty case in backings dashboard

stevestreza-ksr:
# 📲 What

Fixes an error in backings dashboard that prevents the list from loading when there are 0 items.

# 🛠 How

Changes in #2291 added a check for preventing the list from refreshing unnecessarily (which causes jumping in the list). The issue was caused by the deduplication code, which prevents the screen from reloading unless the values from the paginator are different. However, this returns an empty array for unloaded/loading/error states as well as an empty list of values from the API, which prevents the results from exiting the "loading" case.

Instead, we only do this check if the data has actually loaded. So the check also compares if the results enum is in a "hasLoaded" state, which is already defined for these cases. This fixes the "loading" -> "empty" transition which shows the view correctly.

Also, this undoes some unit test changes that were added in #2291, because apparently I should've listened to those. 😅

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
@@ -113,7 +113,7 @@ final class PPOViewModel: ObservableObject, PPOViewModelInputs, PPOViewModelOutp
         }
       })
       .removeDuplicates(by: { left, right in
-        left.values == right.values
+        left.hasLoaded && right.hasLoaded && left.values == right.values
       })
       .receive(on: RunLoop.main)
       .sink(receiveValue: { results in
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
@@ -112,6 +112,9 @@ final class PPOViewModel: ObservableObject, PPOViewModelInputs, PPOViewModelOutp
           return false
         }
       })
+      // SwiftUI List jumps around if reloading frequently. This code path prevents
+      // unnecessary reloads if the data has loaded. If the values are the same and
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
```diff
@@ -113,7 +113,7 @@ final class PPOViewModel: ObservableObject, PPOViewModelInputs, PPOViewModelOutp
         }
       })
       .removeDuplicates(by: { left, right in
-        left.values == right.values
+        left.hasLoaded && right.hasLoaded && left.values == right.values
```
amy-at-kickstarter, line 116: Can you throw in a short explanatory comment here?

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModelTests.swift
amy-at-kickstarter, line 1: Can you throw in a unit test case specifically for the empty load case? Skimmed the test names and didn't see one.

amy-at-kickstarter (approved):
LGTM, two non-blocking suggestions

ifosli (approved):
I agree with Amy's suggestions but otherwise nice fix!

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModel.swift
stevestreza-ksr: Done!

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOViewModelTests.swift
stevestreza-ksr: Added one and verified that this test would have passed prior to #2291, failed with #2291, and passes now (so this test would have caught this issue).

PR #2302, opened 2025-02-25
MBL-2075: Add backingActionCount to PPOUserSettings

amy-at-kickstarter:
# 📲 What

Add `backingActionAction` to GraphQL and `PPOUserSettings`

# 🤔 Why

The updated PPO designs include a badge, with the action count, on the tab bar. Fetching this from the server will allow us to keep that tab bar badge up-to-date.

Note that this field currently isn't used anywhere; I primarily tested this with the ol' breakpoint method.
Screenshot 2025-02-25 at 11 20 45 AM

I'll make sure to add more tests when I'm actually hooking it up.


Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -226,15 +226,17 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
           return
         }
 
-        let ppoSettings = user
-          .flatMap { $0.me.ppoHasAction }
-          .flatMap { PPOUserSettings(hasAction: $0) }
```
amy-at-kickstarter, line 231: AFAIK nothing was relying on `PPOUserSettings` being `nil`. Tests still pass.

scottkicks (approved):

Library/PPOUserSettings.swift
```diff
@@ -2,12 +2,17 @@ import KsApi
 
 public struct PPOUserSettings: EncodableType, Decodable, Equatable {
   public let hasAction: Bool
+  public let backingActionCount: Int
```
stevestreza-ksr, line 5: Do we still need `hasAction`? Like is there ever a case where `backingActionCount > 0` but `hasAction = false`? Or `hasAction = true` but `backingActionCount == 0`?

stevestreza-ksr (approved):

Library/PPOUserSettings.swift
amy-at-kickstarter: I don't believe so - I'll be able to clean this up once the new code is hooked up.

PR #2305, opened 2025-02-26
Account for tab bar height in child EmptyStatesViewController in ActivitiesViewController

amy-at-kickstarter:
# 📲 What

Move code that adjusts the bottom of `EmptyStatesViewController` out of the view model, and into its parent, `ActivitiesViewController`. 

# 🤔 Why

`ActivitiesViewController` is a `UITableViewController` which means it has different behavior for its safe content insets.

I'd like to use `EmptyStatesViewController` elsewhere in the app, with the `.activity` state, but this extra layout step was making it difficult to have the empty VC have a _consistent_ layout when used elsewhere.

It still works well and looks good when used in `ActivitiesViewController`. Probably off by some vertical pixels somewhere or other, but I think it's fine.

Screenshot 2025-02-26 at 5 47 06 PM


Library/ViewModels/EmptyStatesViewModel.swift
```diff
@@ -79,7 +79,7 @@ public final class EmptyStatesViewModel: EmptyStatesViewModelType, EmptyStatesVi
       .ignoreValues()
 
     self.bottomLayoutConstraintConstant = emptyState
-      .map { Styles.grid(5) + ($0 == .activity ? 50.0 : 0) }
```
amy-at-kickstarter, line 82: `.activity` was only being used in `ActivitiesViewController`, sensibly enough.

Kickstarter-iOS/Features/EmptyStates/Controller/__Snapshots__/EmptyStatesViewControllerTests/testEmptyStates._activity_lang_de.png
amy-at-kickstarter, line 1: When used on its own, yes, it's 50 px lower. But used in the tab bar, it should have the same appearance.

Kickstarter-iOS/Features/Activities/Controller/ActivitiesViewController.swift
```diff
@@ -62,6 +62,10 @@ internal final class ActivitiesViewController: UITableViewController {
     let emptyVC = EmptyStatesViewController.configuredWith(emptyState: .activity)
     self.emptyStatesController = emptyVC
     emptyVC.delegate = self
+    // Because the ActivitiesViewController is a UITableViewController, it wasn't automatically accounting
+    // for the root tab bar height in this child VC. Adding the additional height makes this layout correctly.
+    emptyVC.additionalSafeAreaInsets = UIEdgeInsets(bottom: 50)
```
scottkicks, line 67: will this additional inset always be 50? I'm not positive that tab bar heights are the same across devices. 
Also wondering if it's worth having this in a Constants or some shared code in case we come across a similar situation in the future with a different UITableViewController. 

maybe I'm just over thinking this static CGFloat 😅

scottkicks (approved):
LGTM 👍 

Just a note about the static edge inset value.

Kickstarter-iOS/Features/Activities/Controller/ActivitiesViewController.swift
amy-at-kickstarter: Fair point! Let me see if I can do something smarter here, like actually get the tab bar height.

amy-at-kickstarter: Ha! I spent some time specifically using the tab bar height, and found a different layout issue. I actually think the 50 is to get this button to line up with the sign up button on the profile page. I updated the comment as such.

Since this is a one off, I don't feel bad about having it in here with a comment (instead of in a Constants file).

stevestreza-ksr (approved):

PR #2309, opened 2025-02-27
New design system colors

jovaniks:
# 📲 What

**Adaptive Colors** system to have colors organized, and dynamic color retrieval from `.xcassets`.

- Introduced a **protocol-based** approach to dynamically construct color names.
- Implemented structured namespaces (`background/`, `border/`, `text/`) for clarity.

# 🛠 How

- Created a `AdaptativeColors` protocol to handle color retrieval dynamically.
- Defined `Colors` struct with **namespaced enums** (`Background`, `Border`, `Text`).
- Implemented a single method (`adaptative()`) to fetch colors dynamically based on enum type.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2045)

image

# ♿️ Accessibility 

- [ ] Supports Dark Mode via `.xcassets`.
- [ ] Uses color contrast-aware asset catalog colors

# ✅ Acceptance criteria

- [ ] Colors are retrieved correctly based on the namespace (`background/`, `border/`, `text/`).
- [ ] No breaking changes to existing UI elements.

# ⏰ TODO

- [ ] Gather feedback from designers to ensure correct color mapping.


Library/Styles/Colors/AdaptiveColors.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    switch Self.self {
+    case is Colors.Background.Type:
+      namespace = "background/"
+    case is Colors.Border.Type:
+      namespace = "border/"
+    case is Colors.Text.Type:
+      namespace = "text/"
+    default:
+      namespace = ""
```
stevestreza-ksr, line 24: Based on this PR it looks like all the colors are defined in those three prefixes, which would suggest that we wouldn't expect the default case to be hit. Is that correct? If this should NOT be hit and we cant count on the compiler catching it, can we add an assertion failure here?

Library/Styles/Colors/AdaptiveColors.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    // Constructs the full color name for asset lookup.
+    let colorName = "\(namespace)\(self.rawValue)"
+
+    // Retrieves the color from the asset catalog, defaulting to white if unavailable.
+    return UIColor(named: colorName) ?? .white
```
stevestreza-ksr, line 31: Similarly add an assertion failure here if we don't get that named color.

stevestreza-ksr (changes requested):

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftFormat found issues:

| File | Rules |
| ---- | ----- |
| Library/Styles/Colors/AdaptiveColors.swift:29:1 | warning: (trailingSpace) Remove trailing space at end of a line. |
| Library/Styles/Colors/AdaptiveColors.swift:34:1 | warning: (trailingSpace) Remove trailing space at end of a line. |


</p>

Library/Styles/Colors/AdaptiveColors.swift
jovaniks: That's a good point! However, I think we should keep the default case without an `assertionFailure` because we might want to define new color categories outside of a namespace.

stevestreza-ksr: If we do that in the future, someone adding those colors would hit the assertion and they could remove it. In the current state of the code it looks like a mistake to add it here, so IMO it should get treated that way.

jovaniks: That's a good point. I see your reasoning here, and it makes sense to catch unexpected cases early rather than silently defaulting to an empty namespace. I'll go ahead and add the assertionFailure to make it more explicit when a new color category is introduced outside of the expected namespaces.

stevestreza-ksr (approved):

Library/Styles/Colors/AdaptiveColors.swift
```diff
@@ -0,0 +1,40 @@
...
+    // Constructs the full color name for asset lookup.
+    let colorName = "\(namespace)\(self.rawValue)"
+
+    // Retrieves the color from the asset catalog, defaulting to white if unavailable.
+    if let color = UIColor(named: colorName) {
+      return color
+    }
+
+    assertionFailure("⚠️ Color not found in asset catalog: \(colorName)")
+    return .white
```
amy-at-kickstarter, line 38: Nit: since most colors are used for text and UI elements, instead of backgrounds, it would be better if this was a color visible on the background. How about returning a default kickstarter green? But this is an edge case, I assume with the assert we'd never get here. 🤷‍♀️

Library/Styles/Colors/NewDesignSystemColors.swift
```diff
@@ -0,0 +1,40 @@
...
+    case action
+    case actionDisabled
+    case actionPressed
+    case dangerBold
+    case dangerBoldPressed
+    case dangerDisabled
+    case facebookDefault
+    case inverseDisabled
+    case inversePressed
+    case surfacePrimary
```
amy-at-kickstarter, line 19: What does `surface` mean here? Is this a background?

amy-at-kickstarter (approved):
LGTM, one non-blocking nitpick.

Library/Styles/Colors/NewDesignSystemColors.swift
jovaniks: That's correct, `surface` refers to a background color, which is why it's under the `Colors.Background` enum. The idea is to stay consistent with the naming conventions used in Figma.

image

Library/Styles/Colors/AdaptiveColors.swift
jovaniks: I was returning `.white` thinking of a neutral color, but I kind of agree — returning a green would likely be more visible in most cases. Which green are you thinking of, `.ksr_create_700`?

PR #2310, opened 2025-02-27
New design system buttons

jovaniks:
# 📲 What

This PR introduces the new **Design System Buttons** by creating a `KSRButton` subclass, which is fully configurable via `KSRButtonStyle`.

### 🔹 Key Changes:
- **Created `KSRButton`**: A new `UIButton` subclass for consistent button styling.
- **Introduced `KSRButtonStyle`**: Defines different button styles such as `.filled`, `.outlined`, `.borderless`, etc.
- **Added `KSRButtonStyleConfiguration` protocol**:
  - Ensures each button style provides **background color, title color, font, border properties, etc.**
  - Allows **seamless styling and state handling** (`normal`, `highlighted`, `disabled`).
- **Integrated with new Design System Colors (`AdaptiveColors`)** for **dynamic color adaptation**.
- **Added loading state support** with a centered `UIActivityIndicatorView`.

# 🛠 How

- `KSRButton` **subclasses `UIButton`** and applies styles via `KSRButtonStyle`.
- `KSRButtonStyle` **implements `KSRButtonStyleConfiguration`**, defining:
  - Background, border, text colors
  - States (`highlighted`, `disabled`, etc.)
  - Fonts and corner radius
- Uses **`AdaptiveColors`** for **Dark Mode & color consistency**.
- Introduces a **loading state** (`startLoading()`, `stopLoading()`).
- Uses **`UIButton.Configuration`** for modern button styling.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2045)

| Light mode ☀️ | Dark mode 🌑 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-02-27 at 11 11 01 | Simulator Screen Recording - iPhone 16 Pro - 2025-03-10 at 12 04 07 |

# ✅ Acceptance criteria

- [x] **New buttons match the Figma design**
- [x] **Works across all button states (`normal`, `highlighted`, `disabled`)**
- [x] **Loading state is displayed correctly**
- [x] **No breaking changes for existing buttons**

# ⏰ TODO

- [ ] Get **UX & Design Team feedback** on final implementation.
- [ ] Replace the current buttons with the new ones.
- [ ] Create the SwiftUI version of this


Kickstarter-iOS/Features/DesignSystem/Controller/DesignSystemViewController.swift
```diff
@@ -184,6 +184,39 @@ final class DesignSystemViewController: UIViewController {
     )
       |> ksr_addArrangedSubviewsToStackView()
 
+    // New Design System Fonts
```
ifosli, line 187: Nit: Should this comment be a `MARK`? And I'm guessing it should say new "buttons" not "fonts"

Library/Styles/Buttons/KSRButton.swift
```diff
@@ -0,0 +1,52 @@
...
+
+  private func setupConstraints() {
+    self.activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+    NSLayoutConstraint.activate([
+      self.activityIndicator.centerXAnchor.constraint(equalTo: self.centerXAnchor),
+      self.activityIndicator.centerYAnchor.constraint(equalTo: self.centerYAnchor)
+    ])
+  }
+
+  public func starLoading() {
```
ifosli, line 41: nit: `startLoading` not `starLoading`

Library/Styles/Buttons/KSRButton.swift
```diff
@@ -0,0 +1,52 @@
...
+  @available(*, unavailable)
+  required init?(coder _: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  private func setupButton() {
+    self.applyStyleConfiguration(self.style)
+
+    self.activityIndicator.hidesWhenStopped = true
+    self.activityIndicator.color = self.style.loadingIndicatorColor.mixDarker(0.35)
```
ifosli, line 27: Nit: Just making the color darker seems like it might not be a good solution across light mode and dark mode. Have you tested it for both?

Library/Styles/Buttons/KSRButtonStyle.swift
```diff
@@ -0,0 +1,125 @@
...
+public enum KSRButtonStyle: KSRButtonStyleConfiguration, CaseIterable {
+  case filled
+  case green
+  case filledInverted
+  case filledDestructive
+  case borderless
+  case outlined
+  case outlinedDestructive
+  case borderlessDestructive
+  case facebook
```
ifosli, line 13: Nit: Do we want to represent the button by two separate enums instead? Ie one enum for the color/button type (ex: green/black/facebook/destructive) and one for the button appearance (ex: filled/outlined/borderless)? Feel free to ignore this suggestion if this enum is already representing all the buttons we may want.

Library/Styles/Buttons/KSRButtonStyle.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  public var highlightedBackgroundColor: UIColor {
+    switch self {
+    case .filled: Colors.Background.actionPressed.adaptive()
+    case .green: Colors.Background.accentGreenBoldPressed.adaptive()
+    case .filledDestructive: Colors.Background.dangerBoldPressed.adaptive()
+    case .borderlessDestructive, .outlinedDestructive: Colors.Background.accentRedSubtle.adaptive()
+    case .facebook: self.backgroundColor.mixDarker(0.35)
```
ifosli, line 32: Same concern re: dark mode for the `mixDarker` and `mixLighter` in this class

Library/Styles/Fonts/Fonts.swift
```diff
@@ -134,6 +134,11 @@ extension UIFont {
   public static func ksr_bodyXXS(size: CGFloat? = nil) -> UIFont {
     return InterFont.bodyXXS.font(size: size)
   }
+
+  /// regular, 10pt size, TextStyle.caption1**
```
ifosli, line 138: Does this font change size based on dynamic type? Can you add a comment here specifying either way?

Kickstarter-iOS/Features/DesignSystem/Controller/DesignSystemViewController.swift
```diff
@@ -184,6 +184,39 @@ final class DesignSystemViewController: UIViewController {
...
+    let newButtons = UILabel()
+    newButtons.font = .ksr_title1().bolded
+    newButtons.textColor = adaptiveColor(.black)
+    newButtons.adjustsFontForContentSizeCategory = true
+    newButtons.text = "New Buttons"
+    self.buttonsStackView.addArrangedSubview(newButtons)
+
+    KSRButtonStyle.allCases.forEach { buttonStyle in
+      let button = KSRButton(style: buttonStyle)
+      button.setTitle("Button - \(String(describing: buttonStyle))", for: .normal)
```
ifosli, line 197: Can we make at least one of these buttons have a title that spans multiple lines? Buttons are usually just one line, but if they respond to dynamic type or if we get unlucky in other languages, they can have more text than we fit on one line, and it would be useful to see what that looks like.

ifosli (approved):
I left a bunch of comments, but overall this looks great! One pr nit: can you specify in your TODOs what you're planning to do in this pr/where other work is tracked (ie link JIRA tickets unless you're planning to add the work to this pr)?

Also, can you add a screencast for what this looks like in dark mode and with dynamic type, if the button font grows dynamically? I don't think dark mode needs to be a blocker yet, but it's useful to be aware of which parts of our app will work automatically and which will need more work.

Library/Styles/Fonts/Fonts.swift
jovaniks: Yes, this font will adjust based on Dynamic Type settings. I'll add a comment to clarify that.

Kickstarter-iOS/Features/DesignSystem/Controller/DesignSystemViewController.swift
jovaniks: I filed the multiple lines support on this ticket https://kickstarter.atlassian.net/browse/MBL-2184

Library/Styles/Buttons/KSRButtonStyle.swift
jovaniks: This enum is currently representing all the button types and appearances we need, so splitting it into two separate enums doesn't seem necessary at this point. However, we can revisit this if the button variations grow in the future.

Library/Styles/Buttons/KSRButtonStyle.swift
jovaniks: I tested both Light and Dark mode, and they look good. Here are the screenshots for reference:

| Light mode| Dark Mode |
| -- | -- |
| image | image |

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -0,0 +1,95 @@
...
+      UIConfigurationTextAttributesTransformer { [unowned self] config in
+        var newConfig = config
+        newConfig.font = styleConfig.font
+        newConfig.foregroundColor = self.tintColor
+        return newConfig
+      }
+
+    self.configuration?.imageColorTransformer = UIConfigurationColorTransformer { [unowned self] _ in
+      self.tintColor
+    }
```
stevestreza-ksr, line 68: Can you update these to only update `self.configuration` once? Something like

```swift
let buttonConfiguration = self.configuration
buttonConfiguration.contentInsets = buttonContentInsets
…
self.configuration = buttonConfiguration
```

Dont know if this causes re-renders for every change but at least would also remove the optionals in here.

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -0,0 +1,95 @@
...
+    case .highlighted:
+      self.configuration?.background.backgroundColor = styleConfig.highlightedBackgroundColor
+      self.tintColor = styleConfig.highlightedTitleColor
+      self.configuration?.background.strokeColor = styleConfig.highlightedBorderColor
+      self.configuration?.background.strokeWidth = styleConfig.borderWidth
+    default:
+      self.configuration?.background.backgroundColor = styleConfig.backgroundColor
+      self.tintColor = styleConfig.titleColor
+      self.configuration?.background.strokeColor = styleConfig.borderColor
+      self.configuration?.background.strokeWidth = styleConfig.borderWidth
```
stevestreza-ksr, line 92: Same note as above.

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    self.configuration?.imagePadding = Styles.grid(1)
+    self.configuration?.imagePlacement = .leading
+
+    self.configuration?.titleLineBreakMode = .byTruncatingMiddle
+    self.configuration?.titleAlignment = .center
+
+    self.tintColor = styleConfig.titleColor
+
+    self.configurationUpdateHandler = { [unowned self] _ in
```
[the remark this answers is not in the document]
jovaniks: There wasn't a specific reason for using `unowned` here, so I've updated it to weak to avoid potential crashes if self gets deallocated unexpectedly. Thanks for catching that! 👍

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
jovaniks: I've updated this to only update `self.configuration` once, following your suggestion. I also verified that it doesn’t cause any unnecessary re-renders. Thanks for the suggestion!

stevestreza-ksr (approved):

PR #2313, opened 2025-03-05
[SPC] MBL-2165: Connect GraphQL query for similar projects

stevestreza-ksr:
# 📲 What

Adds a new GraphQL query for Similar Projects carousel, along with a basic ProjectCardFragment to populate it. Also adds new functions to ServiceType, Service, and MockService to support querying a GraphQL query directly.

# 🤔 Why

We need data for SPC and this is how we get it.

We also are adding more direct querying of GraphQL to prevent needing to add more endpoints to ServiceType.

# 🛠 How

A new standard GraphQL query was added which includes a fragment for loading similar projects, and hooked into the existing plumbing in SimilarProjectsUseCase.

To support calling this, I also added a method to Service which allows someone to call a GraphQL query directly, along with support in MockService for mocking responses. 

# ✅ Acceptance criteria

- Tests should pass

### DIFF
--- a/KsApi/fragments/ProjectCardFragment.graphql
+++ b/KsApi/fragments/ProjectCardFragment.graphql
@@ -6,4 +6,16 @@ fragment ProjectCardFragment on Project {
   }
   name
   pid
+  state
+  isLaunched
+  deadlineAt
+  percentFunded
+  prelaunchActivated
+  launchedAt
+  goal {
+    ...MoneyFragment
+  }
+  pledged {
+    ...MoneyFragment
+  }
 }
--- a/KsApi/fragments/ProjectCardFragment.graphql
+++ b/KsApi/fragments/ProjectCardFragment.graphql
@@ -1,0 +1,9 @@
+# Minimal version of ProjectFragment, containing only the fields needed for showing project cards.
+fragment ProjectCardFragment on Project {
+  image {
+    id
+    url(width: 1024)
+  }
+  name
+  pid
+}
--- a/Library/UseCases/SimilarProjects/SimilarProject.swift
+++ b/Library/UseCases/SimilarProjects/SimilarProject.swift
@@ -1,5 +1,5 @@
 /// Represents a project that is similar to the currently viewed project.
 public protocol SimilarProject {
   /// The identifier for the project.
-  var pid: String { get }
+  var projectID: String { get }
 }
--- a/Library/UseCases/SimilarProjects/SimilarProject.swift
+++ b/Library/UseCases/SimilarProjects/SimilarProject.swift
@@ -1,5 +1,24 @@
+import Foundation
+import Kingfisher
+import KsApi
+
 /// Represents a project that is similar to the currently viewed project.
 public protocol SimilarProject {
   /// The identifier for the project.
   var projectID: String { get }
+
+  var image: Kingfisher.Source { get }
+
+  var name: String { get }
+
+  var isLaunched: Bool { get }
+  var isPrelaunchActivated: Bool { get }
+
+  var launchedAt: Date? { get }
+  var deadlineAt: Date? { get }
+  var percentFunded: Int { get }
+
+  var state: Project.State { get }
+  var goal: Money? { get }
+  var pledged: Money? { get }
 }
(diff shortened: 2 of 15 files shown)


Library/UseCases/SimilarProjects/SimilarProject.swift
```diff
@@ -1,5 +1,5 @@
 /// Represents a project that is similar to the currently viewed project.
 public protocol SimilarProject {
   /// The identifier for the project.
-  var pid: String { get }
+  var projectID: String { get }
```
amy-at-kickstarter, line 4: You may want to double-check if this gets you an integer ID or a base64-encoded GraphQL ID.

Library/UseCases/SimilarProjects/SimilarProjectsUseCaseTests.swift
```diff
@@ -8,69 +8,113 @@ final class SimilarProjectsUseCaseTests: TestCase {
   private var useCase: SimilarProjectsUseCase!
   private let projectTappedObserver = TestObserver<SimilarProject, Never>()
   private let similarProjectsObserver = TestObserver<SimilarProjectsState, Never>()
+  private var mockService: MockService!
 
   override func setUp() {
     super.setUp()
+
+    // Create mock data for similar projects
+    let mockProjectNodes: [GraphAPI.FetchSimilarProjectsQuery.Data.Project.Node?] = [
```
amy-at-kickstarter, line 17: One improvement that's been in the back of my head for a while - if we upgrade Apollo, we can get their [nice mock object framework](https://www.apollographql.com/docs/ios/testing/test-mocks#mocking-graphql-responses-with-generated-test-mocks). That will make it much easier for us to use GraphQL data in tests.

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -599,6 +599,16 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
             targetUserId: "\(blockedUserId)"
           )
       }
+
+    // MARK: Similar Projects
+
+    freshProjectAndRefTag
+      .map { "\($0.0.id)" }
```
amy-at-kickstarter, line 606: This one took me a sec to understand. Can you add an explanatory comment or decompose the arguments like:
```
.map { {(project, _) in "\(project.id)" }
```

More abstractly, I'm a little surprised `projectPageViewModel` doesn't just have a `project` signal, and that it's combined with the ref tag like this 🤔

amy-at-kickstarter (approved):
LGTM! I like this pattern and how it will keep the queries much closer to where they're used.

KsApi/fragments/ProjectCardFragment.graphql
scottkicks, line 1: We also need the following properties, I believe
- `project.state`
-` project.stats` (specifically the `goal` and `pledged`) and `project.dates.deadline`  to calculate the "percent funded" label and progress bar.
- `project.displayPrelaunch`, `project.prelaunchActivated`, and `project.dates.launchedAt` in order to determine if a project is a Prelaunch Project.

[designs for reference](https://www.figma.com/design/a9GQ3iQn7XzRUxbpE3cVpE/Search-%26-Discover?node-id=612-69397&t=xrV5LuLnIMLF6h93-4)

scottkicks (changes requested):
LGTM but I the Project Cards need access to a few more properties. I commented on which ones directly on the fragment file.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/ProjectPageViewModel.swift
stevestreza-ksr: Updated to `.map { project, _ in "\(project.id)" }`

I can refactor to a signal that only returns the project, but I don't know the full implications of the refTag and can't be sure that it won't break something, so I have skipped it for now.

KsApi/fragments/ProjectCardFragment.graphql
stevestreza-ksr: Properties have been added

scottkicks (approved):
Nice!

Library/UseCases/SimilarProjects/SimilarProject.swift
Arkariang: from graph
`pid` is the project ID without b64 encoding
`id` is the project ID with b64 encoding 
For concrete examples take a look here, we had a very similar conversation in android -> https://github.com/kickstarter/android-oss/pull/2246/files#r1981871824

PR #2314, opened 2025-03-06
[MBL-2141] Default pledge cta button and retry stack view to hidden

ifosli:
# 📲 What

When the pledge cta view is loaded but not yet configured with data, default the UI to show the loading indicator and nothing else.

I also updated all the lazy inits in this class to move away from Prelude while I was changing stuff here anyways.

# 🤔 Why

Before this change, whenever we'd open a project page with just the project id, the pledge cta would, by default, try to lay out all its possible UI states. Usually, this led to the loading indicator and the "retry" button showing. I considered updating the pledge cta config data to no longer require a project in the loading state, but this led to a lot of tedious test updates (and the pledge CTA still required a full `Project` class once loaded). Instead, I did this the quick and easy way, and set the pledge cta and the retry stack view to be hidden when they're initialized.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2141)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-03-06 at 10 32 19 | Simulator Screen Recording - iPhone 16 Pro - 2025-03-06 at 10 41 18 |

# ✅ Acceptance criteria

- [x] Loading state looks reasonable when project page is opened with just a project id
- [x] Pledge CTA still functions as expected when project page is opened with a `Project`

### DIFF
--- a/Kickstarter-iOS/SharedViews/PledgeCTAContainerView.swift
+++ b/Kickstarter-iOS/SharedViews/PledgeCTAContainerView.swift
@@ -27,37 +27,48 @@ final class PledgeCTAContainerView: UIView {
 
   private lazy var activityIndicator: UIActivityIndicatorView = {
     let indicator = UIActivityIndicatorView(frame: .zero)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+    indicator.translatesAutoresizingMaskIntoConstraints = false
     indicator.startAnimating()
     return indicator
   }()
 
   private lazy var activityIndicatorContainerView: UIView = {
-    UIView(frame: .zero)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+    let view = UIView(frame: .zero)
+    view.translatesAutoresizingMaskIntoConstraints = false
+    return view
   }()
 
   private(set) lazy var pledgeCTAButton: UIButton = {
-    UIButton(type: .custom)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+    let button = UIButton(type: .custom)
+    button.translatesAutoresizingMaskIntoConstraints = false
+    button.isHidden = true // Default to hidden until data is ready.
+    return button
   }()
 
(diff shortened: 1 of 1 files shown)


Kickstarter-iOS/SharedViews/PledgeCTAContainerView.swift
```diff
@@ -27,51 +27,64 @@ final class PledgeCTAContainerView: UIView {
 
   private lazy var activityIndicator: UIActivityIndicatorView = {
     let indicator = UIActivityIndicatorView(frame: .zero)
-      |> \.translatesAutoresizingMaskIntoConstraints .~ false
+    indicator.translatesAutoresizingMaskIntoConstraints = false
```
amy-at-kickstarter, line 30: Woohoo, one less Prelude operator!

amy-at-kickstarter, line 30: Woohoo, fewer Prelude operators! Thanks for cleaning this up.

amy-at-kickstarter (approved):
LGTM. 

One thing I'm wondering - is `retryStackViewIsHidden` not emitting a default value when the view is created? This may also need a fix in the VM level, as well.

ifosli:
> LGTM.
> 
> One thing I'm wondering - is `retryStackViewIsHidden` not emitting a default value when the view is created? This may also need a fix in the VM level, as well.

No, the VM doesn't emit anything until it's configured with data. When the `configureWithData` property is processed, it explicitly skips `nil`. I tried changing that to defaulting to the loading state in the VM, but it looks like a signal based on a mutable property only emits when that property changes. I was hoping it would emit a default value so I could just change this in the vm, but I couldn't get that to work, so I landed on this solution instead. It's definitely possible that I missed something and this could've been fixed on the VM level instead, but I'm pretty confident it doesn't need both. Good question, though!

stevestreza-ksr (approved):

PR #2315, opened 2025-03-06
Update search tests to use orthogonal combos

ifosli:
# 📲 What

Update search tests to use orthogonal combos (and delete newly obsolete snapshots), since this reduces the number of snapshots we need to deal with from 55 to 20.

# ✅ Acceptance criteria

- [x] Tests pass


amy-at-kickstarter (approved):
Nice!

PR #2319, opened 2025-03-10
MBL-2177: Hook up feature flag to new vs. old search controller

amy-at-kickstarter:
# 📲 What

Create a new search view controller and view model, prefixed with `LegacySearch*`, and set it in the tab bar controller based on the new Sort + Filters flag.

# 🤔 Why

Sort & Filters will require some extensive changes to the search view model. Because Search is one of our most important features, we want to easily be able to revert to the previous version of Search in case we introduce a critical bug.

This feature flag - and these legacy classes - will be cleaned up immediately after the launch of the V1 Sort & Filters project. New phases of this project will have new flags. 

# 🛠 How

Copy-and-paste.

# 👀 See
Thanks, @ifosli!


scottkicks (approved):
cool beans

PR #2321, opened 2025-03-10
Refactor BackerDashboardProjectCell to support GraphQL

amy-at-kickstarter:
# 📲 What

- Add the protocol `ProjectCellModel`
- Use a `ProjectCellModel` as input to `BackerDashboardProjectCellViewModel`
- Implement an extension on `Project` to implement `ProjectCellModel`
- Add a GraphQL fragment, `BackerDashboardProjectCellFragment`, which also implements `ProjectCellModel`
- Share UI code between `BackerDashboardProjectCell` and `MostPopularSearchProjectCell`

# 🤔 Why

This is one of the pieces that's been blocking us from using GraphQL in the Search flow. This refactoring will allow us to use a short fragment, `BackerDashboardProjectCellFragment`, to power our old Search UI, instead of the Project page fragment `ProjectFragment`.

# 🛠 How

I did my best to minimize collateral damage with this refactoring:

1. I made the protocol first, and made it match `Project` whenever possible.
2. Then I refactored the cell to use the protocol.
3. Then I added the new fragment, and implemented the protocol.

### DIFF
--- a/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
+++ b/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
@@ -2,10 +2,6 @@ import Foundation
 import KsApi
 
 extension GraphAPI.BackerDashboardProjectCellFragment: BackerDashboardProjectCellViewModel.ProjectCellModel {
-  public var id: Int {
-    return Int(self.projectId) ?? -1
-  }
-
   public var prelaunchActivated: Bool? {
     return Optional.some(self.projectPrelaunchActivated)
   }
--- a/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
+++ b/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
@@ -10,8 +10,8 @@ extension GraphAPI.BackerDashboardProjectCellFragment: BackerDashboardProjectCel
     return Project.State(rawValue: self.projectState.rawValue.lowercased()) ?? Project.State.live
   }
 
-  public var imageURL: String {
-    self.image?.url ?? ""
+  public var imageURL: String? {
+    self.image?.url
   }
 
   public var fundingProgress: Float {
--- a/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
+++ b/Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
@@ -1,0 +1,48 @@
+import Foundation
+import KsApi
+
+extension GraphAPI.BackerDashboardProjectCellFragment: BackerDashboardProjectCellViewModel.ProjectCellModel {
+  public var id: Int {
+    return Int(self.projectId) ?? -1
+  }
+
+  public var prelaunchActivated: Bool? {
+    return Optional.some(self.projectPrelaunchActivated)
+  }
+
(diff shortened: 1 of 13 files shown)


Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+import KsApi
+
+extension GraphAPI.BackerDashboardProjectCellFragment: BackerDashboardProjectCellViewModel.ProjectCellModel {
```
stevestreza-ksr, line 4: There's some cases here of optionals that can have invalid fields (like a blank `imageURL` field string, or a `-1` id). Can these be all optional, or use an intermediate type for validation, or something like that?

Library/ViewModels/BackerDashboardProjectCellViewModel.swift
```diff
@@ -47,14 +47,28 @@ public protocol BackerDashboardProjectCellViewModelType {
 
 public final class BackerDashboardProjectCellViewModel: BackerDashboardProjectCellViewModelType,
   BackerDashboardProjectCellViewModelInputs, BackerDashboardProjectCellViewModelOutputs {
+  public protocol ProjectCellModel {
+    var id: Int { get }
```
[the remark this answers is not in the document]
amy-at-kickstarter: Actually, I'll do you one better and delete this from the protocol - we don't actually use it! I put it in because it felt correct, but, eh, YAGNI.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/ViewModels/BackerDashboardProjectCellFragment+ProjectCellModel.swift
amy-at-kickstarter: Aha, it looks like `Project+ProjectFragment` bottoms out and returns `nil` when certain things (like the photo) aren't set: https://github.com/kickstarter/ios-oss/blob/main/KsApi/models/graphql/adapters/Project%2BProjectFragment.swift#L36

I'll update the protocol to be nullable and improve the handling in the cell itself.

amy-at-kickstarter:
Converting to draft while I write some tests for the GraphQL fragment, then will re-open.

stevestreza-ksr (approved):

Library/ViewModels/BackerDashboardProjectCellViewModel+Utilities.swift
```diff
@@ -0,0 +1,114 @@
+import Foundation
```
amy-at-kickstarter, line 1: `BackerDashboardProjectCellViewModel` and `MostPopularSearchProjectCellViewModel` both implemented these methods - with a couple teeny differences between them. I copied these out of `BackerDashboardProjectCellViewModel` and compared them to the ones in `MostPopularSearchProjectCellViewModel` to ensure correctness.

amy-at-kickstarter: Also note that `MostPopularSearchProjectCellViewModel` is used in the screenshot tests for `SearchViewController` so we should be pretty good on it being the same.

PR #2322, opened 2025-03-11
MBL-2177: Update search tab on app foreground

amy-at-kickstarter:
# 📲 What

Update the search tab on app foreground.

# 🤔 Why

Remote feature flags update a little after launch - _after_ the tab bar is created. This means users on a new version might not see the new Search experience until they reboot the app. This is the same issue we had with PPO, so I was able to re-use the same solution.


Library/ViewModels/RootViewModel.swift
```diff
@@ -10,15 +10,16 @@ public enum RootViewControllerData: Equatable {
   case discovery
   case activities
   case pledgedProjectsAndActivities
-  case search
+  case search(filtersEnabled: Bool)
```
amy-at-kickstarter, line 13: I re-used the existing `enum` case to simplify this change. Everywhere that changes for the enum case `.search` still works.

Library/ViewModels/RootViewModel.swift
```diff
@@ -164,10 +165,15 @@ public final class RootViewModel: RootViewModelType, RootViewModelInputs, RootVi
...
-      }.skipRepeats()
+    let featuredFlagChanged =
+      Signal.merge(
+        self.viewDidLoadProperty.signal,
+        self.applicationWillEnterForegroundSignal.signal
+      ).map { _ in
+        (featurePledgedProjectsOverviewEnabled(), featureSearchFiltersEnabled())
+      }
+      .skipRepeats { lhs, rhs in lhs.0 == rhs.0 && lhs.1 == rhs.1 }
+      .skip(first: 1) // Only fire if applicationWillEnterForeground changes the original values in viewDidLoadProperty.
```
amy-at-kickstarter, line 176: I noticed a bug here - `self.applicationWillEnterForegroundSignal.signal` would always fire at least once, causing an extra re-load even if the feature flag value hadn't changed. I made the tests slightly more specific and fixed this bug.

Library/ViewModels/RootViewModelTests.swift
```diff
@@ -471,6 +471,17 @@ final class RootViewModelTests: TestCase {
         nil,
         "Current user should have a cleared badge on activity"
       )
+
+      self.vm.inputs.applicationWillEnterForeground()
+
+      self.viewControllerNames.assertValueCount(
```
amy-at-kickstarter, line 477: This was the bug!

Library/ViewModels/RootViewModelTests.swift
```diff
@@ -599,6 +610,16 @@ final class RootViewModelTests: TestCase {
     self.filterDiscovery.assertValues([params])
   }
 
+  func testSwitchToSearch() {
```
amy-at-kickstarter, line 613: Adding this for good measure

Library/ViewModels/RootViewModel.swift
stevestreza-ksr: If the goal is to ignore the first `applicationWillEnterForegroundSignal` would it make more sense to move the `skip(1)` to that signal specifically?

stevestreza-ksr (approved):

scottkicks (approved):
LGTM 👍

PR #2323, opened 2025-03-12
[MBL-2162] Feature flag for SPC

stevestreza-ksr:
# 📲 What

This PR adds a feature flag for SPC

# 🛠 How

Usual approach of adding a RemoteConfig flag for it and function to access it.

The SimilarProjectsUseCase uses it to determine whether to load the projects, or just return a hidden state.

Tests were updated.


Library/UseCases/SimilarProjects/SimilarProjectsUseCaseTests.swift
```diff
@@ -11,13 +11,19 @@ final class SimilarProjectsUseCaseTests: TestCase {
 
   override func setUp() {
     super.setUp()
+
+    let remoteConfig = MockRemoteConfigClient()
+    remoteConfig.features["similar_projects_carousel"] = true
+    AppEnvironment.pushEnvironment(remoteConfigClient: remoteConfig)
```
amy-at-kickstarter, line 17: Nice, I like this approach to not doing this in every test.

amy-at-kickstarter (approved):
LGTM once tests are fixed; I'm guessing you need to update the screenshot test for the feature flag.

jovaniks (approved):

PR #2325, opened 2025-03-12
[MBL-2168] [SPC] Add navigation routing when tapping a project

stevestreza-ksr:
# 📲 What

Adds navigation to SPC, so tapping a cell opens a project page.

# 🛠 How

Already had the plumbing set up as part of #2311, this just hooks it up in the view controller. When the `navigateToSimilarProject` output fires, we push a new project page view controller onto the navigation stack.

# ✅ Acceptance criteria

Tests should pass.


Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
```diff
@@ -560,6 +560,23 @@ public final class ProjectPageViewController: UIViewController, MessageBannerVie
...
         messageBanner.showBanner(with: .error, message: Strings.Block_user_fail())
       }
+
+    self.viewModel.outputs.navigateToSimilarProject
+      .observeForUI()
+      .observeValues { [weak self] project in
+        guard let self else { return }
+        let vc = ProjectPageViewController.configuredWith(
+          projectOrParam: Either<Project, Param>.right(.id(project.projectID)),
+          refInfo: RefInfo(.projectPage)
```
stevestreza-ksr, line 570: Making an assumption here on what this should be, as I have no idea how this gets used

scottkicks: I would think that the RefInfo would be used specifically for tracking analytics, but I'm not sure if it's in use right now.

scottkicks (approved):
👍

Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
amy-at-kickstarter: If you dig into `ProjectPageViewModel`, we eventually log the ref tag down here as the referer: https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/ProjectPageViewModel.swift#L396-L403

Seems worth asking if we want a new ref tag for similar projects, otherwise this seems fine to me.

amy-at-kickstarter (approved):
LGTM

Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
stevestreza-ksr: Added a todo for the analytics ticket in [96689dfb0](https://github.com/kickstarter/ios-oss/pull/2325/commits/c7e35406338dbfc28d7077929d8c3fff202e474e)

PR #2331, opened 2025-03-18
Add FetchCategoriesUseCase

amy-at-kickstarter:
# 📲 What

Add `FetchCategoriesUseCase` and implement it in Discover.

# 🤔 Why

This will allow us to share category loading/caching between Discovery and Search. No point in re-writing the wheel, etc.


Library/ViewModels/SimilarProjectsCardViewModel.swift
```diff
@@ -43,9 +43,9 @@ public final class SimilarProjectsCardViewModel: SimilarProjectsCardViewModelTyp
 
     self.projectImageSource = project.map { $0.image }
 
-    self.projectStatus = project.map(projectStatus(for:))
+    self.projectStatus = project.map(self.projectStatus(for:))
```
amy-at-kickstarter, line 46: Picked up by `swiftformat`.

Library/UseCases/FetchCategoriesUseCaseTests.swift
```diff
@@ -0,0 +1,108 @@
...
+    AppEnvironment.current.cache[KSCache.ksr_discoveryFiltersCategories] = cachedCategories
+
+    self.initialObserver.send(value: ())
+
+    self.loadingIndicatorIsVisible.assertDidNotEmitValue()
+    self.categories.assertDidEmitValue()
+    XCTAssertEqual(self.categories.lastValue?.count, 3, "Categories should load directly from cache")
+  }
+
+  func test_fetchError_loadsCategories_WhenInitialSignalSentAgain() {
```
amy-at-kickstarter, line 70: The discovery page is kinda buggy when there's an error - I fixed one small bug, in which the loader never disappeared. I added this test to document that the use case works as expected, even though the _view model_ would need a little refactoring to have nicer error handling. That's out of scope right now.

Library/ViewModels/DiscoveryFiltersViewModel.swift
```diff
@@ -75,40 +75,18 @@ public final class DiscoveryFiltersViewModel: DiscoveryFiltersViewModelType,
       params.map { p in SelectableRow(isSelected: p == selectedRow.params, params: p) }
     }
 
+    self.categoriesUseCase = FetchCategoriesUseCase(initialSignal: self.viewDidLoadProperty.signal)
+
     let categoryId = self.initialSelectedRowProperty.signal.skipNil()
       .map { $0.params.category?.rootId }
 
-    let loaderIsVisible = MutableProperty(false)
```
amy-at-kickstarter, line 81: All copy-pasted into the use case, with a few minor changes.

Library/UseCases/FetchCategoriesUseCase.swift
```diff
@@ -0,0 +1,65 @@
...
+            loaderIsVisible.value = true
+          })
+          .map { (envelope: RootCategoriesEnvelope) in envelope.rootCategories }
+          .materialize()
+      }
+
+    self.loadingIndicatorIsVisible = Signal.merge(
+      loaderIsVisible.signal,
+      categoriesEvent.values().mapConst(false),
+      categoriesEvent.errors().mapConst(false)
```
amy-at-kickstarter, line 45: This is my one tweak from copy-paste - hide the loading spinner if the categories fail to load.

Library/ViewModels/SimilarProjectsCardViewModel.swift
scottkicks: 🤔 this is a little strange. It looks like this is the reason for the failing library tests. Probably has to do with how our formatting views methods with parameters being used in this way. Might need to break it out into brackets or something. I'd have to dig into the formatting code, though 😬

stevestreza-ksr (approved):

PR #2333, opened 2025-03-18
[MBL-2169] [SPC] Adds similar_projects as session_ref_tag for analytics

stevestreza-ksr:
# 📲 What

This updates the navigation behavior for Similar Projects carousel to add the necessary telemetry for tracking efficacy.

# 🤔 Why

# 🛠 How

Most of the tracking is implemented already as long as the `session_ref_tag` is set properly on the Project Page. So when opening a new Project Page from the Similar Projects carousel, this sets that RefTag on the new Project Page, and that carries through checkout flow.

# ✅ Acceptance criteria

Should be able to see `session_ref_tag = similar_projects` when opening an SPC carousel item (but that's not fully working right now).

You can fake a navigation event by enabling the SPC feature flag and adding this to `SimilarProjectsUseCase.swift`'s init method:

```swift
    self.similarProjects.signal
      .compactMap { state -> (any SimilarProject)? in
        guard case let .loaded(projects) = state else {
          return nil
        }
        return projects.randomElement
      }
      .take(first: 1)
      .delay(3, on: AppEnvironment.current.scheduler)
      .filter { _ in Bool.random() }
      .observeForUI()
      .observeValues { project in
        self.projectTapped(project: project)
      }
```

This will endlessly simulate navigation to a similar project.


scottkicks (approved):

amy-at-kickstarter (approved):
LGTM.

Might be helpful to throw in a unit test that validates this, but that's not a blocker to me.

PR #2335, opened 2025-03-19
[MBL-2176] Add SortView and SortViewModel to handle sorting options in SwiftUI

jovaniks:
# 📲 What

Added `SortView` and `SortViewModel` to handle sorting options using SwiftUI.  
Introduced `BottomSheetPresenter` to display `SortView` as a bottom sheet with dynamic height adjustment.

# 🛠 How

- Created `SortView` to display sorting options using a `List`.  
- Created `SortViewModel` to handle state and user interactions with Combine.  
- Introduced `SortOption` enum to define available sorting options.  

### **BottomSheetPresenter**
- Created `BottomSheetPresenter` to present views as bottom sheets with dynamic height based on content.  
- Implemented `BottomSheetPresentationController` using `UIPresentationController`:
  - Adjusts height dynamically using `systemLayoutSizeFitting` for precise content-based sizing.  
  - Applies a corner radius on the top corners of the sheet.  
  - Added a `dimmingView` with an alpha background to visually separate the sheet from the background.  

 This presenter will be used to present any **SwiftUI** view or **UIKit-based** view as a bottom sheet with dynamic sizing.

 Usage example:
 ```swift
 let view = SortView(viewModel: viewModel)
 let hostingController = UIHostingController(rootView: view)
 let presenter = BottomSheetPresenter()
 presenter.present(viewController: hostingController, from: self)
 ```

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MLB-2176) 
- [Figma](https://www.figma.com/design/a9GQ3iQn7XzRUxbpE3cVpE/Search-%26-Discover?node-id=617-90097&t=P1HxGJPKIOMtUEke-4)

Simulator Screen Recording - iPhone 16 Pro - 2025-03-19 at 18 19 49

# ✅ Acceptance criteria

- [x] Sorting options update correctly upon user interaction.  
- [x] State reflects the selected sorting option.  
- [x] Closing the view triggers the expected callback. 
- [x] Bottom sheet size dynamically adjusts to content height.  
- [x] Background dims when the sheet is presented. 

# ⏰ TODO

- [x] Finalize translations for `SortOption` names and other strings ([MLB-2204](https://kickstarter.atlassian.net/browse/MLB-2204))
- [ ] Test dynamic resizing with different content sizes.


Kickstarter-iOS/Features/SortFilter/SortView/SortView.swift
```diff
@@ -0,0 +1,112 @@
...
+
+      self.sortOptionList
+
+      Spacer()
+    }
+    .background(Colors.Background.surfacePrimary.swiftUIColor())
+
+    // Handle actions
+    .onReceive(self.viewModel.selectedSortOption) { sortOption in
+      self.onSelectedSort?(sortOption)
```
amy-at-kickstarter, line 21: Same nit as the other one, this should also capture `[weak self]`

amy-at-kickstarter (changes requested):
One change - would it be possible to make this sheet resize based on the size of its content? Sorry, you asked me and I gave you the wrong answer! The sort sheet needs to resize (since it will only have 5 or 6 options); the category filter sheet will have more options, so it doesn't matter if it resizes.

amy-at-kickstarter (approved):

Kickstarter-iOS/SharedViews/Presenters/BottomSheetPresenter.swift
```diff
@@ -0,0 +1,110 @@
+import Library
```
amy-at-kickstarter, line 1: I'm actually really surprised this requires so much custom code to get a bottom sheet of a specific size. I played around a little bit with `preferredContentSize` but didn't have any luck. We're sure that Apple doesn't support this, and we need a custom presenter?

amy-at-kickstarter: Aha, thank you for the investigation! Based on this - I think you should ship `BottomSheetPresenter`, but we should only use it as a fallback for iOS 15, as you suggest. That way, when we drop iOS 15 (presumably soon, there's a ticket for it), it's obvious we can clean up `BottomSheetPresenter`.

amy-at-kickstarter: Also `Detent` is my new vocabulary word today!

Kickstarter-iOS/SharedViews/Presenters/BottomSheetPresenter.swift
```diff
@@ -0,0 +1,153 @@
...
+  /// - For iOS 16+, it uses `UISheetPresentationController` to support dynamic height detents.
+  /// - For earlier versions, it falls back to a custom `UIPresentationController`.
+  ///
+  /// - Parameters:
+  ///   - viewController: The view controller to present.
+  ///   - parentViewController: The view controller that presents the bottom sheet.
+  func present(viewController: UIViewController, from parentViewController: UIViewController) {
+    // If running on iOS 16+, use UISheetPresentationController for a modern sheet presentation.
+    if #available(iOS 16.0, *), viewController.sheetPresentationController != nil {
+      self.sheetPrensent(viewController: viewController, from: parentViewController)
```
jovaniks, line 34: I updated the implementation to use `UISheetPresentationController` with a custom detent for iOS 16+. This allows us to define a dynamic height directly through the `sheetPresentationController`.
For devices running iOS 15, the `BottomSheetPresenter` is still being used as a fallback to ensure compatibility.

Kickstarter-iOS/SharedViews/Presenters/BottomSheetPresenter.swift
jovaniks: I’ve already made the changes in `BottomSheetPresenter` to centralize the presentation logic. It now handles the validation internally, it’ll use `UISheetPresentationController` for iOS 16+ and fall back to `BottomSheetPresenter` for iOS 15 and earlier.

scottkicks (approved):
LGTM! Good call with the fallback solution for iOS 15. I was coming to the same conclusion when I was looking into building a bottom sheet for Backer Dashboard V2!

PR #2338, opened 2025-03-19
Rename methods that swiftformat keeps breaking

amy-at-kickstarter:
# 📲 What

Rename two helper methods in `SimilarProjectsCardViewModel`.

# 🤔 Why

These methods matched variable names in `SimilarProjectsCardViewModel`; `swiftformat` kept trying to be clever and rename them to `self.method`, which breaks (because there is no such thing).


scottkicks (approved):
🎉

PR #2340, opened 2025-03-19
[MBL-2166] Wire Up New SPC Cell To Project Page Table View

scottkicks:
# 🤔 Why

Now that the SPC UI cells and project card are ready to be used, we can get them into the ProjectPage table view.

# 🛠 How

- Configures the new cell in `ProjectPageViewControllerDataSource` 
- Updates `ProjectPageViewModel` and `ProjectPageViewController` to load the SimilarProjects fetched via `SimilarProjectsUseCase` into the datasource/cell

- Also makes sure that tapping a card is wired into `SimilarProjectsUseCase.similarProjectTapped`

# 👀 See

| Project Cards Loaded | Project Cards Being Tapped |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-03-18 at 16 13 58 | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-03-19 at 17 33 02 |

# ✅ Acceptance criteria

- [x] SPC Carousel displays  on Project Pages if there isn't an error or the state is .hidden
- [x] Tapping a card uses the SPC UseCase to present a new Project Page for the selected project.

# ⏰ TODO

- [ ] page control


Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewController.swift
```diff
@@ -938,6 +942,16 @@ extension ProjectPageViewController: UITableViewDelegate {
...
+  public func tableView(_: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+    let cell = self.dataSource.items(in: indexPath.section)[indexPath.row]
+
+    if cell.reusableId == SimilarProjectsTableViewCell.defaultReusableId {
+      return SimilarProjectsCellConstants.collectionViewHeight
+    }
+
+    return UITableView.automaticDimension
+  }
+
```
scottkicks, line 954: The tableview is having trouble understanding it's content view's height when it's child is a collectionview. The best workaround I've come up with is hardcoding the collectionview height.

If any one else wants to dive deeper here with me to find a better solution I'm for it!

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -493,7 +497,7 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
         .skipNil()
 
     // We skip the first one here because on `viewDidLoad` we are setting .overview so we don't need a useless emission here
-    self.updateDataSource = self.projectNavigationSelectorViewDidSelectProperty.signal
+    let dataSourceUpdate = self.projectNavigationSelectorViewDidSelectProperty.signal
```
scottkicks, line 500: I broke out the main `updateDataSource` logic (that configures the rest of the Project Page's tableview) into it's own dataSourceUpdate property to make it easier to combine with `self.similarProjectsUseCase.similarProjects.signal`.

scottkicks:
I'm putting this up for review while I work on fixing some pesky tests.

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -527,6 +531,11 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
       }
       .skip(first: 1)
 
+    self.updateDataSource = Signal.combineLatest(
+      dataSourceUpdate,
+      self.similarProjectsUseCase.similarProjects.signal
+    )
```
scottkicks, line 537: I'm actually not sure we want to couple these two signals together. If the similarProjects signal doesn't complete for some reason we'd still want the normal datasource data to still load.

Anyone have suggestions on how we can make sure that happens, and also if the similar projects signal completes as expected, `self.updateDataSource` still gets emitted?

amy-at-kickstarter: Yeah - another (very similar) approach would be to put that logic inside `similarProjectsUseCase` so that it always emits an empty array or a nil for `similarProjects` on `viewDidLoad`. I've been running into this issue frequently with use cases, and that's where I usually end up.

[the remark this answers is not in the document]
amy-at-kickstarter: Oh man, these long tuple signals are a peeve of mine. If there happens to be time for a little refactoring, I suggest turning it into a `struct` so the whole thing is typed much more cleanly.

amy-at-kickstarter (changes requested):
Requesting a couple changes before this goes out - cleaning up the long tuple (with either a `map` or a `struct`, revisiting the tap gesture, making sure scrolling is handled for all `SimilarProjectState`s, and adding a test to the datasource tests.

Kickstarter-iOS/Features/ProjectPage/SimilarProjects/Cells/SimilarProjectsCollectionViewCell.swift
```diff
@@ -10,15 +14,25 @@ import UIKit
...
+
   // MARK: - Lifecycle
 
   override init(frame: CGRect) {
     super.init(frame: frame)
 
     self.configureViews()
     self.bindStyles()
+
+    let gesture = UITapGestureRecognizer(target: self, action: #selector(self.handleTap(_:)))
```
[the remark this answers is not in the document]
scottkicks: Yeah the challenge is specifically around passing the project that was tapped up from the collectionview to the ProjectPageViewController and in to SimilarProjectsUseCase. Because of the datasource pattern the table view is using this is how we need to do it I think. 

I guess another way would be to store the loaded similar projects in the Use case and then use indexPath.row to identify which one was tapped.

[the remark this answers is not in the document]
amy-at-kickstarter: That pattern would be more what I expect! If everything is working fine, and the tap recognizer works, I can live with this - but you do lose nice things like having an automatic 'selected' state happen.

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -527,6 +531,21 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
+          .takeWhen(self.viewDidLoadProperty.signal)
+      )
+
+    self.updateDataSource = Signal.combineLatest(
+      dataSourceUpdate,
+      similarProjectsState
+    )
+    .map { dataSource, similarProjects in
+      let (navSection, project, refTag, initialIsExpandedArray, urls) = dataSource
+      return ((navSection, project, refTag, initialIsExpandedArray, urls), similarProjects)
```
amy-at-kickstarter, line 546: Ah, this change doesn't address what I see as the larger issue - nested tuples. This signal should either be a struct or a _flat_ tuple like `(navSection, project, refTag, initialIsExpandedArray, urls, similarProjects)`.

amy-at-kickstarter (changes requested):
Requesting a change to the datasource output before this goes out.

stevestreza-ksr (approved):

PR #2341, opened 2025-03-20
MBL-2182: Create SearchFiltersUseCase

amy-at-kickstarter:
# 📲 What

Create a use case for showing Sort and Filter UI from the Search flow.

# 🤔 Why

This keeps the code nice and neat - pulling all the sort and filter logic into its own class, with its own tests.

# 🛠 How

It's worth noting that I'm using some placeholder model objects, `SearchSortSheet` and `SearchFilterCategoriesSheet`. When I finish integrating everything, these will be replaced with model objects for the new UI that @jovaniks has written.


Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -0,0 +1,189 @@
...
+      self.selectedSortIndexProperty.producer.takeWhen(initialSignal),
+      self.selectedSortIndexProperty.signal
+    )
+    .map { [sortOptions] idx in
+      sortOptions[idx]
+    }
+
+    // This bit of complication is here because we want this use case to emit a
+    // selectedCategory of nil on the initial signal - even if no categories have been downloaded yet.
+    // So if the index is nil, just immediately emit nil.
```
amy-at-kickstarter, line 84: This comment is backed up by a unit test.

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -0,0 +1,189 @@
...
+  * `showSort` - Sends a model object which can be used to display sort options.
+  * `showCategoryFilters` - Sends a model object which can be used to display category filter options. Only sends if `categories` have also been sent.
+ */
+
+public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFiltersUseCaseInputs,
+  SearchFiltersUseCaseUIOutputs, SearchFiltersUseCaseDataOutputs {
+  public init(initialSignal: Signal<Void, Never>, categories: Signal<[KsApi.Category], Never>) {
+    self.categoriesProperty <~ categories
+
+    self.showCategoryFilters = self.selectedCategoryIndexProperty.producer
```
amy-at-kickstarter, line 54: Worth noting - I wrote this use case to take indexes as inputs, instead of taking a model object as input like `selectedCategory(_ category: KsApi.Category)`. I thought that'd make everything a little more loosely coupled, so it will be easier to test and change later. 

In retrospect, I'm not sure this was worth the effort and extra complexity. I may refactor this out in another phase (or when I integrate with the actual UI that @jovaniks wrote.)

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -0,0 +1,189 @@
...
+    .map { [sortOptions] idx in
+      sortOptions[idx]
+    }
+
+    // This bit of complication is here because we want this use case to emit a
+    // selectedCategory of nil on the initial signal - even if no categories have been downloaded yet.
+    // So if the index is nil, just immediately emit nil.
+
+    let selectedCategoryIndex = Signal.merge(
+      self.selectedCategoryIndexProperty.producer.takeWhen(initialSignal),
```
amy-at-kickstarter, line 87: A little trick to make the default value of a `MutableProperty` emit when the page loads. Otherwise, `someMutableProp.signal` only emits _changes_, not the initial value.

Library/UseCases/SearchFiltersUseCaseTests.swift
```diff
@@ -0,0 +1,134 @@
...
+    self.useCase.uiOutputs.showCategoryFilters.observe(self.showCategoryFilters.observer)
+    self.useCase.uiOutputs.showSort.observe(self.showSort.observer)
+  }
+
+  func test_category_onInitialSignal_isNil() {
+    self.selectedCategory.assertDidNotEmitValue()
+
+    self.initialObserver.send(value: ())
+
+    self.selectedCategory.assertLastValue(nil)
```
amy-at-kickstarter, line 37: One of my goals has been to make my use cases have very clear initial and default behavior - I've been burned _way_ too many times by a `Signal.combineLatest` which is missing some important values.

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -0,0 +1,189 @@
...
+  * `selectedSortOption(atIndex:)` - Call this when the user selects a new sort option.
+  * `selectedCategory(atIndex:)` - Call this when the user selects a new category.
+
+ Data Outputs:
+  * `selectedSort` - The currently selected sort option. Defaults to `.popular`. Default value only sent after `initialSignal` occurs.
+  * `selectedCategory` - The currently selected category. Defaults to nil. Default value only sent after `initialSignal` occurs.
+
+ UI Outputs:
+  * `showSort` - Sends a model object which can be used to display sort options.
+  * `showCategoryFilters` - Sends a model object which can be used to display category filter options. Only sends if `categories` have also been sent.
```
stevestreza-ksr, line 46: Can these be moved to be comments on the functions in the inputs/outputs? They'll show up nice when you option-click those functions.

Library/UseCases/SearchFiltersUseCase.swift
stevestreza-ksr: Using model objects over indexes is IMO always gonna be more resilient. Make it the compiler's problem to check for valid values! 😎

stevestreza-ksr (approved):

Library/UseCases/SearchFiltersUseCase.swift
amy-at-kickstarter: Yeah, fair! I think what I've been burned by is situations where we couple our UI way too closely to a model object (I'm looking at you, `BackerDashboardProjectViewCell`), and it's turned me into a paranoiac.

amy-at-kickstarter: @stevestreza-ksr You're gonna love my updated implementation coming soon, because it has not only the correct model objects, but a very lovely protocol-and-generics implementation. 😎

PR #2345, opened 2025-03-21
Make FilterCategoryView generic

amy-at-kickstarter:
# 📲 What

Make `FilterCategoryView` and `FilterCategoryViewModel` take a protocol model object, `FilterCategory`, instead of a concrete object.

# 🤔 Why
Much of our UI is tightly coupled to V1 API objects. I'd like to avoid that for Sort and Filter, because ultimately it's a GraphQL based feature - we should be using GraphQL search params and categories. However, right now we're slightly stuck with the V1 objects, for expediency reasons.

This will make it really easy to directly pass a `KsApi.Category` into the `FilterCategoryView` - but will also make it easy to refactor later.

More broadly, I'd like our UI to be less API-coupled in general, and this is a solid approach.

This can be used easily by adding `extension KsApi.Category: FilterCategory`.

See also: #2346


Kickstarter-iOS/Features/SortFilter/FilterCategory/FilterCategoryViewModel.swift
```diff
@@ -76,20 +82,27 @@ class FilterCategoryViewModel: FilterCategoryViewModelType {
...
-  func isCategorySelected(_ category: FilterCategory) -> Bool {
+  func isCategorySelected(_ category: T) -> Bool {
     return self.currentCategory?.id == category.id
   }
 }
 
-struct FilterCategory: Identifiable, Equatable {
-  let id: String
-  let name: String
+internal enum ConcreteFilterCategory: String, FilterCategory, CaseIterable {
```
amy-at-kickstarter, line 94: This is just here so that the tests and previews have something to work with. It's internal because it's not meant to be used anywhere else.

jovaniks (approved):
Looks great! Refactoring to use a protocol model makes total sense, it'll definitely help with future flexibility and reducing API coupling. Nice work! 👌

scottkicks (approved):
Cool 😎

stevestreza-ksr (approved):

PR #2346, opened 2025-03-21
Make SortView generic

amy-at-kickstarter:
# 📲 What

Make `SortView` and `SortViewModel` take a protocol model object, `SortOption`, instead of a concrete object.

# 🤔 Why
Much of our UI is tightly coupled to V1 API objects. I'd like to avoid that for Sort and Filter, because ultimately it's a GraphQL based feature - we should be using GraphQL search params and categories. However, right now we're slightly stuck with the V1 objects, for expediency reasons.

This will make it really easy to directly pass a `DiscoveryParams.Sort` into the `SortView` - but will also make it easy to refactor later.

More broadly, I'd like our UI to be less API-coupled in general, and this is a solid approach.

This can be used easily by adding `extension DiscoveryParams.Sort: SortOption`.

See also: #2345


jovaniks (approved):
Nice!

stevestreza-ksr (approved):

PR #2348, opened 2025-03-21
MBL-2182: Hook up real UI for Sort and Filter modal sheets

amy-at-kickstarter:
# 📲 What

* Fix `SearchFiltersUseCase` so it actually uses `KsApi.Category` and `DiscoveryParams.Sort`, instead of passing around indices to arrays
* Use the real UI, `SortView` and `FilterCategoryView`, instead of `UIAlertController` to select sort options

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests


Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -199,36 +200,52 @@ internal final class SearchViewController: UITableViewController {
...
+      viewModel.selectCategory(selectedCategory)
     }
 
-    controller.addAction(UIAlertAction(title: "Cancel", style: .cancel))
-
-    controller.popoverPresentationController?.sourceView = self.currentSortAndFilterHeader
+    let filterView = FilterCategoryView(
+      viewModel: viewModel,
+      onSelectedCategory: { [weak self] category in
+        self?.viewModel.inputs.selectedCategory(category)
```
amy-at-kickstarter, line 237: Slight divergence from the original plans - we're updating the search every time a category is selected, not just on 'See Results'. I think this is pretty harmless for now, and it keeps the view model a little more simple.

Kickstarter-iOS/Features/Search/Extensions/Category+FilterCategory.swift
```diff
@@ -0,0 +1,16 @@
...
+extension KsApi.Category: @retroactive Identifiable {}
+// Implements FilterCategory so KsApi.Category can be used in a FilterCategoryView.
+extension KsApi.Category: FilterCategory {
+  public var id: Int {
+    guard let intId = self.intID else {
+      assert(
+        false,
+        "Category is missing an int identifier. It may not display correctly in the filter sheet."
+      )
+      return -1
```
amy-at-kickstarter, line 12: I don't love this default behavior. It also shouldn't ever happen, since we're fetching the categories from GraphQL and they'll always include an ID.

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -40,65 +40,31 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
-      .map { [sortOptions] selectedIdx -> SearchSortSheet in
-        let names = sortOptions.map(sortOptionName(from:))
-
-        return SearchSortSheet(sortNames: names, selectedIndex: selectedIdx)
+      .map { [sortOptions] sort -> SearchSortSheet in
+        SearchSortSheet(sortOptions: sortOptions, selectedOption: sort)
       }
 
     self.selectedSort = Signal.merge(
-      self.selectedSortIndexProperty.producer.takeWhen(initialSignal),
```
amy-at-kickstarter, line 64: Thanks @stevestreza-ksr, this is _way_ simpler this way. 

What I decided was that the use case and `SearchViewModel` can be as tightly coupled to API objects as they want - but I made the UI (`SortView` and `FilterCategoryView`) generic. Hopefully this makes things simpler to get away from API V1 later down the right.

scottkicks (approved):

amy-at-kickstarter:
Thanks @jovaniks! I'll address that cleanup in a separate PR, thanks for calling that out.

PR #2349, opened 2025-03-21
MBL-2212, MBL-2214: Add additional filter options and re-order to make Recommended first

amy-at-kickstarter:
# 📲 What

Add additional filtering options to `DiscoveryParams.Sort`.

# 🤔 Why

This adds a few extra sort options that are available in GraphQL. I also changed the default from Popularity to Recommended, to match Android.

Screenshot 2025-03-21 at 4 50 40 PM

# 🛠 How

Because the Search screen analytics are tightly coupled to `DiscoveryParams`, @Arkariang and I decided it makes the most sense to extend the existing `DiscoveryParams` model instead of creating a new one. However, it gets funny in this PR, because it means we're adding filter types that are really only in here for GraphQL - not for the actual V1 Discover endpoint.

I did my best to protect against those cases by adding `assert`s and sensible defaults everywhere I could.

### DIFF
--- a/Library/Styles/DiscoveryStyles.swift
+++ b/Library/Styles/DiscoveryStyles.swift
@@ -172,5 +172,17 @@ private func string(forSort sort: DiscoveryParams.Sort) -> String {
     return Strings.discovery_sort_types_newest()
   case .popular:
     return Strings.Popular()
+  case .most_backed:
+    assert(
+      false,
+      "most_backed was added for GraphQL compatibililty. It shouldn't actually be used in V1 Discover."
+    )
+    return Strings.discovery_sort_types_most_backed()
+  case .most_funded:
+    assert(
+      false,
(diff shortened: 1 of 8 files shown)


Library/Styles/DiscoveryStyles.swift
```diff
@@ -172,5 +172,17 @@ private func string(forSort sort: DiscoveryParams.Sort) -> String {
     return Strings.discovery_sort_types_newest()
   case .popular:
     return Strings.Popular()
+  case .most_backed:
+    assert(
```
amy-at-kickstarter, line 176: Note: our production builds strip assertions, this should never crash for an actual user. Just for us poor, beleaguered developers.

jovaniks (approved):
LGTM!

scottkicks (approved):
👍

PR #2350, opened 2025-03-24
Fix beta build

stevestreza-ksr:
# 📲 What

Changes a folder reference from "reference files in place" to an older-style manual Xcode group.

# 🤔 Why

Beta builds are broken.

# 🛠 How

Xcode recently added a new "reference files in place" option for folders which lets you automatically add files in that folder to a build without having to add them manually in Xcode. Some older version of Fastlane breaks with this, and we haven't updated it yet, so instead of fixing Fastlane, this removes that type of build option.


scottkicks (approved):
🧐

PR #2351, opened 2025-03-24
Turn guard statement into if statement

amy-at-kickstarter:
# 📲 What

Turn `guard` statement into `if` statement.

# 🤔 Why

It doesn't need to eject on failure, just throw an assert. Also it was breaking the build.

### DIFF
--- a/Library/UseCases/SearchFiltersUseCase.swift
+++ b/Library/UseCases/SearchFiltersUseCase.swift
@@ -119,7 +119,7 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
     }
 
     let index = self.categoriesProperty.value.firstIndex(of: category)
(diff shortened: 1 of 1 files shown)


scottkicks (approved):

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -119,7 +119,7 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
     }
 
     let index = self.categoriesProperty.value.firstIndex(of: category)
-    guard index != nil else {
+    if index == nil {
```
stevestreza-ksr, line 122: This assertion suggests that if index is nil, we should NOT set the selectedCategoryProperty to be that category. Are we sure it is right to let this fall through (and not return early) if this is nil?

PR #2357, opened 2025-03-26
Clear sort and filter options when search query is cleared

amy-at-kickstarter:
# 📲 What

Clear selected sort and filter options when the search query clears, is canceled, or is an empty string.

# 🤔 Why

This behavior feels a little more natural. Otherwise, you can get caught in a weird state where you have no search results, cancel the search, but then have 'retained' filters from your last failed query.

### DIFF
--- a/Library/ViewModels/SearchViewModel.swift
+++ b/Library/ViewModels/SearchViewModel.swift
@@ -371,11 +371,13 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs, 
   fileprivate let cancelButtonPressedProperty = MutableProperty(())
   public func cancelButtonPressed() {
     self.cancelButtonPressedProperty.value = ()
+    self.searchFiltersUseCase.inputs.clearOptions()
   }
 
   fileprivate let clearSearchTextProperty = MutableProperty(())
   public func clearSearchText() {
     self.clearSearchTextProperty.value = ()
+    self.searchFiltersUseCase.inputs.clearOptions()
   }
 
   fileprivate let searchFieldDidBeginEditingProperty = MutableProperty(())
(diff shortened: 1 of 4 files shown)


Library/ViewModels/SearchViewModel.swift
```diff
@@ -386,6 +388,10 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
   fileprivate let searchTextChangedProperty = MutableProperty("")
   public func searchTextChanged(_ searchText: String) {
     self.searchTextChangedProperty.value = searchText
+
+    if searchText.isEmpty {
```
amy-at-kickstarter, line 392: This is kind of imperative and I don't love it. The alternative I considered was something like
```
  queryText
  .filter {$0.isEmpty}
  .observeValues { [searchFiltersUseCase] _ in 
     searchFiltersUseCase.inputs.clearOptions()
   }
```

but that also feels pretty weird. I feel like this needs to be an `inputs` call because it modifies some properties, but maybe there's a more clever way to chain these signals?

stevestreza-ksr: These are effectively the same but the latter seems more in line with the existing VM pattern, so between those I'd pick that.

Either way could definitely ticket the improvements to the use case for future refactoring.

stevestreza-ksr (approved):

Library/ViewModels/SearchViewModel.swift
amy-at-kickstarter: The good thing is this will become moot soon, since we'll be copying the behavior that Android does! (In that case, the pills are always visible, and there's no need for clearing behavior at all.)

PR #2358, opened 2025-03-26
Remove PPOv1 feature flag

stevestreza-ksr:
# 📲 What

Removes the PPO feature flag.

# 🤔 Why

The feature is fully rolled out and we don't need it any more.

# 🛠 How

Deleted the flag and the helper function. Then went through every build failure and fixed it based on whether it was expecting a true or false value. Then went through and updated a ton of tests.


amy-at-kickstarter (approved):
👍 NICE

jovaniks (approved):
:rocket:

PR #2360, opened 2025-03-27
[MBL-2178] Upgrade minimum OS version to 16.0

stevestreza-ksr:
Upgrades the minimum version from 15.0 to 16.0.


jovaniks (approved):
Great! This allows us to remove the `BottomSheetPresenter`, previously used as a fallback for `UISheetPresentationController` on iOS 15 and earlier

amy-at-kickstarter (approved):
Wow that was easy 👍

PR #2365, opened 2025-04-01
[MBL-2202] PART II - Use "image(named:)" to load images

jovaniks:
# 📲 What

Update image loading to use the shared `image(named:)` helper instead of `UIImage(named:)`.

This ensures images are correctly resolved from the appropriate bundle in Unit Tests.

# 🛠 How

- Replace direct usage of `UIImage(named:)` with the `image(named:)` utility function.
- This function respects `AppEnvironment.current.mainBundle`, ensuring the correct bundle is used during testing.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2202)
- Check snapshots updated on this PR.

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |

# ✅ Acceptance criteria

- [x] Images are loaded correctly in Unit Tests.
- [x] Snapshot tests involving image assets pass as expected.

# ⏰ TODO

- [ ] Fonts are being handled in a separate PR.

# 📌 Notes

- This is part of the broader effort to improve asset handling in Unit Tests.


jovaniks (commented):
Heads up! This PR includes many snapshot updates. To make the review easier, feel free to filter by `.swift` files to focus on the actual code changes.

amy-at-kickstarter (changes requested):
Can you call these with the a explicit `Library.image(named:)`, instead of the short-hand `image(named:)`? Since it's such a generically named function, the code is much clearer to me when I can see where that function is coming from at a glance.

scottkicks (approved):
LGTM and would like to see `Library.` per Amy's request also.

jovaniks:
> Can you call these with the a explicit `Library.image(named:)`, instead of the short-hand `image(named:)`? Since it's such a generically named function, the code is much clearer to me when I can see where that function is coming from at a glance.

@amy-at-kickstarter Yep! The compiler is actually requiring us to use the module prefix here — otherwise, it interprets `image(named:)` as an instance method. So we’re explicitly calling `Library.image(named:)` to avoid ambiguity and ensure the right method is invoked.

image

PR #2371, opened 2025-04-07
[NO TICKET] Expose The Backing.Order Object

scottkicks:
# 📲 What

Updates our GraphAPI to expose the `Order` object that's on Backing.

# 🤔 Why

When someone backs via Pledge Management, an Order object is created. 
The backend is intentionally using these Order objects to keep these user separate from crowdfund/late pledge backers.

We'll use this object to help determine when to show a web view in place of the View Your Pledge screen.

More details can be found in the [SPIKE](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/3618504789/PM+Crunch+-+Net+new+backers+spike)

# 🛠 How

Updates our GraphAPI and decodes it in the usual way.

# 👀 See

Screenshot 2025-04-07 at 2 38 08 PM

# ✅ Acceptance criteria

- [x] All tests pass
- [x] Order object is now accessible via Backing


KsApi/models/templates/BackingTemplates.swift
```diff
@@ -14,6 +14,7 @@ extension Backing {
     isLatePledge: false,
     locationId: 1,
     locationName: "United States",
+    order: .init(checkoutState: .complete, currency: "USD", total: 12),
```
amy-at-kickstarter, line 17: I think this should be blank on the default template, since it's a new case (and possibly an edge case?)

amy-at-kickstarter (approved):
LGTM. My hope would be that we're able to use GraphQL directly wherever we're implementing this - instead of needing to use the API V1 `Backing` object. Is the `Order` going somewhere we're stuck using V1?

jovaniks (approved):

PR #2372, opened 2025-04-08
[MBL-2261] Display in USD for US-dollar-based Canadian projects

jovaniks:
# 📲 What

Fixes a bug where the currency symbol **CA$** was shown incorrectly on the pledge page for Canadian projects that use USD.

# 🛠 How

- Updated `PledgeViewModel` (standard pledges) to determine the country based on `data.project.stats.currency` instead of `data.project.country`.
- Updated `PostCampaignCheckoutViewModel` (late pledges) to use the same logic via `projectCountry`.

This ensures the correct formatting is applied for USD-based projects regardless of their origin country.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2261)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 13 05 55 | Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 13 31 16 |
| Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 13 05 09 | Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 13 01 23 |

# ✅ Acceptance criteria

- [x] Reward and shipping amounts display in USD for US-dollar-based Canadian projects.
- [x] Both standard and late pledge flows reflect this correction.

# ⏰ TODO

- [ ] Coordinate QA testing on staging.


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -323,10 +323,13 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs,
           data.context == .fixPaymentMethod &&
             data.project.personalization.backing?.isLatePledge == true
         )
+
+      let projectCountry = projectCountry(forCurrency: data.project.stats.currency) ?? data.project.country
```
amy-at-kickstarter, line 327: While this fixes the problem, there's a bigger underlying issue. The code now converts the project currency into a country, and then (if I'm correct), converts the country _back_ into a currency code when it gets to the formatter: https://github.com/kickstarter/ios-oss/blob/main/Library/Format.swift#L123-L127. I believe this back-and-forth between project country and currency is probably what led to this bug in the first place.

Can you take a look a little deeper and see if we can pass the _currency_ into the pledge summary view and the formatter, instead of passing the country around?

amy-at-kickstarter (changes requested):
Can you investigate a little deeper down the chain? Maybe we can fix a whole category of currency bugs along with this bug.

Library/ViewModels/PledgeViewModel.swift
jovaniks: Thanks, Amy! That totally makes sense — I agree there's some unnecessary back-and-forth going on.

That said, after reviewing the formatter logic (`currencySymbol`), it does rely on `Project.Country` for specific rules (like `.us`, `.sg`, `"kr"`, `"Fr"`, etc.), which are not always directly inferable from the `currency`. So replacing all usages of `country` with `currency` would mean we'd need to rethink how we encode those rules.

That said, I’ll take a deeper look to see:
- If we can cleanly extract these edge cases into something based on `currency` instead.
- Or at least minimize the need to pass both `currency` and `country`.

Appreciate the nudge to look further down the chain — definitely seems worth cleaning this up more broadly.

amy-at-kickstarter: Passing both would make sense to me! This also reminds me of the issue we had with language + currency recently.

jovaniks: As we discussed on Slack, we’ve decided to simplify the logic and move forward by only passing the `currencyCode` into the `currencySymbol` function.

We’ll handle a broader formatter cleanup in a [follow-up ticket](https://kickstarter.atlassian.net/browse/MBL-2354), as that change touches many files and would be better addressed separately.

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -323,10 +323,13 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs,
...
+
       let rewardsData = PostCampaignRewardsSummaryViewData(
         rewards: data.rewards,
         selectedQuantities: data.selectedQuantities,
-        projectCountry: data.project.country,
+        projectCountry: projectCountry,
```
amy-at-kickstarter, line 332: Before you merge this - can you verify nothing _else_ is using the `projectCountry` you're passing down, other than currency formatting code? For example, `PledgePaymentMethodCellViewModel` references `projectCountry` for an error message: https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/PledgePaymentMethodCellViewModel.swift#L113

jovaniks: I just reviewed it, the `projectCountry` in `PostCampaignRewardsSummaryViewData` is only used for currency formatting purposes.
Regarding the reference in `PledgePaymentMethodCellViewModel`, that value actually comes from a different data flow. You can see where it’s set in this file:
https://github.com/kickstarter/ios-oss/blob/main/Library/ViewModels/PledgePaymentMethodsViewModel.swift#L574

Let me know if you spot anything else I should double-check!

stevestreza-ksr (approved):

PR #2383, opened 2025-04-14
MBL-2222: Add project state filters to Search

amy-at-kickstarter:
# 📲 What

Implement project state filters for Phase 2 of the sort & filters project.

project-state-filters

* Added project state filters to `SearchFiltersUseCase` and `SearchViewModel`
* Show placeholder pills for 'All' filters and project state filters
* Show placeholder UI for choosing project filters
* Refactor `SearchFiltersUseCase` to make it easier to add more filter types in the future
* Refactor the data types used to show the filter modals

# 🤔 Why

I like to implement scaffolding like this before I go in and write the UI - it makes it easier for me to hook everything up later. Might as well sort out the hard problems first before making it pretty.

### DIFF
--- a/Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
+++ b/Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
@@ -39,13 +39,8 @@ internal final class SearchViewController: UITableViewController {
 
     self.viewModel.inputs.viewDidLoad()
 
-    let pillView = SearchFiltersHeaderView(didTapPill: { pill in
-      switch pill.filterType {
-      case .sort:
-        self.sortButtonTapped()
-      case .category:
-        self.categoryButtonTapped()
-      }
+    let pillView = SearchFiltersHeaderView(didTapPill: { [weak self] pill in
+      self?.viewModel.inputs.tappedButton(forFilterType: pill.filterType)
     }).environmentObject(self.sortAndFilterViewModel)
 
     let sortAndFilterHeader = UIHostingController(rootView: pillView)
@@ -179,16 +174,17 @@ internal final class SearchViewController: UITableViewController {
         self?.changeSearchFieldFocus(focus: $0, animated: $1)
       }
 
-    self.viewModel.outputs.showSort
+    self.viewModel.outputs.showFilters
       .observeForControllerAction()
-      .observeValues { [weak self] sheet in
-        self?.showSort(sheet)
-      }
-
-    self.viewModel.outputs.showCategoryFilters
-      .observeForControllerAction()
-      .observeValues { [weak self] sheet in
-        self?.showCategories(sheet)
+      .observeValues { [weak self] options, type in
+        switch type {
+        case .all:
+          self?.showAllFilters(options)
+        case .category:
+          self?.showCategories(options.category)
+        case .sort:
+          self?.showSort(options.sort)
+        }
       }
 
     self.showSortAndFilterHeader <~ self.viewModel.outputs.showSortAndFilterHeader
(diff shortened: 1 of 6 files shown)


Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -179,16 +174,17 @@ internal final class SearchViewController: UITableViewController {
...
       }
 
-    self.viewModel.outputs.showSort
-      .observeForControllerAction()
-      .observeValues { [weak self] sheet in
-        self?.showSort(sheet)
-      }
-
-    self.viewModel.outputs.showCategoryFilters
+    self.viewModel.outputs.showFilters
```
amy-at-kickstarter, line 177: Since we'll now be showing many filter types, I decided it was going to be very verbose for the view model to have individual methods for each. Instead, the view model outputs the filter options data (a `SearchFilterOptions` struct) and a `SearchFilterModalType` which tells you which one to show.

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -146,9 +138,7 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
     DiscoveryParams.State.successful
   ]
 
-  public let showCategoryFilters: Signal<SearchFilterCategoriesSheet, Never>
-  public let showSort: Signal<SearchSortSheet, Never>
-
+  public var showFilters: Signal<(SearchFilterOptions, SearchFilterModalType), Never>
```
amy-at-kickstarter, line 141: Normally I don't like emitting tuples, but it actually seems like the clearest solution here!

Library/UseCases/SearchFiltersUseCase.swift
```diff
@@ -198,12 +188,89 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
+  var pills: [SearchFilterPill] = []
+
+  pills.append(SearchFilterPill(
+    isHighlighted: sort != SearchFiltersUseCase.defaultSortOption,
+    filterType: .sort,
+    buttonType: .image("icon-sort")
+  ))
+
+  if featureSearchFilterByProjectStatusEnabled() {
+    pills.append(SearchFilterPill(
```
amy-at-kickstarter, line 208: Project state pills and the 'All' filters pills are only shown if this flag is on. When the flag is off, the search is always filtering to all project states. This way, I can ship my 'ugly' version and go back and make it pretty later.

scottkicks (approved):
Nice! Love how readable this is

jovaniks (approved):
LGTM

Library/ViewModels/SearchViewModel.swift
```diff
@@ -427,12 +410,20 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
   public let showSortAndFilterHeader: Signal<Bool, Never>
   public let projectsAndTitle: Signal<(Bool, [SearchResultCard]), Never>
 
-  public var showSort: Signal<SearchSortSheet, Never> {
-    return self.searchFiltersUseCase.showSort
+  public var showFilters: Signal<(SearchFilterOptions, SearchFilterModalType), Never> {
+    return self.searchFiltersUseCase.showFilters
```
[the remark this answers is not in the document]
amy-at-kickstarter: Ah, yeah, this is the convention I've been using - adding a computed property to the view model that passes along the signal from the use case. I'm open to other patterns, though! This way is a bit wordy and annoying when you make changes like this.

PR #2387, opened 2025-04-15
[MBL-2328] Fix image/audio element loading in campaign from search

stevestreza-ksr:
# 📲 What

Fixes image and audio loading when opening a project from search.

# 🤔 Why

So people can see images in search results.

# 🛠 How

Search loads without a Project instance, and instead passes a param (like an id or slug) of a project. The project page then loads that project one time.

The existing signal chain was assuming that the project page would be opened via a Project instance, so the signals would receive two projects. The first one being partial, the second one being complete.

Since it was making this assumption, it was skipping the first project returned. When opening via a param, this is the ONLY project being returned. So it was not doing the work to load images (and, it turns out, audio elements and I think page view events). So images and audio elements were not being shown.

This fixes by removing the `skip` call and replacing it with a `.compactMap({ project.extendedProjectProperties })`. This accomplishes the same thing when loading via a Project, but also makes it work when loading via a Param.

### DIFF
--- a/Library/ViewModels/ProjectPageViewModel.swift
+++ b/Library/ViewModels/ProjectPageViewModel.swift
@@ -389,9 +389,7 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
       }
 
     let trackFreshProjectAndRefTagViewed: Signal<(Project, RefTag?), Never> = Signal.zip(
-      freshProjectAndRefTag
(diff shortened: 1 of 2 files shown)


amy-at-kickstarter (approved):
LGTM, can we add a unit test to exercise this case?

jovaniks (approved):

stevestreza-ksr:
Unit test added. Fails without the fix, passes with it.

PR #2392, opened 2025-04-17
MBL-2353: Fix off-by-one error in search results

amy-at-kickstarter:
# 📲 What

Fix an off-by-one error when tapping on results in search.

# 🤔 Why

This happened because I added the search results title as a cell, instead of as a section header - so `indexPath.row` became off-by-one.

I double-checked and this bug doesn't exist in the release app, only the beta version.


jovaniks (approved):

Kickstarter-iOS/Features/Search/Datasource/SearchDataSource.swift
```diff
@@ -3,6 +3,8 @@ import Library
 import Prelude
 import UIKit
 
+typealias TitleRow = Void
```
scottkicks, line 6: nice!

scottkicks (approved):

PR #2393, opened 2025-04-18
 MBL-2227: Refactor ownership of selected filter state 

amy-at-kickstarter:
# 📲 What

Refactor ownership of selected filter state.

Instead of outputting `ReactiveSwift` signals for selected filters and the header "pill" views, `SearchFiltersUseCase` owns and updates an `ObserveableObject` called `SelectedSearchFilters`.

This makes it easy to keep all of the SwiftUI views in sync: each can use an `@ObservedObject` to view the current filter state. Calls to update the state are still passed to `SearchFiltersUseCase`, using the inputs on the view model and closures passed into SwiftUI. 

Untitled_Artwork 10

# 🤔 Why

The state management between SwiftUI and UIKit was getting really messy - especially since multiple modals ultimately needed to read/update the same state.

My goals here were:
* Keep `SearchFiltersUseCase` as the overall source-of-truth for all search filter related state
* Make it easy to dynamically update SwiftUI views as search filter state changed

# 🛠 How

Two notes here:
1.  I didn't make broader refactors to `FilterCategoryView` or `SortView`, because these are working fine as-is. We can consider whether a more SwiftUI-focused refactor would help here; ultimately those views could just have something like a `@Binding var selectedCategory: KsApi.Category`.
2. I did break my lovingly-written generic abstractions to do this, and had to axe some preview code. I really _wanted_ all of this UI to be very model agnostic, but it was no longer a realistic goal. I think nailing the data flow here is more important than everything being perfectly decoupled.


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderView.swift
```diff
@@ -1,27 +1,28 @@
...
-    self.pills = pills
+  var body: some View {
+    SearchFiltersHeaderView(
+      didTapPill: self.didTapPill,
+      pills: self.selectedFilters.pills
+    )
   }
 }
 
 struct SearchFiltersHeaderView: View {
```
amy-at-kickstarter, line 16: `SearchFiltersHeaderView` is now dumb/non-reactive, and has a reactive wrapper around it `SelectedSearchFiltersHeaderView` that the UI can use.

Library/UseCases/SearchFilters/SearchFilterOptions.swift
amy-at-kickstarter, line 1: Moved this into its own file.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -185,72 +195,7 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
 
   public var inputs: SearchFiltersUseCaseInputs { return self }
   public var uiOutputs: SearchFiltersUseCaseUIOutputs { return self }
-  public var dataOuputs: SearchFiltersUseCaseDataOutputs { return self }
-}
-
-private func filterPills(
```
amy-at-kickstarter, line 191: Moved into `SelectedSearchFilters`

jovaniks (approved):
Everything looks solid! Great architectural cleanup keeping `SearchFiltersUseCase` as the source of truth. The shift to `ObservableObject` seems cleanly executed and makes the SwiftUI integration feel much smoother.

One minor note (or question): I noticed the explicit use of `internal`, which is Swift’s default access level — is that a style convention across the codebase? Just curious since I’ve seen it pop up in a few places.

scottkicks (approved):
Sorry this took me a bit to wrap my head around the intention and how the old approach is being changed. Thanks for the handy diagram.

This makes sense to me. I like that the header view isn't reactive now..

Sorry about your model agnostic ambitions lol

amy-at-kickstarter:
> One minor note (or question): I noticed the explicit use of `internal`, which is Swift’s default access level — is that a style convention across the codebase? Just curious since I’ve seen it pop up in a few places.

Great question! There's no specific standard. I'll bring it up in our next iOS sync.

amy-at-kickstarter:
> Sorry this took me a bit to wrap my head around the intention and how the old approach is being changed. Thanks for the handy diagram.
> 
> This makes sense to me. I like that the header view isn't reactive now..
> 
> Sorry about your model agnostic ambitions lol

Ha, yeah, it took me a while to wrap my head around what I was even _doing_ here in the first place! Glad my doodles helped. Eventually I hope that we have some solid and well-understood patterns for SwiftUI <-> UIKit conversations.

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderView.swift
stevestreza-ksr: "Dumb SwiftUI view wrapped by a view model style SwiftUI view that wraps it" is a pattern I'm thinking about instead of "SwiftUI view that has a separate view model", fwiw

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -53,19 +54,16 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
     )
     .takePairWhen(self.tappedFilterTypeSignal)
     .map { a, b in (a.0, a.1, a.2, a.3, b) }
-    .map { [sortOptions, stateOptions] categories, category, sort, state, pill in
+    .map { [sortOptions, stateOptions] categories, category, sort, _, pill in
```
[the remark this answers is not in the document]
amy-at-kickstarter: Whoop, good catch. Will clean this up.

PR #2397, opened 2025-04-23
[MBL-2360] Show Tracking Number Cards In Activity

scottkicks:
# 📲 What

Exposes a new Activity.Category type as well as trackingNumber and and trackingUrl properties retrieved from our `/v1/activities` endpoint and uses them to populate our new [PM Tracking Number Cards](https://www.figma.com/design/yYATJ3ZjcwSPDEdM6OnOSX/Pledge-Management-Updates?node-id=3-505&t=9V1g3ZT1KeLrha0q-4).

# 🤔 Why

We want to show users who have backed a project via Pledge Management who chose a shippable reward that their reward has shipped and that they can easily open up their shipment tracking info/status in a browser.

# 🛠 How

The plumbing for showing these new cards has already been done!
So, all we need to do is:
- Expose the new category and properties on Activity
- Use that data to populate the cards
- Open a browser using the `trackingUrl` when the "Track shipment" button is tapped. 

# 👀 See

_FYI: The number and URL for this test project is "testing!" and a testing url._ 

|  |  |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-04-23 at 09 31 40 |  |

# ✅ Acceptance criteria

- [x] Tracking cards are shown for all shipped rewards for a project that was backed via Pledge Management.
- [x] Tapping "Track shipment" button opens the tracking URL in a browser
- [x] Card views match the design and display the project name/image, tracking number, track shipment button, and title label.


Kickstarter-iOS/Features/Activities/Datasource/ActivitiesDataSource.swift
```diff
@@ -50,6 +50,9 @@ internal final class ActivitiesDataSource: ValueCellDataSource {
         self.appendRow(value: activity, cellClass: ActivityFriendFollowCell.self, toSection: section)
       case .cancellation, .failure, .launch, .success, .suspension:
         self.appendRow(value: activity, cellClass: ActivityProjectStatusCell.self, toSection: section)
+      case .shipped:
+        // shipped data is loaded via `load(rewardTrackingData:...)` below on line: 62
```
amy-at-kickstarter, line 54: Can you add an `assert(false)` here ? That way we'll know if this is called by accident.

scottkicks: I don't think this can be called by accident. This will technically still be called when there are .shipped activities, but we're just handling that case in a separate load method.

Library/ViewModels/ActivitiesViewModel.swift
```diff
@@ -288,21 +288,26 @@ public final class ActivitiesViewModel: ActivitiesViewModelType, ActitiviesViewM
...
           )
-          return [RewardTrackingActivitiesCellData(trackingData: trackingData, project: $0)]
-        #else
-          return []
-        #endif
+
+          data.append(RewardTrackingActivitiesCellData(trackingData: trackingData, project: project))
+        }
+
+        return data
```
jovaniks, line 310: We could consider refactoring this block to use `compactMap` instead of manually building the array with `forEach` and `append`. It would reduce the nesting and make the intent a bit clearer. Here’s a quick idea:

```swift
self.rewardTrackingData = self.activities.signal
  .filter { _ in featureRewardShipmentTrackingEnabled() }
  .map { activities in
    activities
      .filter { $0.category == .shipped }
      .compactMap { activity -> RewardTrackingActivitiesCellData? in
        guard
          let project = activity.project,
          let trackingUrl = activity.trackingUrl,
          let trackingURL = URL(string: trackingUrl)
        else {
          return nil
        }

        let trackingData = RewardTrackingDetailsViewData(
          trackingNumber: activity.trackingNumber,
          trackingURL: trackingURL
        )

        return RewardTrackingActivitiesCellData(
          trackingData: trackingData,
          project: project
        )
      }
  }
```

I also like the idea of unwrapping `trackingURL` directly within the `guard` statement. Let me know what you think, happy to tweak as needed!

jovaniks (changes requested):
This looks great overall! 🙌
Just one small concern: I’d suggest avoiding the force unwrap on `trackingURL`. Maybe we could unwrap it safely using `guard let` to prevent any potential crashes if the URL is invalid.

Library/ViewModels/ActivitiesViewModel.swift
```diff
@@ -288,21 +288,26 @@ public final class ActivitiesViewModel: ActivitiesViewModelType, ActitiviesViewM
     // This is temporary until the backend implementation is completed.
```
jovaniks, line 288: Can these comment lines be removed now?

Library/ViewModels/ActivitiesViewModel.swift
scottkicks: great suggestion! will do!

jovaniks (approved):
Nice, it looks amazing

PR #2402, opened 2025-04-24
MBL-2220: (Part one)  Update search filters and search view models to support phase 2 search features

amy-at-kickstarter:
# 📲 What

Make some updates to `SearchViewModel` and `SearchFiltersUseCase` for the new parent filter page:

* Support resetting individual pages of the filter modal
* Rename `SearchFilterModalType` case `.all` to `.allFilters` for clarity
* Rename `clearOptions` to `clearedQueryText` to differentiate clearing the search query from clearing filters
* Update `SearchFiltersUseCaseTests` to test all of this (and make some updates for clarity, too)

# 🤔 Why

The parent filter page has complex behavior for the "Reset" button. The "Reset" button changes title and behavior based on which page its on top of. All of these changes support that.


jovaniks (approved):
Nice!!!

Library/UseCases/SearchFilters/SelectedSearchFilters.swift
```diff
@@ -29,62 +48,64 @@ public class SelectedSearchFilters: ObservableObject {
...
 
-private func filterPills(
-  fromSelectedSort sort: DiscoveryParams.Sort,
-  category: KsApi.Category?,
-  state: DiscoveryParams.State
-) -> [SearchFilterPill] {
-  let hasCategory = category != nil
-  let hasState = state != SearchFiltersUseCase.defaultStateOption
+  private func updatePills() {
+    let filterCount = [self.hasCategory, self.hasProjectState].reduce(0) { filterCount, hasFilter in
```
stevestreza-ksr, line 72: A little hard to scan this quickly to understand what it's doing. I think this is just counting how many conditions are active, so maybe replace this with `[self.hasCategory, self.hasProjectState].count(where: { $0 == true })`

stevestreza-ksr (approved):
Lgtm aside from one note

Library/UseCases/SearchFilters/SelectedSearchFilters.swift
amy-at-kickstarter: Fixed! Good suggestion.

PR #2410, opened 2025-04-30
Add ObservedSignal dynamic property wrapper

amy-at-kickstarter:
# 📲 What

This property wrapper makes it easy to dynamically update a SwiftUI view when a ReactiveSwift signal emits a value. This is useful if you want to mix together SwiftUI and ReactiveSwift.

# 🛠 How

Here's the simplest example: a view that takes a `Signal` on `init`, and updates itself when the signal emits a value:
```swift
public struct SimpleExampleView: View {
  @ObservedSignal var count: Int

  init(withCountSignal signal: Signal<Int, Never>) {
    _count = ObservedSignal<Int>(
      initialValue: 0,
      withSignal: signal
    )
  }

  public var body: some View {
    VStack {
      Text("The count is \(self.count)")
    }
  }
}
```

Here's another example. If you have a ReactiveSwift view model, you can use it in SwiftUI by wrapping its output signals in `@ObservedSignal`:

```swift
public protocol SomeReactiveSwiftViewModelType {
  var countSignal: Signal<Int, Never> { get }
  func increaseCount()
}

public struct ExampleView : View {
  let viewModel: SomeReactiveSwiftViewModelType
  @ObservedSignal var count: Int

  init(withViewModel viewModel: SomeReactiveSwiftViewModelType) {
    self.viewModel = viewModel
    _count = ObservedSignal<Int>(initialValue: 0, withSignal: viewModel.countSignal)
  }

  public var body: some View {
    VStack {
      Text("The count is \(self.count)")
      Button("Increase!") {
        self.viewModel.increaseCount()
      }
    }
  }
}
```

# 🤔  Why?

Android has an equivalent to this for Compose, called `subscribeAsState` (see their [Compose best practices](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2452357206/Compose+KS+Best+practices) here). Apparently it's been very handy for their migration to Compose! By creating an equivalent wrapper, my hope is that we can make some bigger strides towards moving code over to SwiftUI, without having to worry about porting everything to Combine or `ObservableObject`s.


Library/SwiftUI+Extensions/ObservedSignal.swift
```diff
@@ -0,0 +1,31 @@
...
+private class ObservableSignalWrapper<T>: ObservableObject {
+  private let (lifetime, token) = Lifetime.make()
+
+  @Published var value: T
+
+  init(initialValue value: T, withSignal signal: Signal<T, Never>) {
+    self.value = value
+
+    signal
+      .take(during: self.lifetime)
```
amy-at-kickstarter, line 13: @stevestreza-ksr I did some digging and apparently this is the way to make sure that there's not a retain issue with the observer. I tested this empirically by throwing some breakpoints in `deinit`, both in `ObservableSignalWrapper` and the various test view models I wrote.

scottkicks: So is the idea here that when the owner of the `@ObservedSignal` gets deallocated(`SimpleExampleView` from your example), then the Lifetime token expires cancelling the signal?

amy-at-kickstarter: Correct!

scottkicks (approved):
Thought I had approved this already 😅 . LGTM

Curious if you've found a good candidate for trying this out on?
It'd be nice to see it in action!

Library/SwiftUI+Extensions/ObservedSignal.swift
stevestreza-ksr: Scanning through the code in ReactiveSwift I think I get what's going on, so that seems to make sense.

Library/SwiftUI+Extensions/ObservedSignal.swift
```diff
@@ -0,0 +1,33 @@
...
+        self?.value = newValue
+      }
+  }
+}
+
+/// A property wrapper that is useful for bridging ReactiveSwift signals and SwiftUI views.
+/// A SwiftUI view that uses an `ObservedSignal` will update when its `ObservedSignal` emits any value.
+@propertyWrapper
+public struct ObservedSignal<T>: DynamicProperty {
+  @ObservedObject private var wrapper: ObservableSignalWrapper<T>
```
stevestreza-ksr, line 24: Not 100% sure on @ObservedObject here, @StateObject may be better, but that's in the weeds on property wrappers that I'm not totally sure on. If this doesn't cause the ObservableSignalWrapper to re-init every time it updates, I'm fine with that.

stevestreza-ksr (approved):
Looks good! It may also make sense to include a variant that takes a `Property` rather than a `Signal`. Would also be fun to see a compatible implementation of `onReceive(_:perform:)` vs the Combine method, but that's more an intellectual itch to scratch than anything that needs to happen here. 

I would just say to make sure:
- Any time the `Signal` emits, the view re-renders properly
- The `@ObservedObject` behavior I listed in the comments works as expected

Like this demo code should render every second. (Also you can see how passing a `Property` might make sense instead of a signal + a value)

```swift
let timer = SignalProducer.timer(
    interval: .seconds(1), 
    on: AppEnvironment.current.scheduler
)
  .map { date in DateFormatter().string(from: date) }
let signal = Property(initial: "", then: timer).signal

struct TestSignalView: View {
    @ObservedSignalWrapper(initialValue: "", withSignal: signal) var text

    var body: some View {
        Text(self.text)
    }
}
```

ifosli (approved):
Nice! I didn't think very deeply about the swiftUI/combine specifics, so full disclosure, if anything is wrong there I probably didn't catch it (I saw there were already 2 approvals). If you did want another opinion on any of those specifics, let me know, and I'd be happy to dig into it.

Library/SwiftUI+Extensions/ObservedSignal.swift
amy-at-kickstarter: My understanding is that `@ObservedObject` implies no ownership, where `@StateObject` makes SwiftUI maintain the lifecycle of the object. The part I'm not entirely clear on is how _ReactiveSwift_ manages the lifecycle of signals and observers. My understanding/empirical testing was that the `observe` call causes the observer (i.e. the `ObserveableSignalWrapper`)  to be retained as long as `lifetime` is active. So in this case, ReactiveSwift implicitly owns the wrapper, making `@StateObject` redundant.

PR #2411, opened 2025-04-30
Fix crash caused by tapping on title row in search results.

amy-at-kickstarter:
# 📲 What

Fix crash caused by tapping on title row in search results.

# 🤔 Why

This crash occurred because I didn't checking for the title row in the method `indexOfProject(forCellAtIndexPath)`. If you tapped on the title row, this method returns `-1`, causing a crash when the view model tries to fetch `projects[-1]`.

### DIFF
--- a/Kickstarter-iOS/Features/Search/Datasource/SearchDataSource.swift
+++ b/Kickstarter-iOS/Features/Search/Datasource/SearchDataSource.swift
@@ -1,6 +1,5 @@
 import KsApi
 import Library
-import Prelude
 import UIKit
 
 typealias TitleRow = Void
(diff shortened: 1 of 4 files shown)


Kickstarter-iOS/Features/Search/Datasource/SearchDataSource.swift
```diff
@@ -1,6 +1,5 @@
 import KsApi
 import Library
-import Prelude
```
amy-at-kickstarter, line 3: Wasn't needed.

Kickstarter-iOS/Features/Search/Datasource/SearchDataSourceTests.swift
amy-at-kickstarter, line 1: Added some unit tests to exercise this more thoroughly!

scottkicks (approved):
Nice!

jovaniks (approved):

PR #2413, opened 2025-05-01
MBL-2395: Refactor FilterCategoryView to be driven by state from FilterRootView

amy-at-kickstarter:
# 📲 What

Refactor `FilterCategoryView` to be driven by external state from `FilterRootView`, using a `@Binding` and a passed-in set of `Category` objects.

This includes:
* Change data sent to `FilterCategoryView` on its `init` method
* Delete `FilterCategoryViewModel`; move any code it needed into `FilterCategoryView` and `SearchFilterCategory`
* Give up on making `FilterCategoryView` a beautiful generic container; directly link it to the `KsApi.Category` class.

This PR depends on #2415 .

# 🤔 Why

There was a conflict in state ownership, which went something like this:
* `FilterCategoryView` created a `FilterCategoryViewModel`(stored as a `@StateObject`) on `init`
* However, once `FilterCategoryView` was placed inside of a `FilterRootView`, it could be re-created and re-rendered many times. `init` suddenly could be called more than once.
* There was code in `FilterCategoryView.init` which tried to update the `FilterCategoryViewModel`, but `@StateObject` tells SwiftUI to own a _single_ view model for the life time of the category view.
* This led to unexpected update behavior. Multiple copies of `FilterCategoryViewModel` were created, and changes sent to some of those copies were lost.
* This led to unexpected/unexplained behavior when trying to reset this view.

The correct solution here is to respect that the "selected filters" state is owned _outside_ of `FilterCategoryView`. This PR makes `FilterCategoryView` into a much "dumber" component. It now uses a binding to update state that is ultimately held by `SearchFiltersUseCase`.


Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -272,17 +272,8 @@ internal final class SearchViewController: UITableViewController {
...
-      if let category = maybeCategory {
-        if let parent = category.parent {
-          self?.viewModel.inputs.selectedCategory(.subcategory(parent, category))
-        } else {
-          self?.viewModel.inputs.selectedCategory(.rootCategory(category))
-        }
-      } else {
-        self?.viewModel.inputs.selectedCategory(.none)
-      }
+    filterView.onSelectedCategory = { [weak self] category in
```
amy-at-kickstarter, line 275: `FilterCategoryView` now returns a `SearchFiltersCategory` and not a `Category`

Kickstarter-iOS/Features/SortFilter/FilterCategory/__Snapshots__/FilterCategoryViewTest/testFilterCategoryView_CategoriesList.1.png
amy-at-kickstarter, line 1: Because I gave up on making it generic, the screenshot test now shows template categories instead of generic options 😢

Kickstarter-iOS/Features/SortFilter/FilterCategory/FilterCategoryViewModel_PhaseOne.swift
ifosli, line 1: Should this file be deleted? If not, would you mind explaining why it exists? And maybe we should consider renaming it?

ifosli (approved):
Overall, I think this makes a lot of sense! The context in the pr description was really helpful, so thanks for explaining! I just had one comment about a file that I think should just be deleted; if I'm wrong about that, let me know and I can take another pass!

Kickstarter-iOS/Features/SortFilter/FilterCategory/FilterCategoryViewModel_PhaseOne.swift
```diff
@@ -90,3 +95,23 @@ class FilterCategoryViewModel_PhaseOne<T: FilterCategory>: FilterCategoryViewMod
     return self.currentCategory?.id == category.id
   }
 }
+
+internal enum ConcreteFilterCategory: String, FilterCategory, CaseIterable {
```
jovaniks, line 99: Not seeing any current usage of this, is it safe to delete?

jovaniks (approved):
LGTM!!! just a small nitpick noted

Kickstarter-iOS/Features/SortFilter/FilterCategory/FilterCategoryViewModel_PhaseOne.swift
amy-at-kickstarter: Ah, so I branched `FilterCategoryView` when I shipped Search Phase 2. If you look in `SearchViewController`, this class is still used here: https://github.com/kickstarter/ios-oss/blob/main/Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift#L220-L229

Once we clean up `featureSearchByProjectStatusEnabled()` we can remove this file.

Kickstarter-iOS/Features/SortFilter/FilterCategory/FilterCategoryViewModel_PhaseOne.swift
amy-at-kickstarter: This is still used in the tests for the `PhaseOne` view model and View

PR #2414, opened 2025-05-05
Refactor UpdateBackingInput construction to reduce conditionals and i…

jovaniks:
# 📲 What

Refactors the construction of `UpdateBackingInput` to reduce inline conditional complexity and improve readability within `PledgeViewModel`.

# 🤔 Why

The previous implementation had several inline ternary operators and conditionals that made the constructor logic hard to follow and error-prone. In particular, `latePledge` handling and Apple Pay cases introduced subtle branching behaviors.

Additionally, a bug was found where omitting the amount on late pledges caused validation errors from the backend:

> “Validation failed: Amount Sorry, please enter a pledge amount greater than or equal to €150 to select that reward”

This refactor aligns the logic with web behavior and makes late pledges include the amount, resulting in a clearer rejection when appropriate:

> “Sorry, the project has ended.”

# 🛠 How

- Introduced `baseInput`, a dedicated internal constructor that handles common input field mapping.
- Introduced `buildFullBackingInput`, which prepares values like `amount`, `rewardIds`, and `locationId` for regular pledge updates.
- Handled `.changePaymentMethod` and `.fixPaymentMethod` paths separately, skipping unnecessary values.
- Updated late pledge logic to always include the amount (per updated requirement).

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screenshot - iPhone 16 Pro - 2025-05-05 at 10 15 29 | Simulator Screenshot - iPhone 16 Pro - 2025-05-05 at 10 17 09 |

# ✅ Acceptance criteria

- [x] Pledge context logic is now separated and easier to follow.
- [x] Late pledges include amount and behave consistently with web.
- [x] Test PASSED


Library/UpdateBackingInput+ConstructorTests.swift
```diff
@@ -57,12 +57,12 @@ final class UpdateBackingInput_ConstructorTests: TestCase {
 
     let input = UpdateBackingInput.input(from: data, isApplePay: false)
```
amy-at-kickstarter, line 58: Can you clarify why this test had to change? Shouldn't this behavior be unaltered?

amy-at-kickstarter (commented):
I've got a question before approving this!

scottkicks (approved):
LGMT 
➕ 1 to Amy's comments

Library/UpdateBackingInput+Constructor.swift
```diff
@@ -6,33 +6,58 @@ extension UpdateBackingInput {
...
-    // Check if this is a fix errored pledge context. If it is, do not include fields that cannot
-    // be changed; amount, locationId, and rewardIds.
-    let isFixPledge = updateBackingData.pledgeContext == .fixPaymentMethod
-
-    // Check if this is a change payment method and PLOT pledge; if so, only include paymentSourceId or applePay.
-    let isChangePaymentMethodAndPlot = updateBackingData
-      .pledgeContext == .changePaymentMethod && updateBackingData.backing.paymentIncrements.count > 0
-
-    let shouldOmitAmount = updateBackingData.backing
-      .isLatePledge || isFixPledge || isChangePaymentMethodAndPlot
```
jovaniks, line 26: Now late pledges include the amount to avoid the error _Validation failed: Amount Sorry, please..._

Library/UpdateBackingInput+ConstructorTests.swift
jovaniks: I updated it so that all `.changePaymentMethod` operations ( even if they’re not PLOT) only send what’s needed (e.g. `paymentSourceId` or `applePay`). I ran several tests with and without PLOT flows, and everything is working as expected.

amy-at-kickstarter (approved):

PR #2415, opened 2025-05-06
Refactor selected category in search filters to use new SearchFiltersCategory type

amy-at-kickstarter:
# 📲 What

Represent the selected search category as a `SearchFiltersCategory`, an `enum` which differentiates between root categories and subcategories.

# 🤔 Why

It makes the most sense for the "source of truth" for the selected category to live in `SearchFiltersUseCase`, instead of in `FilterCategoryView`. That means that `SearchFiltersUseCase` needs to keep track of whether or not the selected category is a root category, a subcategory, or no category - so that `FilterCategoryView` can retrieve that information.

This enum makes that clearer than using a tuple `(Category, Category)` type.

Note that I'm not updating `FilterCategoryView` in this PR - that is in #2413 .


Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -225,13 +225,19 @@ internal final class SearchViewController: UITableViewController {
 
     let viewModel = FilterCategoryViewModel_PhaseOne<KsApi.Category>(
       with: sheet.category.categories,
-      selectedCategory: self.viewModel.outputs.selectedFilters.category
+      selectedCategory: self.viewModel.outputs.selectedFilters.category.category
```
amy-at-kickstarter, line 228: The name of this property isn't super descriptive - it's "get the actual value out of the enum that we need to use to tell the API what category to filter to."

Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -225,13 +225,19 @@ internal final class SearchViewController: UITableViewController {
...
     )
 
     let filterView = FilterCategoryView_PhaseOne(
       viewModel: viewModel,
-      onSelectedCategory: { [weak self] category in
-        self?.viewModel.inputs.selectedCategory(category)
+      onSelectedCategory: { [weak self] maybeCategory in
```
ifosli, line 233: nit (not blocking): `maybeCategory` doesn't feel like an intuitive name to me; I was wondering what the other option(s) were until I read the code. Thoughts on `nullableCategory` or `optionalCategory` or even just `category` instead? I personally like to just do ex `category` followed by `if let category {`, but I'm not sure off the top of my head if that's a common pattern in our codebase or not.

Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -266,8 +272,17 @@ internal final class SearchViewController: UITableViewController {
     filterView.onSelectedProjectState = { [weak self] state in
       self?.viewModel.inputs.selectedProjectState(state)
     }
-    filterView.onSelectedCategory = { [weak self] category in
-      self?.viewModel.inputs.selectedCategory(category)
+    filterView.onSelectedCategory = { [weak self] maybeCategory in
```
ifosli, line 275: same here

Library/UseCases/SearchFilters/SearchFiltersCategory.swift
```diff
@@ -0,0 +1,41 @@
+import KsApi
+
+public enum SearchFiltersCategory: Equatable {
+  case none
+  case rootCategory(KsApi.Category)
+  case subcategory(KsApi.Category, KsApi.Category)
```
ifosli, line 6: nit: It would be nice to label these, or at least label the parent, so that we don't forget accidentally swap which is which.

ifosli (approved):
I like this! Just left some nits around naming

Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
jovaniks: I like the idea of naming it just `category`

Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -266,8 +272,17 @@ internal final class SearchViewController: UITableViewController {
...
+
+      if let category = maybeCategory {
+        if let parent = category.parent {
+          self?.viewModel.inputs.selectedCategory(.subcategory(parent, category))
+        } else {
+          self?.viewModel.inputs.selectedCategory(.rootCategory(category))
+        }
+      } else {
+        self?.viewModel.inputs.selectedCategory(.none)
+      }
```
jovaniks, line 285: Suggestion: consider refactoring this block using early returns to simplify the control flow. For example:

```swift
guard let category = maybeCategory else {
  self?.viewModel.inputs.selectedCategory(.none)
  return
}

if let parent = category.parent {
  self?.viewModel.inputs.selectedCategory(.subcategory(parent, category))
  return
}

self?.viewModel.inputs.selectedCategory(.rootCategory(category))
```

This makes the logic easier to follow and avoids nested `if/else` blocks.

Library/UseCases/SearchFilters/SearchFiltersCategory.swift
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  if case let .subcategory(leftRootCategory, leftSubcategory) = lhs,
+     case let .subcategory(rightRootCategory, rightSubcategory) = rhs {
+    return leftRootCategory.id == rightRootCategory.id &&
+      leftSubcategory.id == rightSubcategory.id
+  }
+
+  return false
+}
```
jovaniks, line 41: Minor refactor suggestion: consider simplifying the equality logic using a switch over the tuple of both enums:

```swift
public static func == (lhs: SearchFiltersCategory, rhs: SearchFiltersCategory) -> Bool {
  switch (lhs, rhs) {
  case (.none, .none):
    return true

  case let (.rootCategory(leftCategory), .rootCategory(rightCategory)):
    return leftCategory.id == rightCategory.id

  case let (.subcategory(leftRootCategory, leftSubcategory), 
       .subcategory(rightRootCategory, rightSubcategory)):
    return leftRootCategory.id == rightRootCategory.id && 
       leftSubcategory.id == rightSubcategory.id

  default:
    return false
  }
}
```

jovaniks (approved):
This refactor looks amazing 👌 I just dropped a couple of small suggestions, nothing blocking!

Library/UseCases/SearchFilters/SearchFiltersCategory.swift
amy-at-kickstarter: Nice, I like this. Will do!

PR #2418, opened 2025-05-07
Fix search scroll so that the loader is in the footer, not the header

amy-at-kickstarter:
# 📲 What

Fix search scroll so that the loader is in the footer, not the header

# 🤔 Why

This way the screen loads much more smoothly. This page doesn't have a pull-to-refresh, and results are cleared when you enter a new query, so there's no reason I can figure to put the spinner on top instead of the bottom.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  ugly-search-scroll | pretty-search-scroll |


Kickstarter-iOS/Features/Search/Controller/__Snapshots__/SearchViewControllerTests/testView_PrelaunchProject_InSearch_Success.lang_en_device_pad.png
ifosli, line 1: Not blocking, but would it be reasonable to add a snapshot test that shows the animation icon in the footer? Maybe by reducing the number of projects displayed?

ifosli (approved):
This UI is so much better! (Also, I would love for us to do the same thing to the profile page when backed/saved projects load in)

jovaniks (approved):
Looks great! Just a nit... it would be nice to include the loading indicator in the test snapshot for completeness.

Kickstarter-iOS/Features/Search/Controller/__Snapshots__/SearchViewControllerTests/testView_PrelaunchProject_InSearch_Success.lang_en_device_pad.png
amy-at-kickstarter: Good idea, I'll take a stab at that!

scottkicks (approved):

Kickstarter-iOS/Features/Search/Controller/__Snapshots__/SearchViewControllerTests/testView_PrelaunchProject_InSearch_Success.lang_en_device_pad.png
amy-at-kickstarter: That was a giant pain in the butt but I got it working 😂

PR #2421, opened 2025-05-12
RFC: Refactor design system colors [will not merge, please leave comments + approval]

amy-at-kickstarter:
# 👀 

This PR is put up as a demonstration of this idea. If this RFC is approved, I'll create a new PR, which adds a complete `CoreColor` definition, a complete `SemanticColor` definition, and replaces the existing `Colors` struct with these new classes.

# 📲 What

Refactor the new design system colors:
* Split color definitions into "core" colors and "semantic" colors
* Define colors in code, by hex value, instead of in the Assets catalog
* Add a preview view to the Beta tools, along with a screenshot test

# 🤔 Why

Our design system is split into two parts: there's the core color palette, like `gray/100`, which defines a specific hex value like `#FAFAFA`. Then there are the semantic colors, like `color/background/action`, which reference core colors.

Screenshot 2025-05-12 at 3 06 09 PM

By updating our design system to match that mapping, it's much easier to add or edit design system colors. 

This proposed refactor also makes a few other changes:
1. By defining colors in code instead of in asset catalogs, it's easier to search (and edit) the codebase.
2. By using a `struct` instead of `enum` to define semantic colors, we can create arbitrarily nested color definitions, like `Colors.background.action` instead of `Colors.backgroundAction`.


Library/Styles/Colors/Colors2.swift
```diff
@@ -0,0 +1,35 @@
+import UIKit
+
+public struct Colors2 {
```
amy-at-kickstarter, line 3: This will replace `Colors`, named `Colors2` for clarity in this PR.

Library/Styles/Colors/Colors2.swift
```diff
@@ -0,0 +1,35 @@
...
+  public struct text {
+    public static let primary = SemanticColor(
+      "text/primary",
+      lightMode: .grey_1000,
+      darkMode: .grey_100
+    )
+
+    public static let secondary = SemanticColor(
```
amy-at-kickstarter, line 11: I wanted it to be easy and type-safe to access semantic colors. You can access them like `Colors2.text.secondary` or `Colors2.text.accent.red.disabled`. These should map 1:1 to the design system semantic colors in Figma. 

I considered doing something clever and dynamic here, like implementing `@dynamicMember`, but I feel like this approach is clear enough to implement and very simple.

Library/Styles/Colors/KSRColorPalette.swift
```diff
@@ -0,0 +1,24 @@
...
+
+public enum KSRPaletteColor: Int {
+  case white = 0xFFFFFF
+  case grey_100 = 0xFAFAFA
+  case grey_400 = 0xC9C9C9
+  case grey_700 = 0x4D4D4D
+  case grey_1000 = 0x171717
+  case red_400 = 0xF7BBB7
+  case red_700 = 0x931910
+  // etc
```
amy-at-kickstarter, line 11: This enum is intended to define every possible color in the design system palette. `SemanticColor` uses values from `KSRPaletteColor`.

Kickstarter-iOS/Features/DesignSystem/Controller/ColorsView.swift
```diff
@@ -0,0 +1,84 @@
+import Library
+import SwiftUI
+
+struct ColorsView: View {
+  let semanticColors = [
```
amy-at-kickstarter, line 5: A beta tools view for showing all the light/dark mode semantic colors. Also conveniently demonstrates the usage of the new colors.

Library/Styles/Colors/KSRColorPalette.swift
amy-at-kickstarter: I specifically like that these are defined in code, because it makes it easy if you ever need to search for a hex value you found in Figma.

Library/Styles/Colors/SemanticColor.swift
```diff
@@ -0,0 +1,30 @@
...
+  private let darkModeColor: CoreColor
+  public let name: String
+
+  init(_ name: String, lightMode: CoreColor, darkMode: CoreColor) {
+    self.name = name
+    self.lightModeColor = lightMode
+    self.darkModeColor = darkMode
+  }
+
+  public func uiColor() -> UIColor {
```
amy-at-kickstarter, line 17: @amy-at-kickstarter check if these reload automatically

PR #2425, opened 2025-05-13
[MBL-2427] Support login in survey webview

ifosli:
# 📲 What

If the survey webview is opened without a current logged in user, trigger login.

Note: If they dismiss login without actually logging in, they'll still get into a weird/bad state in the pledge redemption webview. However, if they tap the link again, they get another chance to log in.

# 🤔 Why

If we log in the user before trying to show a survey or pledge redemption flow, we're able to actually display the survey in the webview. This is particularly important for net new backers, since they may open pledge redemption in a webview before they've backed a single project (and thus possibly before they've even created an account).

I'm not 100% sure if old surveys need a user to log in. This is hard to test because if they still exist in prod, they're definitely rare. However, I did confirm that they redirect to login on web.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2427)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Plus - 2025-05-13 at 11 55 33 | Simulator Screen Recording - iPhone 16 Plus - 2025-05-13 at 11 46 49 |

# ✅ Acceptance criteria

- [x] Trying to open a survey deep link now triggers login. If a user successfully logs in, they are then able to see their survey/pledge management.

# ⏰ TODO

- [ ] I'm going to follow up with Tan/UX on if we want to create a new login intent for this case instead of just using `generic`. In the interest of time, I'm using `generic` for now. (I definitely don't want to wait for new strings to be translated before pushing this change.)

### DIFF
--- a/Kickstarter-iOS/Features/SurveyResponse/Controller/SurveyResponseViewController.swift
+++ b/Kickstarter-iOS/Features/SurveyResponse/Controller/SurveyResponseViewController.swift
@@ -11,6 +11,7 @@ internal protocol SurveyResponseViewControllerDelegate: AnyObject {
 
 internal final class SurveyResponseViewController: WebViewController {
   internal weak var delegate: SurveyResponseViewControllerDelegate?
+  private var sessionStartedObserver: Any?
   fileprivate let viewModel: SurveyResponseViewModelType = SurveyResponseViewModel()
 
   internal static func configuredWith(surveyUrl: String)
@@ -34,6 +35,10 @@ internal final class SurveyResponseViewController: WebViewController {
     self.viewModel.inputs.viewDidLoad()
   }
 
+  deinit {
+    self.sessionStartedObserver.doIfSome(NotificationCenter.default.removeObserver)
+  }
+
   internal override func bindViewModel() {
     super.bindViewModel()
 
@@ -42,6 +47,17 @@ internal final class SurveyResponseViewController: WebViewController {
       .observeValues { [weak self] in
         self?.navigationController?.dismiss(animated: true, completion: nil)
         self?.delegate?.surveyResponseViewControllerDismissed()
+      }
+
+    self.viewModel.outputs.goToLoginSignup
+      .observeForControllerAction()
+      .observeValues { [weak self] intent in
+        self?.goToLoginSignup(with: intent)
+      }
+
+    self.sessionStartedObserver = NotificationCenter.default
+      .addObserver(forName: .ksr_sessionStarted, object: nil, queue: nil) { [weak self] _ in
+        self?.viewModel.inputs.userSessionStarted()
       }
 
     self.viewModel.outputs.goToProject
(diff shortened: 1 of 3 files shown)


scottkicks (approved):
> they'll still get into a weird/bad state in the pledge redemption webview. However, if they tap the link again, they get another chance to log in.

This was the edge case I was confused about in slack, but your description of it here cleared things up for me.

Thanks for the speedy fix.

Library/ViewModels/SurveyResponseViewModelTests.swift
```diff
@@ -206,6 +208,36 @@ final class SurveyResponseViewModelTests: TestCase {
     }
   }
 
+  // MARK: - Test login
+
+  func testGoToLoginSignup() {
```
amy-at-kickstarter, line 213: Nit: Test names should have a full description of the test like `test_loggedOutUser_triggersGoToLoginSignup()`

Library/ViewModels/SurveyResponseViewModelTests.swift
```diff
@@ -206,6 +208,36 @@ final class SurveyResponseViewModelTests: TestCase {
...
+      self.webViewLoadRequest.assertValueCount(0)
+
+      self.vm.inputs.userSessionStarted()
+
+      // Request should be sent as soon as user has logged in.
+      self.webViewLoadRequest.assertValueCount(1)
+    }
+  }
+
   // MARK: - Decision policy tests
```
amy-at-kickstarter, line 241: Can you add an additional test, which checks that if the user is already logged in, that the web request loads immediately? (Basically every time I add something to a `combineLatest` I try to test both if that signal does/doesn't fire, because that's been a big source of bugs for me.)

Kickstarter-iOS/Features/SurveyResponse/Controller/SurveyResponseViewController.swift
```diff
@@ -44,6 +49,17 @@ internal final class SurveyResponseViewController: WebViewController {
         self?.delegate?.surveyResponseViewControllerDismissed()
       }
 
+    self.viewModel.outputs.goToLoginSignup
```
amy-at-kickstarter, line 52: For the edge case where the user _doesn't_ log in, or their login is invalid - can we add some default/placeholder UI to this page? Even if it's just a native spinner?

amy-at-kickstarter (approved):
Fix LGTM; two non-blocking suggestions for improvement. (If you can get them in for the hotfix, awesome, otherwise I'm happy with those as fast-follows to `main`.)

Library/ViewModels/SurveyResponseViewModelTests.swift
ifosli: I disagree with this, actually. I think a lot of our tests unnecessarily long names and it makes it tricky to find their corresponding snapshots sometimes. (If there were multiple things that could trigger login, I'd agree, but since there's just one thing that triggers it, I don't think the test name should specify it.) Plus, this naming is consistent with the rest of the tests in this file.

Kickstarter-iOS/Features/SurveyResponse/Controller/SurveyResponseViewController.swift
ifosli: My guess is that that's not possible to do with a webview. But I'll look into it and file a ticket to add a "bad" or loading state if it looks maybe doable.

PR #2427, opened 2025-05-14
[Part 1] MBL-2357: Add CoreColor and SemanticColor

amy-at-kickstarter:
# 📲 What

Add `CoreColor` and `SemanticColor`, two structs which represent colors in our design system.

# 🤔 Why

This does several things:
* Creates a logical separation between our color "palette" and semantic colors
* Forces semantic colors to only use valid core color options
* Makes it fast and easy to add new semantic colors

# 👀 See

See the follow up PRs to this PR, #2428 and #2429.

# 📠 CC
@stevestreza-ksr, @jovaniks


Library/Styles/Colors/SemanticColor.swift
```diff
@@ -0,0 +1,30 @@
...
+  private let darkModeColor: CoreColor
+  public let name: String
+
+  init(_ name: String, lightMode: CoreColor, darkMode: CoreColor) {
+    self.name = name
+    self.lightModeColor = lightMode
+    self.darkModeColor = darkMode
+  }
+
+  public func uiColor() -> UIColor {
```
amy-at-kickstarter, line 17: @ifosli I did some experimentation, and yes, it looks like both UIKit and SwiftUI will automatically reload a color created with this initializer! No subscription to trait changes required.

If you throw a breakpoint in here, it's called when the appearance changes.

## SwiftUI Example

Presenting this view:
```swift
struct DarkModeView: View {
  @SwiftUICore.Environment(\.colorScheme) private var colorScheme
  
  var isDark: Bool {
    self.colorScheme == .dark
  }
  
  var body: some View {
    VStack {
      Text("Dark mode is \(self.isDark ? "on" : "off")")
        .font(InterFont.heading2XL.swiftUIFont())
        .foregroundStyle(Colors2.text.primary.swiftUIColor())
    }
    .background(Colors2.surface.primary.swiftUIColor())
  }
}
```
Looks like this when you toggle dark mode on and off in the simulator:
toggle-darkmode-swiftui


## UIKit Example

Presenting this view controller:
```swift
class DarkViewController: UIViewController {

  var label: UILabel? = nil
  var isDark: Bool {
    self.traitCollection.userInterfaceStyle == .dark
  }

  override func viewDidLoad() {
    self.view.backgroundColor = Colors2.surface.primary.uiColor()
    
    self.label = UILabel()
    self.label?.font = InterFont.heading2XL.font()
    self.label?.textColor = Colors2.text.primary.uiColor()
    self.updateText()
    
    self.view.addSubview(self.label!)
    self.label?.constrainViewToMargins(in: self.view)
  }
  
  //Note this is just here to update the text, not the colors
  override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
    super.traitCollectionDidChange(previousTraitCollection)
    self.updateText()
  }
  
  func updateText() {
    self.label?.text = "Dark mode is \(self.isDark ? "on" : "off")"
  }
 
}
```
Looks like this when you toggle the simulator appearance:
toggle-darkmode-uikit

Library/Styles/Colors/SemanticColor.swift
```diff
@@ -0,0 +1,30 @@
...
+    return UIColor { traits in
+      if traits.userInterfaceStyle == .dark && featureDarkModeEnabled() {
```
amy-at-kickstarter, line 19: Note that this flag check means we can actually feature flag dark mode now 🎉 . This wasn't possible with asset catalogs.

Library/Styles/Colors/SemanticColor.swift
jovaniks, line 17: What do you think about adding an optional `opacity` parameter (defaulting to 1.0) to support transparency?

Currently, there are about 29 places where `withAlphaComponent` is used manually, this could help centralize and simplify those cases.

```swift
  public func uiColor(opacity: CGFloat = 1.0) -> UIColor {
    return UIColor { traits in
      if traits.userInterfaceStyle == .dark && featureDarkModeEnabled() {
        return UIColor(coreColor: self.darkModeColor).withAlphaComponent(opacity)
      } else {
        return UIColor(coreColor: self.lightModeColor).withAlphaComponent(opacity)
      }
    }
  }
  ```

ifosli (approved):
Nice! I also like how this approach makes it possible to implement an override toggle in the app if we ever want to do that

Library/Styles/Colors/SemanticColor.swift
amy-at-kickstarter: Good idea! I'll add this.

PR #2430, opened 2025-05-14
[MBL-2428] Replace hardcoded removal message with localized string

jovaniks:
# 📲 What

Replace the hardcoded string "This comment has been removed for violating..." with the localized version
`Strings.This_comment_has_been_removed_for_violating_kickstarters_community_guidelines()`.

# 🛠 How

Update the call to use:

```swift
Strings.This_comment_has_been_removed_for_violating_kickstarters_community_guidelines(community_guidelines:)
```

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2428)

# ✅ Acceptance criteria

- [ ] The localized string appears correctly in the UI
- [ ] Tests PASS


ifosli (approved):
You and Amy are going to run into a merge conflict here... I left this comment on Amy's pr too: Not blocking, but I personally prefer when we wait to pull in localized strings until they're actually translated; it's easier to check if they're translated or not when you're looking at a diff than if we're trying to confirm that every localized string that's visible to users during release time has been translated. So I'm okay with merging this as it is, but then we need to make sure we thoroughly check strings when we cut our next release build.

amy-at-kickstarter:
> You and Amy are going to run into a merge conflict here... I left this comment on Amy's pr too: Not blocking, but I personally prefer when we wait to pull in localized strings until they're actually translated; it's easier to check if they're translated or not when you're looking at a diff than if we're trying to confirm that every localized string that's visible to users during release time has been translated. So I'm okay with merging this as it is, but then we need to make sure we thoroughly check strings when we cut our next release build.

@ifosli How about we just add `make strings` to our deploy process? Should be an easy change in Circle CI.

jovaniks:
> > You and Amy are going to run into a merge conflict here... I left this comment on Amy's pr too: Not blocking, but I personally prefer when we wait to pull in localized strings until they're actually translated; it's easier to check if they're translated or not when you're looking at a diff than if we're trying to confirm that every localized string that's visible to users during release time has been translated. So I'm okay with merging this as it is, but then we need to make sure we thoroughly check strings when we cut our next release build.
> 
> @ifosli How about we just add `make strings` to our deploy process? Should be an easy change in Circle CI. So long as nobody is pushing breaking changes to strings, it should be safe.

@amy-at-kickstarter The challenge here is around test snapshots. They can fail when existing strings are updated. I see two possible approaches:

1. Integrate `make strings` into the CircleCI pipeline (as Amy suggested). If a build fails due to snapshot mismatches, a developer should jump in and update the affected snapshots.

2. Ensure that anyone (or the devs who pushed strings with incomplete translations) updates both the strings and the corresponding snapshots before the release cut (as Ingerid mentioned).

ankushkushwaha (approved):

PR #2431, opened 2025-05-15
Implement Dark Mode for iOS

amy-at-kickstarter:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -238,6 +238,12 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         enabled ? Analytics.shared().enable() : Analytics.shared().disable()
       }
 
+    self.viewModel.outputs.darkModeEnabled
```
amy-at-kickstarter, line 241: Added some code here to dynamically update the window when the feature flag changes (or on app foreground). This should give us some protection so we can use our feature flag to turn dark mode on and off.

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -88,17 +90,17 @@ extension UIButton {
     switch self.state {
     case .disabled:
       buttonConfiguration.background.backgroundColor = styleConfig.disabledBackgroundColor
-      self.tintColor = styleConfig.disabledTitleColor
+      buttonConfiguration.baseForegroundColor = styleConfig.disabledTitleColor
```
amy-at-kickstarter, line 93: Using `self.tintColor` triggered another configuration update, causing buttons to refresh in an infinite loop. Not sure why enabling dark mode caused this issue.

@jovaniks Any ideas? My change here was a best guess as to the expected behavior. You can duplicate the infinite loop by turning on Dark Mode, and then opening the Design System page in the Beta tools.

Library/UIColor+Mixing.swift
```diff
@@ -2,11 +2,11 @@ import UIKit
 
 extension UIColor {
   public func mixLighter(_ amount: CGFloat) -> UIColor {
-    return self.mix(with: .ksr_white, amount: amount)
+    return self.mix(with: UIColor.white, amount: amount)
```
amy-at-kickstarter, line 5: This was one exception to the find and replace - I thought it would be weird for these colors to flip.

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
jovaniks: I took a look and ran a few tests, your solution seems to work well! 👍

The use of `tintColor` makes sense here, especially when buttons include an icon (like in the attached screen recording).

 Simulator Screen Recording - iPhone 16 Pro - 2025-05-15 at 10 20 29  Simulator Screen Recording - iPhone 16 Pro - 2025-05-15 at 10 08 45

Kickstarter-iOS/Features/Discovery/Views/DiscoveryProjectCategoryView.swift
```diff
@@ -22,14 +22,11 @@ import UIKit
   override func bindStyles() {
     super.bindStyles()
 
-    _ = self.blurView
```
amy-at-kickstarter, line 25: Not sure what this used to be - it was a white, blurry image view behind these icons:
Screenshot 2025-05-19 at 10 15 28 AM

Removed it since it was, well, white-on-white and looked bad in dark mode.

PR #2433, opened 2025-05-19
[Part 1]: Add LegacyColor type, with values for all legacy design system colors

amy-at-kickstarter:
# 📲 What

* Add `LegacyColor` struct and `LegacyColors` struct, which adds values matching all old design system colors in `Colors.swift`. This is meant to live next to `SemanticColor` and the `Colors` struct.
* Add `ColorResolver` to `Environment` so colors can be resolved either dynamically/statically. This allows the test environment to use static UIColors, for equality comparisons.

# 🤔 Why

`LegacyColor` supports dark mode - which means we can now boldly go where the iOS app hasn't gone before.

# 🛠 How

I took a best-guess effort to turn legacy colors into dark mode colors. I found the nearest color palette and inverted its value. So, for example, `create_200` is green, so its dark mode equivalent became `green_800`.


Library/Styles/Colors/NewDesignSystemColors.swift
```diff
@@ -151,6 +151,12 @@ public struct Colors {
         darkMode: .red_1000
       )
 
+      public static let subtle = SemanticColor(
```
amy-at-kickstarter, line 154: `Colors.swift` had values for these, even though they're from the new design system.

Library/Styles/Colors/SemanticColor.swift
```diff
@@ -1,30 +1,58 @@
...
 import UIKit
 
+public protocol AdaptiveColor {
+  var lightModeColor: UIColor { get }
+  var darkModeColor: UIColor { get }
+}
+
+public extension AdaptiveColor {
+  func uiColor(opacity alpha: CGFloat = 1.0) -> UIColor {
+    return AppEnvironment.current.colorResolver.color(
```
amy-at-kickstarter, line 11: Note that the color is now created by this `ColorResolver` in the environment. This is because unit tests couldn't compare two dynamically provided colors. By adding a layer of indirection here, we can use a `MockColorResolver` in our unit tests, which always returns static (instead of dynamic) colors.

ifosli (approved):
We may want to change the dark mode version of some of these colors after product QA (or just change them to semantic versions if they're not working) but I think this approach seems reasonable and should generally work!

Library/Styles/Colors/LegacyColors.swift
```diff
@@ -0,0 +1,128 @@
+import UIKit
+
+@available(
+  *,
+  deprecated,
+  message: "If you are using a legacy color, you should replace it with an appropriate semantic color from the new design system."
+)
```
scottkicks, line 7: nice!

scottkicks (approved):

PR #2434, opened 2025-05-19
[Part 2] Replace all old design system colors from Colors.swift with LegacyColors or SemanticColors

amy-at-kickstarter:
# 📲 What

Replace all old design system colors from `Colors.swift` with either a `LegacyColor` or a `SemanticColor`, as appropriate.

# 🤔 Why

This giant find-and-replace gets us two things:
1. All legacy design system colors are now dynamically-provided, light/dark mode colors
2. It is now very clear that these color helpers are "old" and should no longer be used, in favor of `SemanticColor`

I could have made this much PR much shorter by replacing the code _inside_ `Colors.swift` - instead of porting everything - but I wanted to make sure our design system is extremely clear. This way, we're not tempted to copy-paste old design system colors just because they're "close enough" to what we want.

# 🛠 How

Find and replace.

# 👀 Please note

This fix was 99% automatic. Let your eyes glaze over as it scrolls before you ... take a closer look if anything looks weird. I know this PR looks huge but it should be pretty fast to review!

📠 cc @jovaniks @stevestreza-ksr  📠


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOStyles.swift
```diff
@@ -19,41 +19,41 @@ enum PPOStyles {
...
-    foreground: UIColor.ksr_support_400,
-    background: UIColor.ksr_celebrate_100
+    foreground: LegacyColors.ksr_support_400.uiColor(),
+    background: LegacyColors.ksr_celebrate_100.uiColor()
   )
 
   static let alertColor = (
-    foreground: UIColor.ksr_alert_foreground,
-    background: UIColor.ksr_alert_background
+    foreground: Colors.Text.Accent.Red.bolder.uiColor(),
```
amy-at-kickstarter, line 32: These were in the old design system, but there was a note that they belonged in the _new_ design system.

scottkicks (approved):

Kickstarter-iOS/SharedViews/AlertBanner.swift
```diff
@@ -36,7 +36,7 @@ final class AlertBanner: UIView {
 
     let icon = UIImageView()
     icon.image = Library.image(named: "icon--alert")
-    icon.tintColor = Colors.Icon.danger.uiColor()
+    icon.tintColor = Colors.Text.Accent.Red.bolder.uiColor()
```
amy-at-kickstarter, line 39: Slightly "wrong" from a semantic perspective, but it matches the old behavior.

Library/Styles/ButtonStyles.swift
```diff
@@ -71,13 +71,14 @@ public let blackButtonStyle: ButtonStyle = { button in
...
-  <> UIButton.lens.backgroundColor(for: .disabled) .~ UIColor.ksr_support_700.mixLighter(0.36)
-  <> UIButton.lens.backgroundColor(for: .selected) .~ UIColor.ksr_support_700.mixLighter(0.46)
+  <> UIButton.lens.titleColor(for: .normal) .~ LegacyColors.ksr_white.uiColor()
+  <> UIButton.lens.titleColor(for: .highlighted) .~ LegacyColors.ksr_white.uiColor()
+  <> UIButton.lens.titleColor(for: .disabled) .~ LegacyColors.ksr_support_100.uiColor()
+  <> UIButton.lens.backgroundColor(for: .normal) .~ LegacyColors.ksr_support_700.uiColor()
+  <> UIButton.lens.backgroundColor(for: .highlighted) .~ LegacyColors.ksr_support_700.uiColor()
+  .mixDarker(0.66)
+  <> UIButton.lens.backgroundColor(for: .disabled) .~ LegacyColors.ksr_support_700.uiColor().mixLighter(0.36)
+  <> UIButton.lens.backgroundColor(for: .selected) .~ LegacyColors.ksr_support_700.uiColor().mixLighter(0.46)
```
ifosli, line 81: We'll probably need to update the disabled/selected states for these buttons - I think `mixLighter` is the opposite of what we want in dark mode.

Library/ViewModels/ProjectPamphletSubpageCellViewModelTests.swift
```diff
@@ -21,6 +21,8 @@ internal final class ProjectPamphletSubpageCellViewModelTests: TestCase {
   override func setUp() {
     super.setUp()
 
+    self.vm = ProjectPamphletSubpageCellViewModel()
```
ifosli, line 24: Is this change intentionally in this pr? I'm fine with it being here; just checking!

ifosli (approved):
Looks good! I'm concerned with the number of `mixLighter`/`mixDarker`s that we use (I don't expect any of them to translate well to dark mode) but I think that's a problem for a later pr.

Library/Styles/ButtonStyles.swift
amy-at-kickstarter: Amusingly, I think none of the styles in here are even _used_. So we might be in the clear!

Library/ViewModels/ProjectPamphletSubpageCellViewModelTests.swift
amy-at-kickstarter: Yup - the tests were failing because the `vm` was being instantiated before `pushEnvironment` set up all the mocking code.

PR #2436, opened 2025-05-19
[Part 3] Hand-fix code and storyboards that don't support dark mode

amy-at-kickstarter:
# 📲 What

After find-and-replacing all our old design system colors, we were left with some awkward areas in the app where dark mode didn't apply correctly. 

This largely included storyboards with hard-coded hex color values, and a few places where we incorrectly referenced (say) `UIColor.white` instead of `UIColor.ksr_white`. 

I applied a few different fixes in here:
* Deleting a white background image that was shown on top of a white background
* Changing hardcoded color values in Storyboards to dynamic color values that remained in the Assets catalog (conveniently, these already had dark mode values defined)
* Turning hardcoded color values in Storyboard to `clear`
* Applying background values in colors in code

This probably doesn't cover _ever_ dark mode bug, but it got the most obvious ones that I found in a couple hours of testing. We'll find more, later!

# 🤸‍♀️ But wait, there's more!

I had to re-record some screenshot tests here. While I was recording them, I swapped them all to `orthogonalCombos`, hence the huge amount of deletes.

It looks like the Activity card has an off-by-half-pixel change. The change password view actually looks a little _better_ after my changes to the storyboards.

# 👀  How to test

Dark mode is still off in this PR. If you want to test dark mode, delete the `Appearance` option in our `Info.plist` and turn on `featureDarkModeEnabled()`.

📠 cc @jovaniks @stevestreza-ksr  📠


Kickstarter-iOS/Features/Discovery/Views/DiscoveryProjectCategoryView.swift
```diff
@@ -22,8 +22,11 @@ import UIKit
   override func bindStyles() {
     super.bindStyles()
 
-    _ = self.blurView
```
amy-at-kickstarter, line 25: I'm not sure what this view originally was. `white--gradient--layer` is a white image displayed on a white background, so it's no longer necessary. It looked pretty funny in dark mode, though!

My solution here is to set the view to the same height as the image to eliminate any change to the UI. You'll note in the screenshot tests that everything looks the same.

ifosli (approved):

scottkicks (approved):

PR #2441, opened 2025-05-21
MBL-2351: Add feature flag for filtering by % raised

amy-at-kickstarter:
# 📲 What

Add a feature flag for phase 4 search filters, filtering by % raised.


scottkicks (approved):
👍 
I wonder if there are feature flags we can clean up now?

amy-at-kickstarter:
> 👍 I wonder if there are feature flags we can clean up now?

I cleaned up the previous search flag, but it would be cool if we cleaned up the late pledges one...

jovaniks (approved):

PR #2445, opened 2025-05-22
Update Activity feed for dark mode

amy-at-kickstarter:
# 📲 What

- Replace custom white color value in Storyboards for Activity and Activity preview cells with semantic `white` color from asset catalog
- Add programmatic background colors to Activity cells, wherever they were missing. I specifically used `background/surface/primary` so that the dark mode version looks a little nicer, instead of using `legacy/white`. 

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  
|  |  |

### DIFF
--- a/Kickstarter-iOS/Features/Activities/Storyboard/Activity.storyboard
+++ b/Kickstarter-iOS/Features/Activities/Storyboard/Activity.storyboard
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="22505" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="23094" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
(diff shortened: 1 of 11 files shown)


Kickstarter-iOS/Features/Activities/Storyboard/Activity.storyboard
```diff
@@ -89,7 +90,7 @@
                                                     </subviews>
                                                 </stackView>
                                             </subviews>
-                                            <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                            <color key="backgroundColor" name="white"/>
```
amy-at-kickstarter, line 93: Most of these changes are automatic - these lines where I replace `<color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" />` with `<color key="backgroundColor" name="white"/>` are the actual bug fix.

Kickstarter-iOS/Features/Activities/Views/Cells/ActivityFriendBackingCell.swift
```diff
@@ -95,5 +95,7 @@ internal final class ActivityFriendBackingCell: UITableViewCell, ValueCell {
         topBottom: Styles.gridHalf(5),
         leftRight: Styles.grid(2)
       )
+
+    self.projectTextContainerView.backgroundColor = Colors.Background.Surface.primary.uiColor()
```
amy-at-kickstarter, line 99: I double-checked everywhere I had to set a `backgroundColor` in these storyboards, and also set a `backgroundColor` programmatically, if it was missing.

scottkicks (approved):
ooo nice. these cells look way better

PR #2453, opened 2025-05-30
[MBL-2452] Call to `addUserToSecretRewardGroup` mutation when secret reward token is presented

jovaniks:
# 📲 What

Implements support for unlocking secret rewards via deep link using the `addUserToSecretRewardGroup` mutation.

# 🛠 How

- Added `RewardsUseCase` to handle the logic for rewards/secret rewards.
- Extracted deep link parameters (`projectId` and `secretRewardToken`) and triggered the mutation `addUserToSecretRewardGroup`.
- Preserved the existing `configureWith(...)` function to avoid unnecessary updates in current tests.
- Added a new `configureWith(..., secretRewardToken:)` for cases where the `secretRewardToken` is required.
- Prompted the Login view when a logged-out user attempts to open a secret reward deep link.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2452)

| Secret reward - Logged-out user | Secret reward - Logged-in user |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-05-29 at 17 30 21  | Simulator Screen Recording - iPhone 16 Pro - 2025-05-29 at 17 38 15 |

| Non-secret reward - Logged-out user | Non-secret reward - Logged-in user |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-05-29 at 17 42 06 | Simulator Screen Recording - iPhone 16 Pro - 2025-05-29 at 17 40 01 |

# ✅ Acceptance criteria

- [x] Opening a deep link with `projectId` and `secretRewardToken` triggers the mutation
- [x] Secret reward is added to the list of available rewards
- [x] `addUserToSecretRewardGroup` only runs for logged-in users when `secretRewardToken` is presented.

# ⏰ TODO

- [ ] [MBL-2478](https://kickstarter.atlassian.net/browse/MBL-2478) Evaluate polishing or final tweaks

[MBL-2478]: https://kickstarter.atlassian.net/browse/MBL-2478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -216,32 +230,41 @@ public final class ProjectPageViewModel: ProjectPageViewModelType, ProjectPageVi
...
-        fetchProject(projectOrParam: projectOrParam, shouldPrefix: shouldPrefix)
-          .on(
-            starting: { isLoading.value = true },
-            terminated: { isLoading.value = false }
-          )
-          .map { project in
-            (project, refInfo?.refTag.map(cleanUp(refTag:)))
-          }
-          .materialize()
+      .map { data, shouldPrefix -> (Either<Project, any ProjectPageParam>, RefInfo?, String?, Bool) in
```
amy-at-kickstarter, line 242: Could this just be `map(unpack)` like it was before?

amy-at-kickstarter (approved):
LGTM, some non-blocking suggestions

Library/ViewModels/RewardsUseCase.swift
```diff
@@ -0,0 +1,77 @@
...
+}
+
+public protocol RewardsUseCaseType {
+  var inputs: RewardsUseCaseInputs { get }
+  var outputs: RewardsUseCaseOutputs { get }
+}
+
+public final class RewardsUseCase: RewardsUseCaseType, RewardsUseCaseInputs, RewardsUseCaseOutputs {
+  public init(secretRewardToken: Signal<String?, Never>, userSessionStarted: Signal<Void, Never>) {
+    let goToRewardsTappedSignal = self.goToRewardsTappedProperty.signal
```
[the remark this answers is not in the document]
jovaniks: Rather than being an initial signal, this is a triggered signal that emits when the “View rewards”, “View your rewards”, or “Back this project” button is tapped.

Library/ViewModels/RewardsUseCase.swift
```diff
@@ -0,0 +1,77 @@
...
+  var goToRewards: Signal<Void, Never> { get }
+}
+
+public protocol RewardsUseCaseType {
+  var inputs: RewardsUseCaseInputs { get }
+  var outputs: RewardsUseCaseOutputs { get }
+}
+
+public final class RewardsUseCase: RewardsUseCaseType, RewardsUseCaseInputs, RewardsUseCaseOutputs {
+  public init(secretRewardToken: Signal<String?, Never>, userSessionStarted: Signal<Void, Never>) {
```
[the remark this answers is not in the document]
jovaniks: Good point! Since `RewardsUseCase` is initialized in the ProjectPageViewModel’s constructor and the `secretRewardToken` is only set later during `configureWith`, I don’t think passing a plain `String?` would be appropriate here — unless we refactor the `RewardsUseCase` to allow setting the token via a setter function or a dedicated input method.

Open to suggestions if you think there’s a cleaner way to handle this!

Library/ViewModels/ProjectPageViewModel.swift
```diff
@@ -899,6 +976,30 @@ private func fetchProject(projectOrParam: Either<Project, any ProjectPageParam>,
...
+// TODO: Consider relocating this logic to `RewardsUseCase` to consolidate secret reward handling.
+// Ticket: [MBL-2478](https://kickstarter.atlassian.net/browse/MBL-2478)
+private func addUserToSecretRewardGroupIfNeeded(
+  project: Project,
+  secretRewardToken: String?
+) -> SignalProducer<Void, ErrorEnvelope> {
+  let isUserLoggedIn = AppEnvironment.current.currentUser != nil
+
+  guard isUserLoggedIn, let secretRewardToken = secretRewardToken, !secretRewardToken.isEmpty else {
+    return SignalProducer(value: ())
```
stevestreza-ksr: It looks like this returns a success when these cases are triggered. Just sanity checking this shouldn't return an error here, or a success bool or enum or something.

jovaniks: Sure!, I can return a boolean to indicate whether the `addUserToSecretRewardGroup` mutation was triggered or skipped. In this case, **true** would mean the call was made successfully, while **false** means the conditions weren’t met (e.g. the user is not logged in or the token is missing). If the mutation fails (e.g. invalid token), the error is still handled via the `ErrorEnvelope`, propagated through the [projectError](https://github.com/kickstarter/ios-oss/blob/75a4fadbc4df8f8aed0ac76582eed5f69343fa2b/Library/ViewModels/ProjectPageViewModel.swift#L388) signal.

PR #2454, opened 2025-05-30
[MBL-2452] Add AudienceData.secret to reward fragment

jovaniks:
# 📲 What

Add `AudienceData.secret` to the `RewardFragment` to determine whether a reward is marked as secret.

# 🛠 How

- Update RewardFragment to include the new field AudienceData.secret.
- Introduce isSecretReward property to the Reward model.
- Sort rewards based on the following priority:
 1. Rewards without items
 2. Secret rewards
 4. Other available rewards
 5. Unavailable rewards

# 👀 See

Simulator Screenshot - iPhone 16 Pro - 2025-05-30 at 07 37 18

# ✅ Acceptance criteria

- [ ] `isSecretReward` is correctly mapped from the GraphQL response
- [ ] Rewards are sorted in the correct order

# ⏰ TODO

- [ ] Work on UI to see the Secret reward badge correctly


Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -471,10 +471,20 @@ private func backingAndShippingTotal(for project: Project, and reward: Reward) -
 }
 
 private func allowableSortedProjectRewards(from project: Project) -> [Reward] {
-  let availableRewards = project.rewards.filter { rewardIsAvailable($0) }
```
amy-at-kickstarter, line 474: This is hard to follow as four different map methods. Can you combine them into one loop, like:

```swift
var notReward: [Reward] = []
var unavailableReward: [Reward] = []
var secretRewards: [Reward] = []
var availableRewards: [Reward] = []

for reward in rewards {
  if reward.isNoReward {
    notReward.append(reward)
  } else if { 
    //etc
```

amy-at-kickstarter (changes requested):
Mostly LGTM but a requested changed for clarity in `allowableSortedProjectRewards`. Also, is there a test exercising `allowableSortedProjectRewards`? Would be useful to get a secret reward tested in there.

Library/ViewModels/RewardsCollectionViewModel.swift
jovaniks: Good call!!! I refactored it into a single loop as suggested.

amy-at-kickstarter: Much better, thanks!

amy-at-kickstarter (approved):

PR #2459, opened 2025-06-02
Disable flaky test in ProjectPageViewControllerTests

amy-at-kickstarter:
# 📲 What

Disable a flaky snapshot test, `testLoggedOut_NonBacker_LiveProjectSwitchedToCampaignTab_Success`.

# 🤔 Why

This snapshot test regularly times out. There's nothing I can find that's obviously causing the timeout. By disabling this test, we can see if it's an issue with the overall suite timing out (in which case, _another_ test will become flaky), or if it's isolated to this one.

Either way, hopefully this moves us towards not being annoyed by this one any more.


amy-at-kickstarter:
> Good choice! Also, I did not know you could disable a test like this; I like it!

This suggestion came from @stevestreza-ksr! The test runner only runs functions beginning with `test`.

stevestreza-ksr (approved):

PR #2461, opened 2025-06-03
[MBL-2478] Fix badge overlap on Edit Reward screen when rotating device

jovaniks:
# 📲 What

Fixes a layout issue where the "Secret Reward" and "Your selection" badges overlap when rotating the device in the **Edit reward** screen.

# 🛠 How

- Introduced `currentRewardCardMaskViewTopMarginInset` property to store the current top margin value dynamically.
- Created `updateRewardCardMaskViewLayout()` to apply margin updates based on whether extra spacing is needed (e.g., when the reward is backed and is a secret reward).
- Called `updateRewardCardMaskViewLayout()` during `bindStyles()` to reflect layout changes.
- Removed hardcoded `.zero` margin.

# 👀 See

[Bug ticket](https://kickstarter.atlassian.net/browse/MBL-2479)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-06-03 at 05 55 05 | Simulator Screen Recording - iPhone 16 Pro - 2025-06-03 at 06 27 34 |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] On the Edit reward screen, "Secret Reward" and "Your selection" badges should not overlap.
- [ ] The spacing should adjust correctly when rotating the device.
- [ ] The fix should not affect rewards without both badges visible.

# ⏰ TODO

- [ ] Consider update the layout structure on the [cleanup ticket MBL-2478](https://kickstarter.atlassian.net/browse/MBL-2478)

### DIFF
--- a/Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
+++ b/Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
@@ -38,11 +38,6 @@ public final class RewardCardContainerView: UIView {
       |> \.translatesAutoresizingMaskIntoConstraints .~ false
   }()
 
-  /// Stores the top margin value to be applied to `rewardCardMaskView`.
-  /// This is dynamically adjusted based on whether the reward requires additional spacing
-  /// (e.g., to avoid overlapping badges like "Secret Reward" and "Your selection").
-  private var currentRewardCardMaskViewTopMarginInset = CGFloat.zero
-
   override init(frame: CGRect) {
     super.init(frame: frame)
 
@@ -66,8 +61,6 @@ public final class RewardCardContainerView: UIView {
     _ = self.rewardCardMaskView
       |> checkoutWhiteBackgroundStyle
       |> roundedStyle(cornerRadius: Styles.grid(3))
-
-    self.updateRewardCardMaskViewLayout()
   }
 
   public override func bindViewModel() {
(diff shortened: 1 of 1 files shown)


amy-at-kickstarter (approved):

Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
```diff
@@ -60,8 +65,9 @@ public final class RewardCardContainerView: UIView {
 
     _ = self.rewardCardMaskView
       |> checkoutWhiteBackgroundStyle
-      |> \.layoutMargins .~ .zero
       |> roundedStyle(cornerRadius: Styles.grid(3))
+
+    self.updateRewardCardMaskViewLayout()
```
ifosli: Nit: do we need to update the layout margins in `bindStyles` at all? It looks like this can only change in `configure(with data`. If that's true, and we switch to just updating the layout margin in configure with data, we'd no longer need to store it in a property. (I'm okay with you looking into this as a follow-up if you'd rather and merging this is urgent!)

ifosli (approved):
Nice find! I left a (non-blocking) suggestion that I think would simplify this code if it works, but I haven't tested if it works or not.

Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
jovaniks: Good call! I went ahead and tried this approach — it works as expected. Nice catch!

PR #2462, opened 2025-06-03
MBL-2480: Fix showstopper dark mode design bugs

amy-at-kickstarter:
# 📲 What

Fix 3 design bugs found in dark mode.

# 🤔 Why

This gets dark mode ready to ship!

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot 2025-06-03 at 1 36 46 PM | Screenshot 2025-06-03 at 1 34 47 PM |
| Screenshot 2025-06-03 at 1 36 23 PM | Screenshot 2025-06-03 at 1 35 00 PM |
| Screenshot 2025-06-03 at 1 36 34 PM | Screenshot 2025-06-03 at 1 34 55 PM |

### DIFF
--- a/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewControllerTests.swift
+++ b/Kickstarter-iOS/Features/ProjectPage/Controller/ProjectPageViewControllerTests.swift
@@ -960,7 +960,8 @@ internal final class ProjectPageViewControllerTests: TestCase {
 
   // MARK: - Tab Content Tests
 
-  func testLoggedOut_NonBacker_LiveProjectSwitchedToCampaignTab_Success() {
+  // FIXME: MBL-2463 This test is flaky and frequently times out. Disabling it for the moment.
+  func disabled_testLoggedOut_NonBacker_LiveProjectSwitchedToCampaignTab_Success() {
     let config = Config.template
     let project = Project.cosmicSurgery
       |> Project.lens.photo.full .~ ""
--- a/Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
+++ b/Kickstarter-iOS/Features/RewardsCollection/Views/RewardCardContainerView.swift
@@ -60,7 +60,6 @@ public final class RewardCardContainerView: UIView {
 
     _ = self.rewardCardMaskView
       |> checkoutWhiteBackgroundStyle
-      |> \.layoutMargins .~ .zero
       |> roundedStyle(cornerRadius: Styles.grid(3))
   }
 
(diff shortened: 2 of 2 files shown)


Kickstarter-iOS/Features/LoginTout/Controller/LoginToutViewController.swift
```diff
@@ -521,7 +521,7 @@ private let baseLabelStyle: LabelStyle = { label in
 private let logoImageViewStyle: ImageViewStyle = { imageView in
   imageView
     |> \.image .~ image(named: "kickstarter-logo")?.withRenderingMode(.alwaysTemplate)
-    |> \.tintColor .~ LegacyColors.ksr_create_500.uiColor()
+    |> \.tintColor .~ Colors.Brand.logo.uiColor()
     |> \.contentMode .~ .scaleAspectFit
```
amy-at-kickstarter, line 525: I found 3 logos in the app overall. I spot checked all three.

Kickstarter-iOS/SharedViews/SharedStoryboards/Main.storyboard
```diff
@@ -17,6 +16,8 @@
                         <rect key="frame" x="0.0" y="975" width="768" height="49"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                         <color key="backgroundColor" red="0.0" green="0.0" blue="0.0" alpha="0.0" colorSpace="custom" customColorSpace="sRGB"/>
+                        <tabBarAppearance key="standardAppearance"/>
```
amy-at-kickstarter, line 19: This was a strange one. It looks like our tab bar changes appearances on different tabs - for example, when we're showing the Profile page, it was semi-translucent; but for the Login page, it was completely transparent. I spent a while digging to see if we were manually setting different tab bar appearances for different tabs, but couldn't find anything to that effect.

Eventually, I poked around in the storyboard files and found these checkboxes:
Screenshot 2025-06-03 at 1 42 30 PM

By checking them, the tab bar now always has a default translucent appearance, and the thin 'shadow' line that separates it from app content. I think this is what we always _wanted_ and the fact that the appearance changed on different screens was actually a bug.

It now looks like this on the login page, for example:
fixed-tab-bar

This storyboard is very old - I'm not sure if this behavior was ever intentional, or if it's simply so many version untouched that it started doing something strange. 🤷‍♀️ But I feel good about this fix.

jovaniks (approved):
Nice!!!

Kickstarter-iOS/Features/ProjectPage/Views/ProjectNavigationSelectorView.swift
```diff
@@ -178,7 +178,8 @@ final class ProjectNavigationSelectorView: UIView {
         |> UIButton.lens.title(for: .normal) %~ { _ in section.displayString.uppercased() }
         |> UIButton.lens.accessibilityLabel %~ { _ in section.displayString }
         |> UIButton.lens.titleColor(for: .normal) %~ { _ in LegacyColors.ksr_support_400.uiColor() }
-        |> UIButton.lens.titleColor(for: .selected) %~ { _ in LegacyColors.ksr_trust_500.uiColor() }
+        |> UIButton.lens
+        .titleColor(for: .selected) %~ { _ in LegacyColors.Project.Navigation.selected.uiColor() }
```
ifosli, line 182: nit: visually, the indentation here looks weird to me. It would be nice to indent the second line a little. If our formatter insists on this as-is, though, feel free to leave it. (I don't think it's worth going back and fixing unless you're touching this file again)

PR #2464, opened 2025-06-04
Fix another round of dark mode showstoppers

amy-at-kickstarter:
# 📲 What

Fix another round of show-stopper bugs in Dark Mode.

* Fix contrast issues on Login page
* Fix contract issue with Search bar
* Fix hard-coded white in Newsletters page

Screenshot 2025-06-04 at 4 43 38 PM
Screenshot 2025-06-04 at 4 43 22 PM


jovaniks (approved):

ifosli (approved):
Looks good! Just one nit - it looks like the facebook logo doesn't change color/isn't visible in dark mode. If it's not too inconvenient, it would be nice to show the logo in the same color as the facebook text.

jovaniks:
> Looks good! Just one nit - it looks like the facebook logo doesn't change color/isn't visible in dark mode. If it's not too inconvenient, it would be nice to show the logo in the same color as the facebook text.

Good call, to fix it just set the "Render as" option in the asset catalog to "Template Image".

amy-at-kickstarter:
> Looks good! Just one nit - it looks like the facebook logo doesn't change color/isn't visible in dark mode. If it's not too inconvenient, it would be nice to show the logo in the same color as the facebook text.

Whoops, great catch! Will fix that before merging.

PR #2469, opened 2025-06-10
 Improve contrast in Dark Mode for "Successful" banner and "Your Selection" tag

jovaniks:
# 📲 What

Improves visual clarity in dark mode by updating the color used for "Successful" banner and "Your Selection" tag

# 🛠 How

Updated the `ksr_trust_500` darkMode value to use`#7271FF`.

# 👀 See

- Tickets:
  - [MBL-2486](https://kickstarter.atlassian.net/browse/MBL-2486)
  - [MBL-2487](https://kickstarter.atlassian.net/browse/MBL-2487)

| Before 🐛 | After 🦋 |
| --- | --- |
| image | Simulator Screenshot - iPhone 16 Pro - 2025-06-10 at 06 44 53 |
| image | Simulator Screenshot - iPhone 16 Pro - 2025-06-10 at 06 17 09 |

# ✅ Acceptance criteria

- [ ]   "Successful" banner and "Your Selection" tag are easier to read

[MBL-2486]: https://kickstarter.atlassian.net/browse/MBL-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MBL-2487]: https://kickstarter.atlassian.net/browse/MBL-2487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ifosli (approved):
Looks much better! Just checking: since you changed the dark mode value of this color instead of adding an extra legacy semantic color, did you check on the other places this color is used? My guess is this version will look better there, too, but only if we're using black text in all those places. Since this probably looks better everywhere, I'm okay with just making this change, but it would be nice to confirm.

jovaniks:
> Looks much better! Just checking: since you changed the dark mode value of this color instead of adding an extra legacy semantic color, did you check on the other places this color is used? My guess is this version will look better there, too, but only if we're using black text in all those places. Since this probably looks better everywhere, I'm okay with just making this change, but it would be nice to confirm.

Yes, I did a search for all the usages of `ksr_trust_500`, and in every case it's used as a background color paired with `ksr_white` for the text. Given that, I decided to update the dark mode value directly instead of introducing a new legacy color variant.

amy-at-kickstarter (approved):
I have the same question as Ingerid - looks like this color isn't used very frequently, though. It'd be great if you could spot check it!

PR #2472, opened 2025-06-11
Fix crash when dark mode is enabled

amy-at-kickstarter:
# 📲 What

The HTML initialization method of `NSAttributedString` isn't thread safe (and, I suspect, also isn't background safe). With dark mode enabled, the app takes a light/dark screenshot pair when the app moves into the background. This triggers a trait collection change, which triggers a call to `bindStyles`, which ultimately triggered this attributed string to crash.

# 🤔 Why

Fixing this will allow us to patch the app and turn dark mode back on.

# 👀 See

This crash has existed in our app for a long time, Dark Mode merely made it worse. The biggest culprit in our app is  `String+SimpleHTML`. I'd like to apply a more general fix there, but I'm frustratingly, I'm unable to duplicate this issue reliably. Even more interestingly, that code _already_ checks that we're already on the main thread, so it's not clear to me _how_ it's crashing. Suggestions for repro steps and debugging welcome!


Kickstarter-iOS/Features/PledgeView/Views/PledgeViewCTAContainerView.swift
ifosli, line 1: There's also an html-based attributed string related to plot that should be moved out of bind styles: `self.pledgeImmediatelyLabel.attributedText = pledgeImmediatelyText()`

ifosli (approved):
There's also the `applyDisclaimerTextViewStyle` that creates an NSAttributedString behind the scenes. This is only used in `FacebookConfirmationViewController` so I very much doubt we hit it often, but it might be nice to fix it while we're fixing stuff. iirc, this one doesn't even use our html helper, so it could cause the dark mode crash or the other variety.

stevestreza-ksr (approved):

scottkicks (commented):
✅

PR #2476, opened 2025-06-12
Secret reward's autoscrolling

jovaniks:
# 📲 What

Add support for automatic scroll to the first secret reward when opening the project via a deeplink that includes a `secretRewardToken`.

# 🛠 How

- Extended `goToRewards` output from `ProjectPageViewModel` to include `secretRewardToken`.
- Propagated the token through `ProjectPageViewController` → `RewardsCollectionViewController` → `RewardsCollectionViewModel`.
- Replaced `scrollToBackedRewardIndexPath` with a more general-purpose `scrollToRewardIndexPath`.
- Modified the scroll logic to prioritize secret reward index if a token is present; otherwise fall back to previously backed reward (existing behaviour).

# 👀 See

- [MBL-2451](https://kickstarter.atlassian.net/browse/MBL-2451)

Simulator Screen Recording - iPhone 16 Pro - 2025-06-12 at 09 34 07

# ✅ Acceptance criteria

- [x] Scrolling occurs automatically when `secretRewardToken` is passed
- [x] Only triggers if project has at least one secret reward
- [x] Defaults to existing scroll-to-backed-reward behavior otherwise

### DIFF
--- a/Library/ViewModels/RewardsCollectionViewModel.swift
+++ b/Library/ViewModels/RewardsCollectionViewModel.swift
@@ -9,7 +9,12 @@ public enum RewardsCollectionViewContext {
 }
 
 public protocol RewardsCollectionViewModelInputs {
-  func configure(with project: Project, refTag: RefTag?, context: RewardsCollectionViewContext)
+  func configure(
+    with project: Project,
+    refTag: RefTag?,
+    context: RewardsCollectionViewContext,
+    secretRewardToken: String?
+  )
   func confirmedEditReward()
   func pledgeShippingLocationViewControllerDidUpdate(_ shimmerLoadingViewIsHidden: Bool)
   func rewardCellShouldShowDividerLine(_ show: Bool)
@@ -33,7 +38,7 @@ public protocol RewardsCollectionViewModelOutputs {
   var reloadDataWithValues: Signal<[RewardCardViewData], Never> { get }
   var rewardsCollectionViewIsHidden: Signal<Bool, Never> { get }
   var rewardsCollectionViewFooterIsHidden: Signal<Bool, Never> { get }
-  var scrollToBackedRewardIndexPath: Signal<IndexPath, Never> { get }
+  var scrollToRewardIndexPath: Signal<IndexPath, Never> { get }
   var shippingLocationViewHidden: Signal<Bool, Never> { get }
   var showEditRewardConfirmationPrompt: Signal<(String, String), Never> { get }
   var title: Signal<String, Never> { get }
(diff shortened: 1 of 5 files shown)


Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -452,7 +478,27 @@ private func shouldTriggerEditRewardPrompt(_ data: PledgeViewData) -> Bool {
...
+    return firstSecretRewardIndexPath(rewards: rewards)
+  }
+
+  return backedRewardIndexPath(project, rewards: rewards)
+}
+
+private func firstSecretRewardIndexPath(rewards: [Reward]) -> IndexPath? {
+  return rewards.firstIndex(where: { $0.isSecretReward })
+    .flatMap { IndexPath(row: $0, section: 0) }
+}
```
scottkicks, line 499: Are secret rewards provided to us in any particular order?
I'm wondering, if a project has multiple secret rewards, would we want to always scroll to the first one? Or would tokens be associated with specific secret rewards so that we could scroll to the most relevant one?

Library/ViewModels/RewardsCollectionViewModelTests.swift
```diff
@@ -33,7 +33,7 @@ final class RewardsCollectionViewModelTests: TestCase {
     let testProject = Project.template
       |> Project.lens.rewardData.rewards .~ rewards
 
-    self.vm.configure(with: testProject, refTag: nil, context: .createPledge)
+    self.vm.configure(with: testProject, refTag: nil, context: .createPledge, secretRewardToken: "34342")
```
scottkicks, line 39: Do we need to add any test cases to cover when secretRewardTokens are/are not present?

scottkicks (changes requested):
👍 Just want to double check that we have tests for this new behavior.

Also added a non-blocking, more general, question.

Library/ViewModels/RewardsCollectionViewModelTests.swift
jovaniks: Good call! I’ve added tests to cover the following cases:
	• Scrolls to the first secret reward when a secret reward token is provided.
	• Scrolls to the backed reward when the project is backed.
	• Does not scroll when the project is not backed and no secret reward token is present.

Library/ViewModels/RewardsCollectionViewModel.swift
jovaniks: As far as I know, there's currently no way to associate a token with a specific secret reward — they're not ordered or linked that way. To support that behavior, we'd likely need some changes on the backend.

scottkicks (approved):
Thanks for adding the tests!

PR #2477, opened 2025-06-16
MBL-915: Replace LegacyInterceptorProvider

amy-at-kickstarter:
# 📲 What

Apollo 1.0 and beyond remove the LegacyInterceptorProvider class. This replaces it, and also removes LegacyParsingInterceptor from our interceptor chain, which has no equivalent in 1.0.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests

### DIFF
--- a/KsApi/extensions/ApolloInterceptors.swift
+++ b/KsApi/extensions/ApolloInterceptors.swift
@@ -1,27 +1,60 @@
 import Apollo
+import Foundation
 
-// MARK: - NetworkInterceptorProvider
-
-class NetworkInterceptorProvider: LegacyInterceptorProvider {
+/// This is based on `LegacyInterceptorProvider` from Apollo version 0.x.
+/// That's provider is longer included in Apollo 1.x, but we want to continue to have the same
+/// behavior. This could, potentailly, be cleaned up to remove unnecessary interceptors.
+class
+NetworkInterceptorProvider: InterceptorProvider {
   private let additionalHeaders: () -> [String: String]
+  private let client: URLSessionClient
+  private let store: ApolloStore
 
-  override func interceptors<Operation: GraphQLOperation>(for operation: Operation) -> [ApolloInterceptor] {
-    return [HeadersInterceptor(self.additionalHeaders)] + super.interceptors(for: operation)
+  /// Designated initializer
+  ///
+  /// - Parameters:
+  ///   - store: The `ApolloStore` to use when reading from or writing to the cache. Make sure you pass the same store to the `ApolloClient` instance you're planning to use.
+  public init(
+    store: ApolloStore,
+    additionalHeaders: @escaping () -> [String: String]
+  ) {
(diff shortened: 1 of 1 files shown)


KsApi/extensions/ApolloInterceptors.swift
```diff
@@ -1,27 +1,60 @@
...
+import Foundation
 
-// MARK: - NetworkInterceptorProvider
-
-class NetworkInterceptorProvider: LegacyInterceptorProvider {
+/// This is based on `LegacyInterceptorProvider` from Apollo version 0.x.
+/// That's provider is longer included in Apollo 1.x, but we want to continue to have the same
+/// behavior. This could, potentailly, be cleaned up to remove unnecessary interceptors.
+class
+NetworkInterceptorProvider: InterceptorProvider {
```
amy-at-kickstarter: This class is now an amalgamation of what we had implemented in `NetworkInterceptorProvider`, and what Apollo provided in `LegacyInterceptorProvider`: https://github.com/apollographql/apollo-ios/blob/0.44.0/Sources/Apollo/InterceptorProvider.swift#L31-L72

Essentially, I copy-pasted `LegacyInterceptorProvider`, harcoded the defaults we weren't changing (`shouldInvalidateClientOnDeinit` and `client`), and added `HeadersInterceptor` to the stack.

It _should_ be 100% equivalent to the old `InterceptorProvider`, just rejiggered a bit.

amy-at-kickstarter:
Regarding feature flags! I'm considering gating this behind a flag, but because the Apollo client is set up at launch, I'm not really sure it would do us much good. Thoughts?

KsApi/extensions/ApolloInterceptors.swift
stevestreza-ksr: Funky line breaking here

KsApi/extensions/ApolloInterceptors.swift
```diff
@@ -1,27 +1,60 @@
...
+      NetworkFetchInterceptor(client: self.client),
+      ResponseCodeInterceptor(),
+      LegacyParsingInterceptor(cacheKeyForObject: self.store.cacheKeyForObject),
+      AutomaticPersistedQueryInterceptor(),
+      LegacyCacheReadInterceptor(store: self.store)
+    ]
+  }
+
+  func additionalErrorInterceptor<Operation: GraphQLOperation>(for _: Operation) -> ApolloErrorInterceptor? {
+    return nil
```
stevestreza-ksr, line 45: Maybe toss in a FIXME to do something more useful with error handling if this is where we can collect them for any query.

PR #2480, opened 2025-06-17
[MBL-2498] Include 'Incremental' in UpdateBackingInput

scottkicks:
# 📲 What

Add an `incremental: Boolean` argument to the updateBacking mutation.

# 🤔 Why

This flag will allow the backend to distinguish between standard and incremental (PLOT) backing updates when backers are updating their pledges.

# 🛠 How

Updates the UpdateBackingInput, Constructor, and call site in PledgeViewModel.

# ✅ Acceptance criteria

Added some test code to put the "Edit Reward" option back in the manage pledge flow and verified that the `incremental` field returns `true` for PLOT pledges and `false` for other pledges.

- [x] New incremental field on UpdateBackingInput is always `false`. (until we start building this feature out more.
- [x] Updating any existing peldge, that isn't a PLOT pledge, still functions as expected.
- [x] All tests pass


Library/ViewModels/PledgeViewModel.swift
```diff
@@ -588,7 +589,9 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs,
       selectedShippingRule,
       self.paymentMethodsUseCase.dataOutputs.selectedPaymentSource,
       applePayParamsData,
-      context
+      context,
+      self.pledgeOverTimeUseCase.outputs
+        .showPledgeOverTimeUI /// Used in UpdateBackingData.increment to inform the backend if the updated pledge is a PLOT pledge.
```
scottkicks: I think it'll be nicer to have a new output dedicated to informing this view model that PLOT has been selected in an edit (update) context.

We have a ticket to update logic on our checkout screen [here](https://kickstarter.atlassian.net/browse/MBL-2500) so I think we can make that update then.

KsApi/mutations/inputs/UpdateBackingInput.swift
```diff
@@ -16,6 +17,7 @@ public struct UpdateBackingInput: GraphMutationInput, Encodable {
    - parameter applePay: The optional ApplePayParams to update.
    - parameter id: The GraphID of the Backing.
    - parameter locationId: The optional ID of the ShippingRule's Location.
+   - parameter incremental: The Boolean that represents the type of backing made. Regular or Incremental (PLOT).
```
jovaniks, line 20: nice

Library/ViewModels/PledgeViewModel.swift
jovaniks: I think we can use `selectedPaymentPlan` to check if the user's selection is PLOT or not.

Eg.
```swift
incremental: selectedPaymentPlan == .pledgeOverTime
```

Library/UpdateBackingInput+ConstructorTests.swift
```diff
@@ -311,7 +320,8 @@ final class UpdateBackingInput_ConstructorTests: TestCase {
       paymentSourceId: "123",
       setupIntentClientSecret: nil,
       applePayParams: applePayParams,
-      pledgeContext: .fixPaymentMethod
+      pledgeContext: .fixPaymentMethod,
+      incremental: false
     )
 
     let input = UpdateBackingInput.input(from: data, isApplePay: true)
```
amy-at-kickstarter, line 327: Can you add a test where `incremental` is `true`?

amy-at-kickstarter (approved):
LGTM, could just use another test or two.

PR #2484, opened 2025-06-18
Add FetchLocationsUseCase

amy-at-kickstarter:
# 📲 What

Add `FetchLocationsUseCase`. 

This is a small use case for getting default and suggested locations from the server. It will be used by `SearchFiltersUseCase`.

# 🤔 Why

Pulling this behavior out into its own use case makes it easy to test these network calls, and easier to test and hook up `SearchFiltersUseCase`.

### DIFF
--- a/KsApi/models/graphql/adapters/Location+LocationFragment.swift
+++ b/KsApi/models/graphql/adapters/Location+LocationFragment.swift
@@ -20,11 +20,8 @@ extension Location {
     }
 
     return nodes.compactMap { node in
-      if let fragment = node?.fragments.locationFragment {
-        KsApi.Location.location(from: fragment)
-      } else {
-        nil
-      }
+      guard let fragment = node?.fragments.locationFragment else { return nil }
+      return KsApi.Location.location(from: fragment)
     }
   }
 
@@ -34,11 +31,8 @@ extension Location {
     }
 
     return nodes.compactMap { node in
-      if let fragment = node?.fragments.locationFragment {
-        KsApi.Location.location(from: fragment)
-      } else {
-        nil
-      }
+      guard let fragment = node?.fragments.locationFragment else { return nil }
+      return KsApi.Location.location(from: fragment)
     }
   }
 }
--- a/KsApi/models/graphql/adapters/Location+LocationFragment.swift
+++ b/KsApi/models/graphql/adapters/Location+LocationFragment.swift
@@ -13,4 +13,32 @@ extension Location {
       name: locationFragment.name
     )
   }
+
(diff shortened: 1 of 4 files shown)


scottkicks (approved):
Cool! I wonder if this could be adapted for the locations selector on the rewards screen.

amy-at-kickstarter:
> Cool! I wonder if this could be adapted for the locations selector on the rewards screen.

Good idea!

KsApi/models/graphql/adapters/Location+LocationFragment.swift
```diff
@@ -13,4 +13,32 @@ extension Location {
...
+      return []
+    }
+
+    return nodes.compactMap { node in
+      if let fragment = node?.fragments.locationFragment {
+        KsApi.Location.location(from: fragment)
+      } else {
+        nil
+      }
+    }
```
jovaniks, line 25: Just a minor suggestion, not a big deal at all. This version might look a bit cleaner/fancier from my perspective:

```swift
  return nodes.compactMap { node in
    guard let fragment = node?.fragments.locationFragment else { return nil }
    return KsApi.Location.location(from: fragment)
  }
```

KsApi/models/graphql/adapters/Location+LocationFragment.swift
```diff
@@ -13,4 +13,32 @@ extension Location {
...
+  }
```
jovaniks, line 37: Same thing here, but again - just my inclination/preference

Library/UseCases/FetchLocations/FetchLocationsUseCase.swift
```diff
@@ -0,0 +1,72 @@
...
+public protocol FetchLocationsUseCaseDataOutputs {
+  /// Emits the list of default locations. Emits empty array on initial signal.
+  var defaultLocations: Signal<[Location], Never> { get }
+
+  /// Emits locations related to the query inputted with `searchedForLocations`. Emits empty array on initial signal.
+  var suggestedLocations: Signal<[Location], Never> { get }
+}
+
+/// A use case for fetching locations for the purpose of filtering projects.
+public final class FetchLocationsUseCase: FetchLocationsUseCaseType, FetchLocationsUseCaseInputs,
```
jovaniks, line 23: Love this `FetchLocationsUseCase`

jovaniks (approved):
Love this new usecase ❤️. Just a couple of suggestion, not blockers!

KsApi/models/graphql/adapters/Location+LocationFragment.swift
amy-at-kickstarter: Yeah, I like that better too! Will fix.

PR #2494, opened 2025-06-24
Refactor LocationView to improve clarity and fix a couple of bugs

amy-at-kickstarter:
# 📲 What

* Move `DefaultLocationsList` back into `LocationView`, instead of making it its own component. This lets me make a function call instead of passing more data via a `@Binding`.
* Fixed interactions between the default search locations, and locations set in the search bar
* Moved the search bar text into a `@State` in `LocationView`

# 🤔 Why

I had several bugs where the search bar and the default search locations weren't playing nice with each other.  See inline comments for a broader description.


Kickstarter-iOS/Features/SortFilter/LocationView/LocationView.swift
```diff
@@ -30,23 +63,48 @@ public struct LocationView: View {
           .searchCompletion(location.displayableName)
       }
     }
+    .onAppear {
+      self.searchText = self.selectedLocation?.displayableName ?? ""
```
amy-at-kickstarter, line 67: Instead of having the `searchText` owned by `FilterRootView`, I moved it back into `LocationView` and update it with the correct value `onAppear`. This ownership patterns makes more sense to me than what I had before.

Kickstarter-iOS/Features/SortFilter/LocationView/LocationView.swift
```diff
@@ -30,23 +63,48 @@ public struct LocationView: View {
...
+  }
+
+  func didChangeSearchText(_ newValue: String) {
+    // If the search bar is already showing the displayableName
+    // of the selected location, no action is needed.
+    if let location = self.selectedLocation {
+      if location.displayableName == newValue {
         return
       }
-
```
amy-at-kickstarter, line 40: One thing I've been experimenting with in SwiftUI is moving away from MVVM - this view is the closest I've come to deciding I _needed_ a view model again. Ultimately, I tried it out, and it didn't make the code any more clear or less buggy. Hence, all these verbose comments, instead.

The big issue I've been having: the search bar is powered by the `@Binding var searchText: String`. We need to respond to changes in the `searchText` (updating the selected location, updating autosuggest results). But we also need to update the `searchText` manually (when a location is selected from the defaults list). Because `onChange` doesn't differentiate between user-generated changes and programmatically generated changes, I kept running into weird and thorny little bugs where the state of this screen would get all mixed up.

I eventually hammered it out, and now it works great, but I don't love how this solution felt a bit fragile. It's unclear if this is a philosophical limitation of `@Binding`, or if it would be fixed by moving to later versions of SwiftUI which support more complex `onChange` behavior.

Kickstarter-iOS/Features/SortFilter/FilterRootView/FilterRootView.swift
```diff
@@ -63,9 +62,6 @@ struct FilterRootView: View {
     }
 
     self.searchFilters = searchFilters
-
-    let selectedLocationName = searchFilters.location.selectedLocation?.displayableName
-    _locationSearchText = State(initialValue: selectedLocationName ?? "")
```
amy-at-kickstarter, line 68: This worked fine, but this pattern is somewhat dangerous. So refactored it out in favor of `onAppear` to get the behavior I needed.

Kickstarter-iOS/Features/SortFilter/LocationView/DefaultLocationsList.swift
amy-at-kickstarter, line 1: Deleted and moved into `LocationView`.

Kickstarter-iOS/Features/SortFilter/LocationView/LocationView.swift
```diff
@@ -30,23 +63,48 @@ public struct LocationView: View {
...
+    self.selectedLocation = nil
+    self.onSearchedForLocations(newValue)
+  }
+
+  func selectedDefaultLocation(_ location: Location?) {
+    self.selectedLocation = location
+
+    // If a location is selected from the default location list,
+    // update the search bar.
+    // This will also trigger `didChangeSearchText`.
```
amy-at-kickstarter, line 103: This is the thing that drove me a bit mad.

ifosli (approved):
Looks good! Thanks for explaining all the things!

Kickstarter-iOS/Features/SortFilter/LocationView/LocationView.swift
scottkicks: Really interesting! I've seen some devs beginning to move away from MVVM, usually to a Model-View approach, but I'm not sure it's the way to go when views need to react in more complicated ways like this.

Are these limitations of `onChange` solved with the latest SwiftUI? Or are you saying that they may be addressed in the future, at which point we can consider adopting your other approach?

scottkicks (approved):

Kickstarter-iOS/Features/SortFilter/LocationView/LocationView.swift
amy-at-kickstarter: iOS 17 adds a couple new versions of `onChange` and changes its behavior slightly - including one that sends both the new and old values. I haven't tried writing this with the later version of `onChange`, but I do think having the new/old values would make it a little easier to reason about what's happening here. 

Still, it doesn't address the fundamental weirdness of having a read-write `@Binding` like this.

amy-at-kickstarter: @stevestreza-ksr Curious about your take on this as well.

PR #2495, opened 2025-06-25
Add a new query to fetch the PLOT-related date in the Manage Pledge flow

jovaniks:
# 📲 What

Adds a new GraphQL query called `FetchProjectPledgeOverTimeData` to fetch PLOT-related project information from the backend.

# 🤔 Why

The Manage Pledge flow for PLOT-enabled projects requires certain project-level PLOT fields that are currently only available via GraphQL.

Since updating the existing V1 endpoint is not feasible at this time, this new query will allow us to fetch just the necessary PLOT fields while keeping the rest of the flow intact.

This query enables the function `fetchProjectPledgeOverTimeData` to retrieve the required information when editing PLOT-enabled pledges.

# 🛠 How
- Introduced a new query: `FetchProjectPledgeOverTimeData`
	
```graphql
query FetchProjectPledgeOverTimeData($projectId: Int!) {
  project(pid: $projectId) {
    __typename
    isPledgeOverTimeAllowed
    pledgeOverTimeCollectionPlanChargeExplanation
    pledgeOverTimeCollectionPlanChargedAsNPayments
    pledgeOverTimeCollectionPlanShortPitch
    pledgeOverTimeMinimumExplanation
  }
}
```
# 👀 See

- [MBL-2548](https://kickstarter.atlassian.net/browse/MBL-2548)

Simulator Screen Recording - iPhone 16 Pro - 2025-06-25 at 08 23 36

# ⏰ TODO

- [ ] Ideally move forward to swift the fetch project from V1 to GraphQL

[MBL-2548]: https://kickstarter.atlassian.net/browse/MBL-2548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KsApi/ServiceType.swift
```diff
@@ -240,6 +240,8 @@ public protocol ServiceType {
   func fetchProject(projectParam: Param, configCurrency: String?)
     -> SignalProducer<Project.ProjectPamphletData, ErrorEnvelope>
 
+  func fetchProjectPlotData(projectId: Int) -> SignalProducer<ProjectPLOTDataEnvelope, ErrorEnvelope>
```
scottkicks: nit: I know we use `PLOT` elsewhere in the app, but I think we should start reducing the acronyms we use to make it easier to understand code context in the future for ourselves and newer engineers that might be onboarded.

Especially since we're not consistent across the board about when to use PLOT and when we use the full spelled out `pledgeOverTime` in our method/property names already. 

```suggestion
  func fetchProjectPledgeOverTimeData(projectId: Int) -> SignalProducer<ProjectPLOTDataEnvelope, ErrorEnvelope>
```

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -491,6 +494,28 @@ private func fetchProjectRewards(project: Project) -> SignalProducer<Project, Er
     }
 }
 
+private func fetchProjectPlotData(project: Project) -> SignalProducer<Project, ErrorEnvelope> {
```
scottkicks: nit: same suggestion here

```suggestion
private func fetchProjectPledgeOverTimeData(project: Project) -> SignalProducer<Project, ErrorEnvelope> {
```

KsApi/models/ProjectPLOTDataEnvelope.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import ReactiveSwift
+
+public struct ProjectPLOTDataEnvelope {
```
scottkicks: nit: same suggestion here

```suggestion
public struct ProjectPledgeOverTimeDataEnvelope {
```

KsApi/models/ProjectPledgeOverTimeDataEnvelope.swift
```diff
@@ -0,0 +1,37 @@
...
+extension ProjectPledgeOverTimeDataEnvelope {
+  static func envelopeProducer(
+    from data: GraphAPI.FetchProjectPledgeOverTimeDataQuery
+      .Data
+  ) -> SignalProducer<ProjectPledgeOverTimeDataEnvelope, ErrorEnvelope> {
+    guard let project = data.project else {
+      return SignalProducer(error: .couldNotParseJSON)
+    }
+
+    let envelope = ProjectPledgeOverTimeDataEnvelope(
```
amy-at-kickstarter, line 23: It looks like GraphQL defines all of these as optional. These envelope values should be `String?` instead of `String`, so then you can get rid of these empty string placeholder values.

amy-at-kickstarter (changes requested):
LGTM but requesting a small change in nullability!

KsApi/models/ProjectPledgeOverTimeDataEnvelope.swift
jovaniks: I initially had them as optional but changed them to non-optional to align with the Project model. No big deal, can revert if needed.

amy-at-kickstarter: Mostly I'm always a bit suspicious of passing around empty placeholder strings! I find `nil` clearer.

jovaniks: Great, I'll go ahead and make the change.

scottkicks (approved):

KsApi/models/ProjectPledgeOverTimeDataEnvelope.swift
jovaniks: Done! PLOT-related fields are now nullable both in `ProjectPledgeOverTimeDataEnvelope` and in the `Project` model.

KsApi/models/graphql/adapters/GraphAPI.UpdateBackingInput+UpdateBackingInput.swift
```diff
@@ -9,7 +9,8 @@ extension GraphAPI.UpdateBackingInput {
       locationId: input.locationId,
       paymentSourceId: input.paymentSourceId,
       intentClientSecret: input.setupIntentClientSecret,
-      applePay: GraphAPI.ApplePayInput.from(input.applePay)
+      applePay: GraphAPI.ApplePayInput.from(input.applePay),
+      incremental: input.incremental
```
jovaniks, line 13: Adds the missing incremental field to ensure we can pass this value through the `UpdateBacking` mutation.

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -94,6 +94,14 @@ public final class ManagePledgeViewModel:
           .switchMap { project in
             fetchProjectRewards(project: project)
           }
+          .switchMap { project in
+            // Only fetch pledge over time data if the feature flag is enabled
+            guard featureEditPledgeOverTimeEnabled() else {
+              return SignalProducer(value: project)
+            }
+
+            return fetchProjectPledgeOverTimeData(project: project)
```
amy-at-kickstarter, line 103: If I'm understanding correctly, this will fire three requests in a row for PLOT projects: the first V1 request to fetch the project, then a second GraphQL request to fetch the rewards, and a third GraphQL request to fetch the PLOT data. Chaining together requests like this is ultimately going to lead to a much slower user experience, since it's three network round trips. 

Since `fetchProjectRewards` is a GraphQL request already, could you fold the `pledgeOverTime` fragment into the second request?

I don't want to block this PR on this, so perhaps you can file a follow-up ticket for this issue. I assume PLOT pledges are a minority of pledges, but I do think this is worth fixing.

amy-at-kickstarter (approved):
Largely LGTM but I have a specific refactoring or follow-up request.

Library/ViewModels/ManagePledgeViewModel.swift
jovaniks: That's a good call!!! I will go for it on Monday in a follow-up ticket!

KsApi/models/ProjectPledgeOverTimeDataEnvelope.swift
jovaniks: Follow-up ticket https://kickstarter.atlassian.net/browse/MBL-2578

PR #2496, opened 2025-06-25
MBL-2521: Support pledged filter in GraphQL and DiscoveryParams

amy-at-kickstarter:
# 📲 What

Add the `pledged` parameter to `SearchQuery`, and code to `DiscoveryParams` to support it. 

# 🤔 Why

`pledged` will power the "Amount Raised" filters.

### DIFF
--- a/KsApi/models/DiscoveryParams.swift
+++ b/KsApi/models/DiscoveryParams.swift
@@ -21,6 +21,7 @@ public struct DiscoveryParams {
   public var state: State?
   public var percentRaised: PercentRaisedBucket?
   public var location: Location?
+  public var pledged: PledgedBucket?
 
   public enum State: String, Decodable {
     case all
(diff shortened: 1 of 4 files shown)


scottkicks (approved):

KsApi/models/DiscoveryParams.swift
```diff
@@ -59,6 +60,23 @@ public struct DiscoveryParams {
...
+    case bucket_1
+
+    /// Range from 10,000 to 100,000 USD
+    case bucket_2
+
+    /// Range from 100,000 to 1,000,000 USD
+    case bucket_3
+
+    /// Range from 1,000,000 to Infinity USD
+    case bucket_4
```
stevestreza-ksr, line 77: Not super loving these names but not really sure if there's a better one that would fit these criteria

PR #2498, opened 2025-06-25
[MBL-2500] Add Edit PLOT PledgeViewContext

scottkicks:
# 📲 What

- Adds a new `PledgeViewContext.editPledgeOverTime` that gets passed to the checkout flow via `PledgeViewData`.
- Also updates logic that replaces "Edit reward" with "Edit pledge" so that it is shown for all backed PLOT allowed projects instead of just PLOT pledges.

# 🤔 Why

- This will make sure that the correct UI (other than the PLOT component itself) is hidden/shown on the pledge view screen when in the Edit PLOT context. 
   - This follows the same behavior as the updateReward context.
   - Logic to show the PLOT component is being worked on in a separate PR.
- We want backers to be able to update their peldges from PLOT -> Pledge in full _and_ Pledge in full -> PLOT.

# ✅ Acceptance criteria

- [x] Checkout screen matches that of the update reward flow
- [x] "Edit pledge" menu option shows for all PLOT pledged projects in the ManagePledge flow.

### DIFF
--- a/KsApi/GraphAPI.swift
+++ b/KsApi/GraphAPI.swift
@@ -4641,6 +4641,7 @@ public enum GraphAPI {
     case chCurrencySelector
     case dkCurrencySelector
     case noCurrencySelector
+    case plUsdCurrency_2025
     case seCurrencySelector
     case datalakeFeEvents
     case creatorDemographicsSurvey
@@ -4767,6 +4768,7 @@ public enum GraphAPI {
         case "ch_currency_selector": self = .chCurrencySelector
         case "dk_currency_selector": self = .dkCurrencySelector
         case "no_currency_selector": self = .noCurrencySelector
+        case "pl_usd_currency_2025": self = .plUsdCurrency_2025
         case "se_currency_selector": self = .seCurrencySelector
         case "datalake_fe_events": self = .datalakeFeEvents
         case "creator_demographics_survey": self = .creatorDemographicsSurvey
@@ -4894,6 +4896,7 @@ public enum GraphAPI {
         case .chCurrencySelector: return "ch_currency_selector"
         case .dkCurrencySelector: return "dk_currency_selector"
         case .noCurrencySelector: return "no_currency_selector"
+        case .plUsdCurrency_2025: return "pl_usd_currency_2025"
         case .seCurrencySelector: return "se_currency_selector"
         case .datalakeFeEvents: return "datalake_fe_events"
         case .creatorDemographicsSurvey: return "creator_demographics_survey"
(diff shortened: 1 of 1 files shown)


Library/ViewModels/RewardsCollectionViewModel.swift
```diff
@@ -164,6 +164,13 @@ public final class RewardsCollectionViewModel: RewardsCollectionViewModelType,
         featurePostCampaignPledgeEnabled() && project.isInPostCampaignPledgingPhase
           ? PledgeViewContext.latePledge
           : PledgeViewContext.pledge
+
+      /// Differentiating between updating a reward for a regular pledge and updating a Pledge Over Time pledge.
+      let isPledgeOverTime = project.personalization.backing?.paymentIncrements.isEmpty == false
```
jovaniks: I'm thinking we should apply the same validation as in `ManagePledgeViewModel`.
So the context for `updatePledgeContext` would be based on PLOT-allowed projects, not just PLOT-pledges.
That way, backers with a **Pledge-In-Full** pledge would still have the ability to switch to **Pledge-Over-Time** if the project is PLOT allowed.

Does that sound right?

```
let isPledgeOverTime = project.isPledgeOverTimeAllowed == true
```

scottkicks: Ah yes good catch!

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -529,7 +529,8 @@ private func actionSheetMenuOptionsFor(
   var actions = ManagePledgeAlertAction.allCases
     .filter { $0 != .viewRewards && $0 != .editPledgeOverTimePledge }
 
-  if isPledgeOverTime(with: backing) {
```
amy-at-kickstarter, line 532: Can this function be cleaned up, or is it still useful elsewhere?

Library/ViewModels/PledgeViewModel.swift
```diff
@@ -615,7 +615,7 @@ public class PledgeViewModel: PledgeViewModelType, PledgeViewModelInputs,
         setupIntentClientSecret: nil,
         applePayParams: applePayParams,
         context,
-        selectedPaymentPlan == .pledgeOverTime
+        incremental: selectedPaymentPlan == .pledgeOverTime
```
amy-at-kickstarter, line 618: 👍 Nice to have this named

Library/ViewModels/ManagePledgeViewModel.swift
scottkicks: Yeah It's still being used to determine when to show the payment schedule on the Manage Pledge screen, so better to keep it for now.

jovaniks: That's right! The validation Scott mentioned still needs to stay for now. Maybe we could consider renaming it to something like `isPledgeOverTimePledge` to make the function’s purpose clearer.

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -529,7 +529,8 @@ private func actionSheetMenuOptionsFor(
...
+  /// Enable the 'Edit pledge' option for all PLOT-enabled projects.
+  if project.isPledgeOverTimeAllowed == true {
```
jovaniks, line 533: Nice!!!

jovaniks (approved):
Cool!!!

PR #2505, opened 2025-06-26
MBL-2521, MBL-2519, MBL-2514: Add filter for amount raised

amy-at-kickstarter:
# 📲 What

Add filtering by amount raised.

amount-raised

# 🤔 Why

Another phase of search filters!

### DIFF
--- a/Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
+++ b/Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
@@ -1,0 +1,39 @@
+import KsApi
+import Library
+import SwiftUI
+
+public struct AmountRaisedView: View {
+  var buckets: [DiscoveryParams.AmountRaisedBucket]
+  @Binding var selectedBucket: DiscoveryParams.AmountRaisedBucket?
+
+  public var body: some View {
+    VStack(alignment: .leading, spacing: Constants.spacing) {
+      ForEach(self.buckets) { bucket in
+        Button {
+          self.selectedBucket = bucket
+        } label: {
+          HStack(spacing: Constants.buttonLabelSpacing) {
+            RadioButton(isSelected: bucket == self.selectedBucket)
+            Text(bucket.title)
+              .font(InterFont.bodyLG.swiftUIFont())
+              .foregroundStyle(Colors.Text.primary.swiftUIColor())
+          }
+        }
+      }
+    }
+    .padding(Constants.padding)
+    .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
+  }
+
+  internal enum Constants {
+    static let padding: CGFloat = 24.0
+    static let spacing: CGFloat = 24.0
+    static let buttonLabelSpacing: CGFloat = 8.0
+  }
+}
+
+extension DiscoveryParams.AmountRaisedBucket: @retroactive Identifiable {
+  public var id: Int {
+    return self.hashValue
+  }
+}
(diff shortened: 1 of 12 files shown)


Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
```diff
@@ -0,0 +1,39 @@
+import KsApi
+import Library
+import SwiftUI
+
+public struct AmountRaisedView: View {
+  var buckets: [DiscoveryParams.AmountRaisedBucket]
+  @Binding var selectedBucket: DiscoveryParams.AmountRaisedBucket?
+
+  public var body: some View {
```
amy-at-kickstarter, line 9: I tinkered with making a common element for these radio button lists, but it actually didn't make the code any _less_ verbose. I'm going to sit on that PR for a bit and decide if it seems worth it or not.

For now, this is virtually identical to `PercentRaisedView`.

scottkicks (approved):
LGTM

Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
```diff
@@ -0,0 +1,39 @@
...
+  internal enum Constants {
+    static let padding: CGFloat = 24.0
+    static let spacing: CGFloat = 24.0
+    static let buttonLabelSpacing: CGFloat = 8.0
+  }
+}
+
+extension DiscoveryParams.AmountRaisedBucket: @retroactive Identifiable {
+  public var id: Int {
+    return self.hashValue
```
stevestreza-ksr: What's going on here? Are we ever expecting this value to change? The id should be consistent if the value is mutated, SwiftUI will use this to add/remove cells rather than just updating existing ones, which can break animations.

Kickstarter-iOS/Features/SortFilter/FilterRootView/FilterRootView.swift
```diff
@@ -141,6 +162,13 @@ struct FilterRootView: View {
     )
   }
 
+  var amountRaisedModal: some View {
```
stevestreza-ksr, line 169: For the sake of consistency, can all of these be updated to be `@ViewBuilder` types? They aren't strictly needed right now but if someone updates these later it might break in a way that isn't obvious to fix.

Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
amy-at-kickstarter: Nope, I wouldn't expect it to change - since `AmountRaisedBucket` is an `enum`. This is just here for the `ForEach` which displays the radio button for all of these.

I suppose it would be cleaner to do it like

```swift
public enum AmountRaisedBucket: Int, Identifiable {
  case bucket_0 = 1
  //etc

  public var id: Int {
    return self.rawValue
  }
}
```

Would that make more sense to you?

Kickstarter-iOS/Features/SortFilter/FilterRootView/FilterRootView.swift
amy-at-kickstarter: Can do!

Kickstarter-iOS/Features/SortFilter/AmountRaisedView/AmountRaisedView.swift
amy-at-kickstarter: @stevestreza-ksr Went ahead and just fixed that for all the `enum`s that use `Identifiable`, much clearer now: c60b6860b

jovaniks (approved):
LGTM!

PR #2507, opened 2025-06-30
Hotfix: Preselect the original payment plan when editing a Pledge Over Time

jovaniks:
# 📲 What

This PR improves the default selection logic for the payment plan in the "Edit Pledge Over Time" flow.

Specifically:
	- The default selected plan will now reflect the backer's original selection (i.e., if they selected **Pledge In Full** originally, that will remain selected when editing).
	- Previously, the default was always **Pledge In Full**, regardless of the backer's choice.

# 🛠 How

This change uses `backing.paymentIncrements.isEmpty` to determine whether **Pledge Over Time** or **Pledge In Full** should be preselected.

- Retain PLOT pre-selection only if the new pledge amount meets the eligibility threshold (amountIsPledgeOverTimeEligible)

Added a new variable:

```swift
let isPledgeOverTimePreSelected = project.personalization.backing?.paymentIncrements.isEmpty == false
let isPledgeOverTimeEligible = paymentPlan.amountIsPledgeOverTimeEligible

let defaultPlan = isPledgeOverTimePreSelected && isPledgeOverTimeEligible ? 
    PledgePaymentPlansType.pledgeOverTime :
    PledgePaymentPlansType.pledgeInFull
```

- Also update `notChangingPaymentMethod` to include the `.editPledgeOverTime` context, this allow to the PLOT Edit Pledge to continue with the backing updates.

The `notChangingPaymentMethod` now looks like:
```swift
let notChangingPaymentMethod = context.map { context in
      if context.isUpdating {
        return context == .updateReward || context == .editPledgeOverTime
      }

      return false
    }
```

# 👀 See

Simulator Screen Recording - iPhone 16 Pro - 2025-06-30 at 12 40 18

# ✅ Acceptance criteria

- [ ] Pre-selected option load as default
- [ ] Retain PLOT pre-selection only if the new pledge amount meets the eligibility threshold (amountIsPledgeOverTimeEligible)

### DIFF
--- a/Kickstarter-iOS/Features/PledgeView/Controllers/PledgeViewController.swift
+++ b/Kickstarter-iOS/Features/PledgeView/Controllers/PledgeViewController.swift
@@ -377,6 +377,7 @@ final class PledgeViewController: UIViewController,
 
     self.localPickupLocationView.rac.hidden = self.viewModel.outputs.localPickupViewHidden
     self.paymentMethodsViewController.view.rac.hidden = self.viewModel.outputs.paymentMethodsViewHidden
+    self.paymentSectionLabel.rac.hidden = self.viewModel.outputs.paymentMethodsViewHidden
 
     self.viewModel.outputs.title
       .observeForUI()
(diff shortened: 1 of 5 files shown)


scottkicks (approved):
Can we also double check that this hasn't impacted the regular (non-edit) plot flow before merging?

PR #2509, opened 2025-07-01
MBL-915: Update all fragment tests in KsApi to use testGraphObject instead of initializers

amy-at-kickstarter:
# 📲 What

Replace direct calls to `GraphQLSelectSet.init(jsonObject:variables:)` with calls to `testGraphObject(jsonObject:variables)`

# 🤔 Why

Apollo deprecates `init(jsonObject:variables:)`. By consolidating all of these under a helper function, it's easier to clean them up for the Apollo upgrade.


KsApi/models/graphql/adapters/UserCreditCards+UserFragmentTests.swift
```diff
@@ -88,3 +32,311 @@ class UserCreditCards_UserFragmentTests: XCTestCase {
     }
   }
 }
+
+let validJSONWithStoredCards = """
```
amy-at-kickstarter, line 36: I had to re-record the JSON for this test to use `testGraphObject`.

ifosli (approved):

jovaniks (approved):
LGTM! I like the idea of using a helper method, it simplifies future updates 🤌

PR #2512, opened 2025-07-01
[MBL-2547] Show "Go to Pledge Manager" button for existing backers

jovaniks:
# 📲 What

Currently, we only show the go to pledge manager button on the project page for net new backers. We should also show it for existing backers!

# 🛠 How

Show go to pledge manager flow for net-new backers || existing backers.

# 👀 See

- [MBL-2547](https://kickstarter.atlassian.net/browse/MBL-2547)

Simulator Screen Recording - iPhone 16 Pro - 2025-07-01 at 12 47 06

# ✅ Acceptance criteria

- [x] "Go to Pledge Manager" should show for both **Net New backers** and **Existing Backers**

[MBL-2547]: https://kickstarter.atlassian.net/browse/MBL-2547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/KsApi/models/Project.swift
+++ b/KsApi/models/Project.swift
@@ -510,11 +510,11 @@ extension Project: GraphIDBridging {
 extension Project.Video: Decodable {}
 
 extension Project {
-  public var acceptsNetNewBackersForPM: Bool {
+  public var pledgeManagementAvailable: Bool {
     let isBacking = self.personalizationIsBacking ?? false
     let lastWaveActive = self.lastWave?.active ?? false
     let pmAcceptsNewBackers = self.pledgeManager?.acceptsNewBackers ?? false
 
-    return !isBacking && lastWaveActive && pmAcceptsNewBackers
+    return lastWaveActive && (pmAcceptsNewBackers || isBacking)
   }
 }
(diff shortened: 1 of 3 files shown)


KsApi/models/Project.swift
```diff
@@ -510,11 +510,11 @@ extension Project: GraphIDBridging {
...
 
 extension Project {
-  public var acceptsNetNewBackersForPM: Bool {
+  public var pledgeManagementAvailable: Bool {
     let isBacking = self.personalizationIsBacking ?? false
     let lastWaveActive = self.lastWave?.active ?? false
     let pmAcceptsNewBackers = self.pledgeManager?.acceptsNewBackers ?? false
 
-    return !isBacking && lastWaveActive && pmAcceptsNewBackers
+    return lastWaveActive && (pmAcceptsNewBackers || isBacking)
```
ifosli, line 518: nit: Should this show for all current backers? Ex if someone pledged without a reward, they may not have anything to fill out in the pledge manager (I'm not sure if they're able to access the pledge manager at all). I'm okay with keeping this logic for now, but I do think we should try to move the "should this user see the pledge manager" logic to the backend sooner rather than later.

jovaniks: Actually, I asked Leigh something similar. From what I understand, we want both backers and net new backers to have access to the pledge manager so they can still pledge for PM rewards. But I agree, there should also be a better path for existing backers to view their current pledge while still having the option to access the PM if needed.

PR #2514, opened 2025-07-02
Fix: Hide Collection Plan component on change payment method flow

jovaniks:
# 📲 What

Hide the Collection Plan component when changing the payment method.
Now, the Collection Plan will only be shown in .`pledge` and .`editPledgeOverTime` contexts.

# 🛠 How

- Updated PledgeOverTimeUseCase to listen to context as part of the eligibility check for showing the Pledge Over Time UI.
- Added `.isAny(of: [.pledge, .editPledgeOverTime])` validation.
- Modified tests in `PledgeOverTimeUseCaseTests` to cover the new behavior across different contexts.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-07-02 at 08 37 13 | Simulator Screen Recording - iPhone 16 Pro - 2025-07-02 at 08 35 37 |

# ✅ Acceptance criteria

- [ ] Collection Plan should only appear on `.pledge` and `.editPledgeOverTime` contexts.
- [ ] Should not appear on `.changePaymentMethod`, `.fixPaymentMethod`, `.latePledge`, `.update`, and `.updateReward`.
- [ ] Unit tests cover both showing and hiding scenarios.


Library/ViewModels/PledgeOverTimeUseCase.swift
```diff
@@ -42,9 +42,17 @@ public protocol PledgeOverTimeUseCaseType {
 
 public final class PledgeOverTimeUseCase: PledgeOverTimeUseCaseType, PledgeOverTimeUseCaseInputs,
   PledgeOverTimeUseCaseOutputs {
-  init(project: Signal<Project, Never>, pledgeTotal: Signal<Double, Never>) {
-    let pledgeOverTimeUIEnabled = project.signal
-      .map { ($0.isPledgeOverTimeAllowed ?? false) && featurePledgeOverTimeEnabled() }
+  init(
+    project: Signal<Project, Never>,
+    pledgeTotal: Signal<Double, Never>,
+    context: Signal<PledgeViewContext, Never>
```
jovaniks, line 48: Now we inject the context so the PLOT UI only shows when the context `. pledge` or `.editPledgeOverTime`

Library/ViewModels/PledgeOverTimeUseCase.swift
```diff
@@ -42,9 +42,17 @@ public protocol PledgeOverTimeUseCaseType {
...
+  ) {
+    let pledgeOverTimeUIEnabled = Signal.combineLatest(project, context)
+      .map { project, context -> Bool in
+        project.isPledgeOverTimeAllowed == true
+          && featurePledgeOverTimeEnabled()
+          && context.isAny(of: .pledge, .editPledgeOverTime)
```
ifosli, line 54: Should we also show the plot ui in the edit pledge (`.updateReward` and maybe `.update` if that still exists?) context? For example, if someone is editing their pledge for a project that allows pledge over time, but their initial pledge didn't use it, will they be able to easily change their pledge to one that uses PLOT?

jovaniks: Yeah, we're actually setting the `editPledgeOverTime` context for any project where PLOT is allowed, regardless of whether the original pledge was "Pledge in Full" or "Pledge Over Time".

PR #2516, opened 2025-07-03
MBL-915: Implement placeholders for GraphQLNullable and GraphQLEnum

amy-at-kickstarter:
# 📲 What

Add placeholder code for `GraphQLNullable` and `GraphQLEnum`.

# 🤔 Why

Apollo 1.0 adds two new wrapper types, `GraphQLNullable` and `GraphQLEnum`. These are used to wrap all optional inputs and enum inputs, respectively, to queries and mutations. `GraphQLNullable` differentiates between `.some(value)`, `.none` and `.null`, and `GraphQLEnum` differentiates between `.case(value)` and `.unknown`.

Our queries and mutations have quite a few optional inputs. That means fixing this is one of the larger manual changes we need to make to upgrade Apollo. To the best of my knowledge, I've never seen us pass a `null` to GraphQL, only `nil` - so my solution was to wrap all of our nullable inputs with a helper function that returns `.none` or `.some`, as appropriate. 

In this PR, I pulled out all those updates to add wrappers into one commit, and implemented a _placeholder_ wrapper (which does nothing). This way, we can ship the bulk of this change before we upgrade Apollo, simplifying the final gigantic PR it that we'll have for the upgrade.

It's possible I missed some (there was a lot of merging and cherry-picking to develop this PR) but I'm fairly sure I got the bulk of these changes.


stevestreza-ksr (approved):
Not 100% sure how this fits in yet but I'm sure that'll be visible in the final PR so 👍

amy-at-kickstarter:
> Not 100% sure how this fits in yet but I'm sure that'll be visible in the final PR so 👍

Oh, yeah, I should have put an example in the PR description! Basically we have to make a bunch of rote changes that look like
```swift
let query = GraphAPI.SomeQuery(
   foo: foo.isSome ? GraphQLNullable.some(foo) : GraphQLNullable.none
)
```

To simplify it, I'm adding a helper `.someOrNil` like
```swift
let query = GraphAPI.SomeQuery(
   foo: GraphQLNullable.someOrNil(foo)
)
```

This change adds `.someOrNil` and puts it in all the places we need, but since we're still on Apollo 0.44, the wrapper does nothing.

PR #2518, opened 2025-07-07
Simplify RemoteConfigFeatureFlagToolsViewController

amy-at-kickstarter:
# 📲 What

- Delete its snapshot test. This is not a user-facing screen, and the snapshot test mostly just annoyed us when we made pull requests. 
- Use `featureEnabled(feature:)` helper function directly, instead of implementing an extra switch statement.
- Delete unused `defaultValue` parameter in `featureEnabled(feature:`)

# 🤔 Why

Adding and removing feature flags is something we do several times every sprint. This PRs are slightly more complicated than they need to be, for such a simple change. 

This small changes will make that frequent task just a little bit shorter to write and ship.


Library/ViewModels/RemoteConfigFeatureFlagToolsViewModel.swift
```diff
@@ -32,8 +32,7 @@ public final class RemoteConfigFeatureFlagToolsViewModel: RemoteConfigFeatureFla
       self.viewDidLoadProperty.signal,
       didUpdateUserDefaultsAndUI
     )
-    .map { _ in AppEnvironment.current.remoteConfigClient?.allFeatures() }
-    .skipNil()
+    .map { _ in RemoteConfigFeature.allCases }
```
amy-at-kickstarter, line 35: `AppEnvironment.current.remoteConfigClient?.allFeatures()` returns `RemoteConfigFeature.allCases`, _except_ in Debug builds, when `remoteConfigClient` is `nil`. This should let us set flags for debug builds, finally!

Library/RemoteConfig/RemoteConfigFeature+Helpers.swift
```diff
@@ -12,7 +12,7 @@ private func featureEnabled(feature: RemoteConfigFeature, defaultValue: Bool = f
     return valueFromRemoteConfig
```
amy-at-kickstarter, line 12: `valueFromRemoteConfig` is only `nil` in debug builds, because `remoteConfigClient` is `nil`. It's never `nil` in Beta builds, because Firebase always returns something. This is why `defaultValue` had no effect.

It would be nice to fix this so we can set sensible defaults values in the beta, but that's work for another PR.

Library/ViewModels/RemoteConfigFeatureFlagToolsViewModel.swift
```diff
@@ -84,45 +83,7 @@ public final class RemoteConfigFeatureFlagToolsViewModel: RemoteConfigFeatureFla
...
-    return featureSearchFilterByAmountRaised()
-  }
-}
-
-/** Returns the value of the User Defaults key in the AppEnvironment.
- */
-private func getValueFromUserDefaults(for feature: RemoteConfigFeature) -> Bool? {
-  return AppEnvironment.current.userDefaults
-    .remoteConfigFeatureFlags[feature.rawValue]
+  return featureEnabled(feature: feature)
```
amy-at-kickstarter, line 86: Just removing a step of complexity, since each of those helper functions called `featureEnabled` anyways.

Kickstarter-iOS/Features/RemoteConfigFeatureFlagTools/Controller/__Snapshots__/RemoteConfigFeatureFlagToolsViewControllerTests/testRemoteConfigFeatureFlagToolsViewController.1.png
stevestreza-ksr, line 1: 🥳 🎆 🎉 🔥

PR #2519, opened 2025-07-07
MBL-2558: Add pills for filtering search by 'show only' toggles

amy-at-kickstarter:
# 📲 What

Add new pills for the "Show Only" toggles.
new toggles

# 🛠 How

* I added new `SearchFilterPill.FilterType` values to support this. They seem juuuust different enough from a dropdown to be worth it.
* I refactored `SearchFilterPill.FilterType` a little bit, to pass in a `UIImage` intead of an icon name. Otherwise, we would have an ambiguous enum type like `.toggle(String, String)`.
* Added a feature flag (such a short change now!) and an asset, too, in separate PRs.


Library/UseCases/SearchFilters/SearchFilterPill.swift
```diff
@@ -24,12 +24,18 @@ public struct SearchFilterPill: Identifiable {
...
+    case projectsWeLove
+    case saved
+    case following
+    case recommended
   }
 
   /// How the pill should be rendered.
   public enum ButtonType {
-    case image(String)
+    case image(UIImage)
```
amy-at-kickstarter, line 35: Made these associated types make slightly more sense.

Kickstarter-iOS/Features/SortFilter/HeaderView/PillButtons.swift
```diff
@@ -111,7 +111,6 @@ internal struct ImagePillButton: View {
     Button(action: self.action) {
       HStack {
         Image(uiImage: self.image)
-          .renderingMode(.template)
```
amy-at-kickstarter, line 114: Moving this out, for consistency

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -335,6 +335,14 @@ private func filterModal(toShowForPill pill: SearchFilterPill.FilterType) -> Sea
     modalType = .location
   case .amountRaised:
     modalType = .amountRaised
+  case .projectsWeLove:
```
amy-at-kickstarter, line 271: These will be toggles on the filter root page.

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderView.swift
```diff
@@ -70,12 +86,22 @@ let previewPills = [
   SearchFilterPill(
     isHighlighted: true,
     filterType: .sort,
-    buttonType: .image("icon-sort")
+    buttonType: .image(UIImage.checkmark)
```
ifosli, line 89: Nit: Just checking if this change is intentional? It looks unrelated but I don't mind it.

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderViewTests.swift
```diff
@@ -14,13 +14,17 @@ final class SearchFiltersHeaderViewTests: TestCase {
       ]
     ).forEach {
       device, contentSize in
-      let size = device.deviceSize(in: .portrait)
+      var size = device.deviceSize(in: .portrait)
+
+      let icon1 = Library.image(named: "shortcut-icon-k")!.withRenderingMode(.alwaysTemplate)
+      let icon2 = Library.image(named: "icon-sort")!.withRenderingMode(.alwaysTemplate)
+      let icon3 = Library.image(named: "icon-pwl")!
```
ifosli, line 21: Nit: Since these icons are logically matched to specific filter types, I think it makes more sense to name them that way (ex `shortcutIcon`, `sortIcon`, `pwlIcon`). But, since this is a test and it doesn't matter if the images align with the filter type, I don't feel strongly about this.

ifosli (approved):

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderView.swift
amy-at-kickstarter: Yup, intentional. I change this enum type from `.image(String)` to `.image(UIImage)`. I used a system image instead of one from our library because bundles get wonky with preview items.

Kickstarter-iOS/Features/SortFilter/HeaderView/SearchFiltersHeaderViewTests.swift
amy-at-kickstarter: Fair! These are mostly just here to exercise the UI in a generic way - I don't actually care what they look like, as long as they're consistent.

PR #2521, opened 2025-07-07
Use security origin to hopefully make survey webview decision policy support 3DS

ifosli:
# 📲 What

Use target frame security origin host to determine if navigation action is allowed. I think this will allow 3DS requests to show (I expect stripe to still be the security origin host) but I'm not sure.

Note: I didn't add a unit test for this behavior because I couldn't figure out how to test this. As far as I can tell, `WKFrameInfo` objects cannot be initialized with data/updated with test data.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2588)

# ✅ Acceptance criteria

- [ ] Nothing breaks

# ⏰ TODO

- I'm planning to follow up on this after it's submitted, either to get someone to test it in the beta app, or I'll wait until the next release and see if the metrics this bug was causing are looking better. If it ends up not fixing the issue after all, I'll revert these changes and probably recommend we get rid of the security policy instead.

### DIFF
--- a/Library/ViewModels/SurveyResponseViewModel.swift
+++ b/Library/ViewModels/SurveyResponseViewModel.swift
@@ -75,8 +75,9 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
       }
       .skipNil()
 
-    let newRequest = self.policyForNavigationActionProperty.signal.skipNil()
-      .map { action in action.request }
+    let newNavigationAction = self.policyForNavigationActionProperty.signal.skipNil()
+
+    let newRequest = newNavigationAction.map { action in action.request }
 
     let newSurveyRequest = newRequest
       .filter { request in
@@ -126,11 +127,13 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
           }
       }
 
-    self.policyDecisionProperty <~ newRequest
-      .map { request in
-        if isStripeRequest(request) {
+    self.policyDecisionProperty <~ newNavigationAction
+      .map { action in
+        if isStripeNavigationAction(action) {
           return true
         }
+
+        let request = action.request
 
         if !AppEnvironment.current.apiService.isPrepared(request: request) {
           return false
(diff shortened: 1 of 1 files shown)


Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -191,11 +194,23 @@ private func isSurvey(request: URLRequest) -> Bool {
   return true
 }
 
-// Returns true if the url host is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
-private func isStripeRequest(_ request: URLRequest) -> Bool {
+// Returns true if either the url host or the target frame's security origin
+// is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
+private func isStripeNavigationAction(_ actionData: WKNavigationActionData) -> Bool {
+  print(actionData.request.url ?? "no url")
```
scottkicks: This might be a leftover print statement.

scottkicks (approved):
Happy to help beta test this when ready!

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -191,11 +194,23 @@ private func isSurvey(request: URLRequest) -> Bool {
...
+  }
+  return false
+}
+
+private func isStripeHost(_ host: String) -> Bool {
   let stripeDomains = ["stripe.com", "stripe.network", "stripecdn.com"]
 
-  guard let host = request.url?.host?.lowercased() else { return false }
-  let withoutSubdomain = host.split(separator: ".").suffix(2).joined(separator: ".")
+  let withoutSubdomain = host.lowercased().split(separator: ".").suffix(2).joined(separator: ".")
```
amy-at-kickstarter, line 213: Regarding unit tests - is there one for this func `isStripeHost`? That's the one bit that seems worth validating to me. Just making a few positive and negative tests.

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -191,11 +194,23 @@ private func isSurvey(request: URLRequest) -> Bool {
...
-// Returns true if the url host is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
-private func isStripeRequest(_ request: URLRequest) -> Bool {
+// Returns true if either the url host or the target frame's security origin
+// is of the form *.stripe.com, *.stripe.network, or *.stripecdn.com.
+private func isStripeNavigationAction(_ actionData: WKNavigationActionData) -> Bool {
+  print(actionData.request.url ?? "no url")
+  if let host = actionData.request.url?.host, isStripeHost(host) {
+    return true
+  }
+  if let securityOriginHost = actionData.navigationAction.targetFrame?.securityOrigin.host,
```
amy-at-kickstarter, line 203: I don't know a ton about security origins. Can you run this by infra during office hours to double-check this is the correct test?

amy-at-kickstarter (changes requested):
One request for a specific unit test! And a suggestion to check with infra. Otherwise, let's try it out.

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: Yeah, we've got multiple tests for the stripe host already! They're the 5 last tests in `SurveyResponseViewModelTests` (at the bottom of the file if you don't consider helpers). I added these tests when I first allowlisted stripe as a host last year.

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: To be honest, I don't expect infra to know webviews well enough to tell us if this is the correct test (or what else we should be testing). But I can ask them. There's a reason this pr is experimental (as in I hope it works but I won't know until after it's merged and can be tested).

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: Whoops, thanks! Fixed

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: I did check with infra, and this is not something they know. So we'll try it, and see what it does 🤷

PR #2522, opened 2025-07-08
Refactor interface to SearchFiltersUseCase

amy-at-kickstarter:
# 📲 What

Refactor the interface to `SearchFiltersUseCase`, and clean up its `init` method.

# 🤔 Why

Now that we have _many_ filter types, all the plumbing required to set it up was getting very wordy. Since I'm about to add 4 more toggles (which means 4 more filters!), it was going to be too unwieldy.

# 🛠 How

* Instead of having individual methods `selectedFoo(value)`, have one method `selectedFilter` which takes a `SearchFilterEvent` called `foo`. So `selectedFoo` becomes `selectedFilter(.foo(value))`
* Add an extension to `MutableProperty` which simplifies the logic "give me this property's signal, but also its value when an initial signal is set".
* Create the `SearchFilters` output object in one line, because it took up too much screen space

### DIFF
--- a/Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
+++ b/Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
@@ -64,35 +64,14 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
         return modalType
       }
 
-    self.selectedSort = Signal.merge(
-      self.selectedSortProperty.producer.takeWhen(initialSignal),
-      self.selectedSortProperty.signal
-    )
-
-    self.selectedCategory = Signal.merge(
-      self.selectedCategoryProperty.producer.takeWhen(initialSignal),
-      self.selectedCategoryProperty.signal
-    )
-
-    self.selectedState = Signal.merge(
-      self.selectedStateProperty.producer.takeWhen(initialSignal),
-      self.selectedStateProperty.signal
-    )
-
-    self.selectedPercentRaisedBucket = Signal.merge(
-      self.selectedPercentRaisedBucketProperty.producer.takeWhen(initialSignal),
-      self.selectedPercentRaisedBucketProperty.signal
-    )
-
-    self.selectedLocation = Signal.merge(
-      self.selectedLocationProperty.producer.takeWhen(initialSignal),
-      self.selectedLocationProperty.signal
-    )
-
-    self.selectedAmountRaisedBucket = Signal.merge(
-      self.selectedAmountRaisedBucketProperty.producer.takeWhen(initialSignal),
-      self.selectedAmountRaisedBucketProperty.signal
-    )
+    self.selectedSort = self.selectedSortProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedCategory = self.selectedCategoryProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedState = self.selectedStateProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedPercentRaisedBucket = self.selectedPercentRaisedBucketProperty
+      .signal(takeInitialValueWhen: initialSignal)
+    self.selectedLocation = self.selectedLocationProperty.signal(takeInitialValueWhen: initialSignal)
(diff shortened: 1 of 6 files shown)


Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -72,71 +64,38 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
-
-    let locationOptions = SearchFilters.LocationOptions(
-      defaultLocations: self.defaultLocationsProperty.value,
-      suggestedLocations: self.suggestedLocationsProperty.value
-    )
-
-    let amountRaisedOptions = SearchFilters.AmountRaisedOptions(
-      buckets: DiscoveryParams.AmountRaisedBucket.allCases
-    )
+    self.selectedSort = self.selectedSortProperty.signal(takeInitialValueWhen: initialSignal)
```
amy-at-kickstarter, line 67: If this would be useful elsewhere, I could make this helper public; but I think we rarely use `MutableProperty` like I'm using it here. This is a much more 'stateful' piece of code than the rest of our app.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -13,18 +13,8 @@ public protocol SearchFiltersUseCaseInputs {
   func tappedButton(forFilterType: SearchFilterPill.FilterType)
   /// Call this when the clears their query and the sort options should reset.
   func clearedQueryText()
-  /// Call this when the user selects a new sort option.
```
amy-at-kickstarter, line 16: 👋 So much simpler!

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -72,71 +64,38 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
+    self.selectedSort = self.selectedSortProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedCategory = self.selectedCategoryProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedState = self.selectedStateProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedPercentRaisedBucket = self.selectedPercentRaisedBucketProperty
+      .signal(takeInitialValueWhen: initialSignal)
+    self.selectedLocation = self.selectedLocationProperty.signal(takeInitialValueWhen: initialSignal)
+    self.selectedAmountRaisedBucket = self.selectedAmountRaisedBucketProperty
+      .signal(takeInitialValueWhen: initialSignal)
 
     self.searchFilters = SearchFilters(
```
amy-at-kickstarter, line 76: This is all the same, I just got rid of the intermediary variables. I feel like reading this as one big block is actually simpler than skimming all the sub-parts.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -270,47 +229,21 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
-      let categories = self.categoriesProperty.value
-      let subcategories = categories.lazy.flatMap { $0.subcategories?.nodes ?? [] }
-      let exists = categories.contains(category) || subcategories.contains(category)
-      if !exists {
-        assert(false, "Selected category should be one of the categories set in SearchFiltersUseCase.")
-      }
+  public func selectedFilter(_ event: SearchFilterEvent) {
+    switch event {
+    case let .sort(sort):
+      self.selectedSortProperty.value = sort
```
amy-at-kickstarter, line 235: I deleted the internal asserts from the methods `selectedCategory` and `selectedSortOption`. In retrospect, I don't feel like they're adding much value.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
jovaniks: Totally! So much simpler and sleeker 💅

ifosli (approved):

PR #2523, opened 2025-07-08
MBL-2560: Update SearchQuery and DiscoveryParams for search filter toggles

amy-at-kickstarter:
# 📲 What

Update `SearchQuery` to support the toggle booleans, and update the `DiscoveryParams` -> `SearchQuery` adapter code.

### DIFF
--- a/Library/ViewModels/SearchViewModel+GraphQL.swift
+++ b/Library/ViewModels/SearchViewModel+GraphQL.swift
@@ -87,10 +87,33 @@ extension GraphAPI.SearchQuery {
     } else {
       categoryId = nil
     }
+
     let state = GraphAPI.PublicProjectState.from(discovery: params.state ?? .all)
     let raised = GraphAPI.RaisedBuckets.from(discovery: params.percentRaised)
     let locationId = params.location?.graphID
     let pledged = GraphAPI.PledgedBuckets.from(discovery: params.amountRaised)
+
+    let showRecommended = params.social
+    let showSavedProjects = params.starred
+    let showProjectsWeLove = params.staffPicks
+    let showFollowing = params.social
+
+    assert(
+      showRecommended != .some(false),
+      "Setting showRecommended to false would return only non-recommended results. Did you mean to set it to nil?"
+    )
+    assert(
+      showSavedProjects != .some(false),
+      "Setting showSavedProjects to false would return only unsaved results. Did you mean to set it to nil?"
+    )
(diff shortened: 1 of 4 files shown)


Library/ViewModels/SearchViewModel+GraphQL.swift
```diff
@@ -87,11 +87,34 @@ extension GraphAPI.SearchQuery {
     } else {
       categoryId = nil
     }
+
     let state = GraphAPI.PublicProjectState.from(discovery: params.state ?? .all)
     let raised = GraphAPI.RaisedBuckets.from(discovery: params.percentRaised)
     let locationId = params.location?.graphID
     let pledged = GraphAPI.PledgedBuckets.from(discovery: params.amountRaised)
 
+    let showRecommended = params.social
```
amy-at-kickstarter, line 96: I've made the naming reflect how the toggles will be labeled, instead of what they're named in the API.

Library/ViewModels/SearchViewModel+GraphQL.swift
```diff
@@ -87,11 +87,34 @@ extension GraphAPI.SearchQuery {
...
+    let showSavedProjects = params.starred
+    let showProjectsWeLove = params.staffPicks
+    let showFollowing = params.social
```
amy-at-kickstarter, line 99: By using the existing `DiscoveryParams` options, we'll get some amount of analytics logging for free.

Library/ViewModels/SearchViewModel+GraphQL.swift
scottkicks: Maybe it's worth discussing with backend how we can align on the same naming?

Other names aren't exact though I guess. Like starred vs saved projects and staffPicks vs projects we love.

amy-at-kickstarter: Yeah, if I had to take a guess, I'm assuming that many of these properties were named when the features themselves had different names. So it's probably that the user-facing names changed over time.

Would you find it clearer if I left everything as its 'backend' name, and just had the correct labels in the UI? 🤔

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #2525, opened 2025-07-08
[MBL-2595] Give cards identifiers that won't change on reload

ifosli:
# 📲 What

Give the PPO cards an identifier based on the project id, tier type, and card actions. This means that on a refresh where the cards don't change (for example, if a user opened a webview and closed it again immediately to check what project they tapped on), swiftUI knows that the cards didn't change, and the user keeps their scroll position.

Note: This isn't a perfect solution; if a card is removed (ex because a survey is filled out or a payment is fixed), the user jumps back to the top of the list. In the future, where these cards would be replaced by "reward funded" cards, the card ids would change, and the user would also be jumped to the top of the list. We should revisit this solution once we implement v2 of PPO - that's when we expect users to have enough cards to deal with that they really don't want to lose their spot.

# 🤔 Why

To me, basing the id on the project id, tier type, and actions feels safe. I can't imagine a future where we won't want that to be a unique identifier. (Changing any other fields just means the card updates smoothly in place to show new info).

I considered making the identifier **equal** to the project id. This would mean that ex if a card changes from an "survey" card to a future "reward delivered?" card, swiftui would know that it's the same card (though it's probably moved in the list). I think just using the project id might be the correct solution in the long run, but I don't feel 100% certain that we'll never have multiple cards for the same project/backing at some point in the future, and we don't currently have support for a general "funded project" card, so this would be hard to test.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2595)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro Max - 2025-07-08 at 15 53 14 | Simulator Screen Recording - iPhone 16 Pro Max - 2025-07-08 at 15 51 24 |
| Fake names change only (after): | Fake names and fake ids change (after): |
| Simulator Screen Recording - iPhone 16 Plus - 2025-07-08 at 15 18 12 | Simulator Screen Recording - iPhone 16 Plus - 2025-07-08 at 15 15 45 |

# ✅ Acceptance criteria

- [x] Cards load/list scrolls as smoothly
- [x] Interacting with card no longer jumps to top of list (when cards are unchanged)


Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -35,7 +35,11 @@ public struct PPOProjectCardModel: Identifiable, Equatable, Hashable {
 
   // MARK: - Identifiable
 
-  public let id = UUID()
```
ifosli, line 38: This line caused the card to get a unique identifier even when the exact same info was fetched from the server, which told swiftUI to reload our list.

Library/PaginatingList.swift
```diff
@@ -48,7 +48,7 @@ public struct PaginatingList<Data, Cell, Header>: View where
         .listRowBackground(Color.clear)
       ForEach(self.data) { item in
         self.content(item)
-          .tag(item)
```
stevestreza-ksr: While we don't use it for PPO, please keep the `tag(item)` for selection support in PaginatingList. It should be able to exist alongside `id(item.id)`.

stevestreza-ksr (changes requested):
Looks good aside from one change, will 👍 when it is fixed!

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
```diff
@@ -35,7 +35,11 @@ public struct PPOProjectCardModel: Identifiable, Equatable, Hashable {
...
+  // Create the card's id from the project id, tier type, and actions.
+  // If any other fields change, the card should be considered the same card, just modified.
+  public var id: String {
+    "\(self.projectId)-\(self.tierType)-\(self.actions)"
```
amy-at-kickstarter: `self.actions` prints kind of weirdly - for example, I got
```"293202887-openSurvey-(Kickstarter_Framework.PPOProjectCardModel.Action.completeSurvey, nil)"```
for a card on PPOTest7.

You could throw in a map to format the `actions` in a more readable way, but it should work correctly even if you don't!

amy-at-kickstarter (approved):

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
ifosli: Actions are a tuple, so I don't think `map` is the way to go, but I updated this to use the actions' ids instead. (Ex: "1370106883-fixPayment-Fix Payment red-" for a card with one fix payment action where the second action is nil). I wasn't planning to make this readable initially, but it's a good idea. I do think the other option for making it readable is to not include the actions in the id; I don't _think_ we'll ever have multiple cards with the same project id and tier type, anyways.

Library/PaginatingList.swift
ifosli: Ah, I'd misread the documentation here. Thanks! (Tags get created automatically, but only if the `id` is the same type as the `item` and I'd missed the second part.) I just switched this back to `tag` - the `id` of the item looks like it gets used automatically.

Kickstarter-iOS/Features/PledgedProjectsOverview/CardView/PPOProjectCardModel.swift
amy-at-kickstarter: I like that solution! Works for me.

stevestreza-ksr (approved):

PR #2526, opened 2025-07-09
[MBL-2419] Adds UseCase For Onboarding Flow

scottkicks:
# 📲 What

Adds a new `OnboardingUserCase` that:
- Outputs a list of `OnboardingItem`s.
- Determines if user permisions have been granted or need to be presented for Push Notifications and App Tracking Transparecy.
- Emits onboarding analytic events.
- Initiating the login/signup flow.

### [Onboarding Designs Here](https://www.figma.com/design/cCLSxffRzm3jCVkQd0F4fw/Native-Onboarding-Flow?node-id=45-4755&m=dev)

# 🛠 How

- Adds new `OnboardingUseCase` + `OnboardingUseCaseTests`
- Adds a new LoginIntentType since we might be launching the login flow from Onboarding.

# ✅ Acceptance criteria

- [x] New UseCaseTests pass

# ⏰ NEXT UP

- [ ] Build Onboarding UI and connect it to this new UseCase
- [ ] Add activity tracking lottie files + new Background image asset


Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+        let canRequestAppTrackingTransparency = appTrackingTransparency
+          .advertisingIdentifier == nil && appTrackingTransparency.shouldRequestAuthorizationStatus()
+
+        if canRequestAppTrackingTransparency {
+          appTrackingTransparency.requestAndSetAuthorizationStatus()
+        }
+      }
+
+    _ = self.goToNextItemTappedSignal.signal
+      .map { itemType in
```
jovaniks: Since the purpose of this is to track events and not return a value, I believe the better method to use here would be `observeValues` instead.

jovaniks (approved):
LGTM! Just one nit - consider using `observeValues` instead of map since this is only tracking events and doesn’t need to return a value.

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+      return true
+    default:
+      return false
+    }
+  }
+}
+
+public struct OnboardingItem: Identifiable, Equatable {
+  public var id: UUID = .init()
+  var title: String
```
amy-at-kickstarter: Nit: should these bet `let` instead of `var`?

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+  var subTitle: String
+  var lottieView: LottieAnimationView = .init()
+  var type: OnboardingItemType
+
+  // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
+  public static let allItems: [OnboardingItem] = [
+    .init(
+      title: "Welcome to Kickstarter",
```
amy-at-kickstarter: For search, I started copying Android and adding "FPO:" ("for position only") in my hardcoded strings. Just to make it clear in the UI that they're temporary.

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+      subTitle: """
+      Use our app to discover and support creative projects. Browse by category, find projects near you, or explore our “Projects We Love” picks.
+      """,
+      lottieView: .init(name: localizedOnboardingLottieFile(for: .welcome, in: .main) ?? "", bundle: .main),
```
amy-at-kickstarter: It would make more sense to have a `guard` statement here, instead of this placeholder string empty string - since if the Lottie file is missing, the onboarding item would just be broken.

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+      lottieView: .init(
+        name: localizedOnboardingLottieFile(for: .enableNotifications, in: .main) ?? "",
+        bundle: .main
+      ),
+      type: .enableNotifications
+    ),
+    .init(
+      title: "Personalize your experince",
+      subTitle: """
+      Allow tracking to help us improve your
```
amy-at-kickstarter: Weird line breaks here but I know it's a temp string.

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+  * `goToNextItemTapped(item: OnboardingItemType)` -  Moves to the next onboarding flow item.
+
+ UI Outputs:
+  * `onboardingItems` - Returns a list of `OnboardingItem` used to populate the views for each section of the flow.
+  * `goToLoginSignup()` - The user tapped the login button. Triggers `goToLoginSignupTapped`.
+
+ Data Outputs:
+  * `completedGetNotifiedRequest` - The user has completed interacting with the notifications permission system dialog.
+  * `completedAllowTrackingRequest` - The user has completed interacting with the AppTrackingTransparency  permission system dialog.
+  */
```
amy-at-kickstarter, line 95: Love these comments 👍

Library/OnboardingUseCase.swift
```diff
@@ -0,0 +1,210 @@
...
+      .map { _ in () }
+
+    self.completedAllowTrackingRequest = self.allowTrackingTappedSignal.signal
+      .map { _ in
+        let appTrackingTransparency = AppEnvironment.current.appTrackingTransparency
+        let canRequestAppTrackingTransparency = appTrackingTransparency
+          .advertisingIdentifier == nil && appTrackingTransparency.shouldRequestAuthorizationStatus()
+
+        if canRequestAppTrackingTransparency {
+          appTrackingTransparency.requestAndSetAuthorizationStatus()
```
amy-at-kickstarter: `map` isn't intended to be used for side effects - calling `requestAndSetAuthorizationStatus` will actually trigger the popup.

Instead, you should have an output signal (like `triggerAppTrackingTransparencyPopup`) which your onboarding view controller subscribes to and calls `appTrackingTransparency.requestAndSetAuthorizationStatus()`.

amy-at-kickstarter (changes requested):
One requested change - move the code that triggers the app tracking transparency popup out of the call to `Signal.map`.

The rest are just nits!

amy-at-kickstarter (approved):

PR #2530, opened 2025-07-10
MBL-2560: Hook up 'show only' search toggles

amy-at-kickstarter:
# 📲 What

Hook up the `show only` search toggles.
* Make the toggles toggle when you tap on them in the header view
* Send the toggle values to the search query, so they take effect
* Add some unit tests

toggles-sorta-working

### DIFF
--- a/Library/ViewModels/SearchViewModel+GraphQL.swift
+++ b/Library/ViewModels/SearchViewModel+GraphQL.swift
@@ -144,16 +144,12 @@ extension DiscoveryParams {
     params.percentRaised = percentRaised
     params.location = location
     params.amountRaised = amountRaised
-    params.recommended = trueOrNil(toggles.recommended)
-    params.starred = trueOrNil(toggles.savedProjects)
-    params.staffPicks = trueOrNil(toggles.projectsWeLove)
-    params.social = trueOrNil(toggles.following)
+    params.recommended = toggles.recommended
+    params.starred = toggles.savedProjects
+    params.staffPicks = toggles.projectsWeLove
+    params.social = toggles.following
     return params
   }
-}
-
-func trueOrNil(_ value: Bool) -> Bool? {
-  value ? true : nil
 }
 
 extension GraphAPI.SearchQuery.Data.Project.Node: @retroactive Equatable {}


Library/UseCases/SearchFilters/SearchFilters.swift
```diff
@@ -232,6 +251,17 @@ public class SearchFilters: ObservableObject {
       )
     }
 
+    if featureSearchFilterByShowOnlyToggles() && AppEnvironment.current.currentUser.isSome {
```
amy-at-kickstarter, line 254: Rearranged the pills a bit. I checked the designs, and the toggles aren't all in one group!

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -41,6 +41,8 @@ public protocol SearchFiltersUseCaseDataOutputs {
   var selectedLocation: Signal<Location?, Never> { get }
   /// The currently selected amount raised bucket. Defaults to `nil`. Default value only sent after `initialSignal` occurs.
   var selectedAmountRaisedBucket: Signal<DiscoveryParams.AmountRaisedBucket?, Never> { get }
+  /// The currently selected 'Show Only' toggles. All toggles default to `false`. Default value only sent after `initialSignal` occurs.
+  var selectedToggles: Signal<SearchFilterToggles, Never> { get }
```
amy-at-kickstarter, line 45: Consolidated these into a single `dataOutput` for two reasons:

1. They eventually get stuffed in a `combineLatest` in `SearchViewModel`, and I was concerned I was going to swap two anonymous booleans if I did that.
2. The `combineLatest` in `SearchViewModel` actually ran out of space - apparently it can only combine 10 signals!

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -95,28 +106,50 @@ public final class SearchFiltersUseCase: SearchFiltersUseCaseType, SearchFilters
...
+    self.showFilters = tappedModalPill
+      .map { pill in
+        let modalType = filterModal(toShowForPill: pill)
+        return modalType
+      }
+
+    tappedTogglePill
+      .observeValues { [weak self] pill in
+        self?.toggleFilter(ofType: pill)
+      }
```
amy-at-kickstarter, line 130: As noted in a few places in comments, these toggles don't actually show the filter page. They just toggle the value when you tap the pill. So some changes were needed in this plumbing.

Library/ViewModels/SearchViewModel+GraphQL.swift
```diff
@@ -93,28 +93,11 @@ extension GraphAPI.SearchQuery {
     let locationId = params.location?.graphID
     let pledged = GraphAPI.PledgedBuckets.from(discovery: params.amountRaised)
 
-    let showRecommended = params.social
+    let showRecommended = params.recommended
     let showSavedProjects = params.starred
     let showProjectsWeLove = params.staffPicks
     let showFollowing = params.social
 
-    assert(
```
amy-at-kickstarter, line 101: Moved this nullability logic elsewhere, cleaned up these asserts.

Library/ViewModels/SearchViewModel+GraphQL.swift
```diff
@@ -93,28 +93,11 @@ extension GraphAPI.SearchQuery {
     let locationId = params.location?.graphID
     let pledged = GraphAPI.PledgedBuckets.from(discovery: params.amountRaised)
 
-    let showRecommended = params.social
+    let showRecommended = params.recommended
```
amy-at-kickstarter, line 96: Whoops!

Library/ViewModels/SearchViewModel+GraphQL.swift
```diff
@@ -160,10 +144,18 @@ extension DiscoveryParams {
...
     params.amountRaised = amountRaised
+    params.recommended = trueOrNil(toggles.recommended)
+    params.starred = trueOrNil(toggles.savedProjects)
+    params.staffPicks = trueOrNil(toggles.projectsWeLove)
+    params.social = trueOrNil(toggles.following)
     return params
   }
 }
 
+func trueOrNil(_ value: Bool) -> Bool? {
```
amy-at-kickstarter: Truth or Dare for Swift

Kickstarter-iOS/Features/SortFilter/FilterRootView/FilterRootView.swift
```diff
@@ -12,48 +12,6 @@ struct FilterRootView: View {
   var onResults: (() -> Void)? = nil
   var onClose: (() -> Void)? = nil
 
-  private var selectedCategory: Binding<SearchFiltersCategory> {
```
amy-at-kickstarter, line 15: Moved all these to the bottom of the file in an extension. This list is getting long!

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Kickstarter-iOS/Features/SortFilter/FilterRootView/FilterRootView.swift
```diff
@@ -310,6 +265,94 @@ extension View {
   }
 }
 
+// MARK: - Bindings
+
+extension FilterRootView {
+  private var selectedCategory: Binding<SearchFiltersCategory> {
```
amy-at-kickstarter, line 271: I experimented with writing a clever little helper function, with keypaths and generics, to make these bindings shorter. But I decided it was _too_ clever, and anyone debugging it would curse my name. 

In retrospect, @stevestreza-ksr was right, and I don't love `@Binding` that much for passing data into SwiftUI. However, for control elements, it's a necessary evil - I found having a shadow `@State` which was monitored in some way even _buggier_.

I'm feeling a lot more in tune with SwiftUI after all this work, and I've gained a healthy skepticism of its state management mechanisms.

amy-at-kickstarter: If folks are interested I'd be happy to talk about this in a squad chat.

Library/UseCases/SearchFilters/SearchFilterPill.swift
```diff
@@ -28,6 +28,23 @@ public struct SearchFilterPill: Identifiable {
     case saved
     case following
     case recommended
+
+    /// If `true`, this pill is a boolean value that can be toggled directly from the search header.
+    /// Tapping on this pill doesn't open a modal.
+    var isToggle: Bool {
+      switch self {
+      case .projectsWeLove:
```
ifosli, line 36: nit: I think this might actually be easier to read if you combined all the show only toggles into one case; that way it's super clear that they all behave the same, and it's less lines of code. But up to you; I'm also fine with the way this is now.

Library/UseCases/SearchFilters/SearchFilters.swift
```diff
@@ -125,7 +139,8 @@ public class SearchFilters: ObservableObject {
     projectState: DiscoveryParams.State,
     percentRaisedBucket: DiscoveryParams.PercentRaisedBucket?,
     location: Location?,
-    amountRaisedBucket: DiscoveryParams.AmountRaisedBucket?
+    amountRaisedBucket: DiscoveryParams.AmountRaisedBucket?,
+    toggles: SearchFilterToggles
```
ifosli, line 143: This might be outside the scope of this pr, but I'm confused by the naming here. What's the difference between `SearchFilterToggles` and `ShowOnlyOptions`? If they're basically the same thing, could we just define one of these classes and use it in both places?

Library/ViewModels/SearchViewModel+GraphQL.swift
ifosli: Just checking: are you sure we want to treat `nil` as `true` instead of treating `nil` as `false`? I feel like the latter is more common, and what I would expect.

ifosli (approved):
A few questions/nits, but looks good overall!

scottkicks (approved):
LGTM 👍 
Thanks for the extra clarifying comments

Library/ViewModels/SearchViewModel+GraphQL.swift
amy-at-kickstarter: Whoops, nope, that's the opposite of what I want! I checked with @santiDotIO and apparently the backend treats `nil` and `false` the same, so I'm just going to clean up this helper and set `false` directly.

Library/UseCases/SearchFilters/SearchFilters.swift
amy-at-kickstarter: Yeah, no difference. I actually did originally make them the same object, but it felt a bit weird aesthetically to have everything in `SearchFilterOptions` be an internal `struct` except for this one specific case. So this was me being intentionally a little wordy, in the interest of preserving consistency.

Library/UseCases/SearchFilters/SearchFilterPill.swift
amy-at-kickstarter: Hmm, yeah, good point. The toggles are a bit clunky with how they worked out!

PR #2532, opened 2025-07-10
[MBL-2597] Use survey webview

ifosli:
# 📲 What

Replace all instances of `PledgeManagementDetailsWebViewController` with `SurveyResponseViewController` instead. The latter is the webview we've had for a while, with deep link support for things like opening the project page and a built-in login redirect. The former looks like it was added accidentally during crunch.

The only user-visible change is that the webview title disappears. We discussed this in the pledge-redemption-mobile-squad on slack. I'll put up a separate PR to add a title to the survey webview during the pledge redemption flow only.

# ✅ Acceptance criteria

- [ ] Buttons that used to open pledge management or backing details in a survey webview still do

# ⏰ TODO

- Add a title to the survey webview when a user is in the actual pledge redemption flow (will come in a separate pr)


Library/ViewModels/MessagesViewModel.swift
```diff
@@ -43,7 +43,7 @@ public protocol MessagesViewModelOutputs {
   var goToBacking: Signal<ManagePledgeViewParamConfigData, Never> { get }
 
   /// Emits when we should go to the Pledge Management View Pledge webview.
-  var goToPledgeManagementViewPledge: Signal<URL, Never> { get }
+  var goToPledgeManagementViewPledge: Signal<String, Never> { get }
```
amy-at-kickstarter, line 46: Nit: Does `SurveyResponseViewController` expect a `String` instead of a `URL`? I prefer the clearer typing of having all these outputs be a URL in the first place; it would be nice to keep that.

amy-at-kickstarter (approved):
LGTM

Library/ViewModels/MessagesViewModel.swift
ifosli: Yeah, `SurveyResponseViewController` does expect a `String`. I kind of like only having to actually create the url once (in the `SurveyResponseViewModel`), since the `URL` initializer can return nil - it makes the code a little simpler everywhere else. But I see what you mean about clearer typing. I'm tempted to leave this the way it is right now, but if you feel strongly about this I can update the VC to take in `URL`s instead (and just update the other call sites to create the `URL` right before presenting the web view).

PR #2382, opened 2025-04-14
[MBL-2316] Replace Native View Pledge Screen with WebView for PM Backers

jovaniks:
# 📲 What

Replaces the View Pledge native screen with a web view for backers who completed their pledge via Pledge Management (PM), whether as original backers or net new backers.

# 🛠 How

- Added logic to rely on `backing.order` to detect PM-completed pledges and switch to web view.
- Swapped the endpoint from `backing/details` to `backing/survey_responses`, which avoids the login issue and still renders the backing details.
- The backing URL is retrieved from the GraphQL `backingDetailsPageRoute` field to decouple from hardcoded paths (this matches the PPO approach).
- Applied this logic both in `ProjectPageViewController` and `MessagesViewController`.
- Ensured creators continue using V1 fetch for their flow.

# 👀 See

- [Ticket](https://kickstarter.atlassian.net/browse/MBL-2316)

| Project Page | Messages |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Pro - 2025-04-16 at 07 52 37 | Simulator Screen Recording - iPhone 16 Pro - 2025-04-16 at 07 53 34 |

# ✅ Acceptance criteria

- [x] PM backers open backing in webview
- [x] Creators still open View Pledge natively
- [x] No login prompt shown in webview
- [x] Works from both Messages and Project Page

PR #2380, opened 2025-04-10
[MBL-2317] View Your Pledge Web View

scottkicks:
# 📲 What

Creates a new web view controller that will display a user's backing details.

# 🤔 Why

For backers who pledged to a campaign and completed PM or who just completed PM, we'll show a web view when they want to see their backing details. 

This is a crunch time approach. There are plans to update our native screen with PM backing details, etc., once we're out of crunch time 🤞.

# 🛠 How

- Creates a new ViewController that inherits from our WebViewController class.
- Load the existing backing details URL on viewDidLoad using a passed in `Project`.

This is just the view controller. The next PR will wire it up and use the new `Backing.Order` object, and the feature flag, to determine when to show this web view and when to show the existing native screen.

The entry points will include the Project Page and Messages.

**[Checkout the SPIKE for more details](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/3627319298/Net+New+Backers+Crunch+Time+SPIKE)**

# 👀 See

I replaced the creator profile webview with this new one for testing purposes.

|  |  |
| --- | --- |
| Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-04-10 at 09 43 52 |  |

# ✅ Acceptance criteria

- [x] when presented this new web view controller loads the backing details page.

# ⏰ TODO

- [ ] wire this up using the A/C I mentioned above.

PR #2534, opened 2025-07-14
[MBL-2597][b] Show title in the survey webview in the pledge management flow

ifosli:
# 📲 What

Show a title in the survey webview during the pledge redemption flow, since this is something we can recognize by url. 

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2597)

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Recording - iPhone 16 Plus - 2025-07-14 at 10 32 29 | Simulator Screen Recording - iPhone 16 Plus - 2025-07-14 at 10 30 18 |

# ✅ Acceptance criteria

- [x] The survey/pledge management webview shows a title during the pledge management flow
- [x] The survey/pledge management webview has no title when viewing backing details/messages/survey/pledge redemption tabs.


Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -79,6 +83,25 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
...
+        if isSurvey(request: request) {
+          // Only update the title based on the main survey url.
+          return request.url
+        }
+        return nil
+      }
+      .map { (surveyUrl: URL) -> String? in
+        // The pledge management flow has its own url, so show a title for this url.
+        // The other urls shown in this dashboard are not unique (navigating between the pages
+        // they load doesn't change the url), so show no title instead.
```
scottkicks: nit: the way this comment line is broken up is a little odd. maybe that's a swift formatter issue.

scottkicks (approved):
Nice! This web view looks much cleaner with a title.

Just called out a formatting nit.

Library/ViewModels/SurveyResponseViewModel.swift
```diff
@@ -79,6 +83,25 @@ public final class SurveyResponseViewModel: SurveyResponseViewModelType {
...
+        if surveyUrl.lastPathComponent == "redeem" {
```
jovaniks, line 98: On the **"PM View your pledge"** flow, the title was `Strings.Backing_details()` when navigating to **View Your Pledge** (`backing.backingDetailsPageRoute`) with the route **/survey_responses**.
With the current changes, this route no longer shows a title.

I know this route is also used in other flows, so removing the title might make sense - just wanted to flag it in case we want to revisit that logic.

[Original PR reference](https://github.com/kickstarter/ios-oss/pull/2380/files#diff-6b4c6d27f20e04a1b40ccb521317afff1a43076c3fb69cd6626fb96ccd59b580R18)
[View your pledge - PR](https://github.com/kickstarter/ios-oss/pull/2382)

jovaniks (approved):
LGTM! ✅ Just a quick nit - confirming we're okay with dropping the `Strings.Backing_details()` title in the **View Your Pledge** flow (`/survey_responses`).

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: No, the way it's formatted is on me. I kept the first sentence on its own line because it felt like a separate thought. The second sentence is too long to fit on one line, so I did a line break close to and before 100 characters. I'll change it to move `the pages` to the next line so that the lines are a more consistent length, unless you'd prefer something else?

Library/ViewModels/SurveyResponseViewModel.swift
ifosli: Yeah, that's intentional. It doesn't make sense to have a title in the survey webview when viewing an individual tab since the different tabs should have different titles (but we can't tell which one the user is accessing from the client). Specifically, we can know what url the webview is opened from, but we cannot tell if they then switch tabs. The "Backing Details" title would've needed to apply for the backing details tab, the survey responses tab, the pledge management tab, and the messages tab, but for most of these, it's a misleading title.See https://kickstarter.slack.com/archives/C073CT2HGAH/p1752100432706729 for more details on this discussion.

PR #2540, opened 2025-07-16
[MBL-2420] Onboarding UI + ViewModel

scottkicks:
# 📲 What

Adds new Onboarding Views and an `OnboardingViewModel` with test data.

This is just the initial plumbing. Wiring up the system level permission dialogs and triggering this flow at the right time will be done separately. 

# 🛠 How

#### UI
- The main parent component here is `OnboardingView`. 
    - It's a SwiftUI view that displays and animates each new Lottie file, a progress view at the top of the screen, and each onboarding section into view when tapping each CTA.

- Within this parent view there are 2 main subviews:
   - `OnboardingItemView`: manages all of the content for each section (title, subtitle, lottie animation, and CTAs)
   - `CallToActionView`: manages the buttons at the bottom of each section.

- We want the onboarding flow to match [the designs](https://www.figma.com/design/cCLSxffRzm3jCVkQd0F4fw/Native-Onboarding-Flow?node-id=266-4207&m=dev) regardless of whether Dark or Light mode is enabled.

#### View Model
- A simple `OnboardingViewModel`, with tests, for now. More changes/refactoring might occur when I get to that piece of the puzzel.

# 👀 See

I'm having some trouble with the UI where the CTAs are pushed down when there's more than 3 lines of subtitle. 
I haven't found a solution that looks good.
I've tried a scrollview approach and allowing the font size to change dynamically.

If anyone has another suggestion, that'd be helpful!

| Before 🐛 | After 🦋 |
| --- | --- |
|  | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-07-16 at 15 10 04 |

# ✅ Acceptance criteria

- [x] Onboarding UI matches designs
- [x] ViewModel has passing and effective unit tests

# ⏰ TODO

- [ ] Launch onboarding at the correct time
- [ ] Trigger permissions dialog
- [ ] Trigger Login flow


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Library/OnboardingUseCase.swift
scottkicks, line 1: I removed the PN permissions output from this Use Case. The view model can responsible for triggering the popup. It didn't make much sense to me for the input signal to go through this extra layer without the use case being needed for any specific logic.

Kickstarter-iOS/Features/Onboarding/OnboardingStyles.swift
```diff
@@ -0,0 +1,16 @@
...
+import SwiftUI
+
+enum OnboardingStyles {
+  static let title = UIFont.ksr_title2().bolded
+  static let subtitle = UIFont.ksr_bodyLG()
+
+  static let ctaFont = UIFont.ksr_bodyLG().bolded
+
+  static let backgroundColor = Color(.hex(0x06E584))
+  static let progressBarTintColor = Color(.hex(0x00743D))
```
ifosli, line 12: Can we add the colors here to our design system if we don't already have comparable colors? Or are we not planning to have colors change in light vs dark mode?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -0,0 +1,67 @@
...
+
+struct CallToActionView: View {
+  let item: OnboardingItem
+  let animationDuration: CGFloat
+  let onPrimaryTap: () -> Void
+  let onSecondaryTap: () -> Void
+  let onLoginSignup: () -> Void
+
+  var body: some View {
+    // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
```
ifosli, line 18: Can you add "FPO" to each of the strings below, so that it's easy to see that these are not the translated versions?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -0,0 +1,67 @@
...
+
+  @ViewBuilder
+  private func primaryButton(title: String, action: @escaping () -> Void) -> some View {
+    Button(action: action) {
+      Text(title)
+        .font(Font(OnboardingStyles.ctaFont))
+        .foregroundColor(.white)
+        .frame(maxWidth: .infinity)
+        .padding()
+        .background(.black)
```
ifosli: Nit: put the button colors in the style file, too? Also, same thing applies here on if we want these colors to change based on light vs dark mode

Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
```diff
@@ -0,0 +1,59 @@
...
+  let onPrimaryTap: () -> Void
+  let onSecondaryTap: () -> Void
+  let onLoginSignup: () -> Void
+
+  var body: some View {
+    VStack {
+      VStack(spacing: Constants.titleSubtitleSpacing) {
+        Text(self.item.title)
+          .font(Font(OnboardingStyles.title))
+          .lineLimit(1)
```
ifosli: Are you sure we want a line limit on the title that we provide? It seems like it'd be more helpful to allow the title to expand over multiple lines

Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
```diff
@@ -0,0 +1,59 @@
...
+          .multilineTextAlignment(.center)
+          .frame(maxWidth: .infinity)
+
+        Text(self.item.subtitle)
+          .font(Font(OnboardingStyles.subtitle))
+          .lineLimit(4)
```
ifosli, line 36: Same here, re line limit

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -0,0 +1,115 @@
...
+  static let verticalSpacing: CGFloat = 24
+}
+
+struct OnboardingView: View {
+  @ObservedObject var viewModel: OnboardingViewModel
+  @State private var currentIndex: Int = 0
+  @Namespace private var animation
+
+  private var progress: Double {
+    Double(self.currentIndex + 1) / Double(self.viewModel.onboardingItems.count)
```
ifosli, line 25: Nit: Is it ever possible for the number of onboarding items to be zero? If so, we should add custom logic for that instead of dividing by zero.

Kickstarter-iOS/Features/Onboarding/Views/ResizableLottieView.swift
```diff
@@ -0,0 +1,23 @@
...
+    let animationView = self.onboardingItem.lottieView
+    animationView.contentMode = .scaleAspectFit
+    animationView.loopMode = .loop
+    return animationView
+  }
+
+  func updateUIView(_ uiView: LottieAnimationView, context _: Context) {
+    if self.isVisible && !uiView.isAnimationPlaying {
+      uiView.play()
+    } else if !self.isVisible {
```
ifosli, line 19: Nit: Is there a reason you're checking `isAnimationPlaying` before calling `play` but not before calling `stop`? (I'm mostly just curious about this one!)

Library/OnboardingViewModel.swift
```diff
@@ -0,0 +1,97 @@
...
+
+public final class OnboardingViewModel: OnboardingViewModelType {
+  private let useCase: OnboardingUseCaseType
+  private var cancellables = Set<AnyCancellable>()
+  public var onboardingItems: [OnboardingItem] = []
+
+  @Published public private(set) var currentIndex: Int = 0
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(UUID())
```
ifosli: I'm concerned about this - won't `UUID()` produce a new id each time it's called; ie every time the `hash` function is called on this class it produces a different result? Why does the view model need to be hashable anyways?

Library/OnboardingViewModelTests.swift
```diff
@@ -0,0 +1,89 @@
...
+    appTrackingTransparency.requestAndSetAuthorizationStatusFlag = true
+    appTrackingTransparency.shouldRequestAuthStatus = true
+
+    withEnvironment(
+      appTrackingTransparency: appTrackingTransparency
+    ) {
+      var cancellables: [AnyCancellable] = []
+
+      let expectation = expectation(description: "Waiting for action to be performed")
+      var triggeredPAppTrackingTransparencyPopup = false
```
ifosli: nit: spelling; should this be `triggeredAppTrackingTransparencyPopup` instead of `triggeredPAppTrackingTransparencyPopup`?

ifosli (approved):
I have some minor nits and a couple concerns about line limits and the hash function. I'm going to just mark this as approved bc I don't think there's any major changes I feel the need to re-review, but feel free to ping me if you want me to take another look.

To make things look better for more lines of text: I wonder if you could put just some of the view in a scrollview? Ex maybe just the text, or maybe everything except the buttons at the bottom.

Also, maybe outside the scope of this pr, but a) are you adding/testing voiceover support for this screen? and b) will it support dynamic type (this would make some kind of scrolling basically required)? I definitely think the answer to both these questions should be yes, but I don't mind if it's something you handle in a separate pr!

Kickstarter-iOS/Features/Onboarding/OnboardingStyles.swift
scottkicks: Exactly. The plan is for these colors to be static and not change based on LM/DM.

Kickstarter-iOS/Features/Onboarding/Views/ResizableLottieView.swift
scottkicks: We check isAnimationPlaying before calling play() so we don't accidentally restart the animation or waste resources if it's already running. 

Side effects that can happen when making redundant play() calls on Lottie animations
- restarts from the beginning (causing glitches)
- redundant internal updates
- unnecessary CPU/GPU usage

But stop() is harmless to call even if it’s already stopped, so I don't think there's a need to guard it.

PR #2546, opened 2025-07-23
Jluna/mbl 2640/refunded and refundedamount

jovaniks:
# 📲 What

Added support for the new `refundedAmount` field and the `PaymentIncrementState.REFUNDED` enum to the GraphQL and model mapping. 

# 🛠 How

- Added `includeRefundedAmount` as a GraphQL variable in:
  - `BuildPaymentPlanQuery`
  - `FetchBacking`
  - `FetchUserBackings`
- Updated `PaymentIncrement` GraphQL fragment to include:
  - `refundedAmount { amountFormattedInProjectNativeCurrency, currency }`
- Added `.refunded` to `PaymentIncrementState` enum
- Updated `PledgePaymentIncrement` and related models to include:
  - `refundedAmount: PledgePaymentIncrementAmount?`

# 👀 See

- [MBL-2640](https://kickstarter.atlassian.net/browse/MBL-2640)

# ✅ Acceptance criteria

- [ ] The app is working properly
- [ ] Test pledges with refunded increments PASS

# ⏰ TODO

- [ ] Test with real data, waiting for test users/pledges


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

GraphAPI/Fragments/PaymentIncrementFragment.graphql.swift
ifosli, line 1: @amy-at-kickstarter can we get these graphAPI generated files set up in github as generated files again?

KsApi/Service.swift
```diff
@@ -521,7 +523,8 @@ public struct Service: ServiceType {
             id: "\(id)",
             withStoredCards: withStoredCards,
             includeShippingRules: true,
-            includeLocalPickup: true
+            includeLocalPickup: true,
+            includeRefundedAmount: true
```
ifosli, line 556: Nit: Why does this query need to include the refunded amount but the `FetchUserBackingsQuery` doesn't? Can you add a comment explaining it (unless this is just temporary while you're still implementing refunds)?

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
```diff
@@ -21,6 +21,15 @@ extension PledgePaymentIncrement {
...
       self.stateReason = PledgePaymentIncrementStateReason(rawValue: stateReason)
     }
+
+    if let refundedAmount = fragment.refundedAmount {
+      self.refundedAmount = PledgePaymentIncrementAmount(
+        currency: refundedAmount.currency,
+        amountFormattedInProjectNativeCurrency: refundedAmount.amountFormattedInProjectNativeCurrency
+      )
+    } else {
+      self.refundedAmount = nil
```
ifosli, line 31: To clarify, we set this refunded amount to the same thing if a) we didn't fetch refunded amounts or b) there isn't a refund, correct? I'm concerned that that will lead to bugs where we don't realize there are places we should've fetched the refunded amount, but didn't. Could we just always fetch it when we're fetching payment increments?

KsApi/models/PledgePaymentIncrement+GraphAPITests.swift
```diff
@@ -28,6 +30,37 @@ final class PledgePaymentIncrementGraphAPITests: XCTestCase {
...
+        "__typename": "PaymentIncrement",
+        "amount": {
+          "__typename": "PaymentIncrementAmount",
+          "amountFormattedInProjectNativeCurrency": "$99.75",
+          "currency": "USD"
+        },
+        "scheduledCollection": "2025-03-31T10:29:19-04:00",
+        "state": "COLLECTED",
+        "stateReason": "REQUIRES_ACTION",
+        "refundedAmount": null
```
ifosli: If we're specifically testing the refunded amount in this test, can you set it to something nonnil? (And set `includeRefundedAmount` to `true`?) Also, it might be nice to take advantage of our new apollo tools instead of using a jsonString here!

ifosli (changes requested):
I have concerns with having `includeRefundedAmount` as a parameter instead of always including it. Could you either a) change it to always be included or b) explain why you decided on this approach and add some comments to the code clarifying what makes some fetches need this and others don't?

KsApi/Service.swift
jovaniks: The `refundedAmounts` are only needed in the **ManagePledge** flow, specifically where we display the **Payment Scheduled** component. That’s the only place we actually use this new field from the payment increments.

Outside of that, it’s not necessary to include it. In fact, it was recommended by the Payments team to only fetch this data when strictly needed, since it’s a bit expensive on the backend.

To help clarify this, I created a separate function `fetchBackingWithIncrementsRefundedAmount`, which should make the intent clearer. So if we do need the backing with increment-level `refundedAmount`, we should use that one — otherwise, we stick to `fetchBacking` or `FetchUserBackingsQuery` depending on the case.

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
jovaniks: As I mentioned above, we only want to fetch `refundedAmount` when strictly necessary to avoid slowing down the response due to the extra backend processing this field requires.
Arleigh actually mentioned they’ve been tweaking the query to improve performance, but still recommends not requesting this field unless it’s really needed.

jovaniks: Would you suggest a different approach here?
I created a new function `fetchBackingWithIncrementsRefundedAmount` to explicitly fetch the `refundedAmount` only in the ManagePledge flow where it’s needed.
Do you think it would make sense to instead create a dedicated fragment for this use case, or do you have another idea in mind?

KsApi/models/PledgePaymentIncrement+GraphAPITests.swift
```diff
@@ -1,55 +1,75 @@
...
+
+    XCTAssertNotNil(increment)
+    XCTAssertEqual(increment?.amount.currency, "USD")
+    XCTAssertEqual(increment?.amount.amountFormattedInProjectNativeCurrency, "$99.75")
+    XCTAssertEqual(increment?.scheduledCollection, 1_743_431_359.0)
+    XCTAssertEqual(increment?.state, .collected)
+    XCTAssertNil(increment?.stateReason)
+    XCTAssertNotNil(increment?.refundedAmount)
+    XCTAssertEqual(increment?.refundedAmount?.currency, "USD")
+    XCTAssertEqual(increment?.refundedAmount?.amountFormattedInProjectNativeCurrency, "$55.50")
```
jovaniks, line 60: Test for `refundedAmount` using the new Apollo tools

KsApi/models/PledgePaymentIncrement+GraphAPITests.swift
jovaniks: Updated!

GraphAPI/Fragments/PaymentIncrementFragment.graphql.swift
amy-at-kickstarter: Yup, see #2551

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
amy-at-kickstarter: Two approaches come to mind:

1. Add a private enum to `PledgePaymentIncrement` like `RefundType` with the cases `none`, `unknown` and `some(Money)`
2. Refactor this so that the refund amount isn't part of the default `PaymentIncrementFragment` but is instead a new fragment like `PaymentIncrementBackingFragment` which is also requested in the `FetchBackingQuery` used for pledge management. Then it is clearly something separate that you handle in that flow, not in the `PledgePaymentIncrement` object directly

KsApi/Service.swift
```diff
@@ -521,7 +523,37 @@ public struct Service: ServiceType {
             id: "\(id)",
             withStoredCards: withStoredCards,
             includeShippingRules: true,
-            includeLocalPickup: true
+            includeLocalPickup: true,
+            includeRefundedAmount: false
```
ifosli, line 527: Nit: If you end up writing a new fragment that includes the refunded amount, I think you'll be able to delete it from this query (and the UserBackingsQuery) - which would make the code a little easier to read.

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
ifosli: I like these suggestions, too!

ifosli (approved):
I still think it'd be better to have a way to differentiate between "we didn't fetch the refunded amount" and "there hasn't been a refund" in the data type itself, but it doesn't need to be a blocker if it's going to add a lot of extra work. Everything else looks good now! I like the new tests!

KsApi/models/PledgePaymentIncrement+GraphAPITests.swift
amy-at-kickstarter: Thanks for refactoring this!

amy-at-kickstarter (approved):
I agreed with @ifosli - the difference between no refund and `nil` could be clearer, but I don't think it's a blocker since it's only used on one screen.

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
amy-at-kickstarter: Oh, another idea - you could explicitly set the refund to a `$0` money object if no refund is present. That might also be clearer.

jovaniks: Thinking about the `RefundType` approach, what kind of value would represent that the refund was not requested, and might be `unknown`?

Here’s how I’m interpreting the cases:
	•	`none`: there’s no refund associated with the increment
	•	`some(Money)`: a refund exists
	•	`unknown`: the refund status is unclear because the field was not requested or fetched

Does that sound right?

amy-at-kickstarter: Yup, that sounds right to me.

Reading it again, how about this? Might be even clearer:
* `notRefunded`: There's no refunded associated with this increment
* `partialRefund(Money)`: Some amount was partially refunded
* `fullRefund(Money)`: Some amount was fully refunded
* `unknown`: It was not fetched

jovaniks:
I’d like to move forward with merging this PR. The refactor and feedback raised here have already been addressed in [this other PR](https://github.com/kickstarter/ios-oss/pull/2569/files), so this one can go in as-is.

KsApi/models/PledgePaymentIncrement+GraphAPI.swift
jovaniks: Thanks, this was super helpful! I included the refactor in [this PR](https://github.com/kickstarter/ios-oss/pull/2569). I’ll go ahead and merge this one, and we can continue the discussion on the refactor PR if needed.

PR #2547, opened 2025-07-23
[MBL-2422] Launch Login/Signup Flow

scottkicks:
# 📲 What

- Launches the existing log in/sign up flow
- Dismisses the onboarding flow when tapping the close icon, "X", or "Explore the app" at end of the flow.
- Stores a new UserDefaults key that that tells us that the user has completed the onboarding flow previously.

# 🛠 How

When the log in or sign up button is tapped, we can emit a new Notification `ksr_goToLoginFromOnboarding` that the AppDelegate listens for and respond to.

The AppDelegate can then fires it's ViewModel's existing `gotToLoginWithIntent` output to trigger our existing log in flow.
   - Making sure to dismiss the onboarding flow first to avoid and any odd UIKit/SwiftUI navigation jankieness.

Also stores a new, `hasSeenOnboarding`,  key when the onboarding flow dismisses in UserDefaults.

# 👀 See

|  | After 🦋 |
| --- | --- |
|  | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-07-23 at 12 39 19 |

# ✅ Acceptance criteria

- [x] Log in flow triggers and functions as normal when tapping the log in button
- [x] We can now exit/close the onboarding flow from the "x" icon and "Explore the app" button
- [x] Onboarding flow full functions as expected with permission dialogs and logging in
- [x] Onboarding flow is not shown again if already completed and/or permission dialogs have been been previously interacted with.

# ⏰ TODO

- [ ] Translations and analytics


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -101,6 +101,13 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.viewModel.outputs.goToLoginWithIntent
       .observeForControllerAction()
       .observeValues { [weak self] intent in
+        /// Dismiss OnboardingView if present so that we can correclty present the LoginToutViewController.
```
amy-at-kickstarter: Nit: typo `correclty`

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -101,6 +101,13 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        if let onboardingView = self?.rootTabBarController?
+          .presentedViewController as? UIHostingController<OnboardingView> {
+          onboardingView.dismiss(animated: true)
+          AppEnvironment.current.userDefaults.set(true, forKey: AppKeys.hasSeenOnboarding.rawValue)
```
amy-at-kickstarter, line 108: Would it be simpler for `rootTabBarController` to just keep a `weak` reference to the onboarding view?

Library/KeyValueStoreType.swift
```diff
@@ -129,6 +131,15 @@ extension KeyValueStoreType {
     }
   }
 
+  public var hasSeenOnboarding: Bool {
+    get {
+      return self.bool(forKey: AppKeys.hasSeenOnboarding.rawValue)
```
amy-at-kickstarter, line 136: Does it matter that this is device-specific instead of user-specific? (If you create a new user, should you see onboarding again?)

scottkicks: Yeah we've decided that this will be device specific since system permissions for Push Notifications and ATT need to be granted per device.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -295,7 +298,7 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
     let pushTokenRegistrationStartedEvents = Signal.merge(
       self.didAcceptReceivingRemoteNotificationsProperty.signal,
       pushNotificationsPreviouslyAuthorized
-        .filter { isTrue($0) && featureOnboardingFlowEnabled() == false }
+        .filter { isPreviouslyAuthorzied in isPreviouslyAuthorzied }
```
ifosli, line 299: nit: The wording here feels confusing to me. Assuming you no longer need the flag check, thoughts on changing this back to just a `.filter(isTrue)` instead?

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -308,12 +311,19 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
+        AppEnvironment.current.pushRegistrationType.hasAuthorizedNotifications()
+          .filter { isAuthorized in isAuthorized == false }
+          .compactMap { _ in
+            guard let context = showDialog.userInfo?.values.first as? PushNotificationDialog.Context,
+                  PushNotificationDialog.canShowDialog(for: context) else {
+              return nil
+            }
+
+            return showDialog
+          }
```
ifosli: Assuming this change is also here intentionally, can you update the pr title/description? Based on the title, I wasn't expecting to see any push notification changes (I'd prefer to see these changes in a separate pr, but it's a little late for that now, maybe)

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -303,12 +311,19 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
-      .filter {
-        if let context = $0.userInfo?.values.first as? PushNotificationDialog.Context {
-          return PushNotificationDialog.canShowDialog(for: context)
-        }
-        return false
+    self.showAlert = self.showNotificationDialogProperty.signal
+      .skipNil()
+      .flatMap { showDialog in
+        AppEnvironment.current.pushRegistrationType.hasAuthorizedNotifications()
+          .filter { isAuthorized in isAuthorized == false }
```
ifosli: This looks like a significant change. I could be wrong about this, but my understanding is that we show the push notification pop up a fair amount, including for users that have already turned on notifications. For example, if I back a projects, I might get a notification popup about notifications for that project or the project's category. With this change, it seems like someone that's already turned on some notifications will not be prompted to add more notifications. Do I have that right? If so, are you sure we want to make this change?

Kickstarter-iOS/AppDelegateViewModelTests.swift
```diff
@@ -998,7 +998,7 @@ final class AppDelegateViewModelTests: TestCase {
       self.vm.inputs.applicationDidFinishLaunching(application: UIApplication.shared, launchOptions: [:])
       self.vm.inputs.userSessionStarted()
 
-      self.pushRegistrationStarted.assertValueCount(0)
+      self.pushRegistrationStarted.assertValueCount(1)
```
ifosli, line 999: I'm so confused by the push notification changes in this pr. Can you please explain why the expectation in this test changed the way it did?

Kickstarter-iOS/AppDelegateViewModelTests.swift
```diff
@@ -2401,6 +2415,90 @@ final class AppDelegateViewModelTests: TestCase {
...
+    ]
+
+    MockPushRegistration.hasAuthorizedNotificationsProducer = .init(value: false)
+    MockPushRegistration.registerProducer = .init(value: true)
+
+    let appTrackingTransparency = MockAppTrackingTransparency()
+    appTrackingTransparency.requestAndSetAuthorizationStatusFlag = true
+    appTrackingTransparency.shouldRequestAuthStatus = true
+
+    userDefaults.set(true, forKey: AppKeys.hasSeenOnboarding.rawValue)
```
ifosli, line 2430: nit: should the `userDefaults` be in the environment/be set locally in this test only so that it doesn't affect other tests?

Kickstarter-iOS/AppDelegateViewModelTests.swift
```diff
@@ -2401,6 +2415,90 @@ final class AppDelegateViewModelTests: TestCase {
...
+    ]
+
+    MockPushRegistration.hasAuthorizedNotificationsProducer = .init(value: false)
+    MockPushRegistration.registerProducer = .init(value: true)
+
+    let appTrackingTransparency = MockAppTrackingTransparency()
+    appTrackingTransparency.requestAndSetAuthorizationStatusFlag = true
+    appTrackingTransparency.shouldRequestAuthStatus = true
+
+    userDefaults.set(false, forKey: AppKeys.hasSeenOnboarding.rawValue)
```
ifosli: nit: Will this key ever actually be set to false if a user hasn't seen the onboarding flow yet, or will it just be unset? If it's the latter, that's what this test should test, too.

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -31,7 +31,7 @@ struct CallToActionView: View {
         self.secondaryButton(title: "FPO: Not right now", action: self.onSecondaryTap)
 
       case .loginSignUp:
-        self.primaryButton(title: "FPO: Sign up or log in", action: self.onLoginSignup, for: self.item)
+        self.primaryButton(title: "FPO: Sign up or log in", action: self.onPrimaryTap, for: self.item)
```
ifosli, line 33: If you're no longer using the `onLoginSignup` completion, can you delete it?

Library/KeyValueStoreType.swift
ifosli: I second Amy's question. If a new user logs in on a device that's already seen the onboarding flow (for a different user), should they see the onboarding flow for the new user as well? They definitely need to re-register for push notifications, but I'm not sure how the tracking permissions work.

ifosli (changes requested):
General login flow and onboarding tracking looks good; just some nits. I'm really confused about what's going on with the push notification changes in this pr, though; can you either update the code, pull it out into a separate pr, or explain it to me?

Kickstarter-iOS/AppDelegateViewModel.swift
scottkicks: Oh I wasn't aware that we intentionally show this alert when users have already turned on notifications. That would makes sense then for this change not to be necessary. I'll revert. 

Thanks for sharing that extra context!

Kickstarter-iOS/AppDelegateViewModel.swift
scottkicks: replied to your other comment on this 😄

Kickstarter-iOS/AppDelegateViewModelTests.swift
scottkicks: no longer needed since I've reverted my changes to the notifications alert logic.

Kickstarter-iOS/AppDelegateViewModelTests.swift
scottkicks: I'm a little confused. It is being set locally for this test. this `userDefaults` is actually the Mock coming from TestCase.

Kickstarter-iOS/AppDelegateViewModelTests.swift
scottkicks: user defaults will return false if it technically hasn't been set yet. I can still throw an AssertFalse down here though to make that a little clearer in this test!

Library/KeyValueStoreType.swift
scottkicks: App permissions like these are determined at a device level, so if the push notification system dialog is shown then the user logs out and then a new user logs in, we won't be able to present the system permission dialogs. However, out custom notification alert will still show on log in still informing the user that they can go and enable their notifications.

Kickstarter-iOS/AppDelegateViewModelTests.swift
ifosli: Oh sorry; based on the code, I thought it looked like `userDefaults` was shared between all tests in this file. (So running them out of order could lead to weirdness.) If I'm wrong about that, I'm happy with the test as-is.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -769,7 +779,17 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
...
       .skipRepeats()
       .ksr_delay(.seconds(1), on: AppEnvironment.current.scheduler)
-      .filter { isTrue($0) && featureOnboardingFlowEnabled() == false }
+      .filter { applicationIsActive in
+        /// Only attempt to request authorization outside of onboarding if the application is active and the user has seen the onboarding flow.
+        /// We don't want to request authorzation in the onboarding flow unless they've tapped the "all tracking" CTA.
+        let hasSeenOnboarding = AppEnvironment.current.userDefaults.hasSeenOnboarding == true
+
+        if featureOnboardingFlowEnabled() == true {
+          return applicationIsActive && AppEnvironment.current.userDefaults.hasSeenOnboarding == true
```
ifosli: nit: use the `hasSeenOnboarding` variable here, too, instead of calculating it twice?

Kickstarter-iOS/AppDelegateViewModelTests.swift
ifosli: Some of these lines still need to be reverted, too

PR #2550, opened 2025-07-24
Feat/adyer/mbl 2646/dark mode bug

amy-at-kickstarter:
# 📲 What

1. Make sure that activity sample cells respond correctly to dark mode.
2. Delete the dark mode feature flag (our OLDEST feature flag 🎉 )

dark mode activity samples

# 🤔 Why

Becca reported an issue in which the activity sample cell has the wrong text color. I wrote some screenshot tests for it and did some manual QA.

The only way I was able to duplicate issues with those cells was by toggling the feature flag, not toggling the interface style itself. My best guess is that the remote feature flag changing state fairly close to launch time led to mis-mismatched colors in some of the app's state.

Since dark mode is now on at 100%, I just cleaned up the flag - and hopefully that resolves any future issues!


Kickstarter-iOS/Features/Discovery/Datasource/DiscoveryProjectsDataSource.swift
```diff
@@ -18,11 +18,11 @@ internal final class DiscoveryProjectsDataSource: ValueCellDataSource {
     activities.forEach { activity in
       switch activity.category {
       case .backing:
-        self.set(values: [activity], cellClass: ActivitySampleBackingCell.self, inSection: section)
```
amy-at-kickstarter, line 21: Since only one activity is ever shown, this should have the same effect in practice. (The cells are cleared right above these lines.) I needed to make this change so I could have my multi-cell screenshot test.

Library/ViewModels/RemoteConfigFeatureFlagToolsViewModelTests.swift
```diff
@@ -63,7 +63,7 @@ final class RemoteConfigFlagToolsViewModelTests: TestCase {
   func testUpdateUserDefaultsWithFeature_FeatureIsEnabled() {
     let mockRemoteConfigClient = MockRemoteConfigClient()
       |> \.features .~ [
-        RemoteConfigFeature.darkModeEnabled.rawValue: false
+        RemoteConfigFeature.editPledgeOverTimeEnabled.rawValue: false
```
amy-at-kickstarter, line 66: These tests had nothing to do with dark mode. I just picked another flag off the top of the list.

Kickstarter-iOS/Features/Discovery/Controller/__Snapshots__/DiscoveryPageViewControllerTests/testView_Activity_AllActivities.lang_es_device_pad_light.png
amy-at-kickstarter, line 1: Note that all 3 card types now have a shadow background. Before, the backing card had a border background.

Kickstarter-iOS/Features/BackerDashboardProjects/Controller/BackerDashboardProjectsViewControllerTests.swift
```diff
@@ -57,7 +57,6 @@ internal final class BackerDashboardProjectsViewControllerTests: TestCase {
 
     let darkModeOn = MockRemoteConfigClient()
     darkModeOn.features = [
-      RemoteConfigFeature.darkModeEnabled.rawValue: true,
       RemoteConfigFeature.newDesignSystem.rawValue: true
```
ifosli, line 60: nit: Considering we had UI issues when the dark mode flag was on and the new design system was off, I think we should try to also clean up the new design system flag while we're at it. (Not in this pr, but would be nice to do this week!)

ifosli (approved):
🎉

Kickstarter-iOS/Features/BackerDashboardProjects/Controller/BackerDashboardProjectsViewControllerTests.swift
amy-at-kickstarter: Yup! Ticketed that for this sprint.

PR #2554, opened 2025-07-28
[MBL-2618] Rename search empty state cell

ifosli:
# 📲 What

Rename old empty state cell. I originally had this in the pr where I added the new search empty state cell, but github did a really bad job of marking files correctly as moved/new when viewing all commits at once, so I'm doing separate prs instead.

This cell and it's corresponding test will be deleted once the "use new empty state search" flag is cleaned up.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2618)

# ✅ Acceptance criteria

- [x] Tests pass


Library/ViewModels/SearchLegacyEmptyStateCellViewModel.swift
ifosli, line 1: Note: This file is just the original file with new names; I think github isn't able to tell that it's the same file because so many of the lines changed (due to renames)

Kickstarter-iOS/Features/Search/Views/Cell/SearchLegacyEmptyStateCell.swift
```diff
@@ -39,6 +39,11 @@ internal final class SearchEmptyStateCell: UITableViewCell, ValueCell {
 
     _ = self.rootStackView
       |> UIStackView.lens.layoutMargins .~ .init(topBottom: Styles.grid(4), leftRight: Styles.grid(4))
+
+    let filterResultsButton = UIButton(type: .custom)
+    filterResultsButton.setTitle("Remove all filters test", for: .normal)
+
+    self.rootStackView.addSubview(filterResultsButton)
```
amy-at-kickstarter: Was this unintentionally left in here?

amy-at-kickstarter (approved):
LGTM except for some leftover test code!

Kickstarter-iOS/Features/Search/Views/Cell/SearchLegacyEmptyStateCell.swift
ifosli: Oh oops, thanks for catching that!

PR #2555, opened 2025-07-28
[MBL-2618] Add new empty state to search

ifosli:
# 📲 What

Create new search empty state (behind a flag just in case). 

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2618)
[Figma](https://www.figma.com/design/a9GQ3iQn7XzRUxbpE3cVpE/Search---Discover?node-id=3392-157464&t=UjAb6fz1D2soYB7q-0)

| Old UI | After (filters only) |
| --- | --- | 
| image | image |
|  After (query only) | After (query and filters) |
|image | image |
# ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type (except for a line break button issue that I'm fixing in a separate pr for visibility)

# ✅ Acceptance criteria

- [x] New empty state UI shows when it should, and in the correct state
- [x] Old empty state UI is unchanged

# ⏰ TODO

- I'll add string translations once they're ready, in a separate pr


Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
```diff
@@ -95,7 +95,15 @@ internal final class SearchViewController: UITableViewController {
     self.viewModel.outputs.showEmptyState
       .observeForUI()
       .observeValues { [weak self] params, visible in
-        self?.dataSource.load(params: params, visible: visible)
+        if featureSearchNewEmptyState() {
+          let data = SearchEmptyStateSearchData(
+            query: params.query,
+            hasFilters: self?.viewModel.outputs.searchFilters.hasFilters == true
```
ifosli, line 101: I wanted to construct this data object in the view model. However, the `searchFilters` is a property that's not accessible when we create our signals in the view model (since they're created in init, and swift didn't like me trying to access `self` before it was fully initialized). If anyone has ideas for a better way to handle this, I'm open to suggestions.

Library/UseCases/SearchFilters/SearchFilters.swift
```diff
@@ -84,7 +84,7 @@ public class SearchFilters: ObservableObject {
     .count(where: { $0 == true })
   }
 
-  var hasFilters: Bool {
+  public var hasFilters: Bool {
```
ifosli, line 87: This needs to be explicitly marked `public` to be accessible outside `Library`. I left most of this class alone, but the `filterCount` property feels linked enough to the `hasFilters` property that I made them both public.

Library/ViewModels/SearchEmptyStateCellViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+import ReactiveSwift
+
+public struct SearchEmptyStateSearchData {
+  public let query: String?
+  public let hasFilters: Bool
+
+  public init(query: String?, hasFilters: Bool) {
+    self.query = query
+    self.hasFilters = hasFilters
+  }
```
ifosli, line 11: Since I'm initializing this outside of `Library`, the initializer needs to be explicitly created in order to be marked `public`

Library/ViewModels/SearchViewModel.swift
```diff
@@ -231,13 +231,31 @@ public final class SearchViewModel: SearchViewModelType, SearchViewModelInputs,
...
+    let shouldShowEmptyState = Signal.merge(
+      shouldShowNewEmptyState.compactMap {
+        if featureSearchNewEmptyState() { return $0 }
+        return nil
+      },
+      shouldShowOldEmptyState.compactMap {
+        if featureSearchNewEmptyState() { return nil }
+        return $0
+      }
+    )
```
ifosli, line 262: These code changes are a little convoluted, but basically: with the new empty state, we need to take filters into account in order to know if the search query changed. I wanted to leave the old logic alone, for rollback safety. The `compactMap` here ensures that we're using the most recent value of the feature flag, in order to keep this and the check in the view controller in sync. This code will get cleaned up when the feature flag is deleted.

Kickstarter-iOS/Features/Search/Controller/SearchViewController.swift
amy-at-kickstarter: Given the various weirdnesses of how I implemented this, I think this solution is good! It'd take a bigger refactor to fix this plumbing.

Kickstarter-iOS/Features/Search/Controller/SearchViewControllerTests.swift
```diff
@@ -128,6 +128,58 @@ internal final class SearchViewContollerTests: TestCase {
...
+    let mockRemoteConfig = MockRemoteConfigClient()
+    mockRemoteConfig.features = [
+      RemoteConfigFeature.searchNewEmptyState.rawValue: true,
+      RemoteConfigFeature.darkModeEnabled.rawValue: true,
+      RemoteConfigFeature.newDesignSystem.rawValue: true
+    ]
+
+    var emptyStateTypeIterator = 0
+
+    orthogonalCombos(
```
amy-at-kickstarter, line 147: 😎

Library/ViewModels/SearchEmptyStateCellViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+}
+
+public final class SearchEmptyStateCellViewModel: SearchEmptyStateCellViewModelType,
+  SearchEmptyStateCellViewModelInputs, SearchEmptyStateCellViewModelOutputs {
+  public init() {
+    self.titleText = self.searchDataProperty.signal
+      .skipNil()
+      .map { searchData in
+        if let query = searchData.query, query.count > 0 {
+          return "FPO: No results for \"\(query)\""
```
amy-at-kickstarter, line 36: When this gets translated, it'd be nice to use pretty quotes instead of plain ones

Kickstarter-iOS/Features/Search/Views/Cell/SearchEmptyStateCell.swift
```diff
@@ -0,0 +1,108 @@
...
+
+  internal override func bindViewModel() {
+    super.bindViewModel()
+
+    self.titleLabel.rac.text = self.viewModel.outputs.titleText
+    self.subtitleLabel.rac.text = self.viewModel.outputs.subtitleText
+    self.viewModel.outputs.hideClearFiltersButton
+      .observeForUI()
+      .observeValues { [weak self] isHidden in
+        self?.clearFiltersButton.isHidden = isHidden
```
amy-at-kickstarter: Huh, isn't there a `rac` property for `isHidden`?

Library/ViewModels/SearchViewModel.swift
amy-at-kickstarter: This would also be a good comment!

Kickstarter-iOS/Features/Search/Views/Cell/SearchEmptyStateCell.swift
```diff
@@ -0,0 +1,108 @@
...
+internal final class SearchEmptyStateCell: UITableViewCell, ValueCell {
+  typealias Value = SearchEmptyStateSearchData
+
+  fileprivate let viewModel: SearchEmptyStateCellViewModelType = SearchEmptyStateCellViewModel()
+
+  weak var delegate: SearchEmptyStateCellDelegate?
+
+  private var titleLabel: UILabel = UILabel()
+  private var subtitleLabel: UILabel = UILabel()
+  private var clearFiltersButton: UIButton
```
jovaniks: Although the property is currently declared as `UIButton`, it is actually instantiated as a `KSRButton`, which applies custom styling and behavior. This could be misleading for other developers, as `KSRButton` does not behave exactly like a native `UIButton`.

To improve type safety and clarity, we should consider updating the property type to `KSRButton`. This makes the intent explicit and helps avoid confusion when working with custom styles, etc.

Library/ViewModels/SearchEmptyStateCellViewModel.swift
ifosli: Yeah, the translated string has pretty quotes!

Kickstarter-iOS/Features/Search/Views/Cell/SearchEmptyStateCell.swift
ifosli: Oh you're right! I thought there was one, too, but I wasn't able to find it; turns out the `rac` is named `hidden` instead 😅

Kickstarter-iOS/Features/Search/Views/Cell/SearchEmptyStateCell.swift
ifosli: I used UIButton here since that's the part I care about (KSRButton is a convenient initializer) but I don't feel strongly about it, so I'll update it.

Kickstarter-iOS/Features/Search/Views/Cell/SearchLegacyEmptyStateCell.swift
ifosli, line 1: I made this change in the base branch after putting up this pr. I added it to this branch as well to keep them in sync, but I guess git remembers the state of the base branch when this pr was created, instead. Lesson learned

PR #2556, opened 2025-07-28
Stop truncating KSRButton text

ifosli:
# 📲 What

The UIKit version of the `KSRButton` was (probably accidentally) set up so it truncated text if the button text didn't fit on one line. I'm updating the button to wrap text (by word) instead.

I did a quick code search and it looks like the only usage of this button is in the `RewardTrackingDetailsView`. I'm not sure how to make that view show up, so I didn't properly test these changes, but I looked at the snapshot tests and it definitely looks like that's a button that'll benefit from this change, too.

# 👀 See

[Jira for the empty state search ticket](https://kickstarter.atlassian.net/browse/MBL-2618) (I was testing dynamic type for the new UI when I found this issue)

# ✅ Acceptance criteria

- [ ] Tests pass

### DIFF
--- a/Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
+++ b/Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
@@ -45,7 +45,7 @@ extension UIButton {
     buttonConfiguration.imagePadding = Styles.grid(1)
     buttonConfiguration.imagePlacement = .leading
 
-    buttonConfiguration.titleLineBreakMode = .byTruncatingMiddle
+    buttonConfiguration.titleLineBreakMode = .byWordWrapping
     buttonConfiguration.titleAlignment = .center
 
     self.tintColor = styleConfig.titleColor


scottkicks (approved):
👍  

Setting up a project that will show reward tracking, to get the Reward Tracking Details to show, is somewhat complicated and requires some backend help right now. 

I think using the snapshots as a reference for now makes sense!

jovaniks (approved):
I finally figured out why this setting was there — the [legacy buttons](https://github.com/kickstarter/ios-oss/blob/main/Library/Styles/ButtonStyles.swift#L34) use the same configuration. So we’re just mirroring the behavior from those styles.

To test this, we can use any button in the app with the **New Design System** feature flag enabled, since they also apply the same `KSRButtonStyleConfiguration.applyStyleConfiguration(_:)` method as `KSRButton`. So in theory, they should behave the same way.

PR #2562, opened 2025-07-29
[MBL-2569] Onboarding Analytics

scottkicks:
# 📲 What

Adds analytics events to cover the onboarding flow.

- Expected event structure and coverage can be found in [this doc](https://docs.google.com/document/d/1wDgnjIS7NgCDYlSR1Dn7i1T51OWjTOy2XRa9aboaTEE/edit?tab=t.0).
- Worked with Yun to make sure android and ios are aligned as closely as possible.

Also includes some small cleanup.

# ✅ Acceptance criteria

- [x] All events that insights expects (listed in [this doc](https://docs.google.com/document/d/1wDgnjIS7NgCDYlSR1Dn7i1T51OWjTOy2XRa9aboaTEE/edit?tab=t.0)) are accounted for and called with the correct structure
- [x] All events that insights expects (listed in [this doc](https://docs.google.com/document/d/1wDgnjIS7NgCDYlSR1Dn7i1T51OWjTOy2XRa9aboaTEE/edit?tab=t.0)) are accounted for and called at the correct times in the onboarding flow.


Library/OnboardingUseCase.swift
```diff
@@ -38,6 +39,18 @@ public struct OnboardingItem: Identifiable, Equatable {
...
+    title: String,
+    subtitle: String,
+    lottieView: LottieAnimationView = .init(),
+    type: OnboardingItemType
+  ) {
+    self.title = title
+    self.subtitle = subtitle
+    self.lottieView = lottieView
+    self.type = type
+  }
```
scottkicks, line 53: Added this public init to make initializing this type in my tests.

Library/OnboardingUseCase.swift
```diff
@@ -55,14 +68,10 @@ public protocol OnboardingUseCaseUIInputs {
 
   /// Call when a user taps on the "Next", "Not right now", or "Explore the app" buttons.
   func goToNextItemTapped(item: OnboardingItemType)
-
-  func goToLoginSignupTapped()
```
scottkicks, line 59: removing this login signup logic. It's no longer being used since the log in flow is triggered via a Notification. Just missed this in my cleanup.

Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -258,8 +258,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/stripe/stripe-ios",
       "state" : {
-        "revision" : "8a35266fab2e0ba116dec7d651ae69a0ca7bc7b7",
-        "version" : "23.27.2"
+        "revision" : "d20549e5a83b69d47f041ce2ba9af789e3e1a626",
+        "version" : "23.15.0"
```
ifosli: Nit: This looks completely unrelated - can it be reverted easily? If not, I'm fine with leaving it.

Library/OnboardingViewModel.swift
```diff
@@ -64,25 +62,40 @@ public final class OnboardingViewModel: OnboardingViewModelType {
 
   // MARK: - Inputs
 
+  public func onAppear() {
+    AppEnvironment.current.ksrAnalytics.trackOnboardingPageViewed(at: .welcome)
+    AppEnvironment.current.ksrAnalytics.trackOnboardingPageButtonTapped(
+      context: .onboardingSignUpLogIn,
+      item: .welcome
+    )
```
ifosli: Just confirming that users tap a button in order to enter the onboarding flow? If not, what's this button tap tracking?

Library/OnboardingViewModel.swift
```diff
@@ -28,35 +37,100 @@ public final class OnboardingViewModel: OnboardingViewModelType {
...
+
+    _ = self.goToNextItemTappedSignal.signal
+      .on(value: { item in
+        AppEnvironment.current.ksrAnalytics.trackOnboardingPageViewed(at: item.type)
+        AppEnvironment.current.ksrAnalytics.trackOnboardingPageButtonTapped(
+          context: .onboardingNext,
+          item: item.type
+        )
+      })
+      .observeValues { _ in }
```
amy-at-kickstarter: Is there a reason this is handled in `on` instead of in `observeValues`? I don't think it matters since they both are valid for side effects, but we also don't use `on` very much.

Library/OnboardingViewModelTests.swift
```diff
@@ -87,17 +86,106 @@ final class OnboardingViewModelTest: TestCase {
...
+  func testOnAppear_FiresPageViewedAnalyticsEvents() {
+    self.viewModel.inputs.onAppear()
 
-    self.viewModel.goToLoginSignupTapped()
+    XCTAssertEqual(["Page Viewed"], self.segmentTrackingClient.events)
+    XCTAssertEqual("onboarding", self.segmentTrackingClient.properties.last?["context_page"] as? String)
+    XCTAssertEqual("welcome", self.segmentTrackingClient.properties.last?["context_section"] as? String)
+  }
+
+  func testDidCompleteAppTrackingDialog_FiresAnalyticsEvent() {
```
amy-at-kickstarter, line 103: Ha, I was going to ask if making these log without a signal imeded your ability to test them. Looks like it doesn't!

Library/OnboardingUseCase.swift
ifosli: I think it's too late now, but in the future, this should've been its own pr. Smaller prs are easier to review, and this change isn't really related to analytics

amy-at-kickstarter (approved):

Library/OnboardingViewModel.swift
amy-at-kickstarter: Also, it's inconsistent that this set of tracking data is logged by observing a signal, whereas the other tracking calls are made directly from the input functions like in `onAppear`. Can you call this one directly, instead - or re-use this approach and add a signal for tracking the other events?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -155,17 +154,22 @@ public struct OnboardingView: View {
...
-      self.hasSeenOnboarding()
-      self.dismiss()
+      self.handleClose()
     }
   }
 
+  private func handleClose() {
+    self.hasSeenOnboarding()
+    self.dismiss()
+    self.viewModel.inputs.onboardingFlowEnded()
```
ifosli: Question: will this get handled correctly when `onboardingFlowEnded` is called after `dismiss`? I would expect `dismiss` to also clean up the view (so referencing `self` no longer makes sense)

Library/OnboardingViewModel.swift
```diff
@@ -56,10 +61,14 @@ public final class OnboardingViewModel: OnboardingViewModelType {
     lhs.id == rhs.id
   }
 
+  public var inputs: OnboardingViewModelInputs { return self }
+  public var outputs: OnboardingViewModelOutputs { return self }
```
ifosli, line 65: nit: Are you planning to define the inputs as inputs only and the outputs as outputs only? If not, I don't see the point in defining `inputs` and `outputs` at all

Library/OnboardingUseCase.swift
```diff
@@ -116,6 +116,9 @@ public final class OnboardingUseCase: OnboardingUseCaseType, OnboardingUseCaseUI
               /// If already authorized, do nothing.
               return .empty
             } else {
+              AppEnvironment.current.ksrAnalytics
+                .trackSystemPermissionsDialogViewed(on: .onboardingNotificationsDialog)
+
               /// Otherwise, trigger the system dialog to request authorization.
```
ifosli: nit: I think this comment would make more sense above the tracking code (as the first thing in the else block)

Library/OnboardingViewModel.swift
```diff
@@ -42,7 +45,24 @@ public final class OnboardingViewModel: OnboardingViewModelType, Equatable & Ide
     self.onboardingItems = self.useCase.uiOutputs.onboardingItems
     self.didCompletePushNotificationSystemDialog = self.useCase.outputs
       .didCompletePushNotificationSystemDialog
+      .map { _ in
```
ifosli: Nit: `map` doesn't feel like the correct signal to use here, but I'm not sure what the correct signal is, so I don't consider this blocking. Maybe `observe` would be better?

Library/OnboardingViewModel.swift
```diff
@@ -42,7 +45,24 @@ public final class OnboardingViewModel: OnboardingViewModelType, Equatable & Ide
...
+
+        return ()
+      }
+
     self.triggerAppTrackingTransparencyPopup = self.useCase.outputs.triggerAppTrackingTransparencyDialog
+      .map { _ in
+        AppEnvironment.current.ksrAnalytics
+          .trackSystemPermissionsDialogViewed(on: .onboardingAppTrackingDialog)
+
+        return ()
```
ifosli, line 56: Same for this signal about the `map` feeling weird when the input and outputs aren't used/changed

Library/OnboardingViewModel.swift
```diff
@@ -68,11 +88,6 @@ public final class OnboardingViewModel: OnboardingViewModelType, Equatable & Ide
 
   public func onAppear() {
     AppEnvironment.current.ksrAnalytics.trackOnboardingPageViewed(at: .welcome)
-
-    AppEnvironment.current.ksrAnalytics.trackOnboardingPageButtonTapped(
-      context: .onboardingSignUpLogIn,
-      item: .welcome
-    )
```
ifosli: Ah, I see - this code gets deleted in a new commit.

Library/Tracking/AppTrackingTransparency.swift
```diff
@@ -20,7 +20,10 @@ public class AppTrackingTransparency: AppTrackingTransparencyType {
     self.updateAdvertisingIdentifier()
   }
 
-  public func requestAndSetAuthorizationStatus(_ completion: (() -> Void)? = nil) {
+  public func requestAndSetAuthorizationStatus(_ completion: (
+    (ATTrackingManager.AuthorizationStatus)
+      -> Void
+  )? = nil) {
```
ifosli: nit: The line break here got weird; I'm guessing the formatter chose it but maybe see if it'll also accept something different on fewer lines? Ex maybe putting the entire completion param on its own line?

Library/Tracking/KSRAnalytics.swift
```diff
@@ -1316,6 +1325,42 @@ public final class KSRAnalytics {
...
+  /**
+   Call when a system permissions dialog is viewed in the onboarding flow.
+   */
+  public func trackSystemPermissionsDialogViewed(on sectionContext: SectionContext) {
+    let props = contextProperties(page: .onboarding, sectionContext: sectionContext)
+    self.track(event: SegmentEvent.pageViewed.rawValue, properties: props)
+  }
+
+  /**
+   Call when the user has finished interacting with the Push Notification permissions dialog that was presented in the onboarding flow..
```
ifosli: Nit: Looks like you got an extra `.` in this comment and the next one

Library/OnboardingUseCase.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 import KsApi
 import Lottie
 import ReactiveSwift
+import UIKit
```
ifosli, line 5: nit: Was this new import intentional? Generally, we don't need `UIKit` in `Library`, but I think there are definitely exceptions to that

Library/OnboardingViewModelTests.swift
ifosli, line 1: Nice! I appreciate the comments in these tests!

Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
ifosli, line 1: Ah, you fixed this, too!

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -80,6 +80,9 @@ public struct OnboardingView: View {
 
       /// Handle push notification system dialog completion
       self.viewModel.outputs.didCompletePushNotificationSystemDialog.observeValues {
+        UNUserNotificationCenter.current().getNotificationSettings { settings in
```
ifosli: If this needs to be on the main thread, shouldn't you add an `observeForUI` before the `observeValues`? (You should be able to run code on the main thread from within the view model itself, too, if that would be cleaner!)

ifosli (approved):
Overall, looks good! I did review this commit-by-commit, so some of my questions got resolved by a future commit (but I didn't always go back and find my original comments to delete them). I added a bunch of comments, but most of them are minor nits. Also, for future reference, it'd be nice if a pr this big got split into multiple prs instead (maybe one for each type of notification, approx the way your commits were organized)

Kickstarter.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
scottkicks: resolved in a later commit!

Library/OnboardingViewModel.swift
scottkicks: resolved in a later commit!

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
scottkicks: It does and I verified this in my tests, but I can move it up so that future devs aren't wondering the same.

Library/OnboardingViewModel.swift
scottkicks: I want to stay consistent with our current view model pattern and making a clear separation between inputs and outputs with these protocols is still helpful imo.

Library/OnboardingViewModel.swift
scottkicks: resolved in a later commit!

Library/Tracking/AppTrackingTransparency.swift
scottkicks: The formatter was being very finicky with this, but I'll try moving the entire completion param down and see how that works out!

Library/OnboardingUseCase.swift
scottkicks: Yes this is intentional. In the `OnboardingItem`'s init, since I'm having the LottieAnimationView's default value be `.init()` we need UIKit to be imported. 

Or at least this is what the compiler complains about.

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
scottkicks: Ah, yeah, I originally had this logic in my view model, and it seemed to work fine, but my unit tests were failing due to a threading issue. I wasn't seeing a way to make it thread safe from there, so I thought moving it up to the view would be fine.

PR #2564, opened 2025-07-30
Remove Onboarding Feature Flag Gate

scottkicks:
# What

This PR updates the logic for triggering the onboarding flow by relying solely on permission-related signals (i.e. whether the user has already interacted with the push notification and App Tracking Transparency dialogs) rather than a remote config flag.

In practice, this achieves the same behavior as the existing remote config flag:
- **New users** who haven't seen either system dialog, or the onboarding flow, will be eligible to see onboarding.
- **Existing users** will have already seen the App Tracking Transparency dialog (at minimum), and thus will not see onboarding again.

# 🤔 Why

The onboarding flow is currently gated behind a Remote Config feature flag; however, there's a race condition where fresh installs sometimes initialize before Remote Config activates, causing the flag to default to false—even though it's set to true in the dashboard.

# 🛠 How

Removed the Remote Config flag from the gating logic.

Onboarding is now triggered only if both:
- The user has not yet authorized push notifications or AppTrackingTransparency

This makes sure that:
- Existing users (who have seen the tracking dialog) are not shown onboarding again.
- New users reliably enter the onboarding flow.

# ✅ Acceptance criteria

- [x] Onboarding flow still functions as expected
- [x] All Tests pass
- [x] Existing users experience no changes

# ⏰ TODO

- [ ] Remove the actual feature flag code in a follow up PR


Kickstarter-iOS/AppDelegateViewModelTests.swift
scottkicks, line 1: Did some needed test clean up along the way 🧹

amy-at-kickstarter (approved):

PR #2565, opened 2025-07-30
[MBL-2417] Onboarding String Translations

scottkicks:
# 📲 What

Updates onboarding strings

Note: We don't have some translations for the accessibility labels/hints in the OnboardingView yet 

# 👀 See

| | After 🦋 |
| --- | --- |
|  | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-07-28 at 11 47 29 |

# ✅ Acceptance criteria

- [x] Hardcoded strings use Localized strings
- [x] All tests pass


Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -70,33 +73,30 @@ struct CallToActionView: View {
...
 
-  // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
   private func accessibilityLabel(for item: OnboardingItem) -> LocalizedStringKey {
     switch item.type {
-    case .welcome:
-      "FPO: Welcome"
-    case .saveProjects:
-      "FPO: Save Projects."
+    case .welcome, .saveProjects:
+      LocalizedStringKey(stringLiteral: Strings.project_checkout_navigation_next())
```
amy-at-kickstarter, line 79: I'm unfamiliar with `LocalizedStringKey`, what does this do for us?

amy-at-kickstarter: Also, should this be a new string just for onboarding? It's confusing that it's a `project_checkout` string.

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -70,33 +73,30 @@ struct CallToActionView: View {
...
     }
   }
 
-  // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
+  // TODO: Add accessibility translations [mbl-2418]
   private func accessibilityHint(for item: OnboardingItem) -> LocalizedStringKey {
     switch item.type {
     case .welcome, .saveProjects:
-      "FPO: Navigate to next oboarding view."
+      LocalizedStringKey(stringLiteral: "Navigate to next oboarding view.")
```
amy-at-kickstarter: Typo

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
scottkicks: Accessibility labels need a LocalizedStringKey, and unfortunately, a "Next" string wasn't translated specifically for onboarding, so I'm just making use of the only existing one for now.

ifosli: I think using correct strings should generally be release blockers, though there can be exceptions. Have you confirmed if this (and the other missing strings) should count as release blockers? And if they should, do you need help making sure they get translated asap?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -70,33 +73,30 @@ struct CallToActionView: View {
...
+      LocalizedStringKey(stringLiteral: "Navigate to the next oboarding view.")
     case .enableNotifications:
-      "FPO: Enable Push Notifications."
+      LocalizedStringKey(stringLiteral: "Enable Push Notifications.")
     case .allowTracking:
-      "FPO: Allow App Tracking."
+      LocalizedStringKey(stringLiteral: "Allow App Tracking.")
     case .loginSignUp:
-      "FPO: Go to the Login and Sign Up screen."
+      LocalizedStringKey(stringLiteral: "Go to the Login and Sign Up screen.")
```
ifosli: Nit: If we're planning to release this feature before getting string translations in (I'm hoping we're not but just in case) have you considered using the button strings as accessibility strings (temporarily)? I think that would make more sense than sudden english (that probably ends up getting very mispronounced).

Kickstarter-iOS/Features/Onboarding/Views/OnboardingCTAView.swift
```diff
@@ -70,33 +73,30 @@ struct CallToActionView: View {
...
     }
   }
 
-  // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
+  // TODO: Add accessibility translations [mbl-2418]
   private func accessibilityHint(for item: OnboardingItem) -> LocalizedStringKey {
     switch item.type {
     case .welcome, .saveProjects:
-      "FPO: Navigate to next oboarding view."
+      LocalizedStringKey(stringLiteral: "Navigate to the next oboarding view.")
```
ifosli: "onboarding" not "oboarding"

Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
```diff
@@ -43,8 +43,8 @@ struct OnboardingItemView: View {
       ResizableLottieView(onboardingItem: self.item, isVisible: true)
         .frame(maxWidth: .infinity)
         .padding(.top, Constants.lottieViewTopPadding)
-        // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
-        .accessibilityLabel(Text("FPO: An example animation for the \(self.item.type) feature."))
+        // TODO: Add accessibility translations [mbl-2418]
+        .accessibilityLabel(Text("An example animation for the \(self.item.type) feature."))
```
ifosli: Hmm, I have concerns about this string. Will `self.item.type` be translated, too? If there are few enough types here, it'd probably be better to do separate strings for each one.

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -33,42 +33,42 @@ public struct OnboardingView: View {
...
-                onSecondaryTap: { self.handleSecondaryTap(for: item) }
-              )
-              .accessibilityElement(children: .contain)
-              // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
-              .accessibilityHint("FPO: Tap the next button to continue.")
-              .transition(.asymmetric(
-                insertion: .move(edge: .trailing).combined(with: .opacity),
-                removal: .move(edge: .leading).combined(with: .opacity)
-              ))
+    GeometryReader { _ in
```
ifosli: What's the benefit of adding a GeometryReader here, when it's unused?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -33,42 +33,42 @@ public struct OnboardingView: View {
...
-      Image(OnboardingStyles.backgroundImage)
-        .resizable()
-        .scaledToFit()
-        .ignoresSafeArea()
-
-      VStack {
-        self.ProgressBarView()
-          .accessibilityElement(children: .combine)
-          // TODO: Update hardcoded strings with translations [mbl-2417](https://kickstarter.atlassian.net/browse/MBL-2417)
-          .accessibilityLabel("FPO: Onboarding Progress Bar")
```
ifosli, line 48: Nit: Add a todo to add this back in, once it's translated? At least if the accessibility label will actually provide useful info. ("Progress bar" isn't particularly useful if we don't know what the progress is; something like "Page _ of _" would be more useful. Or this element can just be invisible to voiceover if we don't have a useful label for it.)

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -33,42 +33,42 @@ public struct OnboardingView: View {
...
+            ForEach(Array(self.onboardingItems.enumerated()), id: \.element.id) { index, item in
+              if index == self.currentIndex {
+                OnboardingItemView(
+                  item: item,
+                  progress: self.progress,
+                  onPrimaryTap: { self.handlePrimaryTap(for: item) },
+                  onSecondaryTap: { self.handleSecondaryTap(for: item) }
+                )
+                .accessibilityElement(children: .contain)
+                .accessibilityHint(Strings.project_checkout_navigation_next())
```
ifosli: This string doesn't contain the same info as the "FPO". I'm not sure it's useful as an accessibility hint at all like this. I'm not sure, though. Either way, can you add a "todo" to replace it with the correct string once it's ready?

ifosli (approved):
I have concerns about the missing strings and some minor changes that need to happen. Marking this as approved bc I don't think I need to re-review, though. If you consider this pr to unblock translations for the release, can you confirm with Tan before you head out? I'd expect the release to still be blocked on missing translations (and if that's the case, we should try to get them translated asap).

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
scottkicks: ah crap. that's a missed change from a merged conflict.

Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
```diff
@@ -26,25 +26,21 @@ struct OnboardingItemView: View {
...
-          .accessibilityHint(Text("Onboarding screen"))
+          .accessibilityLabel(Text(self.accessibilityLabel(for: self.item)))
 
         Text(self.item.subtitle)
           .font(Font(OnboardingStyles.subtitle))
           .lineLimit(4)
           .multilineTextAlignment(.center)
           .frame(maxWidth: .infinity)
-          .accessibilityLabel(Text(self.item.subtitle))
+          .accessibilityLabel(Text(self.accessibilityLabel(for: self.item)))
```
amy-at-kickstarter, line 36: Follow-up suggestion: you could write a custom view modifier to shorten this and combine all the accessibility logic for `OnboardingItemView`

amy-at-kickstarter (approved):

PR #2567, opened 2025-07-30
Default The Onboarding Feature Flag To True On App Launch

scottkicks:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests

### DIFF
--- a/Kickstarter-iOS/AppDelegateViewModel.swift
+++ b/Kickstarter-iOS/AppDelegateViewModel.swift
@@ -782,11 +782,7 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
         /// We don't want to request authorzation in the onboarding flow unless they've tapped the "all tracking" CTA.
         let hasSeenOnboarding = AppEnvironment.current.userDefaults.hasSeenOnboarding == true
 
-        if featureOnboardingFlowEnabled() == true {
-          return applicationIsActive && hasSeenOnboarding
-        }
-
-        return applicationIsActive && !hasSeenOnboarding
+        return applicationIsActive && hasSeenOnboarding
       }
       .map { _ in AppEnvironment.current.appTrackingTransparency }
       .map { appTrackingTransparency in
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -782,11 +782,7 @@ public final class AppDelegateViewModel: AppDelegateViewModelType, AppDelegateVi
         /// We don't want to request authorzation in the onboarding flow unless they've tapped the "all tracking" CTA.
         let hasSeenOnboarding = AppEnvironment.current.userDefaults.hasSeenOnboarding == true
 
-        if featureOnboardingFlowEnabled() == true {
-          return applicationIsActive && hasSeenOnboarding
-        }
-
-        return applicationIsActive && !hasSeenOnboarding
```
scottkicks, line 789: There's actually still a race condition that occurs even after forcing the flag value to `true` in RemoteConfigFeature+Helpers.

when the app first boots up, `featureOnboardingFlowEnabled()` returns `true` because of the [new logic added in this PR](https://github.com/kickstarter/ios-oss/pull/2567/files#:~:text=if%20feature%20%3D%3D,%7D), This is because, until the the app has fully finished launching, the remoteConfigClient is technically uninitialized (`nil`), but once the app has loaded up, remoteConfigClient gets configured, ignoring the new logic, `featureOnboardingFlowEnabled()` returns `false` for a few seconds until eventually returning the correct dashboard value ([currently `true` for the beta](https://console.firebase.google.com/u/0/project/kickstarter-ios-beta/config/env/firebase))

My solution is to remove the flag from here and rely on the `hasSeenOnboarding` value in user defaults to determine when to attempt to show the ATT dialog, as well as Apples internal logic to only show the dialog once per device.

Hopefully that makes sense. This is a weird one.

Kickstarter-iOS/AppDelegateViewModelTests.swift
scottkicks, line 1: Cleaned up some tests around the ATT dialog along the way.

scottkicks:
closing in favor of this approach instead https://github.com/kickstarter/ios-oss/pull/2564

PR #2572, opened 2025-08-04
MBL-2653: Add spacing constants to design system

amy-at-kickstarter:
# 📲 What

Add `Spacing.unit_xx` and `Dimensions` to the app.

# 🤔 Why

This creates code that is equivalent to the `Spacing/xx` and `Dimensions` constants in Figma. Having a direct mapping to Figma will make it easier for us to follow the new design system when we write code.

# 🛠 How

I landed on naming these as static constants on a global struct, like
```swift
Spacing.unit_04
```

~~I went back and forth on the interface for `Spacing`. What I originally wanted was an `enum` or `struct` so we could access it like~~
```swift
Spacing.01
```
~~However, you can't start a Swift variable or case name with a number. The closest I could get (readability-wise) would be to use an underscore, like~~
```swift
Spacing._01
```
~~but to my eyes, that makes the constant look _private_.~~

~~Instead, I flattened these out into global constants, so they're simply~~
```swift
Spacing_01
```

~~Suggestions or dissenting opinions are welcome on naming these - since we're going to look at them _everywhere_ I want to get the convention correct!~~


ifosli (approved):
I'm fine with this as it is now, but I agree with you that an enum would be nice. I don't have any great ideas for how to name that, though. I think if we were doing an enum, I'd add an extra word in there (ex `Spacing.core_01`) but I could not think of a word that I definitely like better than just doing global constants.

stevestreza-ksr:
These should be definitely encapsulated in _something_, even if just static accessors like:

```swift
public enum SpacingConstants {
    public static let spacing01: CGFloat = 4
    ...
}
```

Right now they are adding stuff to the global public scope, which should be avoided, especially since we don't really want people accessing these directly without good reason.

stevestreza-ksr (changes requested):
These should be definitely encapsulated in _something_, even if just static accessors like:

```swift
public enum SpacingConstants {
    public static let spacing01: CGFloat = 4
    ...
}
```

Right now they are adding stuff to the global public scope, which should be avoided, especially since we don't really want people accessing these directly without good reason.

amy-at-kickstarter:
@stevestreza-ksr Ah, fair. Any opinions on naming? I rather liked the look of `Spacing._01` but I know it's...unorthodox.

amy-at-kickstarter:
I settled on `unit_01`. Seems clear and short enough.

stevestreza-ksr (approved):

PR #2573, opened 2025-08-04
MBL-2653: Replace hardcoded spacing constants with Spacing_xx and Dimensions

amy-at-kickstarter:
# 📲 What

Replace hardcoded spacing constants with their equivalent `Spacing_xx` or `Dimensions`.

# 🤔 Why

This will help us keep newer areas of the codebase in line with our design system, and it improves code clarity.

# 🛠 How

I focused on places where we defined `Constants`, and where we used values which were equivalent to `Spacing_01` through `Spacing_06`. I ignored files that still relied heavily on `Styles.grid`, since those are basically still in the old design system.

### DIFF
--- a/Library/Styles/BaseStyles.swift
+++ b/Library/Styles/BaseStyles.swift
@@ -3,7 +3,6 @@ import Prelude_UIKit
 import UIKit
 
 public enum Styles {
-  public static let cornerRadius: CGFloat = 4.0
   public static let minTouchSize: CGSize = CGSize(width: 44, height: 44)
   public static let projectPageLeftRightInset: CGFloat = Styles.grid(3)
   public static let projectPageTopBottomInset: CGFloat = Styles.grid(4)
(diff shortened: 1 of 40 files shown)


Library/Styles/BaseStyles.swift
```diff
@@ -155,7 +154,10 @@ public let separatorStyleDark: ViewStyle = { view in
 
  - returns: A view transformer that rounds corners.
  */
-public func roundedStyle<V: UIViewProtocol>(cornerRadius r: CGFloat = Styles.cornerRadius) -> ((V) -> V) {
+public func roundedStyle<V: UIViewProtocol>(
+  cornerRadius r: CGFloat = Dimension.CornerRadius
+    .small
+) -> ((V) -> V) {
```
ifosli: Nit: looks like the formatter did something kind of awkward to the line breaking here - do you want to see if you can make it more readable? If nothing else, I think combining line 158 and 159 would help

amy-at-kickstarter: Fixed!

amy-at-kickstarter:
> Looks good! Outside the scope of this pr, but do we have a plan for how we want to handle iPad in the new design system? I didn't think of it until just now, but it would be nice if iPad's sometimes bigger spacing was baked into the design system, instead of us having to do it on a per-view basis.

Good question! The best way to handle it in this system would be to add a semantic spacing to `Dimensions`, when that's necessary. Right now, everything here is a direct copy from Figma, but we can always bring that up when we get into iPad designs.

PR #2574, opened 2025-08-04
MBL-2657: Deprecate Styles.grid and Styles.gridHalf

amy-at-kickstarter:
# 📲 What

Mark `Styles.grid` and `Styles.gridHalf` as deprecated.

# 🤔 Why

These use an old design system constant for spacing (`6.0` instead of `4.0`). This reduces the tempation to use old spacing values from the old design system.


ifosli (approved):

stevestreza-ksr (approved):
See my comment in #2572 as it relates to these deprecation messages.

PR #2579, opened 2025-08-05
Onboarding Colors Should Be Static

scottkicks:
# 📲 What

Hardcodes the colors in OnboardingStyles so that they don't respond to Light/Dark Mode preferences.

# 🤔 Why

We intentionally want users to see the same UI for this flow.

# 👀 See

| Before (Dark Mode Enabled) 🐛 | After (Dark Mode Enabled) 🦋 |
| --- | --- |
| old-broken | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-08-05 at 11 19 41 |

# ✅ Acceptance criteria

- [x] Onboarding flow UI element colors remain the same regardless of light mode or dark mode settings


Kickstarter-iOS/Features/Onboarding/OnboardingStyles.swift
```diff
@@ -3,18 +3,27 @@ import Library
 import SwiftUI
 
 enum OnboardingStyles {
```
amy-at-kickstarter: Instead of hardcoding the color values, you should use [overrideUserInterfaceStyle](https://developer.apple.com/documentation/uikit/uiview/overrideuserinterfacestyle) to hard-code this to light mode.

That way, you can continue to use design system styles and colors.

scottkicks: Oh i didn't realize UIHostingController had this. Great call.

amy-at-kickstarter (approved):

ifosli (approved):

PR #2580, opened 2025-08-05
New Design System Button Disabled State Button Title Color

scottkicks:
…ation

# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests


Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -63,7 +63,7 @@ extension UIButton {
 
         var newConfig = config
         newConfig.font = styleConfig.font
-        newConfig.foregroundColor = self.tintColor
```
scottkicks, line 66: the foregroundColor gets set in `func updateColors(...)` already and is called before this removed line on `line: 56`.

`newConfig.foregroundColor = self.tintColor` was overwriting that.

amy-at-kickstarter (approved):
Good find!

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
ifosli: Looks good! I did notice we also reference the tint color in line 71 - I'm not sure how often that's used, but I think that might have similar problems due to not taking the button state into account. Up to you if you want to/have time to check on it, though!

ifosli (approved):

PR #2586, opened 2025-08-07
Fix slow type inference in FacebookResetPasswordViewController

amy-at-kickstarter:
# 📲 What

Fix compile warning about `bindStyles` in  `FacebookResetPasswordViewController` being very slow to type check.

This changes makes the method compile in 27.23ms instead of 20+ seconds.

# 🤔 Why

I've been doing a little background exploration to understand if deprecating Prelude would speed up our builds - and if so, how _much_ and _why_.

This PR came out of that exploration; I was [reading more about the Swift type checker](https://danielchasehooper.com/posts/why-swift-is-slow/?continueFlag=7ec373c00bdf01cb02736c00eeaadcc8) and wanted to understand why this function was so slow. My best guess is that the type checker solver has a hard time resolving these key paths (with very common names) to a specific type; defining them as `Foo.Lens.bar` instead of `\.bar` simplifies that for the compiler. When they're appended together as one long expression, the compiler panics.

This change has a trivial impact, but I thought it was neat - since I fixed it, I figured I might as well ship the fix!


amy-at-kickstarter:
> Alternatively just remove all the Prelude ;)

SOON 😉

scottkicks (approved):

PR #2589, opened 2025-08-11
[MBL-2715] Remove Subtitle Line Limit

scottkicks:
# 🤔 Why

German translations have some longer text that we don't want truncated.

I think the `.lineLimit` property was left over from trying to maintain the view height and button placement a few PRs ago.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| 179d251d-347e-4669-8311-028396645c6e | Simulator Screen Recording - iPhone 15 Pro 17 5 - 2025-08-11 at 14 04 42 |

### DIFF
--- a/Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
+++ b/Kickstarter-iOS/Features/Onboarding/Views/OnboardingItemView.swift
@@ -30,7 +30,6 @@ struct OnboardingItemView: View {
 
         Text(self.item.subtitle)
           .font(Font(OnboardingStyles.subtitle))
-          .lineLimit(4)
           .multilineTextAlignment(.center)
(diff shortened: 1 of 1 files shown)


ifosli (approved):
Looks nice! Just checking: if this font size can grow dynamically, have you checked what the view looks like for really large font sizes as well? I'd be nice to check if this change causes any layout issues for large text (though not a blocker for this pr; I think if there's layout issues for large text, 4 lines should be enough to have spotted them already).

scottkicks:
> have you checked what the view looks like for really large font sizes as well?

Yeah! I checked this and all text looks good at larger sizes. The status bar gets slight pushed up, but it's still visible and the close icon still looks good 👍

stevestreza-ksr (approved):

PR #2592, opened 2025-08-18
Include GraphAPI and GraphAPITestMocks targets using SPM

amy-at-kickstarter:
# 📲 What

* Add the `GraphAPI` and `GraphAPITestMocks` targets to the project as a local swift package, instead of as individual targets.
* Link the Apollo library statically, instead of dynamically
* Link `GraphAPI` and `GraphAPITestMocks` statically, instead of dynamically

Screenshot 2025-08-19 at 9 33 52 AM

# 🤔 Why

* Using SPM means we have fewer targets to configure and manage in XCode. We had some signing issues last release because of how these targets were configured - my hope is that fewer targets = fewer places we can make that kind of mistake.
* Using SPM also means all these files and folders are added automatically, so you no longer have to individually add generated files.
* We link basically all of our dependencies statically and not dynamically, so this is more consistent with the rest of the app.

# 👀 How

I created a beta distribution build to test this - the build builds and runs!


nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KsApi/models/graphql/adapters/WatchProjectResponseEnvelope+WatchProjectMutation.DataTests.swift
```diff
@@ -7,7 +7,7 @@ import XCTest
 final class WatchProjectResponseEnvelope_WatchProjectMutationTests: XCTestCase {
   func test_envelopeFrom() {
     let mock = Mock<GraphAPITestMocks.Mutation>()
-    mock.watchProject = Mock<GraphAPITestMocks.WatchProjectPayload>()
```
scottkicks, line 10: Why do we need these changes?

scottkicks (approved):
LGTM. Excited to see how this impacts!

Just one clarifying change question

KsApi/models/graphql/adapters/WatchProjectResponseEnvelope+WatchProjectMutation.DataTests.swift
amy-at-kickstarter: Ah! So it turns out there's a bug with the generator - or maybe with our GraphQL schema. We have two mutations, `WatchProjectMutation` and `UnwatchProjectMutation`, both of which return a `watchProject` object. However, they each return two different _types_. Every time we re-generate our GraphQL objects, it switches which type it defines, and one of these tests breaks 🤦‍♀️ 

The tests aren't terrible useful, so one solution would be to delete them. I'm not sure we can fix the backend issue.

ifosli (approved):

PR #2594, opened 2025-08-19
Feat/adyer/mbl 2658/cleanup old ds code

amy-at-kickstarter:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests

### DIFF
--- a/Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemColors.swift
+++ b/Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemColors.swift
@@ -2,58 +2,6 @@ import Prelude
 import Prelude_UIKit
 import UIKit
 
-public enum DesignSystemColors: String {
-  // MARK: - Greens
-
-  case create100
-  case create300
-  case create500
-  case create700
-
-  // MARK: - Greys
-
-  case black
-  case support100
-  case support200
-  case support300
-  case support400
-  case support500
-  case support700
-  case white
-
-  // MARK: - Blues
-
-  case trust100
-  case trust300
-  case trust500
-  case trust700
-
-  // MARK: - Corals
-
-  case celebrate100
-  case celebrate300
-  case celebrate500
-  case celebrate700
-
-  // MARK: - Functional
-
-  case alert
-  case cellSeparator
-  case facebookBlue
-  case inform
-  case warn
-}
-
-extension DesignSystemColors {
-  public func load() -> UIColor {
-    UIColor(named: self.rawValue) ?? .white
-  }
-}
-
-public func adaptiveColor(_ style: DesignSystemColors) -> UIColor {
-  style.load()
-}
-
 public let verticalComponentStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
   stackView
     |> verticalStackViewStyle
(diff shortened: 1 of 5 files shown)


Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemColors.swift
```diff
@@ -2,58 +2,6 @@ import Prelude
 import Prelude_UIKit
 import UIKit
 
-public enum DesignSystemColors: String {
```
amy-at-kickstarter, line 5: These were only ever used in the design system demo screen. I replaced them with their equivalent `LegacyColors`. (Conveniently, they all had similar names, so I did it with one global find-and-replace regex.)

Kickstarter-iOS/Features/DesignSystem/Colors/DesignSystemColors.swift
```diff
@@ -76,61 +24,6 @@ public let alertStackViewStyle: StackViewStyle = { (stackView: UIStackView) in
     |> \.layer.cornerRadius .~ 6
 }
 
-// MARK: - Buttons
-
-public let adaptiveGreenButtonStyle = baseButtonStyle
```
amy-at-kickstarter, line 81: These were only used in the demo screen, and were effectively replaced by `KSRButtonStyle`.

Kickstarter-iOS/Features/DesignSystem/Controller/DesignSystemViewController.swift
```diff
@@ -24,14 +24,6 @@ final class DesignSystemViewController: UIViewController {
 
   private let buttonsLabel = UILabel()
   private let buttonsStackView: UIStackView = { UIStackView(frame: .zero) }()
-  private let primaryGreenButton = UIButton(type: .custom)
```
amy-at-kickstarter, line 27: Now that the new design system is on, these buttons and fonts were redundant. The design system screen was showing "old" and "new" buttons, and "old" and "new" fonts, but their appearance was identical!

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -42,7 +42,7 @@ extension UIButton {
 
     buttonConfiguration.background.cornerRadius = styleConfig.cornerRadius
 
-    buttonConfiguration.imagePadding = Styles.grid(1)
+    buttonConfiguration.imagePadding = 6.0
```
ifosli, line 45: I think we may want to revisit the padding in the designs (both there and the `grid(2) -> unit_3`). I know when I've added UI components in the past, the figma has often had spacing in multiples of `4`, but I've converted them to multiples of `6` to work with our design system instead. I'm okay with this for this pr, but I think it'd be useful to go through our components that we're definitely keeping during the app re-write to make sure spacing is what we want it to be in the long run.

ifosli (approved):
Nice! I accidentally left a comment outside of this review, but for the "updating old spacing to new spacing" part, I think it'd be useful to revisit designs instead of matching values exactly (unless you're intentionally matching old spacing for now?)

scottkicks (approved):
👍

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
amy-at-kickstarter: I spoke with Alison about this specifically - apparently `6.0` is a _very old_ design spacing constant, and no longer in use anywhere else at the company. All the design system spacing values now use `4.0`. 

I did experiment with completely deleting `Styles.grid` by replacing it with equivalent values (like you suggest, `.grid(2)` -> `Spacing.unit_03`), but I ended up with an odd mix of `Styles.grid` and `Spacing` in the same file, which I liked even less.

For now, my approach is "old code that uses old spacing can continue to do so; new code should always use new spacing." I agree that we should clean it up anywhere we keep old components; my assumption is those will end up being refreshed to match the new design system anyways!

PR #2595, opened 2025-08-20
MBL-2658: Remove ColorResolver abstraction

amy-at-kickstarter:
# 📲 What

Refactor `AdaptiveColor` to cache dynamic colors, so we can remove the `ColorResolver` abstraction.

# 🤔 Why

I added the `ColorResolver` abstraction to fix a unit testing problem: dynamically-provided `UIColor`s can't be compared with `==`, so many of our view model tests were breaking. The `ColorResolver` replaced those dynamic `UIColor` objects with static ones in test environments.

However, when I went to pull the design system into its own module, this abstraction meant our color code was dependent on `AppEnvironment` via the `ColorResolver`.

To clean up that dependency, I did some experimentation and found that you _can_ compare dynamically provided `UIColor`, but only with pointer equality. Since we create each color only once (via `static let someColor = ...`), caching the dynamically generated color fixes the tests, with no extra abstraction needed.


ifosli (approved):
Very nice!

PR #2598, opened 2025-08-21
MBL-2658: Move design system code into new package named KDS

amy-at-kickstarter:
# 📲 What

Move our all our new design system code into a local Swift package, named KDS.

# 🤔 Why

* Pulling out domain-specific code into separate packages should improve our build times, by flattening our (Very vertical) dependency structure.
* Modularizing code forces us to think more carefully about the dependencies it requires, ultimately forcing us to make smarter decisions about class/method visibility and dependencies. 
* The design system is a neatly packaged set of features, so it is a great first candidate for this kind of modularization.
* SPM packages are nifty and mean we don't need to individually define files in our Xcode project.

# 🛠 How

I learned some interesting things throughout this process - these are called out specifically in code comments.

# 👀 See
Screenshot 2025-08-21 at 5 59 25 PM

### DIFF
--- a/KDS/Package.swift
+++ b/KDS/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.4
+// swift-tools-version: 6.0
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription


KDS/Package.swift
```diff
@@ -0,0 +1,36 @@
...
+        .process("Fonts/Inter-VariableFont.ttf")
+      ]
+    ),
+    .testTarget(
+      name: "KDSTests",
+      dependencies: ["KDS"]
+    )
+  ],
+  swiftLanguageModes: [
+    .v5
```
amy-at-kickstarter, line 39: Specifically calling out this package as Swift 5.0 and iOS 16.0.

KDS/Sources/KDS/Colors/CoreColor.swift
```diff
@@ -87,7 +87,7 @@ public enum CoreColor: Int {
   case blue_10 = 0x050543
 }
 
-extension UIColor {
+public extension UIColor {
```
amy-at-kickstarter: I went down a really deep rabbit hole of figuring out which methods (specifically around color generation) should be public, and which should not be.

It started when I made `UIColor.hex` internal to the `KDS` package. This was actually really nifty, because it automatically flagged the handful of places in the code where we're manually creating colors off of the design system! But then I went pretty deep trying to figure out how to handle these one-off color cases...and eventually I came around to just making the method public again.

I wanted to highlight this because it _is_ a very powerful tool for enforcing code conventions - you can be really intentional about only exposing a small surface area from your package/library, in a way that we typically don't bother to do.

KDS/Sources/KDS/Fonts/CustomFont.swift
```diff
@@ -8,15 +9,34 @@ public protocol CustomFont {
...
 
+extension CustomFontAccessible {
+  public static func registerFont() {
+    for fontName in self.fontFileNames {
+      guard let fontURL = Bundle.module.url(forResource: fontName, withExtension: "ttf") else {
+        continue
+      }
+
+      var error: Unmanaged<CFError>?
+      CTFontManagerRegisterFontsForURL(fontURL as CFURL, .process, &error)
```
amy-at-kickstarter: Spiffy trick I found for registering fonts. This meant we can remove the `Fonts` section from `Info.plist`.

KDS/Sources/KDS/Fonts/CustomFont.swift
```diff
@@ -8,15 +9,34 @@ public protocol CustomFont {
...
+      assert(error == nil, "Error registering font: \(String(describing: error))")
```
amy-at-kickstarter: I went down another rabbithole making sure asserts work as we expect - they do, in fact, run in debug builds, but are stripped from release builds.

It would be good to verify if this is because SPM inherits build flags from the target building the SPM module (in this case, `Library`), or if SPM automatically adds the flag to strip assertions.

You can add custom swift build flags to an SPM module, so it's not totally clear to me how the inheritance of flags works.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -43,6 +44,8 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     UIImageView.appearance(whenContainedInInstancesOf: [UITabBar.self])
       .accessibilityIgnoresInvertColors = true
 
+    InterFont.registerFont()
```
amy-at-kickstarter: I believe we can now add this to `Kickstarter-Framework-Tests` and get our screenshot tests all building with `Interfont`.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -2,6 +2,7 @@ import AppboyKit
 import FBSDKCoreKit
 import Firebase
 import Foundation
+import KDS
```
amy-at-kickstarter, line 6: Downside of modularization: 300 files that just add `import KDS`.

Kickstarter-iOS/Features/FacebookResetPasswordViewController/Controller/FacebookResetPasswordViewController.swift
```diff
@@ -252,14 +253,3 @@ private let textFieldStyle: TextFieldStyle = { textField in
     |> UITextField.lens.layer.borderWidth .~ 1
     |> UITextField.lens.returnKeyType .~ .done
 }
-
-private let savePasswordButtonStyle: ButtonStyle = { button in
```
amy-at-kickstarter, line 256: Unused, deleted because it required `mixLighter` which I made `internal` to KDS.

Kickstarter-iOS/Info.plist
```diff
@@ -137,18 +145,5 @@
...
 	</array>
-	<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_ANALYTICS_STORAGE</key>
-	<false/>
-	<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_STORAGE</key>
-	<false/>
-	<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_USER_DATA</key>
-	<false/>
-	<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_PERSONALIZATION_SIGNALS</key>
-	<false/>
-	<key>UIAppFonts</key>
```
amy-at-kickstarter: See note above about fonts. Perhaps we should leave this line, though?

Kickstarter.xcodeproj/xcshareddata/xcschemes/Library-iOS.xcscheme
```diff
@@ -57,6 +57,16 @@
                ReferencedContainer = "container:Kickstarter.xcodeproj">
             </BuildableReference>
          </TestableReference>
+         <TestableReference
```
amy-at-kickstarter, line 60: This is one weird bit of SPM I can't quite figure out. I can add the KDS scheme, and we have a target `KDSTests`, but I can't just _run_ `KDSTests`. Instead, I had to add them to another schema, in this case `Library-iOS`. I think this is because Xcode assumes an SPM module is an _external_ dependency, and why would you want to test an _external_ dependency all the time?

I'm not sure if there's a better workaround. Maybe using "Test Plans"? Thoughts welcome.

Library/Styles/PLOTSemanticColors.swift
```diff
@@ -1,28 +1,29 @@
+import KDS
 import UIKit
 
-// This extension contains custom semantic colors that are not part of the official Figma design set.
+// This extension contains custom semantic colors for PLOT that are not part of the official Figma design set.
```
amy-at-kickstarter, line 4: These all are related to PLOT, so I rescoped them and left them in Library instead of KDS.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

KDS/Package.swift
ifosli: This seems inconsistent with the `swift-tools-version: 6.0` at the top of the file (commented out). Can you explain why they should both be here or delete one of them?

Kickstarter.xcodeproj/xcshareddata/xcschemes/Library-iOS.xcscheme
```diff
@@ -78,6 +88,13 @@
             ReferencedContainer = "container:Kickstarter.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
+      <EnvironmentVariables>
```
amy-at-kickstarter, line 91: When you call `Bundle.module`, it looks for bundles inside the framework (like `$(BUILT_PRODUCTS_DIR)/Library.framework/KDS_KDS.bundle`). However, when you're building the framework alone (i.e. for tests), the bundle isn't copied there. By setting this environment variable in Debug, we can tell the compiler to look for the bundle in `$(BUILT_PRODUCT_DIRS)/KDS_KDS.bundle` instead. 

This is necessary to keep our tests from crashing when you call `Bundle.module`.
Screenshot 2025-08-28 at 10 54 12 AM

KDS/Package.swift
amy-at-kickstarter: Fixed!

amy-at-kickstarter: Just kidding, that breaks on CircleCI! Somewhat confusing why there are two definitions in this file, but looks like it's necessary.

Kickstarter.xcodeproj/xcshareddata/xcschemes/Library-iOS.xcscheme
amy-at-kickstarter: OK, I did figure this out. I had to add a schema _to KDS_, not _to Kickstarter_. You can now run the KDS tests from the Kickstarter project using CMD+U.

I still left this added to the Library schema, because I don't want to build a whole new CircleCI job just for these tests.

KDS/Package.swift
ifosli: Weird! Thanks for trying though

ifosli (approved):

PR #2599, opened 2025-08-21
Braze upgrade

ifosli:
# 📲 What

Upgrade braze 🎉 

# 🤔 Why

We've needed to do this upgrade for a while...

## Braze automatic vs manual integration
In the newest version of braze, there's two ways to handle push notifications/in-app messages; manual and automatic. I first attempted a manual integration, since that would've allowed us to keep using a lot of our existing code (and would've meant I did this in an extra step, using the `BrazeKitCompat` library). However, for some reason that I wasn't able to figure out, the manual integration didn't work. It compiled and I inspected the braze object in the debugger and it _claimed_ to be a manual integration, but our manual tracking code was never run. Instead of spending more time trying to figure out what was going on, I set up the automatic integration immediately instead.

The automatic integration has the benefit of requiring less code, so hopefully it'll be easy to maintain.

# 🛠 How

This ended up not being a terrible amount of lines of code changed, so feel free to review all commits at once, if you prefer. I've split this into 5 commits:
1. Delete old braze tracking code (no longer needed in any version with the automatic integration)
2. Bulk of the upgrade (this compiles but doesn't work perfectly)
3. Update how we handle braze deeplinks (we need to use a delegate function; otherwise braze will handle deeplinks for us by opening them in safari 🤦 )
4. Include ids in the segment identify events (this matches behavior in the old version of segment)
5. Re-add in-app message support (uses the same delegate as push deeplinks)

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2664)

# ✅ Acceptance criteria

- [x] Segment debugger track, identify, and lifecycle events contain the same information
- [x] Braze push notifications and inapp messages work (deeplinks work as expected, and interactions are tracked)
- [ ] aws notifications still work (deeplinks are handled correctly; I haven't tested this on the latest code; I'll re-rest once I get another notification in prod. I may merge this pr before that happens, but I'll make sure it gets tested before the next release)

# ⏰ TODO

I added two tickets for follow-up work here:
- [Investigate why braze push notification registration doesn't work immediately](https://kickstarter.atlassian.net/browse/MBL-2744)
- [Revisit decisions I made during this migration in order to match behavior as much as possible](https://kickstarter.atlassian.net/browse/MBL-2742)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -146,7 +155,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.viewModel.outputs.registerPushTokenInSegment
       .observeForUI()
       .observeValues { token in
-        Analytics.shared().registeredForRemoteNotifications(withDeviceToken: token)
```
ifosli, line 149: `Analytics.shared()` isn't really a thing in the new version of segment (they now support multiple instances). It's supposed to be backwards compatible, but they recommend writing our own `Analytics` extension that holds a reference to the analytics object to ensure that it works. Letting our `AppDelegate` keep a reference instead is simpler and much safer than relying on segment's default implementation.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -250,23 +259,47 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            configuration.push.automation = true
+            configuration.push.automation.requestAuthorizationAtLaunch = false
+            // TODO(MBL-2742): Change the logger level to `info` or `error` if it gets tedious.
+            configuration.logger.level = .debug
+          }
+        ) { [weak self] braze in
+          guard let self else { return }
+          braze.delegate = self
+
+          braze.inAppMessagePresenter = BrazeUI.BrazeInAppMessageUI()
```
ifosli: This line of code is needed for push notification support, and also requires that we use the `SegmentBrazeUI` import instead of `SegmentBraze`. And the way this package works, once we use `SegmentBrazeUI` in one target, that's the version we need across the app.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -24,6 +25,9 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
   var window: UIWindow?
   fileprivate let viewModel: AppDelegateViewModelType = AppDelegateViewModel()
   fileprivate var disposables: [any Disposable] = []
+  fileprivate var brazeSubscription: BrazeKit.Braze.Cancellable?
```
ifosli, line 30: We need to keep a reference to the subscription, or it gets cancelled. It's a custom braze type of cancellable, so it seemed easiest to me to store it as its own variable, but if anyone feels strongly about combining it with the `disposables` list above and has thoughts on how to do, I'm open to suggestions!

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -250,23 +259,47 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+          if let userId = AppEnvironment.current.currentUser?.id {
+            braze.changeUser(userId: String(userId))
+          }
+
+          self.brazeSubscription = braze.notifications
+            .subscribeToUpdates(payloadTypes: [.opened]) { [weak self] payload in
+              guard let self else { return }
+              // TODO(MBL-2742): Once migration is stable, revisit if braze should update the rootTabBar.
+              if let rootTabBarController = self.rootTabBarController {
+                rootTabBarController.didReceiveBadgeValue(payload.badge)
```
ifosli: I kept this code purely because it got called in our custom code when we handled any opened notification. I'd guess we only need it for aws notifications, but it feels safer to keep this same logic in the braze notification path for now.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -526,20 +543,12 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
...
-  func handleAppboyURL(_ url: URL?, from _: ABKChannel, withExtras _: [AnyHashable: Any]?) -> Bool {
-    self.viewModel.inputs.urlFromBrazeInAppNotification(url)
+// MARK: - BrazeDelegate
 
-    return true
+extension AppDelegate: BrazeDelegate {
+  func braze(_: BrazeKit.Braze, shouldOpenURL context: BrazeKit.Braze.URLContext) -> Bool {
+    // Custom handle all urls instead of letting braze try to open them.
+    self.viewModel.inputs.urlFromBrazeNotification(context.url)
+    return false
```
ifosli, line 565: Without this override, braze tries to open all our deep links in safari. This is the main reason why I jumped straight to the new version of braze instead of using `BrazeKitCompat`; even with `BrazeKitCompat`, if we need to intercept urls here.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -80,7 +77,7 @@ public protocol AppDelegateViewModelInputs {
   func showNotificationDialog(notification: Notification)
 
   /// Call when Braze in-app notifications send a valid URL.
-  func urlFromBrazeInAppNotification(_ url: URL?)
+  func urlFromBrazeNotification(_ url: URL?)
```
ifosli, line 80: Renamed because this now gets called for push notifications and in-app messages, instead of us needing to parse the push notification payload

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -146,7 +155,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.viewModel.outputs.registerPushTokenInSegment
       .observeForUI()
       .observeValues { token in
-        Analytics.shared().registeredForRemoteNotifications(withDeviceToken: token)
+        self.analytics?.registeredForRemoteNotifications(deviceToken: token)
```
scottkicks, line 160: Would it be helpful to add a debug log or guard here? Could make debugging easier if self.analytics is ever nil unexpectedly.

Kickstarter-iOS/AppDelegate.swift
scottkicks: Do we ever cancel this explicitly? (e.g. on app close or login/logout?) 

Maybe Braze handles that for us. If so then keeping this separate makes sense to me since the type differs. A comment explaining why it's separate, and maybe something about its lifecycle handling could be helpful.

Kickstarter-iOS/AppDelegateViewModel.swift
```diff
@@ -1,8 +1,8 @@
-import AppboyKit
 import KsApi
 import Library
 import Prelude
 import ReactiveSwift
+import UIKit
```
scottkicks, line 5: Do we need UIKit in this ViewModel?

scottkicks (commented):
🎉🎉🎉

Added just a could of comments. 
Would it be helpful to schedule a little QA party as an iOS team to get more AC testing in?

I also noticed that `NotificationService.swift` has some AppBoy references and properties. Do we still need those?

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -503,18 +528,10 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
   }
 
   func userNotificationCenter(
-    _ center: UNUserNotificationCenter,
+    _: UNUserNotificationCenter,
```
amy-at-kickstarter, line 540: Nit: should remove this now that it's unused.

Kickstarter-iOS/AppDelegate.swift
amy-at-kickstarter: Do we ever _want_ to allow this? For example, if it's a URL that's unsupported in some other way from the app?

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
```diff
@@ -10,7 +10,7 @@
             ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
             <ActionContent
                title = "Run Script"
-               scriptText = "# Type a script or drag a script file from your workspace to insert its path.&#10;bash &quot;$BUILT_PRODUCTS_DIR/Appboy_iOS_SDK_AppboyKit.bundle/Appboy.bundle/appboy-spm-cleanup.sh&quot;&#10;">
```
amy-at-kickstarter, line 13: Can this step be deleted completely?

amy-at-kickstarter (commented):
A couple nits/questions to start! 

One question that I have nowhere to put in the code: have you tested rich push notifications and the rich push notification target as part of this change?

Kickstarter-iOS/AppDelegate.swift
ifosli: I'm currently not canceling it explicitly. We could cancel it on logout, but this subscription doesn't affect whether or not the app receives push notifications; it just determines if braze calls this block when it handles them in our app. So I don't think it's necessary. On app close, the AppDelegate should be deallocated, which will stop our braze subscription, too. I'll add a comment for it, though!

Kickstarter-iOS/AppDelegate.swift
ifosli: Good question! But no, we don't need to; we already have a "open in safari" path in the app delegate that we fall back to in our custom deep link handling. It's possible that we could some kind of "if deep link, handle in app. else, let braze handle" logic here instead, but it would just make this more complicated than it needs to be.

Kickstarter-iOS/AppDelegate.swift
ifosli: That's a good idea. I'm going to add a note to the "attempt to fix push notification registration ticket" (https://kickstarter.atlassian.net/browse/MBL-2744) to add in a guard here (assuming this line stays)

Kickstarter-iOS/AppDelegateViewModel.swift
ifosli: Yeah, we do. It was implicitly included in the `AppboyKit` import before I deleted that, so after that got deleted, I needed to re-add UIKit explicitly. (It's because we're using types like `UIApplication` and `UIViewController` in this view model)

Kickstarter-iOS/AppDelegate.swift
ifosli: Sorry, explain what you mean? The method is still being used and it's a delegate method so it needs the same function signature (afaik); it's just the `center` param that's no longer being used.

ifosli:
> A couple nits/questions to start!
> 
> One question that I have nowhere to put in the code: have you tested rich push notifications and the rich push notification target as part of this change?

Yeah, I've tested rich push and it still works as it should :)

Kickstarter.xcodeproj/xcshareddata/xcschemes/Kickstarter-iOS.xcscheme
ifosli: Yeah; I left it because it looked like it might've come with the project by default, but I'll delete it completely!

ifosli:
> 🎉🎉🎉
> 
> Added just a could of comments. Would it be helpful to schedule a little QA party as an iOS team to get more AC testing in?
> 
> I also noticed that `NotificationService.swift` has some AppBoy references and properties. Do we still need those?

Re: QA party: My plan here is to create a testflight build (in addition to the beta build) once these changes are approved/merged. Becca is helping a lot with setting up braze stuff for testing. I was planning to have us do stuff async, but I'll check with her if she wants to do a QA party. Thanks for the suggestion!

NotificationService.swift is our rich push extension. We should upgrade this at some point to the latest recommended approach, but because it's not required for the main migration ([source](https://braze-inc.github.io/braze-swift-sdk/documentation/braze/appboy-migration-guide/#Rich-Push-Notifications)) I'm leaving it alone for now. Thanks for the reminder, though; I've added this to the "braze follow-up" ticket for now.

Kickstarter-iOS/AppDelegate.swift
amy-at-kickstarter: Ah, what I meant was,  "you should delete the `center` parameter completely since it's unused".

Kickstarter-iOS/AppDelegate.swift
amy-at-kickstarter: Huh, that's so odd that they made their own type. Is it documented at all?

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -250,23 +261,47 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+          guard let self else { return }
+          braze.delegate = self
+
+          braze.inAppMessagePresenter = BrazeUI.BrazeInAppMessageUI()
+
+          if let userId = AppEnvironment.current.currentUser?.id {
+            braze.changeUser(userId: String(userId))
+          }
+
+          self.brazeSubscription = braze.notifications
```
amy-at-kickstarter: Nit: there's quite a bit nested in this `BrazeDestination` callback. Can you pull it out into its own method, like

```swift
let brazeDestination = BrazeDestination(
  additionalConfiguration: { configuration in
    //etc
  }
) { [weak self] braze in
     self.configureBraze(braze)
}
```

Kickstarter-iOS/AppDelegate.swift
amy-at-kickstarter: Just to clarify - is Braze handling _all_ our notifications now, including AWS notifications? Was it always this way? It's a little confusing to me that they're mixed up like this.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -250,23 +261,47 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                rootTabBarController.didReceiveBadgeValue(payload.badge)
+              }
+            }
         }
 
-        Analytics.setup(with: configuration)
-        AppEnvironment.current.ksrAnalytics.configureSegmentClient(Analytics.shared())
+        configuration.add(plugin: brazeDestination)
+
+        let middleware = BrazeDebounceMiddlewarePlugin()
```
amy-at-kickstarter, line 266: Is the debounce middleware still required after this upgrade, or did Braze add its own debouncing?

Kickstarter-iOS/AppDelegate.swift
ifosli: Nope! At least according to the braze docs, they swizzle the notifications-related methods, but they only do something if it's a braze notification; if not, our normal delegate code gets called instead. So this specific method only gets called when a braze notification is opened (which is why I have that TODO to revisit if we need to include this logic at all). I have tested this but not super thoroughly, since aws notifications are hard to test.

Kickstarter-iOS/AppDelegate.swift
ifosli: Just here: https://braze-inc.github.io/braze-swift-sdk/documentation/brazekit/braze/notifications-swift.class/subscribetoupdates(payloadtypes:_:) or here's the cancellable itself: https://braze-inc.github.io/braze-swift-sdk/documentation/brazekit/braze/cancellable.

Kickstarter-iOS/AppDelegate.swift
ifosli: Still required :( (It's specifically required because of the segment braze integration)

ifosli:
> LGTM, assuming all of the manual testing checks out.
> 
> A few general thoughts:
> 
> 1. Can you make it more clear, in code, what we expect to be handled by Braze, and what we expect to handle ourselves or elsewhere? Based on this PR, it seems like Braze is acting as the entry point for _everything_, including our AWS notifications and all of our URL opens.
> 2. You already noted this in your follow-up ticket, but it would be nice to make sure we're taking advantage of everything new that we're getting. Specifically: the simplified code for rich push, and possibly cleaning up the debounce middleware.

Yeah, I added comments to the places where we handle notifications. Just to be clear, I'm not planning to do any additional manual testing before merging this; I want to test aws notifications in a testflight build, which I'm planning to create after this pr gets merged.

Kickstarter-iOS/AppDelegate.swift
ifosli: Done!

amy-at-kickstarter:
> > LGTM, assuming all of the manual testing checks out.
> > A few general thoughts:
> > 
> > 1. Can you make it more clear, in code, what we expect to be handled by Braze, and what we expect to handle ourselves or elsewhere? Based on this PR, it seems like Braze is acting as the entry point for _everything_, including our AWS notifications and all of our URL opens.
> > 2. You already noted this in your follow-up ticket, but it would be nice to make sure we're taking advantage of everything new that we're getting. Specifically: the simplified code for rich push, and possibly cleaning up the debounce middleware.
> 
> Yeah, I added comments to the places where we handle notifications. Just to be clear, I'm not planning to do any additional manual testing before merging this; I want to test aws notifications in a testflight build, which I'm planning to create after this pr gets merged.

Ah, by documenting in code, I meant add a stub method like
```swift
func shouldHandlePushNotificationSomehow(payload: PayloadType) -> Bool {
  return true
}
```

My thought was that will make it a little more future-proof if we ever need to extend or change the handling, even if the method isn't doing anything _more_ than a comment.

Kickstarter-iOS/AppDelegate.swift
stevestreza-ksr: We only really use disposables by letting memory management clean them up, so I'm fine with it being separate.

If you wanted to push it into the array, easiest way would just be to make it conform to the `Disposable` protocol from ReactiveSwift. https://github.com/ReactiveCocoa/ReactiveSwift/blob/a44317ce38b5bcce173b03f5d36cfdc939e1768b/Sources/Disposable.swift#L23

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -146,7 +157,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     self.viewModel.outputs.registerPushTokenInSegment
       .observeForUI()
       .observeValues { token in
-        Analytics.shared().registeredForRemoteNotifications(withDeviceToken: token)
+        self.analytics?.registeredForRemoteNotifications(deviceToken: token)
```
stevestreza-ksr, line 160: Weakify self here (it's an app delegate so it's not going anywhere but also there's a bunch of `[weak self]` in the rest of the file so may as well be consistent and safe)

stevestreza-ksr (approved):
Thanks for taking this on! 🎉

Kickstarter-iOS/AppDelegate.swift
ifosli: Good call! We were already using weak self in this block (but the strongified version was behind a rename). I've fixed it so we're consistently using the strongified version in this block (by renaming the strongified version).

Kickstarter-iOS/AppDelegate.swift
ifosli: Hmm, yeah, that would probably work; thanks for the suggestion! I'm going to just leave it separate since it sounds like we're all in agreement that it's fine as is, but I'll keep this kind of strategy in mind in the future.

PR #2601, opened 2025-08-26
MBL-2658: Load Interfont directly from Library bundle

amy-at-kickstarter:
# 📲 What

* Load `InterFont` directly from the Library bundle, instead of including it in the main app target.
* Re-record all screenshots to correctly use `InterFont` instead of `systemDefaultFont`.

# 🤔 Why

I've been working on pulling out our design system into its own SPM module. Our font code is an important part of the design system - so I wanted it to be included in the new KDS bundle, as well.

However, even though `InterFont` and `CustomFont` were part of our `Library-iOS` target, we had to include the font in every single app bundle for it to be available. Even then, it wasn't correctly being used in our screenshot tests.

This change makes `InterFont` more neatly encapsulated in Library, and then it will be a very easy port to move it over to KDS. As a side effect, this fixes the screenshot issue, so I went ahead and re-recorded them.

# 🛠 How

See comments inline.

I separated the code changes and the screenshot changes into separate commits, so they should be easier to review that way.

# ☑️ Tasks

- [x] Spot check screenshots
- [x] Test app with large accessibility sizes
- [x] Manually test that default font is loaded correctly if an error is thrown

# 👀 Follow ups

I filed a ticket, [MBL-2771](https://kickstarter.atlassian.net/browse/MBL-2771), to address font weight changes that came with using Interfont in the app.

[MBL-2771]: https://kickstarter.atlassian.net/browse/MBL-2771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -43,6 +43,8 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     UIImageView.appearance(whenContainedInInstancesOf: [UITabBar.self])
       .accessibilityIgnoresInvertColors = true
 
+    InterFont.registerFontIfUnregistered()
```
amy-at-kickstarter, line 46: I added some code that automatically registers `Interfont` the first time we try to access it. However, that's just to get our tests running - in production, it makes more sense to register it deterministically on launch.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -8,20 +9,45 @@ public protocol CustomFont {
...
+  static var fontFileURLs: [URL]? { get }
+  static var isRegistered: Bool { get set }
   var fontName: String { get }
   var boldFontName: String { get }
 }
 
 extension UIFont {
+  static func registerFontFromFileURL(_ fontURL: URL) throws {
+    var cfError: Unmanaged<CFError>?
+    let registered = CTFontManagerRegisterFontsForURL(fontURL as CFURL, .process, &cfError)
```
amy-at-kickstarter, line 22: This is the magic that lets us replace the font lines in `Info.plist`.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -8,20 +9,45 @@ public protocol CustomFont {
...
+    if !registered, let error = cfError?.takeRetainedValue() as? Error {
+      throw error
+    }
+  }
+
   static func customFont(with fontConfig: CustomFont & CustomFontAccessible, size: CGFloat? = nil) -> UIFont {
     let fontName = UIAccessibility.isBoldTextEnabled ? fontConfig.boldFontName : fontConfig.fontName
+
+    let fontConfigType = type(of: fontConfig)
+    if !fontConfigType.registerFontIfUnregistered() {
```
amy-at-kickstarter: This registers the font the first time you try to access it, if it wasn't already registered.

I found this was necessary to get our tests running smoothly. We call `UIFont.ksr_some_font_variant` in many `init` methods - which meant that it was frequently called before `XCTestCase.setUp`.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -8,20 +9,45 @@ public protocol CustomFont {
...
+      assert(
+        false,
+        "Tried to loading custom font \(fontName) but was unable to. Using system default font, instead."
+      )
       return self.defaultSystemFont(with: fontConfig, size: size)
     }
 
     let metrics = UIFontMetrics(forTextStyle: fontConfig.textStyle)
-    let finalFont = metrics.scaledFont(for: font)
+    let finalFont = metrics.scaledFont(for: font, compatibleWith: UITraitCollection.current)
```
amy-at-kickstarter, line 44: I had to add this to have accessibility sizes correctly reflected in the screenshot tests. The app itself works fine without this line.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -49,3 +75,34 @@ extension CustomFont {
...
+    }
+
+    self.isRegistered = true
+    return true
+  }
+
+  /// Registers the font if `self.isRegistered` is false.
+  /// - Returns: `true` if the font has been successfully registered and is ready for use; `false` if a backup font should be used.
+  public static func registerFontIfUnregistered() -> Bool {
+    if self.isRegistered {
```
amy-at-kickstarter, line 89: If you call `CTFontManagerRegisterFontsForURL` more than once for the same file/font name, it throws an error. Hence this extra guard.

Library/Styles/Fonts/InterFont.swift
```diff
@@ -75,8 +75,32 @@ public enum InterFont: CustomFont, CaseIterable {
...
+    get {
+      registeredInterfont
+    }
+    set {
+      registeredInterfont = newValue
+    }
+  }
+
+  public static var fontFileURLs: [URL]? {
+    guard let bundle = Bundle.init(identifier: "com.Library-iOS") else {
```
amy-at-kickstarter, line 91: This will be easy to swap over to the KDS bundle when I move this code.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -8,20 +9,39 @@ public protocol CustomFont {
...
+    if !registered, let error = cfError?.takeRetainedValue() as? Error {
+      throw error
+    }
+  }
+
   static func customFont(with fontConfig: CustomFont & CustomFontAccessible, size: CGFloat? = nil) -> UIFont {
     let fontName = UIAccessibility.isBoldTextEnabled ? fontConfig.boldFontName : fontConfig.fontName
+
+    let fontConfigType = type(of: fontConfig)
+    fontConfigType.registerFontIfUnregistered()
```
amy-at-kickstarter, line 33: This fixes fonts in our tests. We frequently call `UIFont.ksr_some_font` before `TestCase.setUp` occurs, so this was a necessary workaround.

Library/Styles/Fonts/CustomFont.swift
```diff
@@ -49,3 +69,34 @@ extension CustomFont {
     return Font(self.font(size: size))
   }
 }
+
+extension CustomFontAccessible {
+  /// Registers the font files so that they're available for use in the process.
+  /// - Returns: `true` if the font has been successfully registered and is ready for use; `false` if a backup font should be used.
```
ifosli: Looks like this return value here and in the comment below are outdated - maybe replace with a line about throwing an error if font registration fails? Or just delete it.

ifosli (approved):
Please fix those comment return types before merging, but otherwise, looks good!

Library/Styles/Fonts/CustomFont.swift
amy-at-kickstarter: Whoops, good catch!

PR #2602, opened 2025-08-26
[DO NOT MERGE] Next Gen Architecture SPIKE

scottkicks:
# 📲 What

The meat of this SPIKE is written up in [**this doc**](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/3864494081/Next+Gen.+Architecture+Test+SPIKE)

I suggest reviewing alongside the SPIKE doc. You should be able to go commit by commit, but I did expand on the apollo wrapper in the last commit so keep that in mind.

The goal of this is to prove we can build a new Swift 6 module locally and with CircleCI that:
- Imports GraphAPI and KsApi. ✅
- Uses SwiftUI + @Observable. ✅
- Does async/await GraphQL via Apollo (no Combine/ReactiveSwift). ✅
- Uses an AsyncSequence for “type-to-search” behavior. ✅
- Builds locally (Xcode 16 / Swift 6) ✅ and is CI-friendly ✅

# ⏰ TODO

- [ ] Add some tests around the Service and New AsyncApolloClient
- [ ] Put up a PR for the Apollo wrapper

### DIFF
--- a/NextGenFeature/Networking/AsyncApolloClient.swift
+++ b/NextGenFeature/Networking/AsyncApolloClient.swift
@@ -2,16 +2,20 @@
 import Foundation
 import GraphAPI
 
-/*
- An async/await wrapper around Apollo 1.x.
- */
-
-/// Using @unchecked because Apollo types aren’t fully Sendable-annotated.
+/// Minimal async/await wrapper for Apollo
 public final class AsyncApolloClient: @unchecked Sendable {
-  private let client: ApolloClientProtocol /// Existing Apollo protocol
+  private let client: ApolloClientProtocol
+  // One place to decide where Apollo invokes its completion.
+  private let callbackQueue: DispatchQueue
 
-  public init(client: ApolloClientProtocol) {
+  /// `callbackQueue` defaults to a background queue; pass `.main` if you truly want main.
+  public init(
+    client: ApolloClientProtocol,
+    /// using `.userInitiated` for lower latency/scheduling the work the user is waiting on  promptly. `.background` can sometimes get deprioritized.
+    callbackQueue: DispatchQueue = .global(qos: .userInitiated)
+  ) {
     self.client = client
+    self.callbackQueue = callbackQueue
   }
 
   // MARK: - Queries
@@ -19,22 +23,16 @@ public final class AsyncApolloClient: @unchecked Sendable {
   public func fetch<Query: GraphQLQuery>(
     _ query: Query,
     cachePolicy: CachePolicy = .fetchIgnoringCacheCompletely,
-    contextIdentifier: UUID? = nil,
-    queue: DispatchQueue = .main
+    contextIdentifier: UUID? = nil
   ) async throws -> GraphQLResult<Query.Data> {
-    try await withCheckedThrowingContinuation { (cont: CheckedContinuation<
-      GraphQLResult<Query.Data>,
-      Error
-    >) in
-      /// Kick off Apollo's callback-style request
+    try await withCheckedThrowingContinuation { cont in
       _ = self.client.fetch(
         query: query,
         cachePolicy: cachePolicy,
         contextIdentifier: contextIdentifier,
         context: nil,
-        queue: queue
+        queue: self.callbackQueue
       ) { result in
-        /// resume via async/await
         switch result {
         case let .success(value): cont.resume(returning: value)
         case let .failure(error): cont.resume(throwing: error)
@@ -47,21 +45,15 @@ public final class AsyncApolloClient: @unchecked Sendable {
 
   public func perform<Mutation: GraphQLMutation>(
     _ mutation: Mutation,
-    publishResultToStore: Bool = true,
-    queue: DispatchQueue = .main
+    publishResultToStore: Bool = true
   ) async throws -> GraphQLResult<Mutation.Data> {
-    try await withCheckedThrowingContinuation { (cont: CheckedContinuation<
-      GraphQLResult<Mutation.Data>,
-      Error
-    >) in
-      /// Kick off Apollo's callback-style request
+    try await withCheckedThrowingContinuation { cont in
       _ = self.client.perform(
         mutation: mutation,
         publishResultToStore: publishResultToStore,
         context: nil,
-        queue: queue
+        queue: self.callbackQueue
       ) { result in
-        /// resume via async/await
         switch result {
         case let .success(value): cont.resume(returning: value)
         case let .failure(error): cont.resume(throwing: error)


.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ test_job: &test_job
         # Keep simulator in sync with `Library/TestHelpers/TestCase.swift` and `Makefile`
         # iOS 17.5 is minimum supported runtime by circle-ci -> 16.0 (16A242d)
         # more info: https://circle-macos-docs.s3.amazonaws.com/image-manifest/v15048/manifest.txt
-        version: "17.5"
+        version: "18.0"
```
scottkicks, line 46: This is the change that CircleCI required to build successfully on `beta-dist`.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
```diff
@@ -398,6 +398,14 @@ public struct SearchFilterToggles {
...
   public var projectsWeLove: Bool
   public var following: Bool
+
+  /// Making a public initiializer for Next Gen Architecture testing.
+  public init(recommended: Bool, savedProjects: Bool, projectsWeLove: Bool, following: Bool) {
+    self.recommended = recommended
+    self.savedProjects = savedProjects
+    self.projectsWeLove = projectsWeLove
+    self.following = following
+  }
```
scottkicks, line 408: This is just so I could initialize the GraphQL query in `NextGenProjectSearchService`

Makefile
```diff
@@ -14,7 +14,7 @@ COMMIT ?= $(CIRCLE_SHA1)
 CURRENT_BRANCH ?= $(CIRCLE_BRANCH)
 
 ifeq ($(PLATFORM),iOS)
-	DESTINATION ?= 'platform=iOS Simulator,name=$(IPHONE_NAME),OS=$(IOS_VERSION)'
```
scottkicks, line 17: removing this also appeased CircleCI. It was requiring a destination OS version of 18 and just updating the config.yml wasn't enough apparently.

NextGenFeature/Networking/NextGenProjectSearchService.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    let result = try await self.apollo.fetch(GraphAPI.SearchQuery.from(discoveryParams: params))
+
+    let nodes = result.data?.projects?.nodes ?? []
+
+    return nodes.compactMap { node -> NextGenSearchResult? in
+      guard let name = node?.name else { return nil }
+
+      return NextGenSearchResult(id: UUID(), name: name)
+    }
```
scottkicks, line 51: This feels a little messy to me and could prob use a helper  that mroe effectively handles the optionals at least.

NextGenFeature/ViewModels/NextGenSearchViewModel.swift
```diff
@@ -0,0 +1,117 @@
...
+    self.bindTask?.cancel()
+
+    self.bindTask = Task {
+      for await text in stream {
+        try? await Task.sleep(nanoseconds: 250_000_000) // ~250ms
+
+        if Task.isCancelled { return }
+
+        await self.searchFor(query: text)
+      }
```
scottkicks, line 76: a handy way to loop over values that arrive over time from an AsyncSequence

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -154,13 +158,22 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
       .observeValues { [weak self] in
         guard let rootTabBarController = self?.rootTabBarController else { return }
 
-        let onboardingVC = UIHostingController(rootView: OnboardingView(viewModel: OnboardingViewModel()))
+        /// Replacing Onboarding screen with Next Gen Architecture Search View for testing.
+        let nextGenSearchVC = UIHostingController(
+          rootView: NextGenFeature.NextGenSearchView(
+            viewModel: NextGenFeature.NextGenSearchViewModel(
```
amy-at-kickstarter: My guess is that we want our default practice to be creating the view model with `@State` inside the `View`, and pass in dependencies like the `SearchService` as `@Environment` variables. (Or that it should be some kind of singleton, like `AppEnvironment.service`.

I did use this pass-in-the-view-model pattern a bunch in Search, but I had weird requirements for externally managed state.

Library/UseCases/SearchFilters/SearchFiltersUseCase.swift
amy-at-kickstarter: Probably worth noting that we're going to have to make many small decisions like this if we move to feature-based-modules! I've run into this a lot with the `KDS` module.

NextGenFeature/Networking/NextGenProjectSearchService.swift
amy-at-kickstarter: For next gen architecture, my hope is we just use straight-up Apollo objects, unless we have need for custom behavior or wrappers. You could try adding a query like

```
query NextGenSearch(
  $term: String,
  $sort: ProjectSort,
  $first: Int,
  $cursor: String
) {
  projects(
    term: $term,
    sort: $sort,
    //whatever other params you need for the demo...
    after: $cursor,
    first: $first
  ) {
    nodes {
       id
       name
    }
    totalCount
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```

 and then you'll get a custom object definition from `GraphAPI` like `NextGenSearchQuery.Node.Project` you can use in your view.

Kickstarter-iOS/AppDelegate.swift
scottkicks: So something like

```
@State private var viewModel = NextGenSearchViewModel(
    service: AppEnvironment.nextGen.projectSearchService
  )
```

amy-at-kickstarter: Yup!

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -154,13 +158,22 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
...
+              service: NextGenProjectSearchService(
+                apollo: AsyncApolloClient(client: GraphQL.shared.client)
+              )
+            )
+          )
+        )
         /// Onboarding flow should not respond to light/dark mode preference.
-        onboardingVC.overrideUserInterfaceStyle = .light
-        onboardingVC.modalPresentationStyle = .fullScreen
+        nextGenSearchVC.overrideUserInterfaceStyle = .light
```
amy-at-kickstarter, line 164: I assume there's nothing that would break with dark mode with the next-gen architecture - we already use dark mode with SwiftUI - but would be good to remove this just to be sure.

Kickstarter.xcodeproj/xcshareddata/xcschemes/NextGenFeature.xcscheme
```diff
@@ -0,0 +1,67 @@
...
+      buildConfiguration = "Release"
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      savedToolIdentifier = ""
+      useCustomWorkingDirectory = "NO"
+      debugDocumentVersioning = "YES">
+      <MacroExpansion>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "6075B3C22E5374A700CA4A46"
+            BuildableName = "NextGenFeature.framework"
```
amy-at-kickstarter, line 54: This is added as a regular target, correct? We talked a bit about creating features as SPM modules instead of targets - I think it might be impossible right now (since it will depend on Library/KsApi which are targets), but might be good to verify that.

NextGenFeature/Networking/AsyncApolloClient.swift
```diff
@@ -0,0 +1,72 @@
...
+
+  public init(client: ApolloClientProtocol) {
+    self.client = client
+  }
+
+  // MARK: - Queries
+  
+  public func fetch<Query: GraphQLQuery>(
+    _ query: Query,
+    cachePolicy: CachePolicy = .default,
```
amy-at-kickstarter: Right now we always us `.fetchIgnoringCacheCompletely`. I think we should stick with that as the default, and we can override it as necessary for fetches that are actually cacheable.

NextGenFeature/Networking/NextGenProjectSearchService.swift
```diff
@@ -0,0 +1,53 @@
...
+public protocol NextGenProjectSearchServicing {
+  func searchProjects(matching term: String) async throws -> [NextGenSearchResult]
+}
+
+/// Calls Apollo via our minimal async/await wrapper.
+@MainActor
+public struct NextGenProjectSearchService: NextGenProjectSearchServicing {
+  private let apollo: AsyncApolloClient
+
+  public init(apollo: AsyncApolloClient) {
```
amy-at-kickstarter, line 17: Thoughts on how we might handle this in the real app? Would we fold this into `Service`, or actually create a new client like `NextGenService`?

NextGenFeature/ViewModels/NextGenSearchViewModel.swift
```diff
@@ -0,0 +1,117 @@
...
+  public private(set) var isLoading: Bool = false
+  public private(set) var statusText: String = "Idle"
+
+  private let service: any NextGenProjectSearchServicing
+  private let textChangesStream: AsyncStream<String>
+  private let textChangesContinuation: AsyncStream<String>.Continuation
+  private var bindTask: Task<Void, Never>?
+  private var currentRequest: Task<Void, Never>?
+
+  public init(service: any NextGenProjectSearchServicing) {
```
amy-at-kickstarter: Noted this before, but instead of passing in data on `init`, this should be initialized by the `View` and then dependencies can be passed in on `onAppear` (or handled as singletons)

NextGenFeature/Views/NextGenSearchView.swift
```diff
@@ -0,0 +1,109 @@
...
+  static let horizontalPadding: CGFloat = 20
+  static let titleSubtitleSpacing: CGFloat = 12
+  static let verticalPadding: CGFloat = 20
+  static let verticalSpacing: CGFloat = 24
+}
+
+public struct NextGenSearchView: View {
+  @SwiftUI.Environment(\.dismiss) private var dismiss
+  @State private var viewModel: NextGenSearchViewModel
+  @Namespace private var animation
```
amy-at-kickstarter: This one is new to me, what is `@Namespace`?

NextGenFeature/Views/NextGenSearchView.swift
```diff
@@ -0,0 +1,109 @@
...
+
+  public init(viewModel: NextGenSearchViewModel) {
+    _viewModel = State(initialValue: viewModel)
```
amy-at-kickstarter: This pattern led to a lot of SwiftUI state bugs for me - instead, you should initialize this like `@State private var viewModel: NextGenSearchViewModel = NextGenSearchViewModel()`

NextGenFeatureTests/NextGenFeatureTests.swift
```diff
@@ -0,0 +1,6 @@
+@testable import NextGenFeature
+import Testing
+
+struct NextGenFeatureTests {
+  @Test func example() async throws {}
```
amy-at-kickstarter, line 5: Definitely curious to see what happens when you get testing spun up - did Apple give us everything we need to test async methods, or do we need helper methods?

NextGenFeature/Views/NextGenSearchView.swift
```diff
@@ -0,0 +1,109 @@
...
+          .font(.title3)
+          .foregroundColor(.secondary)
+          .padding(Constants.closeIconPadding)
+      }
+    }
+    .padding(.top, Constants.verticalPadding)
+    .padding(.horizontal, Constants.horizontalPadding)
+  }
+
+  private func SearchFieldView(vm: NextGenSearchViewModel) -> some View {
```
amy-at-kickstarter: It miiiight be more canonical to pass the `vm` down as an `@Environment` variable, instead.

NextGenFeature/Views/NextGenSearchView.swift
```diff
@@ -0,0 +1,109 @@
...
+    TextField(
+      "Search projects…",
+      text: Binding(
```
amy-at-kickstarter: Custom bindings gave me a lot of trouble in search. Is it possible to do this with something like

```swift
@State searchText = ""
///view code
.onChange(searchText) { newValue
  vm.inputs.searchTextChanged($0)
}
```

or does that break all the nice debouncing?

If we *do* need to push state up into the view model, maybe we need a custom macro to make wrapping these into a `Binding` easier.

Kickstarter-iOS/AppDelegate.swift
scottkicks: Ah yea this is just test code to launch my next gen screen. It's replacing the onboarding flow init logic.

NextGenFeature/ViewModels/NextGenSearchViewModel.swift
```diff
@@ -0,0 +1,117 @@
...
+  public func onAppear() {}
+
+  public func searchTextChanged(_ text: String) { self.searchQuery = text }
+
+  private func bindTyping(stream: AsyncStream<String>) {
+    self.bindTask?.cancel()
+
+    self.bindTask = Task {
+      for await text in stream {
+        try? await Task.sleep(nanoseconds: 250_000_000) // ~250ms
```
amy-at-kickstarter, line 71: Ah, is this doing a debounce?

NextGenFeature/Views/NextGenSearchView.swift
scottkicks: Ah this is leftover from some animation stuff. Basically, its a property wrapper that creates a Namespace.ID for the lifetime of the view’s identity. It doesn’t animate anything by itself. it just labels views as belonging together so that SwiftUI can relate the two when animating. I've always used it when using `matchedGeometryEffect`.

Kickstarter.xcodeproj/xcshareddata/xcschemes/NextGenFeature.xcscheme
scottkicks: Yes exactly. I also didn't think that would work, but didn't actually try.

NextGenFeature/Networking/NextGenProjectSearchService.swift
scottkicks: My vote is to try and keep our new patterns outside of existing ones, but we could also opt to start by folding into existing classes like `Service` and move things out if they start getting overwhelmed.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

NextGenFeature/ViewModels/NextGenSearchViewModel.swift
scottkicks: Yep!

Makefile
ifosli: The `IOS_VERSION` here is set in line 9 - updating it there and the other two places in the comment would be better than just deleting it. Also, calling out that if we _do_ update our simulator version, we will need to re-record all our snapshots.

NextGenFeature/Networking/AsyncApolloClient.swift
```diff
@@ -0,0 +1,72 @@
...
+    self.client = client
+  }
+
+  // MARK: - Queries
+
+  public func fetch<Query: GraphQLQuery>(
+    _ query: Query,
+    cachePolicy: CachePolicy = .fetchIgnoringCacheCompletely,
+    contextIdentifier: UUID? = nil,
+    queue: DispatchQueue = .main
```
ifosli: I'm curious: is this code that we specifically want to run on the main queue? I'm not familiar enough with Apollo or async/await to know if that completion should run on the main queue. If the `queue` param controls where the completion is run, and it needs to run on the main queue, what do you think of removing the `queue` from the params entirely? And if it doesn't need to run on the main queue, would it make more sense for this wrapper to own its own dedicated queue? It's not doing complicated logic, so I don't have a strong opinion, but I am curious why you default to the main queue.

NextGenFeature/ViewModels/NextGenSearchViewModelTests.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    XCTAssertEqual(vm.results, [])
+    XCTAssertEqual(vm.statusText, "Idle")
+    XCTAssertEqual(vm.isLoading, false)
+
+    vm.searchTextChanged("ca")
+    vm.searchTextChanged("cats")
+
+    // debounce (250ms) + our fake delay (10ms)
+    try? await Task.sleep(nanoseconds: 300_000_000)
```
ifosli, line 55: Hmm, when we actually implement this, do we want to be able to change the debounce value for tests? Ex by setting it in the appEnvironment, which I think we currently do, or by passing it in as an optional parameter on the view model's init?

NextGenFeature/Networking/AsyncApolloClient.swift
scottkicks: Ah good call out. 
The `queue` you pass to Apollo controls which queue Apollo calls your completion on, which I think might actually matter with an `async/await` approach because after `await` things will continue on the caller’s executor/actor (@MainActor in this test case), not on Apollo’s callback queue.

So, you're right. I think this is a legacy convenience. I'll work on having a dedicated queue that this wrapper can own so we can remove the queue parameter.

NextGenFeature/ViewModels/NextGenSearchViewModelTests.swift
scottkicks: I think that's a good approach! We'll be able to do that when we get to implementation. I wasn't too concerned with details like this for this POC, but we'll likely go with something similar!

NextGenFeatureTests/NextGenFeatureTests.swift
scottkicks: Yea I think so. My plan is to make some dummy mutation/query data + a mock apollo client to pass into this wrapper that we can test to make sure the expected results/errors are returned and on the correct thread. 

I got pulled into some Onboarding changes and have had a sec to write up a test class yet.

scottkicks:
@amy-at-kickstarter @ifosli thanks for the review and feedback! I used Android's Coroutines Best Practices doc to write up a Best Practices doc of our own for this here where we can continue the discussion and have a source of truth

https://kickstarter.atlassian.net/wiki/spaces/NT/pages/3896082452/Swift+6+Apollo+Async+Await+Best+Practices

I'll go ahead and close this PR for now. The new doc links to this in case we want to come back to it.

PR #2603, opened 2025-08-26
[MBL-2725] Fix disabled button font colors

ifosli:
# 📲 What

Ensure our buttons get rendered with colors of our choice, even when they're in the disabled state.

# 🤔 Why

I did some experimenting, and the `baseForegroundColor` that we set in our button configuration has zero effect on the text color of the button when it's rendered in a `disabled` state. So I'm adding the `foregroundColor` that was deleted in https://github.com/kickstarter/ios-oss/pull/2580 back in, but this time set to the correct color.

I had to add a new semantic color as part of this change; our green buttons in their disabled state in the current design system wanted the exact same font color and background color 😅  so I added a new semantic color to represent green action button font color (mid gray on green in light mode and dark gray/black on mid gray in dark mode).

Note: This pr does NOT fix the underlying problem that these buttons don't have sufficient contrast in our design system. I talked to Alison and she's looking into it - it looks like this may be an issue across kickstarter, so in that case, we'll want to fix this on both web and mobile.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2725)

| Before | After (light mode) | After (dark mode) |
| --- | --- | --- |
| image | Screenshot 2025-08-26 at 11 50 15 AM | Screenshot 2025-08-26 at 11 50 31 AM |
| image | Screenshot 2025-08-26 at 11 49 49 AM | Screenshot 2025-08-26 at 11 49 08 AM |

# ✅ Acceptance criteria

- [x] Disabled buttons match our figma mocks in [light mode](https://www.figma.com/design/enEdulL4R1rsMcMttfxJRr/Kickstarter-Design-System---Mobile?node-id=25-1579&p=f&m=dev) and [dark mode](https://www.figma.com/design/U0zRgbvCPa0bRB6VucLq86/iOS-Dark-Mode?node-id=56-1994&p=f&m=dev)


Kickstarter-iOS/Features/FacebookResetPasswordViewController/Controller/__Snapshots__/FacebookResetPasswordViewControllerTests/testFacebookResetPasswordViewController.lang_de_device_phone4_7inch.png
ifosli, line 1: Most of the snapshot changes are this green disabled button. The text color apple chose for us is fairly similar to the color we want (a little greener and a little lighter I think). The screenshots are different enough that they needed to be re-recorded for tests to pass.

Kickstarter-iOS/Features/RewardsCollection/Views/__Snapshots__/RewardCardContainerViewTests/testNonLive_BackedProject_BackedReward.AvailableAddOnsReward_lang_en_device_phone4_7inch.png
ifosli, line 1: Here the font change is more obvious and looks much better!

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
```diff
@@ -63,6 +63,10 @@ extension UIButton {
 
         var newConfig = config
         newConfig.font = styleConfig.font
+        // The `foregroundColor` set gets used directly when determining font color.
+        // Without this override, the `baseForegroundColor` in the configuration does
+        // not actually get used when determining text color for disabled buttons.
```
amy-at-kickstarter, line 68: Is this documented anywhere in `UIButtonConfiguration`? Or was this purely trial and error? I'm wondering if there's some underlying misunderstanding with how we're using the configuration code.

amy-at-kickstarter (approved):
Whoops, meant to approve.

Library/Styles/Buttons/KSRButtonStyleConfiguration.swift
ifosli: Hmm, it's not really documented. I looked into it based on this comment and apple says "The button configuration may transform the base color before applying it to foreground views." ([source](https://developer.apple.com/documentation/uikit/uibuttonconfiguration/baseforegroundcolor#Discussion)). They don't say what to do if we _don't_ want them to transform it. I did find multiple options for how to set the font color, but I ultimately decided on this one because we'd had it in the code already.

PR #2606, opened 2025-09-03
[MBL-2743] Onboarding - Move App Tracking View & Update Analytics Events

scottkicks:
# 📲 What

Moves the App Tracking View to be before the Push Notifications View.

# 🤔 Why

We want to be able to track as much data as we can if users have consented to AppTracking.
Today, if users consent after the notifications view we won't know what they actually did on that screen.

I also updated our app tracking analytic event so that we don't send anything if user "denied" consent permissions. That would defeat the purpose of denying!

# 👀 See

| | After 🦋 |
| --- | --- |
|  | Simulator Screen Recording - iPhone 16 Pro - 2025-09-03 at 08 36 00 |

# ✅ Acceptance criteria

- [x] AppTracking View is shown just before the Notifications View
- [x] Events are only sent if user has consented to app tracking

### DIFF
--- a/Library/OnboardingUseCase.swift
+++ b/Library/OnboardingUseCase.swift
@@ -201,15 +201,15 @@ private func allOnboardingItems(
       in: bundle
     ),
     makeOnboardingItem(
-      title: Strings.onboarding_stay_in_the_know_title(),
-      subTitle: Strings.onboarding_stay_in_the_know_subtitle(),
-      type: .enableNotifications,
-      in: bundle
-    ),
-    makeOnboardingItem(
       title: Strings.onboarding_personalize_your_experience_title(),
       subTitle: Strings.onboarding_personalize_your_experience_subtitle(),
       type: .allowTracking,
+      in: bundle
+    ),
+    makeOnboardingItem(
+      title: Strings.onboarding_stay_in_the_know_title(),
+      subTitle: Strings.onboarding_stay_in_the_know_subtitle(),
+      type: .enableNotifications,
       in: bundle
     ),
     makeOnboardingItem(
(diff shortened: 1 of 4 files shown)


Library/OnboardingUseCase.swift
```diff
@@ -141,11 +141,6 @@ public final class OnboardingUseCase: OnboardingUseCaseType, OnboardingUseCaseUI
       .mapConst(())
 
     _ = self.goToNextItemTappedSignal.signal
-      .observeValues { itemType in
-        if itemType.isPermissionType {
-          // TODO: Emit analytic tracking events here so that we know that the user opted skipped push notifications and/or AppTrackingTransparency views
-        }
-      }
```
scottkicks, line 148: We decided that our insights dashboards can rely on the paged viewed events, that are triggered when the system dialogs are presented, to tell us if users skipped the app tracking or notifications CTAs. 

That is, if the user has consented to apptracking.

Library/OnboardingViewModel.swift
```diff
@@ -90,17 +90,19 @@ public final class OnboardingViewModel: OnboardingViewModelType, Equatable & Ide
...
     AppEnvironment.current.ksrAnalytics.trackPushNotificationPermissionsDialogInteraction(
       .onboardingNotificationsDialog,
       authStatus: authStatus
     )
   }
 
   public func didCompleteAppTrackingDialog(with authStatus: ATTrackingManager.AuthorizationStatus) {
-    /// Send analytics event when the user has finished interacting with the AppTracking  system dialog. `authStatus` let's insights know whether they allowed or denied permissions.
+    /// Send analytics event if the user has allowed AppTracking permission.
+    guard authStatus == .authorized else { return }
```
scottkicks, line 102: this guard is a little redundant since all of our analytics are gated behind whether AppTracking permission have been granted, but I think it adds more clarity to this method.

stevestreza-ksr (approved):
Reviewing since Amy is out. I'm not too familiar with onboarding but this follows easily enough, and the changes make sense to me.

PR #2607, opened 2025-09-03
[MBL-2694] Add missing semantic colors

ifosli:
# 📲 What

Add in remaining semantic colors to our design system with a caveat: I'm not adding in colors with an alpha component (ie partially transparent). I've called them out in the ticket, and will handle them in a separate pr/maybe in a separate ticket.

# 👀 See

[Figma](https://www.figma.com/design/enEdulL4R1rsMcMttfxJRr/Kickstarter-Design-System---Mobile?node-id=1-12&p=f&vars=1&var-id=37-1797&m=dev)

[Jira](https://kickstarter.atlassian.net/browse/MBL-2694)

# ✅ Acceptance criteria

- [x] Colors are added in correctly (description string matches the class structure and color values are correct)
- [x] Tests pass

# ⏰ TODO

- I've called out some remaining colors that aren't handled in this pr. I'm planning to look more into these next week (I'll check in with Amy if she has thoughts on how she wants them handled).


KDS/Sources/KDS/Colors/Colors.swift
```diff
@@ -284,25 +452,91 @@ public struct Colors {
...
+    )
+
+    public static let inverse = SemanticColor(
+      "icon/inverse",
+      lightMode: .gray_300,
+      darkMode: .gray_800
+    )
+
+    public static let pressed = SemanticColor(
+      "icon/primary",
```
scottkicks: should this be "icon/pressed"? I see there's a primary just below this one

ifosli: Yeah, good catch!

PR #2608, opened 2025-09-03
[MBL-2774] Auto Scale Onboarding Views For Smaller Devices

scottkicks:
# 📲 What

Update onboarding views to account for smaller screen sizes.

# 🤔 Why

The new Onboarding view is cut off on small phones - the progress bar is missing, and the secondary button is pushed towards the bottom.

For example, here’s how it looks on the iPhone SE:
image

# 🛠 How

From what I've found, SwiftUI doesn’t have a way to auto-scale an entire view tree to fit a device's height if the content height exceeds it. `minimumScaleFactor` only works on `Text`, and `ViewThatFits` swaps layouts, but it doesn’t scale them.

`scaleEffect` was also causing blurry text depending on the font size which was no good either.

I ended up putting the text and lottie view in a scrollview and pinning the CTA to the bottom safe area inset.

I ran this by Alison and she gave the ok as well. It's not my favorite that you can see the main text behind the CTAs on larger font sizes, but we decided that adding a background to that area wasn't a better alternative.

# 👀 See

| SE | SE Large Text | 15 Pro |
| --- | --- | --- |
| SE normal font | SE Large Font | 15 pro |

# ✅ Acceptance criteria

- [x] Onboarding flow view content is always visible regardless of device size.

### DIFF
--- a/Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
+++ b/Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
@@ -29,13 +29,6 @@ public struct OnboardingView: View {
     return Double(self.currentIndex + 1) / onboardingItemsCount
   }
 
-  @SwiftUI.Environment(\.sizeCategory) private var sizeCategory
-
-  /// SE / older small devices are ~667pt tall
-  private var isSmallDevice: Bool {
-    UIScreen.main.bounds.height < 700
-  }
-
   public init(viewModel: OnboardingViewModel) {
     self.viewModel = viewModel
   }
@@ -52,6 +45,7 @@ public struct OnboardingView: View {
       VStack {
         self.ProgressBarView()
           .accessibilityElement(children: .combine)
+        // TODO: Add accessibility translations [mbl-2418]
 
         ZStack {
           ForEach(Array(self.onboardingItems.enumerated()), id: \.element.id) { index, item in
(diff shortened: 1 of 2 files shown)


Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -45,7 +52,6 @@ public struct OnboardingView: View {
       VStack {
         self.ProgressBarView()
           .accessibilityElement(children: .combine)
-        // TODO: Add accessibility translations [mbl-2418]
```
ifosli: Double-checking: Did these translations already get added?

Kickstarter-iOS/Features/Onboarding/Views/OnboardingView.swift
```diff
@@ -29,6 +29,13 @@ public struct OnboardingView: View {
     return Double(self.currentIndex + 1) / onboardingItemsCount
   }
 
+  @SwiftUI.Environment(\.sizeCategory) private var sizeCategory
```
ifosli: Nit: Looks like this is unused

ifosli (approved):
I think this'll make the view better than it was, but hardcoding values like this scares me a little. For example, have you tested what the view looks like on a phone (smaller or larger) with accessibility large text? And the largest normal size? (How sure are you that this scaling is always sufficient?)

PR #2609, opened 2025-09-04
[MBL-2744] Fix push notification registration

ifosli:
# 📲 What

The segment/braze upgrade broke our push notification registration for some edge cases. Fix push token setup by
1. Ensuring that we tell Braze when new users log in (this fixes the Braze issue where we had to quit and relaunch the app after deleting and reinstalling the app)
2. Telling Braze to never attempt to automatically set up push notifications or push registration. This doesn't fully fix aws notification registration (it still sometimes gets in a weird state when you uninstall and reinstall the app), but it seems to help

# See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2744)

# ✅ Acceptance criteria

- [x] Braze notifications work on a fresh install (after user has logged in and we give the app a little time to actually finish push registration)
- [x] ~aws notifications work even after uninstalling app, reinstalling, and logging back in as same user~ aws notifications sometimes work after uninstalling app, reinstalling, and logging back in as same user. (In current app store version, we see the same issue sometimes, too. It might still be happening more frequently after braze upgrade than before, but these changes make behavior pretty similar.)

### DIFF
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -161,10 +161,10 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         print("📲 [Push Registration] Push token successfully registered (\(token)) ✨")
       }
 
-    self.viewModel.outputs.registerPushTokenInSegment
+    self.viewModel.outputs.registerPushTokenInBraze
       .observeForUI()
       .observeValues { token in
-        self.analytics?.registeredForRemoteNotifications(deviceToken: token)
+        self.braze?.notifications.register(deviceToken: token)
       }
 
     self.viewModel.outputs.triggerOnboardingFlow
@@ -520,6 +520,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     let automation: BrazeKit.Braze.Configuration.Push.Automation = true
     automation.automaticSetup = false
     automation.requestAuthorizationAtLaunch = false
+    automation.registerDeviceToken = false
     return automation
   }
 }
--- a/Kickstarter-iOS/AppDelegate.swift
+++ b/Kickstarter-iOS/AppDelegate.swift
@@ -48,7 +48,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     // Braze expects to be configured immediately, but segment destination plugins are initialized
     // async. This method bridges that gap.
     // https://www.braze.com/docs/developer_guide/sdk_integration#swift_step-2-set-up-delayed-initialization-optional
-    Braze.prepareForDelayedInitialization(pushAutomation: .init(automaticSetup: false))
+    Braze.prepareForDelayedInitialization(pushAutomation: self.configuredBrazePushAutmation())
 
     UIView.doBadSwizzleStuff()
     UIViewController.doBadSwizzleStuff()
@@ -478,9 +478,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     return BrazeDestination(
       additionalConfiguration: { configuration in
         configuration.triggerMinimumTimeInterval = 5
-        configuration.push.automation = true
-        configuration.push.automation.requestAuthorizationAtLaunch = false
-        configuration.push.automation.automaticSetup = false
+        configuration.push.automation = self.configuredBrazePushAutmation()
         // TODO(MBL-2742): Change the logger level to `info` or `error` if it gets tedious.
         configuration.logger.level = .debug
       }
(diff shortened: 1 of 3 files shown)


Kickstarter-iOS/AppDelegate.swift
```diff
@@ -31,6 +31,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
   fileprivate var brazeSubscription: BrazeKit.Braze.Cancellable?
 
   private var analytics: Segment.Analytics?
+  private weak var braze: Braze?
```
ifosli, line 34: We need to hold onto this in order to change braze user after a user logs in (this is required for push to work correctly; the identify event tracked through segment is not sufficient)

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -157,12 +161,6 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         print("📲 [Push Registration] Push token successfully registered (\(token)) ✨")
       }
 
-    self.viewModel.outputs.registerPushTokenInSegment
-      .observeForUI()
-      .observeValues { token in
-        self.analytics?.registeredForRemoteNotifications(deviceToken: token)
```
ifosli, line 163: Now that I've verified that braze automatic push token registration works, we no longer need this call. (Upgraded segment doesn't actually do anything for push notifications unless we get another plugin.) I left it initially in case it was doing something/in case I wanted to call braze directly here to register the push token in braze. But, now that I'm sure the automatic push token registration was not the issue, it can be deleted

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -47,7 +48,7 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     // Braze expects to be configured immediately, but segment destination plugins are initialized
     // async. This method bridges that gap.
     // https://www.braze.com/docs/developer_guide/sdk_integration#swift_step-2-set-up-delayed-initialization-optional
-    BrazeDestination.prepareForDelayedInitialization()
+    Braze.prepareForDelayedInitialization(pushAutomation: self.configuredBrazePushAutmation())
```
ifosli: Annoyingly enough, while the `BrazeDestination` exposes this method for us, it does NOT include any of the optional parameters. I learned today that when we call this method without parameters, full push automation gets turned on immediately. (This answers my previous question about how to turn that automation off, though the automatic setup has grown on me.) When we call this method directly on `Braze` (which is the only thing the `BrazeDestination` copy was doing anyways), we get the control we need over the setup.

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -480,14 +478,14 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
     return BrazeDestination(
       additionalConfiguration: { configuration in
         configuration.triggerMinimumTimeInterval = 5
-        configuration.push.automation = true
-        configuration.push.automation.requestAuthorizationAtLaunch = false
+        configuration.push.automation = self.configuredBrazePushAutmation()
```
stevestreza-ksr: Nit: typo in `configuredBrazePushAutmation` missing an o in `Automation`

Kickstarter-iOS/AppDelegate.swift
```diff
@@ -508,6 +506,16 @@ internal final class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
   }
+
+  // This configuration object defines how much automation Braze does. It gets set both when
+  // we configure `BrazeDestination` and when we call `Braze.prepareForDelayedInitialization`.
+  // https://braze-inc.github.io/braze-swift-sdk/documentation/brazekit/braze/configuration-swift.class/push-swift.class/automation-swift.class/
+  private func configuredBrazePushAutmation() -> BrazeKit.Braze.Configuration.Push.Automation {
```
stevestreza-ksr: Can this be driven out of the view model?

Kickstarter-iOS/AppDelegate.swift
ifosli: Nice catch!

Kickstarter-iOS/AppDelegate.swift
ifosli: Hmm, I don't think it makes sense to move it; I don't see a file it would belong in, so we'd need to create a separate file just for this method. More importantly, the main braze configuration code is in the method right above in this file and cannot be moved out, and I prefer to keep these methods together. If you feel strongly about this, though, and have specific ideas for how you'd like to see it implemented, feel free to let me know!

amy-at-kickstarter (approved):

PR #2611, opened 2025-09-05
[MBL-2755] Update Facebook SDK

stevestreza-ksr:
# 📲 What

Updates Facebook dependency.

# 🤔 Why

Needed support for iOS 18 and next gen architecture.

# ✅ Acceptance criteria

Tests should pass. Functions relating to this library should not have regressions.

### DIFF
--- a/Library/MockFacebookSDK.swift
+++ b/Library/MockFacebookSDK.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Library
 import UIKit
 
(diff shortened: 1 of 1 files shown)


stevestreza-ksr:
@ifosli I reset the review as I did in fact run in to login issues with Facebook. There were some duplicate symbols being complained about, so I moved all the code for Facebook to a custom type in Library, which fixed the issue. Logging in works on my device now. Shouldn't be anything structurally different.

PR #2616, opened 2025-09-09
Delete some files that were deprecated in 2022

amy-at-kickstarter:
# 📲 What

Delete files labeled `DEPRECATED_09_07_2022`.

# 🤔 Why

I don't think we're going to need these any more. As part of project file cleanup, it's good to delete totally unused files.


Kickstarter.xcodeproj/project.pbxproj
```diff
@@ -5167,7 +5134,6 @@
 				19450C1F28C81FEF00C60F97 /* RewardsCollection */,
 				1937A70928C93BC800DD732D /* RewardPledgeNavigation */,
 				3344A7742D9C4D9900D7DE91 /* RewardTracking */,
-				1937A70A28C93BD100DD732D /* RiskMessaging */,
```
amy-at-kickstarter, line 5170: This was an empty group I deleted as well.

ifosli (approved):

PR #2619, opened 2025-09-10
[MBL-2690] Add New ApolloGraphClient Target & Scheme

scottkicks:
# 📲 What

Adds a new target, that supports Swift 6, that will house the new async/await apollo wrapper.

Empty for now. I'll add the new wrapper code separately. 

It'll be easier to pull this branch and open the app to see the new target code/settings.

# 🤔 Why Does This Need To Be In A New Target?

- Isolation & clear API
- Testability
- Reuse
- Future-proofing: Easy to swap Apollo, or migrate to Apollo 2.0 when available

# 🛠 How

- Creates the new target with tests
- Sets swift version to 6.0
- Link Apollo, KsApi, and Library
- Adds a new scheme that can be built

# ✅ Acceptance criteria

- [x] App build and all tests pass
- [x] New target builds with linked frameworks and binaries.

# ⏰ TODO

- Add the new apollo wrapper to this new target.

### DIFF
--- a/ApolloGraphClient/ApolloGraphClient.docc/ApolloGraphClient.md
+++ b/ApolloGraphClient/ApolloGraphClient.docc/ApolloGraphClient.md
@@ -1,0 +1,13 @@
+# ``ApolloGraphClient``
+
+<!--@START_MENU_TOKEN@-->Summary<!--@END_MENU_TOKEN@-->
+
+## Overview
+
+<!--@START_MENU_TOKEN@-->Text<!--@END_MENU_TOKEN@-->
+
+## Topics
+
+### <!--@START_MENU_TOKEN@-->Group<!--@END_MENU_TOKEN@-->
+
+- <!--@START_MENU_TOKEN@-->``Symbol``<!--@END_MENU_TOKEN@-->
--- a/ApolloGraphClient/ApolloGraphClient.h
+++ b/ApolloGraphClient/ApolloGraphClient.h
@@ -1,0 +1,11 @@
+#import <Foundation/Foundation.h>
+
+//! Project version number for ApolloGraphClient.
+FOUNDATION_EXPORT double ApolloGraphClientVersionNumber;
+
+//! Project version string for ApolloGraphClient.
+FOUNDATION_EXPORT const unsigned char ApolloGraphClientVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <ApolloGraphClient/PublicHeader.h>
+
+
--- a/ApolloGraphClientTests/ApolloGraphClientTests.swift
+++ b/ApolloGraphClientTests/ApolloGraphClientTests.swift
@@ -1,0 +1,8 @@
+@testable import ApolloGraphClient
+import Testing
+
+struct ApolloGraphClientTests {
+  @Test func example() async throws {
+    // Write your test here and use APIs like `#expect(...)` to check expected conditions.
+  }
+}
--- a/Kickstarter.xcodeproj/project.pbxproj
+++ b/Kickstarter.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 54;
+	objectVersion = 70;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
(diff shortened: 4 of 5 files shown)


ApolloGraphClient/ApolloGraphClient.docc/ApolloGraphClient.md
```diff
@@ -0,0 +1,13 @@
+# ``ApolloGraphClient``
```
amy-at-kickstarter, line 1: Obviously this isn't up for review yet, but a question! Could this live in `GraphAPI`, instead? Or am I misunderstanding what you're doing here?

scottkicks: Maybe a different target name would be helpful. What would you name this new target? I guess it doesn't have to be Apollo, or graph, specific. The new wrapper is just the initial use case

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ApolloGraphClient/ApolloGraphClient.docc/ApolloGraphClient.md
amy-at-kickstarter: I'm just thinking about project structure - why would the Apollo wrapper need its own target, versus just living with the other GraphQL/Apollo code? Or were other things going to go in there, too?

scottkicks: Ah I've been thinking that we would try to keep all of our new architecture code in a new swift 6 target. 

This wrapper could live with our other GraphQL/Apollo code though.

amy-at-kickstarter: Aaah gotcha. So are you proposing this target for all the new architecture code? Or just adding a small dependency that the new code can import?

A couple thoughts:
1. If this is just a small dependency to add the wrapper, I do think it would make sense to either lump it in with existing GraphQL code (unless that's an issue with Swift 6.0), or make it a SPM module instead of a target. I don't love maintaining extra targets, if we can avoid it.
2. You're pointing at a bigger question, which is: how do we structure the new app code in the current app structure? Definitely we need a new target for Swift 6, but beyond that, are we going to make vertical 'slice' modules (like a `Project` target, a `Checkout` target, etc. etc.) or one big module (like `Kickstarter-New-Framework`)? More problematically, what targets/SPM modules will our new architecture code need to import? It'd be helpful if you could sketch this out in the new architecture doc so we could workshop it a bit. @ifosli and I have been playing with the _existing_ project structure, and seeing how we might break it down, and we landed on some pretty thorny questions. (For example: will our new module need to import the _old_ app code, just so we can link between old pages and new pages? Can we pull out some critical functionality into its own module, like Analytics, without bringing in all of API v1?)

scottkicks: Excellent questions. For this Apollo wrapper, at least in my POC, I needed to import KsApi and Library to access things like our GraphQL SearchQuery. I'm playing around with just adding this wrapper and tests alongside our existing GraphQL code in KsApi and I'm running into some confusing Apollo compile errors in my tests that I can't figure out.

I was more suggesting one big target for newer code rather than a bunch of individual targets, but I'm not sure of the technical benefits of one vs the other. 

I wonder if we should sync on this as a team. Especially if you and Ingerid have been throwing some ideas around alread.

scottkicks: It's starting to sound like we could benefit from building POCs for specific use cases to answer some of those thorny questions...

scottkicks: @amy-at-kickstarter I have a [PR](https://github.com/kickstarter/ios-oss/pull/2621) building right now that adds this new wrapper into `KsApi` for now. We can always move it if/when we decide on the project structure we want.

scottkicks:
Closing this PR out until the team re-aligns on our the project structure that we want moving forward. I've made a PR that adds our new async/await apollo wrapper to `KsApi` target in the meantime.

PR #2621, opened 2025-09-11
[MBL-2690] Async/Await Apollo Wrapper

scottkicks:
# 📲 What

Adds a lightweight `AsyncApolloClient` to `KsApi` with some unit tests.

This wrapper converts Apollo’s callback-based fetch and perform methods into modern Swift async/await APIs.

# 🛠 How

- Adds`AsyncApolloClient`:  a minimal async/await bridge around `ApolloClientProtocol`.
   - Supports both fetch (queries) and perform (mutations).
- Adds `MockApolloClient` for testing.
   - simulates success/failure cases
   - records callbackQueue changes
- Adds `AsyncApolloClientTests`
   - Tests success and failure paths for both fetch and perform.

# ✅ Acceptance criteria

- [x] `AsyncApolloClient` builds and passes unit tests.
- [x] Existing Apollo code is unaffected.
- [x] Tests cover success/failure and queue forwarding.

# ⏰ TODO

- [ ] Consider adding Task cancellation support. If fetch or perform are cancelled, we should cancel the underlying Apollo request, too. Otherwise the network work keeps running pointlessly and may still call the completion later.

### DIFF
--- a/KsApi/AsyncApolloClient.swift
+++ b/KsApi/AsyncApolloClient.swift
@@ -1,0 +1,80 @@
+@preconcurrency import Apollo
+import Foundation
+import GraphAPI
+
+/**
+ An async/await wrapper around `ApolloClientProtocol`.
+
+ Purpose:
+    - bridge Apollo’s callback APIs with Swift Concurrency.
+
+ Callback queue:
+    - all Apollo completions run on `callbackQueue`(defaults to `.global(qos: .userInitiated)`.
+
+ Methods:
+    - `fetch(_:)`: forwards `cachePolicy`/`contextIdentifier`
+      - returns `GraphQLResult`or throws an Apollo error.
+    - `perform(_:)`: forwards `publishResultToStore`
+      - returns `GraphQLResult`or throws an Apollo error.
+
+ Concurrency: marked `@unchecked Sendable`. queue hopping is controlled by `callbackQueue`.
+
+ */
+public final class AsyncApolloClient: @unchecked Sendable {
+  private let client: ApolloClientProtocol
+  private let callbackQueue: DispatchQueue
+
+  /// `callbackQueue` defaults to a background queue..
+  public init(
+    client: ApolloClientProtocol,
+    /// using `.userInitiated` for scheduling the work the user is waiting on  more efficiently.. `.background` can sometimes get deprioritized.
+    callbackQueue: DispatchQueue = .global(qos: .userInitiated)
+  ) {
+    self.client = client
+    self.callbackQueue = callbackQueue
+  }
+
+  // MARK: - Queries
+
+  public func fetch<Query: GraphQLQuery>(
+    _ query: Query,
+    cachePolicy: CachePolicy = .fetchIgnoringCacheCompletely,
+    contextIdentifier: UUID? = nil
+  ) async throws -> GraphQLResult<Query.Data> {
+    try await withCheckedThrowingContinuation { cont in
+      _ = self.client.fetch(
+        query: query,
+        cachePolicy: cachePolicy,
+        contextIdentifier: contextIdentifier,
+        context: nil,
+        queue: self.callbackQueue
+      ) { result in
+        switch result {
+        case let .success(value): cont.resume(returning: value)
+        case let .failure(error): cont.resume(throwing: error)
+        }
+      }
+    }
+  }
+
+  // MARK: - Mutations
+
+  public func perform<Mutation: GraphQLMutation>(
+    _ mutation: Mutation,
+    publishResultToStore: Bool = true
+  ) async throws -> GraphQLResult<Mutation.Data> {
+    try await withCheckedThrowingContinuation { cont in
+      _ = self.client.perform(
+        mutation: mutation,
+        publishResultToStore: publishResultToStore,
+        context: nil,
+        queue: self.callbackQueue
+      ) { result in
+        switch result {
+        case let .success(value): cont.resume(returning: value)
+        case let .failure(error): cont.resume(throwing: error)
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


KsApi/AsyncApolloClient.swift
```diff
@@ -0,0 +1,80 @@
...
+    try await withCheckedThrowingContinuation { cont in
+      _ = self.client.fetch(
+        query: query,
+        cachePolicy: cachePolicy,
+        contextIdentifier: contextIdentifier,
+        context: nil,
+        queue: self.callbackQueue
+      ) { result in
+        switch result {
+        case let .success(value): cont.resume(returning: value)
```
amy-at-kickstarter, line 53: Given this is using a background thread for `callbackQueue` - will this correctly resume on the same thread that called the `fetch`? I assume as much, based on the docs for `withCheckedThrowingContinuation`, but double-checking here.

KsApi/AsyncApolloClientTests.swift
```diff
@@ -0,0 +1,108 @@
...
+    let mock = MockApolloClient()
+    mock.fetchShouldSucceed = false
+
+    let queue = DispatchQueue(label: "fetchQueue")
+    let asyncClient = AsyncApolloClient(client: mock, callbackQueue: queue)
+
+    do {
+      _ = try await asyncClient.fetch(StubQuery())
+
+      XCTFail("Expected to throw an error but didn't")
```
amy-at-kickstarter: There is an XCTest helper for this - `XCTAssertThrowsError`

KsApi/MockApolloClient.swift
```diff
@@ -0,0 +1,140 @@
...
+    query _: Query,
+    cachePolicy _: Apollo.CachePolicy,
+    contextIdentifier _: UUID?,
+    context _: (any Apollo.RequestContext)?,
+    queue: DispatchQueue,
+    resultHandler: Apollo.GraphQLResultHandler<Query.Data>?
+  ) -> Cancellable {
+    self.lastFetchQueue = queue
+
+    /// Mimic Apollo getting called on the provided queue
```
amy-at-kickstarter, line 44: I like that you wrote these, although I'm a bit nervous they may not end up being accurate to Apollo's internal behavior. Have you done some smoke testing with actual GraphQL requests in the app using this wrapper?

KsApi/AsyncApolloClient.swift
```diff
@@ -0,0 +1,80 @@
...
+    publishResultToStore: Bool = true
+  ) async throws -> GraphQLResult<Mutation.Data> {
+    try await withCheckedThrowingContinuation { cont in
+      _ = self.client.perform(
+        mutation: mutation,
+        publishResultToStore: publishResultToStore,
+        context: nil,
+        queue: self.callbackQueue
+      ) { result in
+        switch result {
```
stevestreza-ksr, line 73: Nit: Since this is a Result type, you could just call `cont.resume(returning: result)` directly, theoretically

KsApi/AsyncApolloClient.swift
scottkicks: Yea it won’t _necessarily_ resume on the same thread. It resumes on the same executor/actor as the awaiting task.
So, If the caller is on @MainActor, the continuation resume should jump back to main.

You can technically call `cont.resume...` on any thread I believe.

KsApi/MockApolloClient.swift
scottkicks: Mostly just with the SearchQuery in my POC. I could stand to test others

scottkicks: did some testing with a few different GraphQL requests and all is looking good. 
Here's an example of fetching a project by ID that you can quick plug into `Service` if you'd like to play around with the wrapper a bit.

```
public func asyncApolloWrapperSmokeTesting() async -> Void {
    let asyncApollo = AsyncApolloClient(client: GraphQL.shared.client, callbackQueue: .global(qos: .userInitiated))
    
    do {

      let result = try await asyncApollo.fetch(GraphAPI.FetchProjectByIdQuery(projectId: 2129414155, withStoredCards: false))

      if let project = result.data?.project {
        let name      = project.name
        let slug      = project.slug
        let state     = project.state.rawValue
        let goal      = project.goal
        let pledged   = project.pledged
        let currency  = project.currency
        let creator   = project.creator?.name
        
        print("""
            name: \(name)
            slug: \(slug)
            state: \(state)
            goal: \(goal) \(currency)
            pledged: \(pledged) \(currency)
            creator: \(creator)
            """)
      } else {
        print("Nil result. errors: \(result.errors)")
      }
    } catch {
      print(error.localizedDescription)
    }
  }
```

KsApi/AsyncApolloClientTests.swift
```diff
@@ -0,0 +1,108 @@
...
+    let mock = MockApolloClient()
+    mock.fetchShouldSucceed = false
+
+    let queue = DispatchQueue(label: "fetchQueue")
+    let asyncClient = AsyncApolloClient(client: mock, callbackQueue: queue)
+
+    do {
+      _ = try await asyncClient.fetch(StubQuery())
+
+      XCTAssertThrowsError("Expected to throw an error.")
```
amy-at-kickstarter: Ah, I believe you need to call this like 
`XCTAssertThrowsError(try await asyncClient.fetch(StubQuery()), "Expected to throw error)`

scottkicks: Actually putting the fetch directly into the assert leads to this compile error

`'async' call in an autoclosure that does not support concurrency`

stevestreza-ksr (approved):

KsApi/AsyncApolloClientTests.swift
amy-at-kickstarter: Ah, unfortunate. In that case I think you just want the plan `XCTFail`!

scottkicks: Sure thing!

PR #2624, opened 2025-09-15
Make swiftformat mandatory

amy-at-kickstarter:
# 📲 What

Make our `swiftformat` checks mandatory.

# 🤔 Why

Since most of us use a pre-commit hook for `swiftformat`, having errors creep into `main` can block other engineers from making commits. 

# 👀 See

If `swiftformat` fails, it will show up as a failed check by Danger:

Screenshot 2025-09-15 at 12 49 38 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,15 @@ jobs:
           # This is for Danger, which requires a bit more git history
           fetch_depth: 100 
       - run:
+          name: Install Bundler
+          command: bundle check || bundle install
+          environment:
+            BUNDLE_JOBS: 4
+            BUNDLE_RETRY: 3
+      - run:
+          name: Danger
+          command: bin/danger.sh
+      - run:
           name: SPM SSH Workaround
           command: *spm_ssh_workaround
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,15 @@ jobs:
       - git-shallow-clone/checkout:
           # This is for Danger, which requires a bit more git history
           fetch_depth: 100 
+      - run:
```
amy-at-kickstarter, line 101: SPM is a pretty slow step. Moving these two steps above SPM means we'll get `swiftformat` failures much sooner after you push.

Dangerfile
```diff
@@ -2,7 +2,7 @@
 warn("Big PR") if git.lines_of_code > 500
 
 # SwiftFormat
+swiftformat.additional_message = "This PR violates our formatting conventions. Run ./bin/format.sh to fix."
 swiftformat.binary_path = "bin/swiftformat"
 swiftformat.additional_args = "--config .swiftformat --swiftversion 5"
-swiftformat.exclude = %w(Library/Strings.swift Library/Styles/Colors.swift)
```
amy-at-kickstarter, line 7: These were already excluded in `.swiftformat`.

scottkicks (approved):

PR #2625, opened 2025-09-15
MBL-2782: Add swiftlint and run automatically with Danger

amy-at-kickstarter:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

# 🏎 Performance

# ✅ Acceptance criteria

- [ ] Environment to test on
- [ ] Issues to look out for

# ⏰ TODO

- [ ] Will add more tests


amy-at-kickstarter:
> I like it! I'm not sure about the "making it mandatory" part- I'd like to see exactly how it works first. (Ex I think a file length rule is interesting and it's a good idea in general, but I'd still like to be able to ex add a new import without needing to refactor a file.) But definitely open to exploring this!

Good point! There would be a couple workarounds for that specific case:
1. You can disable a rule for a file or group of lines
2. The linter provides a warning at a shorter file length, and only becomes an error a few hundred lines later. So there'd be a lot of prior warning that the file was getting close to the limit.

But either way, I do expect there to be a lot of rules-tuning before we got to that point.

ifosli (approved):

PR #2627, opened 2025-09-17
[MBL-2766] Remove IBDesignable

stevestreza-ksr:
Removes a use of IBDesignable, which is deprecated in iOS 18.


amy-at-kickstarter (approved):
RIP, `@IBDesignable`.

PR #2628, opened 2025-09-17
[MBL-2783] Show fix payment button

ifosli:
# 📲 What

Show the fix payment button if a backing is in the errored state, regardless of if we can show the payment method as well. After talking to Tae and looking at the code, I decided to move the fix payment button logic into the main `ManagePledgeViewController` instead of keeping it in the `ManagePledgePaymentMethodView`. This makes it easier to keep the payment method section and the fix payment button view logic in sync.

Note: I was never able to repro the "fix button doesn't show" so I'm not 100% sure if the underlying cause was that we didn't have a payment method to show. However, this seems like a likely fix. (When I tested this without payment method, I manually set the payment method to `nil`.)

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2783)

| Before 🐛 | After 🦋 |
| --- | --- |
| image | image |
| image | image|

# ✅ Acceptance criteria

- [x] Fix button shows for errored pledges
- [x] Fix button does not show for normal pledges
- [x] Button works/UI still looks fine for dynamic type


Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -191,11 +192,20 @@ public final class ManagePledgeViewModel:
       .take(until: backing.ignoreValues())
       .ignoreValues()
 
+    self.fixButtonHidden = backing
+      // For PLOT pledges, the fix button is always hidden because the fix action is handled by the `PledgeStatusLabelView`.
```
ifosli: This comment/logic is copied from the `ManagePledgePaymentMethodViewModel`. I did not copy the "this is temporary" text or the ticket; after talking to Scott about the state of that ticket, it's been marked as "won't do."

nativeksr:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
ManagePledgeViewControllerTests.swift | 684 | File should contain 400 lines or less: currently contains 684 (file_length)
ManagePledgeViewModel.swift | 701 | File should contain 400 lines or less: currently contains 701 (file_length)
ManagePledgeViewModel.swift | 64 | Class body should span 250 lines or less excluding comments and whitespace: currently spans 325 lines (type_body_length)
#### Errors

File | Line | Reason |
| --- | ----- | ----- |
ManagePledgeViewControllerTests.swift | 8 | Class body should span 350 lines or less excluding comments and whitespace: currently spans 507 lines (type_body_length)
ManagePledgeViewModel.swift | 66 | Initializer body should span 100 lines or less excluding comments and whitespace: currently spans 257 lines (function_body_length)
ManagePledgeViewModelTests.swift | 1871 | File should contain 1000 lines or less: currently contains 1871 (file_length)
ManagePledgeViewModelTests.swift | 922 | Function body should span 100 lines or less excluding comments and whitespace: currently spans 139 lines (function_body_length)
ManagePledgeViewModelTests.swift | 9 | Class body should span 350 lines or less excluding comments and whitespace: currently spans 1428 lines (type_body_length)


</p>

Kickstarter-iOS/Features/ManagePledge/Controllers/ManagePledgeViewController.swift
```diff
@@ -198,10 +210,16 @@ final class ManagePledgeViewController: UIViewController, MessageBannerViewContr
 
     self.pledgeDetailsSectionLabel.rac.text = self.viewModel.outputs.pledgeDetailsSectionLabelText
 
+    self.viewModel.outputs.fixButtonHidden
+      .observeForUI()
+      .observeValues { [weak self] hidden in
+        self?.fixPaymentButton.isHidden = hidden
```
amy-at-kickstarter: Is this _always_ output after the view is initialized? Or do you need to set `isHidden = true` by default when you create `fixPaymentButton`?

Kickstarter-iOS/Features/ManagePledge/Controllers/__Snapshots__/ManagePledgeViewControllerTests/testView_ErroredBacking.lang_en_device_phone4_7inch.png
amy-at-kickstarter, line 1: Nit: even though this screenshot is correct, it's a bit confusing that the actual fix button isn't on the screen. For this test, how hard would it be to scroll the fix button into view?

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -191,11 +192,20 @@ public final class ManagePledgeViewModel:
...
+
     self.paymentMethodViewHidden = Signal.combineLatest(
+      self.fixButtonHidden,
       userIsCreatorOfProject,
       backing.map { backing in backing.paymentSource }
     )
-    .map { userIsCreatorOfProject, creditCard in userIsCreatorOfProject || creditCard == nil }
+    .map { fixButtonHidden, userIsCreatorOfProject, creditCard in
+      if userIsCreatorOfProject { return true }
+      if !fixButtonHidden { return false }
```
amy-at-kickstarter: Nit: the multiple negatives here ("is not hidden = false") took me an extra second to parse. This could use a comment.

amy-at-kickstarter (commented):
Always showing the 'Fix Payment' button makes sense to me, and moving the view model logic up a level (from `ManagePledgePaymentMethodViewModel` into `ManagePledgeViewModel`) also makes sense. But I'm not sure so about moving the button itself from `ManagePledgePaymentMethodView` into `ManagePledgeViewController`. It feels like it logically still lives there, even if it's not always displayed - making the pledge controller manage two views in sync feels a little bit more complex.

As a different option, what about making the payment method view 'dumber'? The manage pledge page could still control and pass down whether the fix button could be shown, using `ManagePledgePaymentMethodViewData`:

```swift
public struct ManagePledgePaymentMethodViewData: Equatable {
  public let showFixPaymentButton: Bool //replaces backingState and isPledgeOverTime
  public let expirationDate: String?
  public let lastFour: String?
  public let creditCardType: CreditCardType?
  public let paymentType: PaymentType?
}
```

and then the manage view can handle whether to hide/show the credit card information and the button, based on whether they're set or not.

Library/ViewModels/ManagePledgeViewModel.swift
ifosli: Added comments!

Kickstarter-iOS/Features/ManagePledge/Controllers/__Snapshots__/ManagePledgeViewControllerTests/testView_ErroredBacking.lang_en_device_phone4_7inch.png
ifosli: It turned out to be pretty easy; thanks for the suggestion!

ifosli:
> Always showing the 'Fix Payment' button makes sense to me, and moving the view model logic up a level (from `ManagePledgePaymentMethodViewModel` into `ManagePledgeViewModel`) also makes sense. But I'm not sure so about moving the button itself from `ManagePledgePaymentMethodView` into `ManagePledgeViewController`. It feels like it logically still lives there, even if it's not always displayed - making the pledge controller manage two views in sync feels a little bit more complex.
> 
> As a different option, what about making the payment method view 'dumber'? The manage pledge page could still control and pass down whether the fix button could be shown, using `ManagePledgePaymentMethodViewData`:
> 
> ```swift
> public struct ManagePledgePaymentMethodViewData: Equatable {
>   public let showFixPaymentButton: Bool //replaces backingState and isPledgeOverTime
>   public let expirationDate: String?
>   public let lastFour: String?
>   public let creditCardType: CreditCardType?
>   public let paymentType: PaymentType?
> }
> ```
> 
> and then the manage view can handle whether to hide/show the credit card information and the button, based on whether they're set or not.

I like this! Thanks for the suggestion! I've updated the code to do this instead (feel free to look at changes from the last commit if you want, but it's basically a re-write). I re-tested and re-recorded "after" snapshots, even though they ended up looking the same.

Kickstarter-iOS/Features/ManagePledge/Controllers/ManagePledgeViewControllerTests.swift
```diff
@@ -351,6 +351,13 @@ final class ManagePledgeViewControllerTests: TestCase {
         controller.tableView.layoutIfNeeded()
         controller.tableView.reloadData()
 
+        // Scroll down to top of table view to ensure the fix payment button is on screen.
```
amy-at-kickstarter, line 418: Nice 👍

Kickstarter-iOS/Features/ManagePledge/Views/ManagePledgePaymentMethodView.swift
```diff
@@ -74,9 +76,6 @@ final class ManagePledgePaymentMethodView: UIView {
       action: #selector(ManagePledgePaymentMethodView.fixButtonTapped),
       for: .touchUpInside
     )
-
-    self.fixButton.setContentHuggingPriority(.required, for: .horizontal)
```
amy-at-kickstarter, line 78: I assume these aren't necessary any more?

Library/ViewModels/ManagePledgePaymentMethodViewModel.swift
```diff
@@ -82,9 +81,7 @@ public final class ManagePledgePaymentMethodViewModel: ManagePledgePaymentMethod
       .map { Strings.Credit_card_expiration(expiration_date: $0) }
 
     self.fixButtonHidden = self.configureWithDataSignal
-      // TODO: these changes are temporary and will likely be removed when we get to the native implementation in this ticket [MBL-2012](https://kickstarter.atlassian.net/browse/MBL-2012)
-      // For PLOT pledges, the fix button is always hidden because the fix action is handled by the `PledgeStatusLabelView`.
-      .map { $0.backingState != .errored || $0.isPledgeOverTime }
+      .map { !$0.showFixPaymentButton }
```
amy-at-kickstarter, line 84: Aaah so much nicer.

Library/ViewModels/ManagePledgePaymentMethodViewModelTests.swift
```diff
@@ -142,58 +138,4 @@ final class ManagePledgePaymentMethodViewModelTests: TestCase {
     self.fixButtonHidden.assertValues([false])
     self.notifyDelegateFixButtonTapped.assertValueCount(1)
   }
-
-  func testFixButton_IsHidden_WhenPlotErrored() {
```
amy-at-kickstarter, line 146: Nice to see this moved out to `ManagePledgeViewModelTests`, makes so much more sense 👍

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -193,9 +193,16 @@ public final class ManagePledgeViewModel:
...
+      backing
     )
-    .map { userIsCreatorOfProject, creditCard in userIsCreatorOfProject || creditCard == nil }
+    .map { userIsCreatorOfProject, backing in
+      // Never show payment method view if the user is the creator.
+      if userIsCreatorOfProject { return true }
+      // Always show payment method view if the fix button should be shown.
+      if shouldShowFixPaymentMethodButton(for: backing) { return false }
+      // Otherwise, show payment method view if we have a payment method to display.
+      return backing.paymentSource == nil
```
amy-at-kickstarter, line 204: 👍 Much clearer

Library/ViewModels/ManagePledgeViewModel.swift
```diff
@@ -193,9 +193,16 @@ public final class ManagePledgeViewModel:
 
     self.paymentMethodViewHidden = Signal.combineLatest(
       userIsCreatorOfProject,
-      backing.map { backing in backing.paymentSource }
+      backing
     )
-    .map { userIsCreatorOfProject, creditCard in userIsCreatorOfProject || creditCard == nil }
+    .map { userIsCreatorOfProject, backing in
+      // Never show payment method view if the user is the creator.
+      if userIsCreatorOfProject { return true }
```
amy-at-kickstarter, line 200: As an aside, can a creator even _get_ in this state? Seems pretty weird, they'd have to have a backing on their own project?

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -130,12 +130,154 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
+    withEnvironment(apiService: mockService) {
+      self.vm.inputs.configureWith((Param.slug("project-slug"), Param.id(1)))
+
+      self.vm.inputs.viewDidLoad()
+
+      self.configurePaymentMethodView.assertDidNotEmitValue()
+
+      self.scheduler.advance()
+
+      self.configurePaymentMethodView.assertValues([pledgePaymentMethodViewData])
```
amy-at-kickstarter: Nit 1: should add a failure message here that explains the whole thing with the PLOT fix payment button.
Nit 2: I prefer `assertLastValue` over `assertValues` because it's less brittle.

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -130,12 +130,154 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
+    let envelope = ProjectAndBackingEnvelope(project: project, backing: backing)
+
+    let mockService = MockService(
+      fetchManagePledgeViewBackingResult: .success(envelope),
+      fetchProjectResult: .success(project),
+      fetchProjectRewardsResult: .success([.template])
+    )
+
+    let pledgePaymentMethodViewData = ManagePledgePaymentMethodViewData(
+      showFixPaymentButton: false,
```
amy-at-kickstarter, line 278: Is this the same situation, where the fix button with belong to the PLOT view?

Library/ViewModels/ManagePledgeViewModelTests.swift
```diff
@@ -1490,6 +1631,30 @@ internal final class ManagePledgeViewModelTests: TestCase {
...
+      fetchProjectRewardsResult: .success([.template])
+    )
+
+    withEnvironment(apiService: mockService) {
+      self.vm.inputs.configureWith((Param.slug("project-slug"), Param.id(1)))
+      self.vm.inputs.viewDidLoad()
+
+      self.scheduler.advance()
+
+      self.paymentMethodViewHidden.assertValues([true])
```
amy-at-kickstarter: Do you also have a view model or screenshot test for the "no payment method but please fix" state? Would be good to enshrine this fix in a test.

amy-at-kickstarter (approved):
LGTM, with some nits/suggestions!

Kickstarter-iOS/Features/ManagePledge/Views/ManagePledgePaymentMethodView.swift
ifosli: Correct! The fix button is no longer "fighting for space" since it's on its own line.

Library/ViewModels/ManagePledgeViewModel.swift
ifosli: I don't think so. But it wouldn't surprise me if we used the manage pledge screen for creators in the past (ex to see a specific backer's pledge) so it felt safer to leave this check, just in case.

Library/ViewModels/ManagePledgeViewModelTests.swift
ifosli: Updated! I swapped the `assertValues` to `assertLastValue` in the tests I've touched; I also like this one better, but I'd originally left it as `assertLastValue` to be consistent with the rest of the file.

I didn't add a failure message, since there's unrelated fields in the data that could error. But I did add a comment explaining the plot thing where the expected data is constructed.

Library/ViewModels/ManagePledgeViewModelTests.swift
ifosli: Yeah, I don't know how plot handles authentication needed, but I think it does - this is basically just me moving the plot tests from the payment method view tests.

Library/ViewModels/ManagePledgeViewModelTests.swift
ifosli: Yeah, this is already tested in the `testConfigurePaymentMethodViewController_showFixPayment_noPaymentMethod` test above. I didn't initially add a snapshot test since this is very much an edge case, but I'll add a snapshot test, too.

PR #2630, opened 2025-09-22
[MBL-2091] Add PPO v2 tier types

ifosli:
# 📲 What

Reorganize how tierTypes are defined, add tierTypes to the PPO fetch, and add new tier types for the backings dash v2 cards.

In order to support v2 of PPO, we need to add the specific tier types we're requesting to the PPO fetch. I decided to do this by centralizing all tiertype logic/definitions in one file, including parsing the tierType we get from the fetch, and helpers for the graphQL types that should be included in v1 and v2.

Because the cards aren't rendered yet, this has no visual effect. I did briefly check that the raw query response does include new card types when PPOv2 flag is on.

# 👀 See

[Jira](https://kickstarter.atlassian.net/browse/MBL-2091)

# ✅ Acceptance criteria

- [x] PPO v1 looks the same
- [x] Inspecting the query response for v2 shows additional cards

### DIFF
--- a/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
+++ b/Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
@@ -29,8 +29,8 @@ public enum PPOTierType: String {
     ]
   }
 
-  public static func projectAlertGraphQLTypes() -> [GraphQLEnum<PledgeProjectsOverviewSort>] {
-    return self.projectAlertTypes().map { $0.toGraphQLEnum() }
+  public static func projectAlertGraphQLTypes() -> [PledgeProjectsOverviewSort] {
+    return self.projectAlertTypes().map { $0.toPledgeProjectsOverviewSort() }
   }
 
   public static func fundedProjectTypes() -> [PPOTierType] {
@@ -43,12 +43,8 @@ public enum PPOTierType: String {
     ]
   }
 
-  public static func fundedProjectGraphQLTypes() -> [GraphQLEnum<PledgeProjectsOverviewSort>] {
-    return self.fundedProjectTypes().map { $0.toGraphQLEnum() }
-  }
-
-  public func toGraphQLEnum() -> GraphQLEnum<PledgeProjectsOverviewSort> {
-    return GraphQLEnum.case(self.toPledgeProjectsOverviewSort())
+  public static func fundedProjectGraphQLTypes() -> [PledgeProjectsOverviewSort] {
+    return self.fundedProjectTypes().map { $0.toPledgeProjectsOverviewSort() }
   }
 
   public func toPledgeProjectsOverviewSort() -> PledgeProjectsOverviewSort {
--- a/KsApi/Service.swift
+++ b/KsApi/Service.swift
@@ -994,7 +994,7 @@ public struct Service: ServiceType {
   }
 
   public func fetchPledgedProjects(
-    tierTypes: [GraphQLEnum<PledgeProjectsOverviewSort>],
+    tierTypes: [PledgeProjectsOverviewSort],
     cursor: String? = nil,
     limit: Int? = nil
   ) -> AnyPublisher<GraphAPI.FetchPledgedProjectsQuery.Data, ErrorEnvelope> {
@@ -1002,7 +1002,7 @@ public struct Service: ServiceType {
       .fetch(query: GraphAPI.FetchPledgedProjectsQuery(
         first: GraphQLNullable.someOrNil(limit),
         after: GraphQLNullable.someOrNil(cursor),
-        tierTypes: GraphQLNullable.some(tierTypes)
+        tierTypes: GraphQLNullable.some(tierTypes.map { GraphQLEnum($0) })
       ))
       .eraseToAnyPublisher()
   }
--- a/KsApi/MockService.swift
+++ b/KsApi/MockService.swift
@@ -1925,7 +1925,7 @@
     }
 
     func fetchPledgedProjects(
-      tierTypes _: [GraphQLEnum<PledgeProjectsOverviewSort>],
+      tierTypes _: [PledgeProjectsOverviewSort],
       cursor _: String?,
       limit _: Int?
     ) -> AnyPublisher<GraphAPI.FetchPledgedProjectsQuery.Data, ErrorEnvelope> {
--- a/KsApi/ServiceType.swift
+++ b/KsApi/ServiceType.swift
@@ -441,7 +441,7 @@ public protocol ServiceType {
 
   /// Fetch data for the pledged projects overview.
   func fetchPledgedProjects(
-    tierTypes: [GraphQLEnum<PledgeProjectsOverviewSort>],
+    tierTypes: [PledgeProjectsOverviewSort],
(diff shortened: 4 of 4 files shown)


graphql/graphql-schema.json
ifosli, line 1: @amy-at-kickstarter any changes here were automatic but 70k+ lines deleted makes me nervous. Anything I should do to manually check on these changes?

amy-at-kickstarter: A couple thoughts:

1. You might be the first person to have re-generated the schema. The old schema was pretty-printed JSON, and the new one is one big single-line blob. My guess is that's why it's showing this crazy line count.
2. You could spot-check the generated GraphAPI objects and make sure nothing crazy happened, just to make sure the schema is still good.
3. Perhaps this should be marked in `.gitattributes` as `linguist-generated` instead of `linguist-vendored`? I wouldn't expect this to affect the line count!

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import GraphAPI
+import Kingfisher
+import KsApi
+import Library
+
+public enum PPOTierType: String {
+  // Source of truth for enum values can be found at
+  // https://github.com/kickstarter/kickstarter/blob/main/app/models/open_search/pledged_projects_overview.rb
```
amy-at-kickstarter, line 9: Feels a little funny to me to reference a private repo from a public one in this way - how about
```swift
// Source of truth for enum values can be found in OpenSearch::PledgedProjectsOverview
```

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import GraphAPI
+import Kingfisher
+import KsApi
+import Library
+
+public enum PPOTierType: String {
```
amy-at-kickstarter, line 7: I like the cleanup of making this into a proper enum type.

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
```diff
@@ -0,0 +1,69 @@
...
+      .authenticateCard,
+      .pledgeManagement
+    ]
+  }
+
+  public static func projectAlertGraphQLTypes() -> [GraphQLEnum<PledgeProjectsOverviewSort>] {
+    return self.projectAlertTypes().map { $0.toGraphQLEnum() }
+  }
+
+  public static func backingsDashV2Types() -> [PPOTierType] {
```
amy-at-kickstarter: Nit: slightly confusing to me that it's "project alert types" vs "v2 types". Should it be "alert types" vs "all types"? Or maybe "v1" vs "v2" types? I know what you mean now, but I think this will be easy to forget in 6 months.

KsApi/Service.swift
```diff
@@ -994,13 +994,15 @@ public struct Service: ServiceType {
   }
 
   public func fetchPledgedProjects(
+    tierTypes: [GraphQLEnum<PledgeProjectsOverviewSort>],
```
amy-at-kickstarter: Nit: I don't think we expose `GraphQLEnum`, or more generally, GraphQL API types, in the interface of `Service`. It would be a little cleaner to just take `[PPOTierType]` as the input and transform it internally.

nativeksr:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
PPOProjectCardModel+Parsing.swift | 14 | Initializer body should span 50 lines or less excluding comments and whitespace: currently spans 80 lines (function_body_length)
PPOViewModel.swift | 447 | File should contain 400 lines or less: currently contains 447 (file_length)
PPOViewModel.swift | 92 | Class body should span 250 lines or less excluding comments and whitespace: currently spans 251 lines (type_body_length)
ServiceType.swift | 659 | File should contain 400 lines or less: currently contains 659 (file_length)
#### Errors

File | Line | Reason |
| --- | ----- | ----- |
PPOViewModel.swift | 93 | Initializer body should span 100 lines or less excluding comments and whitespace: currently spans 178 lines (function_body_length)
MockService.swift | 2078 | File should contain 1000 lines or less: currently contains 2078 (file_length)
MockService.swift | 9 | Struct body should span 350 lines or less excluding comments and whitespace: currently spans 1533 lines (type_body_length)
Service.swift | 1009 | File should contain 1000 lines or less: currently contains 1009 (file_length)
Service.swift | 19 | Struct body should span 350 lines or less excluding comments and whitespace: currently spans 823 lines (type_body_length)


</p>

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
ifosli: Hmm, that would also work, and I'm okay with changing it if you feel strongly about it, but I definitely prefer the link. It's so convenient to use. (I was very pleased to see I'd already added it last time I touched this code; it made my work easier in this pr - this link isn't newly added; the comment just moved.) You have a point that it's a bit weird to have private links in a public repo, but I think their value outweighs that (both for linking to the main repo, and for referencing jira tickets). Let me know if you feel strongly about this, and I'll change it.

graphql/graphql-schema.json
ifosli: That makes sense, for the line count. I tried switching to `linguist-generated` and it did not affect the line count, so I'm planning to change it back and just let this line count be weird.

Just to confirm, the info represented in `graphql-schema.json` is also displayed in our `GraphAPI` target, in more readable format, right? Because I looked at the `.graphql.swift` files that were also generated, and I didn't see any concerning changes there, so if I'm understanding our new apollo correctly, then I think I'm no longer concerned.

Kickstarter-iOS/Features/PledgedProjectsOverview/PPOTierTypes.swift
ifosli: Hmm, I'll update this to `fundedProjectTypes` instead of `backingsDashV2Types`. That's more consistent with the project alert naming (which I want to keep bc I think I'll also need it to count the number of alerts), and it lets me pretty easily use the same naming scheme for v3/v4 cards, once we add them.

KsApi/Service.swift
ifosli: Updated to use `[PledgeProjectsOverviewSort]` instead.

graphql/graphql-schema.json
amy-at-kickstarter: Correct - `graphql-schema` is a JSON file, downloaded from Staging, which describes _everything_ that our GraphQL endpoints can do. The generated `.graphql.swift` files represent the mutations, fragments, and queries that we actually use in the app - so they are generated from all our `.graphql` files combined with `graphql-schema`.

